PR #57, opened 2016-08-29
BOOKKEEPER-945 Add counters to track the activity of auditor and repl…

rithin-shetty:
…ication workers

Once we enable auto recovery, auditor and replication workers start their activity.
Today there is no way to monitor it using counters. This change introduces the
following counters to track various activities of auditor and replication workers like:

- Time taken by auditor to build the bookie->ledger list
- No. of under replicated ledgers detected
- Time taken by auditor to publish the under replicated ledger list
- Time taken by auditor to check all the ledgers in the cluster
- No. of ledgers replicated by each replication worker
- No. of entries and bytes of data read and written by each replication worker
- Auditor can also report the distribution of ledgers within the cluster: how many bookies own a piece of ledger, etc.


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperAdmin.java
```diff
@@ -148,13 +149,13 @@ public BookKeeperAdmin(ClientConfiguration conf) throws IOException, Interrupted
      * @param bkc
      *            - bookkeeper instance
      */
-    public BookKeeperAdmin(final BookKeeper bkc) {
+    public BookKeeperAdmin(final BookKeeper bkc, StatsLogger statsLogger) {
```
eolivelli: Can we leave the one-parameter constructor too ? I use BookKeeperAdmin to manage Bookies from Java code and this change will break binary compatibilty

rithin-shetty: Sure, I'll leave the old constructor. I should have noticed that it was defined public...

rithin-shetty:
The following tests have failed. These pass on my machine. I think these are flappers:

testBookie(org.apache.bookkeeper.benchmark.TestBenchmark)  Time elapsed: 63.403 sec  <<< ERROR! java.lang.Exception: test timed out after 60000 milliseconds
testReadThroughputLatency(org.apache.bookkeeper.benchmark.TestBenchmark)  Time elapsed: 64.472 sec  <<< ERROR! java.lang.Exception: test timed out after 60000 milliseconds

sijie:
The change looks good to me. +1

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperAdmin.java
sijie: @rithin-shetty do you want to address @eolivelli 's comment? so I can merge this change.

rithin-shetty: I did make the change @eolivelli asked. He approved my changes after that...

sijie: ah. nvm. I missed the old constructor. 

I am going to merge this change. thanks.

PR #58, opened 2016-09-14
BOOKKEEPER-946: Provide an option to delay auto recovery of lost bookies

rithin-shetty:
If auto recovery is enabled, and a bookie goes down for upgrade or even if it looses zk connection
intermittently, the auditor detects it as a lost bookie and starts under replication detection and
the replication workers on other bookie nodes start replicating the under replicated ledgers. All
of this stops once the bookie comes up but by then a few ledgers would get replicated. Given the
fact that we have multiple copies of data, it is probably not necessary to start the recovery as
soon as a bookie goes down. We can wait for an hour or so and then start recovery. This should
cover cases like planned upgrade, intermittent network connectivity loss, etc.

This change:
    1) Provides a bookie option 'lostBookieRecoveryDelay' in secs, which when set to a non zero value,
       will delay the start of recovery by that number of seconds. By default, this option is set to 0;
       which means the audit is not delayed.
    2) If another bookie were to go down in this interval, the recovery is immediately started and the
       one scheduled for future is canceled.
    3) Adds counters to track how many audits were delayed(#1) and how many scheduled audits were
       canceled due to multiple bookie failures(#2).
    4) Three junit tests to verify the new feature.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
@@ -98,6 +98,10 @@ public class Auditor implements BookiesListener {
     private final Counter numLedgersChecked;
     private final OpStatsLogger numFragmentsPerLedger;
     private final OpStatsLogger numBookiesPerLedger;
+    private final Counter numBookieAuditsDelayed;
+    private final Counter numDelayedBookieAuditsCancelled;
+    private volatile Future<?> auditTask;
+    private Set<String> bookiesToBeAudited = Sets.newHashSet();
 
     public Auditor(final String bookieIdentifier, ServerConfiguration conf,
                    ZooKeeper zkc, StatsLogger statsLogger) throws UnavailableException {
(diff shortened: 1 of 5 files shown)


bookkeeper-server/src/test/java/org/apache/bookkeeper/replication/AuditorLedgerCheckerTest.java
```diff
@@ -321,6 +325,120 @@ public void testReadOnlyBookieShutdown() throws Exception {
...
+        // we have enabled periodic bookie check
+        _testDelayedAuditOfLostBookies();
+    }
+
+    /**
+     * Test audit of bookies is delayed when one bookie is down. But when
+     * another one goes down, the audit is started immediately.
+     */
+    @Test(timeout=60000)
+    public void testDelayedAuditWithMultipleBookieFailures() throws Exception {
```
eolivelli: This testcase on my laptop fails sometimes, I have attached the logs to the JIRA

rithin-shetty: Thanks @eolivelli for running the tests. I'm having a bit of a trouble reproducing the issue. I ran this one tests in a loop 100 times and it passed every single time! From your logs it appears that there was a zk timeout. That may have contributed to the failure.

rithin-shetty: I added some delay before starting testDelayedAuditWithMultipleBookieFailures() so that the initial bookie check finishes. If the system is busy the initial check can be very slow and could end up doing the audit of lost bookies. @eolivelli, hopefully this should address the test failure on your setup.

rithin-shetty: Thanks @eolivelli for trying it out. I had to use my old desktop to reproduce the issue :)

sboobna:
If I understand correctly, for this feature to help during upgrades, each bookie should be upgraded in a separate ```lostBookieRecoveryDelay``` window. Else, when a 2nd bookie is taken down to upgrade, it will anyways fall back to the current behavior.

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
```diff
@@ -196,20 +204,39 @@ public void run() {
...
-                            auditBookies();
+                            if (conf.getLostBookieRecoveryDelay() == 0 ||
+                                lostBookies.size() > 1 ||
+                                auditTask != null) {
+                                // 1) If more than one bookie is down, start the audit immediately;
+                                // 2) If we had scheduled an audit earlier for a lost bookie and in
+                                //    the meantime another bookie goes down, let us not delay recovery;
+                                //    we cancel the previously scheduled audit before starting the new one
+                                if (auditTask != null) {
+                                    if (auditTask.cancel(false)) {
```
sijie: if the auditTask can not be cancelled, you will be starting another audit task, no?

rithin-shetty: The documentation for cancel return code says  "[it returns] false if the task could not be cancelled, typically because it has already completed normally; true otherwise"

So I decided to schedule another one just to be safe. Running it twice shouldn't cause any issues though. Let me know if you want me to not schedule one if cancel fails.

sijie:
@rithin-shetty can you answer @sboobna 's question?

rithin-shetty:
@sijie I answered @sboobna's question in the Jira(https://issues.apache.org/jira/browse/BOOKKEEPER-946) itself. There doesn't seem to be a way to reply to his comment here in git; it doesn't give a reply button. He did find an issue with the code. It wasn't handling the rolling upgrades properly. I've fixed it now and pushed it; the latest diffs should show you the new code.

sboobna:
LGTM 👍

jvrao:
+1 LGTM

PR #60, opened 2016-10-03
BOOKKEEPER-924 addEntry() is susceptible to spurious wakeups

eolivelli:
Use Java8 CompletableFuture instead of SyncCounter

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeper.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeper.java
@@ -60,6 +60,8 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
 
 /**
  * BookKeeper client. We assume there is one single writer to a ledger at any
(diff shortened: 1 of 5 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeper.java
```diff
@@ -627,27 +618,15 @@ public LedgerHandle createLedgerAdv(int ensSize, int writeQuorumSize, int ackQuo
...
                              new SyncCreateCallback(), counter, customMetadata);
 
-        /*
-         * Wait
-         */
-        counter.block(0);
-        if (counter.getrc() != BKException.Code.OK) {
-            LOG.error("Error while creating ledger : {}", counter.getrc());
-            throw BKException.create(counter.getrc());
-        } else if (counter.getLh() == null) {
```
sijie: There is a validation about the returned value. It seems your change dropped the behavior.

sijie: @eolivelli in theory, there should be no null ledgerhandle returned with status code == OK. the validation was there to prevent unexpected condition (e.g. bugs). if possible, I'd suggest not changing the behavior when refactoring this piece of code. does that make sense to you?

eolivelli: @sijie OK. I will update the PR soon

sijie:
@eolivelli the change looks good to me +1

PR #61, opened 2016-10-14
BOOKKEEPER-952: Fix RegionAwareEnsemblePlacementPolicy

yzang:
allocate bookies evenly as much as possible across all regions

https://issues.apache.org/jira/browse/BOOKKEEPER-952

RB_ID=880463

PR #62, opened 2016-10-19
BOOKKEEPER-903: metaformat command fix

reddycharan:
metaformat command should delete underreplicated ledger
znodes and lock nodes


sijie:
The jenkins failure was due to the issue at #61 (not related to this change)

sijie:
LGTM +1

PR #63, opened 2016-10-19
BOOKKEEPER-955: in listLedgers method currentRange variable has to be…

reddycharan:
… updated to the next iterator

in BookKeeperAdmin listLedgers method currentRange variable is not getting updated to next iterator
when it has run out of elements


sijie:
The failure test case was due to #61 (not related to this change)

jvrao:
+1 change looks good.

sijie:
LGTM +1

PR #65, opened 2016-10-19
BOOKKEEPER-957: LedgerHandleAdv fixes

reddycharan:
- making addToLength call synchronized
- adding op to pendingAddOps after doing sanity check


sijie:
LGTM +1

PR #67, opened 2016-10-20
BOOKKEEPER-959 ClientAuthProvider and BookieAuthProvider Public API used Protobuf Shaded classes

eolivelli:
This fix removes the explicit usage of protobuf from ClientAuthProvider and BookieAuthProvider API, since protobuf library is shaded and relocated on the distributed public version of BookKeeper


merlimat:
The change looks OK to me. It's much easier to get around compared to the Protobuf extension. I guess we (Yahoo) are the only current users of this framework but I prefer to break the compatibility right now and have a better solution. 

The only point I'd like to make sure is how to transition a running cluster from no-auth to auth enabled and back. Eg. if a client is sending auth commands, will the bookie reply OK if auth is disabled?

eolivelli:
Thanks @merlimat for the review.
Do you think we can remove the use of the "extensions"  and use a field for the payload ?

At least for the transition from no-auth to auth I can submit a fix on this PR or is it better to create a new JIRA ?

my idea is to make the Bookie answer a standard message "authentication-disabled" to clients which is trying to make any kind of auth, so:
- on the server side the connection goes to "authenticated" state
- on the client side falls back to a "authenticated" state
this standard message will be processed from BookKeeper client system code and not passed to the auth plugin

we have to implement a "require.authentication" flag on the Bookie side, to let the admin set it to false, in order to deal with legacy clients (and Bookies) which do not perform authentication.
On a rolling restart we upgrade Bookie to the new version with require.authentication=false, then upgrade all the clients to the new version and/or auth plugin, then we restart again each Bookie ith require.authentication=true

this simple implementation will allow to switch from no-auth to auth, but not from auth-1-plugin to auth-2-plugin

does it sound good  ?

merlimat:
As a sequence to enable authentication I would say something like: 
 1. Upgrade to BK-4.5 in clients and bookies, auth disabled
 2. Enable auth in BK clients (bookies should let clients connect)
 3. Configure bookies to require auth

eolivelli:
@merlimat I'm going to update the PR this way:
- remove the extension from AuthMessage, and make 'payload' a fixed, required, field
- rename AuthMessageData to AuthToken
- fix the auth required/auth disabled flow (I will update https://cwiki.apache.org/confluence/display/BOOKKEEPER/BP-3+-+Security+support with specs)

is the change on protobuf OK for you, as we know you are the only user of such message ?
is there any other way to deal with compatibility for you ?

merlimat:
Sounds good. 

> is the change on protobuf OK for you, as we know you are the only user of such message ?

Yes

> is there any other way to deal with compatibility for you ?

If the bookies with auth turned off will accept a client with auth turned on... then we can manage the transition (eg: turn temporarily off the auth and deploy the new code). But the same requirement is needed to enable auth on an existing cluster.

eolivelli:
I have updated the PR with the close() method.
I think it is aligned with the specs and the discussion

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieSideConnectionPeer.java
```diff
@@ -0,0 +1,63 @@
...
+package org.apache.bookkeeper.bookie;
+
+import java.net.SocketAddress;
+import java.util.Collection;
+import org.apache.bookkeeper.auth.BookKeeperPrincipal;
+
+/**
+ * Represents the connection to a BookKeeper client, from the Bookie side 
+ */
+public interface BookieSideConnectionPeer {
```
sijie: Two comments:

- it seems that BookieSideConnectionPeer has exactly same interface with ClientSideConnectionPeer. Can we have one common parent class.
- Can we remove 'Side' from the name?

How about have the common interface 'ConnectionPeer' and have 'ServerConnectionPeer' and 'ClientConnectionPeer' extend from the parent interface?

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/AuthHandler.java
```diff
@@ -200,6 +229,7 @@ public void operationComplete(int rc, AuthMessage newam) {
             public void operationComplete(int rc, Void v) {
                 if (rc == BKException.Code.OK) {
                     authenticated = true;
+                    LOG.error("Authentication success on server side");
```
sijie: 'info'?

eolivelli:
Thanks @sijie for the review.
I have applied your suggestions.

sijie:
@eolivelli LGTM +1.

@merlimat  can you take a look too?

sijie:
ping @merlimat

sijie:
ping @merlimat

sijie:
@merlimat do you mind taking a look one more time before I merge this one?

zzsongsg:
+1 merging this pull request.

sijie:
merged this pull request in 13d668f. thank you @eolivelli

PR #68, opened 2016-10-31
BOOKKEEPER-912 Allow EnsemblePlacementPolicy to choose bookies using ledger custom data (multitenancy support)

eolivelli:


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LocalBookieEnsemblePlacementPolicy.java
```diff
@@ -74,7 +74,7 @@ public void uninitalize() {
     }
 
     @Override
-    public BookieSocketAddress replaceBookie(int ensembleSize, int writeQuorumSize, int ackQuorumSize, Collection<BookieSocketAddress> currentEnsemble, BookieSocketAddress bookieToReplace, Set<BookieSocketAddress> excludeBookies) throws BKNotEnoughBookiesException {
+    public BookieSocketAddress replaceBookie(int ensembleSize, int writeQuorumSize, int ackQuorumSize, java.util.Map<String, byte[]> customMetadata, Collection<BookieSocketAddress> currentEnsemble, BookieSocketAddress bookieToReplace, Set<BookieSocketAddress> excludeBookies) throws BKNotEnoughBookiesException {
```
sijie: break this line into multiple lines. it seems to be too long.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LocalBookieEnsemblePlacementPolicy.java
```diff
@@ -89,8 +89,7 @@ public BookieSocketAddress replaceBookie(int ensembleSize, int writeQuorumSize,
     }
 
     @Override
-    public ArrayList<BookieSocketAddress> newEnsemble(int ensembleSize, int writeQuorumSize, int ackQuorumSize,
-            Set<BookieSocketAddress> excludeBookies) throws BKNotEnoughBookiesException {
+    public ArrayList<BookieSocketAddress> newEnsemble(int ensembleSize, int writeQuorumSize, int ackQuorumSize, java.util.Map<String, byte[]> customMetadata, Set<BookieSocketAddress> excludeBookies) throws BKNotEnoughBookiesException {
```
sijie: break into new lines?

eolivelli:
@sijie new lines added. thank you for the review

sijie:
@eolivelli LGTM +1

eolivelli:
The idea is that you can.implemement a custom policy which takes into account the custometadata. I am going to use this feature at least for these two cases:
1) for.multitenancy, that is that a ledger is created for a tenant/user and the policy will choose the best bookies according to the user' SLA. The internals of this policy is custom and depends on the project
2) in order to define different behaviors depending on the type of ledger. I have many different applications/services that share the same BookKeeper cluster and only one single policy can't work for every usecase. Using metadata you can implement a custom policy which forwards request to other policies depending on the specs written in metadata

PR #71, opened 2016-11-02
BOOKKEEPER-963: Allow to use multiple journals in bookie

merlimat:
By configuring multiple journals, we can take advantage of the IO of multiple
disks to increase the write throughput of a single bookie.

Each journal will have its own journal and sync threads and writes will be
assigned to a particular journal by hashing on the ledger id.

In addition to using multiple physical disks, there can improvements even by
using multiple journal on a single SSD device, because these disks can handle
well multiple concurrent writes in different blocks of the disk.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
@@ -84,6 +84,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.Lists;
 
 import static org.apache.bookkeeper.bookie.BookKeeperServerStats.BOOKIE_ADD_ENTRY;
 import static org.apache.bookkeeper.bookie.BookKeeperServerStats.BOOKIE_ADD_ENTRY_BYTES;
@@ -105,14 +106,14 @@ public class Bookie extends BookieCriticalThread {
 
     private final static Logger LOG = LoggerFactory.getLogger(Bookie.class);
 
-    final File journalDirectory;
+    final List<File> journalDirectories;
     final ServerConfiguration conf;
 
     final SyncThread syncThread;
     final LedgerManagerFactory ledgerManagerFactory;
     final LedgerManager ledgerManager;
     final LedgerStorage ledgerStorage;
-    final Journal journal;
+    final List<Journal> journals;
 
     final HandleFactory handles;
 
@@ -321,28 +322,37 @@ public class Bookie extends BookieCriticalThread {
             allLedgerDirs.addAll(indexDirsManager.getAllLedgerDirs());
         }
         if (zk == null) { // exists only for testing, just make sure directories are correct
-            checkDirectoryStructure(journalDirectory);
+            for (File journalDirectory : journalDirectories) {
+                checkDirectoryStructure(journalDirectory);
+            }
             for (File dir : allLedgerDirs) {
-                    checkDirectoryStructure(dir);
+                checkDirectoryStructure(dir);
             }
             return;
         }
         try {
             boolean newEnv = false;
             List<File> missedCookieDirs = new ArrayList<File>();
-            Cookie journalCookie = null;
-            // try to read cookie from journal directory.
-            try {
-                journalCookie = Cookie.readFromDirectory(journalDirectory);
-                if (journalCookie.isBookieHostCreatedFromIp()) {
-                    conf.setUseHostNameAsBookieID(false);
-                } else {
-                    conf.setUseHostNameAsBookieID(true);
+            List<Cookie> journalCookies = Lists.newArrayList();
+
+            // try to read cookie from journal directory
+            for (File journalDirectory : journalDirectories) {
+                try {
+                    checkDirectoryStructure(journalDirectory);
+                    Cookie journalCookie = Cookie.readFromDirectory(journalDirectory);
+                    journalCookies.add(journalCookie);
+
+                    if (journalCookie.isBookieHostCreatedFromIp()) {
+                        conf.setUseHostNameAsBookieID(false);
+                    } else {
+                        conf.setUseHostNameAsBookieID(true);
+                    }
+                } catch (FileNotFoundException fnf) {
+                    newEnv = true;
+                    missedCookieDirs.add(journalDirectory);
                 }
-            } catch (FileNotFoundException fnf) {
-                newEnv = true;
-                missedCookieDirs.add(journalDirectory);
             }
+
             String instanceId = getInstanceId(zk);
             Cookie.Builder builder = Cookie.generateCookie(conf);
             if (null != instanceId) {
@@ -357,10 +367,15 @@ public class Bookie extends BookieCriticalThread {
                 // 1) new environment or
                 // 2) done only metadata format and started bookie server.
             }
-            checkDirectoryStructure(journalDirectory);
 
-            if(!newEnv){
-                journalCookie.verify(masterCookie);
+            for (File journalDirectory : journalDirectories) {
+                checkDirectoryStructure(journalDirectory);
+            }
+
+            if (!newEnv){
+                for (Cookie journalCookie : journalCookies) {
+                    journalCookie.verify(masterCookie);
+                }
             }
             for (File dir : allLedgerDirs) {
                 checkDirectoryStructure(dir);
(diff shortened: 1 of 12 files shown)


sijie:
@merlimat out of curiosity, do you have any numbers to share about using multiple journal disks?

/cc @franckcuny

merlimat:
@sijie Unfortunately not. We did the perf test long time back and I haven't repeated it since.

sijie:
@merlimat do we need to have a tool to upgrade the cookie file?

Can you rebase this change?

jvrao:
Crap I lost all my review comments; Will start from the beginning again.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -321,28 +322,37 @@ private void checkEnvironment(ZooKeeper zk) throws BookieException, IOException
...
+            for (File journalDirectory : journalDirectories) {
+                try {
+                    checkDirectoryStructure(journalDirectory);
+                    Cookie journalCookie = Cookie.readFromDirectory(journalDirectory);
+                    journalCookies.add(journalCookie);
+
+                    if (journalCookie.isBookieHostCreatedFromIp()) {
+                        conf.setUseHostNameAsBookieID(false);
+                    } else {
+                        conf.setUseHostNameAsBookieID(true);
```
jvrao: Why are you updating conf in the loop?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -357,10 +367,15 @@ private void checkEnvironment(ZooKeeper zk) throws BookieException, IOException
                 // 1) new environment or
                 // 2) done only metadata format and started bookie server.
             }
-            checkDirectoryStructure(journalDirectory);
 
-            if(!newEnv){
-                journalCookie.verify(masterCookie);
+            for (File journalDirectory : journalDirectories) {
+                checkDirectoryStructure(journalDirectory);
+            }
```
jvrao: You have called checkDirectoryStructure() above in the loop. Why again?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -500,16 +522,28 @@ public Bookie(ServerConfiguration conf, StatsLogger statsLogger)
...
+        if (journalDirectories.size() == 1) {
+            journals = Lists.newArrayList(new Journal(journalDirectories.get(0), conf, ledgerDirsManager, statsLogger
+                    .scope(JOURNAL_SCOPE)));
+        } else {
+            journals = Lists.newArrayList();
+            for (int idx = 0; idx < journalDirectories.size(); idx++) {
+                journals.add(new Journal(journalDirectories.get(idx), conf, ledgerDirsManager, statsLogger
+                        .scope(JOURNAL_SCOPE + "-" + idx)));
+            }
+        }
```
jvrao: Why are you handling ==1 case differently? Is it just to define the scope? I think even if we have one having scope with idx(0) makes more sense;

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -1044,11 +1085,17 @@ public boolean isRunning() {
...
+            for (Journal journal : journals) {
+                journal.start();
+            }
+
+            // wait until journals quit
+            for (Journal journal : journals) {
+                journal.join();
+            }
+
+            LOG.info("Journal threads quits.");
```
jvrao: I think this should be LOG.debug or even remove it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -1343,38 +1393,39 @@ synchronized public void waitZero() throws InterruptedException {
...
+                        // data.
+                        if (force) {
+                            confirm = true;
+                        } else {
+                            confirm = false;
+                        }
                     } else {
-                        confirm = false;
+                        confirm = alreadyConfirmed || IOUtils.confirmPrompt("Are you sure to format Bookie data..?");
+                        alreadyConfirmed = true;
```
jvrao: This logic is confusing. All you need to do is, move confirm outside the for loop and in the else case, if (!confirmed) IOUtils... will do the job. You don't need alreadyConfirmed flag.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/UpdateCookieCmdTest.java
```diff
@@ -192,9 +192,12 @@ private void updateCookie(String option, String optionVal, boolean useHostNameAs
...
-        File journalDir = Bookie.getCurrentDirectory(conf.getJournalDir());
-        Cookie jCookie = Cookie.readFromDirectory(journalDir);
-        jCookie.verify(cookie);
+        for (File journalDir : conf.getJournalDirs()) {
+            journalDir = Bookie.getCurrentDirectory(journalDir);
+            Cookie jCookie = Cookie.readFromDirectory(journalDir);
+            jCookie.verify(cookie);
+        }
+
         File[] ledgerDir = Bookie.getCurrentDirectories(conf.getLedgerDirs());
```
jvrao: Looks like tests for this feature are missing.
- At lest few of the journal tests needs to be implemented with multiple journals 
- Looks like ; if there are multiple journal stanzas with NULL, that many journal files are created in the current dir; not sure if that is intended; also no checks for uniqueness of each journal dir

reddycharan:
new BookieShell command - ExpandStorageCmd (expandstorage) has been added by @rithin-shetty , it should also be updated as part of this change

BOOKKEEPER-948: Provide an option to add more ledger/index directories to a bookie
https://github.com/apache/bookkeeper/commit/66515e74d3fb03d21cad4eccb446a90ab7e94597

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/CheckpointSourceList.java
```diff
@@ -0,0 +1,97 @@
...
+            if (o == Checkpoint.MAX) {
+                return -1;
+            } else if (o == Checkpoint.MIN) {
+                return 1;
+            }
+
+            checkArgument(o instanceof CheckpointList);
+            CheckpointList other = (CheckpointList) o;
+            if (checkpoints.size() != other.checkpoints.size()) {
+                return Integer.compare(checkpoints.size(), other.checkpoints.size());
```
reddycharan: in which case, this (different sizes) can happen

govind-menon:
I'll take over the JIRA and PR (unless @merlimat wants to continue).

merlimat:
@govind-menon Please go ahead

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
merlimat: Yes, but changing the scope name might break existing metrics tools that expect the name without the index

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
merlimat: I don't remember exactly, though, I was just replicating the same steps being done on the single journal directory

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
merlimat: I agree it would much cleaner, but it would be slightly different compared to the existing logic that only asks for confirmation if the directory exists. I wouldn't mind changing that though.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/CheckpointSourceList.java
```diff
@@ -0,0 +1,97 @@
...
+
+        public CheckpointList(CheckpointSourceList source) {
+            this.source = source;
+            this.checkpoints = Lists.newArrayListWithCapacity(source.checkpointSourcesList.size());
+            for (CheckpointSource checkpointSource : source.checkpointSourcesList) {
+                checkpoints.add(checkpointSource.newCheckpoint());
+            }
+        }
+
+        private void checkpointComplete(boolean compact) throws IOException {
```
merlimat: The `compact` was already present in the `Checkpoint` interface. According to the javadoc: 

```
     * @param compact
     *          Flag to compact old checkpoints.
```

The point of `CheckPointSourcesList` is to act as a meta-checkpoint source, delegating to all the individual sources, so we just need to forward the `compact` parameter that we get passed on.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/CheckpointSourceList.java
merlimat: They shouldn't differ, in the case of journals, since the checkpoint source will always be composed by the same N journal sources. I agree, though, that the code here should be clearer and more explicit.

sijie:
#115 is taking over this PR.

mind closing this PR?

PR #72, opened 2016-11-02
BOOKKEEPER-964: Add concurrent maps and sets for primitive types

merlimat:
In BookKeeper there are many instances of maps and sets that use ledger id
and entry ids as keys or values. JDK concurrent collections have the overhead
of boxing all the primitive values into objects (eg: long --> Long) that would
need to be allocated from the heap. In addition to that, JDK map implementations
are closed hash tables and they will require at least one more allocation to hold
the linked-list/tree node.

There are already available libraries that offer primitive collections with
zero-allocation, but none of these support concurrent maps/sets.

We have added a handful of specializations, all based on the same implementation
idea. We have a hash table which is broken down into multiple sections. Each
sections, on its own, is an open hash table with linear probing, protected by
a stamped lock.

All insertions, lookups and iterations on these collections are allocation free.

```
ConcurrentLongHashMap: Map<long, Object>
ConcurrentLongHashSet: Set<long>
ConcurrentLongLongHashMap: Map<long, long>
ConcurrentLongLongPairHashMap: Map< Pair<long, long>, Pair<long, long> >
ConcurrentOpenHashMap: Map<Object, Object>
ConcurrentOpenHashSet: Set<Object>
```

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/util/collections/ConcurrentLongHashMap.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/util/collections/ConcurrentLongHashMap.java
@@ -1,0 +1,494 @@
+/*
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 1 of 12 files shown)


merlimat:
Note: this PR and Jira are only relative to adding the maps implementations. I'll submit later more changes that depend on these.

sijie:
@merlimat this sounds interesting. I assumed this is for performance (especially for memory allocation). Do you have any idea how much memory that it can help reduce here?

merlimat:
My primary concern here was about the number of allocations, especially of allocation that could be slightly long-lived (eg: item stays in the map for few seconds), that are more likely to be pushed to old-gen whenever a young GC cycle is performed.

For example in the `Map<Long, Long>` case you would have 3 allocations per-insertion (long boxing + linked list node to hold key and value). 

Also, every `get()`, `containsKey()` operation would need 1 allocation to box the long key.

In general Objects in 64bits JVM will have a minimum size of 16 bytes. http://stackoverflow.com/questions/258120/what-is-the-memory-consumption-of-an-object-in-java

PR #73, opened 2016-11-08
BOOKKEEPER-965: Long Polling Part I: Changes in the write path

robindh:
This is the first in the series of changes for enabling long polling between bookkeeper client and the bookkeeper server. The changes were originally implemented in the Twitter fork and these pull request combine multiple commits from Twitter's bookkeeper fork as they include not only the changes made initially but also bug fixes added since. 

The first change captures the changes on the write path (AddEntry). We track the last add confirmed in the FileInfo so that we can trigger actions when the value is updated

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileInfo.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileInfo.java
@@ -80,6 +80,9 @@ class FileInfo {
     private int stateBits;
     private boolean needFlushHeader = false;
 
+    // lac
+    private Long lac = null;
+
     // file access mode
     protected String mode;
 
@@ -88,6 +91,17 @@ class FileInfo {
 
         this.masterKey = masterKey;
         mode = "rw";
+    }
+
+    synchronized Long getLastAddConfirmed() {
+        return lac;
+    }
+
+    synchronized long setLastAddConfirmed(long lac) {
+        if (null == this.lac || this.lac < lac) {
+            this.lac = lac;
+        }
+        return this.lac;
     }
 
     public File getLf() {
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
@@ -1305,6 +1305,11 @@ public class Bookie extends BookieCriticalThread {
         }
     }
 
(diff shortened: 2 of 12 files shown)


sijie:
LGTM +1 

@jvrao this is our first change about the long poll read request. I know you guys are also working on the lac related change. Can you also take a look this change to make sure it will be compatible with your change.

jvrao:
This directly corresponds to what we have been using and have local changes. You are aware of our change. Is this something you worked on recently? if so, I am wondering why to reinvent the wheel instead of using our patch.

sijie:
@jvrao no, this is the old change at Twitter's branch. we are merging these changes back.

sijie:
@jvrao @merlimat can any of you take a look? so we can send the subsequent changes out.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileInfo.java
```diff
@@ -80,6 +80,9 @@
     private int stateBits;
     private boolean needFlushHeader = false;
 
+    // lac
+    private Long lac = null;
```
merlimat: Is this used as an optimization to quickly find the highest LAC sent by the writer?

sijie: yes. it will also be used for the subsequent long poll change. so we can detect the lac changes and satisfy any long poll wait requests.

sijie:
@jvrao if there is no objection from you, I'd like to merge this and move on sending the subsequent requests.

jvrao:
@sijie  Sorry responding after it is pulled in.  Do you know how-much savings you are getting with this? I have changes in similar lines but to keep explicitLac in FileInfo. Read path may be little more complicated. I will watch out for it. In the mean time I will push my changes too.

sijie:
@jvrao what do you mean saving? you mean the long poll change?

jvrao:
@sijie  Saving = perf benefit. 
This whole thing is to cache the LAC in FileInfo rather than reaching out on the disk and read it.
IMO the last entry must be in buffer cache and no need to hit the disk. So I was little surprised to see you went this far to cache the LAC in FileInfo, hence wondering if you have any measurements
on how much perf benefit you got with this change.

sijie:
@jvrao gotcha. we didn't measure that in particularly. the most benefit we have with this is we can have long poll requests satisfied when lac is advanced. FileInfo is the best place to us to implement this feature.

PR #75, opened 2016-11-09
BOOKKEEPER-966: change bookieServer cmdline to make conf-file and option co-exist

jiazhai:
Currently, when using bookieServer cmdline to start a bookie, you will either give it a cofiguration file by "-c booke.conf"; or add some options like "<bookie_port> <zk_servers> <journal_dir> <ledger_dir [ledger_dir]>" in a fix sequential.
It may not satisfy some of the requirement. So changed it to be co-exist for configuration file and options.

By this change, it will first use settings in configuration file; and then use options to overwrite some of the settings, if there are some options provided.

Here is an example after this change:
```
BookieServer -c bookie.conf -z localhost:2181 -m /bookkeeper/ledgers -p 3181 -j /mnt/journal -l "/mnt/ledger1 /mnt/ledger2 /mnt/ledger3”
```
Here, in this command:
-z is for “Zookeeper client instance”;
-m is for "Zookeeper ledgers root path for bookies";
-p is for "bookie service port exported";
-j is for "bookie journal directory";
-l is for "bookie ledgers directories".


bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieServer.java
```diff
@@ -270,6 +272,13 @@ public void run() {
                 "Start Autorecovery service Bookie server");
         bkOpts.addOption("readOnly", false,
                 "Force Start a ReadOnly Bookie server");
+        bkOpts.addOption("z", "zkserver", true, "Zookeeper Server");
+        bkOpts.addOption("m", "zkledgerpath", true, "Zookeeper ledgers root path");
+        bkOpts.addOption("p", "bookieport", true, "bookie port exported");
+        bkOpts.addOption("j", "journal", true, "bookie journal directory");
+        Option ledgerDirs = new Option ("l", "ledgerdirs", true, "bookie ledgers directories");
```
sijie: @jiazhai can you add another option for index dirs?

jiazhai: OK. got it. will add and update this.

jiazhai:
Jenkins error seems not related with this change. Opened tickiet BOOKKEEPER-975 for it.

eolivelli:
Can you consider to add a more general flag to configure any proprty bu name?
Something like -x option=value ?
I think that it can be tricky to implement but it will let us pass any new parameter without code changes.
I wonder if we can also add support for system properties, like -Djava.security.config=myjaas.conf

sijie:
@eolivelli +1 for your idea.

jiazhai:
@eolivelli That idea is great. How about open another new issue to tracking this requirement?

sijie:
@jiazhai I am okay with opening a new issue for the -x change. @eolivelli what do you think?

eolivelli:
@sijie @jiazhai OK.  I will create the new issue for advanced options.
This changes loooks good to me.
+1

PR #77, opened 2016-11-11
BOOKKEEPER-968 Entry log flushes at configurable chunks

dlg99:
With this patch one can configure interval (in bytes) for entry log to flush writes to the disk.

### DIFF
--- a/bookkeeper-server/conf/bk_server.conf
+++ b/bookkeeper-server/conf/bk_server.conf
@@ -270,6 +270,13 @@ zkTimeout=10000
 # The interval is specified in seconds.
 #auditorPeriodicBookieCheckInterval=86400
 
+# Entry log flush interval in bytes.
+# Default is 0. 0 or less disables this feature and effectively flush
+# happens on log rotation.
+# Flushing in smaller chunks but more frequently reduces spikes in disk
+# I/O. Flushing too frequently may also affect performance negatively.
+#flushEntrylogBytes=0
+
 # How long to wait, in seconds, before starting auto recovery of a lost bookie
 #lostBookieRecoveryDelay=0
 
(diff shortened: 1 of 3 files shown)


dlg99:
Failed test seems to be unrelated/flapper

merlimat:
@dlg99 Have you considered the approach of using a rate limiter (eg: guava `RateLimiter`) with a configurable max number of MBytes/s to write when flushing? 
It might be a bit inconvenient to properly tune for one's system, but setting the rate slightly below the disk write capacity should do the trick.

merlimat:
On the other hand, we might even be able to do some auto-tuning: If I was only able to write let's say 200 MB in 1 second without the throttling. Next time I'll try to throttle at 190MB/s to not saturate the disk.

dlg99:
@merlimat limiting write rate won't help. One can write with relatively low write rate but since entry log only gets fileChannel.force() on log rotation we end up with a lot of writes cached in memory. 
Once OS (when cache is too full or because of explicit fileChannel.force()) decide that it is time to flush the data to disk we have no control over the rate of writes. It does it at max possible speed, flushing 2048M on ssd with 400M/sec write throughput takes 5 sec if there are no other writes/reads. 
I.e. in our production env I have configured flush to happen at every 10MB of entry log writes and so far I am seeing a lot less read latency spikes. I'll get some charts to show during meetup.

One can experiment with linux config (dirty_write_bytes / backround_dirty_write_bytes IIRC) but these are OS-wide setting (not per disk) and will affect other writes, i.e. I would not want to decrease these parameters to 10-50M for the rotational disk where we write application logs. Limiting these to the range of hundreds MB does not help much with the specific problem on hands.

jiazhai:
May be we could add items in doc/bookkeeperConfigParams.textile along with this changes.

In this link: 
https://cwiki.apache.org/confluence/display/BOOKKEEPER/Contributing+to+BookKeeper
it said:
New server configuration settings should be added and documented in bookkeeper-server/conf/bk_server.conf, and also documented in doc/bookkeeperConfigParams.textile.

sijie:
+1 on what @jiazhai suggested. we should document all the new configurations.

jvrao:
+1 on the patch.
@sijie we may need a jira item for what @jiazhai suggested.

sijie:
@jvrao yup. another jira is fine. We need some jiras for documenting the settings introduced in 4.5.0 and mark them as blockers for 4.5.0 release.

sijie:
+1 for this change

PR #79, opened 2016-11-19
BOOKKEEPER-970: Bump the zookeeper version to 3.5

sijie:
in DL, we need to leverage the asynchronous version of 'multi' in zookeeper. so this jira is to bump the zookeeper version to 3.5 to support async multi.


eolivelli:
Is 3.5 client compatible with 3.4 servers?

sijie:
@eolivelli yes. 3.5 client is compatible with 3.4 servers.

sijie:
@eolivelli yes. we used the new 3.5 API (e.g. asyncMulti and removeWatches) in DL.

eolivelli:
I have double checked and now a new compile-time dependency appeared: net.java.dev.javacc:javacc:jar:5.0:compile

I have already opened a JIRA for that problem https://issues.apache.org/jira/browse/ZOOKEEPER-2460
mean while can we exclude it from BK pom ?

thanks

sijie:
@eolivelli I excluded the javacc dependency

eolivelli:
@sijie now LGTM +1

PR #81, opened 2016-11-22
BOOKKEEPER-753: Bookie should run garbage collection before startup when all directories became full

sijie:
- run gc before startup if all directories are full
- move some initialization into #init() function
- add some more stats


jvrao:
Started looking into it; We solved this problem in two simple steps.
- Enable force compaction.
- Allow entrylog creation for compaction even if the disk usage is above threshold but disk has space.

sijie:
@jvrao enable force compaction is fine. the change here won't actually conflict with your change though

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookKeeperServerStats.java
```diff
@@ -42,6 +42,7 @@
     public final static String BOOKIE_ADD_ENTRY = "BOOKIE_ADD_ENTRY";
     public final static String BOOKIE_RECOVERY_ADD_ENTRY = "BOOKIE_RECOVERY_ADD_ENTRY";
     public final static String BOOKIE_READ_ENTRY = "BOOKIE_READ_ENTRY";
+    public final static String BOOKIE_READ_LAST_CONFIRMED = "BOOKIE_READ_LAST_CONFIRMED";
```
eolivelli: this new stat is not related to the issue.
for me it is fine anyway, but I wonder if we have to commit it in a separate issue

eolivelli:
@sijie there are some unanswered questions on this pr, could you please answer so that we can push go on with it?
Thanks

eolivelli:
IMHO we should restart the thread about this patch. There are some other ongpoing  works  on compaction.
@sijie @jvrao what do you think?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookKeeperServerStats.java
```diff
@@ -42,6 +42,7 @@
...
```
sijie: yeah. it might be from a different commit. but it is a bit hard for me to separate that out. If you guys are really concerned about this, I will do that effort. Otherwise, I will leave it here.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -304,6 +304,36 @@ public void resumeMinorGC() {
...
     }
 
+
+    /**
+     * gc ledger storage.
+     */
+    void gc() {
+        LOG.info("Starting garbage collecting ledger storage.");
+
+        LOG.info("Garbage collecting deleted ledgers' index files.");
```
sijie: turned to LOG.debug().

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -304,6 +304,36 @@ public void resumeMinorGC() {
...
+        // Extract all of the ledger ID's that comprise all of the entry logs
+        // (except for the current new one which is still being written to).
+        entryLogMetaMap = extractMetaFromEntryLogs(entryLogMetaMap);
+
+        // if it isn't running, break to not access zookeeper
+        if (!running) {
+            return;
+        }
+
+        LOG.info("Garbage collecting deleted ledgers' index files again just in case ledgers deleted during scanning.");
```
sijie: turned to LOG.debug

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -304,6 +304,36 @@ public void resumeMinorGC() {
...
+        // gc inactive/deleted ledgers again, just in case ledgers are deleted during scanning entry logs
+        doGcLedgers();
```
sijie: if the scan takes too longer time, it makes sense to scan the ledgers again to figure out what ledgers might have been deleted. I added a timer to explicitly check the scan time.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -304,6 +304,36 @@ public void resumeMinorGC() {
...
+
+        LOG.info("Garbage collecting empty entry log files.");
+        // gc entry logs
+        doGcEntryLogs();
```
sijie: In the case of your disk is full, the bookie cannot replay the journal. so the bookie can't actually start to reclaim the disk. The logic below is to ensure bookie can still do gc when the disks are full.

`
        try {
            checkDiskSpace();
        } catch (NoWritableLedgerDirException nwlde) {
            LOG.info("Ledger storage is already full : ", nwlde);
            // if there is no writable ledger dir, we should try reclaimSpace
            reclaimDiskSpace();
            // check disk again if we still can't have enough room for replaying journal, fail it
            checkDiskSpace();
        }
`

sijie:
@jvrao I answered the questions above (new pr will be pushed)

/cc @merlimat for another review.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
jvrao: Not really true. I have made a simple change, and the code is very very simple. I will push the code soon. This is a huge hammer for simple fix.

jvrao: Here is the gist of what I did:


+    public List<File> getWritableLedgerDirsForNewLog()
+        throws NoWritableLedgerDirException {
+
+        if (!writableLedgerDirectories.isEmpty()) {
+            return writableLedgerDirectories;
+        }
+
+        // If Force GC is not allowed under no space
+        if (forceGCAllowWhenNoSpace == false) {
+            String errMsg = "All ledger directories are non writable and force GC is not enabled.";
+            NoWritableLedgerDirException e = new NoWritableLedgerDirException(errMsg);
+            LOG.error(errMsg, e);
+            throw e;
+        }
+
+        // We don't have writable Ledger Dirs.
+        // That means we must have turned readonly but the compaction
+        // must have started running and it needs to allocate
+        // a new log file to move forward with the compaction.
+        List<File> fullLedgerDirsToAccomodateNewEntryLog = new ArrayList<File>();
+        for (File dir: this.ledgerDirectories) {
+            // Pick dirs which can accommodate little more than an entry log.
+            if (dir.getUsableSpace() > (this.entryLogSize * 1.2) ) {
+                fullLedgerDirsToAccomodateNewEntryLog.add(dir);
+            }
+        }
+
+        if (!fullLedgerDirsToAccomodateNewEntryLog.isEmpty()) {
+            LOG.info("No writable ledger dirs. Trying to go beyond to accomodate compaction."
+                    + "Dirs that can accomodate new entryLog are: {}", fullLedgerDirsToAccomodateNewEntryLog);
+            return fullLedgerDirsToAccomodateNewEntryLog;
+        }
+
+        // We will reach here when we have no option of creating a new log file for compaction
+        String errMsg = "All ledger directories are non writable and no reserved space left for creating entry log file.";
+        NoWritableLedgerDirException e = new NoWritableLedgerDirException(errMsg);
+        LOG.error(errMsg, e);
+        throw e;
+    }

--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
@@ -549,7 +549,7 @@ public class EntryLogger {
          * Allocate a new log file.
          */
         BufferedLogChannel allocateNewLog() throws IOException {
-            List<File> list = ledgerDirsManager.getWritableLedgerDirs();
+            List<File> list = ledgerDirsManager.getWritableLedgerDirsForNewLog();
             Collections.shuffle(list);
             // It would better not to overwrite existing entry log files
             File newLogFile = null;

eolivelli:
@jvrao  @merlimat ping
for me is still +1

jvrao:
I agree that this change won't conflict with my changes, but I really don't see much value in this change with our changes. @reddycharan did we send pull request on this one?

reddycharan:
@jvrao I haven't sent most of recent commits to community yet. one way or the other they are dependent on https://github.com/apache/bookkeeper/pull/108. I'm waiting for this commit to get merged for quite a long time.

@sijie @dlg99 @eolivelli

sijie:
@jvrao even you enforce force compaction or allow entry log creation, it doesn't really address the issue that this pull request addresses. the change is useful when all your disks are full (full here can mean the physical space is full). This change allows bookies have the ability to reclaim and recover from disks are all full.

sijie:
@jvrao thoughts?

eolivelli:
@sijie @jvrao @reddycharan 
do you think we can proceed with this patch ? or is there any conflict with new patches @reddycharan  is going to push ?

reddycharan:
How we handle when ledgerdirs are full when Bookie is started is little different
- if isReadOnlyModeEnabled is true, then we try to bring up bookie in ReadOnlyMode and provide ability to create EntryLog file or Index file for replaying Journal or for doing compaction, though storage has reached threshold.

In your change, you try to do force GC in this case and if still it hasn't come below threshold then you fail, if it has come below threshold then you start Bookie in RW mode.

So I believe, theoretically speaking both the features can coexist. Like following
- it tries to do force GC if storage is full during start
- if still it cannt find space, then start Bookie in RO mode and provide ability to create entrylog/index file for replaying Journal or for compaction.

But for pushing my code on top of yours I've to deal with bunch of conflicts since we are touching the same code components. 

Also, I dont think your commits are in sync with current repo code. You need to incorporate changes for handling LedgerDirsMonitor/LedgerDirsManager decoupling which we did with this change - https://github.com/apache/bookkeeper/commit/efd8ec26926d2e75f51d6d576543a0dcc116b83f .

@sijie @eolivelli @jvrao

sijie:
@reddycharan yes. I think the features can co-exist. I think the question is - are you guys okay with this change? I am trying to move on merging twitter's branch.

reddycharan:
your changes might not be required, because we are letting Bookie to start in RO only mode when disks are full and within gcWaitTime period it will do gc anyhow. Also, with your change it might impact Bookie startup time, since gc is being done as part of startup synchronously. 

I can send my CR within a day, if you want to evaluate it.

sijie:
@reddycharan I've explained that above. in the case, you don't actually have any physical disk space for compaction or open any new entry log files, your approach still doesn't work. this patch is addressing that. it allows bringing any 100% full bookies back.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -322,6 +323,42 @@ public void resumeMinorGC() {
...
+        entryLogMetaMap = extractMetaFromEntryLogs(entryLogMetaMap);
+
+        // if it isn't running, break to not access zookeeper
+        if (!running) {
+            return;
+        }
+
+        long elapsedTimeOnScan = scanStopwatch.elapsedMillis();
+
+        if (elapsedTimeOnScan >= gcWaitTime) {
```
reddycharan: not sure how this logic works, here you are instantiating scanStopwatch in this method and calculating time taken for extractMetaFromEntryLogs and comparing it with gcWaitTime

reddycharan:
@sijie ok, it is bit of an extreme case for it to happen (since BookieStorageThreshold will be generally configured at safe levels) and even in the case of eventuality (by misconfiguring or when bookie receives very high traffic before next DISK_CHECK during the DISK_CHECK_INTERVAL), we should be ok to lose a bookie because of write quorum.

Being said that, I don't have a strong reason to say no for providing extra safety mechanism / recoverability option. It should be ok, but we have to work on conflicts.

sijie:
@reddycharan sure. it happened to us before, that's why the code exists. I am trying to reach a consensus on this change. If you guys are okay with this change, I am going to rebase my change.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
sijie: yes. if the extract/scanning time is longer than gcWaitTime, we gc ledgers again. because during this period, ledgers might be already deleted.

eolivelli:
jenkins job crashed, I restarted it
https://builds.apache.org/job/bookkeeper-precommit-pullrequest/54/console

eolivelli:
CI complete with due tests failures, I think they are not releated to the patch

Failed tests: 
  TestSpeculativeRead.testSpeculativeReadFirstReadCompleteIsOk:260
  TestBackwardCompat.testCompat410:647 Shouldn't be able to write

Tests run: 1167, Failures: 2, Errors: 0, Skipped: 0

eolivelli:
sorry my comments are about #127, not this issue

reddycharan:
@sijie It should be ok for us, I don't have any in principle objection. 

But will it possible to wait for one more day before you rebase. Let me send my fix (ability to create EntryLog and IndexFile when ledgerdirs are full), so that it would be easier for me since I dont have to rebase on top of your changes. There is considerable amount of code refactoring in your commit but mine isn't as big of a change as yours and shouldn't be that complicated for you. 

I'll send pull request for mine by the end of today. LedgerDirsMonitor/LedgerDirsManager code refactoring commit made me to delay these commits.

sijie:
Sure

sijie:
@jvrao @reddycharan #190 is merged. let me know what is your thought of this pull request.

eolivelli:
@jvrao @reddycharan this is another long standing patch. Do you think we can ship this ?

sigee:
IMHO
It is a terrible practice to use the garbage collector because the use of it could introduce an over load to the software that may be even worst than on the memory, the garbage collector has his own thread which is not possible to control plus depending on the algorithm used by the gc could take more time and is consider very inefficient, you should check your software if it worst with the help of the gc because it is definitely broke, **a good solution must not depend on the gc**.

eolivelli:
@sigee we are not talking about JVM GC but the GC of the Bookie that is the auto compaction of data on disks, please take a look at GarbageCollectorThread

sijie:
this change has a huge conflict with Charan's change. Since it is not a blocker, I am going to close this one for now, so it won't block the 4.5.0 release.

jvrao:
Awesome @sijie.

PR #83, opened 2016-11-25
BOOKKEEPER-971: update bk codahale stats provider version

jiazhai:
Update bk stats provider: from codahale to yammer.
Currently io.dropwizard.metrics 3.1.0 is used most widely. will change version to 3.1.0.

And change CodahaleMetricsProvider.getMetrics() to public, since this would be used outside package.

And a small bug here:
```
            File outdir;
            if (Strings.isNullOrEmpty(prefix)) {  < === if (!)
                outdir = new File(csvDir, prefix);
            } else {
                outdir = new File(csvDir);
            }
```


sijie:
LGTM +1

PR #86, opened 2016-11-30
BOOKKEEPER-967: New testsuite for RackPlacement

reddycharan:
Added New Testsuite for testing
RackAwareEnsemblePlacementPolicy using
ScriptBasedMapping. It works
only on Unix based OS, because of Shell
script.

This testsuite has similar testscenarios as in
TestRackawareEnsemblePlacementPolicy.java.

Also here it is assumed that the script file
has execution permission.


sijie:
LGTM +1

PR #91, opened 2016-12-13
BOOKKEEPER-984: Fix BookieClientTest.testWriteGaps

sijie:
in commit 9359d682a1598e30364eca6021d976f911e055b2, the third field (last add confirmed) in the message. changed the test to include the lac in the test so the test logic can work as expected.


eolivelli:
LGTM +1
The test is not failing anymore on my laptop

PR #92, opened 2016-12-13
BOOKKEEPER-986: Handle memtable flush failure

reddycharan:
- If the memtable flush is failed previously then
for the next addEntry call it will try to flush the
existing snapshot


sijie:
The change looks good to me. +1

PR #101, opened 2016-12-28
BOOKKEEPER-907: EntryFormatter should be configur.

reddycharan:
for ReadLedgerEntriesCmd, EntryFormatter should
be configurable and HexDumpEntryFormatter should
be one of them.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/util/HexDumpEntryFormatter.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/util/HexDumpEntryFormatter.java
@@ -1,0 +1,53 @@
+/*
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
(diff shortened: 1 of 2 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/util/HexDumpEntryFormatter.java
```diff
@@ -0,0 +1,32 @@
+package org.apache.bookkeeper.util;
```
sijie: missing apache header

reddycharan: added header

zzsongsg:
+1 LGTM

sijie:
+1 merged in master as ca7e5fd

PR #102, opened 2016-12-29
BOOKKEEPER-992: ReadLog Command Enhancement

reddycharan:
- Take arguments lid and eid and prints just that entry
- Take arguments of position range and print entries
which are located in that position range


sijie:
looks good to me +1

PR #104, opened 2017-02-06
BOOKKEEPER-998: Increased the max entry size to 5MB

revans2:
Full disclosure, most of these changes were actually done by @merlimat For this I am mostly just putting them back into open source.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
@@ -174,6 +174,7 @@ public class EntryLogger {
     private final long flushIntervalInBytes;
     private final boolean doRegularFlushes;
     private long bytesWrittenSinceLastFlush = 0;
+    private final int maxSaneEntrySize;
 
     final ServerConfiguration conf;
     /**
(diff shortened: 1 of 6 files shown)


eolivelli:
@revans2 
IMHO it would be better to make it configurable as system property or in ServerConfiguration it would be even better, so it would be more "testable".
I'm not aware of some system property naming convention, maybe you can suggest one and then the discuss is open.

IMHO the max entry size can be related to the max Netty frame size but maybe we can use two different limits. For instance on the wire we need headers
@sijie what to you think ?

I think that at Salesforce they make large use of big entries (as they store BLOBs on BK), maybe JV as some number to share

For me, I'm starting a new project which will need to store "huge" (more than 10 MB) entries and so I will need to set this limit to a greater value that 5MB

revans2:
@eolivelli Thanks I'll look at making them separate.

If someone from salesforce wants to take a look that would be great.  If they are storing large BLOBs on BK there is a high likelihood that they have similar code too.

revans2:
@eolivelli I added in configs for both the server and the client.  Please take a look again.

sijie:
I think 1MB limitation was coming before we used netty for the network transportation. Increasing it to 5MB sounds okay to me. I don't think it breaks any backward compatibility issue. I don't have any strong preference here on using a system property or configuration setting for it.

jvrao:
1MB is good enough for us at least for now. Did not test anything larger.

revans2:
Thanks everyone I moved the configs to AbstractConfiguration and squashed the commits.  Hopefully this is good to go now.

revans2:
@eolivelli and @jvrao is there anything more I need to do to get this merged in?

eolivelli:
For me is +1 but I am not a committer. Usually @sijie merges PR or maybe @merlimat

jvrao:
+1 LGTM.

PR #105, opened 2017-02-08
BOOKKEEPER-999: Stop leaking threads in BookKeeper client

revans2:


jvrao:
+1 LGTM

eolivelli:
+1 this fix is good for me

maybe we should consider deprecating that constructor (add a @Deprecated annocation), in favour of the official Builder for BookKeeper client

@sijie  @jvrao  what do you think ?

zzsongsg:
+1 looks good to me.

sijie:
sorry for late response @eolivelli 

+1 LGTM

jiazhai:
+1, LGTM.

revans2:
Sorry I took so long to respond.  This kind of fell off my radar for a bit.  

@eolivelli  and @jiazhai 
I am fine with deprecating that constructor but it is not going to be a simple change.  I don't like deprecating an API and continuing to use it internally.  This constructor is used in a lot of places, 28 by by my count, and some of them are not going to be simple to replace.

There is no public replacement API, except  

```
public BookKeeper(ClientConfiguration conf, ZooKeeper zk, ClientSocketChannelFactory channelFactory)
```

which is not used at all, and looks like it too was intended to be an internal API that should be deprecated.  So if I do deprecate it I think I would rather do it on a separate JIRA/pull request.

eolivelli:
@revans2 
I see your point. Let's create a separate JIRA.

In general we need to clean up the "public" API, but I think this task will be for 4.6.0, now we need to finish 4.5.0.

revans2:
I filed https://issues.apache.org/jira/browse/BOOKKEEPER-1011 for this.

PR #107, opened 2017-02-10
BOOKKEEPER-1003 Fix TestDiskChecker so it can be used on /dev/shm

kishorvpatil:
The test generates a threshold which is negative if used on a disk where
the usable disk size is very close to the total disk size. A negative
threshold is invalid so the test fails in the wrong way. This fix stops
the threshold from going negative.

Full disclosure, most of these changes were actually done by @ivankelly  For this I am mostly just putting them back into open source.


kishorvpatil:
The unit test failures are unrelated.

jvrao:
+1 LGTM

PR #108, opened 2017-02-13
    BOOKKEEPER-944: LowWaterMark Storage Threshold

reddycharan:
BOOKKEEPER-944: LowWaterMark Storage Threshold

LowWaterMark Storage Threshold and code refactoring

    - Current implementation toggles READONLY status of the bookie as soon as a directory usage falls below the disk storage threshold.
      Added LowWaterMark parameter that limits such switches.
    	1. Bookie transition from RW to RONLY only when all the ledger dirs usage > HWM (storage threshold)
    	2. Bookie transition from RONLY to  RW only when total system disk usage (ledger/index disks) capacity is < LWM
    	3. When bookie is in RW mode all disks which are < HWM (storage threshold) are RW
    - refactored code and separated LedgerDirsMonitor from LedgerDirsManager, to remove circular dependency
      between LedgerDirsManager and LedgerDirsMonitor and also it improves testability by making them separate classes.
      It becomes easier to do functional/unit level testing and fault-injection testing at LedgerDirsMonitor class level.
    - relevant testcases

Author: Andrey Yegorov <ayegorov@salesforce.com>
Co-Author: Charan Reddy Guttapalem <cguttapalem@salesforce.com>

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDirsMonitor.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDirsMonitor.java
@@ -1,0 +1,191 @@
+/**
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+
+package org.apache.bookkeeper.bookie;
+
+import java.io.File;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.ConcurrentMap;
+
+import org.apache.bookkeeper.bookie.LedgerDirsManager.LedgerDirsListener;
+import org.apache.bookkeeper.bookie.LedgerDirsManager.NoWritableLedgerDirException;
+import org.apache.bookkeeper.conf.ServerConfiguration;
+import org.apache.bookkeeper.util.DiskChecker;
+import org.apache.bookkeeper.util.DiskChecker.DiskErrorException;
+import org.apache.bookkeeper.util.DiskChecker.DiskOutOfSpaceException;
+import org.apache.bookkeeper.util.DiskChecker.DiskWarnThresholdException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Thread to monitor the disk space periodically.
+ */
+class LedgerDirsMonitor extends BookieThread {
+    private final static Logger LOG = LoggerFactory.getLogger(LedgerDirsMonitor.class);
+    
+    private final int interval;
+    private final ServerConfiguration conf;
+    private final ConcurrentMap<File, Float> diskUsages;
+    private final DiskChecker diskChecker;
+    private final LedgerDirsManager ldm;
+
+    public LedgerDirsMonitor(final ServerConfiguration conf, 
+            final DiskChecker diskChecker,
+            final LedgerDirsManager ldm) {
+        super("LedgerDirsMonitorThread");
+        this.interval = conf.getDiskCheckInterval();
+        this.conf = conf;
+        this.diskChecker = diskChecker;
+        this.diskUsages = ldm.getDiskUsages();
+        this.ldm = ldm;
+    }
+
+    @Override
+    public void run() {
+        while (true) {
+            try {
+                List<File> writableDirs = ldm.getWritableLedgerDirs();
+                // Check all writable dirs disk space usage.
+                for (File dir : writableDirs) {
+                    try {
+                        diskUsages.put(dir, diskChecker.checkDir(dir));
+                    } catch (DiskErrorException e) {
+                        LOG.error("Ledger directory {} failed on disk checking : ", dir, e);
+                        // Notify disk failure to all listeners
+                        for (LedgerDirsListener listener : ldm.getListeners()) {
+                            listener.diskFailed(dir);
+                        }
+                    } catch (DiskWarnThresholdException e) {
+                        LOG.warn("Ledger directory {} is almost full.", dir);
+                        diskUsages.put(dir, e.getUsage());
+                        for (LedgerDirsListener listener : ldm.getListeners()) {
+                            listener.diskAlmostFull(dir);
+                        }
+                    } catch (DiskOutOfSpaceException e) {
+                        LOG.error("Ledger directory {} is out-of-space.", dir);
+                        diskUsages.put(dir, e.getUsage());
+                        // Notify disk full to all listeners
+                        ldm.addToFilledDirs(dir);
+                    }
+                }
+                // Let's get NoWritableLedgerDirException without waiting for the next iteration
+                // in case we are out of writable dirs
+                // otherwise for the duration of {interval} we end up in the state where 
+                // bookie cannot get writable dir but considered to be writable 
+                ldm.getWritableLedgerDirs();
+            } catch (NoWritableLedgerDirException e) {
+                for (LedgerDirsListener listener : ldm.getListeners()) {
+                    listener.allDisksFull();
+                }
+            }
+
+            List<File> fullfilledDirs = new ArrayList<File>(ldm.getFullFilledLedgerDirs());
+            boolean hasWritableLedgerDirs = ldm.hasWritableLedgerDirs();
+            float totalDiskUsage = 0;
+
+            // When bookie is in READONLY mode .i.e there are no writableLedgerDirs:
+            // - Check if the total disk usage is below DiskLowWaterMarkUsageThreshold.
+            // - If So, walk through the entire list of fullfilledDirs and add them back to writableLedgerDirs list if
+            // their usage is < conf.getDiskUsageThreshold.
+            if (hasWritableLedgerDirs || (totalDiskUsage = diskChecker.getTotalDiskUsage(ldm.getAllLedgerDirs())) < conf
+                    .getDiskLowWaterMarkUsageThreshold()) {
+                // Check all full-filled disk space usage
+                for (File dir : fullfilledDirs) {
+                    try {
+                        diskUsages.put(dir, diskChecker.checkDir(dir));
+                        ldm.addToWritableDirs(dir, true);
+                    } catch (DiskErrorException e) {
+                        // Notify disk failure to all the listeners
+                        for (LedgerDirsListener listener : ldm.getListeners()) {
+                            listener.diskFailed(dir);
+                        }
+                    } catch (DiskWarnThresholdException e) {
+                        diskUsages.put(dir, e.getUsage());
+                        // the full-filled dir become writable but still above
+                        // warn threshold
+                        ldm.addToWritableDirs(dir, false);
+                    } catch (DiskOutOfSpaceException e) {
+                        // the full-filled dir is still full-filled
+                        diskUsages.put(dir, e.getUsage());
+                    }
+                }
+            } else {
+                LOG.debug(
+                        "Current TotalDiskUsage: {} is greater than LWMThreshold: {}. So not adding any filledDir to WritableDirsList",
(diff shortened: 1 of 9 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDirsMonitor.java
```diff
@@ -0,0 +1,200 @@
...
+import org.apache.bookkeeper.util.DiskChecker.DiskErrorException;
+import org.apache.bookkeeper.util.DiskChecker.DiskOutOfSpaceException;
+import org.apache.bookkeeper.util.DiskChecker.DiskWarnThresholdException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Thread to monitor the disk space periodically.
+ */
+class LedgerDirsMonitor extends BookieThread {
```
revans2: Why is this not a BookieCriticalThread?  To me if we stop being able to monitor the size of the dirs it is a really bad thing and requires more then just an error log.  I am still coming up to speed on BK so I easily could be wrong here.

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/DiskChecker.java
```diff
@@ -155,6 +156,29 @@ float checkDiskFull(File dir) throws DiskOutOfSpaceException, DiskWarnThresholdE
...
+    public float getTotalDiskUsage(List<File> dirs) {
+        if (dirs == null || dirs.isEmpty()) {
+            throw new IllegalArgumentException(
+                    "list argument of getTotalDiskUsage is not supposed to be null or empty");
+        }
+        long totalUsableSpace = 0;
+        long totalSpace = 0;
+        for (File dir : dirs) {
+            totalUsableSpace += dir.getUsableSpace();
+            totalSpace += dir.getTotalSpace();
```
revans2: very minor: These will only work if each of the directories are on separate partitions, which is what a user should be doing anyways.  It might be worth using the Files API with FileStore to dedupe these, or at least output some sort of a warning somewhere if they are not the separate partitions.

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/DiskChecker.java
```diff
@@ -191,9 +215,8 @@ public float checkDir(File dir) throws DiskErrorException,
      *
      * @param diskSpaceThreshold
      */
-    @VisibleForTesting
-    void setDiskSpaceThreshold(float diskSpaceThreshold, float diskUsageWarnThreshold) {
-        validateThreshold(diskSpaceThreshold, diskSpaceThreshold);
+    public void setDiskSpaceThreshold(float diskSpaceThreshold, float diskUsageWarnThreshold) {
```
revans2: I don't see this called anywhere in the patch.  Is there a reason to make it public?

sijie:
@reddycharan can you address @revans2 's comments?

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/DiskChecker.java
reddycharan: it is used in TestDiskChecker. But I agree it doesn't need to be public.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDirsMonitor.java
reddycharan: in this commit we just refactored LedgerDirsMonitor code and brought it to a new class. It is BookieThread in the existing code, so we wish to keep it as it is.

reddycharan:
still working on this..will get back in couple of days

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/DiskChecker.java
reddycharan: valid point! internally we made assumption that ledger/index dirs wont be on the same partition. but current implementation doesn't restrict user from doing this. I think it needs little more than warn message. Logged Feature story - https://issues.apache.org/jira/browse/BOOKKEEPER-1033 . Will take this issue separately, since it needs to be handled in couple of other places as well.

reddycharan:
in the initially version we added code for Bookie to start in RO mode (if RO mode is enabled) when all disks are full. But in the commit I created today I removed that part, since it needs more than what we initially thought. Created separate Feature for it - https://issues.apache.org/jira/browse/BOOKKEEPER-1034 will follow up with that feature separately.

reddycharan:
@sijie @revans2 

Addressed your comments and sent new commit, please take a look. 

Thanks.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDirsMonitor.java
```diff
@@ -0,0 +1,191 @@
...
+
+    @Override
+    public void run() {
+        while (true) {
+            try {
+                List<File> writableDirs = ldm.getWritableLedgerDirs();
+                // Check all writable dirs disk space usage.
+                for (File dir : writableDirs) {
+                    try {
+                        diskUsages.put(dir, diskChecker.checkDir(dir));
```
revans2: I am really confused by the refactoring here. LedgerDirsMonitor is very highly coupled to the implementation of LedgerDirsManager.  Here it is reaching into and modifying the internal state of the Manager.  On this line the diskUsages are only reported to a metric so it would be simple to move the metric here, but it seems really odd to have half the callbacks from the Manager be done by the Monitor.

I just don't see much upside to separating them into separate files and potentially a lot of downside.

dlg99: @revans2 It's been over 6 months since I worked on this so may have forgotten some details.
Main point of separation of LedgerDirsMonitor and LedgerDirsManager was to break circular dependency between these two and reduce coupling. Functionally everything is same as it was minus manager having reference to monitor and dealing with its lifecycle (that is done explicitly now).
Circular dependency between two components is a strong smell of bad design (from complicating one's understanding of the code to complicating test mocks etc.), I hope we agree on that.
I left callbacks as they were (monitor calling manager's callbacks directly instead of notifying it and letting it call its callback). I agree that this can be refactored further but I did not want to get ratholed into perfecting this over other task on hands.
There is no reason for manager to handle monitor's state so that dependency is dropped. IMO it simplifies one's understanding of what's going on with that code.
I do not see problem with class being in separate file.  if you do please elaborate.

revans2: I agree that circular dependencies are bad, but so is having one class reach into the guts of another class to modify its internal state.  I do not see any reduction in coupling between the classes.  LedgerDirsManager is broken without a LedgerDirsMonitor also running (most of the callbacks just will not happen without it).  LedgerDirsMonitor cannot be instantiated without a LedgerDirsManager for it to go reach into.  All I see is that a single file was broken into two files and some APIs were made public that were not before.

To me the original LedgerDirsMonitor class does not really feel like a strict circular dependency.  It feels like an implementation detail of LedgerDirsManager that is hidden (private).

If you really want them separate then make a separate LedgerDirsMonitorCallback class.  Have the Monitor then call into instances of that callback that are registered with it.  Then have the LedgerDirsManager take a Monitor class in its constructor and register itself with it. The Manager can listen to the Monitor callbacks then update its own state internally and make its own callbacks.  This way you can never instantiate a LedgerDirsManager that will not function properly, and LedgerDirsMonitor has no knowledge of the LedgerDirsManager.

dlg99: @revans2 LDMonitor calls LDManager's callbacks directly instead of providing some level of inderection, as it did before. 
This may be changed later. It is not a goal of this change. The goal was to make relationships between two components more visible and to remove circular dependency while not changing functionality too much at this moment.

What changed is:

previously LDManager created LDMonitor and kept reference to it, managed its lifecycle etc (see "private final LedgerDirsMonitor monitor" and around in old LDManager).
At the same time LDMonitor called some of LDManager's methods to get writable dirs and notified LDManager's listeners of changes, all this implicitly and not clearly visible from the code. 
so LDManager knew about LDMonitor and LDMonitor knew about LDManager.
Maybe it made sense for folks who worked with this longer than I am, but for me it took some time to rationalize relationships between these two components.

now: LDManager does not know anything about LDMonitor. 
LDMonitor takes instance of LDManager as explicit parameter and calls its callbacks etc as before. Monitor does know about manager internals (as before) but we can consider this as a way to trigger event listeners and handle event subscriptions. Roles are better separated now.

LedgerDirsMonitor class is not public and can only be used from current package.

revans2: @dlg99 It looks like more discussion is not going to change your opinion on this.  I am a -0 on this patch the way it is and leave it at that.

eolivelli:
Overall looks good to me.
+1
I am not a fan of the refactor of the LedgerDirsMonitir but I am fine with it anyway

But I would like to have an approval from @sijie or @merlimat before merging

reddycharan:
@sijie its been quite long, can you please merge this commit. There are multiple commits from our side which depend on this change and we are holding them.

@jvrao @dlg99

eolivelli:
If @sijie is OK I can push the patch myself, or any other committer

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDirsMonitor.java
```diff
@@ -0,0 +1,191 @@
...
+
+            List<File> fullfilledDirs = new ArrayList<File>(ldm.getFullFilledLedgerDirs());
+            boolean hasWritableLedgerDirs = ldm.hasWritableLedgerDirs();
+            float totalDiskUsage = 0;
+
+            // When bookie is in READONLY mode .i.e there are no writableLedgerDirs:
+            // - Check if the total disk usage is below DiskLowWaterMarkUsageThreshold.
+            // - If So, walk through the entire list of fullfilledDirs and add them back to writableLedgerDirs list if
+            // their usage is < conf.getDiskUsageThreshold.
+            if (hasWritableLedgerDirs || (totalDiskUsage = diskChecker.getTotalDiskUsage(ldm.getAllLedgerDirs())) < conf
```
sijie: If I understand the change correctly, this is the only core piece for this change. I am not convinced about why do you need to move LedgerDirsMonitor out of LedgerDirsManager. The original implementation of LedgerDirsMonitor is an inner class that used by LedgerDirsManager for monitoring the disks. The LedgerDirsMonitor isn't exposed to any other classes.

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
```diff
@@ -104,6 +104,7 @@
     //Disk utilization
     protected final static String DISK_USAGE_THRESHOLD = "diskUsageThreshold";
     protected final static String DISK_USAGE_WARN_THRESHOLD = "diskUsageWarnThreshold";
+    protected final static String DISK_USAGE_LWM_THRESHOLD = "diskUsageLwmThreshold";
```
sijie: Do you need to validate the settings between warn_threshold and lwm_threshold?

Can you also add this setting to bookie template conf file?

reddycharan: warn-threshold can be lower or greater than lwm_threshold, so validation is not required.

I'll add this setting to template conf file.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDirsMonitor.java
reddycharan: yes, it is the core piece for this change.

@dlg99  (Andrey) tried to explain in the earlier comments, why he decoupled LedgerDirsMonitor from LedgerDirsManager.

sijie: @reddycharan I tried to read through the comments. I am not convinced about the refactor. Can you hold on the refactor and do the change here first?

jvrao: @dlg99 @reddycharan What is the down side of not doing the refactor?

eolivelli:
@reddycharan this patch has now some conflict,
can you rebase and address @sijie  comments so that we can get this merged ?

reddycharan:
@sijie @eolivelli  resolved merge conflicts and added more description to the commit as we discussed in todays call. Thanks.

eolivelli:
+1 for me. Thank you @reddycharan

sijie:
LGTM +1

sijie:
merged this. thank you @reddycharan @dlg99

PR #110, opened 2017-02-16
BOOKKEEPER-391 Support Kerberos authentication of bookkeeper

eolivelli:
This patch contains a very basic AuthProvider which uses JAAS and so enables the usage or GSSAPI/Kerberos for BookKeeper authentication

### DIFF
--- a/bookkeeper-server/pom.xml
+++ b/bookkeeper-server/pom.xml
@@ -139,7 +139,7 @@
     <dependency>
       <groupId>commons-io</groupId>
       <artifactId>commons-io</artifactId>
-      <version>2.1</version>
+      <version>2.4</version>
     </dependency>
     <dependency>
       <groupId>net.java.dev.jna</groupId>
@@ -216,9 +216,23 @@
       <artifactId>netty-all</artifactId>
       <version>${netty.version}</version>
     </dependency>
+    <dependency>
+        <groupId>org.apache.hadoop</groupId>
+        <artifactId>hadoop-minikdc</artifactId>
+        <version>2.7.3</version>
+        <scope>test</scope>
+    </dependency>          
   </dependencies>
   <build>
     <plugins>
+      <plugin>
+            <!-- for mini-kdc -->
+            <groupId>org.apache.felix</groupId>
+            <artifactId>maven-bundle-plugin</artifactId>
+            <version>3.2.0</version>
+            <inherited>true</inherited>
+            <extensions>true</extensions>
+      </plugin>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-shade-plugin</artifactId>
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/sasl/SASLBookieAuthProvider.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/sasl/SASLBookieAuthProvider.java
@@ -1,0 +1,75 @@
+/**
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+package org.apache.bookkeeper.sasl;
+
+import java.io.IOException;
+import java.util.regex.Pattern;
+import javax.security.auth.Subject;
+import javax.security.auth.login.LoginException;
+import javax.security.sasl.SaslException;
+import org.apache.bookkeeper.auth.AuthCallbacks;
+import org.apache.bookkeeper.auth.AuthToken;
+import org.apache.bookkeeper.auth.BookieAuthProvider;
+import org.apache.bookkeeper.bookie.BookieConnectionPeer;
+import org.apache.bookkeeper.client.BKException;
+import org.apache.bookkeeper.conf.ServerConfiguration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class SASLBookieAuthProvider implements BookieAuthProvider {
+
+    private static final Logger LOG = LoggerFactory.getLogger(SASLBookieAuthProvider.class);
+
+    private SaslServerState server;
+    private final AuthCallbacks.GenericCallback<Void> completeCb;
+    private final Pattern allowedIdsPattern;
+
+    SASLBookieAuthProvider(BookieConnectionPeer addr, AuthCallbacks.GenericCallback<Void> completeCb,
+        ServerConfiguration serverConfiguration, Subject subject, Pattern allowedIdsPattern) {
+        this.completeCb = completeCb;
+        this.allowedIdsPattern = allowedIdsPattern;
+        try {
+            server = new SaslServerState(serverConfiguration, subject, allowedIdsPattern);
+        } catch (IOException | LoginException error) {
+            LOG.error("Error while booting SASL server", error);
+            completeCb.operationComplete(BKException.Code.UnauthorizedAccessException, null);
+        }
+    }
+
+    @Override
+    public void process(AuthToken m, AuthCallbacks.GenericCallback<AuthToken> cb) {
+        try {
+            byte[] clientSideToken = m.getData();
+            byte[] response = server.response(clientSideToken);
+            if (response != null) {
+                cb.operationComplete(BKException.Code.OK, AuthToken.wrap(response));
+            }
+            if (server.isComplete()) {
+                completeCb.operationComplete(BKException.Code.OK, null);
+            }
+        } catch (SaslException err) {
+            LOG.debug("SASL error", err);
+            completeCb.operationComplete(BKException.Code.UnauthorizedAccessException, null);
+        }
+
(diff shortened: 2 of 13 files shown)


eolivelli:
@sijie  I have rebased the patch

bookkeeper-server/pom.xml
```diff
@@ -178,10 +178,24 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
+        <groupId>org.apache.hadoop</groupId>
+        <artifactId>hadoop-minikdc</artifactId>
```
sijie: Why do we need introduce hadoop dependencies?

eolivelli: Scope is "test" and it is needed in order to have a working Kerberos server. I'm using the one used in ZooKeeper.
Maybe we can switch to http://directory.apache.org/kerby/ if preferred (ZooKeeper uses both of them for different branches)

eolivelli:
@sijie  
JIRA for documentation https://issues.apache.org/jira/browse/BOOKKEEPER-1016

eolivelli:
@revans2 can you take a look at this patch ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/sasl/SaslServerState.java
```diff
@@ -0,0 +1,294 @@
...
+        }
+
+        private void handleRealmCallback(RealmCallback rc) {
+            LOG.debug("client supplied realm: " + rc.getDefaultText());
+            rc.setText(rc.getDefaultText());
+        }
+
+        private void handleAuthorizeCallback(AuthorizeCallback ac) {
+            String authenticationID = ac.getAuthenticationID();
+            String authorizationID = ac.getAuthorizationID();
```
revans2: If I remember correctly SASL has built in the ability to pretend to be someone else.  That is the difference between the authenticationID and the authorizationID.  The authenticationID is who the client authenticated as and the authorizationID is who you they want to pretend to be.  Ignoring the authorizationID is fine, but it might be good to fail the SASL attempt if the authrorizationID is set and not set to the same thing as the authenticationID.  This isn't a security issue.  It is mostly that is a client wants to pretend to be someone else and ends up being themselves it might be confusing instead of failing quickly because we just don't support that.

bookkeeper-server/src/main/java/org/apache/bookkeeper/sasl/SASLBookieAuthProvider.java
```diff
@@ -0,0 +1,71 @@
...
+    @Override
+    public void process(AuthToken m, AuthCallbacks.GenericCallback<AuthToken> cb) {
+        try {
+            byte[] clientSideToken = m.getData();
+            byte[] response = server.response(clientSideToken);
+            if (response != null) {
+                cb.operationComplete(BKException.Code.OK, AuthToken.wrap(response));
+            }
+            if (server.isComplete()) {
+                completeCb.operationComplete(BKException.Code.OK, null);
```
revans2: We are missing a critical Authorization piece here.  MD5 provides both simply because there is only one/small number of users being configured.  In the case of Kerberos it only provides Authentication.  So the way the code is currently written anyone with credentials in the KDC will have access to the bookies.  I am not sure if this is what you want, because it is not common to spin up a single KDC for a service.  Instead there is usually a KDC for an organization.

eolivelli:
@revans2 thanks for the review.
I will address all of the comments next week and update the patch.

I do not need MD5 for my projects/in production so I can drop it, honestly I have taken it from ZooKeeper. So I'm going to drop it

eolivelli:
@revans2 I have double checked and with some other product (i.e. ZooKeeper) we are using legacy MD5 in some production deployment.
As in this implementation I would like to remain as much similar to ZK as possible I would like to keep MD5 as well, maybe we can write some "Warning" in the documentation

Is is OK for you ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/sasl/SaslClientState.java
```diff
@@ -0,0 +1,207 @@
...
+            ? clientConfiguration.getString(SaslConstants.JAAS_AUDITOR_SECTION_NAME, SaslConstants.JAAS_DEFAULT_AUDITOR_SECTION_NAME)
+            : clientConfiguration.getString(SaslConstants.JAAS_CLIENT_SECTION_NAME, SaslConstants.JAAS_DEFAULT_CLIENT_SECTION_NAME);
+        AppConfigurationEntry[] entries = Configuration.getConfiguration()
+            .getAppConfigurationEntry(configurationEntry);
+        if (entries == null) {
+            LOG.info("No JAAS Configuration found with section BookKeeper");
+            return null;
+        }
+        try {
+            LoginContext loginContext = new LoginContext(configurationEntry, new ClientCallbackHandler(null));
```
eolivelli: I have checked my code.
My implementation creates a new LoginContext (in SASLBookieAuthProvider/SaslServerState) for every incoming connection so there is no need to perform a TGT renewal, in fact I'm not running in such a problem in "staging".
I see it can be a performance/resource problem, in by staging env I have a limited number of clients (10 to 100) and I have never got into problems.

The same happens at the client-side.

@revans2 
do you think we must cache the TGT as ZooKeeper ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/sasl/SASLBookieAuthProvider.java
eolivelli: @revans2 
Do you have any suggestion ? maybe we can validate the principal and force it to be bookkeeper@XXXX and for the server side something like bookie@XXXX (for Kerberos case)

I can't find any validation code in ZooKeeper (SaslServerCallbackHandler)

revans2:
@eolivelli I am OK with having MD5 in there, but it is horribly insecure and should not be used in production.  If we do put it in we should deprecate it very quickly afterwards and ask people to move to kerberos.

https://en.wikipedia.org/wiki/MD5

The fact that ZK still uses it is on my TODO list as it should be replaced with something better.

eolivelli:
@revans2 
I have pushed the TGT refresh logic for the bookie side, I'm going to do the same for the client-side

I will push other commits soon in order to address all of your comments (TGT refresh on the client side, logging, authenticationID vs authorizationID...)

Meanwhile I'm asking you some suggestion about the validation of the client side principal. I did not find any example on ZooKeeper code.

bookkeeper-server/src/main/java/org/apache/bookkeeper/sasl/SaslServerState.java
eolivelli: @revans2 I have added a safety check

revans2:
@eolivelli Validation of the client side principal in ZK is done through ACLs.  The simplest thing to do in our case is to have a white list of clients (possibly with regular expressions).  If you want to do it as a separate JIRA you can, but without it your would need a separate KDC for the cluster to restrict who is and isn't allowed to access the bookie.

bookkeeper-server/src/main/java/org/apache/bookkeeper/sasl/SASLBookieAuthProvider.java
eolivelli: @revans2  Added a basic regexp validation

eolivelli:
@revans2  @sijie 
I have addressed all of the questions/issues. This PR is ready for a new review.
Thank you

bookkeeper-server/src/main/java/org/apache/bookkeeper/sasl/TGTRefreshThread.java
```diff
@@ -0,0 +1,256 @@
...
+                nextRefresh = now + MIN_TIME_BEFORE_RELOGIN;
+                nextRefreshDate = new Date(nextRefresh);
+                LOG.warn("No TGT found: will try again at {}", nextRefreshDate);
+            } else {
+                nextRefresh = getRefreshTime(tgt);
+                long expiry = tgt.getEndTime().getTime();
+                Date expiryDate = new Date(expiry);
+                if ((container.isUsingTicketCache()) && (tgt.getEndTime().equals(tgt.getRenewTill()))) {
+                    Object[] logPayload = {expiryDate, container.getPrincipal(), container.getPrincipal()};
+                    LOG.error("The TGT cannot be renewed beyond the next expiry date: {}." + "This process will not be able to authenticate new SASL connections after that " + "time (for example, it will not be authenticate a new connection with a Zookeeper " + "Quorum member).  Ask your system administrator to either increase the " + "'renew until' time by doing : 'modprinc -maxrenewlife {}' within " + "kadmin, or instead, to generate a keytab for {}. Because the TGT's " + "expiry cannot be further extended by refreshing, exiting refresh thread now.", logPayload);
```
revans2: This error message looks like a copy and paste.  It might be confusing to talk about zookeeper, when in fact we are bookkeeper here.

eolivelli:
@revans2 I looked through all the messages and changed the reference to ZooKeeper.
I have clipped long lines too

revans2:
+1 still looks great.

eolivelli:
@sijie if you are still OK with this patch can you merge it ?

sijie:
@eolivelli can you rebase and merge it? +1 from me

bookkeeper-server/src/main/java/org/apache/bookkeeper/sasl/SaslServerState.java
EronWright: To add some color, this is similar to the proxy user support that is pervasive in Hadoop.  For example, you configure HDFS to trust the YARN service to access files on behalf of users that are interacting with YARN.

In the Bookkeeper context, the service would be configured to treat the user referenced by `authenticationID` as a superuser account that is allowed to impersonate other users.  After authentication succeeds, you use the user referenced by `authorizationId` as the basis for authorization decisions, i.e. as the effective user.

eolivelli: Thank you @EronWright 
This patch has already been integrated into main codebase.
I think that "impersionation" will be very interesting once we introduce ACLs or quotas....
Can you file an issue in order to track better your idea ?

PR #111, opened 2017-02-17
BOOKKEEPER-1007 Explicit LAC: make the interval configurable in ms

eolivelli:


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ExplicitLacFlushPolicy.java
```diff
@@ -111,7 +110,7 @@ public String toString() {
             };
             try {
                 scheduledFuture = lh.bk.mainWorkerPool.scheduleAtFixedRateOrdered(lh.getId(), updateLacTask,
-                        explicitLacIntervalInSec, explicitLacIntervalInSec, SECONDS);
+                        explicitLacIntervalInSec, explicitLacIntervalInSec, TimeUnit.MILLISECONDS);
```
revans2: So the variable names say that they are in Seconds but the timeunit says to interpret it as milliseconds.  Please at least rename the variables.

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ClientConfiguration.java
```diff
@@ -610,7 +610,7 @@ public int getExplictLacInterval() {
     /**
      * Set the interval to check the need for sending an explicit LAC.
      * @param interval
-     *        Number of seconds between checking the need for sending an explict LAC.
+     *        Number of milli seconds between checking the need for sending an explict LAC.
      * @return Client configuration.
      */
     public ClientConfiguration setExplictLacInterval(int interval) {
```
revans2: This feels really bad from a backwards comparability standpoint.  Could we add in a new method that lets you set milliseconds but update this API to call into the other but converting seconds to milliseconds?

eolivelli: @revans2 thanks for the review. I have addressed the local variable naming issue.

This feature has been contributed by SalesForce by merging their local fork. @jvrao are you ok with this change ?

The feature on the "Apache version" is very "new" and so I think there is no compatibility issue actually.

I have filed the PR now because I would like to address the issue as soon as possible, so that we are sure that no one is using it in production

revans2: @eolivelli thanks for pointing that out.  I am still coming up to speed on all of BK, so if `setExplictLacInterval` has not been in an official release yet, then I am fine with just changing it.

revans2:
With the renamed variables I am +1 (non-binding)

jvrao:
+1 
LGTM. We need to make sure to change our configuration to reflect this.

PR #112, opened 2017-02-17
BOOKKEEPER-552: 64 Bits Ledger ID Generation

knusbaum:
This PR adds 63-bit ID generation to Bookkeeper and modifies the hierarchical ledger structure to be backwards compatible.


reddycharan:
Hey @knusbaum we already pushed our commit for 63 bit ledgerID. Can you please check that - 

https://github.com/apache/bookkeeper/commit/42e8f1294f206cbe51a6af669cf605833b78bf42

knusbaum:
@reddycharan Yes, this builds on that. It makes it backwards compatible and builds the 64-bit id generation stuff which was missing.

reddycharan:
@knusbaum can you please explain briefly, what you are doing in this change and why LongHierarchicalLedgermanager is being removed.

knusbaum:
LongHierarchicalLedgerManager is being moved to HierarchicalLedgerManager and HierarchicalLedgerManager is being moved to LegacyHierarchicalLedgerManager.

LongHierarchicalLedgerManager (now HierarchicalLedgerManager) has been made backwards-compatible so that existing Bookkeeper instances running HierarchicalLedgerManager can be upgraded to the new code, which will allow 64-bit IDs.

This is achieved by having the HierarchicalLedgerManager proxy requests to the LegacyHierarchicalLedgerManager for ids which are less than Integer.MAX_VALUE
For ids greater than that, the new code is used.

We also add a new id generator. The existing ZkLedgerIdGenerator is limited to 31 bits (about 2 million) ledgers because ZooKeeper's sequential id generator uses a signed int internally. IDs start going negative if you overflow it, which messes some stuff up. 

The new ID generator uses a multi-level directory scheme to overcome this.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BKException.java
```diff
@@ -142,6 +144,7 @@ public static BKException create(int code) {
         int UnauthorizedAccessException = -102;
         int UnclosedFragmentException = -103;
         int WriteOnReadOnlyBookieException = -104;
+        int LedgerIdOverflowException = -106;
```
revans2: What happened to -105?  Was it previously already used?  Or is it part of code Yahoo intends to push back?

knusbaum: Part of our pushback. I wasn't sure how best to handle the discrepancy. Mixing them up now seems undesirable.

revans2: Can we at least put in a comment about it?

knusbaum:
@reddycharan Do you see any problems with this?

jvrao:
Couple of questions:
1. Charan with this change how does LongHierarchicalLedgerManager users (Salesforce) has backward compatibility?
2.  @knusbaum Why not keep LongHierarchicalLedgerManager around? even when HierarchicalLedgerManager supporting both ids.?

knusbaum:
@jvrao @reddycharan 

So my plan is this: Keep LongHierarchicalLedgerManager as is currently in master, move HierarchicalLedgerManager to LegacyHierarchicalLedgerManager and create the new backwards-compatible one as HierarchicalLedgerManager. This way, if you're already using LongHierarchicalLedgerManager, it will still work. If you're using the current HierarchicalLedgerManager, that will work, and if you (for some unknown reason) want to use the old legacy one, it still exists as well.

Thoughts?

reddycharan:
It sounds good that LongHLM is not changed. For HierarchicalLM also, it sounds good, but I'm not the best person to comment since we are not using it. Thanks @knusbaum.

knusbaum:
Closing in favor of #114

PR #49, opened 2016-06-17
BOOKKEEPER-930 Option to disable Bookie networking

eolivelli:

PR #116, opened 2017-03-14
BOOKKEEPER-1008 Move to Netty4.1

kishorekasi:


eolivelli:
I have starting lloking at the PR 
first comments on new dependencies:
- netty-tcnative-1.1.33.Fork25.pom -> there is a 2.0.0.Final version, what do you think ? (https://mvnrepository.com/artifact/io.netty/netty-tcnative) 
- use of os-maven-plugin +  netty-tcnative-boringssl-static-1.1.33.Fork25-linux-x86_64.jar -> I think it would be better to include only a specific version, the build depends on the OS which is running the build so IMHO it is not reproducible (I have Fedora for instance)
- OK for "boringssl-static"  other versions are very OS dependant

@revans2 which version of the native libs are you using ?

eolivelli:
On my laptop the NetworkLessBookieTest is hung

I think you have dropped the work done on BOOKKEEPER-930, that included a refactor to introduce more ChannelManagers

eolivelli:
A note on "netty-all" dependency:
I would prefer to include the dependency only to the needed subcomponents, because with "all" we are imposing a lot of garbage to clients

revans2:
Minor nit: line 108 of BookKeeper.java ownChannelFactory really should be renamed to ownEventLoopGroup (it is not a channel factory any more)

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieNettyServer.java
```diff
@@ -112,19 +151,121 @@ boolean isRunning() {
...
-            allChannels.setReadable(true).awaitUninterruptibly();
+            for (Channel channel : allChannels) {
+                channel.config().setAutoRead(true);
+            }
             suspensionLock.notifyAll();
         }
     }
 
+    private void listenOn(InetSocketAddress address, BookieSocketAddress bookieAddress) throws InterruptedException {
+        if (!conf.isDisableServerSocketBind()) {
```
revans2: For reasons of code coupling, correctness, and code reuse I really would prefer to see us refactor the code.  I think all of this can be turned into something like.

```
private void listenOn(InetSocketAddress address, BookieSocketAddress bookieAddress) throws InterruptedException {
    if (eventLoopGroup != null) {
        listenOn(eventLoopGroup, address, bookieAddress);
    }
    if (jvmEventLoopGroup != null) {
        listenOn(jvmEventLoopGroup, address, bookieAddress);
    }
}

private void listenOn(EventLoopGroup elg, InetSocketAddress address, BookieSocketAddress bookieAddress) throws InterruptedException {
            ServerBootstrap bootstrap = new ServerBootstrap();
            bootstrap.childOption(ChannelOption.ALLOCATOR, new PooledByteBufAllocator(true));
            bootstrap.group(eventLoopGroup, eventLoopGroup);
            bootstrap.childOption(ChannelOption.TCP_NODELAY, conf.getServerTcpNoDelay());
            bootstrap.childOption(ChannelOption.SO_LINGER, conf.getServerSockLinger());
            bootstrap.childOption(ChannelOption.RCVBUF_ALLOCATOR,
                    new AdaptiveRecvByteBufAllocator(conf.getRecvByteBufAllocatorSizeMin(),
                            conf.getRecvByteBufAllocatorSizeInitial(), conf.getRecvByteBufAllocatorSizeMax()));

            if (elg instanceof DefaultEventLoopGroup) {
                bootstrap.channel(LocalServerChannel.class);
            } else if (elg instanceof EpollEventLoopGroup) {
                bootstrap.channel(EpollServerSocketChannel.class);
            } else if (elg instanceof NioEventLoopGroup) {
                jbootstrap.channel(NioServerSocketChannel.class);
            } else {
                throw new IllegalArgumentException("Don't know how to handle an EventLoopGroup of type " + elg.getClass());
            }

            bootstrap.childHandler(new BookieChannelInitializer());
            if (elg instanceof DefaultEventLoopGroup) {
                bootstrap.bind(bookieAddress.getLocalAddress()).sync();
                LocalBookiesRegistry.registerLocalBookieAddress(bookieAddress);
            } else {
                // Bind and start to accept incoming connections
                bootstrap.bind(address.getAddress(), address.getPort()).sync();
            }
}

...

class BookieChannelInitializer extends ChannelInitializer<Channel> {
        @Override
        public void initChannel(Channel ch) throws Exception {
            synchronized (suspensionLock) {
                while (suspended) {
                    suspensionLock.wait();
                }
            }
            BookieSideConnectionPeerContextHandler contextHandler = new BookieSideConnectionPeerContextHandler();
            ChannelPipeline pipeline = ch.pipeline();
            pipeline.addLast("lengthbaseddecoder",
                    new LengthFieldBasedFrameDecoder(maxMessageSize, 0, 4, 0, 4));
            pipeline.addLast("lengthprepender", new LengthFieldPrepender(4));

            pipeline.addLast("bookieProtoDecoder", requestDecoder);
            pipeline.addLast("bookieProtoEncoder", responseEncoder);
            pipeline.addLast("bookieAuthHandler",
                    new AuthHandler.ServerSideHandler(contextHandler.getConnectionPeer(), authProviderFactory));

            ChannelInboundHandler requestHandler = isRunning.get() ? new BookieRequestHandler(conf,
                    requestProcessor, allChannels) : new RejectRequestHandler();

            pipeline.addLast("bookieRequestHandler", requestHandler);
            pipeline.addLast("contextHandler", contextHandler);
        }
}

```

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieNettyServer.java
```diff
@@ -112,19 +151,121 @@ boolean isRunning() {
...
+                    pipeline.addLast("lengthprepender", new LengthFieldPrepender(4));
+
+                    pipeline.addLast("bookieProtoDecoder", requestDecoder);
+                    pipeline.addLast("bookieProtoEncoder", responseEncoder);
+                    pipeline.addLast("bookieAuthHandler", new AuthHandler.ServerSideHandler(contextHandler.getConnectionPeer(), authProviderFactory));
+
+                    ChannelInboundHandler requestHandler = isRunning.get()
+                            ? new BookieRequestHandler(conf, requestProcessor, allChannels) : new RejectRequestHandler();
+                    pipeline.addLast("bookieRequestHandler", requestHandler);
+
```
revans2: contextHandler is missing here too.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieNettyServer.java
```diff
@@ -133,13 +274,19 @@ void shutdown() {
...
         allChannels.close().awaitUninterruptibly();
-        for (ChannelManager channel : channels) {
-            channel.close();
+
+        if (eventLoopGroup != null) {
+            eventLoopGroup.shutdownGracefully();
         }
+        if (jvmEventLoopGroup != null) {
+            jvmEventLoopGroup.shutdownGracefully();
+            LocalBookiesRegistry.registerLocalBookieAddress(bookieAddress);
```
revans2: Shoudn't this be unregister?  and as a nit I would expect that we would want to unregister first and then shutdown afterwards.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieNettyServer.java
```diff
@@ -190,56 +337,25 @@ public BookieConnectionPeer getConnectionPeer() {
         }
 
         @Override
-        public void channelBound(ChannelHandlerContext ctx, ChannelStateEvent e) throws Exception {
-            channel = ctx.getChannel();
+        public void channelActive(ChannelHandlerContext ctx) throws Exception {
+            channel = ctx.channel();
```
revans2: Do we want to call `super.channelRegister(ctx);` too so that it propagates to others in the chain?

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieNettyServer.java
```diff
@@ -190,56 +337,25 @@ public BookieConnectionPeer getConnectionPeer() {
...
-
     }
 
     private static class CleanupChannelGroup extends DefaultChannelGroup {
 
         private AtomicBoolean closed = new AtomicBoolean(false);
 
-        CleanupChannelGroup() {
-            super("BookieChannelGroup");
+        public CleanupChannelGroup(EventLoopGroup eventLoopGroup) {
```
revans2: nit the public is not needed for a private static class.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
```diff
@@ -269,56 +272,94 @@ private void completeOperation(GenericCallback<PerChannelBookieClient> op, int r
...
-        bootstrap.setOption("child.sendBufferSize", conf.getClientSendBufferSize());
-        bootstrap.setOption("child.receiveBufferSize", conf.getClientReceiveBufferSize());
-        bootstrap.setOption("writeBufferLowWaterMark", conf.getClientWriteBufferLowWaterMark());
-        bootstrap.setOption("writeBufferHighWaterMark", conf.getClientWriteBufferHighWaterMark());
-        SocketAddress bookieAddr = addr.getSocketAddress();        
-        if (channelFactory instanceof LocalClientChannelFactory) {
+        // Set up the ClientBootStrap so we can create a new Channel connection to the bookie.
+        Bootstrap bootstrap = new Bootstrap();
+        bootstrap.group(eventLoopGroup);
+        if (SystemUtils.IS_OS_LINUX && eventLoopGroup instanceof EpollEventLoopGroup) {
```
revans2: nit The IS_OS_LINUX check is not needed.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
```diff
@@ -1277,6 +1295,7 @@ public void writeComplete(int rc, long ledgerId, long entryId, BookieSocketAddre
                     cancelTimeout();
                     if (pcbc.addEntryOpLogger != null) {
                         long latency = MathUtils.elapsedNanos(startTime);
+                        // LOG.info("Write latency: {}", latency);
```
revans2: nit please remove this

bookkeeper-server/src/test/java/org/apache/bookkeeper/proto/NetworkLessBookieTest.java
```diff
@@ -68,12 +71,11 @@ public void testUseLocalBookie() throws Exception {
         }
 
         for (BookieServer bk : bs) {
-            for (ChannelManager channel : bk.nettyServer.channels) {
-                if (! (channel instanceof VMLocalChannelManager)) {
+            for (Channel channel : bk.nettyServer.allChannels) {
```
revans2: This test will not be valid because of the reasons I described above, with the allChannels being stomped on.  I would suggest you also check that eventLoopGroup is null.

bookkeeper-server/src/test/java/org/apache/bookkeeper/test/BookieClientTest.java
```diff
@@ -228,22 +224,19 @@ public void testWriteGaps() throws Exception {
...
-        bb.putLong(lid);
-        bb.putLong(eid);
-        bb.putLong(eid-1);
-        bb.putInt(i);
-        bb.flip();
-        return ChannelBuffers.wrappedBuffer(bb);
+    private ByteBuf createByteBuffer(int i, long lid, long eid) {
+        ByteBuf bb = Unpooled.buffer(4 + 16);
+        bb.writeLong(lid);
+        bb.writeLong(eid);
```
revans2: This is missing the translation of `bb.putLong(eid-1);`, and the allocation needs to be updated to match it.

revans2:
Also it looks like Jenkins failed with some findbugs issues.  The logs say that it timed out but when I run it I get

```
INFO] --- findbugs-maven-plugin:3.0.3:check (default-cli) @ bookkeeper-server ---
[INFO] BugInstance size is 4
[INFO] Error size is 0
[INFO] Total bugs: 4
[INFO] Dead store to ledgerId in org.apache.bookkeeper.bookie.FileInfo.setExplicitLac(ByteBuffer) [org.apache.bookkeeper.bookie.FileInfo] At FileInfo.java:[line 140] DLS_DEAD_LOCAL_STORE
[INFO] Inconsistent synchronization of org.apache.bookkeeper.bookie.FileInfo.explicitLac; locked 85% of time [org.apache.bookkeeper.bookie.FileInfo, org.apache.bookkeeper.bookie.FileInfo, org.apache.bookkeeper.bookie.FileInfo, org.apache.bookkeeper.bookie.FileInfo, org.apache.bookkeeper.bookie.FileInfo, org.apache.bookkeeper.bookie.FileInfo, org.apache.bookkeeper.bookie.FileInfo, org.apache.bookkeeper.bookie.FileInfo, org.apache.bookkeeper.bookie.FileInfo, org.apache.bookkeeper.bookie.FileInfo, org.apache.bookkeeper.bookie.FileInfo, org.apache.bookkeeper.bookie.FileInfo, org.apache.bookkeeper.bookie.FileInfo, org.apache.bookkeeper.bookie.FileInfo] Unsynchronized access at FileInfo.java:[line 118]Unsynchronized access at FileInfo.java:[line 145]Synchronized access at FileInfo.java:[line 121]Synchronized access at FileInfo.java:[line 122]Synchronized access at FileInfo.java:[line 123]Synchronized access at FileInfo.java:[line 124]Synchronized access at FileInfo.java:[line 125]Synchronized access at FileInfo.java:[line 134]Synchronized access at FileInfo.java:[line 137]Synchronized access at FileInfo.java:[line 138]Synchronized access at FileInfo.java:[line 140]Synchronized access at FileInfo.java:[line 135]Synchronized access at FileInfo.java:[line 141]Synchronized access at FileInfo.java:[line 143] IS2_INCONSISTENT_SYNC
[INFO] Inconsistent synchronization of org.apache.bookkeeper.client.LedgerHandle.length; locked 90% of time [org.apache.bookkeeper.client.ReadOnlyLedgerHandle$MetadataUpdater, org.apache.bookkeeper.client.LedgerHandle, org.apache.bookkeeper.client.LedgerHandle, org.apache.bookkeeper.client.LedgerHandle, org.apache.bookkeeper.client.LedgerHandle, org.apache.bookkeeper.client.LedgerHandle, org.apache.bookkeeper.client.LedgerHandle, org.apache.bookkeeper.client.LedgerHandle, org.apache.bookkeeper.client.LedgerHandle, org.apache.bookkeeper.client.LedgerRecoveryOp, org.apache.bookkeeper.client.LedgerHandle$2] Unsynchronized access at ReadOnlyLedgerHandle.java:[line 59]Synchronized access at LedgerHandle.java:[line 259]Synchronized access at LedgerHandle.java:[line 778]Synchronized access at LedgerHandle.java:[line 778]Synchronized access at LedgerHandle.java:[line 1362]Synchronized access at LedgerHandle.java:[line 249]Synchronized access at LedgerHandle.java:[line 249]Synchronized access at LedgerHandle.java:[line 250]Synchronized access at LedgerHandle.java:[line 977]Synchronized access at LedgerRecoveryOp.java:[line 151]Synchronized access at LedgerHandle.java:[line 379] IS2_INCONSISTENT_SYNC
[INFO] Futile attempt to change max pool size of ScheduledThreadPoolExecutor in new org.apache.bookkeeper.util.OrderedSafeExecutor(String, int, ThreadFactory, StatsLogger, boolean, long) [org.apache.bookkeeper.util.OrderedSafeExecutor] At OrderedSafeExecutor.java:[line 186] DMI_FUTILE_ATTEMPT_TO_CHANGE_MAXPOOL_SIZE_OF_SCHEDULED_THREAD_POOL_EXECUTOR
[INFO]
```

All of these look like they are unrelated to the patch though

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieNettyServer.java
eolivelli: @revans2 in  BOOKKEEPER-930 we refactored the code in order to have a more clear concept of ChannelManager. I think it would be better to leave the code in the form of BOOKKEEPER-930 which is more modular as @sijie proposed at the time of implementing  BOOKKEEPER-930

bookkeeper-server/pom.xml
```diff
@@ -178,8 +208,26 @@
...
+    <dependency>
+      <groupId>io.netty</groupId>
+      <artifactId>netty-all</artifactId>
+      <version>${netty.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>io.netty</groupId>
+      <artifactId>netty-tcnative-boringssl-static</artifactId>
+      <version>${netty-tcnative.version}</version>
+      <classifier>${os.detected.classifier}</classifier>
```
eolivelli: I prefer to use
 ```
<dependency>
            <groupId>io.netty</groupId>
            <artifactId>netty-transport-native-epoll</artifactId>
            <version>${libs.netty4}</version>
            <scope>compile</scope>
            <classifier>linux-x86_64</classifier>
        </dependency>
        <dependency>
            <groupId>io.netty</groupId>
            <artifactId>netty-tcnative-boringssl-static</artifactId>
            <version>${libs.netty4.tcnative}</version>
            <scope>compile</scope>            
        </dependency>
```

without the ${os.detected.classifier} but with linux-x86_64

File: bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieNettyServer.java (hunk 1 above)
Comment (revans2) [inline bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieNettyServer.java]: @eolivelli I would be fine with something like that too.  My main goal is that we should reduce/eliminate the amount of code that is a copy/paste and that we reduce the amount of coupling between different pieces of the code.

File: bookkeeper-server/pom.xml (hunk 10 above)
Comment (revans2) [inline bookkeeper-server/pom.xml]: For me the issue mostly is making sure that we document clearly when a release (binary or otherwise) is arch specific, which features depend on that arch specific code, and if there is a simple way to work around it (like rebuild on the platform you want to run on) how to do that.

Hard coding the architecture to x86_64 does not feel ideal to me.

**Comment by kishorekasi** [comment]
Thanks Enrico and Bobby. I will address the comments and update the PR.

Kishore

File: bookkeeper-server/pom.xml (hunk 10 above)
Comment (sijie) [inline bookkeeper-server/pom.xml]: I'd like to know more reasons on why we need to introduce arch specific dependencies.

File: bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeper.java [hunk 11]
```diff
@@ -225,10 +233,10 @@ private static ZooKeeper validateZooKeeper(ZooKeeper zk) throws NullPointerExcep
         return zk;
     }
 
-    private static ClientSocketChannelFactory validateChannelFactory(ClientSocketChannelFactory factory)
+    private static EventLoopGroup validateChannelFactory(EventLoopGroup eventLoopGroup)
```
sijie: +1 on this comment. we should rename it.

kishorekasi:
I am currently little busy with another priority at work and haven't spent
time on this in the past week. I hope to get back to this soon.

Kishore

eolivelli:
At the last meeting (2017-03-23) we discussed about this PR and the inclusion of OS-specific dependencies.
We all agreed that is it important to document this feature and how to configure the classpath in order to achive best performances for a specific OS

I suggest using only the raw dependency without the native library, this will make the build non-platform-dependant:
<dependency>
    <groupId>io.netty</groupId>
    <artifactId>netty-transport-native-epoll</artifactId>
   <version>${libs.netty4}</version>
</dependency>

And document how to add the OS-specific version on the classpath like this:

<dependency>
   <groupId>io.netty</groupId>
   <artifactId>netty-transport-native-epoll</artifactId>
   <version>${libs.netty4}</version>
   <scope>compile</scope>
   <classifier>linux-x86_64</classifier>
</dependency>

Beware that the raw dependency will import only Java classes and so we will have to handle gracefully exceptions like this an fallback to Nio:

java.lang.UnsatisfiedLinkError: no netty-transport-native-epoll in java.library.path
	at java.lang.ClassLoader.loadLibrary(ClassLoader.java:1867)
	at java.lang.Runtime.loadLibrary0(Runtime.java:870)
	at java.lang.System.loadLibrary(System.java:1122)
	at io.netty.util.internal.NativeLibraryUtil.loadLibrary(NativeLibraryUtil.java:38)
	at io.netty.util.internal.NativeLibraryLoader.loadLibrary(NativeLibraryLoader.java:259)
	at io.netty.util.internal.NativeLibraryLoader.load(NativeLibraryLoader.java:199)
	at io.netty.channel.epoll.Native.loadNativeLibrary(Native.java:269)
	at io.netty.channel.epoll.Native.<clinit>(Native.java:64)
	at io.netty.channel.epoll.IovArray.<clinit>(IovArray.java:57)
	at io.netty.channel.epoll.EpollEventLoop.<init>(EpollEventLoop.java:54)

we should also remove checks like "SystemUtils.IS_OS_LINUX"

sijie:
@kishorekasi any updates?

eolivelli:
@kishorekasi could you push your current local version so that me and @revans2 can help you with failing tests ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeper.java
```diff
@@ -272,9 +276,9 @@ public BookKeeper(ClientConfiguration conf, ZooKeeper zk)
      * @throws InterruptedException
      * @throws KeeperException if the passed zk handle is not connected
      */
-    public BookKeeper(ClientConfiguration conf, ZooKeeper zk, ClientSocketChannelFactory channelFactory)
+    public BookKeeper(ClientConfiguration conf, ZooKeeper zk, EventLoopGroup eventLoopGroup)
```
kishorekasi: fixed

kishorekasi:
Enrico,

I have updated the pull request with my local changes.

Kishore

eolivelli:
Thank you @kishorekasi.

At this point this branch fails to build
What would be the best way to help you?

kishorekasi:
Just back from vacation. There seems to be new commits to the repo. I will resolve the conflicts and generate a new pull request.

merlimat:
Hi @kishorekasi, is this the latest version of the patch you were mentioning yesterday?

kishorekasi:
Yes @merlimat

merlimat:
Ok, I see only ~5 tests are failing there, not the 100s I understood yesterday :) 

https://builds.apache.org/job/bookkeeper-master-git-pullrequest/309/

merlimat:
Let me take a look

kishorekasi:
I fixed a merge conflict. Also as Netty channel and write and flush semantics have changed in 4.x; made few more changes.

eolivelli:
On my laptop (Linux fedora 24) I have:

> testDuplicateBookieServerStartup(org.apache.bookkeeper.bookie.BookieInitializationTest)  Time elapsed: 0.028 sec  <<< ERROR!
> io.netty.channel.unix.Errors$NativeIoException: bind() failed: Address already in use
> 
The test is failing because using the NativeIoException is not a BindException
I suggest to change the 'catch' clause to a more naive 'catch IOException'

eolivelli:
@kishorekasi adding simple debug logs fixes some error on the SlowBookieTest, I think we have race conditions on ByteBuf usage, I am checking now the flow of the pendingOps in case of network error.

kishorekasi:
@eolivelli - I fixed a refcnt bug with ByteBuf and consistently pass SlowBookieTest now; updated the pull request. I am not able to see the jenkins build output, waiting for this next run to address if there are any more test failures.

eolivelli:
I have checked out the lastest version of this PR But I am still seeing this kind of errors, for instance in BookieFailuresTest. Did you push your local code to GitHub ?
```

Tests in error: 
  BookieFailureTest.testAsyncBK4:121->auxTestReadWriteAsyncSingleClient:183->Object.wait:-2 » TestTimedOut
```

> 2017-04-28 13:09:43,496 - ERROR - [BookKeeperClientWorker-orderedsafeexecutor-0-0:SafeRunnable@33] - Unexpected throwable caught 
> io.netty.util.IllegalReferenceCountException: refCnt: 0, increment: 1
>         at io.netty.buffer.AbstractReferenceCountedByteBuf.retain0(AbstractReferenceCountedByteBuf.java:68)
>         at io.netty.buffer.AbstractReferenceCountedByteBuf.retain(AbstractReferenceCountedByteBuf.java:53)
>         at io.netty.buffer.CompositeByteBuf.retain(CompositeByteBuf.java:1906)
>         at io.netty.buffer.CompositeByteBuf.retain(CompositeByteBuf.java:44)
>         at org.apache.bookkeeper.proto.BookieClient.addEntry(BookieClient.java:223)
>         at org.apache.bookkeeper.client.PendingAddOp.sendWriteRequest(PendingAddOp.java:103)
>         at org.apache.bookkeeper.client.PendingAddOp.unsetSuccessAndSendWriteRequest(PendingAddOp.java:169)
>         at org.apache.bookkeeper.client.LedgerHandle.unsetSuccessAndSendWriteRequest(LedgerHandle.java:1347)
>         at org.apache.bookkeeper.client.LedgerHandle$ChangeEnsembleCb.safeOperationComplete(LedgerHandle.java:1211)
>         at org.apache.bookkeeper.client.LedgerHandle$ChangeEnsembleCb.safeOperationComplete(LedgerHandle.java:1181)
>         at org.apache.bookkeeper.util.OrderedSafeExecutor$OrderedSafeGenericCallback$1.safeRun(OrderedSafeExecutor.java:443)
>         at org.apache.bookkeeper.util.SafeRunnable.run(SafeRunnable.java:31)
>         at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
>         at java.util.concurrent.FutureTask.run(FutureTask.java:266)
>         at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:180)
>         at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:293)
>         at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
>         at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
>         at java.lang.Thread.run(Thread.java:745)
>

eolivelli:
@kishorekasi  Maybe you can add commits without squashing, this way we can track more easily the  new changes you are pushing in. Thanks

bookkeeper-server/pom.xml
eolivelli: @revans2  @sijie @kishorekasi I think it is better to drop the  linux-x86_64 classifier, so that we will import only the pura Java code without any arch specific dependency

bookkeeper-server/src/test/java/org/apache/bookkeeper/test/BookieClientTest.java
eolivelli: @kishorekasi this is the cause of the failure of test testWriteGaps() (I remember the same problem with that test was solved by @sijie some time ago)
The fix is very simple, just update the method with the current master code adapted to ByteBuf
```
        ByteBuf bb = Unpooled.buffer(4 + 24);
        bb.writeLong(lid);
        bb.writeLong(eid);
        bb.writeLong(eid-1);
        bb.writeInt(i);
        return bb;
```
I hope that helps

eolivelli:
I am locally testing a more cleaner dependency set against netty

```

<dependency>
            <groupId>io.netty</groupId>
            <artifactId>netty-common</artifactId>
            <version>${netty.version}</version>
        </dependency>
        <dependency>
            <groupId>io.netty</groupId>
            <artifactId>netty-codec</artifactId>
            <version>${netty.version}</version>
        </dependency>
        <dependency>
            <groupId>io.netty</groupId>
            <artifactId>netty-transport</artifactId>
            <version>${netty.version}</version>
        </dependency>
        <dependency>
            <groupId>io.netty</groupId>
            <artifactId>netty-handler</artifactId>
            <version>${netty.version}</version>
        </dependency>
         <dependency>
            <groupId>io.netty</groupId>
            <artifactId>netty-transport-native-epoll</artifactId>
            <version>${netty.version}</version>
            
        </dependency>
    <dependency>
      <groupId>io.netty</groupId>
      <artifactId>netty-tcnative-boringssl-static</artifactId>
      <version>${netty-tcnative.version}</version>
      <!--<classifier>${os.detected.classifier}</classifier>-->
    </dependency>
```

eolivelli:
@kishorekasi 
I noticed we are using Netty 4.1.7, on 4.1.9 release notes I see https://github.com/netty/netty/issues/6303

see 
http://netty.io/news/2017/03/10/4-0-45-Final-4-1-9-Final.html

On my projects I am using 4.1.9 now, maybe you can try to upgrade as well

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/TestGetBookieInfoTimeout.java
```diff
@@ -62,8 +65,17 @@ public TestGetBookieInfoTimeout() {
     @Before
     public void setUp() throws Exception {
         super.setUp();
-        channelFactory = new NioClientSocketChannelFactory(Executors.newCachedThreadPool(), Executors
-                .newCachedThreadPool());
+        if (SystemUtils.IS_OS_LINUX) {
+        	try {
+        		eventLoopGroup = new EpollEventLoopGroup();
+        	} catch (Exception e) {
```
eolivelli: you MUST  catch (Throwable) and not Exception, as you do in other parts.
Maybe it would be better to have a common utility function to create all the EventLoopGroups

in fact running without the native epoll support libraries on the library path make the test fail

eolivelli:
with the dependency set (with no epoll native libs, only java dep and with the needed netty parts) which I suggested in previous comments and the fix to testWriteGaps I get only 5 failures, maybe due to the bug on ref counted buffers.

eolivelli:
I got these other errors which can be due to some resource leak
```
recoverWithoutPasswordInConf[7](org.apache.bookkeeper.client.BookieRecoveryTest)  Time elapsed: 0.299 sec  <<< ERROR!
java.lang.IllegalStateException: failed to create a child event loop
        at org.apache.bookkeeper.client.BookieRecoveryTest.recoverWithoutPasswordInConf(BookieRecoveryTest.java:846)
Caused by: io.netty.channel.ChannelException: failed to open a new selector
        at org.apache.bookkeeper.client.BookieRecoveryTest.recoverWithoutPasswordInConf(BookieRecoveryTest.java:846)
Caused by: java.io.IOException: Too many open files
        at org.apache.bookkeeper.client.BookieRecoveryTest.recoverWithoutPasswordInConf(BookieRecoveryTest.java:846)

testSyncBookieRecoveryToSpecificBookie[7](org.apache.bookkeeper.client.BookieRecoveryTest)  Time elapsed: 2.27 sec  <<< ERROR!
java.lang.IllegalStateException: failed to create a child event loop
        at org.apache.bookkeeper.client.BookieRecoveryTest.setUp(BookieRecoveryTest.java:112)
Caused by: io.netty.channel.ChannelException: failed to open a new selector
        at org.apache.bookkeeper.client.BookieRecoveryTest.setUp(BookieRecoveryTest.java:112)
Caused by: java.io.IOException: Too many open files
        at org.apache.bookkeeper.client.BookieRecoveryTest.setUp(BookieRecoveryTest.java:112)

```

eolivelli:
Thank you @kishorekasi 
now running the full suite I get sometimes this error 
```
Caused by: java.io.IOException: Too many open files
        at sun.nio.ch.IOUtil.makePipe(Native Method)
        at sun.nio.ch.EPollSelectorImpl.<init>(EPollSelectorImpl.java:65)
        at sun.nio.ch.EPollSelectorProvider.openSelector(EPollSelectorProvider.java:36)
        at io.netty.channel.nio.NioEventLoop.openSelector(NioEventLoop.java:155)
        ... 50 more

```

Maybe we have some leak

merlimat:
I think there was a socket leak in the original commit. The leak was fixed in a subsequent commit in the `yahoo-4.3` branch: yahoo/bookkeeper@f19c1d8c67caec2006962574adcf483d40fe6dc7

I have rebased that commit on this PR at https://github.com/merlimat/bookkeeper/commit/68a711577fe8121f5179462a0c5ee0fa61ddb86d

@kishorekasi Can you apply this commit to your branch?

bookkeeper-server/pom.xml
merlimat: I think we should stick with `netty-all` dependency since it will include the binaries for all supported platforms.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/AuthHandler.java
```diff
@@ -307,20 +297,18 @@ public void messageReceived(ChannelHandlerContext ctx,
...
+                    BookkeeperProtocol.Request req = (BookkeeperProtocol.Request) msg;
                     if (req.getHeader().getOperation()
                             == BookkeeperProtocol.OperationType.AUTH) {
-                        super.writeRequested(ctx, e);
+                        super.write(ctx, msg, promise);
                     } else {
-                        waitingForAuth.add(e);
+                        waitingForAuth.add(msg);
                     }
                 } // else just drop
```
merlimat: I think the implementation of the auth provider here was to accumulate all the responses while the auth handshake is being performed.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieClient.java
```diff
@@ -423,26 +423,21 @@ public void writeComplete(int rc, long ledger, long entry, BookieSocketAddress a
         Counter counter = new Counter();
         byte hello[] = "hello".getBytes(UTF_8);
         long ledger = Long.parseLong(args[2]);
-        ThreadFactoryBuilder tfb = new ThreadFactoryBuilder();
-        ClientSocketChannelFactory channelFactory = new NioClientSocketChannelFactory(
-                Executors.newCachedThreadPool(tfb.setNameFormat(
-                        "BookKeeper-NIOBoss-%d").build()),
-                Executors.newCachedThreadPool(tfb.setNameFormat(
-                        "BookKeeper-NIOWorker-%d").build()));
+        EventLoopGroup eventLoopGroup = new NioEventLoopGroup(1);
```
merlimat: This was just on the `main()` method that is not really used in practice anywhere

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieNettyServer.java
```diff
@@ -84,24 +105,42 @@
...
         if (!conf.isDisableServerSocketBind()) {
-            channels.add(new NioServerSocketChannelManager());
+            ThreadFactory threadFactory = new ThreadFactoryBuilder().setNameFormat("bookie-io-%s").build();
+            final int numThreads = Runtime.getRuntime().availableProcessors() * 2;
+
+            EventLoopGroup eventLoopGroup;
+            if (SystemUtils.IS_OS_LINUX) {
+                try {
+                    eventLoopGroup = new EpollEventLoopGroup(numThreads, threadFactory);
+                } catch (Throwable t) {
```
merlimat: In Pulsar we are catching multiple exceptions: 

`ExceptionInInitializerError | NoClassDefFoundError | UnsatisfiedLinkError e`

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieNettyServer.java
```diff
@@ -112,19 +151,121 @@ boolean isRunning() {
...
+                    pipeline.addLast("lengthprepender", new LengthFieldPrepender(4));
+    
+                    pipeline.addLast("bookieProtoDecoder", requestDecoder);
+                    pipeline.addLast("bookieProtoEncoder", responseEncoder);
+                    pipeline.addLast("bookieAuthHandler", new AuthHandler.ServerSideHandler(contextHandler.getConnectionPeer(), authProviderFactory));
+    
+                    ChannelInboundHandler requestHandler = isRunning.get()
+                            ? new BookieRequestHandler(conf, requestProcessor, allChannels) : new RejectRequestHandler();
+                    pipeline.addLast("bookieRequestHandler", requestHandler);
+
```
merlimat: It's just being passed to the `AuthHandler.ServerSideHandler` at line 206: 

```java
pipeline.addLast("bookieAuthHandler", 
     new AuthHandler.ServerSideHandler(contextHandler.getConnectionPeer(), 
         authProviderFactory));
```

merlimat:
@kishorekasi @eolivelli @revans2 Addressed Bobby comments in a new commit on top of this PR. 

https://github.com/merlimat/bookkeeper/commits/netty-4.1

kishorekasi:
Thanks @merlimat I will pick those changes

eolivelli:
@merlimat  @kishorekasi 
Even on @merlimat brach I get a lot of too many open files, like this
```
2017-05-03 09:33:12,232 - ERROR - [ZKC-connect-executor-0:ZooKeeperClient$1$1@125] - Failed to create zookeeper instance to 127.0.0.1:15000
java.io.IOException: Couldn't instantiate org.apache.zookeeper.ClientCnxnSocketNIO
        at org.apache.zookeeper.ZooKeeper.getClientCnxnSocket(ZooKeeper.java:2732)
        at org.apache.zookeeper.ZooKeeper.<init>(ZooKeeper.java:728)
        at org.apache.zookeeper.ZooKeeper.<init>(ZooKeeper.java:786)
        at org.apache.zookeeper.ZooKeeper.<init>(ZooKeeper.java:653)
        at org.apache.bookkeeper.zookeeper.ZooKeeperClient.createZooKeeper(ZooKeeperClient.java:324)
        at org.apache.bookkeeper.zookeeper.ZooKeeperClient$1$1.call(ZooKeeperClient.java:123)
        at org.apache.bookkeeper.zookeeper.ZooKeeperClient$1$1.call(ZooKeeperClient.java:114)
        at org.apache.bookkeeper.zookeeper.ZooWorker.syncCallWithRetries(ZooWorker.java:141)
        at org.apache.bookkeeper.zookeeper.ZooKeeperClient$1.call(ZooKeeperClient.java:114)
        at org.apache.bookkeeper.zookeeper.ZooKeeperClient$1.call(ZooKeeperClient.java:109)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
        at java.lang.Thread.run(Thread.java:745)
Caused by: java.io.IOException: Too many open files
        at sun.nio.ch.EPollArrayWrapper.epollCreate(Native Method)
        at sun.nio.ch.EPollArrayWrapper.<init>(EPollArrayWrapper.java:130)
        at sun.nio.ch.EPollSelectorImpl.<init>(EPollSelectorImpl.java:69)
        at sun.nio.ch.EPollSelectorProvider.openSelector(EPollSelectorProvider.java:36)
        at java.nio.channels.Selector.open(Selector.java:227)
        at org.apache.zookeeper.ClientCnxnSocketNIO.<init>(ClientCnxnSocketNIO.java:44)
        at sun.reflect.GeneratedConstructorAccessor7.newInstance(Unknown Source)
        at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
        at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
        at java.lang.Class.newInstance(Class.java:442)
        at org.apache.zookeeper.ZooKeeper.getClientCnxnSocket(ZooKeeper.java:2730)
        ... 13 more

```
I'm running tests on two different laptops.
Don't you get errors like those ?

revans2:
@eolivelli I get them too.  It may be differences in the default value of ulimit.  Probably the simplest thing to do it to try and trace down who is leaking all of the file descriptors.  It most likely is a missed close and falling back to GC to issue the close.

kishorekasi:
@revans2 I have seen "too many open files" in the past. Recently I have not seen them. But, I had to increase ulimit on my dev machines for a different reason. That explains why I am not seeing them locally

merlimat:
I didn't see the problem locally either. I'm trying now with a file leak detector agent: http://file-leak-detector.kohsuke.org/ to see if it gives any clue.

merlimat:
It seems entry log file descriptors are being leaked: https://gist.github.com/merlimat/18ecfc83b7312625755610d4d59bac9b

Investigating on why

merlimat:
I have been testing quite a while with the leak detector and across all the leaked files I haven't seen any since netty 4 created socket.

What I mostly saw with the file leak detector has been a whole bunch of file leaks in different tests. Eg: 
 * Tests that instantiate bookies and call `shutdown()` without calling `start()` before are creating and initializing the ledger storage but not closing it, leaking threads and several fds
 * ZooKeeperClient sometimes doesn't shutdown the zk handle if the test completes too quickly, leaking sockets.
 * Several tests are passing bad config, so the bookie/client start gets exception (on purpose) and then doesn't clean up some partial objects.
 * ... 

I don't think there is an easy fix for the issue. A bigger refactor would be needed to address them. 

Also, I've noticed that in `pom.xml` there was already:
```xml
<forkMode>always</forkMode>
```

This directive was intended to use a fresh JVM fork for each test, but it's deprecated in maven and it doesn't seem to be working. 

I'd propose to fix that and to use in the surefire configuration:

```xml
<reuseForks>true</reuseForks>
```

I'll create a separate Jira for this.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
```diff
@@ -253,6 +251,7 @@ void submitCallback(final int rc) {
             addOpLogger.registerSuccessfulEvent(latencyNanos, TimeUnit.NANOSECONDS);
         }
         cb.addComplete(rc, lh, entryId, ctx);
+        ReferenceCountUtil.release(toSend);
```
merlimat: I don't think this change is necessary. We can release the before before trigger the callback, since we are not going to touch it anymore at this point.

merlimat:
@kishorekasi can you also rebase on top of last master? This should reduce the count of tests failing

merlimat:
@kishorekasi I have fixed the issues with the `BookieFailureTest`. I have applied multiple ref-count fixes as well from the yahoo branch.. pushed to https://github.com/merlimat/bookkeeper/commits/netty-4.1
Please take a look. I'll try to get to the others tests now.

merlimat:
@kishorekasi  I'll push my branch in a new PR to trigger Jenkins with my changes as well

kishorekasi:
@eolivelli @sijie @merlimat With my latest commit this evening, all tests seem to pass now. Let me know if you are seeing any test failures in your local runs.

eolivelli:
@kishorekasi it is great to see all tests now passing !

@kishorekasi did you take a look to latest @merlimat commit on this issue ?

@kishorekasi @sijie @revans2  Now that all the tests are passing I think we should restore the refactor of https://issues.apache.org/jira/browse/BOOKKEEPER-930, the change is quite simple, it is just a better code, see the PR https://github.com/apache/bookkeeper/pull/49 and the discussion about the refactor

kishorekasi:
@eolivelli I have not looked into @merlimat changes. I will take a look and add them if they are not already addressed.

merlimat:
@kishorekasi I get a clean build on mac from my branch. I'm looking at the reasons it's failing with netty-all dependency on linux. I think we shouldn't be forcing the nio transport on linux.

merlimat:
@kishorekasi I think I fixed the epoll test issues. Added a couple of small commits on my branch. All tests are now passing on local linux.

sijie:
@kishorekasi I merged #138 .  do you mind closing this pull request?

PR #118, opened 2017-03-23
BOOKKEEPER-1013 - Fix findbugs errors on latest master

sijie:
- 100% synchronization on _explicitLac_ on FileInfo
- 100% synchronization on _length_ on LedgerHandle
- fixed the print format issue on journal dir names array
- fix checkpointsource list hashCode, equals()


jvrao:
LGTM +1

PR #121, opened 2017-03-31
BOOKKEEPER-1019 Support for reading entries after LAC

eolivelli:
This patch introduces a new client-side configuration option to allow reads outside the boundary of the local LastAddConfirmed value.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
@@ -470,12 +470,36 @@ public class LedgerHandle implements AutoCloseable {
      * @param lastEntry
      *          id of last entry of sequence (included)
      *
+     * @see #asyncReadEntries(long, long, org.apache.bookkeeper.client.AsyncCallback.ReadCallback, java.lang.Object)
      */
     public Enumeration<LedgerEntry> readEntries(long firstEntry, long lastEntry)
             throws InterruptedException, BKException {
         CompletableFuture<Enumeration<LedgerEntry>> counter = new CompletableFuture<>();
 
         asyncReadEntries(firstEntry, lastEntry, new SyncReadCallback(), counter);
+
+        return SynchCallbackUtils.waitForResult(counter);
+    }
+
+    /**
+     * Read a sequence of entries synchronously, allowing to read after the LastAddConfirmed range.<br>
+     * This is the same of
+     * {@link #asyncReadUnconfirmedEntries(long, long, org.apache.bookkeeper.client.AsyncCallback.ReadCallback, java.lang.Object) }
+     *
+     * @param firstEntry
+     *          id of first entry of sequence (included)
+     * @param lastEntry
+     *          id of last entry of sequence (included)
+     *
+     * @see #readEntries(long, long)
+     * @see #asyncReadUnconfirmedEntries(long, long, org.apache.bookkeeper.client.AsyncCallback.ReadCallback, java.lang.Object)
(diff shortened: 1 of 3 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -481,6 +481,28 @@ public String toString() {
     }
 
     /**
+     * Read a sequence of entries synchronously. This is the same of
+     * {@link #readEntries(long, long)  }
+     * but it lets the client to read without checking the local value of LastAddConfirmed, so that it is possibile to
+     * read entries for which the writer has not received the acknowledge yet
```
sijie: Please document the behaviors on a few cases:

- reading entries within LAC
- reading entries beyond LAC
- reading entries (some within LAC, some beyond LAC)

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookKeeperTest.java
```diff
@@ -432,5 +432,126 @@ public void testReadHandleWithExplicitLAC() throws Exception {
         rlh.close();
         wlh.close();
         bkcWithExplicitLAC.close();
-    }	
+    }
+
+    @Test(timeout = 60000)
+    public void testReadAfterLastAddConfirmed() throws Exception {
```
sijie: add test cases about read behaviors that I mentioned as above

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookKeeperTest.java
```diff
@@ -432,5 +432,126 @@ public void testReadHandleWithExplicitLAC() throws Exception {
...
+                Assert.assertTrue(
+                    "Expected LAC of rlh: " + (numOfEntries - 2) + " actual LAC of rlh: " + rlh.getLastAddConfirmed(),
+                    (rlh.getLastAddConfirmed() == (numOfEntries - 2)));
+
+                Assert.assertFalse(writeLh.isClosed());
+
+                // without readUnconfirmedEntries we are not able to read all of the entries
+                try {
+                    rlh.readEntries(0, numOfEntries - 1);
+                    fail();
```
sijie: please add message in "fail()" function.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookKeeperTest.java
```diff
@@ -432,5 +432,126 @@ public void testReadHandleWithExplicitLAC() throws Exception {
...
```
sijie: please add message in "fail()" function.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookKeeperTest.java
```diff
@@ -432,5 +432,126 @@ public void testReadHandleWithExplicitLAC() throws Exception {
...
+                    Assert.assertTrue("Expected entry String: " + ("foobar" + entryId)
+                        + " actual entry String: " + entryString,
+                        entryString.equals("foobar" + entryId));
+                    entryId++;
+                }
+            }
+
+            try {
+                writeLh.close();
+                fail();
```
sijie: please add message in "fail()" function.

eolivelli:
@sijie thank you for your review
I have tried to fix Javadocs and be more explicit about all the cases, please check.

I have added new check in the testcase, I hope to have covered all the cases

sijie:
@merlimat do you mind taking a look at this one?

eolivelli:
@jiazhai thank you for your review
we decided not to add a configuration option, but a brand new method readUnconfirmedEntries, this way it is more explicit that the client is reading entries out side of the last-add-confirmed protocol

PR #122, opened 2017-04-03
BOOKKEEPER-1020 Fix Explicit LAC tests on master

eolivelli:


bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookKeeperTest.java
```diff
@@ -401,7 +401,7 @@ public void testReadHandleWithExplicitLAC() throws Exception {
         // lh.getExplicitLastAddConfirmed() will be <
         // lh.getPiggyBackedLastAddConfirmed(),
         // so it wont make explicit writelac in the first run
-        Thread.sleep((2 * explictLacInterval + 1) * 1000);
```
govind-menon: Can we rename it to explicitLacIntervalMillis?

eolivelli: @govind-menon done
thank you for your review

PR #125, opened 2017-04-04
BOOKKEEPER-997: Wire protocol change for supporting long poll

sijie:
adding fields in ReadRequest to support piggying back lac on read responses and support long poll requests.

### DIFF
--- a/bookkeeper-server/src/main/proto/BookkeeperProtocol.proto
+++ b/bookkeeper-server/src/main/proto/BookkeeperProtocol.proto
@@ -85,6 +85,7 @@ message Request {
 message ReadRequest {
     enum Flag {
         FENCE_LEDGER = 1;
+        ENTRY_PIGGYBACK = 2;
     }
     optional Flag flag = 100;
     required int64 ledgerId = 1;
@@ -92,6 +93,10 @@ message ReadRequest {
     required int64 entryId = 2;
     // Used while fencing a ledger.
     optional bytes masterKey = 3;
+    // Used for waiting on last add confirmed update
+    optional int64 previousLAC = 4;
+    // Used as a timeout for the long polling request
+    optional int64 timeOut = 5;
 }
 
 message AddRequest {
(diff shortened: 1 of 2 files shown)


eolivelli:
@sijie 
Maybe it will be better to run protobuf compiler e commit new java classes in this commit, so that anyone who will run the compiler will not find inconsistent .java sources

sijie:
@eolivelli that works for me. can you explore the solution for it?

eolivelli:
@sijie 
I meant to run simply this command and "git add" the generated classes to the commit
`mvn clean install -Pprotobuf -DskipTests`

I can pick up your patch if you want

sijie:
oh, I understand what you are saying now. The reason that we didn't make it automatically generate the protobuf classes with the build is not every one installed protoc. If there's a plugin can automatically install protoc, I am good with that. 

For this pull requests, I will manually generate the proto classes. Feel free to create a JIRA for automate this in the build.

eolivelli:
I think that to generate  automatically sources is not useful and would keep the build less reproducible and slow.

I was saying that usually we commit the generated clasdses together with the change to the protocol file.
I image you are going to do it in another commit but it would ve clearer if we do it in thia commit.

sijie:
@eolivelli okay. so I think we are in the same page. I will add the generated protobuf classes in this review. I forgot that.

sijie:
add the generated protobuf file.

sijie:
tag @merlimat for reviews

bookkeeper-server/src/main/proto/BookkeeperProtocol.proto
```diff
@@ -85,13 +85,18 @@ message Request {
...
     }
     optional Flag flag = 100;
     required int64 ledgerId = 1;
     // entryId will be -1 for reading the LAST_ADD_CONFIRMED entry.
     required int64 entryId = 2;
     // Used while fencing a ledger.
     optional bytes masterKey = 3;
+    // Used for waiting on last add confirmed update
+    optional int64 previousLAC = 4;
+    // Used as a timeout for the long polling request
```
merlimat: Please add time unit for the timeout. Just a mention in the comment would be good

sijie: done.

sijie:
ping @merlimat

PR #126, opened 2017-04-04
BOOKKEEPER-1018: Allow client to select older V2 protocol (no protobuf)

govind-menon:
Tested manually - running all tests locally. Will tag contributors. when ready for review (for now putting up to test via Screwdriver and manual review)

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
@@ -18,11 +18,15 @@
 package org.apache.bookkeeper.proto;
 
 import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.MappedByteBuffer;
 import java.nio.channels.ClosedChannelException;
 import java.util.ArrayDeque;
 import java.util.Collections;
+import java.util.Collection;
 import java.util.Queue;
 import java.util.Set;
+import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
@@ -98,7 +102,7 @@ import com.google.common.collect.Sets;
 import com.google.protobuf.ByteString;
 import com.google.protobuf.ExtensionRegistry;
 import java.net.SocketAddress;
-import java.util.Collection;
+
 import org.apache.bookkeeper.auth.BookKeeperPrincipal;
 import org.jboss.netty.channel.ChannelFactory;
 import org.apache.bookkeeper.client.ClientConnectionPeer;
@@ -146,6 +150,8 @@ public class PerChannelBookieClient extends SimpleChannelHandler implements Chan
     private final OpStatsLogger readLacTimeoutOpLogger;
     private final OpStatsLogger getBookieInfoOpLogger;
     private final OpStatsLogger getBookieInfoTimeoutOpLogger;
+
+    private final boolean useV2WireProtocol;
 
     /**
      * The following member variables do not need to be concurrent, or volatile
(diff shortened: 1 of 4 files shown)


govind-menon:
Tagging @sijie and @eolivelli  for review. Will amend test cases in a follow up.

eolivelli:
@sijie  @govind-menon @merlimat 
IMHO we should at least add a test case for the most simple cases (addEntry, readEntries, readLAC...) + basic auth

So that in the future we will not forget about this "legacy" option, as actual is not "legacy" at all if Yahoo/Pulsar is using it in production
I wonder if there is a way to automatically run the full suite with the legacy protocol and if this is really "interesting".

Issue:
should we implement all new features on the v2 protocol as well ? 
I'm thinking for instance to STARTTLS and to Long Poll Reads which have not been merged yet

revans2:
@eolivelli I totally agree we do need some test cases for this.  The reason that we use the older API is because of the large amount of object allocation that happens with protocol buffers.  I have heard similar complains when HDFS moved to protocol buffers, but it was not fixed there either.

eolivelli:
@revans2  I'm really OK with supporting the legacy protocol, let's find a way to guarantee the quality of that mode too.
Maybe we could add  some system property to the set suite and configure a Jenkins Job to periodically run the whole suite with the V2 protocol
@sijie  @fpj   do you think it will be possible ?

govind-menon:
@eolivelli I can add a V2 test suite (there may be some duplication which I'll try to refactor out).

eolivelli:
@govind-menon +1 for me maybe just cover the basic feature (addEntry, readEntries, readLAC...) + basic auth

we will need to add new testcases for future protocol changes as well mabye we should discuss this in a separate email thread

merlimat:
@eolivelli 
> So that in the future we will not forget about this "legacy" option, as actual is not "legacy" at all if Yahoo/Pulsar is using it in production

I think that for long term we should converge to one single wire protocol that is both extensible and resource efficient. 

BK v2 protocol has its own issues as well (which is why protobuf was introduced): 
 * Fixed binary protocol, cannot add new fields and have easy upgrade
 * Doesn't have a request id. For example if a client tries to read the same entry twice, it cannot distinguish the response for each request, thus needing additional logic in the client to keep track of it.

sijie:
@merlimat @eolivelli - what is the deal for this pr? shall we add the tests now or doing it in a separated jira?

govind-menon:
I was thinking of doing it in a separate JIRA. @merlimat  Had suggested a way of running the existing test suite with the V2 protocol set.

eolivelli:
+1 for separate JIRA

Overall LGTM

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
```diff
@@ -803,6 +856,7 @@ void errorOutReadKey(final CompletionKey key) {
     }
 
     void errorOutReadKey(final CompletionKey key, final int rc) {
+        LOG.info("Removing completion key: {}", key);
```
eolivelli: @govind-menon can you remove or change to  "debug" level ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
```diff
@@ -835,6 +889,7 @@ void errorOutWriteLacKey(final CompletionKey key) {
     }
 
     void errorOutWriteLacKey(final CompletionKey key, final int rc) {
+        LOG.info("Removing completion key: {}", key);
```
eolivelli: @govind-menon can you remove or change to  "debug" level ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
```diff
@@ -859,6 +914,7 @@ void errorOutReadLacKey(final CompletionKey key) {
     }
 
     void errorOutReadLacKey(final CompletionKey key, final int rc) {
+        LOG.info("Removing completion key: {}", key);
```
eolivelli: @govind-menon can you remove or change to  "debug" level ?

eolivelli:
I'm sorry I have missed some LOG.info lines
Maybe can you @govind-menon  remove them or change to 'debug' level ?

govind-menon:
@eolivelli @sijie Comments addressed.

govind-menon:
@sijie Good to merge? As we discussed in the meeting the test cases and other parts of the API will be a separate PR

PR #130, opened 2017-04-10
BOOKKEEPER-1031: close the ledger handle in ReplicationWorker.rerepli…

athanatos:
…cate

Otherwise, we build up an unbounded set of Listeners in the
AbstractZkLedgerManager listenerSet structure which never go
away.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/ReplicationWorker.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/ReplicationWorker.java
@@ -225,9 +225,60 @@ public class ReplicationWorker implements Runnable {
     private boolean rereplicate(long ledgerIdToReplicate) throws InterruptedException, BKException,
             UnavailableException {
         LOG.debug("Going to replicate the fragments of the ledger: {}", ledgerIdToReplicate);
-        LedgerHandle lh;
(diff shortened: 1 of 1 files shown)


eolivelli:
@sijie can you add Samuel to the contributors in JIRA so that the issue can be assigned to him?

athanatos:
Build's failing due to

Tests in error:
  BookieRecoveryTest.ensurePasswordUsedForOldLedgers[4] » BKNoSuchLedgerExists
  BookieRecoveryTest.ensurePasswordUsedForOldLedgers[5] » BKNoSuchLedgerExists

which seems to be broken in master.

I did a bisect and found that

https://github.com/apache/bookkeeper/commit/42e8f1294f206cbe51a6af669cf605833b78bf42 introduced both a hang in that test and that test failure.  https://github.com/apache/bookkeeper/commit/32ebf0ad68f85daffdec571e28206339a4d09443 fixed the hang uncovering the test failure.

sijie:
@athanatos I think those two are known failures. I think @eolivelli knew about it.

athanatos:
Cool, what's the next step?

eolivelli:
We are going to merge as soon as possible. Thank you

sijie:
merging now

PR #134, opened 2017-04-28
BOOKKEEPER-1042 Deploy SNAPSHOTS of master to public Apache Repository

eolivelli:
Make all pom.xml files have a -SNAPSHOT version.

Currently we have the 'compats' subprojects which have a strange versioning scheme.
Actual versioning scheme for that poms will clash will other versions of the project, because they are fixed and do not depend on the common parent pom.

Changing the version will make the build more reproducible and it is a prerequisite for the deployment of -SNAPSHOT version to the official Apache Snapshots Repository

### DIFF
--- a/bookkeeper-server/pom.xml
+++ b/bookkeeper-server/pom.xml
@@ -151,7 +151,7 @@
     <dependency>
       <groupId>org.apache.bookkeeper</groupId>
       <artifactId>bookkeeper-server-compat400</artifactId>
-      <version>4.0.0</version>
+      <version>${project.version}</version>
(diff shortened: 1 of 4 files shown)


eolivelli:
@merlimat I think that those project need to have the same version as the other modules.
Actually they are used to create artifacts for compat tests but for instance the artifacts which support version 4.5.0  are different from the ones for the 4.4.0 version, see the protobuf and guava upgrades for instance.
If they had to be 'fixed' I think they should be kept outside the project, but I think this is not the case, in fact if it was the case we could have used directly the released artifacts

PR #136, opened 2017-05-04
BOOKKEEPER-1046: Avoid long to Long conversion in OrderedSafeExecutor task submit

merlimat:
When submitting tasks to an OrderedSafeExecutor, most of the time a ledger id is being passed. Given that the method accepts and Object, the primitive `long` is boxed into a `Long` allocated on the heap.

Added specific method overload to directly accept longs as the key in the OrderedSafeExecutor.


eolivelli:
Overall is OK for me, one little change
can you drop the old method as it is no more used ?

`public void submitOrdered(Object orderingKey, SafeRunnable r) {`

BK code does not use it directly, it is an internal class.

eolivelli:
All JUnit tests are passing

merlimat:
> can you drop the old method as it is no more used ?

@eolivelli We still use it in Pulsar/ManagedLedger to have tasks in the executor executed in order by topic name

PR #140, opened 2017-05-09
BOOKKEEPER-1050: Cache journalFormatVersionToWrite when starting Journal

merlimat:
Reading the journal version format from `ServiceConfiguration` each time is inefficient.
`ServiceConfiguration` is based on Java properties which is based on a String to object hashtable. Each read implies acquiring a mutex and converting from object to int.


merlimat:
@eolivelli Good point. This is a merge from a very old commit.. probably the alignment size was not there at the point. I'll fix that.

PR #141, opened 2017-05-09
BOOKKEEPER-1051: Fast shutdown for GarbageCollectorThread

merlimat:
Several unit tests are taking very long time to complete (eg: `BookieLedgerIndexTest` taking ~10 minutes). 
The reason is that these tests are playing with the ZK quorum shutting it down and after the test succeeds, the shutdown phase is taking long time, since we try to do graceful shutdown with 1min wait time.
I think is better to interrupt immediately the garbage collector thread when shutting down the bookie.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
@@ -487,11 +487,9 @@ public class GarbageCollectorThread extends SafeRunnable {
             // Wait till the thread stops compacting
             Thread.sleep(100);
         }
-        gcExecutor.shutdown();
-        if (gcExecutor.awaitTermination(60, TimeUnit.SECONDS)) {
-            LOG.warn("GC executor did not shut down in 60 seconds. Killing");
-            gcExecutor.shutdownNow();
-        }
+
+        // Interrupt GC executor thread
+        gcExecutor.shutdownNow();
     }
 
     /**


merlimat:
Build time reduces from 2h to 30min!

jvrao:
Can someone explain me why it won't affect production bookies?
@eolivelli ?

eolivelli:
@jvrao the main explanation is that we already call shutdownNow if the pool does not end execution within 1 minute.

In my personal usecases I usually kill bookies (kill -9) and do not wait for graceful shutdown

jvrao:
Ok so you are talking about your own production issue. But I don't believe it is right thing to change production code to speed up testing. I believe we need more elaborate fix to take care of junit tests without affecting production code path. @eolivelli @merlimat

eolivelli:
@jvrao  we can add a configuration property for "fast shutdown".

IMHO I don't think  that this change really impact bookies even in case of graceful shutdown.
The prev version already called shutdownNow after a timeout.
With this change a InterruptedException will be more likely to happen, but code should already handle this case, so it won't be an issue.
If we find cases of buggy handling of InterruptedException then we should fix them as the InterruptedException can be thrown even without this change.

I have just commented about this new behavior just for completeness and because I worried about this case, but from my point of view this is a safe change

merlimat:
@jvrao sorry for merging this quickly, should have waited for more feedback. 

The change here is to interrupt the Bookie GC thread immediately when shutting down. The interruption only happens if that thread is waiting on _something_. For the GC thread that means waiting on ZK reads. 

I don't see any problem in shutting down that thread quickly, because the GC anyway picks it up again after restart.

The other part is waiting until the thread is done doing its work. That we are not doing even today, given that bookie GC can take minutes or hours.

PR #146, opened 2017-05-10
BOOKKEEPER-1058: Ignore already deleted ledger on replication audit

merlimat:
Replication auditor should skip ledgers that were deleted since the auditing was started.


jvrao:
@athanatos did we do the same fix internally?

PR #149, opened 2017-05-10
BOOKKEEPER-1061: BookieWatcher should not do ZK blocking operations from ZK async callback thread

merlimat:
In some cases, the BookieWatcher can get the ZK event thread stuck. This happens when a ZK blocking request is issued from a ZK callback thread.

We should decouple the blocking requests in a separate executor to avoid deadlocking ZK client.


jvrao:
@athanatos FYI

PR #152, opened 2017-05-11
BOOKKEEPER-1065: OrderedSafeExecutor should only have 1 thread per bucket

merlimat:


merlimat:
Ok, this is the error that findbugs gives: 

```
[INFO] --- findbugs-maven-plugin:3.0.4:check (default-cli) @ bookkeeper-server ---
[INFO] BugInstance size is 1
[INFO] Error size is 0
[INFO] Total bugs: 1
[INFO] Futile attempt to change max pool size of ScheduledThreadPoolExecutor in new org.apache.bookkeeper.util.OrderedSafeExecutor(String, int, ThreadFactory, StatsLogger, boolean, long) [org.apache.bookkeeper.util.OrderedSafeExecutor] At OrderedSafeExecutor.java:[line 186] DMI_FUTILE_ATTEMPT_TO_CHANGE_MAXPOOL_SIZE_OF_SCHEDULED_THREAD_POOL_EXECUTOR
[INFO] 
```

eolivelli:
Yes recently we removed a similar line due to findbugs warning

merlimat:
I got a bit scared when I saw `ScheduledThreadPoolExecutor` is pass `MAX_INT` as the max thread pool size.. but I think the current code is correct. 

I just did a quick verification and indeed it doesn't create more than one thread :

```java
public static void main(String[] args) {
        ScheduledThreadPoolExecutor e = new ScheduledThreadPoolExecutor(1,
                new ThreadFactoryBuilder().setNameFormat("test-%d").build());

        for (int i = 0; i < 10; i++) {
            final int n = i;
            e.submit(() -> {
                System.out.println( Thread.currentThread().getName() + " |  Start task " + n);
                try {
                    Thread.sleep(1000);

                    System.out.println( Thread.currentThread().getName() + " |  Done task " + n);
                } catch (InterruptedException e1) {
                    e1.printStackTrace();
                }
            });
        }

        e.shutdown();
    }
```

Closing this PR as no fix is needed

PR #153, opened 2017-05-11
BOOKKEEPER-1066: Introduce GrowableArrayBlockingQueue

merlimat:
In multiple places, (eg: journal, ordered executor, etc..), we are using `LinkedBlockingQueue` instances to pass objects between threads.

The `LinkedBlockingQueue` differs from the `ArrayBlockingQueue` in that it doesn't require to define a max queue size, though, being implemented with a linked list, it requires to allocates list nodes each time an item is added.

We can use a `GrowableArrayBlockingQueue` that behaves in the same way as the `LinkedBlockingQueue`, but it's implemented with an array that can be resized when the queue reaches the capacity.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/util/MathUtils.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/util/MathUtils.java
@@ -35,6 +35,10 @@ public class MathUtils {
 
     }
 
+    public static int findNextPositivePowerOfTwo(final int value) {
+        return 1 << (32 - Integer.numberOfLeadingZeros(value - 1));
+    }
+
     /**
      * Current time from some arbitrary time base in the past, counting in
      * milliseconds, and not affected by settimeofday or similar system clock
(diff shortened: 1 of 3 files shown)


jvrao:
Interesting and on the paper sounds great. But do we have any perf testing or benchmark? I believe we need to have a benchmark / ref to run and compare any perf improvements. I have seen extremely interesting results for changes which made absolute sense on paper. :)

merlimat:
@jvrao I don't have a benchmark for this specific change, though I agree it might make sense to add one.

The points here are:
 * In some places we are using `LinkedBlockingQueue` which uses the same head/tail mutex schema but with a linked list instead of the array
 * I don't expect the throughput / cpu utilization to be different between the 2
 * This implementation doesn't require to create a LinkedList node for each value, so it creates no garbage. This can be significant since we are enqueuing quite a bit of items (Journal thread, force-sync thread, OrderedSafeExcutor). For each BK request there were several of these (cannot remember exact number, though I believe ~3 in Bookie and another ~3 in client)

PR #154, opened 2017-05-15
BOOKKEEPER-1067: Add Prometheus stats provider

merlimat:
Prometheus (https://prometheus.io) is a metrics collection system, similar but much more flexible than graphite.

It would be good to expose the Bookie and BookKeeper client stats directly so that a Prometheus instance can collect them (and also check the process status and add alerts).


jvrao:
Awesome!! @kishorekasi

PR #156, opened 2017-05-15
BOOKKEEPER-1055: Optimize handling of masterKey in case it is empty

merlimat:
On each request client and bookies are exchanging the ledger masterKey, which is a 20 bytes MAC digest of the ledger password.

For each request there is a considerable overhead in allocating byte arrays when parsing the add/read requests. 

If the client is a passing an empty password, we should optimize the data path to skip all allocations (related to the masterKey) and instead rely on a static byte array.


eolivelli:
I was trying to merge, but FindBugs check failed, 
@merlimat can you fix the warnings ?

```
[INFO] BugInstance size is 3
[INFO] Error size is 0
[INFO] Total bugs: 3
[INFO] new org.apache.bookkeeper.client.MacDigestManager(long, byte[]) may expose internal representation by storing an externally mutable object into MacDigestManager.passwd [org.apache.bookkeeper.client.MacDigestManager] At MacDigestManager.java:[line 63] EI_EXPOSE_REP2
[INFO] org.apache.bookkeeper.client.MacDigestManager.DIGEST_ALGORITHM isn't final but should be [org.apache.bookkeeper.client.MacDigestManager] At MacDigestManager.java:[line 38] MS_SHOULD_BE_FINAL
[INFO] org.apache.bookkeeper.client.MacDigestManager.KEY_ALGORITHM isn't final but should be [org.apache.bookkeeper.client.MacDigestManager] At MacDigestManager.java:[line 39] MS_SHOULD_BE_FINAL

```

merlimat:
@eolivelli Sure, I'll fix that. Thanks for checking it out !

PR #162, opened 2017-05-25
BOOKKEEPER-1075: BK LedgerMetadata: more memory-efficient parsing of configs

sijie:
It is the contribution from Alex Yarmula

commit 9d9d7dd26235a9beda4421b7bed750fea1789076
Author: Alex Yarmula <ak@twitter.com>
Date: Wed Sep 23 05:57:30 2015 -0700

BK LedgerMetadata: more memory-efficient parsing of configs
Looking at the most prevalent client-side memory allocations, I noticed that we allocate 4KB every time we open a ledger. This is caused by allocating a 4KB buffer (in TextFormat.toStringBuilder) to account for the maximum possible Protobufs message, which is unnecessary in our case: we know the exact size of the metadata ( << 500 B) and don't need to allocate more.
TextFormat.merge(Readable, Message.Builder) is the current method we use. This changes to use TextFormat.merge(CharSequence, Message.Builder), which avoids the extra 4K allocation conversion + an extra StringBuilder.

RB_ID=745700

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
@@ -29,6 +29,7 @@ import org.slf4j.LoggerFactory;
 import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.StringReader;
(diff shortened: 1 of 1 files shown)


eolivelli:
+1 LGTM

PR #163, opened 2017-05-25
Bookkeeper-1077: Allow configuration journal/ledger paths for local bookkeeper.

sijie:


bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
```diff
@@ -646,9 +656,18 @@ public ServerConfiguration setJournalDirsName(String[] journalDirs) {
...
+
+    /**
+     * Get dir names to store ledger data
+     *
+     * @return ledger dir names, if not provided return null
+     */
     public String[] getLedgerDirNames() {
         String[] ledgerDirs = this.getStringArray(LEDGER_DIRS);
-        if (null == ledgerDirs) {
+        if ((null == ledgerDirs) || (0 == ledgerDirs.length)) {
```
eolivelli: @sijie 
just to double check:
I wonder if this change can be dangerous, as if the user configures and 'empty' string for the ledgersDirs it will fallback to /tmp without knowing.

I think that the 'cookie' on ZK will prevent the bookie from starting, so this change is safe
Am I correct ?

sijie: For a bookie joining first time, there is no cookie. So cookie doesn't prevent this. 

but the behavior is same as not configuring ledger dirs. I am fine changing the behavior to be more strict but this would have to be a separate jira.

PR #164, opened 2017-05-25
BOOKKEEPER-1078: Local BookKeeper enhancements for testability

sijie:
BookKeeper: Local Bookkeeper enhancements for testability

1. Allow creating local bookies without always starting a zookeeper server - This is required as tests may want to create and use their own instance of a test zookeeper
2. Allow using non default zookeeper host and more importantly non default ZK port
3. Allowing the caller to specify the initial port for the bookies
4. Optionally shutdown bookies when the bookie thread exits


sijie:
This change is based on #163

eolivelli:
@sijie  I have merged #163 but it seems that there is a conflict, can you resolve it ?

sijie:
updated.

eolivelli:
@sijie I am going to merge this, I you are currently doing merges too I can hold off

PR #169, opened 2017-05-28
BOOKKEEPER-1081: a docker file for Apache BookKeeper

jiazhai:
A docker file for Apache BookKeeper.  This docker image will start a new bookie instance.
Steps to run it:
1, start zookeeper service;
2, run this docker image, e.g:
```
$ docker run -t -i -e "ZK_URL=192.168.100.100:2181" bookkeeper/bookie /bin/bash
```

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,24 @@
+FROM java:openjdk-8-jre-alpine
+
+MAINTAINER bookkeeper community 
+
+RUN apk add --no-cache wget bash \
+&& mkdir -p /opt \
+&& cd /opt \
+&& wget -q http://www.apache.org/dist/bookkeeper/bookkeeper-4.4.0/bookkeeper-server-4.4.0-bin.tar.gz \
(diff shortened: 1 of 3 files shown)


docker/Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM java:openjdk-8-jre-alpine
```
eolivelli: Is it working with alpine?
On alpine there are issues with netty native

docker/README.md
```diff
@@ -0,0 +1,17 @@
+# docker-distributedlog
```
eolivelli: Typo from cut and paste from dl?

jiazhai: Thanks. will change it.

docker/Dockerfile
jiazhai: How to verify this? There seems no error reported when I start 3 bookies, and did some simple write.

jiazhai:
Since there is a dup related work, will close this PR first, discuss and collaborate together with @eolivelli and @caiok to make it done.

PR #171, opened 2017-06-01
BOOKKEEPER-1083: Improvements on OrderedSafeExecutor

sijie:
- use listeningscheduledexecutorservice for the threads
    - as a general util class, expose chooseThread to allow applications use specific thread
    - add a method to force shutdown executor


eolivelli:
+1

PR #176, opened 2017-06-01
BOOKKEEPER-1086: Ledger Recovery - Refactor PendingReadOp

sijie:
this change is the first part of improving ledger recovery. it is basically a refactor change, which:

- abstract an interface for LedgerEntryRequest in PendingReadOp
- rename current implementation to SequenceReadRequest, which read the entry in the sequence of quorum.


eolivelli:
@sijie did you tell that we should mark some other contributor for this sequence of patches? Exact names?

sijie:
@eolivelli no this is from my changes.

PR #177, opened 2017-06-01
BOOKKEEPER-1087: Ledger Recovery (part 2) - Add a parallel reading request in PendingReadOp

sijie:
THIS CHANGE IS BASED ON #176 (you can review f0fb89c)

bookkeeper recovery improvement (part-2): add a parallel reading request in PendingReadOp

- add a parallel reading request in PendingReadOp
- allow PendingReadOp to configure whether to do parallel reading or not
- add flag in ClientConfiguration to allow configuring whether to do parallel reading in LedgerRecoveryOp or not.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerRecoveryOp.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerRecoveryOp.java
@@ -49,6 +49,7 @@ class LedgerRecoveryOp implements ReadCallback, AddCallback {
     long entryToRead;
     // keep a copy of metadata for recovery.
     LedgerMetadata metadataForRecovery;
+    boolean parallelRead = false;
 
     GenericCallback<Void> cb;
 
@@ -73,6 +74,11 @@ class LedgerRecoveryOp implements ReadCallback, AddCallback {
         callbackDone = new AtomicBoolean(false);
         this.cb = cb;
         this.lh = lh;
+    }
+
+    LedgerRecoveryOp parallelRead(boolean enabled) {
+        this.parallelRead = enabled;
+        return this;
     }
 
     public void initiate() {
@@ -110,7 +116,7 @@ class LedgerRecoveryOp implements ReadCallback, AddCallback {
         if (!callbackDone.get()) {
             entryToRead++;
             try {
-                new RecoveryReadOp(lh, lh.bk.scheduler, entryToRead, entryToRead, this, null).initiate();
+                new RecoveryReadOp(lh, lh.bk.scheduler, entryToRead, entryToRead, this, null).parallelRead(parallelRead).initiate();
             } catch (InterruptedException e) {
                 readComplete(BKException.Code.InterruptedException, lh, null, null);
             }
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
@@ -104,25 +104,26 @@ class PendingReadOp implements Enumeration<LedgerEntry>, ReadEntryCallback {
          * @return return true if we managed to complete the entry;
          *         otherwise return false if the read entry is not complete or it is already completed before
          */
-        boolean complete(InetSocketAddress host, final ChannelBuffer buffer) {
-            ChannelBufferInputStream is;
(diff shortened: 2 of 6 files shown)


eolivelli:
@sijie  I have merged #176 maybe you will have to rebase

sijie:
@eolivelli I think a better approach is to complete the reviews for all the sub-tasks for a given master JIRA before merging. otherwise, I have to rebase every time you merged.

for example, all the ledger recovery changes are sub-tasks of https://issues.apache.org/jira/browse/BOOKKEEPER-757. I haven't sent them out yet.

If you are okay, you can defer merging to me.

eolivelli:
@sijie got it. Sorry. I will only proceed with reviews. Just wanted to help

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
```diff
@@ -129,6 +131,22 @@ boolean complete(BookieSocketAddress host, final ByteBuf buffer) {
         }
 
         /**
+         * Fail the request with given result code <i>rc</i>.
+         *
+         * @param rc
+         *          result code to fail the request.
+         * @return true if we managed to fail the entry; otherwise return false if it already failed or completed.
```
jvrao: what will be case - when the control is here (i.e fail) the read is completed but not failed? If so may be we need more explanation on that scenario. May be two parallel reads, one fail and other success, ?

sijie: this change is to reading all replicas in parallel at the same time. the read completes whenever receiving a success response. A failure might come after a success response.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerRecoveryOp.java
```diff
@@ -110,7 +116,7 @@ private void doRecoveryRead() {
         if (!callbackDone.get()) {
             entryToRead++;
             try {
-                new RecoveryReadOp(lh, lh.bk.scheduler, entryToRead, entryToRead, this, null).initiate();
+                new RecoveryReadOp(lh, lh.bk.scheduler, entryToRead, entryToRead, this, null).parallelRead(parallelRead).initiate();
```
jvrao: You are assuming that if parallelRead is enabled, it is used for recoveryRead. I bet majority is fine with slow recovery but use faster/low latency parallel reads for application demanding workloads. So we need another configuration variable to allow parallel reads (if enabled) to use during recovery.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
```diff
@@ -191,6 +209,52 @@ public String toString() {
...
+        @Override
+        void read() {
+            for (int bookieIndex : writeSet) {
+                InetSocketAddress to = ensemble.get(bookieIndex);
+                try {
+                    sendReadTo(to, this);
+                } catch (InterruptedException ie) {
+                    LOG.error("Interrupted reading entry {} : ", this, ie);
+                    Thread.currentThread().interrupt();
+                    fail(BKException.Code.InterruptedException);
```
jvrao: Any read failure is a fail()? One success is good right?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerRecoveryOp.java
sijie: @jvrao : this parallel read flag is controlled by the setting "bk.getConf().getEnableParallelRecoveryRead()". currently it is for recovery only

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
sijie: yes. we don't fail the request when encountering a read failure (see logErrorAndReattemptRead). we fail here is because when you received a "InterrupteException", it means the thread is interrupted. it is most likely happening during closing the client, so we fail the request immediately.

sijie:
@merlimat @jvrao please review this. if there is no objections, I'd like to complete the ledger recovery changes and move on.

sijie:
All CI passed. Merging this now.

PR #178, opened 2017-06-01
BOOKKEEPER-1088: Ledger Recovery (part-3) - Add a ReadEntryListener to callback on individual request

sijie:
THIS CHANGE IS BASED ON #177  (you can review 868a3c8 for the only change that belongs to BOOKKEEPER-1088).

bookkeeper recovery improvement (part-3): add a ReadEntryListener to callback on individual request.

- add read entry listener which allow doing batch read, but callback on individual entries in sequence. so in recovery op, we could issue batch reads, then on each individual callback do add entry and stop when received NoSuchEntry.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,43 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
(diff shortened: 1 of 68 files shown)


jvrao:
LGTM

eolivelli:
LGTM +1

PR #179, opened 2017-06-01
BOOKKEEPER-1090: Use LOG.isDebugEnabled() to avoid unexpected allocations

merlimat:
Using `LOG.debug(...)` can lead to multiple unexpected memory allocation, even when the logger it's turned off.
For example, int and long parameter are boxed into Integer and Long objects and the var-arg parameters are using an `Object[]` to hold
them.
We should guard all usages of `LOG.debug()` with the `if (LOG.isDebugEnabled()` guard.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
@@ -734,7 +734,9 @@ public class Bookie extends BookieCriticalThread {
                 long ledgerId = recBuff.getLong();
                 long entryId = recBuff.getLong();
                 try {
-                    LOG.debug("Replay journal - ledger id : {}, entry id : {}.", ledgerId, entryId);
+                    if (LOG.isDebugEnabled()) {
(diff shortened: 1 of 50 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -1453,7 +1459,9 @@ public void addEntry(ByteBuf entry, WriteCallback cb, Object ctx, byte[] masterK
             bb.flip();
 
             FutureWriteCallback fwc = new FutureWriteCallback();
-            LOG.debug("record fenced state for ledger {} in journal.", ledgerId);
+            if (LOG.isDebugEnabled()) {
```
jvrao: Not a java expert, but isn't there any way to declare a macro to do this job instead of calling it every time??

jvrao:
can't we come up with a macro to wrap this? Chane looks fine to me though. +1

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
merlimat: For the good (and for the bad) there are no macros in Java.

PR #180, opened 2017-06-02
BOOKKEEPER-1093: Piggyback LAC on ReadResponse

sijie:
This change is based #178 - (you can review git sha 40ca8c2)

bookkeeper: LAC piggyback at read response

        - bookie server changes
          * cache maximum lac in file info
          * provide getLastAddConfirmed & setLastAddConfirmed in ledger storage
          * addEntry will set its lac thru setLastAddConfirmed
          * readEntry will read latest lac and piggyback
        - client change
          * check whether the response has lac piggybacked, if there is lac update lac in its corresponding ledger handle.


sijie:
@merlimat review?

eolivelli:
@sijie I did a more deep review and it is all great for me.

Do you think it would be useful to add a test for  readUnconfirmedEntries ? Actually the implementation is shared with readEntries so maybe we can live without such new test

b.q. Just to understand better  
How are you using this feature to improve speed of "tailing reads" ? I see that the LAC is advanced without calling readLastAddConfirmed but there is no guarantee that it will actually advance so sometimes you will need to call readLastAddConfirmed anyway, isn't it ?.
How can you distinguish the case of dead writer/no more data/closed ledger ? by catching exceptions ?

sijie:
you still need to call "readLastAddConfirmed". This change is not "long poll", it is "piggy-back lac". It is an optimization on reducing the times of calling readLastAddConfirmed. In a tailing case, most of the time you should call getLastAddConfirmed, you keep reading until your current read entry is moving beyond the value that #getLastAddConfirmed, which means you caught up with the writer then you switch to call #readLastAddConfirmed. since lac is piggy-back, so the value returned by #readLastConfirmed is advancing, hence you don't need to call #readLastAddConfirmed that often.

sijie:
@merlimat ??

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerRecoveryOp.java
```diff
@@ -40,22 +38,31 @@
...
     static final Logger LOG = LoggerFactory.getLogger(LedgerRecoveryOp.class);
-    LedgerHandle lh;
-    AtomicLong readCount, writeCount;
-    AtomicBoolean readDone;
-    AtomicBoolean callbackDone;
-    long entryToRead;
+
+    final LedgerHandle lh;
+    final AtomicLong readCount, writeCount;
+    final AtomicBoolean readDone;
```
merlimat: Minor thing, if we don't need `compareAndSet()` we can just use regular `volatile boolean` variables than don't require to be boxed

sijie: changed to volatile boolean

PR #182, opened 2017-06-02
BOOKKEEPER-1092: Ledger Recovery (Part 5) - Add Test Case for Parallel Ledger Recovery

sijie:
This change is based on #180 (ebf7020 is the change for review)

- Add test case for parallel ledger recovery
- More stats for ledger recovery


sijie:
@merlimat can you review this as well?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -1466,7 +1490,11 @@ void recover(final GenericCallback<Void> cb) {
         if (wasInRecovery) {
             // if metadata is already in recover, dont try to write again,
             // just do the recovery from the starting point
-            new LedgerRecoveryOp(LedgerHandle.this, cb).initiate();
+            new LedgerRecoveryOp(LedgerHandle.this, cb)
+                    .parallelRead(bk.getConf().getEnableParallelRecoveryRead())
```
merlimat: Minor: reading from the AbstractConfiguration object for each entry can be expensive, since it goes to the HashTable and I think it convert from String into boolean even. If possible, cache the conf value into a final member variable int the constructor.

PR #186, opened 2017-06-09
BOOKKEEPER-1096: recursive znode delete

reddycharan:
When ledger is deleted, along with leaf node
all the eligible branch nodes should be
deleted in ZooKeeper.


reddycharan:
@revans2 and @knusbaum FYI, LongHierarchicalLedgerManager needs this fix for https://issues.apache.org/jira/browse/BOOKKEEPER-1096

PR #188, opened 2017-06-13
BOOKKEEPER-1097: GC test when no WritableDirs

reddycharan:
- Functional test validating that Compaction takes place even if there
are no writableledgerdir but there are ledgerdirs according to
LedgerDirsManager.getWritableLedgerDirsForNewLog

- end-to-end testcase of Bookie recovery, incase of Bookie ledgerdir reaching
the threshold, and recovering by forcing the gc/compaction


reddycharan:
Thanks. I think this commit should be good to go.

eolivelli:
@reddycharan the patch is ok. I think @sijie will try merge current patch list in the best order

PR #190, opened 2017-06-13
BOOKKEEPER-1034: Bookie start in RO when diskfull

reddycharan:
When the disk is above threshold, Bookie goes to RO. If we have to restart the
bookie, on the way back, bookie tries to create new entrylog and other files,
which will fail because disk usage is above threshold,
hence bookie refuses to come up. So with this fix we will try to start in RO
mode if RO is enabled.

Also, if bookie has died abruptly then it may missed flushing EntryMemtable and
IndexInMemoryPageManager. So next time when it starts when disc is full
it is failing to create index file and it is shutting down, though we expect it
to start in readonlymode. So Bookie should be able to create index file
though it has reached the diskusagethreshold, while starting the Bookie in
Readonly Mode. But ofcourse there should be some config to safeguard when
disk usable space is so low.

Minor fixes in shutdown logic of Bookie and Bookieserver.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDirsManager.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDirsManager.java
@@ -56,6 +56,7 @@ public class LedgerDirsManager {
             new ConcurrentHashMap<File, Float>();
     private final long entryLogSize;
     private boolean forceGCAllowWhenNoSpace;
+    private long minUsableSizeForIndexFileCreation;
 
     public LedgerDirsManager(ServerConfiguration conf, File[] dirs) {
         this(conf, dirs, NullStatsLogger.INSTANCE);
@@ -70,6 +71,7 @@ public class LedgerDirsManager {
         this.listeners = new ArrayList<LedgerDirsListener>();
         this.forceGCAllowWhenNoSpace = conf.getIsForceGCAllowWhenNoSpace();
         this.entryLogSize = conf.getEntryLogSizeLimit();
+        this.minUsableSizeForIndexFileCreation = conf.getMinUsableSizeForIndexFileCreation();
         for (File dir : dirs) {
             diskUsages.put(dir, 0f);
             String statName = "dir_" + dir.getPath().replace('/', '_') + "_usage";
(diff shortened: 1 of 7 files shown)


reddycharan:
@sijie this is the fix we are having to deal with bookie start when ledgerdirs are full.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDirsManager.java
```diff
@@ -186,32 +186,37 @@ public boolean hasWritableLedgerDirs() {
             LOG.error(errMsg, e);
             throw e;
         }
-
+
```
sijie: remove tailing spaces?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDirsManager.java
```diff
@@ -294,24 +299,51 @@ File pickRandomWritableDir() throws NoWritableLedgerDirException {
...
+                        + " available writable dirs (" + dirs + ") : exclude dir "
                         + excludedDir);
             }
-            candidate = writableDirs.get(idx);
+            candidate = dirs.get(idx);
         }
         return candidate;
     }
-
+
```
sijie: remove tailing spaces

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieServer.java
```diff
@@ -161,11 +161,11 @@ public void resumeProcessing() {
     }
 
     public synchronized void shutdown() {
+        LOG.info("Shutting down BookieServer");
```
sijie: this doesn't seem correct to me

reddycharan: in BookieServer.start method, 'running' is set to true after nettyserver is started, so in BookieServer.shutdown method, nettyserver shutdown should be called atleast before the 'running' check. here i want to make sure that port is freed when BookieServer shutdown is called by not relying on 'running' status, which according to me is not reliable in all edge cases while testing. My main concern with 'running' variable is it is not reliable in the case of partial successful start (partial failure).

moreover I think shutdown methods should be idempotent operation and all the resources shutdown should be called irrespective of 'running' status. Ideally in shutdown methods, all they have to check is if the resource is null or not and if it is not null then it should call shutdown of that resource. Let that resource shutdown logic take care of all the scenarios, like if the resource is not started at all or if it is partially started.

reddycharan:
@sijie if you dont have anymore comments, I can fix the tailing spaces and send new CR.

sijie:
@reddycharan I don't have more comments. please fix the tailing spaces.

reddycharan:
@sijie fixed tailing spaces

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieServer.java
sijie: @reddycharan I think this change broke the TestAuth test case #221

PR #193, opened 2017-06-15
BOOKKEEPER-1086: ZkUnderreplicationManager cache watcher

athanatos:
Previously, getLedgerToReplicate left watches each time it traversed the
tree until it found a suitable replication target.  Since we don't have
a way of canceling watches, these watches tended to get abandoned,
particularly on interior nodes, which aren't changed much.  Thus,
over time, some nodes would build up a very large number of watch.

Instead, introduce a caching mechanism to remember outstanding watches
and avoid ever creating two watches on the same node.

Author: Samuel Just <sjust@salesforce.com>

### DIFF
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/util/SubTreeCacheTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/util/SubTreeCacheTest.java
@@ -1,0 +1,321 @@
+/*
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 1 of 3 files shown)


bookkeeper-server/src/test/java/org/apache/bookkeeper/util/SubTreeCacheTest.java
```diff
@@ -0,0 +1,321 @@
...
+        assertNotFired(watch);
+    }
+
+    @Test(timeout=5000)
+    public void testGuardCancelationExceptional() throws Exception {
+        TestWatch watch = null;
+        try (TestWatchGuard guard = setWatchWithGuard()) {
+            watch = guard;
+            readAssertChildren("/z/a", new String[]{});
+        } catch (Exception e) {
```
eolivelli: Should we fail in this case?

eolivelli:
@sijie I think this change can be merged

eolivelli:
@athanatos can you change the title to "BOOKKEEPER-1098: ZkUnderreplicationManager cache watcher"
our bot is binding comments to the wrong issue

athanatos:
@sijie Oops, I just noticed the comments.  I guess you just changed the title?

bookkeeper-server/src/test/java/org/apache/bookkeeper/util/SubTreeCacheTest.java
athanatos: No, the exception is meant to fire and the test is that the watch is canceled when it does (hence the assertNotFired below).

sijie:
@athanatos yeah, I just changed the title for the correct jira number.

PR #197, opened 2017-06-21
BOOKKEEPER-974: Add an official bookkeeper docker image

caiok:
Docker directory now contains two sub dirs, "4.4.0" that will contain the centos docker build and "4.4.0-alpine" that contains the alpine build.


docker/4.4.0-alpine/run.sh
```diff
@@ -0,0 +1,68 @@
...
+if [[ "${BK_LEDGERS_PATH}" != "" ]]; then
+	sed -r -i "s|^[# ]*zkLedgersRootPath.*=.*|zkLedgersRootPath=${BK_LEDGERS_PATH}|" ${BK_DIR}/conf/bk_server.conf
+fi
+
+diff ${BK_DIR}/conf/bk_server.conf.bak ${BK_DIR}/conf/bk_server.conf || true
+# -------------- #
+
+# -------------- #
+# Wait for zookeeper server
+#set +x
```
fpj: Are these lines supposed to be commented out?

docker/Makefile
```diff
@@ -0,0 +1,180 @@
...
+
+# -------------------------------- #
+
+# Create and run a bookkeeper container with data persisted on local filesystem. It needs the zookkeeper container.
+# In order to launch several bookies, the command need the bookie number
+#   make run-bk BOOKIE=4
+
+run-bk:
+	# Temporary gimmick: clear all data because of bookkeeper blocking check on host / data integrity
+	#-sudo rm -rf $(BK_LOCAL_CONTAINER_DATA_DIR)
```
fpj: Should we get rid of this commented code?

docker/Makefile
```diff
@@ -0,0 +1,180 @@
...
+	mkdir -p $(BK_LOCAL_DATA_DIR) \
+			$(BK_LOCAL_CONTAINER_DATA_DIR) \
+			$(BK_LOCAL_CONTAINER_DATA_DIR)/journal \
+			$(BK_LOCAL_CONTAINER_DATA_DIR)/ledger \
+			$(BK_LOCAL_CONTAINER_DATA_DIR)/index
+	
+	-docker rm -f $(CONTAINER_NAME)
+	docker run -it\
+		--network $(DOCKER_NETWORK) \
+	    --volume $(BK_LOCAL_CONTAINER_DATA_DIR)/journal:/data/journal \
```
fpj: If there are multiple disk devices available, will the different directories be mapped to them?

docker/Makefile
```diff
@@ -0,0 +1,180 @@
...
+		-v $(ZK_LOCAL_DATA_DIR)/datalog:/datalog \
+		-p 2181:2181 \
+		zookeeper
+
+# -------------------------------- #
+
+# Create and run a container running the bookkeeper tutorial application (a simple dice rolling application). It's possibile
+# to run several dice applications in order to simulate a real life concurrent scenario.
+#   make run-dice
+run-dice:
```
fpj: If we are to have this example in the makefile, then it might be a better idea to bring the example into bk.

jiazhai:
There seems be an empty file: "docker/4.4.0/.keep", what is the reason to have it?

docker/4.4.0-alpine/run.sh
```diff
@@ -0,0 +1,68 @@
...
+#	echo -n "."
+#	sleep 2
+#done
+#echo " Connected!"
+#set -x
+# -------------- #
+
+
+# -------------- #
+# Trying to format bookkeeper dir on zookeeper. If the dir already exists, it does nothing.
```
jiazhai: How does it achieve this?  Seems line 62 will always run?

docker/4.4.0-alpine/Dockerfile
```diff
@@ -0,0 +1,59 @@
+FROM java:openjdk-8-jre-alpine
+MAINTAINER Francesco Caliumi <francesco.caliumi@gmail.com>
```
sijie: I am not sure how does it work here.

shall it be "Apache BookKeeper" <dev@bookkeeper.apache.org> ?

docker/4.4.0-alpine/Dockerfile
```diff
@@ -0,0 +1,59 @@
+FROM java:openjdk-8-jre-alpine
```
sijie: is it possible to have a template and generate Docker files based on different JDK?

docker/4.4.0-alpine/Dockerfile
```diff
@@ -0,0 +1,59 @@
...
+MAINTAINER Francesco Caliumi <francesco.caliumi@gmail.com>
+
+# Install required packages
+RUN apk add --no-cache \
+    bash \
+    su-exec
+
+ENV ZK_SERVERS= \
+    BK_USER=bookkeeper \
+    BK_PORT= \
```
sijie: shall we provide a default BK_PORT? and what is the difference between BK_PORT and BK_BUILD_PORT?

docker/4.4.0-alpine/Dockerfile
```diff
@@ -0,0 +1,59 @@
...
+    BK_BUILD_PORT=3181 \
+    BOOKIE_OPTS="" \
+    BK_JOURNAL_DIR=/data/journal \
```
sijie: two comments for the disk layouts

- it might be good to put directories under /data/bookkeeper. so it would be /data/bookkeeper/journal, /data/bookkeeper/ledger, /data/bookkeeper/index.
- also it might be good to allow configuring multiple directories. for example, if user says two journal dirs, it should generate /data/bookkeeper/journal-0 and /data/bookkeeper/journal-1. if user only configures one journal dir, it can be /data/bookkeeper/journal.

docker/4.4.0-alpine/Dockerfile
```diff
@@ -0,0 +1,59 @@
...
+RUN set -x \
+    && apk add --no-cache --virtual .build-deps \
+        gnupg \
+        wget \
+    && mkdir -pv /opt \
+    && cd /opt \
+    && wget -q "https://archive.apache.org/dist/bookkeeper/bookkeeper-${BK_VERSION}/${DISTRO_NAME}.tar.gz" \
+    && wget -q "https://archive.apache.org/dist/bookkeeper/bookkeeper-${BK_VERSION}/${DISTRO_NAME}.tar.gz.asc" \
+    && export GNUPGHOME="$(mktemp -d)" \
+#    && gpg --keyserver ha.pool.sks-keyservers.net --recv-key "$GPG_KEY" \
```
sijie: why they are commented out?

docker/4.4.0-alpine/Dockerfile
```diff
@@ -0,0 +1,59 @@
...
+#    && gpg --batch --verify "$DISTRO_NAME.tar.gz.asc" "$DISTRO_NAME.tar.gz" \
+    && tar -xzf "$DISTRO_NAME.tar.gz" \
+    && rm -r "$GNUPGHOME" "$DISTRO_NAME.tar.gz" "$DISTRO_NAME.tar.gz.asc" \
+    && apk del .build-deps
+
+ENV BK_DIR=/opt/bookkeeper-server-${BK_VERSION}
```
sijie: how about /opt/bookkeeper/${BK_VERSION} ?

docker/4.4.0-alpine/run.sh
```diff
@@ -0,0 +1,68 @@
...
+    exit
+fi
+# -------------- #
+
+# -------------- #
+# Copy input config files in Bookkeeper configuration directory
+cp -vaf /conf/* ${BK_DIR}/conf || true
+chown -R "$BK_USER" ${BK_DIR}/conf
+
+# Bookkeeper setup
```
sijie: for the environment variables, can we have a script:

- in the dockerfile, taking a bookkeeper template configuration and export them as the environment. for example, for a setting 'x', we can export it as 'BK_${X}'.
- in the run.sh script, looping over the template configuration and for each parameters, check if its corresponding environment variable is set or not. if it is set, replace the settings.

docker/4.4.0-alpine/run.sh
sijie: I don't think we should put the format in any entry script. this is **dangerous**. it should be run as a separate admin command.

jiazhai: Put it here will make the automatic deploy easy, or user (like in K8s/DCOS) will add some additional manual  steps for the format things.
Usually only the first bookie should do the format, If the dirs, such as ../ledger/readonly, are there, the following bookies will check and not do format.

docker/4.4.0-alpine/Dockerfile
jiazhai: Second this, Better to provide default values for these ENV.

docker/4.4.0-alpine/Dockerfile
```diff
@@ -0,0 +1,59 @@
+FROM java:openjdk-8-jre-alpine
+MAINTAINER Francesco Caliumi <francesco.caliumi@gmail.com>
+
+# Install required packages
+RUN apk add --no-cache \
+    bash \
+    su-exec
+
+ENV ZK_SERVERS= \
```
merlimat: We should have a way to control all the variables that are there in `bk_server.conf`. For reference take a look at https://github.com/apache/incubator-pulsar/blob/master/docker/scripts/apply-config-from-env.py 

The variables are being taken from the ENV and applied to the config files. 

All the defaults are kept in the conf file.

docker/4.4.0-alpine/Dockerfile
```diff
@@ -0,0 +1,59 @@
...
+ENV DISTRO_NAME=bookkeeper-server-${BK_VERSION}-bin
+	
+# Download Apache Bookkeeper, verify its PGP signature, untar and clean up
+RUN set -x \
+    && apk add --no-cache --virtual .build-deps \
+        gnupg \
+        wget \
+    && mkdir -pv /opt \
+    && cd /opt \
+    && wget -q "https://archive.apache.org/dist/bookkeeper/bookkeeper-${BK_VERSION}/${DISTRO_NAME}.tar.gz" \
```
merlimat: Instead of downloading the release, we should build the docker image from the newly build JARs. That way, as part of a release we can automatically push out the docker images

docker/4.4.0-alpine/Dockerfile
jiazhai: FYI. there is also cmd line command to do the config: https://github.com/apache/bookkeeper/pull/75

docker/4.4.0-alpine/Dockerfile
merlimat: I would remove the version from the directory.. anyway, one container will have just one version of bookkeeper. Having the version means some tools and script will need to be updated each time.

docker/4.4.0-alpine/healthcheck.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+set -x -e -u
+
+# Simple healtcheck on BK port
+nc -z localhost ${BK_PORT}
```
merlimat: This check only probes the port. We should have a more thorough probe here. For example with the coommand `bookkeeper shell bookiesanity`, that creates a ledger, writes a few entries, reads them and deletes the ledger.
In that way most critical paths are being exercised.

docker/4.4.0-alpine/Dockerfile
sijie: +1 for what @merlimat said.

sijie:
@caiok any updates on this pull request?

docker/4.4.0-alpine/Dockerfile
jiazhai: +1. It is a great py.

docker/4.4.0-alpine/run.sh
caiok: I have un-commented them, it could be useful to wait for zk before start bookkeeper. Maybe can be enhanced with a timeout.

docker/Makefile
caiok: Yes

docker/4.4.0-alpine/Dockerfile
caiok: I have usually seen physical persons in MAINTAINER command, but I'm fine with your proposal

docker/4.4.0-alpine/Dockerfile
caiok: Blank is the default so it let the run.sh script to know that the env variable has not been set by "docker run" and let the config file original value. I'm fine anyway in providing a default here with documentation purpose.

docker/4.4.0-alpine/Dockerfile
caiok: @sijie 
1) IMO it's not particularly relevant because it's just a docker container internal dir and no other process are going to run in it out of bk
2) The env variables BK_xxx_DIR are not in my opinion particularly useful to set up, being an internal of docker container. For achieving your configuration one has only to mount that journal-0, journal-1 dirs in docker container. If it's not clear I could explain it more in depth

docker/4.4.0-alpine/Dockerfile
caiok: The docker build process will be executed from docker hub (it will retrieve dockerfile and than make the build every night or so). For creating an image for the the latest commit (the "on build" image, like it's usually named) we need to publish that jars anyway.

docker/4.4.0-alpine/Dockerfile
caiok: I will fix and un-comment them

docker/4.4.0-alpine/Dockerfile
caiok: I agree

docker/4.4.0-alpine/healthcheck.sh
caiok: Interesting suggestion. Doing it every 3 seconds doesn't impact performances, right?

docker/Makefile
caiok: You have to manually mount them. E.g:
--volume /real/path/1:/data/journal/journal-1
--volume /real/path/2:/data/journal/journal-2

docker/Makefile
caiok: Are you saying to bring the tutorial code in bk repo and have a docker build for that image? I think it could be a good idea (so the code will be updated with releases), but it need a general consensus. @sijie @jvrao @jiazhai @eolivelli what do you think about it?

docker/4.4.0-alpine/Dockerfile
caiok: Yes it is. I use to put placeholders and then have a makefile that generates the real dockerfile. It could be a good be a good idea, even for the bookkeeper version (I don't know how much bk changes between releases, but I suppose that the dockerfile, run.sh and healthcheck.sh will remain the same).

I'm not sure if to introduce this change now or in a second step. @jiazhai @fpj @merlimat Any preferences / ideas?

docker/4.4.0-alpine/Dockerfile
caiok: I think this script doesn't handle the commented out option lines, like the majority of options in https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/conf/bk_server.conf . It's a good idea anyway, and I'm fine to execute this script before the current substitutions done in run.sh. Maybe in a second step with a version that handles even commented out options?

docker/4.4.0-alpine/run.sh
caiok: I've messed up someway this line before open the PR, the correct one is obviously "exec bookkeeper shell metaformat -nonInteractive || true"

@sijie This command should ignore format if it find zk dir structure already formatted, right? What kind of issues can it arise? (just my curiosity)
@jiazhai I will do metaformat only if it's passed a non empty env var BK_TRY_METAFORMAT, it should satisfy both requirements.

docker/4.4.0-alpine/run.sh
caiok: Is the first step really needed or we could use the default bk_server.conf and then in run phase use found BK_xxx env variables for substitute confs in it? Furthermore, the first step will brake the possibility to mount an already set configuration file (it would be entirely rewritten).

docker/4.4.0-alpine/Dockerfile
caiok: @sijie @jiazhai  I remembered the reason behind this implementation. We provide to docker user the ability to mount configurations file at run time. These files will be rewritten with the env variables set at run. If we do provide a default for the port, in run.sh we have no way to determine if was passed by the user or it's the default, so it will be rewritten in each case, overwriting the port specified by the user in conf file. I will restore the BK_PORT=[blank] and BK_BUILD_PORT

docker/4.4.0-alpine/run.sh
jiazhai: Looks good to me, but It is better to check the related dirs in zookeeper.

docker/4.4.0/Dockerfile
```diff
@@ -0,0 +1,63 @@
...
+    && mv bookkeeper-server-* bookkeeper \
+    && yum remove -y \
+        wget \
+    && yum clean all
+
+ENV BK_DIR=/opt/bookkeeper
+ENV PATH=$PATH:${BK_DIR}/bin
+
+WORKDIR ${BK_DIR}
+VOLUME ["/conf", "${BK_JOURNAL_DIR}", "${BK_LEDGER_DIR}", "${BK_INDEX_DIR}"]
```
jiazhai: Hi @caiok ,   Seems in this code, the configuration is depends on local dir mounted to "/conf", It is good to run docker locally, but for K8S and DC/OS, it is hard to have one "local dir" contains user specified data, because all disk resources are managed by and shared in K8S or DC/OS, when disk is arriving to satisfy user request, it usually contains nothing.   
This problem also nags me for a while. I even thought to provide a way, which read the configurations from an internet place, but that seems not very user friendly.  Until @merlimat provide the way to read env variables inside docker.  
And this maybe the reason that @sijie, @merlimat and me are so care about the configurations and env variables.

docker/Makefile
jiazhai: It is a good idea. Following @fpj's suggestion, the first step would be bring @ivankelly 's [code](https://github.com/ivankelly/bookkeeper-tutorial) into bookkeeper repo?   We may need @ivankelly's idea and help on this.  :)

docker/4.4.0-alpine/run.sh
caiok: @jiazhai If metaformat has not been performed the bk container could be stuck forever with the current operations order. If otherwise put metaformat before zk check, it could die because zk is not reachable in that moment. Maybe we could maybe do: zk port check, metaformat (if env var provided) and then zk dir check?

jiazhai: Yes, I got it. "nonInteractive" will abort the format, if it as been formatted before.

docker/Makefile
ivankelly: +1 from me on making the tutorial easier to run. What is the purpose of this Makefile? A brief glance suggests it's for running multiple containers together. docker-compose may be better for this.

caiok: Good point, docker-compose was my first-choice, but it lacks a way to start containers in a prefixed order, and this made it impossible to execute metaformat before running bk containers. With last additions to bk image, docker-compose could be used to run zk, the ensemble and tutorials instance.

BTW, I've done some changes in order to use the tutorial code in its own docker image. They can be found here: https://github.com/caiok/bookkeeper-tutorial

Maybe they could be merged before adding the code to bk, but some cleanup it's needed. I could do this if there is consensus about bringing this code in bk.

Even adding a docker-compose example could be much useful then the Makefile one.

jiazhai:
Generally, looks good to me.   We may merge it in for the first stage, and do polish it, while using it. 

@caiok Still have one question of why we have 2 sub dirs --  "4.4.0", and "4.4.0-alpine", it seems not convient.  This way contains a lot of duplicate code, and harder to maintain;  and at last, when Create Automated Docker Build, only one Dockerfile is allowed provided to docker, we will have to make a decision to choose based on CentOS/Alpine sooner or later.

Here are 2 raw ideas from my view: 
1, Let's only keep the [alpine](https://www.brianchristner.io/docker-is-moving-to-alpine-linux/) version this time?  
2, It maybe better to place the version info(4.4.0) into Makefile, and remove dir 4.4.0.  It seems strange to me for dirs named by a version in master branch.

docker/4.4.0-alpine/Dockerfile
sijie: is this approach applied to current pull request? or are you planning to do it in a separate pull request?

docker/4.4.0-alpine/Dockerfile
sijie: ack

docker/4.4.0-alpine/Dockerfile
sijie: I think my point is to allow user configuring multiple directories for journal / ledger / index. I am not sure how it is achieved, by overriding the env variables?

docker/4.4.0-alpine/run.sh
sijie: I think Matteo's py script should already address this, right?

docker/4.4.0-alpine/run.sh
sijie: Ideally I am hesitated to putting any format command in a script that will run every time. That is *DANGEROUS* to me.

@merlimat how do you handle this in pulsar docker image?

docker/Makefile
sijie: Can the script handle this case?

docker/Makefile
sijie: @caiok : I remembered that pulsar is using docker-compose to start zk, bk, and pulsar brokers. @merlimat can help confirm that.

@jiazhai @caiok : can you guys create an issue to bring Ivan's tutorial to bookkeeper-turorials/dice?

caiok:
@jiazhai 
It's ok to have such a structure. We need it in order to provide docker images of older versions. Having a CentOS image is ok too, when licensing concerns not allow to run alpine. Take a look at other "official" images, you will see that no images provide only one tag. The [postgres](https://hub.docker.com/_/postgres/) example is quite complete.

It's normal that centos and alpine version shares a lot of code, unfortunately, but I don't see any solution to that. To reduce code replication between releases of the same os, I think in the end @sijie solution to generate versions from a template is the best.

@sijie 
I missed your question: "for current version, it might be worth having a latest directory. when we cut a release, we copy the latest directory to the targeted release directory. for example, in 4.5.0, we can copy latest to a directory 4.5.0. Is that doable?"
It is possible and simple to create an "onbuild" docker build (not to be confused with "latest" that in docker means the latest stable version) if maybe in jenkins job we run the "package" target and then we publish the produced packet somewhere. Otherwise it's still possible to create a docker build that checkout the last commit (at risk to have something broken, with jenkins we can publish the package only if all tests pass), package it, unpack it in the usual dir and then clean everything it's not needed runtime, but it's require some effort, doubled by alpine + centos dichotomy, so it's better to create a dedicated issue.

docker/4.4.0-alpine/run.sh
caiok: Matteo's py only handles not commented out properties present in original files. It needs to be enhanced with handling of commented ones IMHO, in order to not create confusion (if the doc says "you can set BK_xxxx for setting corresponding BK property" I expect this behavior, if it don't work it's not kind to demand the user dig in the code to figure out the reason). If we provide default conf files with all properties not commented out, then the problem arise if user mount it's own conf files.

docker/4.4.0-alpine/Dockerfile
caiok: No, it hasn't been applied, yet. I just created #260

docker/4.4.0-alpine/Dockerfile
caiok: No, env vars refer to internal container directories and I don't think it's very useful changing them.
Current build already creates three different dirs, so the end user just have to mount the desired host dirs to these dirs. Take a look at makefile example (target run-bk) for seeing this in action. It's just like the common mounting mechanism in unix, the target system will not notice the difference.

docker/Makefile
caiok: #261

jiazhai:
Hi @caiok, 
Postgres example is not a "Automated Docker Build",  While if bookkeeper image goes into [apache](https://hub.docker.com/u/apache/), we could reference the apache examples.

caiok:
@jiazhai Seems that Apache examples are a bit poor on this side. [Here an example](https://hub.docker.com/r/azul/zulu-openjdk/) (the first I found) with an automated build with tags. You can refer to [the official documentation](https://docs.docker.com/docker-hub/builds/#create-an-automated-build) in order to understand how it works. How you can see, you have  a lot of flexibility about tags.

jiazhai:
@caiok, Thanks for the explain.
It is better to handle the code duplications in dirs 4.4.0 and 4.4.0-alpine. If it is hard to resolve, and CentOS version is preferred, It is Ok to keep CentOS version.
It is better to put version information into related version branch, like in branch-4.4 including docker files for docker-4.4.0, not in master. Master branch always keep files for the latest docker file. We could do that in your [example](https://github.com/zulu-openjdk/zulu-openjdk), but that seems a little mess.

caiok:
@jiazhai 
What are your concerns about this code duplication? I think providing a centos image for who is concerned about license and compatibility and an alpine one for who is concerned security and performances could be nice. There is no simple way to use only one copy of the files for both (you need to create a template to "compile" in respective files).

On image versions I fear the branch solution could rapidly became very uncomfortable (and I think this is the reason I never seen it used, in main images). If you released 3 versions and find a dockerfile bug, you have to commit this change on every branch? (you than have one or more commits "detached" from the release tag, that is maybe a little dirty).
I prefer @sijie solution with templates in order to reduce code duplication between different versions but I'm not closed to alternatives (even the branch one) if there is consensus.

caiok:
@sijie @fpj @merlimat Any other doubts or requests for this PR? We need it merged before proceeding with the related issues (and any thoughts on these will be much appreciated too)

sijie:
@caiok : regarding the versioning here, I am not sure whether we need to keep the version here. Ideally the docker image should be built along with an apache release. For example when we cut an apache release 4.5.0, along with the release procedure, we will build the 4.5.0 image and push it to apache docker hub. The docker files for 4.5.0 will live in the branch 4.5.0. The docker files in master will evolve with the changes in next release (for example 4.6.0). That means the docker files for image 4.5.0 will live only in branch 4.5.0 and subsequent changes to 4.5.0 will happen only in branch 4.5.0 and new changes will live in master. How does that sound?

I see a few issues created for followup actions. I am fine with merging this pull request and iterate after that if we can achieve basic consensus here.

@fpj @merlimat : since you were reviewing this pull request, please try to take a look at this pull request when you have time. I'd like to move this forward with 4.5.0 release.

caiok:
@sijie Thanks for the reply. We don't push images on dockerhub, dockerhub checkout our dockerfiles and builds them. Check [this page](https://docs.docker.com/docker-hub/builds/#understand-the-build-process) in order to have an idea of the process.
Current dockerfile will be useless for master version until the package will be released somewhere (but we can create an onbuild dockerfile that checkout and build the code in the container).
Anyway, I'm fine to let version x.y.z of docker build to live on branch x.y.z, if this is the prevalent preference (there are some inconveniences that I explained to Jia, though)

sijie:
@caiok 

Let me try to summarize here. There are two approaches: 

1. the approach suggested in this pull request. a summary for this approach.

- use the docker hub autobuilds.
- download released package in docker build
- maintain different docker files for different versions

pro: you can use docker hub autobuilds the images
cons: 
- you have to manage the docker hub account in order to build the images, because you have to create the builds in docker hub side. in apache, we might not have access to apache docker account.

2. the approach other people suggested in this pull request.

- use the apache/travis ci build, build the image along as part of the CI
- maintain only one copy of docker files for each release/branches.
- push the ci built image to apache docker hub account (e.g. docker push)

pro: the image is built as part of CI and aligned with apache release. and we can simply use an API token from apache docker account to push built image and tag them.
cons: managing versions might be not as easy as the first approach, although I am not very sure.


Approach 2) is aligned with apache release and it is easy to configure if we want to push to apache docker account. 1) will be a bit tricky, not sure how we can create an autobuild in apache docker account.

Hope this summary is clear for the people involved in this pull request.

docker/4.4.0-alpine/Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+
+WORKDIR ${BK_DIR}
+VOLUME ["/conf", "${BK_JOURNAL_DIR}", "${BK_LEDGER_DIR}", "${BK_INDEX_DIR}"]
+
+EXPOSE ${BK_BUILD_PORT}/tcp
+
+COPY run.sh healthcheck.sh /opt/
+
+ENTRYPOINT [ "/bin/bash", "/opt/run.sh" ]
+CMD ["bookkeeper", "bookie"]
```
kishorekasi: @caiok - here is the snippet you asked for (from my local Dockerfile)
CMD /bookkeeper/bookkeeper-server/bin/bookkeeper shell bookieformat -nonInteractive -force -deleteCookie; /bookkeeper/bookkeeper-server/bin/    bookkeeper bookie; /bookkeeper/bookkeeper-server/bin/bookkeeper autorecovery

sijie:
notes about the docker image from today's meeting

- follow what other apache projects are doing [zeppelin](https://issues.apache.org/jira/browse/INFRA-12781)
- using docker autobuild
- dockerfile is to download official bk release image
- remove versions. docker autobuild will look for docker file location and release tags.
- os versions. let's stick to centos-7 and jdk-8.
- configuration: 1) allow providing a configuration from a mounted volume 2) if no configuration file is provided, it will use the template packaged in the binary and take the ENV variables that start with "bk_"
- entry point: we need an entry point to start and stop bookie gracefully
- documentation: docker-compose file to start a cluster for testing
- optional: k8s and dcos instructions (can be separate pull requests)

sijie:
@caiok @jiazhai as we discussed in the community meeting, do you guys want to work together to move this forward?

jiazhai:
@sijie @caiok How about merge this first, and open another issue to improve it following the comments above?

caiok:
Thanks @sijie for the recap of community meeting decisions.

@jiazhai I'm ok with merging this and move on. I will open new issues for what is still missing.

sijie:
@jiazhai @caiok what is going to be reused here? as we discussed, we need to keep only one version. I am wondering why not just close this pull request and start a new one that incorporates the comments. that would be much clearer, right?

jiazhai:
@sijie, I am OK with your idea to start a new one, since this will be changed a lot. 
It depends on @caiok's decision.

caiok:
@sijie @jiazhai I agree, let's close this PR and start a new one. I will work on it as soon as possible.

zhaijack:
@caiok I am also starting do it. How could we collaborate together on it?

caiok:
@jiazhai As a first thing we should decide the issues we want included in this PR. I suggest #327 and the base step for #289 that, if we decide to implement this issue, should change the dir structure from a "all in docker in root dir" to a "general files in root dir and build files in two subdirs: release and onbuild". I prepend for the letter.

This should be quite easy and I'd like to create PR tonight. Then we can divide remaining issue between us and work in parallel. Sounds good to you?

I created #docker-image slack channel so we can decide quickly and report here the decisions.

PR #199, opened 2017-06-21
Issue 207: change the HashSet to BitSet to handle duplicated bookies on reading entries

sijie:

Problem:

If there happens to have duplicated bookies in same ledger (it happened before when ensemble change still has problem), the read doesn't work as expected. It hang forever.

Solution:

Change the HashSet to BitSet to use bookieIndex rather than InetSocketAddress to track the heardfrom bookies set.

---
quickly and easily:

---


jiazhai:
Would you please open an Issue for it, and change the name of this PR as "Issue XXX: Description ...." ?

sijie:
@jiazhai done

PR #200, opened 2017-06-21
Issue 209: Introduce Speculative Read/Read LAC policy

sijie:

    - Separate out the logic that determines when speculative reads are scheduled from the individual requests into a separate SpeculativeRequestExecutionPolicy
    - Implement the default speculative read execution policy based on the current mechanism of using first and max timeout
    - Initialize the policy once in the BookKeeper object and use it for individual requests by keeping information about the requests in the call chain.

---
quickly and easily:

---

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
@@ -29,11 +29,14 @@ import java.util.List;
 import java.util.NoSuchElementException;
 import java.util.Queue;
 import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.Callable;
 import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
+
+import com.google.common.util.concurrent.ListenableFuture;
 
 import org.apache.bookkeeper.client.AsyncCallback.ReadCallback;
 import org.apache.bookkeeper.client.BKException.BKDigestMatchException;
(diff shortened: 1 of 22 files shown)


sijie:
This pull request is based on #199 ( e79d620 is the change for this pull request ).

The change is contributed by Robin Dhamankar

jiazhai:
Would you please open an Issue for it, and change the name of this PR as "Issue XXX: Description ...." ?

sijie:
@jiazhai I think we agreed on the workflow - if there is already a change, people can submit a pull request directly.

eolivelli:
@sijie in https://cwiki.apache.org/confluence/display/BOOKKEEPER/BP-9+-+Github+issues+for+Issue+Tracking
@jiazhai  wrote "Always create an issue for a pull request" as @fpj suggested.

I think we all agreed to what was written in the wiki

jiazhai:
@eolivelli, thanks for the info.  @sijie, excuse me, there maybe some mis-sync regarding this.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
```diff
@@ -565,7 +559,7 @@ protected void submitCallback(int code) {
                 }
             }
             LOG.error("Read of ledger entry failed: L{} E{}-E{}, Heard from {}. First unread entry is {}",
-                    new Object[] { lh.getId(), startEntryId, endEntryId, heardFromHosts, firstUnread });
+                    new Object[] { lh.getId(), startEntryId, endEntryId, heardFromHostsBitSet, firstUnread });
```
jvrao: I would still like to get the bookie inet address as we use it extensively for debugging. With this we get the index only right?

jvrao:
@sijie what is the major drive of this change? Current method of trying speculative reads has any major flaw? I understand this is better organized as speculative reads has its own policy, but trying to understand what use-case drove you to this change at Twitter?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
sijie: I will add the address information in the error messages

sijie:
- in the upcoming changes of long poll requests, we need same speculative policy for long poll requests. abstract the logic as a policy will make things easier and we don't have to duplicate the code for both reads and long poll reads
- abstracting the logic as a policy will help us improving the policy without changing the read code.
- we also extend the default policy to have exponential backup in place.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
sijie: @jvrao I added the heardFromHosts set back in e93a5cb . so the error messages will include the host information. let me know if it is okay to you.

PR #205, opened 2017-06-22
Issue 208: Improve ledger fence logic

sijie:

Problem:
    When bookie receive a fence request and couldn't find any writable dirs for the new index file, it will throw exception. This behavior can be improved, because as long as ledger fence request be persisted in Journal, we can say the fence request succeed. It should not depends on the success of flushing new index file.

Solution:

- Add option to fall back to pick from all directories regardless of writable or not when we getFileInfo for a new ledger
- Return success only when ledger fence request has been persisted in Journal

---
quickly and easily:

---


sijie:
This change is contributed by Yiming from Twitter.

eolivelli:
@sijie maybe you can change the description of the PR dropping the "Be sure to do all of the following to help us incorporate your contribution...." lines

jiazhai:
Seems there is conflict files.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/IndexPersistenceMgr.java
```diff
@@ -177,7 +177,35 @@ private FileInfo putFileInfo(Long ledger, byte masterKey[], File lf, boolean cre
...
+     * Get a new index file for a ledger in a lazy way.
+     *
+     + <p>If fallback is false, this function will throw exception when there are no writable dirs.
+     + If fallback is true and there's no writable dirs, it will ignore the error and pick any dir.
+     + Set fallback to true is useful when we want to delay disk check and just get the File pointer, e.g. fence ledger
+     *
+     * @param ledger
+     *          Ledger id.
+     * @param excludedDir
+     *          The ledger directory to exclude.
```
jvrao: Add new @param fallback here.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/IndexPersistenceMgr.java
```diff
@@ -177,7 +177,35 @@ private FileInfo putFileInfo(Long ledger, byte masterKey[], File lf, boolean cre
      */
     private File getNewLedgerIndexFile(Long ledger, File excludedDir)
                     throws NoWritableLedgerDirException {
-        File dir = ledgerDirsManager.pickRandomWritableDir(excludedDir);
+        return getNewLedgerIndexFile(ledger, excludedDir, false);
```
jvrao: Can you explain situations where fallback is not allowed and where it is OK.
Index files are small, but still it can push dirs to 100% and eventually write can fail.
Looks like this change has *NO* Checks to see the fullness of the randomly picked dir. Any consequences ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDescriptorImpl.java
```diff
@@ -81,6 +85,52 @@ ByteBuf getExplicitLac() {
...
     }
 
+    synchronized SettableFuture<Boolean> fenceAndLogInJournal(Journal journal) throws IOException {
+        boolean success = this.setFenced();
+        if(success) {
+            // fenced for first time, we should add the key to journal ensure we can rebuild.
+            return logFenceEntryInJournal(journal);
+        } else {
+            // If we reach here, it means this ledger has been fenced before.
+            // However, fencing might still be in progress.
```
jvrao: Comment is vague. Please reword it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDescriptorImpl.java
```diff
@@ -81,6 +85,52 @@ ByteBuf getExplicitLac() {
...
+
+    /**
+     * Log the fence ledger entry in Journal so that we can rebuild the state.
+     * @param journal log the fence entry in the Journal
+     * @return A future which will be satisfied when add entry to journal complete
+     */
+    private SettableFuture<Boolean> logFenceEntryInJournal(Journal journal) {
+        SettableFuture<Boolean> result;
+        synchronized (this) {
+            result = logFenceResult = SettableFuture.create();
```
jvrao: When logFenceEntryInJournal is called in line 104, the else case says we have object logFenceResult already. Why are we creating another logFenceResult here???

sijie: Because the fence is failed , line 102 is the else branch of line 96..

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDescriptorImpl.java
sijie: Will do.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/IndexPersistenceMgr.java
sijie: First of all, creating a index file doesn't mean we really write any data to this file. the real file creation happens only when the ledger is flushed. The creation of the index file to hold the states eg fencing state for flushing later on.

Secondly, the problem this issue tries to address here is even the we don't have writable disks, the fence requests should go through since we only need the fence state is set in the file info cache and persisted in the journal. The change of fallback is for this purpose.

Thirdly, it doesn't have to check the fullness, because the creation of the index file happens in the memory at this point. It is deferred to create at flushing time. When a ledger is flushed, it will check the fullness and do the relocate if it is necessary.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDescriptorImpl.java
sijie: done. I have improved the comments.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/IndexPersistenceMgr.java
sijie: done

sijie:
@jvrao I have addressed your comments. please take a look and let me know if it works for you. I'd like to move forward with other pull requests.

sijie:
ping @jvrao

sijie:
@jvrao any more comments?

jvrao:
@sijie thanks for the explanation; ok for me.

eolivelli:
@sijie are going to merge this now? I can merge this and other approved patches today

eolivelli:
@sijie sorry I did not see that it needs a rebase

sijie:
@eolivelli I am currently traveling. I will spend some time on rebasing my changes ASAP when I have a large chunk of time.

Issue #221, opened 2017-06-30
TestAuth#testCloseMethodCalledOnAuthProvider is failing 


sijie:
**BUG REPORT**

1. Please describe the issue you observed:

- What did you do?

Run TestAuth.

- What did you expect to see?

The test should pass.

- What did you see instead?

The test failed. Because the shutdown is called twice after this change 9ddd9e6f9e48b03a57a2c78ec2630303abd49782 . So the assertion fails.

sijie:
@reddycharan I don't think the right approach is to change to the tests. we should just make the shutdown call idempotent.

eolivelli:
@reddycharan @sijie the close method of the factory is expected to be called once.
I am OK with the fact that close methods should be idempotent too but if it is possible it is better to make the call to "close" once. It is cleaner and for the long term it is the best approch

sijie:
sorry I typed the wrong issue id at #437. revert the labels and milestone.

PR #222, opened 2017-06-30
Issue 221: Make netty server shutdown idempotent

sijie:

(see the detailed discussion in #221 )

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieNettyServer.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieNettyServer.java
@@ -85,6 +85,7 @@ class BookieNettyServer {
     final EventLoopGroup jvmEventLoopGroup;
     final RequestProcessor requestProcessor;
     final AtomicBoolean isRunning = new AtomicBoolean(false);
+    final AtomicBoolean isClosed = new AtomicBoolean(false);
     final Object suspensionLock = new Object();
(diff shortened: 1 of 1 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieNettyServer.java
```diff
@@ -279,6 +280,12 @@ void start() {
     void shutdown() {
         LOG.info("Shutting down BookieNettyServer");
         isRunning.set(false);
+
+        if (!isClosed.compareAndSet(false, true)) {
```
eolivelli: IMHO maybe you can move this at the start of the method, so we will not print twice "Shutting down" and doing to isRunning.set

reddycharan: +1 with the comment

+1 with the change

sijie: I did it in this way intentionly. Because current shutdown will be called twice, I would like to have logging to tell, so if we want to debug, we know what happened.

sijie: Same thing applied to isRunning. If the shutdown will be called twice, I would like it to be set eventually.

eolivelli:
Ok for me

PR #220, opened 2017-06-30
BOOKKEEPER-1095: Long Poll - Server and Client Side Changes

sijie:
Descriptions of the changes in this PR:

- changes on FileInfo to support notifications on LAC changes
- a new ReadEntryLongPollV3 processor to process readEntry requests with long poll flags
- a new public API for long poll: readLastConfirmedAndEntry. if it is reading beyond the LAC, it will become a long poll request and wait for advancing LAC on bookie side; if it isn't reading beyond the LAC it will be normal reads.
- also have a speculative mechanism for long poll reads.

PR #224, opened 2017-07-01
BOOKKEEPER-772: Reorder Read Sequence

sijie:

    - for rackware placement policy, the bookie in the same rack will be preferred.
    - for region-aware placement policy, the bookie in the same region will be preferred.
    - for any readonly or unavailable (high score in bookie failure history) bookies, they will be at the last position in the sequence.

This change is based on #220 . Please review gitsha 1f7dccd for the reorder change.


sijie:
@jiazhai ah, this change was based on #220 so we need #220 to be merged first. but anyway, I am going to address the comments as part of #220.

PR #227, opened 2017-07-01
Add missed tests for SortedLedgerStorage

sijie:
Added two missed tests for SortedLedgerStorage. It was missed when we (twitter) merged the SortedLedgerStorage back.


sijie:
It is a trivial merge. so no separate issue is created here.

PR #233, opened 2017-07-04
Issue 232: Add code-coverage report

eolivelli:
Introduce Cobertura And Coveralls.io plugins in order to run code coverage reports

### DIFF
--- a/bookkeeper-server/pom.xml
+++ b/bookkeeper-server/pom.xml
@@ -437,73 +437,23 @@
         <build>
             <plugins>
                 <plugin>
-                    <groupId>org.jacoco</groupId>
-                    <artifactId>jacoco-maven-plugin</artifactId>
-                    <version>0.7.9</version>
-                    <executions>
-                        <execution>
-                            <id>default-instrument</id>
-                            <goals>
-                                <goal>instrument</goal>
-                            </goals>
-                        </execution>
-                        <execution>
-                            <id>default-restore-instrumented-classes</id>
-                            <goals>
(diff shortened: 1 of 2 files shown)


bookkeeper-server/pom.xml
```diff
@@ -432,5 +432,78 @@
...
+                            <goals>
+                                <goal>check</goal>
+                            </goals>
+                            <configuration>
+                                <rules>
+                                    <rule implementation="org.jacoco.maven.RuleConfiguration">
+                                        <element>BUNDLE</element>
+                                        <!--
+                                        <limits>
+                                            <limit implementation="org.jacoco.report.check.Limit">
```
eolivelli: We should define a limit and a criterion to fail the build otherwise we can run the plugin and simply publish the report.
I would prefer to add a limit, maybe on the max number of uncovered lines

eolivelli:
I have added the plugin only  to bookkeeper server because it is the only significant project for this kind of check

sijie:
I think DL also uses jacoco to report code coverage to coveralls.io. It might be good to make these two repo using same mechanism, as part of the progress of merging DL with BK.

eolivelli:
@sijie I will update the patch with the same jacoco config of DL. Infra team did not respond to my email about coveralls.io . Jacoco is a prerequisite and it is working well on bookkeeper-server

eolivelli:
Updated the PR with cobertura + coveralls.io as DL does

sijie:
LGTM +1

PR #238, opened 2017-07-10
Issue-237 Upgrade to ZooKeeper 3.5.3-BETA

eolivelli:

Update ZooKeeper dependency from 3.5.1-alpha to 3.5.3-beta.
This will enable ZooKeeper over SSL

---
quickly and easily:

- [x ] Make sure the PR title is formatted like:
- [ x] Make sure tests pass via `mvn clean apache-rat:check install findbugs:check`.
- [ x] Replace `<Issue # or BOOKKEEPER-#>` in the title with the actual Issue/JIRA number.

---

### DIFF
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/test/ZooKeeperUtil.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/test/ZooKeeperUtil.java
@@ -47,6 +47,12 @@ import org.slf4j.LoggerFactory;
 import static org.junit.Assert.*;
 
 public class ZooKeeperUtil {
+
+    static {
+        // org.apache.zookeeper.test.ClientBase uses FourLetterWordMain, from 3.5.3 four letter words
+        // are disabled by default due to security reasons
+        System.setProperty("zookeeper.4lw.commands.whitelist", "*");
+    }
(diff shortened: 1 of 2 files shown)


eolivelli:
@sijie @jiazhai is this ok for you? We upgraded to 3.5  branch for DL.

jiazhai:
+1, LGTM.

bookkeeper-server/src/test/java/org/apache/bookkeeper/test/ZooKeeperUtil.java
```diff
@@ -47,6 +47,12 @@
 import static org.junit.Assert.*;
 
 public class ZooKeeperUtil {
+
+    static {
```
sijie: if you don't enable this, what will happen? Can you clarify this?

eolivelli: @sijie for 3.5.3 most of four letter words commands are disabled by default in order to prevent DoS attacks. We are using ClientBase to wait for ZK server to start and this uses the fourbletter words API.
Without this change  tests will be very slow because the wait will always timeout

sijie: gotcha. thank you for the explanation.

PR #244, opened 2017-07-14
Issue#243 - asyncAddEntry fails with NPE with LedgerHandlerAdv

eolivelli:
Fix asyncAddEntry on LedgerHandleAdv and clean up the asyncAddEntry API, drops BKException which is never thrown in asynch functions

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
@@ -720,8 +720,7 @@ public class LedgerHandle implements AutoCloseable {
      * @param ctx
      *            some control object
      */
-    public void asyncAddEntry(final long entryId, final byte[] data, final AddCallback cb, final Object ctx)
-            throws BKException {
+    public void asyncAddEntry(final long entryId, final byte[] data, final AddCallback cb, final Object ctx) {
         LOG.error("To use this feature Ledger must be created with createLedgerAdv() interface.");
         cb.addComplete(BKException.Code.IllegalOpException, LedgerHandle.this, entryId, ctx);
     }
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandleAdv.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandleAdv.java
@@ -121,7 +121,7 @@ public class LedgerHandleAdv extends LedgerHandle {
      *            some control object
      */
     @Override
-    public void asyncAddEntry(long entryId, byte[] data, AddCallback cb, Object ctx) throws BKException {
+    public void asyncAddEntry(long entryId, byte[] data, AddCallback cb, Object ctx) {
         asyncAddEntry(entryId, data, 0, data.length, cb, ctx);
     }
 
@@ -165,6 +165,11 @@ public class LedgerHandleAdv extends LedgerHandle {
      */
     @Override
     protected void doAsyncAddEntry(final PendingAddOp op, final ByteBuf data, final AddCallback cb, final Object ctx) {
+        if (op.entryId < 0 ) {
+            super.doAsyncAddEntry(op, data, cb, ctx);
+            return;
+        }
+
         if (throttler != null) {
             throttler.acquire();
         }
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandleAdv.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandleAdv.java
@@ -165,7 +165,9 @@ public class LedgerHandleAdv extends LedgerHandle {
      */
     @Override
     protected void doAsyncAddEntry(final PendingAddOp op, final ByteBuf data, final AddCallback cb, final Object ctx) {
-        if (op.entryId < 0 ) {
+        if (op.entryId == LedgerHandle.INVALID_ENTRY_ID) {
+            // if entryId has not been set we need to fallback to normal LedgerHandle doAsyncAddEntry
+            // and assign a new entryId
             super.doAsyncAddEntry(op, data, cb, ctx);
             return;
         }
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandleAdv.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandleAdv.java
@@ -165,13 +165,6 @@ public class LedgerHandleAdv extends LedgerHandle {
      */
     @Override
     protected void doAsyncAddEntry(final PendingAddOp op, final ByteBuf data, final AddCallback cb, final Object ctx) {
-        if (op.entryId == LedgerHandle.INVALID_ENTRY_ID) {
-            // if entryId has not been set we need to fallback to normal LedgerHandle doAsyncAddEntry
-            // and assign a new entryId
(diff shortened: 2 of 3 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandleAdv.java
```diff
@@ -165,6 +165,11 @@ public void asyncAddEntry(final long entryId, final byte[] data, final int offse
      */
     @Override
     protected void doAsyncAddEntry(final PendingAddOp op, final ByteBuf data, final AddCallback cb, final Object ctx) {
+        if (op.entryId < 0 ) {
```
sijie: when will this happen? I think this change volatiles the semantic of LedgerHandleAdv#asyncAddEntry. if the entry id is negative, we should reject this add with illegal arguments.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -720,8 +720,7 @@ public void asyncAddEntry(final byte[] data, final AddCallback cb,
      * @param ctx
      *            some control object
      */
-    public void asyncAddEntry(final long entryId, final byte[] data, final AddCallback cb, final Object ctx)
-            throws BKException {
+    public void asyncAddEntry(final long entryId, final byte[] data, final AddCallback cb, final Object ctx) {
```
sijie: since we are touching this function now, I'd like to also raise one discussing about the semantic here. Currently it is totally the application's responsibility on managing the semantic on using this API. it would be good to add some constraints on this API:

- if entryId is negative  (the fix @eolivelli tried to fix)
- if entryId is smaller than LAC, that means this entry is a duplicated entry

in both cases, we should fail the addEntry with exceptions.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandleAdv.java
eolivelli: In order to get the npe just see the new  testcase.
My bad I should have added  an example in the issue.
 The idea is that if in LedgerHandleAdv the user does not provide an entry id, as in the normal addEntry method, the call should be handled as standard LedgerHandle does.
I see my fix was too simple.
I will come back on it monday. I apologize for having sent a PR too early.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
eolivelli: @sijie the check with entryId <= 0 was to detect that no entryId was set, I have changed the code with the best way I have found, that is to check if entryId == LedgerHandler.INVALID_ENTRY_ID (-1)

the BKException will never be thrown as the callback will to be called with an error code

in LedgerHandlerAdv there is such a check as @sijie told in asyncAddEntry
this is the code:
```
if ((entryId <= this.lastAddConfirmed) || pendingAddOps.contains(op)) {
            LOG.error("Trying to re-add duplicate entryid:{}", entryId);
            cb.addComplete(BKException.Code.DuplicateEntryIdException,
                    LedgerHandleAdv.this, entryId, ctx);
            return;
        }
```

I noticed also that in LedgerHandlerAdv we missed the ByteBuf variant of the API asyncAddEntry with entryId, we should create a new issue for this

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieWriteLedgerTest.java
```diff
@@ -179,6 +180,27 @@ public void testLedgerCreateAdv() throws Exception {
     }
 
     /**
+     * Verify that LedgerHandleAdv can handle asynchAddEntry without the entryId
+     *
+     * @throws Exception
+     */
+    @Test(timeout = 60000)
+    public void testAsynchAddEntryLedgerCreateAdv() throws Exception {
```
sijie: Asynch => Async

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
sijie: in LedgerHandleAdv, we should not allow adding entries without providing any entry id. Otherwise, the semantic is too complicated to maintain. As it is hard to tell, -1 is a user provided entry id or it is provided by calling addEntry without an entry id.

/cc @jvrao

eolivelli: @sijie as LedgerHandleAdv is a subclass I thought it was automatic that all the simple API should work as in the normal LedgerHandle. If this is this the case then I am ok with making all functions without explicit entryid fail. I have not tested with 4.4.
IMHO if possible it would be better to support the full API, I think that as last option we will need an extra boolean field in PendingAddOp

sijie: @eolivelli I don't remember the exact conversation I had with JV before. but it is better not to mix these two semantics together. 

I am not sure why we need an extra boolean field. You can simply override the methods and fail the requests, can't you?

eolivelli: The boolean flag would be usefull to tell that the entryid in PendingAddOp is not  initialized, an alternative could be to use a java.lang.Long.
But if we decide to fail all the normal addEntry calls this will no be useful.
@jvrao what do you think? Are you using -1 as vakid entryid? Are you using negative values? Cc @reddycharan

sijie: I don't understand why do we need this. The LedgerHandleAdv can just override the method and fail the request before initiating the add ops.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandleAdv.java
jvrao: Why are you using Adv handle and not intended to pass down entryId? I don't understand the usecase.

eolivelli:
@sijie  I have updated the patch with a check on LedgerHandleAdv in order to prevent the client to use addEntry (and overloaded/similar functions) without passing an entryId.

I am not sure that -1 or negative values are valid entryIds, because readEntries does some check and fails the call with negative values

```
        // Little sanity check
        if (firstEntry < 0 || firstEntry > lastEntry) {
            LOG.error("IncorrectParameterException on ledgerId:{} firstEntry:{} lastEntry:{}",
                    new Object[] { ledgerId, firstEntry, lastEntry });
            cb.readComplete(BKException.Code.IncorrectParameterException, this, null, ctx);
            return;
        }
```

@jvrao I have no use case I was only playing with the API and I found a NullPointerException, which IMHO is always a bad bug.
The fact that LedgerHandleAdv is a subclass of LedgerHandle makes it simple for a new user to use the bad API.
For me is OK to prevent calls to addEntry without entryId on LedgerHandleAdv, in fact if you are assigning the entryId externally you cannot use the internal id generator.
I have updated the patch and I have added testcases.

I have tried to use negative entries (and -1 too) but actually it is not possible, @jvrao can you confirm ?

eolivelli:
@sijie I was writing a test case for negative entry id and I discovered the fact

eolivelli:
@sijie  @jvrao can someone of you click on "Approve button", the merge script needs at least an "approval" to work, a simple comment with  '+1' or "LGTM"  is not enough. thanks

PR #246, opened 2017-07-16
Issue 245 Enable Coverall.io and Covertura on Travis configuration file

eolivelli:
Enable Coveralls.io and Cobertura during build on Travis-CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,3 +41,6 @@ script:
 cache:
   directories:
     - $HOME/.m2
+
+after_success:
+  - mvn clean cobertura:cobertura coveralls:report
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,4 +43,4 @@ cache:
     - $HOME/.m2
 
 after_success:
-  - mvn clean cobertura:cobertura coveralls:report
+  - mvn clean install cobertura:cobertura coveralls:report


eolivelli:
I added cobertura on after-success callback but the build failed, tests are not compiling. I will investigate and change the config.
Unfortunately Travis reports the build as successful.
Maybe we should not use the after success callback, but this is the suggested config from Travis

eolivelli:
@sijie take too long to complete and Travis is not happy.
Maybe we should eork on making tests faster on resource limited envs like Travis.
Maybe we can make parametrized test do less work. Cc @merlimat

sijie:
a few comments:

1) currently we only used travis ci for verifying pull requests (e.g. findbugs and checkstyles)
2) if you want to run coverage for each pull requests, you can skip tests for now. but I am not sure if we need to run coverage for every pull requests. I would suggest starting with master.
3) coveralls requires a token to report correctly. you might need to setup a secret env variable in travis ci. or it might be making sense just do this in jenkins. (DL does coverage on jenkins only)

eolivelli:
@sijie  in order to see code coverage you have to run tests anyway.
I am going to look for a jenkins based solution. I am closing this issue

PR #257, opened 2017-07-18
Issue 256: BookieInitializationTest.testWithDiskFullAndAbilityToCreateNewIndexFile failed

sijie:

in c49621b, we allow creating FileInfo even when disk is full to allow fence requests succeed at this case.

it is conflicted with Charan's change. Since there is already a setting to configure the min usable size to allow index file creation. if we configure that to be zero, it is same as allowing index file creation even disks are full. so removing the fallback logic introduced in c49621b.


reddycharan:
+1 LGTM

PR #264, opened 2017-07-19
Issue 263: Enable request rate limiter in zookeeper client

sijie:

Eanble request rate limiter in zookeeper client. The rate limiter logic is in place, but is never enabled. This change is to expose the settings in configurations and enable them in zookeeper client at both client and server.


eolivelli:
@sijie can we use the same configuration parameter in BookKeeper constructor ? I can file another issue

PR #266, opened 2017-07-20
Issue 265: Add persistable bookie status

sijie:

- Add support for persisting bookie status
- Add configuration to enable/disable this feature
- Add test cases
- Improve bookie status stat

It also includes changes for BOOKKEEPER-754 (flush ledger storage after replaying journal)


sijie:
This feature is contributed by @yzang

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
```diff
@@ -0,0 +1,246 @@
...
+        if (!bookieMode.equals(BookieMode.READ_WRITE)) {
+            bookieMode = BookieMode.READ_WRITE;
+            this.lastUpdateTime = System.currentTimeMillis();
+            return true;
+        }
+        return false;
+    }
+
+    synchronized boolean isInReadOnlyMode() {
+        return bookieMode.equals(BookieMode.READ_ONLY);
```
eolivelli: Can't we use '==' for enums ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
```diff
@@ -0,0 +1,246 @@
...
+
+    /**
+     * Write content to the file. If file does not exist, it will create one.
+     *
+     * @param file file that you want to write to
+     * @param body content to write
+     * @throws IOException
+     */
+    private void writeToFile(File file, String body) throws IOException {
+        FileOutputStream fos = new FileOutputStream(file);
```
eolivelli: it is better to use try(FileOutputStream ....) {}

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    /**
+     * Write content to the file. If file does not exist, it will create one.
+     *
+     * @param file file that you want to write to
+     * @param body content to write
+     * @throws IOException
+     */
+    private void writeToFile(File file, String body) throws IOException {
```
eolivelli: this method can be 'static'

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
```diff
@@ -0,0 +1,246 @@
...
+     * @param file file to read from
+     * @return BookieStatus if not error, null if file not exist or any exception happens
+     * @throws IOException
+     */
+    private BookieStatus readFromFile(File file)
+            throws IOException, IllegalArgumentException {
+        if (!file.exists()) {
+            return null;
+        }
+        BufferedReader reader = new BufferedReader(
```
eolivelli: we can use try(BufferedReader...) {}

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
```diff
@@ -0,0 +1,246 @@
...
+     * @return true if successfully read from file, otherwise false.
+     */
+    boolean readFromDirectories(List<File> directories) {
+        boolean success = false;
+        for (File dir : directories) {
+            File statusFile = new File(dir, BOOKIE_STATUS_FILENAME);
+            try {
+                BookieStatus status = readFromFile(statusFile);
+                if (null != status) {
+                    synchronized (status) {
```
eolivelli: status is a local variable do we need a 'synchronized' ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
```diff
@@ -0,0 +1,246 @@
...
+        if (line == null || line.trim().isEmpty()) {
+            LOG.debug("Empty line when parsing bookie status");
+            return null;
+        }
+        String[] parts = line.split(",");
+        if (parts.length == 0) {
+            LOG.debug("Error in parsing bookie status: {}", line);
+            return null;
+        }
+        synchronized (status) {
```
eolivelli: 'status' is local, we no not need 'synchronized'

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieInitializationTest.java
```diff
@@ -265,17 +275,23 @@ public void testDuplicateBookieServerStartup() throws Exception {
...
         // starting bk server with same conf
         try {
-            BookieServer bs2 = new BookieServer(conf);
+            bs2 = new BookieServer(conf);
             bs2.start();
             fail("Should throw BindException, as the bk server is already running!");
         } catch (BindException e) {
-            // Ok
+            Assert.assertTrue("BKServer allowed duplicate startups!",
+                    e.getMessage().contains("Address already in use"));
```
eolivelli: this line of code was dropped in 4.4, because e.getMessage().contains depends on the language of the SO, in Italian for instance it says 'Indirizzo già in uso' and the test cannot be run on Italian machines
see https://issues.apache.org/jira/browse/BOOKKEEPER-904

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieInitializationTest.java
```diff
@@ -632,4 +648,174 @@ private ZooKeeperClient createNewZKClient() throws Exception {
...
+        // restart the bookie should start in writable state
+        bookieServer.shutdown();
+        bookieServer = new BookieServer(conf);
+        bookieServer.start();
+        bookie = bookieServer.getBookie();
+        assertFalse(bookie.isReadOnly());
+        bookieServer.shutdown();
+    }
+
+    private void corrupteFile(File file) throws IOException {
```
eolivelli: typo 'corrupt' ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -149,6 +150,8 @@
     private volatile boolean running = false;
     // Flag identify whether it is in shutting down progress
     private volatile boolean shuttingdown = false;
+    // Bookie status
+    private BookieStatus bookieStatus = new BookieStatus();
```
eribeiro: `bookieStatus` can be made `final` as it its object reference is not expected to change for the life of the Bookie (but the state will).

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -748,7 +751,13 @@ public Number getDefaultValue() {
 
             @Override
             public Number getSample() {
-                return zkRegistered.get() ? (readOnly.get() ? 0 : 1) : -1;
+                if(!zkRegistered.get()){
```
eribeiro: nit: space between `if` and `(`

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
```diff
@@ -0,0 +1,246 @@
...
+     */
+    boolean readFromDirectories(List<File> directories) {
+        boolean success = false;
+        for (File dir : directories) {
+            File statusFile = new File(dir, BOOKIE_STATUS_FILENAME);
+            try {
+                BookieStatus status = readFromFile(statusFile);
+                if (null != status) {
+                    synchronized (status) {
+                        if (status.lastUpdateTime > this.lastUpdateTime) {
```
eribeiro: nit: you could use `maybeUpdate()` method on `Bookie` as below:
```java
private boolean maybeUpdate(Bookie status) {
     if (status.lastUpdateTime > this.lastUpdateTime) {
         this.lastUpdateTime = status.lastUpdateTime;
         this.layoutVersion = status.layoutVersion;
         this.bookieMode = status.bookieMode;
         return true;
     }
     return false;
}
```

then here:
```java
     success = maybeUpdate(status);
```

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
eribeiro: +1

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
```diff
@@ -0,0 +1,246 @@
...
+                }
+            } catch (IOException e) {
+                LOG.warn("IOException while trying to read bookie status from directory {}." +
+                    " This is fine if not all directories failed.", dir);
+            } catch (IllegalArgumentException e ){
+                LOG.warn("IllegalArgumentException while trying to read bookie status from directory {}." +
+                    " This is fine if not all directories failed.", dir);
+            }
+        }
+        if (success) {
```
eribeiro: Sincere doubt: what is the semantics of `success` here? It can be false because the directories are empty (were cleaned?). Or maybe it was a stale file with a stale `lastUpdateTime`? In this case, a `success == false` not necessarily means that it "failed" to retrieve status from disk, but that there was not valid status to retrieve after all. What I am missing?

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
```diff
@@ -1468,6 +1470,28 @@ public float getDiskUsageWarnThreshold() {
...
+     *            - true if persist the bookie status. Otherwise false.
+     * @return ServerConfiguration
+     */
+    public ServerConfiguration setPersistBookieStatusEnabled(boolean enabled) {
+        setProperty(PERSIST_BOOKIE_STATUS_ENABLED, enabled);
+        return this;
+    }
+
+    /**
+     * Get whether the Bookie is force started in read only mode or not
```
eribeiro: Oops, this comment doesn't reflect the code after all. Copy and paste❓

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -857,7 +884,13 @@ public synchronized void start() {
 
         ledgerStorage.start();
 
-        syncThread.start();
+        // check the bookie status to start with
+        if (forceReadOnly.get()) {
```
eribeiro: if force read only and read status from disk are mutually exclusive then this needs to be documented somewhere (that is, which one has precedence), right?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    /**
+     * Write bookie status to multiple directories in best effort
+     *
+     * @param directories list of directories to write to
+     * @return true if any write succeed, false if all failed
+     *
+     */
+    synchronized boolean writeToDirectories(List<File> directories) {
```
eribeiro: As far as I can see, the `boolean` is not used anywhere, so maybe it could be dropped in favor of `void`?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
```diff
@@ -0,0 +1,246 @@
...
+            File statusFile = new File(dir, BOOKIE_STATUS_FILENAME);
+            try {
+                BookieStatus status = readFromFile(statusFile);
+                if (null != status) {
+                    synchronized (status) {
+                        if (status.lastUpdateTime > this.lastUpdateTime) {
+                            this.lastUpdateTime = status.lastUpdateTime;
+                            this.layoutVersion = status.layoutVersion;
+                            this.bookieMode = status.bookieMode;
+                            success = true;
```
eribeiro: Once it has successfully read a status from a file, wouldn't it leave the loop by the means of a break?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
sijie: I am not sure if we need use '=='. equals has more readability and consistency in java world than ==.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
sijie: done

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
sijie: done

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
sijie: yes, forceReadOnly will override status from disk. The forceReadOnly is set when setForceReadOnly is configured. That is documented in service configuration.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
```diff
@@ -0,0 +1,246 @@
...
+import java.io.OutputStreamWriter;
+import java.util.List;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * The status object represents the current status of a bookie instance.
+ */
+public class BookieStatus {
+    static Logger LOG = LoggerFactory.getLogger(BookieStatus.class);
```
sijie: done

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
```diff
@@ -0,0 +1,246 @@
...
+        this.layoutVersion = CURRENT_STATUS_LAYOUT_VERSION;
+        this.lastUpdateTime = INVALID_UPDATE_TIME;
+    }
+
+    private synchronized BookieMode getBookieMode() {
+        return bookieMode;
+    }
+
+    public synchronized boolean isInWritable() {
+        return bookieMode.equals(BookieMode.READ_WRITE);
```
sijie: prefer using equals than == to maintain readability and consistency.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    synchronized boolean setToWritableMode() {
+        if (!bookieMode.equals(BookieMode.READ_WRITE)) {
```
sijie: 1. this is a port change. I would keep naming to the original author if it is not a big issue.

2. prefer equals over ==

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
```diff
@@ -0,0 +1,246 @@
...
+        }
+        return false;
+    }
+
+    synchronized boolean isInReadOnlyMode() {
+        return bookieMode.equals(BookieMode.READ_ONLY);
+    }
+
+    synchronized boolean setToReadOnlyMode() {
+        if (!bookieMode.equals(BookieMode.READ_ONLY)) {
```
sijie: same as above.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
sijie: dropped.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
sijie: @eolivelli the finally block is required for bw. so keep them consistent in final block.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
sijie: changed.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
sijie: @eolivelli yes, we need synchronized. otherwise findbugs will report errors.

@eribeiro you can't synchronize on this because this is a different object.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
sijie: done

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
```diff
@@ -0,0 +1,246 @@
...
+        FileOutputStream fos = new FileOutputStream(file);
+        BufferedWriter bw = null;
+        try {
+            bw = new BufferedWriter(new OutputStreamWriter(fos, UTF_8));
+            bw.write(body);
+        } finally {
+            if (bw != null) {
+                bw.close();
+            }
+            fos.close();
```
sijie: it is needed. because bw can be null.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    /**
+     * Read bookie status from the status files, and update the bookie status if read succeed.
+     * If a status file is not readable or not found, it will skip and try to read from the next file.
+     *
+     * @param directories list of directories that store the status file
+     * @return true if successfully read from file, otherwise false.
+     */
+    boolean readFromDirectories(List<File> directories) {
```
sijie: dropped the return value.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
sijie: no, we need to get the latest valid status

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
sijie: false means there is no valid bookie status (it can be corrupted or empty).

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
```diff
@@ -0,0 +1,246 @@
...
+     * Function to read the bookie status from a single file
+     *
+     * @param file file to read from
+     * @return BookieStatus if not error, null if file not exist or any exception happens
+     * @throws IOException
+     */
+    private BookieStatus readFromFile(File file)
+            throws IOException, IllegalArgumentException {
+        if (!file.exists()) {
+            return null;
```
sijie: this is not an exception, because if there is no file, it means there is not status stored on disks.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
sijie: done

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
sijie: we need to synchronize otherwise findbugs report errors.

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
sijie: done

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieInitializationTest.java
sijie: reverted

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieInitializationTest.java
sijie: done

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieInitializationTest.java
```diff
@@ -632,4 +648,174 @@ private ZooKeeperClient createNewZKClient() throws Exception {
...
+        try {
+            bw = new BufferedWriter(new OutputStreamWriter(fos, UTF_8));
+            byte[] bytes = new byte[64];
+            new Random().nextBytes(bytes);
+            bw.write(new String(bytes));
+        } finally {
+            if (bw != null) {
+                bw.close();
+            }
+            fos.close();
```
sijie: bw can be null

sijie:
addressed the comments.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieInitializationTest.java
eribeiro: Oh. Sure. Sorry.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
eribeiro: No problem. :+1:

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
eribeiro: Fair enough. :+1:

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
eribeiro: :+1:

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
eribeiro: @sijie What about making `status` final?

`final BookieStatus status = readFromFile(statusFile);`

The static code analiser I use (Intellij) reports bugs when synchronizing on non final objects...

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
eribeiro: Yeah, makes sense. Thanks!

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
jvrao: @sijie can you please describe the usecase of this feature?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
jvrao: What if we have multiple ledger directories and have different/conflicting status and last update times in all those files?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
sijie: @jvrao : this allows turn a bookie to readonly or writable on-the-fly and keep the bookie status regarding restarts. so you can hook this with http endpoint and your deployment/administration tools to do automatic operations.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
sijie: the status with latest 'lastUpdateTime' wins.

yzang:
One of the original goals of persisting bookie status was to make it possible in the future to set the bookie to read only through http endpoint and persist the status so that even after the bookie is restarted, it should still run in read only mode.

However, after a second thought, I think the current this commit itself is probably not able to achieve this goal, because when a bookie start, it restore the read only status from disk, it will still transition to writable state as the DiskChecker runs and detect there's writable disk.

So I think maybe in the future, we need to add another status in the file to represent forceReadonlyMode, what do you think? @sijie

sijie:
@yzang : I think the ideal override sequence of write/readonly status would be:

1. if a bookie is started with '--readonly' option, it will be forced readonly (ignoring the bookie status persisted in the disks and ignoring the real disk usage) 
2. if a bookie is started without '--readonly' option, it will first respect the status persisted in the disks (because that is the status set for operation purpose).
3. if a bookie is started without '--readonly' option and there is not status persisted in the disks, that it will transition based on the read disk usage.

3) is already there regarding this change, 1) is achieved by this change, 2) is not ideal without this approach yet, we can make the case one the http endpoint is merged we can enable this there.

I don't think you need another status field in this file. In the BookieStatus, you can distinguish 2) and 3) easily, 2) is the status persisted while 3) is the in-memory status. the final status is if 2) is readonly, the final status will be readonly regarding what status is 3), if 2) is writable, the final status will be the status of 3). Does that sound good? @yzang

yzang:
@sijie That make sense, I understand what you mean, but my concern is:

Even if we persist the bookie status on disk, when we set the bookie to read only through http endpoint, the bookie will still be transitioned back to writable very quickly, because the DiskChecker runs periodically and would do the transition on each check, so there's no point setting the bookie to readonly on the fly, because the status still can't be persisted.

I think we probably need a http endpoint that can set the bookie to forceReadOnly mode or cancel the forceReadOnly mode, so that we can choose to ignore any further transition operation. In order to do that, we need to persist this forceReadOnly option on disk, so that when the bookie get restarted, it should start with --readonly option.

sijie:
@yzang sorry for late response.

I was suggesting the status in BookieStatus is comprised of two parts : one is persisted status read from status file (let's call it 'persisted_status') and the other one is in-memory status that determined by disk usage (let's call it 'in_memory_status'). In non-force-readonly mode, the final status is determined by these two, if persisted_status is readonly, there is not transition will happen from readonly to writable.

it is similar idea as what your proposed, but you don't have to persist another state. you just need to add another in-memory field in the bookie status class, and expose a method 'canTransitionToWritable'. so when disk listener detects disks are freed, it will check bookie status to see if it can transition to writable or not and decide the action.

Does this make sense to you? @yzang

yzang:
Yup, adding a flag to disable auto status transition sounds like a good idea to me.

sijie:
@yzang cool. I assume that would be future improvement, not related to this pull request. for this change, we will make this persistable bookie status available, right?

yzang:
@sijie Yeah, I think this change LGTM.

sijie:
@yzang I would probably defer this to 4.6.0. Is that okay to you?

yzang:
@sijie Sure, feel free to defer, it's not a critical feature to 4.5

sijie:
thank you @yzang

sijie:
Going to pick this up and merge it, if there is no objections.

zhaijack:
Seems there is a confulicting file, may need an update.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
jvrao: So force is only one-way operation; we have no forceWrite option. i.e if bookieStatus says readonly there is no way to start bookie in RW until we change bookieStatus. I am still at large don't understand the need of this change, but as long as we document it well, I am fine. I think it is mostly an operational overhead than letting operational scripts deal with starting bookie in R/RW.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
```diff
@@ -0,0 +1,234 @@
...
+     *
+     */
+    synchronized boolean writeToDirectories(List<File> directories) {
+        boolean success = false;
+        for (File dir : directories) {
+            try {
+                File statusFile = new File(dir, BOOKIE_STATUS_FILENAME);
+                writeToFile(statusFile, toString());
+                success = true;
+            } catch (IOException e) {
```
jvrao: How do you handle a case where you were able to write to few dirs and was not able to write to few dirs? What happens if a disk that has the mode got corrupted/not bootable?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
```diff
@@ -0,0 +1,234 @@
...
+     * @param directories list of directories to write to
+     * @return true if any write succeed, false if all failed
+     *
+     */
+    synchronized boolean writeToDirectories(List<File> directories) {
+        boolean success = false;
+        for (File dir : directories) {
+            try {
+                File statusFile = new File(dir, BOOKIE_STATUS_FILENAME);
+                writeToFile(statusFile, toString());
```
jvrao: Why setting the mode and writing to disk is not an atomic operation? What happens if bookie crash before writing to file, or after writing to a couple of dirs?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
jvrao: :) This is very confusing. It will be a nightmare to figure out the behavior if we have multiple disks and each has weird status, as an operator to know what exactly going on with the bookie.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
sijie: It is not a big deal though: 1) it is a best-effort operation. because it stores locally, the status can be gone when disk becomes broken 2) you might have inconsistent state on the disks, the status file with latest modification time can win.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
sijie: why do you need an atomic operation here? I don't quite get a use case need this to be atomic.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
jvrao: Exactly. That is why I am saying it is confusing. Take this scenario.
1. Bookie has two journal dirs.
2. bookie info persistence was able to save to dir1, and not to dir2.
case-1: Bookie restarted, and dir1 and dir2 are readable. Bookie comes up in read-only.
case-2: Bookie restarted, and dir1 is good/readable but not dir2. Bookie comes up in read-only.
case-3: Bookie restarted, and dir1 is not readable, but dir2 is. Bookie comes up in read-write mode.

We can think to bring down the entire bookie if some of the ledgerdirs are not available, but that is a big hammer. All I am saying is this going to be operational complex to know what to expect exactly. But given that this is disabled by default, I am fine with it.

sijie: it depends. we used same pattern in other internal systems, it isn't actually complicated. it is actually very simple and useful. because before this change, there is no way for you to ensure a bookie stays in readonly state among restarts. you have to specify --readonly or configure this in configuration when starting the bookie, that actually puts more operations overheads when you operate a 500+ bookies cluster.

asfgit:
https://builds.apache.org/job/bookkeeper-precommit-pullrequest-docker/140/

eolivelli:
@sijie  what is the status of this patch ? are we waiting for @jvrao approval ?

sijie:
@eolivelli we are ready to merge it.  @jvrao approved that in the last meeting.

I will let the ci run again.

sijie:
retest this please

asfgit:
https://builds.apache.org/job/bookkeeper-precommit-pullrequest-docker/151/

asfgit:
https://builds.apache.org/job/bookkeeper-precommit-pullrequest-docker/158/

sijie:
CI passed. merging it now.

PR #275, opened 2017-07-22
BOOKKEEPER-1102: Clarify BookieInfoReader and fix associated test flappers

athanatos:
BookieInfoReader:

The previous syncronization logic wasn't really correct, and the logic
at the top of the method was far more complicated than it needed to be.
Restrict bookies to be non-null.  Restructure the code to simply use
the BookieInfoReader instance as a single lock.

One significant behavioral change is that we scan every bookie not in
the map, and we clear from the map bookies which returned an error.

Also, explicitely cache the most recent bookie set reported by the
BookieWatcher.  This eliminates the need to call into BookieWatcher
from getReadWriteBookieInfo and the corresponding error path.  The
periodic scan continues to explicitely check.

Another departure is the addition of an explicit retry-on-error param to
trigger retry if any of the requests failed
(getBookieInfoRetryIntervalSeconds).  We'll only retry the ones that
actually failed (along with any new additions since the last run).  This
is useful because bookie startup triggers the addition of the bookie
node to zk before the bookie actually becomes available for the bookie
info request, so there can be rare races in the unit tests between
BookieInfoReader requesting the info and the bookie actually being up.

Also, add a method to allow tests to wait for updates to be reflected.

PerChannelBookieClient: fix error handling for BookieInfo keys

Passing a key corresponding to a GET_BOOKIE_INFO operation to
errorOutReadKey results in a casting exception, clean up the invalid
calls.

BookKeeperClusterTestCase: add killBookieAndWaitForZK

Should reduce the need for tests to wait for an arbitrary period to let
the cluster "settle".

BookKeeperDiskSpaceWeightedLedgerPlacementTest:

This test was heavily time dependent, and the Thread.sleep values did
not work universally.  Instead, eliminate the arbitrary Thread.sleep
values and instead verify the free space changes on each change.

Also, switch the delay on
testDiskSpaceWeightedBookieSelectionWithPeriodicBookieInfoUpdate
to simply use an atomic boolean to signal the value switch.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookieInfoReader.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookieInfoReader.java
@@ -17,15 +17,17 @@
  */
 package org.apache.bookkeeper.client;
 
+import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import org.apache.commons.collections.CollectionUtils;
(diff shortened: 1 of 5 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookieInfoReader.java
```diff
@@ -80,121 +78,278 @@ public String toString() {
...
+            return infoMap;
+        }
+    }
+    private final BookieInfoMap bookieInfoMap = new BookieInfoMap();
+
+    /**
+     * Tracks whether there is an execution in progress as well as whether
+     * another is pending.
+     */
+    static final int UNQUEUED = 0;
```
sijie: is there any reason not using an Enum?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookieInfoReader.java
```diff
@@ -80,121 +78,278 @@ public String toString() {
...
-                    LOG.debug("Running periodic BookieInfo scan");
+                synchronized (BookieInfoReader.this) {
+                    if (LOG.isDebugEnabled()) {
+                        LOG.debug("Running periodic BookieInfo scan");
+                    }
+                    try {
+                        Collection<BookieSocketAddress> updatedBookies = bk.bookieWatcher.getBookies();
+                        bookieInfoMap.updateBookies(updatedBookies);
+                    } catch (BKException e) {
+                        LOG.info("Got exception {} while querying bookies from watcher, rerunning after {}s",
```
sijie: any reason the put exception as the first parameter? it is a common practice in slf4j to put exception as the last parameter, and you don't need '{}' for exceptions.

eolivelli:
I think that this patch does more than fixing a test case. Maybe we should change the title

eolivelli:
@athanatos is this your current code at Salesforce or is it just a proposal ? as far as I know you are the only current user of DiskSpaceWeightedLedgerPlacement functions

athanatos:
@eolivelli It is our current code at SF as of last week.

eolivelli:
@athanatos could you address @sijie comments ? the patch is almost ready to go

athanatos:
@eolivelli Will do today.

athanatos:
@sijie @eolivelli Ok, updated.

jiazhai:
Seems there is a conflicting file.

PR #276, opened 2017-07-24
Issue-271 LedgerHandle#readEntries leaks ByteBufs

eolivelli:
Add a setNettyUsePooledBuffers() client configuration option to let the user decide to use Pooled vs Unpooled ByteBufs.
Using v2 wire protocol Application receives the original ByteBuf return from Channel and it is responsible for 'releasing' it.

Add assertions on LedgerEntry to prevent the user from accessing data more then once

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerEntry.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerEntry.java
@@ -24,6 +24,7 @@ import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufInputStream;
 
 import java.io.InputStream;
+import org.apache.bookkeeper.conf.ClientConfiguration;
 
 /**
  * Ledger entry. Its a simple tuple containing the ledger id, the entry-id, and
@@ -53,22 +54,49 @@ public class LedgerEntry {
         return length;
     }
 
+    /**
+     * Returns the content of the entry.
+     * This method can be called only once.
+     * While using v2 wire protocol this method will automatically release the internal ByteBuf
+     * @return
+     */
     public byte[] getEntry() {
+        if (data == null) {
+            throw new IllegalStateException("entry content can be accessed only once");
+        }
         byte[] entry = new byte[data.readableBytes()];
         data.readBytes(entry);
         data.release();
+        data = null;
         return entry;
     }
 
+    /**
+     * Returns the content of the entry.
+     * This method can be called only once.
+     * While using v2 wire protocol this method will automatically release the internal ByteBuf when calling the close
+     * method of the returned InputStream
+     *
+     * @return an InputStream which gives access to the content of the entry
+     */
     public InputStream getEntryInputStream() {
-        return new ByteBufInputStream(data);
+        if (data == null) {
+            throw new IllegalStateException("entry content can be accessed only once");
+        }
+        ByteBufInputStream res = new ByteBufInputStream(data);
+        data = null;
+        return res;
     }
 
     /**
      * Return the internal buffer that contains the entry payload.
      * <p>
      *
-     * Note: It is responsibility of the caller to ensure to release the buffer after usage.
+     * Note: Using v2 wire protocol it is responsibility of the caller to ensure to release the buffer after usage.
+     *
+     * @return a ByteBuf which contains the data
+     *
+     * @see ClientConfiguration#setNettyUsePooledBuffers(boolean)
      */
     public ByteBuf getEntryBuffer() {
         return data;
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerEntry.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerEntry.java
@@ -20,6 +20,7 @@
  */
 package org.apache.bookkeeper.client;
 
+import com.google.common.base.Preconditions;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufInputStream;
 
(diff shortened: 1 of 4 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerEntry.java
```diff
@@ -53,22 +54,49 @@ public long getLength() {
         return length;
     }
 
+    /**
+     * Returns the content of the entry.
+     * This method can be called only once.
+     * While using v2 wire protocol this method will automatically release the internal ByteBuf
+     * @return
```
sijie: if you are adding javadoc, please follow the new checkstyle rules.

`
Returns ...

<p>This method can ...
While using ...

@return ....
`

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerEntry.java
```diff
@@ -53,22 +54,49 @@ public long getLength() {
...
+     */
     public byte[] getEntry() {
+        if (data == null) {
```
sijie: I don't understand here:

why do you need this change here? it doesn't seem to be necessary.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerEntry.java
```diff
@@ -53,22 +54,49 @@ public long getLength() {
...
+     * Returns the content of the entry.
+     * This method can be called only once.
+     * While using v2 wire protocol this method will automatically release the internal ByteBuf when calling the close
+     * method of the returned InputStream
+     *
+     * @return an InputStream which gives access to the content of the entry
+     */
     public InputStream getEntryInputStream() {
-        return new ByteBufInputStream(data);
+        if (data == null) {
```
sijie: same question: it doesn't seem to be necessary?

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ClientConfiguration.java
```diff
@@ -114,6 +116,10 @@
     protected final static String ENABLE_BOOKIE_FAILURE_TRACKING = "enableBookieFailureTracking";
     protected final static String BOOKIE_FAILURE_HISTORY_EXPIRATION_MS = "bookieFailureHistoryExpirationMSec";
 
+    // Netty
```
sijie: can you move this parameter to section '  // NIO Parameters '?

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ClientConfiguration.java
```diff
@@ -114,6 +116,10 @@
...
+    protected final static String NETTY_USE_POOLED_BUFFERS = "nettyUsePooledBuffers";
+    protected final static boolean DEFAULT_NETTY_USE_POOLED_BUFFERS = true;
```
sijie: it is not a common practice to have the default value here. I'd prefer stick to current practice. If we are going to change the practice, let's change them all.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerEntry.java
eolivelli: @sijie the first time getEntry is called we will set data to null, this way the client cannot re-use the buffer anymore.
On an eventual "second call" we will throw an IllegalStateException which explains the error.
if I drop the 'if' the user will get a NullPointerException, but this is not very clear and IMHO it is bad practice.

Unfortunately the ByteBuf can be used only once, because we have no clue to reset the original start position for the entry

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerEntry.java
eolivelli: same as above

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ClientConfiguration.java
eolivelli: done

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ClientConfiguration.java
eolivelli: ok, done

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerEntry.java
sijie: @eolivelli 

there are two separate questions here.

1) are you going to change this behavior to support multiple gets?
2) if you are not going to change this behavior, why do we need you change?

The first question is more about whether to keep the behavior same as before. The behavior without your code change is exactly same as before. If you want to change this behavior, you can mark the read position, read the data and reset the read position. but your change doesn't seem to be this case.

for 2) change, without your change, it still throws IlegalStateException because the bytebuf will be released twice. so I don't see a strong reason why do we need your change here. I am not talking about the 'if (data == null)' case, I am talking about the change in general. Because I do not see a strong reason to change this if we are not changing the behavior.

eolivelli: @sijie 
I don't want to change the behavior, I am fine with calling getEntry only once. I only wanted to make it more clear.
When I and my colleague Diego saw actual code we got a little worried and we needed a double check with you and Matteo on the mailing list to be sure that there was no real problem.

Anyway I am ok with reverting for the getEntry() as the IlegalStateException will be thrown in ByteBuf#release().

But for the InputStream case I think it will be too late to see the error at the 'close' of the inputstream. I will revert if you prefer, no problem

sijie: @eolivelli you don't have to revert it. it is good to make it clear. I just want to clarify to make things clear at this pull request. for assertion, I would prefer using guava Preconditions.checkState(...).

eolivelli: @sijie I will update the patch as soon as possibile.
Thank you

PR #297, opened 2017-07-25
Issue 296: Bookie supports ephemeral port

zhaijack:

- add a flag to allow/disable using ephemeral ports
- change the initialization sequence to support ephmeral port
- added two tests on verifying this behavior

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieNettyServer.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieNettyServer.java
@@ -83,7 +83,7 @@ class BookieNettyServer {
     final ServerConfiguration conf;
     final EventLoopGroup eventLoopGroup;
     final EventLoopGroup jvmEventLoopGroup;
-    final RequestProcessor requestProcessor;
+    RequestProcessor requestProcessor;
     final AtomicBoolean isRunning = new AtomicBoolean(false);
     final AtomicBoolean isClosed = new AtomicBoolean(false);
     final Object suspensionLock = new Object();
@@ -140,6 +140,11 @@ class BookieNettyServer {
         listenOn(bindAddress, bookieAddress);
     }
 
+    public BookieNettyServer setRequestProcessor(RequestProcessor processor) {
+        this.requestProcessor = processor;
+        return this;
+    }
+
     boolean isRunning() {
         return isRunning.get();
     }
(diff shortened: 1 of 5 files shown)


zhaijack:
ported the change for @sijie

eolivelli:
@zhaijack @sijie 
In DL tests do you need non-loopback addresses ? Another approach can be to allow port=0 (and automatic choose) on loopback interfaces (which in turn is to be enabled explicitly).

Another approach, only for tests, is to allow "local transport" (enableLocalTransport=true) and disabile "network transport" (disableServerSocketBind=true), this is one of the reasons for which I have introduced such features

sijie:
@eolivelli 

we use loopback addresses. I don't see any reason to combine ephemeral  port with loopback address together. they have different purposes though.

ephemeral port is also different from local transport. using ephemeral port, we are exercising all the network transportation, which is important for all use cases.

eolivelli:
@sijie I understand your point, I was only suggesting another approach, the local-transport is not good in your case.

What about not adding the new configuration option and allowing the use of ephemeral port on loopback address ? loopback is normally used only in tests
my concern is only to not add a dangerous option to be used in production.

eolivelli:
@sijie  sorry I just re-read your comment.
OK for the new option, we are going to use non-loopback + ephemeral for tests

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieNettyServer.java
```diff
@@ -219,7 +225,12 @@ protected void initChannel(SocketChannel ch) throws Exception {
             });
 
             // Bind and start to accept incoming connections
-            bootstrap.bind(address.getAddress(), address.getPort()).sync();
+            Channel listen = bootstrap.bind(address.getAddress(), address.getPort()).sync().channel();
+            if (listen.localAddress() instanceof InetSocketAddress) {
+                if (conf.getBookiePort() == 0) {
+                    conf.setBookiePort(((InetSocketAddress) listen.localAddress()).getPort());
```
eolivelli: @jiazhai @sijie is there any other case in which we "change" the provided configuration object ?
wouldn't it be clearer to add a public getBookieSocketAddress() ?
if we do not have any other case it should be better to treat ServerConfiguration as immutable once the server is started

sijie: @eolivelli currently bookie and netty server are separated, there is not cross referencing in between. The setBookiePort is propagated the real port to bookie for it to advertise its network location. you can't easily do getBookieSocketAddress() without any major refactor between bookie and netty server.

eolivelli: OK for me

PR #298, opened 2017-07-26
Issue 295: LocalBookKeeper fails to start up

zhaijack:

remote the zk port and zk servers from the classes and use the values passed in from the method.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/util/LocalBookKeeper.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/util/LocalBookKeeper.java
@@ -55,17 +55,15 @@ public class LocalBookKeeper {
     }
 
     public LocalBookKeeper(int numberOfBookies) {
-        this(numberOfBookies, 5000, ZooKeeperDefaultHost, ZooKeeperDefaultPort);
+        this(numberOfBookies, 5000);
     }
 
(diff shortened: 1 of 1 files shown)


eolivelli:
I have restarted the job 
https://builds.apache.org/job/bookkeeper-precommit-pullrequest/174/

eolivelli:
Jenkins is broken....I will try on my laptop

ivankelly:
I came across this the other day. It failed to start because there was setLoopbackAllowed needed to be set in the configuration.

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/LocalBookKeeper.java
```diff
@@ -299,6 +297,9 @@ static void startLocalBookiesInternal(ServerConfiguration conf,
                 }
                 throw ie;
             }
+        } catch (Exception e) {
```
merlimat: If we catch the exception here, it means we are not exiting the process. I would say to not catch here.

sijie: +1 on this. if we want logging, we should re-throw it again.

eolivelli:
I Am sorry. I have already closed the pr. I will create a new one and do the proposed fix from @merlimat

Issue #299, opened 2017-07-26
EnsemblePlacementPolicy in 4.5 is not compatible with 4.4 clients


eolivelli:
Is is not possible to use custom EnsemblePlacementPolicy compiled for BookKeeper 4.4 having 4.5 BookKeeper on the classpath.
It would be better to use method overloading and java8 default interface methods to make the transition to 4.5 simpler

eolivelli:
In my specific use-case I have custom policies which extend DefaultEnsemblePlacementPolicy, so a special treatment is to be done on DefaultEnsemblePlacementPolicy to make them work

eolivelli:
Closing as won't fix.
See the discussion on #301 
In this case it is really hard to preserve compatibility, because we are not dealing only with compile time/runtime compatibility, but the problem is about the semantics of the methods

In the future we will discuss how to preserve compatibility for public APIs

PR #301, opened 2017-07-26
Issue-299 EnsemblePlacementPolicy in 4.5 is not compatible with 4.4 clients

eolivelli:
Introduce overloaded methods in EnsemblePlacementPolicy in order to handle compatibility with 4.4 clients.
Modify DefaultEnsemblePlacementPolicy in order to let legacy EnsemblePlacementPolicy which extended DefaultEnsemblePlacementPolicy to work as expected

see issue #299

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/EnsemblePlacementPolicy.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/EnsemblePlacementPolicy.java
@@ -34,12 +34,17 @@ import org.apache.bookkeeper.feature.FeatureProvider;
 import org.apache.bookkeeper.net.BookieSocketAddress;
 import org.apache.bookkeeper.net.DNSToSwitchMapping;
 import org.apache.bookkeeper.stats.StatsLogger;
+import org.apache.commons.configuration.Configuration;
 
 /**
  * Encapsulation of the algorithm that selects a number of bookies from the cluster as an ensemble for storing
  * data, based on the data input as well as the node properties.
  */
 public interface EnsemblePlacementPolicy {
+
+    public default EnsemblePlacementPolicy initialize(Configuration conf) {
+        return this;
+    }
 
     /**
      * Initialize the policy.
@@ -50,16 +55,21 @@ public interface EnsemblePlacementPolicy {
      * @param featureProvider feature provider
      * @param statsLogger stats logger
      */
-    public EnsemblePlacementPolicy initialize(ClientConfiguration conf,
+    public default EnsemblePlacementPolicy initialize(ClientConfiguration conf,
                                               Optional<DNSToSwitchMapping> optionalDnsResolver,
                                               HashedWheelTimer hashedWheelTimer,
                                               FeatureProvider featureProvider,
-                                              StatsLogger statsLogger);
+                                              StatsLogger statsLogger) {
+        return initialize(conf);
+    }
+
+
 
     /**
      * Uninitialize the policy
      */
-    public void uninitalize();
+    public default void uninitalize() {
+    }
 
     /**
      * A consistent view of the cluster (what bookies are available as writable, what bookies are available as
@@ -89,7 +99,27 @@ public interface EnsemblePlacementPolicy {
      * @throws BKNotEnoughBookiesException if not enough bookies available.
      * @return the java.util.ArrayList<org.apache.bookkeeper.net.BookieSocketAddress>
      */
-    public ArrayList<BookieSocketAddress> newEnsemble(int ensembleSize, int writeQuorumSize, int ackQuorumSize, Map<String, byte[]> customMetadata, Set<BookieSocketAddress> excludeBookies) throws BKNotEnoughBookiesException;
+    public default ArrayList<BookieSocketAddress> newEnsemble(int ensembleSize, int writeQuorumSize,
+        int ackQuorumSize, Map<String, byte[]> customMetadata, Set<BookieSocketAddress> excludeBookies) throws BKNotEnoughBookiesException {
+        return newEnsemble(ensembleSize, ackQuorumSize, excludeBookies);
+    }
+
+    /**
+     * Legacy method introduce for compatibility with 4.4 clients
+     * @param ensembleSize
+     *          Ensemble Size
+     * @param quorumSize
+     *          Quorum Size
+     * @param excludeBookies
+     * @return
+     * @throws org.apache.bookkeeper.client.BKException.BKNotEnoughBookiesException
+     *
+     * @see #newEnsemble(int, int, int, java.util.Map, java.util.Set)
+     */
+    public default ArrayList<BookieSocketAddress> newEnsemble(int ensembleSize, int quorumSize,
+            Set<BookieSocketAddress> excludeBookies) throws BKException.BKNotEnoughBookiesException {
+        throw new UnsupportedOperationException("implement newEnsemble");
+    }
 
     /**
      * Choose a new bookie to replace <i>bookieToReplace</i>. If no bookie available in the cluster,
@@ -107,7 +137,24 @@ public interface EnsemblePlacementPolicy {
      * @throws BKNotEnoughBookiesException
      * @return the org.apache.bookkeeper.net.BookieSocketAddress
      */
-    public BookieSocketAddress replaceBookie(int ensembleSize, int writeQuorumSize, int ackQuorumSize, java.util.Map<String, byte[]> customMetadata, Collection<BookieSocketAddress> currentEnsemble, BookieSocketAddress bookieToReplace, Set<BookieSocketAddress> excludeBookies) throws BKNotEnoughBookiesException;
+    public default BookieSocketAddress replaceBookie(int ensembleSize, int writeQuorumSize, int ackQuorumSize,
+        java.util.Map<String, byte[]> customMetadata, Collection<BookieSocketAddress> currentEnsemble,
+        BookieSocketAddress bookieToReplace, Set<BookieSocketAddress> excludeBookies) throws BKNotEnoughBookiesException {
+        return replaceBookie(bookieToReplace, excludeBookies);
+    }
+
+    /**
+     * Legacy method introduce for compatibility with 4.4 clients
+     * @param bookieToReplace bookie to replace
+     * @param excludeBookies bookies that should not be considered as candidate.
+     * @return the org.apache.bookkeeper.net.BookieSocketAddress
+     * @throws org.apache.bookkeeper.client.BKException.BKNotEnoughBookiesException
+     * @see #replaceBookie(int, int, int, java.util.Map, java.util.Collection, org.apache.bookkeeper.net.BookieSocketAddress, java.util.Set)
+     */
+    public default BookieSocketAddress replaceBookie(BookieSocketAddress bookieToReplace,
+            Set<BookieSocketAddress> excludeBookies) throws BKException.BKNotEnoughBookiesException {
+        throw new UnsupportedOperationException("implement replaceBookie");
+    }
 
     /**
      * Reorder the read sequence of a given write quorum <i>writeSet</i>.
@@ -120,8 +167,10 @@ public interface EnsemblePlacementPolicy {
      *          Observed failures on the bookies
      * @return read sequence of bookies
      */
-    public List<Integer> reorderReadSequence(ArrayList<BookieSocketAddress> ensemble,
-                                             List<Integer> writeSet, Map<BookieSocketAddress, Long> bookieFailureHistory);
+    public default List<Integer> reorderReadSequence(ArrayList<BookieSocketAddress> ensemble,
+                                             List<Integer> writeSet, Map<BookieSocketAddress, Long> bookieFailureHistory) {
+        return writeSet;
+    }
 
 
     /**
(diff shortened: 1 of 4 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/EnsemblePlacementPolicy.java
```diff
@@ -107,7 +137,24 @@ public EnsemblePlacementPolicy initialize(ClientConfiguration conf,
...
+
+    /**
+     * Legacy method introduce for compatibility with 4.4 clients
+     * @param bookieToReplace bookie to replace
+     * @param excludeBookies bookies that should not be considered as candidate.
+     * @return the org.apache.bookkeeper.net.BookieSocketAddress
+     * @throws org.apache.bookkeeper.client.BKException.BKNotEnoughBookiesException
+     * @see #replaceBookie(int, int, int, java.util.Map, java.util.Collection, org.apache.bookkeeper.net.BookieSocketAddress, java.util.Set)
+     */
+    public default BookieSocketAddress replaceBookie(BookieSocketAddress bookieToReplace,
```
sijie: same comment as above.

- the default behavior for the new method throws UnsupportedOperationException
- the legacy method should call the new method by default.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/EnsemblePlacementPolicy.java
```diff
@@ -135,12 +184,22 @@ public EnsemblePlacementPolicy initialize(ClientConfiguration conf,
      *          Observed failures on the bookies
      * @return read sequence of bookies
      */
-    public List<Integer> reorderReadLACSequence(ArrayList<BookieSocketAddress> ensemble,
-                                                List<Integer> writeSet, Map<BookieSocketAddress, Long> bookieFailureHistory);
+    public default List<Integer> reorderReadLACSequence(ArrayList<BookieSocketAddress> ensemble,
```
sijie: why not just return writeSet?

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/GenericEnsemblePlacementPolicyTest.java
```diff
@@ -132,9 +132,9 @@ public void testReplaceBookie() throws Exception {
                     lh.addEntry(value);
                 }
             }
-            assertEquals(2, customMetadataOnNewEnsembleStack.size());
```
sijie: this seems to indicate this change changes the placement behavior. in order to keep binary backward compatible, I would think adding method overloading should be good enough, and it should not change the placement behavior.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/LegacyCustomDefaultEnsemblePlacementPolicyTest.java
```diff
@@ -0,0 +1,113 @@
...
+import org.apache.bookkeeper.net.BookieSocketAddress;
+import org.apache.bookkeeper.test.BookKeeperClusterTestCase;
+import org.junit.Assert;
+import static org.junit.Assert.assertEquals;
+import org.junit.Before;
+import org.junit.Test;
+
+/**
+ * Test that policies which extended DefaultEnsemblePlacementPolicy of BookKeeper 4.4 work with current version
+ * @author enrico.olivelli
```
sijie: no author tag.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/LegacyCustomDefaultEnsemblePlacementPolicyTest.java
```diff
@@ -0,0 +1,113 @@
...
+import java.util.concurrent.atomic.AtomicBoolean;
+import org.apache.bookkeeper.net.BookieSocketAddress;
+import org.apache.bookkeeper.test.BookKeeperClusterTestCase;
+import org.junit.Assert;
+import static org.junit.Assert.assertEquals;
+import org.junit.Before;
+import org.junit.Test;
+
+/**
+ * Test that policies which extended DefaultEnsemblePlacementPolicy of BookKeeper 4.4 work with current version
```
sijie: add '.' to the end of this sentence.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/LegacyCustomDefaultEnsemblePlacementPolicyTest.java
```diff
@@ -0,0 +1,113 @@
...
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicBoolean;
+import org.apache.bookkeeper.net.BookieSocketAddress;
+import org.apache.bookkeeper.test.BookKeeperClusterTestCase;
+import org.junit.Assert;
+import static org.junit.Assert.assertEquals;
```
sijie: sort the imports using the checkstyle rules.

import static ...

import non-static ...

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/EnsemblePlacementPolicy.java
```diff
@@ -50,16 +55,21 @@
...
      * @param statsLogger stats logger
      */
-    public EnsemblePlacementPolicy initialize(ClientConfiguration conf,
+    public default EnsemblePlacementPolicy initialize(ClientConfiguration conf,
                                               Optional<DNSToSwitchMapping> optionalDnsResolver,
                                               HashedWheelTimer hashedWheelTimer,
                                               FeatureProvider featureProvider,
-                                              StatsLogger statsLogger);
+                                              StatsLogger statsLogger) {
+        return initialize(conf);
```
sijie: Can you just return 'this' here? and implement initilize(conf) by calling initialize(conf, Optional.empty(), null, NullFeatureProvider.INSTANCe, NullStatsLogger.INSTANCE). because initialize(conf) is a legacy method.

eolivelli: @sijie for this comment and the others, the problem is that we must call legacy "initialize" method of 4.4 policies.
So by default we will call the old initialize(conf) which will be noop

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/EnsemblePlacementPolicy.java
```diff
@@ -50,16 +55,21 @@
...
+    }
+
+
 
     /**
      * Uninitialize the policy
      */
-    public void uninitalize();
+    public default void uninitalize() {
```
eolivelli: ok

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/EnsemblePlacementPolicy.java
```diff
@@ -89,7 +99,27 @@ public EnsemblePlacementPolicy initialize(ClientConfiguration conf,
...
+     *          Ensemble Size
+     * @param quorumSize
+     *          Quorum Size
+     * @param excludeBookies
+     * @return
+     * @throws org.apache.bookkeeper.client.BKException.BKNotEnoughBookiesException
+     *
+     * @see #newEnsemble(int, int, int, java.util.Map, java.util.Set)
+     */
+    public default ArrayList<BookieSocketAddress> newEnsemble(int ensembleSize, int quorumSize,
```
eolivelli: as the prev comment, BookKeeper needs to call "new methods" which in turn will call "old methods".
In my case I always have subclasses of DefaultEnsemblePlacementPolicy, so I can change the public interface as you are proposing at make only subclasses of DefaultEnsemblePlacementPolicy call the 'legacy' methods

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/EnsemblePlacementPolicy.java
eolivelli: sorry, cut and past from Default Policy.  I will clean

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/GenericEnsemblePlacementPolicyTest.java
eolivelli: yes this is a "change", but it is not possible to do in another way, because we need to call the 'legacy' method by default  and the class in the the does not override it.
I have preferred to change the assertion instead of changing the test policy, the fact that the legacy method is called is covered by the new test class

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/LegacyCustomDefaultEnsemblePlacementPolicyTest.java
eolivelli: Ok, I'm sorry, I hope we can add checkstyle soon

eolivelli:
@sijie 
overall I am not very happy with this patch, as I told on the mailing list, addressing the problem with method overloading + default method interface is messy.

I would prefer to break compatibility once and create a better API, packaging method parameters in contextual beans
like

initialize(ClientEnvironment env)
newEnsemble(LedgerSpecs specs)
replaceBookie(LedgerSpecs specs, bookietoreplace, ReplaceContext ctx)
......

/cc @merlimat @jvrao @ivankelly 
in the future maybe we will add support for "Bookie pools/labels", ACLs (so user context....).....

sijie:
@eolivelli can we focus on addressing the problem for 4.5.0? I don't see a point to do any large refactor when close to a release. Also as I comment in the mail list, packing fields into a huge structure has its own problem, we don't have any consensus about this. personally I would defer this until it is really needed (for example as you said, if we want to introduce pools). We don't know yet how we want to deal with pools, it doesn't make any sense to introduce any refactor.

sijie:
@eolivelli as [BP-9](https://cwiki.apache.org/confluence/display/BOOKKEEPER/BP-9+-+Github+issues+for+Issue+Tracking) suggested, we should discuss features/designs in an issue (or an email thread). we should focus on the implementation problem on pull requests, otherwise the ideas/discussions just spread over different places and hard to manage.

eolivelli:
@sijie sure let's fix the problem for 4.5 doing the best as possible.

I do not want large refactors now too.
I will try to address your comments/suggestions soon

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/EnsemblePlacementPolicy.java
sijie: the way I suggested doesn't prevent you calling initialize(conf), no? 

- if your placement policy implements the interface directly, you anyway already implement initialize(conf)
- if you placement policy extends the default ensemble placement policy, this default ensemble placement policy implements the new initialize method hence the old initialize method is already implemented.

Your approach looks messy because the dependency is somehow very strange. if you follow the rule I suggested, it is very clear.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/EnsemblePlacementPolicy.java
sijie: see my comments as above.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/GenericEnsemblePlacementPolicyTest.java
sijie: I still don't understand why do you need to call 'legacy' method.  in 4.4.0, all the legacy methods are interface method (no default behavior), a customized implementation already implements them.

eolivelli: @sijie  for instance I have a custom policy, which implements only old methods like the one in:
LegacyCustomDefaultEnsemblePlacementPolicyTest.java 

this is a real example:
https://github.com/diennea/herddb/blob/master/herddb-core/src/main/java/herddb/cluster/PreferLocalBookiePlacementPolicy.java

this example does not even extends DefaultEnsamblePlacementPolicy, so my current patch will not work :-(

eolivelli:
@sijie maybe a diffent approach would be to add back the legacy methods with a default implementation like 'throw new UnsupportedOperationException'

than in BK code we will try to call the legacy version and in case of UnsupportedOperationException we will call the new implementation

this has a drawback to throw and catch an exception every time

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/GenericEnsemblePlacementPolicyTest.java
sijie: I see. the callers are the new bookkeeper client. if that's the case, I am not sure if we should support such backward compatibility. to me, it is not a simple backward compatibility issue, it is a semantic issue.

for example, the new placement policy introduces ackQuorumSize in newEnsemble and replica settings for replaceBookie. The new bookkeeper client expects the placement policy to be respecting the replica settings when choosing bookies. However the old placement policy implementation doesn't really respect these settings because it was implemented lack of those settings. This is a semantic disconnection between new client behavior and old placement policy implementation. I am not sure if this is a behavior that we should expose to applications, because there might be a lot of unknowns when new client uses old placement policy. same concerns applied even if we want to refactor it to use structures.

but anyway, I am fine the approach in this pull request, but we might consider marking the old methods as deprecated. so that people can upgrade to use new interface in future.

We might need to discuss the audience and stability of interfaces that exposed by bookkeeper in next release.

eolivelli: @sijie in usecases covered by me and my company I am already handling this problem by falling  back to DefaultEnsamblePlacementPolicy in case for non-compatibility.

So I am tempted to close the issue as "won't fix".
I wonder if other companies will have such problems, we should make a clear list of changes and maybe a guide to the migration to 4.5.
As you previously said this would be a 5.0 release, but let's stick to 4.5, we want to close as soon as possible

are you ok with closing the issue as won't fix ?

sijie: @eolivelli I don't know what others will be impacted. I do not see such binary compatibility issue a lot, because typically when a new version is available, most of the people would update their application with a newer version and compile it. this would be detected as compilation issues and fixed. 

Compilation issues are much safer and easier to deal with. In contrast, I did see a lot of runtime exceptions when newer clients uses old dependencies (when there aren't any troubles compiling new clients with old dependencies), e.g. method signatures are changed, method not found, classes not found. Those runtime exceptions are much dangerous than compilation issues. 

>> are you ok with closing the issue as won't fix

I am fine with that.

eolivelli:
Closing as won't fix

PR #309, opened 2017-07-27
Issue-308 LocalBookKeeper.java rethrow Exception in case of failed boot

eolivelli:
Do not swallow generic Exceptions during the boot of LocalBookKeeper and let the caller handle errors


sijie:
Can we keep the log message and rethrow the exception?

eolivelli:
@sijie  done

PR #313, opened 2017-07-28
Issue 300: EnsemblePlacementPolicy must not use com.google.common.base.Optional but java.util.Optional

zhaijack:

use java8 optional instead of guava optional for public interface

- other changes: when touching the files, also fixing checkstyle problems.


eolivelli:
@zhaijack 
I have already sent a PR #302
I am sorry for this duplicate work

eolivelli:
we can go with your patch if you prefer @merlimat already approved my pr and it is ready for merge
you addressed checkstyle too, maybe your patch is better

zhaijack:
Oh, Sorry for it.  my bad. I will close mine.

PR #315, opened 2017-07-28
Issue 314: Enable checkstyle on bookkeeper-server

zhaijack:

- enable checkstyle plugin in bookkeeper-server with a suppression configuration file
- enable sort order rule (it was disabled when the rule was imported)


zhaijack:
doing it now

PR #317, opened 2017-07-28
Issue-316 BookKeeperTest#testReadEntryReleaseByteBufs is failing

eolivelli:
Change the assertion in BookKeeperTest#testReadEntryReleaseByteBufs: expect a NullPointerException instead of a IllegalStateException

### DIFF
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookKeeperTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookKeeperTest.java
@@ -823,12 +823,12 @@ public class BookKeeperTest extends BaseTestCase {
                     try {
                         entry.getEntry();
                         fail("entry data accessed twice");
-                    } catch (IllegalStateException ok){
+                    } catch (NullPointerException ok){
                     }
                     try {
                         entry.getEntryInputStream();
                         fail("entry data accessed twice");
-                    } catch (IllegalStateException ok){
+                    } catch (NullPointerException ok){
                     }
                 }
             }


eolivelli:
There is one problem, I have chagne the assertion and if I run the single test it works as expected, but if I run the all class (BookKeeperTest) it fails at line 753, because ByteBuf#release does not return 'true', which means that this is not the 'last' release (refcount goes to 0)

I think that it is some kind of race condition, due to the fact that launching the full suite the JVM runs faster and faster while compiling code in the background.

I would like that @merlimat  or @sijie  o @kishorekasi could take a look.
If the assertion is not useful I will remove it

zhaijack:
- we should stick to IllegalStateException. before your recent change, the client is throwing IllegalStateException. we should use checkState rather than checkNull in LedgerEntry.
- the useUpooled setting broke the EnDecoder for v2 protocol.

I have a pull request to improve the tests, I will include this in that pull request.

eolivelli:
@zhaijack thank you for looking at this, I am ok with IllegalStateException @sijie are you ok ?

zhaijack:
there is in general a problem with #testReadEntryReleaseByteBufs - it made a very bad assumption that #release() call will be the last call in the tests. that would be true for v3 protocol, because the bytebuf returned in v3 protocol is always a new allocated now. However, in v2 protocol, the bytebuf is actually the buf from netty, netty might not release its references before the test ran.

eolivelli:
@zhaijack That was my suspect. I think it is better to drop the assertTrue(byteBuf.release()) a leave only  "byteBuf.release();"

zhaijack:
@eolivelli I included the fix for BookKeeperTest in #340

eolivelli:
Thank you very much @zhaijack I have checked out locally #340
I am reviewing, will give my +1 soon

sijie:
@eolivelli it seems that #340 addressed the problem here. can we close the pull request?

eolivelli:
already fixed with #340

PR #340, opened 2017-08-01
Issue 339: Use ephemeral port for bookkeeper tests and variable unit test improvements

zhaijack:

- Enable ephemeral port by default for all BookKeeperClusterTestCase based test cases
- Removed Multi* based classes. A lot of unit tests unnecessarily extended from those base classes. Converted them to extend BookKeeperClusterTestCase directly. Confirmed this with @sijie 
- Reduce sleep period for some test cases

LedgerEntry fixes for #317  
- use checkState rather than checkNull to keep the behavior consistent as before
- fix BookKeeperTest#testReadEntryReleaseByteBufs 

TestBackwardCompat:
- removed testCompact400 (400 has been very long time ago, no one is running 400)
- BookKeeperNettyServer should shut down netty server when failed to construct a bookie

TestDiskChecker:
- increased the threshold to 0.99f. be consistent with TestBKConfiguration

BookieInitializationTest
- make each test method using different ledgers path, so registration won't be conflicting between tests

BookieJournalRollingTest
- make sure bookie port is used correctly on restarting bookies.


eolivelli:
@zhaijack great work. I am checking on my laptop

eolivelli:
Overall is OK for me.
There is still some failing test, like these, but it is not always reproducible

```
testManyOpenParallel(org.apache.bookkeeper.client.TestFencing)  Time elapsed: 60.267 sec  <<< ERROR!
org.junit.runners.model.TestTimedOutException: test timed out after 60000 milliseconds
	at org.apache.bookkeeper.client.TestFencing.testManyOpenParallel(TestFencing.java:196)

```


```
testBookieRestartContinuously(org.apache.bookkeeper.bookie.BookieShutdownTest)  Time elapsed: 9.186 sec  <<< ERROR!
org.apache.bookkeeper.bookie.BookieException$InvalidCookieException: 
Cookie [4
bookieHost: "10.168.101.193:43929"
journalDir: "/tmp/bookie3163597913702501683test"
ledgerDirs: "1\t/tmp/bookie3163597913702501683test"
] is not matching with [4
bookieHost: "10.168.101.193:43929"
journalDir: "/tmp/bookie5441052039362779250test"
ledgerDirs: "1\t/tmp/bookie5441052039362779250test"
]
	at org.apache.bookkeeper.bookie.BookieShutdownTest.testBookieRestartContinuously(BookieShutdownTest.java:86)


```

zhaijack:
The problem of BookieShutdownTest#testBookieRestartContinously: it requires restarting bookies 100 times, and each time it should come up with a different port and different ledger directories. so set this test to not use ephemeral port and reduce the number of restarts to 100 to 10, to reduce the total running time of this test.

zhaijack:
TestFencing seems to be a flaky test for a while. I will create a separate issue for that.

PR #350, opened 2017-08-02
Issue 349: Documentation for security feature in 4.5.0

sijie:

- add a section for `security`
- overview to introduce security feature
- encryption and authentication using tls
- sasl authentication
- zookeeper authentication


sijie:
You can review the staging site here: https://sijie.github.io/bookkeeper-staging-site/docs/security/

sijie:
@kishorekasi : can you review the details for tls support?
@eolivelli : can you review the details for SASL and zookeeper authentication?

/cc @lucperkins @merlimat for reviews general reviews.

site/docs/security/tls.md
```diff
@@ -0,0 +1,178 @@
...
+
+## Signing the certificate
+
+The next step is to sign all certificates in the keystore with the CA we generated. First, you need to export the certificate from the keystore:
+
+    keytool -keystore bookie.keystore.jks -alias localhost -certreq -file cert-file
+
+Then sign it with the CA:
+
+    openssl x509 -req -CA ca-cert -CAkey ca-key -in cert-file -out cert-signed -days {validity} -CAcreateserial -passin pass:{ca-password}
```
merlimat: Use some highlighting to mark the parameters to fill up

site/docs/security/tls.md
```diff
@@ -0,0 +1,178 @@
...
+title: Encryption and Authentication using TLS
+prev: ../index.html
+next: ../sasl
+---
+
+Apache BookKeeper allows clients and autorecovery daemons to communicate over TLS, although this is not enabled by default.
+
+## Overview
+
+The bookies and clients need their own key and certificate in order to use TLS. The key is used for encryption, while the
```
merlimat: > clients need their own key and certificate

If you only want to do encryption, in general there's no need for the client to configure a certificate. A self-signed certificate should be created on the spot. (I have not checked if it's actually true in the current implementation)

site/docs/security/sasl.md
```diff
@@ -0,0 +1,197 @@
...
+next: ../zookeeper
+---
+
+Bookies support client authentication via SASL. Currently we only support GSSAPI (Kerberos). We will start
+with a general description of how to configure `SASL` for bookies, clients and autorecovery daemons, followed
+by mechanism-specific details and wrap up with some operational details.
+
+## SASL configuration for Bookies
+
+1. Select the mechanisms to enable in the bookies. `GSSAPI` is the only mechanism currently supported by BookKeeper.
```
eolivelli: We are supporting GSSAPI (kerberos) for production.
There is support for username/password usually good for tests and staging environments.
Actually this is known as DIGEST-MD5, so passwords are not sent on the wire but only digests.
As @revans2 states we should state clearly that DIGEST-MD5 is not suitable for production.

see MD5DigestBookKeeperTest

site/docs/security/sasl.md
```diff
@@ -0,0 +1,197 @@
...
+
+Follow the steps in [GSSAPI (Kerberos)](#kerberos) to configure SASL for the selected mechanism.
+
+## <a name="kerberos"></a> Authentication using SASL/Kerberos
+
+### Prerequisites
+
+#### Kerberos
+
+If your organization is already using a Kerberos server (for example, by using `Active Directory`), there is no need to
```
eolivelli: Honestly I never used Active Directory for Kerberos, I don't know if there is some adapter/proxy to install in order to make it work with Java GSSAPI

site/docs/security/sasl.md
```diff
@@ -0,0 +1,197 @@
...
+install a new server just for BookKeeper. Otherwise you will need to install one, your Linux vendor likely has packages
+for `Kerberos` and a short guide on how to install and configure it ([Ubuntu](https://help.ubuntu.com/community/Kerberos),
+[Redhat](https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/6/html/Managing_Smart_Cards/installing-kerberos.html)).
+Note that if you are using Oracle Java, you will need to download JCE policy files for your Java version and copy them to `$JAVA_HOME/jre/lib/security`.
```
eolivelli: You can run with basic JDK JCE files (no unlimited strenght....) but there is some setup on the KDC to be done, I don't know it is worth to say here

site/docs/security/sasl.md
```diff
@@ -0,0 +1,197 @@
...
+#### Kerberos Principals
+
+If you are using the organization’s Kerberos or Active Directory server, ask your Kerberos administrator for a principal
+for each Bookie in your cluster and for every operating system user that will access BookKeeper with Kerberos authentication
+(via clients and tools).
+
+If you have installed your own Kerberos, you will need to create these principals yourself using the following commands:
+
+
+    sudo /usr/sbin/kadmin.local -q 'addprinc -randkey bookkeeper/{hostname}@{REALM}'
```
eolivelli: this is another example doc
https://www.cloudera.com/documentation/enterprise/latest/topics/cm_sg_s6_user_principals.html

site/docs/security/sasl.md
```diff
@@ -0,0 +1,197 @@
...
+3. Make sure the keytabs configured in the `bookkeeper_jaas.conf` are readable by the operating system user who is starting bookkeeper client.
+
+4. Enable SASL authentication plugin in the client by setting following parameters.
+
+        clientAuthProviderFactoryClass=org.apache.bookkeeper.sasl.SASLClientProviderFactory
+
+
+## Enabling Logging for SASL
+
+To enable SASL debug output, you can set `sun.security.krb5.debug` system property to `true`.
```
eolivelli: Java GSSAPI errors are not clear, most of the times I need to enable debug to understand the cause.
in my cases most of the times the problems are related to reverse-lookup of IP addresses to hostname.
for Kerberos it is very important that DNS is setup coherently both for names and for PTRs

there is also a 'debug=true' option in Krb5LoginModule
```
Server {
  com.sun.security.auth.module.Krb5LoginModule required debug=true
  useKeyTab=true
  keyTab="xxxxx.keytab"
  storeKey=true
  useTicketCache=false
  principal="xxx/xxx@xxx";
};
```

site/docs/security/zookeeper.md
```diff
@@ -0,0 +1,41 @@
...
+---
+
+## New Clusters
+
+To enable `ZooKeeper` authentication on Bookies or Clients, there are two necessary steps:
+
+1. Create a `JAAS` login file and set the appropriate system property to point to it as described in [GSSAPI (Kerberos)](../sasl#notes).
+2. Set the configuration property `zkEnableSecurity` in each bookie to `true`.
+
+The metadata stored in `ZooKeeper` is such that only certain clients will be able to modify the corresponding znodes, but znodes are world readable.
```
eolivelli: I am sorry but we are using ZooDefs.Ids.CREATOR_ALL_ACL, this way only the authenticated and authorized clients will be able to read the contents of znode.

```
      /**
         * This is a completely open ACL .
         */
        public final ArrayList<ACL> OPEN_ACL_UNSAFE = new ArrayList<ACL>(
                Collections.singletonList(new ACL(Perms.ALL, ANYONE_ID_UNSAFE)));

        /**
         * This ACL gives the creators authentication id's all permissions.
         */
        public final ArrayList<ACL> CREATOR_ALL_ACL = new ArrayList<ACL>(
                Collections.singletonList(new ACL(Perms.ALL, AUTH_IDS)));
```

site/docs/security/sasl.md
sijie: technically we don't treat DIGEST-MD5 as a security feature. there were discussions before to remove `passwd` completely, because it doesn't really make any sense.

site/docs/security/sasl.md
sijie: installing kerberos is somehow out of the scope of this documentation, no? this is just giving an example. what are you suggesting here?

site/docs/security/sasl.md
sijie: I don't have real debug experience. I would defer this to a subsequent request if you are willing to contribute.

site/docs/security/zookeeper.md
sijie: will update

site/docs/security/tls.md
sijie: yes. the subsequent sentence explains `key` is for encryption and `certificate` for authentication. is that enough? can you guys suggest a better structure for explaining that?

site/docs/security/sasl.md
eolivelli: My bad, @sijie  forget this comment,
I found in other docs that ActiveDirectory is well supported.

site/docs/security/sasl.md
eolivelli: I think that if users will have trouble they will ask on the user@ mailing list.
It is better to leave the doc as simple as possible, expecially for third-party deps.

So I am OK with this part of the doc as you have written

site/docs/security/sasl.md
eolivelli: would it be good to cite this opportunity for tests/staging ?
I am fine with not writing about it as it is not "secure"

site/docs/security/tls.md
eolivelli: > The bookies and clients need their own key and certificate in order to use TLS

can be reworded as:

> The bookies need their own key and certificate in order to use TLS. Clients can optionally provide a key and a certificate for mutual authentication.

site/docs/security/sasl.md
sijie: I'd prefer not mentioning it at all, because it is not `secure` at all.

site/docs/security/tls.md
kishorekasi: TestTLS#testConnectToTLSClusterNonTLSClient - Tests whether a client, not configured to use TLS, can talk to bookie cluster with TLS enabled. I do not have a test to only encrypt but not authenticate. Not sure if it is even possible.

site/docs/security/tls.md
```diff
@@ -0,0 +1,178 @@
...
+4. `ca-password`: the passphrase of the CA
+5. `cert-file`: the exported, unsigned certificate of the bookie
+6. `cert-signed`: the signed certificate of the bookie
+
+(TBD: add a script to automatically generate truststores and keystores.)
+
+## Configuring Bookies
+
+Bookies support TLS for connections on the same service port. In order to enable TLS, you need to configure `tlsProvider` to be either
+`JDK` or `OpenSSL`. If `OpenSSL` is configured but the open ssl library is not found, it will automatically fall back to use default
```
kishorekasi: Yes, this is important. We will not be using OpenSSL .so library.

site/docs/security/tls.md
```diff
@@ -0,0 +1,178 @@
...
+    # trust store
+    tlsTrustStoreType=JKS
+    tlsTrustStore=/var/private/tls/bookie.truststore.jks
+    tlsTrustStorePasswordPath=/var/private/tls/bookie.truststore.passwd
+
+NOTE: it is important to restrict access to the store files and corresponding password files via filesystem permissions.
+
+Optional settings that are worth considering:
+
+1. tlsClientAuthentication=false: Enable/Disable using TLS for authentication.
```
kishorekasi: This config when enabled will authenticate the other end of the communication channel. It should be enabled on both bookie and client for MutualTLS.

site/docs/security/tls.md
```diff
@@ -0,0 +1,178 @@
...
+2. tlsEnabledCipherSuites= A cipher suite is a named combination of authentication, encryption, MAC and key exchange
```
kishorekasi: Provide a link to all the openssl and JDK cipher suites

site/docs/security/tls.md
```diff
@@ -0,0 +1,178 @@
...
+
+To verify the bookie's keystore and truststore are setup correctly you can run the following command:
+
+    openssl s_client -debug -connect localhost:3181 -tls1
+
+NOTE: TLSv1 should be listed under `tlsEnabledProtocols`.
+
+In the output of this command you should see the server's certificate:
+
+    (TBD)
```
kishorekasi: Add more details here?

site/docs/security/tls.md
```diff
@@ -0,0 +1,178 @@
...
+TLS is supported only for the new BookKeeper client (BookKeeper versions 4.5.0 and higher), the older clients are not
+supported. The configs for TLS will be the same as bookies.
+
+If client authentication is not required by the bookies, the following is a minimal configuration example:
+
+    tlsProvider=OpenSSL
+    clientTrustStore=/var/private/tls/client.truststore.jks
+    clientTrustStorePasswordPath=/var/private/tls/client.truststore.passwd
+
+If client authentication is required, then a keystore must be created for each client, and the bookies' truststores must
```
kishorekasi: provide the config option 'clientAuthentication' example here?

site/docs/security/sasl.md
eolivelli: @sijie ok, do not mention

site/docs/security/sasl.md
eolivelli: drop my comment, any KDC administrator will be able to do the better setup, it is out of the scope of BK docs to provide suggestions about KDC setup

site/docs/security/sasl.md
sijie: changed to use ```shell

site/docs/security/tls.md
sijie: done

site/docs/security/tls.md
sijie: done

site/docs/security/tls.md
sijie: done

site/docs/security/tls.md
sijie: done

site/docs/security/tls.md
sijie: done

site/docs/security/tls.md
sijie: done

site/docs/security/tls.md
sijie: done

site/docs/security/zookeeper.md
sijie: done

sijie:
addressed all comments. @merlimat @kishorekasi @eolivelli

PR #353, opened 2017-08-02
Issue 351: Apache rat check failed on master

zhaijack:

This fix on #352 was adding the exclusion rule in wrong location. this fix is to address that.


zhaijack:
@sijie can you review it again?

Issue #372, opened 2017-08-03
Allow to configure a specific address to advertise the bookie


merlimat:
Currently the bookies advertise itself, by default based on its own IP.

There are ways to control:
 * Which IP to use, by selecting a `listeningInterface`
 * Whether to use the machine hostname, by setting `useHostNameAsBookieID`

Though there is no way to force a particular address which the bookie should use to advertise itself.

This can be necessary in some deployment scenarios for different reasons: 
 * The Machine/VM has multiple IPs. One public and one private. The bookie should bind on the private IP but it should advertise the public IP to clients.
 * The bookie runs inside container (or pods) and the IP might change across restart, so we need a more "stable" bookie identifier (eg: the host machine IP or hostname).

For these reason, we should allow to control the `advertisedAddress` in the bookie config: 

```properties
advertisedAddress=10.0.0.2
...
```

PR #373, opened 2017-08-03
Issue #372: Allow to configure advertised address in bookies

merlimat:
Add option to configure advertised address for bookies. 

Reasons for this are outlined in #372

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
@@ -586,6 +586,10 @@ public class Bookie extends BookieCriticalThread {
         if (conf.getUseHostNameAsBookieID()) {
             hostAddress = inetAddr.getAddress().getCanonicalHostName();
         }
+
+        if (conf.getAdvertisedAddress() != null && conf.getAdvertisedAddress().trim().length() > 0) {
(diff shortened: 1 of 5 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -586,6 +586,10 @@ public static BookieSocketAddress getBookieAddress(ServerConfiguration conf)
         if (conf.getUseHostNameAsBookieID()) {
             hostAddress = inetAddr.getAddress().getCanonicalHostName();
         }
+
+        if (conf.getAdvertisedAddress() != null && conf.getAdvertisedAddress().trim().length() > 0) {
```
eolivelli: Maybe you can rearrage the code in order to skip completely the previous code, like dns queries...

sijie:
@eolivelli let's squeeze this in 4.5.0. I have same problem when I test 4.5 on google cloud.

eolivelli:
Yep @sijie it is what I meant, that for me is ok to have it in 4.5. We are on same page

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
merlimat: Good point, I thought it would make the logic complicated, but actually is better. I've fixed it.

Issue #262, opened 2017-07-19
Useless option "explicitLacInterval" is set to 1 in default bk_server.conf file


eolivelli:
The default bk_server.conf reports

```
# Interval between sending an explicit LAC in seconds
explicitLacInterval = 1
```

This option is only useful on the client side, it should be removed from this file

eolivelli:
@sijie  @zhaijack should be include this in 4.5 ?

sijie:
yes we should include this in 4.5. can you help with that?

eolivelli:
I will send a PR

PR #375, opened 2017-08-03
Issue-262 Useless option explicitLacInterval is set to 1 in default bk_server.conf

eolivelli:
Remove an useless line from default bk_server.conf file


eolivelli:
this is for issue #262

sijie:
yes we need to remove it from yaml file if it exists there.

Issue #377, opened 2017-08-03
Exception in double registering of Prometheus metric collectors


merlimat:
When the `PrometheusMetricsProvider` is enabled in the bookie, there can be cases in which the metrics are registered multiple times, leading to `IllegalArgumentException` errors that can cause the bookie process to restart. 

We should make the metric to be idempotent for the Prometheus provider, such that trying to get a bookie `StatsOpLogger` multiple times will always return the same instance instead of throwing exception.

PR #378, opened 2017-08-03
Issue #377: Make Prometheus stats logger registration idempotent

merlimat:
Make sure the metric are only registered once on Prometheus client lib.

Discusses in #377.


merlimat:
> It is a pity that we need reflection for such tasks. Maybe in future releases of PT client lib the problem will be addressed in a cleaner way.

Yes, it's a bit unfortunate. I'll try to get a PR on their repo.

PR #382, opened 2017-08-04
Issue 381: update old releases url to achives

sijie:

- the cms content has been copied from cms to asf-site branch under /content/archives. it is available under http://bookkeeper.apache.org/test/content/archives/
- update the release url from 'bookkeeper.apache.org/docs/rx.y.z' to '{{ site.baseurl }}achives/rx.y.z'

scripts:

- update the scripts to use clone and copy to avoid trash everything in remote. this would prevent overwrite `archives` directory.

content:

- fix syntax error in `shell.yaml` to be able to build website

### DIFF
--- a/site/_data/cli/shell.yaml
+++ b/site/_data/cli/shell.yaml
@@ -53,11 +53,11 @@ commands:
   description: List the bookies, which are running as either readwrite or readonly mode.
   options:
   - flag: -readwrite
-  description: Print readwrite bookies
+    description: Print readwrite bookies
   - flag: -readonly
-  description: Print readonly bookies
+    description: Print readonly bookies
   - flag: -hostnames
-  description: Also print hostname of the bookie
+    description: Also print hostname of the bookie
 - name: listfilesondisc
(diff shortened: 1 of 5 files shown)


site/_data/cli/shell.yaml
```diff
@@ -119,13 +119,12 @@ commands:
...
 - name: recover
   description: Recover the ledger data for failed bookie.
-  argument: [-deleteCookie] <bookieSrc> [bookieDest]
+  argument: <bookieSrc> [<bookieDest>]
   options:
   - flag: -deleteCookie
     description: Delete cookie node for the bookie.
 - name: simpletest
   description: Simple test to create a ledger and write entries to it.
-  argument: [-ensemble N] [-writeQuorum N] [-ackQuorum N] [-numEntries N]
```
merlimat: Is removing the argument intended or incidental? 

Is he problem related to the "[]"? You could try with the multi line syntax for the yank values

sijie: It is intended. Those are options listed in the options list below.

PR #396, opened 2017-08-05
Issue 390: [CI] Test errors in TestRackawareEnsemblePlacementPolicyUsingScript

sijie:

`Shell` class was ported from hadoop, but it contains a lot of unused code. Especially there is not `HADOOP_HOME` will set, so an exception will be thrown. (although the issue seems to be hidden in mac)

This change is to clean up the `Shell` class. The `Shell` class is only used for executing shell script configured in rackaware placement policy.


zhaijack:
tests passed, will shipit. 👍

PR #404, opened 2017-08-07
Fix zkCli issue

zhaijack:
/opt/zk/bin/zkCli.sh was leaked to replace by /opt/bookkeeper/bin/bookkeeper org.apache.zookeeper.ZooKeeperMain and cause:

/opt/bookkeeper/entrypoint.sh: line 61: /opt/zk/bin/zkCli.sh: No such file or directory


sijie:
nice catch

Issue #405, opened 2017-08-07
Docker image: fix non-privileged user execution


caiok:
**BUG REPORT**

1. Please describe the issue you observed:

- What did you do?

I run bookkeeper using "--user $(id -u)" option in "docker run" 

- What did you expect to see?

Bookeeper process owned by the current user in the host system.  

- What did you see instead?

Process is owned by root

caiok:
I've already fixed, but I can't assign this issue to me. I'll submit a PR asap.

/cc @jiazhai

caiok:
Created PR #406

sijie:
@caiok I assigned this to you. 

(I think only committers can assign the issues. I need to figure out to how to make contributors able to self-assign issues)

jiazhai:
Thanks a lot for the fix @caiok , have already merged it.

PR #406, opened 2017-08-07
ISSUE #405: Docker image: fix non-privileged user execution

caiok:
- Fix non-privileged user execution
- Change the Makefile in order to always use current user when run the bk image
- Fix Makefile in order to run "make run-demo"
- Adjust unneeded packages remove in Dockerfile


caiok:
In order to test this PR just do:
```
cd docker
make build
make run-demo
```
And verify that bk processes in the host system are not running as root

PR #413, opened 2017-08-07
Issue 412: css and js files are broken in `archives` links

sijie:

The content was copied from cms to `archives`. Those pages are using absolute links, so css and js locations are wrong. This change to fix all css, js, image urls to base on '/archives'.


sijie:
This is a change to `asf-site`.

PR #423, opened 2017-08-09
Issue 422: Don't use community layout for documentation pages

sijie:

The layout of documentation page is marked as community. when users navigate to this page, it is a bit difficult for the users to figure out where they are. We should stick documentation pages to doc layout.


sijie:
This is the layout how it looks like: https://sijie.github.io/bookkeeper-staging-site/docs/latest/index.html

PR #425, opened 2017-08-09
BOOKKEEPER-1105: RackAwarePolicy: Failure to map node into rack may result in failure to add other nodes.

dlg99:
- RackAwarePolicy's no longer uses /default-region if rack mapping fails unless required (by RegionAwarePolicy)
- it no longer fails to add rest of nodes after one node's failed addition,
- added unit tests
- added counters for successful/failed bookie adds/removal

UpdateLedgerOpTest failed but it seems to be known/unrelated issue.


bookkeeper-server/src/main/java/org/apache/bookkeeper/net/DNSToSwitchMapping.java
```diff
@@ -41,14 +41,14 @@
...
-     * should return {@link NetworkTopology#DEFAULT_RACK}. This
+     * should return {@link NetworkTopology#DEFAULT_REGION_AND_RACK}. This
      * is what the bundled implementations do, though it is not a formal requirement
      *
      * @param names the list of hosts to resolve (can be empty)
      * @return list of resolved network paths.
      * If <i>names</i> is empty, the returned list is also empty
      */
-    public List<String> resolve(List<String> names);
+    public List<String> resolve(List<String> names, String defaultRack);
```
sijie: This would break the binary compatibility for any people who implemented DNSToSwitchMapping.

The main problem is 

```java
    public static String resolveNetworkLocation(DNSToSwitchMapping dnsResolver, InetSocketAddress addr) {
        List<String> names = new ArrayList<String>(1);
        if (dnsResolver instanceof CachedDNSToSwitchMapping) {
            names.add(addr.getAddress().getHostAddress());
        } else {
            names.add(addr.getHostName());
        }
        // resolve network addresses
        List<String> rNames = dnsResolver.resolve(names);
        String netLoc;
        if (null == rNames) {
            logger.warn("Failed to resolve network location for {}, using default rack for them : {}.", names,
                NetworkTopology.DEFAULT_RACK);
            netLoc = NetworkTopology.DEFAULT_RACK;
        } else {
            netLoc = rNames.get(0);
        }
        return netLoc;
    }
```

I think a clean solution without breaking the binary compatibility is

- Fix the DefaultResolver to allow passing in a `defaultRack` 

```java
class DefaultResolver {

    final String defaultRack = ...;

   DefaultResolver() {
       this(...);
   }

   DefaultResolver(String defaultRack) {
       this.defaultRack = defaultRack;
   }

}
```

the rackware can use new DefaultResolver(`/default-rack`) and regionaware can use new DefaultResolver(`/default-region/default-rack`)

- create a DNSResolverDelegator with default-rack and remove the logic `NetUtils#resolveNetworkLocation` to this delegator.

```java
class DNSResolverDelegator extends DefaultResolver {

    DNSToSwitchMapping resolver;

     DNSResolverDelegator(DNSToSwitchMapping resolver, String defaultRack) {
          super(defaultRack);
          this.resolver= resolver;
     }

     List<String> resolve(List<String> names) {
        List<String> resolvedNames = resolver.resolve(names);
        if (null == resolvedNames) {
            // return default rack
        }
     }

}
```

- on initialization

```java
        if (optionalDnsResolver.isPresent()) {
            dnsResolver = optionalDnsResolver.get();
        } else {
            String dnsResolverName = conf.getString(REPP_DNS_RESOLVER_CLASS, ScriptBasedMapping.class.getName());
            try {
                dnsResolver = ReflectionUtils.newInstance(dnsResolverName, DNSToSwitchMapping.class);
                if (dnsResolver instanceof Configurable) {
                    ((Configurable) dnsResolver).setConf(conf);
                }
            } catch (RuntimeException re) {
                LOG.info("Failed to initialize DNS Resolver {}, used default subnet resolver.", dnsResolverName, re);
                dnsResolver = new DefaultResolver(`default-rack`); // <= add default rack
            }
        }
        this.dnsResolver = new DNSResolverDelegator(dnsResolver, defaultRack);
```

In this way you don't need to change the signature of the methods. How does that sound?

dlg99: makes sense, fixed.

dlg99:
@sijie tests that passed locally failed with TestTimedOutException, similarly to other pull requests earlier. Looks like a jenkins/build machine issue or unrelated build break.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
```diff
@@ -456,6 +572,7 @@ public BookieNode selectFromNetworkLocation(
             LOG.warn("Failed to choose a bookie from {} : "
                      + "excluded {}, fallback to choose bookie randomly from the cluster.",
                      networkLoc, excludeBookies);
+            LOG.warn("Predicate {}  Ensemble {}", predicate, ensemble);
```
sijie: This line doesn't give any useful information for debugging, isn't it?

bookkeeper-server/src/test/java/org/apache/bookkeeper/test/BookKeeperClusterTestCase.java
```diff
@@ -80,36 +80,46 @@
     protected int numBookies;
     protected BookKeeperTestClient bkc;
 
-    protected final ServerConfiguration baseConf = TestBKConfiguration.newServerConfiguration();
+    protected final ServerConfiguration baseConf;
```
sijie: This file seems to be changed for debugging purpose?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicy.java
```diff
@@ -22,6 +22,8 @@
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+
+import com.google.common.base.Preconditions;
```
sijie: unused import?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
```diff
@@ -136,9 +208,11 @@ protected RackawareEnsemblePlacementPolicyImpl initialize(DNSToSwitchMapping dns
                                                               int maxWeightMultiple,
                                                               StatsLogger statsLogger) {
         this.statsLogger = statsLogger;
+        this.bookiesJoinedCounter = statsLogger == null ? null : statsLogger.getOpStatsLogger(BookKeeperServerStats.BOOKIES_JOINED);
```
sijie: when the statsLogger would be null? we should use NullStatsLogger for the `null` case, rather than passing in a `null`.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
```diff
@@ -93,6 +103,62 @@ public void reloadCachedMappings() {
...
+        DNSResolverDecorator(DNSToSwitchMapping resolver, Supplier<String> defaultRackSupplier) {
+            Preconditions.checkNotNull(resolver, "Resolver cannot be null");
+            Preconditions.checkNotNull(defaultRackSupplier, "defaultRackSupplier should not be null");
+            this.defaultRackSupplier = defaultRackSupplier;
+            this.resolver= resolver;
+        }
+
+        public List<String> resolve(List<String> names) {
+            if (names == null) {
+                return new ArrayList<>();
```
sijie: use Collections.emptyList() to avoid constructing an object here.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
```diff
@@ -77,11 +73,25 @@
 
     static class DefaultResolver implements DNSToSwitchMapping {
 
+        final Supplier<String> defaultRackSupplier;
+
+        // for backwards compat
+        public DefaultResolver() {
+            this(()->NetworkTopology.DEFAULT_REGION_AND_RACK);
```
sijie: add space "() -> Network"

bookkeeper-server/src/test/java/org/apache/bookkeeper/test/BookKeeperClusterTestCase.java
dlg99: picked up by mistake, will revert.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
dlg99: Unit tests pass null as logger. will update unit tests, disallow null.

dlg99:
@sijie addressed last set of comments.

Issue #418, opened 2017-08-08
Bump Docker to use 4.5.0 binarary


jiazhai:
Since it is near to bk 4.5.0 release, would like to bump the docker image to use 4.5.0 binary.

in 4.4.0 binary, `bin/bookkeeper org.apache.zookeeper.ZooKeeperMain` will met issue 'Will not attempt to authenticate using SASL (unknown error)'.

eolivelli:
I think that that message is normal for zk clients without auth. It is somehow scary but it is not a problem

eolivelli:
I am ok with this change, it will be possible once we promote a rc to general availability

zhaijack:
@eolivelli. Thanks for the comments. 
In our use case it is a problem. I have verified the zk binary that contained in bookkeeper-4.4.0, with command "bin/bookkeeper org.apache.zookeeper.ZooKeeperMain", the wanted operation could not success, such as the command `ls /` and `create /new_root_dir`.

PR #436, opened 2017-08-11
Issue 418: bump docker to use bk - 4.5.0

zhaijack:
-  bump the docker image to use 4.5.0 binary, update pgp_key and bk version.
-  delete yum install for md5sum and sha1sum, since they are already in coreutils, which is default installed; and left them here will report error: 'No Match for argument: md5sum/sha1sum', while 'yum remove' in docker build.

have tested by running make run-dice.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -20,9 +20,9 @@
 FROM centos:7
 MAINTAINER Apache BookKeeper <dev@bookkeeper.apache.org>
 
-ARG BK_VERSION=4.4.0
+ARG BK_VERSION=4.5.0
 ARG DISTRO_NAME=bookkeeper-server-${BK_VERSION}-bin
-ARG GPG_KEY=B3D56514
+ARG GPG_KEY=FD74402C
 
 ENV BOOKIE_PORT=3181
 EXPOSE $BOOKIE_PORT
@@ -31,7 +31,7 @@ ENV BK_USER=bookkeeper
 # Download Apache Bookkeeper, untar and clean up
 RUN set -x \
     && adduser "${BK_USER}" \
-    && yum install -y java-1.8.0-openjdk-headless wget bash python md5sum sha1sum sudo \
+    && yum install -y java-1.8.0-openjdk-headless wget bash python sudo \
     && mkdir -pv /opt \
     && cd /opt \
     && wget -q "https://archive.apache.org/dist/bookkeeper/bookkeeper-${BK_VERSION}/${DISTRO_NAME}.tar.gz" \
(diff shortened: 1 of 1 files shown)


zhaijack:
@caiok @eolivelli Thanks for your review.
@caiok, Yes , it need cherry pick for 4.5.0, thanks for the reminder. I would like to do it after this merge.
Regarding 4.4.0,  one issue is that the ZooKeeperMain not work as expected (more info contained in issue #418). If support 4.4.0 branch, in docker file, we may need some more change, like  download latest zk, and replace ZooKeeperMain in entrypoint.sh with the downloaded zk, for better support.
That seems not simple cherry pick,  I am not sure if this will bring trouble to code management. 
@sijie , what do you think 4.4.0?

eolivelli:
I think that as 4.5 bookie is fully compatible with 4.4  clients we can support only 4.5  onwards

sijie:
since docker files are added for 4.5.0, I don't think we need to support previous releases. let's keep things simpler to just support 4.5.0 and onwards.

sijie:
FYI. I also cherry-picked the docker changes to branch 4.5. once this merge, the docker change would be in branch 4.5.

However, I'd like tho hold on retagging the 4.5.0 release. because, based on the comment in https://issues.apache.org/jira/browse/INFRA-14756, we *can't* mark the images generated by dockerhub autobuilds as official apache image based on apache policy. so I would like to wait for INFRA's response before we move on. If dockerhub builds is not the right approach to generate the official apache image, we have to take a different approach.

zhaijack:
Got it. Let's wait for the answer.

sijie:
@zhaijack @caiok @eolivelli  based on the response on https://issues.apache.org/jira/browse/INFRA-14756, Dockerhub is not the official channel for hosting 
docker images. that means any docker images we have to mark them as non-official.

as the community, we can still merge this pull request and publish the docker images. however unfortunately these images have to be marked as non-official, we have to state that in documentation if we are writing the docker related documentation.

sijie:
merging this now. 

- merge this to both master and branch-4.5
- git checkout a new branch from tag `release-4.5.0` (because it was created before docker files are added)
- cherry-pick the docker changes to that the temp branch, retag `release-4.5.0` to include the docker files

I will update the `release_guide` about it.

sijie:
cherry-picked all docker related to changes to both `branch-4.5` and `release-4.5.0` (retagged). 

4.5.0 image is built - https://hub.docker.com/r/apache/bookkeeper/tags/

@zhaijack @caiok can you guys verify this one?

zhaijack:
@sijie  Thanks a lot for the work. 
And I have verified 4.5.0 image. it worked well.

caiok:
@sijie Thanks for the work, image 4.5.0 works well.

@sijie @zhaijack Now we should make "latest" tag (the default tag and the most used one) to refer to 4.5.0, because it's still referring to 4.4.0.

zhaijack:
@caiok, latest should be 4.5.0. But seems the auto build failed for not able to connect to gpg server, please reference build details here: 
https://hub.docker.com/r/apache/bookkeeper/builds/bkvstyewexhym5ev94wce4n/
And @sijie has send out a mail which mentioned this.

Issue #430, opened 2017-08-10
Add an instruction in docker README to run standalone cluster


sijie:
**FEATURE REQUEST**

1. Please describe the feature you are requesting.

Currently there is no instruction in docker `README.md` to run standalone cluster. It would be good to write an instruction in `README.md` to run a standalone cluster.

2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have). Are you currently using any workarounds to address this issue?

*nice-to-have*

3. Provide any additional detail on your proposed use case for this feature.

N/A

PR #470, opened 2017-08-25
Issue #430: Add an instruction in docker README to run standalone cluster

ArvinDevel:
Add an instruction in docker README to run standalone cluster, and change the docker image description more accurately

> ---
> quickly and easily:
> 
> 
> ---

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -9,10 +9,28 @@ Apache Bookkeeper is a software project of the Apache Software Foundation, provi
 # How to use this image
 
 Bookkeeper needs [Zookeeper](https://zookeeper.apache.org/) in order to preserve its state and publish its bookies (Bookkeeper servers). The client only need to connect to a Zookeeper server in the ensamble in order to obtain the list of Bookkeeper servers.
+## standalone BookKeeper cluster
+Just like running a BookKeeper cluster in one machine(http://bookkeeper.apache.org/docs/latest/getting-started/run-locally/), you can run a standalone BookKeeper in one docker container, the command is:
+```
+docker run -it --env JAVA_HOME=/usr/lib/jvm/jre-1.8.0 \
+--entrypoint "/bin/bash" apache/bookkeeper:4.5.0 \
+-c "/opt/bookkeeper/bin/bookkeeper localbookie 3"
+```
+Note: you can first start the container, and then execute "bin/bookkeeper localbookie 3" in the container.  
+After that, you can execute BookKeeper shell command(http://bookkeeper.apache.org/docs/latest/reference/cli/) to test the cluster, you need first log into the container, use command below:
+```
+docker exec -it <container id or name> bash
+```
+then run test command, such as:
+```
+ ./bin/bookkeeper shell listbookies -rw
+ 
+ ./bin/bookkeeper shell simpletest
+ 
+```
+## TL;DR -- BookKeeper cluster 
 
-## TL;DR
-
-If you just want to see things working, you can play with Makefile hosted in this project and check its targets for a fairly complex set up example:
+If you want to setup cluster, you can play with Makefile hosted in this project and check its targets for a fairly complex set up example:
 ```
 git clone https://github.com/apache/bookkeeper
 cd bookkeeper/docker
@@ -48,7 +66,7 @@ And initialize the metadata store that bookies will use to store information:
 docker run -it --rm \
     --network "my-bookkeeper-network" \
     --env ZK_URL=my-zookeeper:2181 \
-    bookkeeper \
+    apache/bookkeeper:4.5.0 \
     bookkeeper shell metaformat
 ```
 Now we can start our Bookkeeper ensemble (e.g. with three bookies):
@@ -58,7 +76,7 @@ docker run -it\
     --env ZK_URL=my-zookeeper:2181 \
     --name "bookie1" \
     --hostname "bookie1" \
-    bookkeeper
+    apache/bookkeeper:4.5.0
 ```
 And so on for "bookie2" and "bookie3". We have now our fully functional ensemble, ready to accept clients.
 
@@ -91,7 +109,7 @@ $ docker run --name bookie1 -d \
     -e BK_bookiePort=3181 \                             < == use 1st approach, set bookiePort
     -e BK_zkServers=zk-server1:2181,zk-server2:2181 \   < == use 1st approach, set zookeeper servers
     -e BK_journalPreAllocSizeMB=32 \                    < == use 1st approach, set journalPreAllocSizeMB in [bk_server.conf](https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/conf/bk_server.conf)
-    bookkeeper
+    apache/bookkeeper:4.5.0
 ```
 
 ### Override rules for Bookkeeper configuration
--- a/docker/README.md
+++ b/docker/README.md
@@ -12,9 +12,11 @@ Bookkeeper needs [Zookeeper](https://zookeeper.apache.org/) in order to preserve
 ## standalone BookKeeper cluster
 Just like running a BookKeeper cluster in one machine(http://bookkeeper.apache.org/docs/latest/getting-started/run-locally/), you can run a standalone BookKeeper in one docker container, the command is:
 ```
-docker run -it --env JAVA_HOME=/usr/lib/jvm/jre-1.8.0 \
---entrypoint "/bin/bash" apache/bookkeeper:4.5.0 \
--c "/opt/bookkeeper/bin/bookkeeper localbookie 3"
+docker run -it \
+     --env JAVA_HOME=/usr/lib/jvm/jre-1.8.0 \
+     --entrypoint "/bin/bash" \
+     apache/bookkeeper \
+     -c "/opt/bookkeeper/bin/bookkeeper localbookie 3"
 ```
 Note: you can first start the container, and then execute "bin/bookkeeper localbookie 3" in the container.  
 After that, you can execute BookKeeper shell command(http://bookkeeper.apache.org/docs/latest/reference/cli/) to test the cluster, you need first log into the container, use command below:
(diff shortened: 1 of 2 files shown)


sijie:
@jiazhai @caiok @lucperkins can you guys review this pull request?

docker/README.md
```diff
@@ -9,10 +9,28 @@ Apache Bookkeeper is a software project of the Apache Software Foundation, provi
 # How to use this image
 
 Bookkeeper needs [Zookeeper](https://zookeeper.apache.org/) in order to preserve its state and publish its bookies (Bookkeeper servers). The client only need to connect to a Zookeeper server in the ensamble in order to obtain the list of Bookkeeper servers.
+## standalone BookKeeper cluster
+Just like running a BookKeeper cluster in one machine(http://bookkeeper.apache.org/docs/latest/getting-started/run-locally/), you can run a standalone BookKeeper in one docker container, the command is:
+```
+docker run -it --env JAVA_HOME=/usr/lib/jvm/jre-1.8.0 \
+--entrypoint "/bin/bash" apache/bookkeeper:4.5.0 \
```
jiazhai: To be align with other command, it may be better to put "apache/bookkeeper:4.5.0" at the end of this command.

docker/README.md
```diff
@@ -48,7 +66,7 @@ And initialize the metadata store that bookies will use to store information:
 docker run -it --rm \
     --network "my-bookkeeper-network" \
     --env ZK_URL=my-zookeeper:2181 \
-    bookkeeper \
+    apache/bookkeeper:4.5.0 \
     bookkeeper shell metaformat
```
jiazhai: Thanks for change this @ArvinDevel .
line 64--71 is not necessary for the new image. We could remove it, or change it to something like line-17: 
```
-c "/opt/bookkeeper/bin/bookkeeper shell metaformat -n -f"
```

docker/README.md
```diff
@@ -58,7 +76,7 @@ docker run -it\
     --env ZK_URL=my-zookeeper:2181 \
     --name "bookie1" \
     --hostname "bookie1" \
-    bookkeeper
+    apache/bookkeeper:4.5.0
```
caiok: Tag 4.5.0 is not needed. If you not provide it, docker will use "latest" (that we will always make it pointing to the latest released version, and it is probably the best choice for a new user). If we besides "hardcode" this tag in readme, we need to update it at every release.

docker/README.md
```diff
@@ -9,10 +9,28 @@ Apache Bookkeeper is a software project of the Apache Software Foundation, provi
 # How to use this image
 
 Bookkeeper needs [Zookeeper](https://zookeeper.apache.org/) in order to preserve its state and publish its bookies (Bookkeeper servers). The client only need to connect to a Zookeeper server in the ensamble in order to obtain the list of Bookkeeper servers.
+## standalone BookKeeper cluster
+Just like running a BookKeeper cluster in one machine(http://bookkeeper.apache.org/docs/latest/getting-started/run-locally/), you can run a standalone BookKeeper in one docker container, the command is:
+```
+docker run -it --env JAVA_HOME=/usr/lib/jvm/jre-1.8.0 \
+--entrypoint "/bin/bash" apache/bookkeeper:4.5.0 \
+-c "/opt/bookkeeper/bin/bookkeeper localbookie 3"
```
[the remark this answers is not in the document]
sijie: @caiok all the actions in the entrypoint.sh are done for a single bookie. this command is to start a standalone bookie cluster using one single image. all the operations in entrypoint.sh are not needed by this command. indeed the entrypoint.sh will not actually work for a standalone cluster. so I think @ArvinDevel 's approach is correct here. 

If you think there is another solution for starting a standalone cluster without bypassing entrypoint.sh, please suggest.

docker/README.md
ArvinDevel: ok, thanks, I will change it soon.

ArvinDevel:
During my test, I found somethings.
First, the JAVA_HOME  env is not set in current Docker image, so every time I enter the container to execute the bookkeeper command, I should set it firstly, I think we can set it in entrypoint.sh or somewhere else.
Second,When I use the Makefile to setup cluster in Mac, there exists some error:
If I execute `make run-format`, the info "/opt/bookkeeper/entrypoint.sh: line 51: sudo: command not found" showing that the current image don't have 'sudo' command;
if I execute `make run-bk` after execute `make run-zk`, the info "/opt/bookkeeper/entrypoint.sh: line 62: /opt/zk/bin/zkCli.sh: No such file or directory" showing the image lack 'zkCli.sh'; These two little matter doesn't make sense, we can improve it later.
if I restart the zk by run `make run-zk` again after shutdown it, then the `make run-bk` will have the info" ERROR - [main:Bookie@391] - Couldn't access cookie in zookeeper
org.apache.zookeeper.KeeperException$NodeExistsException: KeeperErrorCode = NodeExists for /ledgers/cookies/172.18.0.3:3181", I found this was due to the binding to local dir, so should we clear the local dir when we start zookeeper? 
Third, current Makefile pull  bookkeeper/bookie:centos docker image from store, and we use apache/bookkeeper in the ReadMe file, should we keep consistence?

jiazhai:
@ArvinDevel , Seems you are running a docker image that is out of date. Please build with latest code, or checkout the latest image.
https://github.com/apache/bookkeeper/blob/master/docker/scripts/entrypoint.sh#L24, this line export JAVA_HOME.  and line 62 neither contains the zkcli command.
https://github.com/apache/bookkeeper/blob/master/docker/Dockerfile#L34 , this line installed sudo.

Regarding the error of "Couldn't access cookie in zookeeper", in Makefile, there is a comments: 
"	@echo "If you want to restart from scratch the application, remove all its data:"
Maybe we could also add it in Readme.

And please help change the inconsistence in make file. Maybe change "VERSION ?= centos" into latest?

ArvinDevel:
@jiazhai I found the JAVA_HOME problem still exists, although I 
 use the latest version. https://github.com/apache/bookkeeper/blob/master/docker/scripts/entrypoint.sh#L24, I think this line export JAVA_HOME not accurately. @caiok can you give an explain to use the incomplete line?

docker/README.md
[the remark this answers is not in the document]
caiok: @sijie If you bypass the entrypoint you don't set the correct env variables (like JAVA_HOME) and don't modify the default package provided configuration (so -e BK_xxxx will not work and ledgers, indexes and logs will not be put in the expected /data/bookkeeper directory). You can't execute bk like a normal user either.

I read again entrypoint, I think we should modify it to not execute all zk related actions (wait for zk, create the root dir and metaformat) if the BK_zkServers env var is not provided. Sounds good to you?

caiok:
@ArvinDevel If you run the image with "--entrypoint xxx" the current entry point will be bypassed, so JAVA_HOME will not be set.

Steps I think you should do:
1) Modify entrypoint.sh in order to skip zk related actions (wait for zk, create zk root dir and metaformat) if BK_zkServers is empty
2) Run the image with this command:
```
docker run -it --rm \
    -v <local_dir_where_you_want_your_data_to_be_persisted>:/data/bookkeeper
    bookkeeper \
    /opt/bookkeeper/bin/bookkeeper localbookie 3
```

@sijie Arvin run into the same cookies issue I faced initially when creating bk docker image. I'm just curios, is there a better way to avoid this error without the need to wipe out data like we currently suggest to do? I don't remember if I had already found a solution at the time (e.g. specifying --hostname at "docker run" in order to avoid random hostname assignation)

zhaijack:
@ArvinDevel  The new changes bring in a lot of extraneous commits. try use "rebase" in git.

jiazhai:
@ArvinDevel The new changes bring in a lot of extraneous commits. try use "rebase" in git.

ArvinDevel:
@caiok Thank you first. I forgot describe the situation in last comment. When I run standalone cluster by bypassing the "entrypoint.sh", there‘s no lack of JAVA_HOME, because I pass it as by "--env JAVA_HOME=...". The situation where I can't find JAVA_HOME env in running bookkeeper container is when I use Makefile, "make run-bk", specifically when using original "entrypoint.sh", so I'm curious your implementation and I found exporting JAVA_HOME line is just "export JAVA_HOME=/usr"(https://github.com/apache/bookkeeper/blob/master/docker/scripts/entrypoint.sh#L24), I think this is incomplete, So I ask your help.

caiok:
@ArvinDevel Just for make sure, have you run Makefile targets after a "docker pull apache/bookkeeper"? This command force the pull of the latest version. I ask because I've tried makefile targets and everything went fine.
What value do you think it should be set in JAVA_HOME? What value have you used for make it working in standalone?

sijie:
@caiok @ArvinDevel : I think standalone mode is very different from the normal bookie container. so I am fine with skipping the default entry.sh, because there is nothing in current entry.sh needed for the current standalone mode. also, `standalone` mode is used for testing purpose. 

Also the current zk operation in entry point is to simplify running in different schedulers, e.g. mesos/marathon. I am fine with @ArvinDevel 's current approach for standalone.

ArvinDevel:
@caiok I'm sure I use the latest apache/bookkeeper, my test is using the step by step, first "make run-zk", then "make run-bk". 
I used "--env JAVA_HOME=/usr/lib/jvm/jre-1.8.0" to set JAVA_HOME when setup standalone cluster, and I think we should set  JAVA_HOME=/usr/lib/jvm/jre-1.8.0 in entrypoint.sh.

caiok:
@sijie Never played much with standalone mode, but I think it will take some configuration in input, maybe just the location for persisted data. How can you deal without the configuration management done in entrypoint.sh? I think security will not be a concern for this use case, but the user switch will not be performed neither.
 
Note: If we proceed in this way we need to specify in Readme  that everything written about image configuration will not apply in standalone mode.

sijie:
@caiok standalone mode runs a standalone zookeeper and N bookies all in same process. it doesn't take any configuration, and writes data to temp files.

The standalone mode and `normal` mode are different process. I don't see any point to consolidate them. What standalone mode needs here are the classes and libraries provided by this image.

caiok:
@sijie Very well, if no configuration is involved I'm ok to bypass entrypoint.sh.

@ArvinDevel I just re-tried to run "make run-demo" (which runs "make run-bk") and everything went fine, so I think the problem could be about something by your side. 
JAVA_HOME=/usr/lib/jvm/jre-1.8.0 works, but JAVA_HOME=/usr works either because the needed files are linked in /usr subdirs by "alternatives" (take a look at /usr/bin/java, for instance). The letter is in my opinion the more flexible choice for the future.

sijie:
@caiok I am not sure why `/usr` is better than `/usr/lib/jvm/jre-1.8.0`. isn't `/usr` is a too broader directory, because it also contains none-java binaries.

`/usr/lib/jvm/jre-1.8.0` is more precise, because you can tell what version of java is used. and also it is a docker image and if you don't change dependency, the location will never be changed. why do you think `/usr` is better?

caiok:
@sijie `/usr` will always be the same (so if we put --env JAVA_HOME=/usr in Readme we don't need to remember to change it when java version change). Java version it's better stated by running `java -version` in the container, but in the end it doesn't matter the which directory is set in JAVA_HOME, so I'm fine in using `/usr/lib/jvm/jre-1.8.0`

sijie:
@caiok - changing the java version is very rare, it only happens when you change the docker dependency. the docker readme can be changed when we change java version. the biggest concern I have using `/usr` for JAVA_HOME: it is not a pure java home directory. I feel it is very awkward to use `/usr` as JAVA_HOME. we should change it in the docker file.

sijie:
@ArvinDevel any idea why do you change the network name and the zookeeper name?

sijie:
@ArvinDevel I see. it makes sense.

@jiazhai @caiok can you guys review @ArvinDevel 's change again?

asfgit:
https://builds.apache.org/job/bookkeeper-precommit-pullrequest-docker/3/

Issue #458, opened 2017-08-22
Stats Provider jars not included in "server-bin" package 


zhaijack:
**FEATURE REQUEST**

1. Please describe the feature you are requesting.
Stats Provider jars not included in "server-bin" package, this will cause error for user to use the release binrary. It would be good to provide a "-all" package besides "server-bin" package.

2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have). Are you currently using any workarounds to address this issue?
should-have

3. Provide any additional detail on your proposed use case for this feature.

bk_server.conf contains the configuration of "statsProviderClass", and user could choose provider between PrometheusMetricsProvider, CodahaleMetricsProvider and Twitter finagle/ostrich/science Provider.  But if user set the provider in bk_server.conf,  run bookies it would cause error, becaue current bin package not contains the provider jars:
```
[root@bookkeeper-1 bookkeeper]# vi conf/bk_server.conf 
[root@bookkeeper-1 bookkeeper]# bookkeeper bookie
JMX enabled by default
2017-08-22 01:26:50,525 - INFO  - [main:BookieServer@301] - Using configuration file /opt/bookkeeper/conf/bk_server.conf
2017-08-22 01:26:50,535 - INFO  - [main:BookieServer@410] - Hello, I'm your bookie, listening on port 3181. ZKServers are on test_zookeeper:2181. Journals are in [/data/bookkeeper/journal]. Ledgers are stored in /data/bookkeeper/ledgers.
2017-08-22 01:26:50,538 - ERROR - [main:BookieServer@435] - Exception running bookie server : 
org.apache.commons.configuration.ConfigurationException: org.apache.commons.configuration.ConfigurationException: java.lang.ClassNotFoundException: org.apache.bookkeeper.stats.CodahaleMetricsProvider
	at org.apache.bookkeeper.util.ReflectionUtils.getClass(ReflectionUtils.java:102)
	at org.apache.bookkeeper.conf.ServerConfiguration.getStatsProviderClass(ServerConfiguration.java:1972)
	at org.apache.bookkeeper.proto.BookieServer.main(BookieServer.java:414)
Caused by: org.apache.commons.configuration.ConfigurationException: java.lang.ClassNotFoundException: org.apache.bookkeeper.stats.CodahaleMetricsProvider
	at org.apache.bookkeeper.util.ReflectionUtils.getClass(ReflectionUtils.java:63)
[...]

PR #500, opened 2017-09-06
Issue 499: make a package contains all the jar

zhaijack:
This is a sub-task for #458:
Make Prometheus provider the default in bk_server.conf,
Include stats providers in a -all package,

> ---
> quickly and easily:
> 
> - [X] Make sure the PR title is formatted like:
> - [X] Replace `<Issue # or BOOKKEEPER-#>` in the title with the actual Issue/JIRA number.
> 
> ---


asfgit:
https://builds.apache.org/job/bookkeeper-precommit-pullrequest-docker/10/

all/pom.xml
```diff
@@ -0,0 +1,121 @@
+<!--
```
sijie: let's rename `all` as `bookkeeper-all` or `bookkeeper-dist` (as what Leigh suggested in apache/distributedlog#175

all/pom.xml
```diff
@@ -0,0 +1,121 @@
...
+            <artifactId>bookkeeper-stats-api</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.bookkeeper.stats</groupId>
+            <artifactId>codahale-metrics-provider</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
```
sijie: let's add twitter-finagle-stats-provider as well.

all/pom.xml
```diff
@@ -0,0 +1,121 @@
...
+        </dependency>
+
+        <!-- bookkeeper.stats -->
+        <dependency>
+            <groupId>org.apache.bookkeeper.stats</groupId>
+            <artifactId>bookkeeper-stats-api</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+
+        <dependency>
```
sijie: it might be worth providing shaded jars for each stats providers, and use shaded jar here. in order to avoid any dependency conflicts between different stats provider implementation.

bookkeeper-server/src/main/resources/LICENSE.bin.txt
```diff
@@ -1,302 +0,0 @@
-
```
sijie: any reason why removing this one and NOTICE.bin.txt?

all/pom.xml
jiazhai: Thanks, will do it.

all/pom.xml
jiazhai: Thanks, will do it.

all/pom.xml
jiazhai: thanks. will do it.

all/pom.xml
zhaijack: Thanks. will change it..

all/pom.xml
zhaijack: Thanks. will change it.

all/pom.xml
zhaijack: Thanks. will add it.

bookkeeper-server/src/main/resources/LICENSE.bin.txt
zhaijack: This seems a dup to README and NOTICE under root dir. so use the ones under root dir.  If this is not right, will add them back.

bookkeeper-server/pom.xml
```diff
@@ -283,7 +283,7 @@
       </plugin>
       <plugin>
         <artifactId>maven-assembly-plugin</artifactId>
-        <version>2.2.1</version>
+        <version>${maven-assembly-plugin.version}</version>
```
eolivelli: @sijie  @zhaijack 
can we drop the assembly plugin from bookkeeper-server ? or it will be a further step ?

zhaijack: Thanks @eolivelli .  I thought it would be better to be similar to DL. in DL there are 2 binary provided, one is -all, while the other is like our bookkeeper-server-bin.  By this way, user will choose which is better for their use case.  
I am OK with both way: drop assemble for bookkeeper-server or not.  
@sijie what do you think of it?

eolivelli: @zhaijack it is OK for me

asfgit:
https://builds.apache.org/job/bookkeeper-precommit-pullrequest-docker/14/

asfgit:
https://builds.apache.org/job/bookkeeper-precommit-pullrequest-docker/17/

Issue #508, opened 2017-09-13
Introduce lifecycle components for managing components in bookie server


zhaijack:
**FEATURE REQUEST**

1. Please describe the feature you are requesting.

In a bookie server, we have multiple components to run, including:

- stats provider
- bookie server (both storage and netty server)
- autorecovery
- http endpoint

It becomes a bit messy on managing components run in a bookie. This issue attempts to introduce lifecycle component for each service component. so we can run these components in a clear way.

This is also help with #491 when we introduce a metadata rpc component. 

2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have). Are you currently using any workarounds to address this issue?

*should-have*

3. Provide any additional detail on your proposed use case for this feature.

N/A

PR #509, opened 2017-09-13
Issue 508: Introduce lifecycle components for managing components in bookie server 

zhaijack:

The motivation of creating lifecycle components is described in #508.

This changes include:

- create bookkeeper-common module, move some common classes (such as interface annotations) to common module.
- add the lifecycle components in bookkeeper-common module.
- add a package `org.apache.bookkeeper.server` for keeping all server related classes.
- rewrite BookieServer using the lifecycle components: stats provider | bookie server | auto recovery | http endpoint
- BookieServer is still kept for backward compatible, all the main functions are delegated to `org.apache.bookkeeper.server.Main`.


asfgit:
https://builds.apache.org/job/bookkeeper-precommit-pullrequest-docker/19/

eolivelli:
@zhaijack @sijie I have started looking into this. It is very interesting. Will come back soon with feedback

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/component/AbstractComponent.java
```diff
@@ -0,0 +1,44 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.bookkeeper.common.component;
+
+import org.apache.bookkeeper.common.conf.ComponentConfiguration;
+
+/**
+ * The base class represents a component in zstream.
```
eolivelli: zstream ? I assume this is a typo or cut and paste

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/component/ComponentStarter.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    /**
+     * Start a component and register a shutdown hook.
+     *
+     * @param component component to start.
+     */
+    public static void startComponent(LifecycleComponent component,
+                                      CountDownLatch aliveLatch) {
+        Runtime.getRuntime().addShutdownHook(new Thread(
```
eolivelli: this is not good.
On unit tests for instance we are going to pollute the Java Runtime with shutdownhooks.
Shutdown hooks introduce a lot of troubles and in my experience I saw a lot of deadlocks.
I would like to drop this if there is no strong evidence of benefit.

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/component/Lifecycle.java
```diff
@@ -0,0 +1,196 @@
...
+ *      stop();
+ *  }
+ *  if (!lifecycleState.moveToClosed()) {
+ *      return;
+ *  }
+ *  // perform close logic here
+ * }
+ * </pre>
+ */
+public class Lifecycle {
```
eolivelli: It seems to me that there is not concurrency control in this class, why ?

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/component/LifecycleListener.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+package org.apache.bookkeeper.common.component;
+
+/**
+ * Listener listening of the lifecycle changes.
+ */
+public interface LifecycleListener {
+
+    void beforeStart();
```
eolivelli: you can consider adding 'default' method implementations, it will make simpler to implement the interface and extend in the guture

bookkeeper-common/src/test/java/org/apache/bookkeeper/common/component/TestComponentStarter.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.concurrent.CountDownLatch;
+import org.apache.bookkeeper.common.component.ComponentStarter.ComponentShutdownHook;
+import org.junit.Test;
+
+/**
+ * Test Case of {@link ComponentStarter}.
+ */
+public class TestComponentStarter {
+
+  @Test(timeout = 10000)
```
eolivelli: I think we dropped "timeouts" in single tests

bookkeeper-server/pom.xml
```diff
@@ -30,16 +30,11 @@
...
   <dependencies>
     <dependency>
-      <groupId>org.apache.bookkeeper.stats</groupId>
-      <artifactId>bookkeeper-stats-api</artifactId>
+      <groupId>org.apache.bookkeeper</groupId>
+      <artifactId>bookkeeper-common</artifactId>
       <version>${project.parent.version}</version>
     </dependency>
     <dependency>
-      <groupId>com.google.guava</groupId>
```
eolivelli: We are shading and relocating guava, I have not checked the final output of the build, is the final result the same as in 4.5, that is that we do not have vanilla "guava" in the classpath ?

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/component/AbstractComponent.java
zhaijack: will fix this.

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/component/Lifecycle.java
zhaijack: this class usually used during statup and teardown. it is usually done by one thread. there is really need a lot of concurrency control attentions. and the caller can take care of this concurrency control in the place when it is used.

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/component/LifecycleListener.java
zhaijack: I don't have a strong preference here. if you feel strong about this, I can change.

bookkeeper-common/src/test/java/org/apache/bookkeeper/common/component/TestComponentStarter.java
zhaijack: will remove it.

bookkeeper-server/pom.xml
zhaijack: this is a good catch. shading is a big pain. I would defer this to a later task, because we might need to shade the classes that use guava.

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/component/ComponentStarter.java
eolivelli: Got it. Do you think we can drop it anyway?

eolivelli: Got it. Do you think we can drop it anyway?

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/component/Lifecycle.java
eolivelli: Ok

eolivelli: Ok

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/component/LifecycleListener.java
eolivelli: As we are introducing now the interface there is no real need. We will do it maybe in future.
Ok for me

bookkeeper-server/pom.xml
eolivelli: Ok  maybe we wil create a new issue. I will take care of think. I am particularly cautious with guava because sometimes at Guava they do not care about compatibility with previous releases

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/component/ComponentStarter.java
sijie: @eolivelli I am wondering why do you want to drop this, especially this is not a new hook introduced by this code. This shutdown hook is an attempt to do a clean shutdown, when the process was killed.

eolivelli: Before this change the hook applies only to the bookie server. Now we are adding an hook for every component.
There will be no control on when these hooks will run and in which order.
We could run into deadlocks if there is no coordination of components and for their own pluggable nature we won't have this ability of coordination.

Maybe this is ok for the BookieServer in order to try to do a clean shutdown of journals and storage. This usually can bring better recovery.

The hook is not guaranteed to be executed,  so functionally we can drop it.

For tests we should disable it anyway because it induces a strong reference to the component and this is a memory leak. We don't suffer about this now because we are always forking new processes.

Sorry for typos, using mobile

sijie: @eolivelli - 

1) this is just a util function to start a lifecycle component. You can use it or not use it. It is up to the caller.
2) the actual place calling this is a stack of lifecycle components, which is still exactly the bookie server. so there is no behavior changes
3) this function isn't actually called by other places except the main method starting the bookie server. so it doesn't run by any tests.

in this patch, we don't change any behavior of existing bookie server, how it starts and stops. we don't introduce new behavior and we don't run this for any components in tests. so I am not sure if we really want to drop it here.

eolivelli: @sijie thanks
My bad I did not see this is just an utility method and how it was used.
I apologize I did not have my laptop to have a global view.

For me is ok

PR #513, opened 2017-09-16
BOOKKEEPER-1106: Introduce write FileInfo cache and read FileInfo cache

yzang:
Problem: when read behind happens, it would quickly read bunch of ledgers, which will evict current active ledgers for writing from the ledger cache. with the ledger being evicted from cache, it would impact the write performance.

This feature is contributed by @sijie , in which we introduced write file info cache and read file info cache to cache the ledger index separately for read and write, so that when catch up read happens, it will not evict the file info for writes.

> ---
> quickly and easily:
> 
> 
> ---


yzang:
Pass all checks locally

asfgit:
https://builds.apache.org/job/bookkeeper-precommit-pullrequest-docker/23/

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/IndexPersistenceMgrTest.java
```diff
@@ -128,4 +130,138 @@ private void evictFileInfoTest(boolean createFile) throws Exception {
             indexPersistenceMgr.close();
         }
     }
+
+    final long lid = 1L;
+    final byte[] masterKey = "write".getBytes();
+
+    @Test(timeout = 60000)
```
eolivelli: Drop explicit timeout please. Recently we choose not to use it

eolivelli:
@yzang  the change would be good to go, if you can fix the @timeout annotation when can ship it

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/IndexPersistenceMgrTest.java
yzang: Sure, I'll fix the timeout

sijie:
@eolivelli @jiazhai if @yzang doesn't have time fixing the timeout annotation, we can have a separate issue for addressing that. let's move forward with this change.

eolivelli:
Ok will merge today

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/IndexPersistenceMgr.java
```diff
@@ -100,71 +118,160 @@ public IndexPersistenceMgr(int pageSize,
...
+        FileInfo fileInfo = notification.getValue();
+        Long ledgerId = notification.getKey();
+        if (null == fileInfo || null == notification.getKey()) {
+            return;
+        }
+        if (notification.wasEvicted()) {
+            evictedLedgersCounter.inc();
+            // we need to acquire the write lock in another thread,
+            // otherwise there could be dead lock happening.
+            evictionThreadPool.execute(() -> {
```
ivankelly: @yzang @sijie 
There's a serious bug here (which also causes the flake in IndexPersistenceMgrTest#testEvictFileInfoWhenUnderlyingFileExists).

When evicting from the cache, the FileInfo is removed from the cache, but only flushed asynchronously. If someone tries to read the FileInfo from the cache before the flush runs, they will read stale information, and then this will be written back when that FileInfo eventually gets evicted.

Do either of you recall what the deadlock mentioned in the comment is? LoadingCache notifications happen in the same thread as the get() so when handleLedgerEviction is run we should already have the FileInfoLock.readLock() held.

eolivelli: @ivankelly do your think this can be a blocker for 4.6 release ?
@jiazhai

sijie: @ivankelly nice catch here. @yzang has the fresh mind about the locking part. However I think the problem exists even before this change, the problem will occur when closing fileinfo happens after removing fileinfo from the fileinfo cache. another thread can load a new fileinfo in between and cause the issue. It wasn't noticed because of the way how we wrote the test case here - the test case was single threaded.

I don't think it is a blocker for `4.6`. It is a very corner case: when a file info was evicted with a cached `fenced` bit and a concurrent read request to load the fileinfo into the cache again. It only happens when a file info was evicted from both write and read fileinfo cache. in reality, it most likely won't happen. because the fileinfo cache is usually large and the read fileinfo cache is twice larger than write fileinfo cache. 

but it should go to `4.6.1` for sure, if necessary we need a fix for `4.5.2` as well.

ivankelly: ya, I don't think it's a blocker for 4.6.0, but for sure for 4.6.1.

The tragic part about this, is that I didn't catch it, CI did, multiple times, but because there were already flakes it wasn't seen as an big issue :/

Looking forward to hearing solutions around the locking.

sijie: @yzang do you mind taking a look at this and comment here?

yzang: @ivankelly Good questions. The deadlock is because eviction is running in the same thread who call getFileInfo. So when a thread call getFileInfo, it already acquires a read lock, and in order to guarantee no race condition, we have to acquire a write lock before we close the fileInfo, but the read lock is already acquired by the same thread, so we'll be stuck here because acquiring a write lock need to wait for read lock to be released. That's why we need to acquire the write lock asynchronously or in another thread to let itself release the read lock first.

As for the problem you mentioned, I don't think it will happen in normal cases. But it might happen in extreme case.
1. When FileInfo is evicted, it's already not in the cache, so whoever wants to read the file wouldn't get stale information
2. There's no impact if only read fileinfo is evicted or write fileinfo is evicted, because we only close the file when both of them are evicted and no one is using them by looking at the ref counting.
3. There's no async flush happening, we're just closing the file channel asynchronously when the ref counter for that FileInfo is 0.
4. If thread A is going to fileInfo, and thread B is getting the fileInfo, inside guava cache, this operation is done in a single thread, so if thread A happens first, then thread B will find nothing in the cache and create a new FileInfo, if thread B happens first, thread A will basically not evicting fileInfo anymore because the fileInfo is just accessed. 

(Extreme case is thread B access the FileInfo and before it has the chance to increase the ref counting, thread A evict the FileInfo from the cache and close the file)

ivankelly: So the deadlock is within the same thread? I was under the impression that if you hold a readlock on a thread, you can upgrade it to the writelock without releasing the readlock, but the documentation is telling that this isn't the case. But ok, this isn't the sort of deadlock which involves many interacting threads, so it should be simple enough to fix. This does need to be fixed though. Even if the sequence of events in unlikely, it breaks our guarantees, and once you have exceptions to guarantees, you have no guarantees at all.

ivankelly: @yzang @sijie

Ended up going down a rabbithole on this today and there's more ways for this to break.
When getFileInfo gives out a FileInfo object, it increases the refcount. If the cache evicts or expires before the caller of getFileInfo is finished, the flush can occur before modifications take place (such as setFenced), this setFenced can be lost, as the final release will not close the file handle. Going to put it down for today though, because it has my head fried.

I don't think guava cache is enough for this. We need something that is aware of reference counts, and gives us control of locking. It shouldn't be super hard, but will end up touching a lot of this code.

ivankelly: Also, #close() can be called when other places still have a reference, which triggers recycling of objects, which could break things spectacularly.

yzang: @ivankelly Hi, flush wouldn't occur before setFence take place. Because when you get out a FileInfo Object, you increment the counter, and when the cache evict that FileInfo it will check if the refcount == 0, if the setFence hasn't completed, refCount will be >=1, so the eviction wouldn't close and flush untilthe setFence operation is completed. So the setFence wouldn't be lost.

sijie: @ivankelly : I have the same feeling with @yzang. It can cause the case that I described in my first comment here, and trigger this flaky test. but it won't cause fence bit physically lost. I would suggest here if you think it will lost, a test case is good to demonstrate the sequence.

I don't think this problem is particularly related to guava cache and reference counting. this corner case exists even before guava cache change. It will happens if there is concurrent loading and eviction.  

regarding `FileInfo#close()`, it is my bad when introducing the watcher/watchable. it is a change after guava cache, so we can move the recycling logic to actual release.

ivankelly: @sijie @yzang there is an issue for sure.

The flaking one and the one I described today (see https://github.com/ivankelly/bookkeeper/commit/15609e571c5e7c166544ed2ca54f1d214518e7ff for demo).

Yes, they're unlikely, but we cannot base guarantees on "unlikely".

I have half a fix already, I'll should get it finished tomorrow morning.

sijie: @ivankelly I think we all agree with we need a fix.

yzang: @ivankelly Could you add an example of which test is flaky?

ivankelly: IndexPersistenceMgrTest.testEvictFileInfoWhenUnderlyingFileExists

I've already a fix in review, see #913

PR #516, opened 2017-09-18
Issue 515: Run bookies using hostIP instead of containerIP in k8s

sijie:

- use `status.hostIP` rather than `status.containerIP`. so the identifier (ip) of a bookie pod will not be changed after it is restarted.
- use `adverstiseAddress` to advertise hostip as the ip of a bookie pod
- change it to use `hierarchical` ledger manager


asfgit:
https://builds.apache.org/job/bookkeeper-precommit-pullrequest-docker/26/

PR #531, opened 2017-10-02
Issue-530 BP-14 BP-14 Relax Durability - protocol changes preview

eolivelli:
This is the first part of #471.
This patch includes changes to the v3 wire protocol and to metadata stored on zookkeeper

On metadata we are goign to introduce the LedgerType attribute for ledgers:
- PD_JOURNAL: Persistent Durability ledgers using the Journal
- VD_JOURNAL: Volatile Durability ledgers using the Journal

On the wire protocol we are going to change AddRequest and AddResponse to supporto VD_JOURNAL protocol changes to the LAC protocol and to introduce the brand new Sync protocol

More patches will come to implement client side changes and bookie side changes:
- LAC Protocol changes for VD_JOURNAL ledgers
- Introduction of the new 'sync' API
- Changes to the journal to support VD_JOURNAL ledgers and 'sync' API

**Please not that this patch is only a preview** and it needs the changes in #510 to support the creation of VD_JOURNAL ledgers.
It uses the org.apache.bookkeper.client.api package just to have the new LedgerType enum

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
@@ -42,7 +42,10 @@ import java.util.TreeMap;
 import static com.google.common.base.Charsets.UTF_8;
 
 import com.google.common.base.Optional;
+import com.google.common.base.Preconditions;
 import com.google.common.collect.Maps;
+import java.util.Collections;
+import org.apache.bookkeeper.client.api.LedgerType;
 
 /**
  * This class encapsulates all the ledger metadata that is persistently stored
@@ -74,6 +77,7 @@ public class LedgerMetadata {
     private long length;
     private long lastEntryId;
     private long ctime;
+    private LedgerType ledgerType;
 
     private LedgerMetadataFormat.State state;
     private SortedMap<Long, ArrayList<BookieSocketAddress>> ensembles =
@@ -85,14 +89,17 @@ public class LedgerMetadata {
     private LedgerMetadataFormat.DigestType digestType;
     private byte[] password;
 
-    private Map<String, byte[]> customMetadata = Maps.newHashMap();
+    private Map<String, byte[]> customMetadata;
 
     public LedgerMetadata(int ensembleSize, int writeQuorumSize, int ackQuorumSize,
-                          BookKeeper.DigestType digestType, byte[] password, Map<String, byte[]> customMetadata) {
+                          BookKeeper.DigestType digestType, byte[] password, Map<String, byte[]> customMetadata,
+                          LedgerType ledgerType) {
+        Preconditions.checkNotNull(ledgerType, "ledgerType is required");
         this.ensembleSize = ensembleSize;
         this.writeQuorumSize = writeQuorumSize;
         this.ackQuorumSize = ackQuorumSize;
         this.ctime = System.currentTimeMillis();
+        this.ledgerType = ledgerType;
 
         /*
          * It is set in PendingReadOp.readEntryComplete, and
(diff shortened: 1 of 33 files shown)


asfgit:
https://builds.apache.org/job/bookkeeper-precommit-pullrequest-docker/57/

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
```diff
@@ -85,14 +89,17 @@
     private LedgerMetadataFormat.DigestType digestType;
     private byte[] password;
 
-    private Map<String, byte[]> customMetadata = Maps.newHashMap();
+    private Map<String, byte[]> customMetadata;
 
     public LedgerMetadata(int ensembleSize, int writeQuorumSize, int ackQuorumSize,
```
sijie: it might be good to consider using builder pattern here. ledger type should have default value.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
```diff
@@ -422,6 +427,18 @@ public static LedgerMetadata parseConfig(byte[] bytes, Version version, Optional
...
+        if (data.hasLedgerType()) {
+            switch (data.getLedgerType()) {
+                case VD_JOURNAL:
+                   lc.ledgerType = LedgerType.VD_JOURNAL;
+                   break;
+                case PD_JOURNAL:
+                   lc.ledgerType = LedgerType.PD_JOURNAL;
+                   break;
+                default:
+                    throw new IllegalStateException("illegal ledgerType "+data.getLedgerType()+" is metadata");
```
sijie: please throw IOException instead of IllegalStateException. because this means the metadata is corrupted or unknown.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtocol.java
```diff
@@ -62,6 +62,16 @@
      */
     public static final int MASTER_KEY_LENGTH = 20;
 
+    /**
```
sijie: are you sure to support ledger type in v2 protocol? if you want to support it, I would suggest you first make sure v2 and v3 support same semantic first before introducing ledger type.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
```diff
@@ -576,6 +572,15 @@ void addEntry(final long ledgerId, byte[] masterKey, final long entryId, ByteBuf
                 addBuilder.setFlag(AddRequest.Flag.RECOVERY_ADD);
             }
 
+            switch (ledgerType) {
+                case VD_JOURNAL:
+                    addBuilder.setLedgerType(AddRequest.LedgerType.VD_JOURNAL);
+                    break;
+                case PD_JOURNAL:
+                    // nothing, this is the default
```
sijie: why? if you introduce this field, you'd better to set it.

bookkeeper-server/src/main/proto/BookkeeperProtocol.proto
```diff
@@ -111,6 +112,12 @@ message AddRequest {
     required int64 entryId = 2;
     required bytes masterKey = 3;
     required bytes body = 4;
+
+    enum LedgerType {
```
sijie: can you move LedgerType out of AddRequest?

bookkeeper-server/src/main/proto/BookkeeperProtocol.proto
```diff
@@ -136,6 +143,13 @@ message GetBookieInfoRequest {
     optional int64 requested = 1;
 }
 
+message SyncRequest {
+    required int64 ledgerId = 1;
+    required bytes masterKey = 2;
+    required int64 firstEntryId = 3;
```
sijie: why do you need first and last entry ids?

sync means flushing any outstanding requests and the response should tell you where does it sync to, no?

sijie:
@eolivelli is this still needed? I saw #642 and #643 are already sent out.

eolivelli:
Closing. Superseded by new set of patches

PR #536, opened 2017-10-04
Issue 535: Prometheus metrics provider fails to initialize

sijie:

- the parameter of `collectorNames` should be Collector rather than String
- add test case for prometheus metrics provider
- rename the classes to under `org.apache.bookkeeper.stats.prometheus` rather than taking the common `stats` package.


asfgit:
https://builds.apache.org/job/bookkeeper-precommit-pullrequest-docker/67/

PR #548, opened 2017-10-10
Issue 547: Allow load extra lifecycle components when running a bookie server.

sijie:

- make a common `ServerLifecycleComponet` for the components loaded by a bookie server
- add a flag to allow start extra components.
- add test cases

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/server/component/package-info.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/server/component/package-info.java
@@ -1,0 +1,21 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/**
+ * The lifecycle components that run on a bookie server.
+ */
+package org.apache.bookkeeper.server.component;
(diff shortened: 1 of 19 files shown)


asfgit:
https://builds.apache.org/job/bookkeeper-precommit-pullrequest-docker/94/

bookkeeper-server/src/main/java/org/apache/bookkeeper/server/component/package-info.java
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/**
+ * The lifecycle components that run on a bookie server.
+ */
+package org.apache.bookkeeper.server.component;
```
eolivelli: no new line ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/ReflectionUtils.java
```diff
@@ -36,6 +35,30 @@
...
+     * @param xface class interface
+     * @return the class object associated with the class or interface with the given string name.
+     */
+    public static <U> Class<? extends U> forName(String className,
+                                                 Class<U> xface) {
+
+        // Construct the class
+        Class<?> theCls;
+        try {
+            theCls = Class.forName(className);
```
eolivelli: I think is does not work very well, it is better to use Thread.currentThread().getContextClassloader()
I know that for bookie started from the command line this is not an issue, but if you embed the bookie in another process the class loader will be important.
I see that anyway the same approach is used in another method in this class.
So I am fine with this change, we can improve it later.

bookkeeper-server/src/main/java/org/apache/bookkeeper/server/component/package-info.java
sijie: these packages are under validation of current checkstyle rules. I don't think there is such one rule.

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/ReflectionUtils.java
sijie: I don't think we have ever used `Thread.currentThread().getContextClassLoader()`. I think that is typically useful in a multi-classloader environment. I am not sure the bookkeeper usage will ever get into this situation.

PR #556, opened 2017-10-12
BOOKKEEPER-1062: Fixed the race in AuditorLedgerCheckerTest

revans2:

This fixes the race condition in the auditor between read only bookies and available bookies.  But because I am not sure of the performance impacts to modifying getReadOnlyBookies, I have opted to add in a new getROBookies.  But I am happy to change it.  This pull request is mostly to show my work and see what others think.

### DIFF
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/replication/AuditorLedgerCheckerTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/replication/AuditorLedgerCheckerTest.java
@@ -40,6 +40,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicInteger;
 import org.apache.bookkeeper.client.AsyncCallback.AddCallback;
 import org.apache.bookkeeper.client.BKException;
@@ -168,6 +169,7 @@ public class AuditorLedgerCheckerTest extends BookKeeperClusterTestCase {
 
         // grace period for publishing the bk-ledger
         LOG.debug("Waiting for ledgers to be marked as under replicated");
+        waitForAuditToComplete();
         underReplicaLatch.await(5, TimeUnit.SECONDS);
         Map<Long, String> urLedgerData = getUrLedgerData(urLedgerList);
         assertEquals("Missed identifying under replicated ledgers", 1,
@@ -296,6 +298,7 @@ public class AuditorLedgerCheckerTest extends BookKeeperClusterTestCase {
 
         // grace period for publishing the bk-ledger
         LOG.debug("Waiting for Auditor to finish ledger check.");
+        waitForAuditToComplete();
         assertFalse("latch should not have completed", underReplicaLatch.await(5, TimeUnit.SECONDS));
     }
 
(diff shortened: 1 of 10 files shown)


asfgit:
https://builds.apache.org/job/bookkeeper-precommit-pullrequest-docker/104/

bookkeeper-server/src/test/java/org/apache/bookkeeper/replication/AuditorLedgerCheckerTest.java
```diff
@@ -767,6 +772,16 @@ public void testDelayedAuditWithRollingUpgrade() throws Exception {
         LOG.info("*****************Test Complete");
     }
 
+    private void waitForAuditToComplete() throws Exception {
```
eolivelli: If there is no Auditor or no AuditTask should we fail the test ?
Or shall we wait for an Auditor to be active ?

bookkeeper-server/src/test/java/org/apache/bookkeeper/replication/AuditorLedgerCheckerTest.java
```diff
@@ -313,19 +315,22 @@ public void testReadOnlyBookieShutdown() throws Exception {
...
-        assertFalse("latch should not have completed", underReplicaLatch.await(5, TimeUnit.SECONDS));
+        waitForAuditToComplete();
+        assertFalse("latch should not have completed", underReplicaLatch.await(1, TimeUnit.SECONDS));
 
         String shutdownBookie = shutdownBookie(bkIndex);
 
         // grace period for publishing the bk-ledger
         LOG.debug("Waiting for ledgers to be marked as under replicated");
+        waitForAuditToComplete();
         underReplicaLatch.await(5, TimeUnit.SECONDS);
```
eolivelli: Would it be better to add an "assertTrue" on the returned value of this "await" ?

revans2: For a future if it times out then it throws a TimeoutException.  So in this case we don't need the assert.

bookkeeper-server/src/test/java/org/apache/bookkeeper/replication/AuditorLedgerCheckerTest.java
revans2: Yes that is potentially a real concern.  I will add it in.  In practice it should not be an issue, because the auditor is started up but the setup of this test class, but we probably want to be sure of it.

revans2:
@eolivelli I have addressed your review comments.

asfgit:
https://builds.apache.org/job/bookkeeper-precommit-pullrequest-docker/107/

revans2:
Looks like there are some merge conflicts and one of the changes I made that worked fine in my local tests is now failing on jenkins.  I'll fix them and put up a fix shortly.

eolivelli:
@revans2 thank you. My bad I should have merged your fix before the other one

revans2:
I fixed the test and squashed my commits.

bookkeeper-server/src/test/java/org/apache/bookkeeper/replication/AuditorLedgerCheckerTest.java
revans2: Sorry I misunderstood the line number.  In this case we expect it to finish early, and we then check the results on the lines just below it.  This is mostly making sure that we wait long enough for the edit to be reflected properly.

asfgit:
https://builds.apache.org/job/bookkeeper-precommit-pullrequest-docker/108/

PR #642, opened 2017-10-18
BP-14 part 1 - metadata and protocol changes

eolivelli:
First part of Stacked Pull Requests Chain for implementation of BP-14

This commit includes:
- introduction of LedgerType in metadata and in protocol
- protobuf changes, for metadata and protocol
- new API for creating WriteHandle/WriteAdvHandle with a given LedgerType
- test cases for builders

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeper.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeper.java
@@ -50,6 +50,7 @@ import org.apache.bookkeeper.client.SyncCallbackUtils.SyncOpenCallback;
 import org.apache.bookkeeper.client.api.BookKeeperBuilder;
 import org.apache.bookkeeper.client.api.CreateBuilder;
 import org.apache.bookkeeper.client.api.DeleteBuilder;
+import org.apache.bookkeeper.client.api.LedgerType;
 import org.apache.bookkeeper.client.api.OpenBuilder;
 import org.apache.bookkeeper.conf.ClientConfiguration;
 import org.apache.bookkeeper.feature.Feature;
@@ -757,7 +758,7 @@ public class BookKeeper implements org.apache.bookkeeper.client.api.BookKeeper {
                 return;
             }
             new LedgerCreateOp(BookKeeper.this, ensSize, writeQuorumSize,
-                               ackQuorumSize, digestType, passwd, cb, ctx, customMetadata)
+                               ackQuorumSize, digestType, passwd, cb, ctx, customMetadata, LedgerType.PD_JOURNAL)
                 .initiate();
         } finally {
             closeLock.readLock().unlock();
@@ -960,7 +961,8 @@ public class BookKeeper implements org.apache.bookkeeper.client.api.BookKeeper {
                 return;
             }
             new LedgerCreateOp(BookKeeper.this, ensSize, writeQuorumSize,
-                               ackQuorumSize, digestType, passwd, cb, ctx, customMetadata).initiateAdv((long)(-1));
+                               ackQuorumSize, digestType, passwd, cb, ctx, customMetadata, LedgerType.PD_JOURNAL)
+                .initiateAdv((long)(-1));
         } finally {
             closeLock.readLock().unlock();
         }
@@ -1068,7 +1070,8 @@ public class BookKeeper implements org.apache.bookkeeper.client.api.BookKeeper {
                 return;
             }
             new LedgerCreateOp(BookKeeper.this, ensSize, writeQuorumSize,
-                               ackQuorumSize, digestType, passwd, cb, ctx, customMetadata).initiateAdv(ledgerId);
+                               ackQuorumSize, digestType, passwd, cb, ctx, customMetadata, LedgerType.PD_JOURNAL)
+                .initiateAdv(ledgerId);
         } finally {
             closeLock.readLock().unlock();
         }
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeper.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeper.java
@@ -63,6 +63,7 @@ import org.apache.bookkeeper.meta.LedgerManagerFactory;
 import org.apache.bookkeeper.net.BookieSocketAddress;
 import org.apache.bookkeeper.net.DNSToSwitchMapping;
 import org.apache.bookkeeper.proto.BookieClient;
+import org.apache.bookkeeper.proto.BookieProtocol;
 import org.apache.bookkeeper.proto.BookkeeperInternalCallbacks.GenericCallback;
 import org.apache.bookkeeper.stats.NullStatsLogger;
 import org.apache.bookkeeper.stats.StatsLogger;
@@ -962,7 +963,7 @@ public class BookKeeper implements org.apache.bookkeeper.client.api.BookKeeper {
             }
             new LedgerCreateOp(BookKeeper.this, ensSize, writeQuorumSize,
                                ackQuorumSize, digestType, passwd, cb, ctx, customMetadata, LedgerType.PD_JOURNAL)
-                .initiateAdv((long)(-1));
+                .initiateAdv(BookieProtocol.INVALID_LEDGER_ID);
         } finally {
             closeLock.readLock().unlock();
         }
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerCreateOp.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerCreateOp.java
@@ -36,6 +36,7 @@ import org.apache.bookkeeper.client.SyncCallbackUtils.SyncCreateAdvCallback;
 import org.apache.bookkeeper.client.SyncCallbackUtils.SyncCreateCallback;
 import org.apache.bookkeeper.client.api.CreateAdvBuilder;
 import org.apache.bookkeeper.client.api.CreateBuilder;
+import org.apache.bookkeeper.client.api.LedgerType;
 import org.apache.bookkeeper.client.api.WriteAdvHandle;
 import org.apache.bookkeeper.client.api.WriteHandle;
 import org.apache.bookkeeper.meta.LedgerIdGenerator;
@@ -62,6 +63,7 @@ class LedgerCreateOp implements GenericCallback<Void> {
     final byte[] passwd;
     final BookKeeper bk;
     final DigestType digestType;
+    final LedgerType ledgerType;
     final long startTime;
     final OpStatsLogger createOpLogger;
     boolean adv = false;
@@ -80,6 +82,8 @@ class LedgerCreateOp implements GenericCallback<Void> {
      *       ack quorum size
      * @param digestType
      *       digest type, either MAC or CRC32
+     * @param ledgerType
+     *       ledger type
      * @param passwd
      *       password
      * @param cb
@@ -91,13 +95,15 @@ class LedgerCreateOp implements GenericCallback<Void> {
      *       preserve the order(e.g. sortedMap) upon later retireval.
      */
     LedgerCreateOp(BookKeeper bk, int ensembleSize, int writeQuorumSize, int ackQuorumSize, DigestType digestType,
-            byte[] passwd, CreateCallback cb, Object ctx, final Map<String, byte[]> customMetadata) {
+            byte[] passwd, CreateCallback cb, Object ctx, final Map<String, byte[]> customMetadata, LedgerType ledgerType) {
         this.bk = bk;
-        this.metadata = new LedgerMetadata(ensembleSize, writeQuorumSize, ackQuorumSize, digestType, passwd, customMetadata);
+        this.metadata = new LedgerMetadata(ensembleSize, writeQuorumSize, ackQuorumSize, digestType, passwd,
+            customMetadata, ledgerType);
         this.digestType = digestType;
         this.passwd = passwd;
         this.cb = cb;
         this.ctx = ctx;
+        this.ledgerType = ledgerType;
         this.startTime = MathUtils.nowInNano();
         this.createOpLogger = bk.getCreateOpLogger();
     }
@@ -218,6 +224,8 @@ class LedgerCreateOp implements GenericCallback<Void> {
         private byte[] builderPassword;
         private org.apache.bookkeeper.client.api.DigestType builderDigestType
             = org.apache.bookkeeper.client.api.DigestType.CRC32;
+        private org.apache.bookkeeper.client.api.LedgerType builderLedgerType
+            = org.apache.bookkeeper.client.api.LedgerType.PD_JOURNAL;
         private Map<String, byte[]> builderCustomMetadata = Collections.emptyMap();
 
         CreateBuilderImpl(BookKeeper bk) {
@@ -261,11 +269,22 @@ class LedgerCreateOp implements GenericCallback<Void> {
         }
 
         @Override
+        public CreateBuilder withLedgerType(LedgerType ledgerType) {
+            this.builderLedgerType = ledgerType;
+            return this;
+        }
+
+        @Override
         public CreateAdvBuilder makeAdv() {
             return new CreateAdvBuilderImpl(this);
         }
 
         private boolean validate() {
+            if (builderLedgerType == null) {
+                LOG.error("invalid null builderLedgerType");
+                return false;
+            }
+
             if (builderWriteQuorumSize > builderEnsembleSize) {
                 LOG.error("invalid writeQuorumSize {} > ensembleSize {}", builderWriteQuorumSize, builderEnsembleSize);
                 return false;
@@ -314,7 +333,7 @@ class LedgerCreateOp implements GenericCallback<Void> {
             }
             LedgerCreateOp op = new LedgerCreateOp(bk, builderEnsembleSize,
                 builderWriteQuorumSize, builderAckQuorumSize, DigestType.fromApiDigestType(builderDigestType),
-                builderPassword, cb, null, builderCustomMetadata);
+                builderPassword, cb, null, builderCustomMetadata, builderLedgerType);
             ReentrantReadWriteLock closeLock = bk.getCloseLock();
             closeLock.readLock().lock();
             try {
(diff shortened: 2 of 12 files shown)


asfgit:
https://builds.apache.org/job/bookkeeper-precommit-pullrequest-docker/115/

bookkeeper-server/src/main/proto/BookkeeperProtocol.proto
```diff
@@ -73,6 +74,14 @@ message BKPacketHeader {
     required uint64 txnId = 3;
 }
 
+/**
+ * Type of Ledger
+ */
+enum LedgerType {
```
sijie: I saw `LedgerType` is defined in both `BookkeeperProtocol` and `DataFormats`. Can we define it in a common place? otherwise, it is going to be hard to maintain. 

one thought, if `LedgerType` is associated with metadata, it should exist in `DataFormats`; and `BookkeeperProtocol` references it.

bookkeeper-server/src/main/proto/BookkeeperProtocol.proto
```diff
@@ -136,6 +147,11 @@ message GetBookieInfoRequest {
     optional int64 requested = 1;
 }
 
+message SyncRequest {
+   required int64 ledgerId = 1;
+   required bytes masterKey = 2;
```
sijie: why do we need `masterKey` here?

bookkeeper-server/src/main/proto/BookkeeperProtocol.proto
```diff
@@ -193,3 +212,9 @@ message GetBookieInfoResponse {
 
 message StartTLSResponse {
 }
+
+message SyncResponse {
+    required StatusCode status = 1;
+    required int64 ledgerId = 2;
+    required int64 lastPersistedEntryId = 3;
```
sijie: If we are using `lastAddSynced`, why can't we standardize the name as `lastAddSynced`?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerCreateOp.java
```diff
@@ -218,6 +224,8 @@ private void createComplete(int rc, LedgerHandle lh) {
         private byte[] builderPassword;
         private org.apache.bookkeeper.client.api.DigestType builderDigestType
             = org.apache.bookkeeper.client.api.DigestType.CRC32;
+        private org.apache.bookkeeper.client.api.LedgerType builderLedgerType
```
sijie: why do you use the fully qualified class name here? don't you already import `LedgerType`?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeper.java
```diff
@@ -960,7 +961,8 @@ public void asyncCreateLedgerAdv(final int ensSize, final int writeQuorumSize, f
                 return;
             }
             new LedgerCreateOp(BookKeeper.this, ensSize, writeQuorumSize,
-                               ackQuorumSize, digestType, passwd, cb, ctx, customMetadata).initiateAdv((long)(-1));
+                               ackQuorumSize, digestType, passwd, cb, ctx, customMetadata, LedgerType.PD_JOURNAL)
+                .initiateAdv((long)(-1));
```
sijie: why did we use cast here? can we change it to `-1L`?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
```diff
@@ -84,12 +85,14 @@
 
     private boolean hasPassword = false;
     private LedgerMetadataFormat.DigestType digestType;
+    private LedgerMetadataFormat.LedgerType ledgerType;
     private byte[] password;
 
     private Map<String, byte[]> customMetadata = Maps.newHashMap();
 
     public LedgerMetadata(int ensembleSize, int writeQuorumSize, int ackQuorumSize,
```
sijie: Can you create an issue for using Builder pattern to construct ledger metadata? we should use builder pattern here.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeper.java
ivankelly: define an INVALID_LEDGER_ID somewhere. is -1 special, or would any negative work? Better to use something other than -1, because -1 is so common.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
ivankelly: may it could be done as part of making the metadata immutable. I had a patch for it before I left yahoo, but it was never merged. I could dig it out, or maybe just redo. It was only a day or two of work if i remember correctly.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerType.java
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * @since 4.6
+ */
+public enum LedgerType {
+    /**
+     * Persistent Durability, using Journal.<br>
+     * Each entry is persisted to the journal and every writes receives and acknowledgement only with the guarantee that
+     * it has been persisted durabily to it (data is fsync'd to the disk)
+     */
+    PD_JOURNAL,
```
ivankelly: These names aren't very descriptive. I think something like FSYNC_REQUIRED and FSYNC_DEFERRED would be clearer.

bookkeeper-server/src/main/proto/BookkeeperProtocol.proto
```diff
@@ -62,6 +62,7 @@ enum OperationType {
     READ_LAC = 7;
     GET_BOOKIE_INFO = 8;
     START_TLS = 9;
+    SYNC = 10;
```
ivankelly: FSYNC. SYNC could mean other things, like in zookeeper there's a sync op which has nothing to do with disks.

bookkeeper-server/src/main/proto/BookkeeperProtocol.proto
```diff
@@ -83,6 +92,7 @@ message Request {
     optional ReadLacRequest readLacRequest = 104;
     optional GetBookieInfoRequest getBookieInfoRequest = 105;
     optional StartTLSRequest startTLSRequest = 106;
+    optional SyncRequest syncRequest = 107;
```
ivankelly: FSyncJournalRequest

bookkeeper-server/src/main/proto/BookkeeperProtocol.proto
```diff
@@ -136,6 +147,11 @@ message GetBookieInfoRequest {
     optional int64 requested = 1;
 }
 
+message SyncRequest {
+   required int64 ledgerId = 1;
```
ivankelly: Why is the ledgerId included at all? Surely when you send it you expect all the journal up until this message to be fsynced.

bookkeeper-server/src/main/proto/BookkeeperProtocol.proto
```diff
@@ -193,3 +212,9 @@ message GetBookieInfoResponse {
 
 message StartTLSResponse {
 }
+
+message SyncResponse {
+    required StatusCode status = 1;
+    required int64 ledgerId = 2;
```
ivankelly: Again, do we need these? If the bookie has acknowledged an unsynced entry, it means when that bookie gets the sync message, the entry will be synced.

bookkeeper-server/src/main/proto/DataFormats.proto
```diff
@@ -58,6 +58,15 @@ message LedgerMetadataFormat {
         optional bytes value = 2;
     }
     repeated cMetadataMapEntry customMetadata = 11;
+
+    /**
+    * Ledger Type
+    */
+    enum LedgerType {
```
ivankelly: This is the third definition of the same enum.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeper.java
eolivelli: ok, but this is not really part of this change. I would like not to review all the usages of -1 and find which -1 are used for INVALID_LEDGER_ID
This line has been changed only in order to add LedgerType.PD_JOURNAL

I will change to -1L

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerCreateOp.java
eolivelli: yup, it was a cut and paste from the DigestType which needs the FQN

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
eolivelli: @ivankelly  thank you I will let you drive this additional change

bookkeeper-server/src/main/proto/BookkeeperProtocol.proto
eolivelli: Maybe the best option is to use an integer, it will be more simple to handle this in the future.
@sijie  @ivankelly  What do you think ?

bookkeeper-server/src/main/proto/BookkeeperProtocol.proto
eolivelli: you are right, it does not make sense.

bookkeeper-server/src/main/proto/BookkeeperProtocol.proto
eolivelli: will do

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerType.java
eolivelli: The more I am working on this the more I am convinced that these names are not fine.

We have to use the word 'journal" because @jvrao is working on a LedgerType without the Journal.
And Volatile is the opposite of Durability

Maybe
SYNC_REQUIRED_WITH_JOURNAL
SYNC_EXPLICIT_WITH_JOURNAL

@sijie @jvrao  ?

bookkeeper-server/src/main/proto/BookkeeperProtocol.proto
eolivelli: But is maps to the 'sync()' function on the API

bookkeeper-server/src/main/proto/BookkeeperProtocol.proto
eolivelli: This won't be used only for journalled bookies

 @jvrao

bookkeeper-server/src/main/proto/BookkeeperProtocol.proto
eolivelli: because this API means 'force and wait for a sync on data for the given ledger"
A bookie has several journals and we want to fsync data only on the journal assigned to the ledger (just for an example).

If you do not have "journal" with this primitive the clients want the bookie to acknowledge that data sent to him for the ledger has been persisted durably

bookkeeper-server/src/main/proto/DataFormats.proto
eolivelli: I would like to use an integer or to use the same type

bookkeeper-server/src/main/proto/BookkeeperProtocol.proto
eolivelli: I am changing this to a top level enum but I am hitting this bug on protoc
https://github.com/google/protobuf/issues/2054

asfgit:
https://builds.apache.org/job/bookkeeper-precommit-pullrequest-docker/119/

bookkeeper-server/src/main/proto/BookkeeperProtocol.proto
eolivelli: @ivankelly  Sorry, I can't understand your comment here

bookkeeper-server/src/main/proto/BookkeeperProtocol.proto
sijie: I am not sure why do you want to use integer. proto3 handles enum pretty well. proto3 is using forNumber, where valueOf is used?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerType.java
sijie: I am fine with REQUIRED and DEFERRED.

eolivelli: sorry, @sijie 

SYNC_REQUIRED_WITH_JOURNAL, SYNC_DEFERRED_WITH_JOURNAL
or
FSYNC_REQUIRED, FSYNC_DEFERRED ?

sijie: FSYNC_REQUIRED_WITH_JOURNAL, FSYNC_DEFERRED_WITH_JOURNAL ?

bookkeeper-server/src/main/proto/BookkeeperProtocol.proto
eolivelli: Anche for this @sijie?

eolivelli: And for this?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerType.java
ivankelly: FSYNC_REQUIRED_ON_JOURNAL, FSYNC_DEFERRED_ON_JOURNAL ?

bookkeeper-server/src/main/proto/BookkeeperProtocol.proto
eolivelli: Rollback. Thisi was needed, in order to access ledger impl on bookie side. See third part

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerType.java
eolivelli: Better 'with' as some ledger type won't have journal

ivankelly: Wouldn't these have a different type though, something like SKIP_JOURNAL?

bookkeeper-server/src/main/proto/BookkeeperProtocol.proto
ivankelly: sync() on which API? @eolivelli 
my issue is that sync is a very generic term, especially given that our system is a replicated log. it could mean may things.

eolivelli: This is an example:
```java
try (WriteHandle wh...) {
     wh.append(data);
     wh.append(data);
     wh.sync();
     wh.append(data);
}
```

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerType.java
sijie: @ivankelly this is per ledger, allowing a storage implementation bypassing journal.

bookkeeper-server/src/main/proto/BookkeeperProtocol.proto
sijie: SyncLedgerRequest or LSyncRequest? be more specific it is a per ledger operation?

bookkeeper-server/src/main/proto/BookkeeperProtocol.proto
sijie: @ivankelly it is a per ledger behavior. the bookie storage implementation can either use one journal, multiple journals or bypassing journal.

bookkeeper-server/src/main/proto/BookkeeperProtocol.proto
ivankelly: ForceRequest? force seems clearer to me. sync is suggesting something more akin to flushing. @sijie @eolivelli

bookkeeper-server/src/main/proto/BookkeeperProtocol.proto
ivankelly: Hmm, I think that sync() api should be renamed to force(). sync() sounds a little weak for this operation.

bookkeeper-server/src/main/proto/BookkeeperProtocol.proto
sijie: force sounds fine to me. it is a java convention

eolivelli: OK, I will rename Sync with Force, so ForceRequest, FORCE, WriteHandle#force()....

I will wait for some time before actually performing the rename, I am preparing other patches and this will be very disruptive

 @jvrao Are you OK with this new name 'force' instead of 'sync' ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerType.java
eolivelli: If we rename 'sync' to force, for the LedgerType @sijie  @ivankelly what is your idea ?
FSYNC_REQUIRED_ON_JOURNAL vs FORCE_REQUIRED_ON_JOURNAL vs AUTOMATIC_FORCE_ON_JOURNAL
and
FSYNC_DEFERRED_ON_JOURNAL vs FORCE_DEFERRED_ON_JOURNAL vs EXPLICIT_FORCE_ON_JOURNAL

and SKIP_JOURNAL (@jvrao)

ivankelly: FORCE_ON_JOURNAL (default)
FORCE_DEFERRED_ON_JOURNAL
SKIP_JOURNAL

eolivelli: @ivankelly  @sijie  @jvrao 
Proposal from Ivan sounds very good to me

FORCE_ON_JOURNAL (default)
FORCE_DEFERRED_ON_JOURNAL

SKIP_JOURNAL (this is a hint for the future, not part of BP-14)

sijie: sgtm

eolivelli:
closing this one, all comments addressed, new patch is  #670

PR #643, opened 2017-10-18
BP-14 part 2 - client side changes - OLD PATCH

eolivelli:
Second part of Stacked Pull Requests Chain for implementation of BP-14.
Please consider only the second commit.

Changes:
- new sync API
- client side changes to handle LAC for VD_JOURNAL ledgers
- additional feature on Mockito client side test cases

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/AsyncCallback.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/AsyncCallback.java
@@ -160,25 +160,6 @@ public interface AsyncCallback {
     }
 
     /**
-     * Async Callback for handling sync responses
-     *
-     * @since 4.0
-     */
-    @InterfaceAudience.Private
-    @InterfaceStability.Stable
-    interface SyncCallback {
-        /**
-         * Callback declaration
-         *
-         * @param rc
-         *          return code
-         * @param lastSyncedEntryId
-         *          id of the last entry which is known to have been persisted durably by a qquorum of bookies
-         */
-        void syncComplete(int rc, long lastSyncedEntryId);
-    }
-
-    /**
      * Async Callback for deleting ledgers.
      *
      * @since 4.0
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/AsyncCallback.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/AsyncCallback.java
@@ -160,6 +160,25 @@ public interface AsyncCallback {
     }
 
     /**
+     * Async Callback for handling sync responses
+     *
+     * @since 4.0
+     */
+    @InterfaceAudience.Private
+    @InterfaceStability.Stable
+    interface SyncCallback {
+        /**
+         * Callback declaration
+         *
+         * @param rc
+         *          return code
+         * @param lastSyncedEntryId
+         *          id of the last entry which is known to have been persisted durably by a qquorum of bookies
+         */
+        void syncComplete(int rc, long lastSyncedEntryId);
+    }
+
+    /**
      * Async Callback for deleting ledgers.
      *
      * @since 4.0
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperClientStats.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperClientStats.java
@@ -32,7 +32,6 @@ public interface BookKeeperClientStats {
     public final static String CREATE_OP = "LEDGER_CREATE";
     public final static String DELETE_OP = "LEDGER_DELETE";
     public final static String OPEN_OP = "LEDGER_OPEN";
-    public final static String SYNC_OP = "LEDGER_SYNC";
     public final static String RECOVER_OP = "LEDGER_RECOVER";
     public final static String LEDGER_RECOVER_READ_ENTRIES = "LEDGER_RECOVER_READ_ENTRIES";
     public final static String LEDGER_RECOVER_ADD_ENTRIES = "LEDGER_RECOVER_ADD_ENTRIES";
@@ -43,6 +42,7 @@ public interface BookKeeperClientStats {
     public final static String READ_OP = "READ_ENTRY";
     public final static String WRITE_LAC_OP = "WRITE_LAC";
     public final static String READ_LAC_OP = "READ_LAC";
+    public final static String SYNC_OP = "SYNC";
     public final static String READ_LAST_CONFIRMED_AND_ENTRY = "READ_LAST_CONFIRMED_AND_ENTRY";
     public final static String READ_LAST_CONFIRMED_AND_ENTRY_RESPONSE = "READ_LAST_CONFIRMED_AND_ENTRY_RESPONSE";
     public final static String PENDING_ADDS = "NUM_PENDING_ADD";
@@ -56,7 +56,7 @@ public interface BookKeeperClientStats {
 
     public final static String CHANNEL_READ_OP = "READ_ENTRY";
     public final static String CHANNEL_TIMEOUT_READ = "TIMEOUT_READ_ENTRY";
-    public final static String CHANNEL_SYNC = "CHANNEL_SYNC";
+    public final static String CHANNEL_SYNC = "SYNC";
     public final static String CHANNEL_ADD_OP = "ADD_ENTRY";
     public final static String CHANNEL_TIMEOUT_ADD = "TIMEOUT_ADD_ENTRY";
     public final static String CHANNEL_WRITE_LAC_OP = "WRITE_LAC";
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperClientStats.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperClientStats.java
@@ -32,6 +32,7 @@ public interface BookKeeperClientStats {
     public final static String CREATE_OP = "LEDGER_CREATE";
     public final static String DELETE_OP = "LEDGER_DELETE";
     public final static String OPEN_OP = "LEDGER_OPEN";
+    public final static String SYNC_OP = "LEDGER_SYNC";
     public final static String RECOVER_OP = "LEDGER_RECOVER";
     public final static String LEDGER_RECOVER_READ_ENTRIES = "LEDGER_RECOVER_READ_ENTRIES";
     public final static String LEDGER_RECOVER_ADD_ENTRIES = "LEDGER_RECOVER_ADD_ENTRIES";
@@ -55,6 +56,7 @@ public interface BookKeeperClientStats {
 
     public final static String CHANNEL_READ_OP = "READ_ENTRY";
     public final static String CHANNEL_TIMEOUT_READ = "TIMEOUT_READ_ENTRY";
+    public final static String CHANNEL_SYNC = "CHANNEL_SYNC";
     public final static String CHANNEL_ADD_OP = "ADD_ENTRY";
     public final static String CHANNEL_TIMEOUT_ADD = "TIMEOUT_ADD_ENTRY";
     public final static String CHANNEL_WRITE_LAC_OP = "WRITE_LAC";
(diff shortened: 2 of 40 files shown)


asfgit:
https://builds.apache.org/job/bookkeeper-precommit-pullrequest-docker/116/

eolivelli:
I forgot to include part of the needed changes, so most of the failed. I willbfix this pr before sending the next

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingSyncOp.java
```diff
@@ -0,0 +1,110 @@
...
+        this.cb = cb;
+        this.lastAddPushed = lh.getLastAddPushed();
+        ackSet = lh.distributionSchedule.getAckSet();
+        syncOpLogger = lh.bk.getSyncOpLogger();
+        this.writeSet = new HashSet<>(lh.distributionSchedule.getWriteSet(lastAddPushed));
+        this.receivedResponseSet = new HashSet<>(writeSet);
+    }
+
+    void sendSyncRequest(int bookieIndex) {
+        lh.bk.getBookieClient().sync(lh.metadata.currentEnsemble.get(bookieIndex), lh.ledgerId, lh.ledgerKey,
```
sijie: 1. it is a bit difficult for me the understand the implementation of sync without server side part. (as I suggest in the overall comment, I would prefer splitting this by features)

2. why do you need to carry `lastAddPushed` here? becase pending adds and sync calls can potentially go on different channels. how can you guarantee the entries added before `sync` already arrive at the bookies? I can't reason about it based the current client code.

as we discussed in the design doc, one simpler implementation can be --- this `sync` method can just simply issue `sync` method to all the bookies in the ensemble, and each bookies will tell its `lastAddSynced`, then the client will calculate the lastAddConfirmed based on individual `lastAddSynced`.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/SyncSupported.java
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * @see WriteHandle
+ * @see WriteAdvHandle
+ *
+ * @since 4.6
+ */
+public interface SyncSupported {
+
+    /**
+     * Waits for all data written by this Handle to have been persisted durably on a quorum of bookies and advances
```
sijie: this comment is confusing.

the behaviors of `sync` between `WriteHandle` and `WriteAdvHandle` are different. especially in `WriteAdvHandle`, you can't wait for all the data written to be persisted, because the entries might be added out of order.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieClient.java
```diff
@@ -224,15 +225,47 @@ private void completeAdd(final int rc,
             executor.submitOrdered(ledgerId, new SafeRunnable() {
                 @Override
                 public void safeRun() {
-                    cb.writeComplete(rc, ledgerId, entryId, addr, ctx);
+                    // TODO: lastAddSyncedEntry, see next commits
```
sijie: avoid `see next commits`

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
```diff
@@ -477,6 +478,30 @@ void connectIfNeededAndDoOp(GenericCallback<PerChannelBookieClient> op) {
...
+        final long txnId = getTxnId();
+        final CompletionKey completionKey = new V3CompletionKey(txnId,
+                                                                OperationType.SYNC);
+        completionObjects.put(completionKey,
+                              new SyncCompletion(completionKey, cb, ledgerId, ctx));
+
+        // Build the request
+        BKPacketHeader.Builder headerBuilder = BKPacketHeader.newBuilder()
+                .setVersion(ProtocolVersion.VERSION_THREE)
+                .setOperation(OperationType.WRITE_LAC)
```
sijie: operation is wrong

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
```diff
@@ -477,6 +478,30 @@ void connectIfNeededAndDoOp(GenericCallback<PerChannelBookieClient> op) {
 
     }
 
+    void sync(final long ledgerId, final byte[] masterKey, final long entryId,
```
sijie: entryId is not used.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RoundRobinDistributionSchedule.java
```diff
@@ -138,4 +116,64 @@ public QuorumCoverageSet getCoverageSet() {
     public boolean hasEntry(long entryId, int bookieIndex) {
         return getWriteSet(entryId).contains(bookieIndex);
     }
+
+    private class AckSetImpl implements AckSet {
```
sijie: My feeling is -- the `lastAddSyncedMap` should not be maintained per entry and the calculation of `LastAddSynced` should not be done per entry.

I think it should be maintained in `LedgerHandle` - each time a bookie advanced its `LastAddSynced`, it will recalculate a new `LastAddConfirmed` based on the overview of `LastAddSynced`, then it advances the pointer according to that.

eolivelli:
Thank you @sijie 
I will address your comments as soon as possible but as you are suggesting it is better to split the work using a different scheme.
I think I will keep this pr open in order to see yiur comments and open a new one

asfgit:
https://builds.apache.org/job/bookkeeper-precommit-pullrequest-docker/120/

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/AsyncCallback.java
```diff
@@ -160,6 +160,25 @@ void readComplete(int rc, LedgerHandle lh, Enumeration<LedgerEntry> seq,
     }
 
     /**
+     * Async Callback for handling sync responses
+     *
+     * @since 4.0
+     */
+    @InterfaceAudience.Private
```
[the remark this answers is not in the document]
eolivelli: it was because I did not want to mark this as "internal use only", anyway I have just dropped it

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/AsyncCallback.java
```diff
@@ -160,6 +160,25 @@ void readComplete(int rc, LedgerHandle lh, Enumeration<LedgerEntry> seq,
...
+    @InterfaceStability.Stable
+    interface SyncCallback {
```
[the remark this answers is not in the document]
eolivelli: dropped

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperClientStats.java
```diff
@@ -32,6 +32,7 @@
     public final static String CREATE_OP = "LEDGER_CREATE";
     public final static String DELETE_OP = "LEDGER_DELETE";
     public final static String OPEN_OP = "LEDGER_OPEN";
+    public final static String SYNC_OP = "LEDGER_SYNC";
```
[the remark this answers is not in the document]
eolivelli: done, renamed to "SYNC"

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperClientStats.java
```diff
@@ -55,13 +56,15 @@
 
     public final static String CHANNEL_READ_OP = "READ_ENTRY";
     public final static String CHANNEL_TIMEOUT_READ = "TIMEOUT_READ_ENTRY";
+    public final static String CHANNEL_SYNC = "CHANNEL_SYNC";
```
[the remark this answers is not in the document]
eolivelli: done, renamed to "SYNC", is is not a problem that is it the same name as SYNC_OP?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -84,15 +85,32 @@
...
+      * This value is used by readers, the the LAC protocol
+      */
     volatile long lastAddConfirmed;
 
+    /**
+      * Last entryId which has been confirmed to be written durably to the bookies,
+      * this is an internal variable used to track the status of entries and to handle correcly the lastAddConfirmed
+      * value
+      */
+     volatile long lastAddSynced;
```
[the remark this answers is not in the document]
eolivelli: you are right!  they contain the same value.
I used that variable in my first prototype.
Thank you very much @sijie

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -362,6 +402,14 @@ public synchronized boolean isClosed() {
         return metadata.isClosed();
     }
 
+    /**
+     * Is ensemble change supported ?
+     * @return
+     */
+    boolean isDelayEnsembleChangeSupported() {
```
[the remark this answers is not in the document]
eolivelli: this method is not even used. dropping it. Thank you @sijie . 

I don't know why I had this in code. I am sorry I would not like to add noise to this complex work

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RoundRobinDistributionSchedule.java
eolivelli: mmm, I think you suggestion is good.
I will try to move this logic into LedgerHandle and not per entry

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/SyncSupported.java
eolivelli: Yes,  WriteAdvHandle is tricky. You can leave gaps, then call sync() and then write entries in the gap.
In this case sync() can only guarantee that data sent to bookies "before calling sync" has been persisted.
Anyway the comment is right: "all data written by this Handle" are guaranteed to have been fsynched to the journal. And the returned value has no real meaning.
Maybe it is  better to have two separate methods, and for WriteAdvHandle return a CompletableFuture<Void>

asfgit:
https://builds.apache.org/job/bookkeeper-precommit-pullrequest-docker/121/

asfgit:
https://builds.apache.org/job/bookkeeper-precommit-pullrequest-docker/122/

asfgit:
https://builds.apache.org/job/bookkeeper-precommit-pullrequest-docker/123/

asfgit:
https://builds.apache.org/job/bookkeeper-precommit-pullrequest-docker/124/

eolivelli:
@sijie I am splitting the patch following your suggestions, some question:

> I think there are three main changes here:
>     1) write entries with ledger type and make sure client and server work with the ledger type bit first, without worrying about advancing lac or piggyback lastAddSynced
>     2) add sync method, make it work for both client and server.
>     3) add lastAddSync piggyback

in 1) do you expect to see that:
- LedgerType is now in protobuf files
- LedgerType is now in CreateBuilder
- LedgerType is in Metadata
- LedgerType is in AddRequest
- on the bookie side LedgerType gets to the Journal without effect
- on the client-side LastAddConfirmed will be computed as usual

in 2) do you expect to see that:
- introduction of client-side "force"
- on the client side the LastAddConfirmed  advances only on force() (for new ledgertype)
- the journal treats as expected the new ledgertype
(this will be the most important change)

in 3) do you expect to see that:
- in AddEntryResponse there is the LastAddSync piggy-backed
- LastAddConfirmed advances even without performing WriteHandle#force ?

I would add an additional step about splitting stats by LedgerType as JV is requiring, I think this can be done as last thing (maybe a separate commit too)

Maybe the "compat tests" could be a separate commit/step too, we have to prepare the story for comat tests against 4.5 or 4.4

Thank you

eolivelli:
@sijie  this is an example of patch 1 "introducing ledgertype":

see above comments
https://github.com/eolivelli/bookkeeper/commit/2a1b27b18f128d895beeb149d8fbdaa7d0480eeb

sijie:
@eolivelli the split looks good to me. +1

eolivelli:
Thank you @sijie I will send patches soon. I will close old pull requests as far as I am sure to have addresses all if the comments.

Other point
the behaviour of protoc on my laptop is different fron yours? My protobuf compiler generates code with warnings as I have already reported. The compiler is dowloaded from maven so we should have the same results

sijie:
please open an issue with the errors that you are seeing.

PR #645, opened 2017-10-18
Fix concurrent v2 reads on the same ledger/entry

ivankelly:
Original fix from by @merlimat in the yahoo-4.3 branch


ivankelly:
There is a later change with tests, but it depends on a bunch of other things in between.

asfgit:
https://builds.apache.org/job/bookkeeper-precommit-pullrequest-docker/118/

PR #646, opened 2017-10-20
Increase timeouts in TestDisableEnsembleChange

ivankelly:
In some narrow cases, if we manage to connect to the bookie, but the bookie
hasn't fully initialized, an add entry message will be dropped, and the
client will wait for the timeout to complete. The timeout is 5 seconds
by default, which was 5x the timeout in the test. This patch increases
the test timeout to 10s, which should give enough time for the addEntry
to fail and be retried.


asfgit:
https://builds.apache.org/job/bookkeeper-precommit-pullrequest-docker/125/

PR #657, opened 2017-10-24
DistributionSchedule uses custom wrapped int[] rather than HashSet

ivankelly:
This avoids the autoboxing on Integers and allocations of cells for
the hashset.

This also implies using wrapped int[] object for the write set in the
DistributionSchedules and PlacementPolicies.

This patch was originally submitted as dc7933b on the yahoo-4.3
branch, though this has been modified extensively to remove the
dependency on carrotsearch hppc and to allow it to work with placement
policies which didn't exist at the time of the original patch.


bookkeeper-server/pom.xml
```diff
@@ -178,6 +178,11 @@
       <artifactId>http-server</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>com.carrotsearch</groupId>
```
eolivelli: Do you have checked license for this?
Is it really needed? We already have several custom collections, maybe we could have our own implementation of IntArrayList without importing a whole third party lib. This will be used on client side.
It is used in EnsemblePlacement policy which is very often customized, at least in my projects.
I would not like to impose such dep on public plugins API

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LocalBookieEnsemblePlacementPolicy.java
```diff
@@ -88,14 +89,16 @@ public BookieSocketAddress replaceBookie(int ensembleSize, int writeQuorumSize,
     }
 
     @Override
-    public List<Integer> reorderReadSequence(ArrayList<BookieSocketAddress> ensemble, List<Integer> writeSet,
-                                             Map<BookieSocketAddress, Long> bookieFailureHistory) {
+    public IntArrayList reorderReadSequence(
```
eolivelli: We are breaking compatibility. We already did break compatibility for policy in 4.5....
Anyway if you have I would like to have a BK  class

asfgit:
https://builds.apache.org/job/bookkeeper-precommit-pullrequest-docker/142/

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LocalBookieEnsemblePlacementPolicy.java
sijie: I am fine with this, if it helps with performance - because the stability of this interface is `Evolving`. so it is okay to break the compatibility in minor release: https://github.com/apache/bookkeeper/blob/master/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/annotation/InterfaceStability.java#L41

https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/EnsemblePlacementPolicy.java#L195

bookkeeper-server/pom.xml
sijie: second to this comment here:

- for license, we need to update src/main/resources/NOTICE.bin.txt
- it is a new dependency introduced on client. It would prefer if we can avoid using this dependency.
- I am not sure what would be the benefits of using IntArrayList. If boxing, unboxing Integer and using an ArrayList is a problem, why can't we use `int[]` because we don't need resizing ability? we can wrap `int[]` into a recycle class, it should be simple and efficient, no?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
```diff
@@ -48,15 +50,15 @@
...
+class PendingAddOp implements WriteCallback, TimerTask, IntProcedure {
     private final static Logger LOG = LoggerFactory.getLogger(PendingAddOp.class);
 
     ByteBuf toSend;
     AddCallback cb;
     Object ctx;
     long entryId;
     int entryLength;
-    Set<Integer> writeSet;
+    IntHashSet writeSet;
```
sijie: Technically I think we don't need Set<Integer> here. I bet an `int[]` would work here as well.

bookkeeper-server/pom.xml
merlimat: > Do you have checked license for this? 
License is Apache v2.

> I am not sure what would be the benefits of using IntArrayList. If boxing, unboxing Integer and using an ArrayList is a problem, why can't we use int[] because we don't need resizing ability? we can wrap int[] into a recycle class, it should be simple and efficient, no? 

Yes, basically the class is a wrapper on top of an `int[]`. I guess the option here would be to either reimplement the "set" logic on the `int[]` or maybe import just the needed class. 

> We already have several custom collections, maybe we could have our own implementation of IntArrayList without importing a whole third party lib

Yes, the other collections were more about having no-boxing and concurrent collections. For non-concurrent primitive collections there are many available implementations (like the HPPC here).

ivankelly:
Looking at this again, bringing in an outside library is overkill. I'm going to try it with just a raw int[]. The placement policy stuff needs to be rewritten as well. It came in after the yahoo changes were made, and undoes a lot of the good work on the read path.

sijie:
@ivankelly what changes were undoed by the placement policy in the master? the `reorder` logic in placement policy isn't enabled by default, and reorder actually provides better latency in terms of aware of location distance. https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java#L97

ivankelly:
Ah, i didn't see it was optional. If enabled, it causes a bunch of allocations. Perhaps this can be ignored for now. This is more important for the write path, and that's allocated per entry, whereas on read it is per block.

ivankelly:
Actually, reading again, it is per entry on read too. Will think about what the best way to proceed is.

ivankelly:
@sijie @merlimat @eolivelli This is ready for review again. Using exclusively int[] now. Removed all allocations from the read reordering also.

asfgit:
https://builds.apache.org/job/bookkeeper-precommit-pullrequest-docker/148/

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicy.java
```diff
@@ -118,17 +118,21 @@ public BookieSocketAddress replaceBookie(
...
     @Override
-    public List<Integer> reorderReadLACSequence(ArrayList<BookieSocketAddress> ensemble,
-                                                List<Integer> writeSet,
-                                                Map<BookieSocketAddress, Long> bookieFailureHistory) {
-        return super.reorderReadLACSequence(ensemble, writeSet, bookieFailureHistory);
+    public int[] reorderReadLACSequence(
+            ArrayList<BookieSocketAddress> ensemble,
+            Map<BookieSocketAddress, Long> bookieFailureHistory,
+            int[] writeSet) {
+        return super.reorderReadLACSequence(ensemble, bookieFailureHistory,
```
eolivelli: Why do we need to call super? We can just not override the method

bookkeeper-server/src/test/java/org/apache/bookkeeper/util/collections/ArrayUtils2Test.java
```diff
@@ -0,0 +1,156 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+package org.apache.bookkeeper.util.collections;
+
+import java.util.Arrays;
```
eolivelli: Nit: reorder imports?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
```diff
@@ -92,7 +93,8 @@ PendingAddOp enableRecoveryAdd() {
 
     void setEntryId(long entryId) {
         this.entryId = entryId;
-        writeSet = new HashSet<Integer>(lh.distributionSchedule.getWriteSet(entryId));
```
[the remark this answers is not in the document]
ivankelly: Good point, updating

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicy.java
ivankelly: I left it as it always was.

bookkeeper-server/src/test/java/org/apache/bookkeeper/util/collections/ArrayUtils2Test.java
ivankelly: Updated, though I'm not sure what order you want (I don't use an IDE).

eolivelli: @ivankelly  for me is OK, I think that we have no clear rules about import order, maybe we should discuss on the list. Maybe checkstyle will impose some ordering

//cc @sijie

eolivelli:
I think the patch is ready. I will wait for the checks.

we need to wait for @sijie final review

ivankelly:
retest
previous seems stuck

bookkeeper-server/src/test/java/org/apache/bookkeeper/util/collections/ArrayUtils2Test.java
ivankelly: Ya, a checkstyle would be good. However turning it on would result in a bunch of errors straight away. Personally, I have little preference for imports, as long as they look neat.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RoundRobinDistributionSchedule.java
```diff
@@ -136,6 +221,7 @@ public QuorumCoverageSet getCoverageSet() {
 
     @Override
     public boolean hasEntry(long entryId, int bookieIndex) {
-        return getWriteSet(entryId).contains(bookieIndex);
+        int[] set = writeSets[(int) (entryId % ensembleSize)];
```
sijie: I think this can be further improved. because it is just checking whether an entry is supposed to be stored in one bookie. I think we can simply calculate it.

bookkeeper-server/src/test/java/org/apache/bookkeeper/util/collections/ArrayUtils2Test.java
sijie: there is already a checkstyle rule. it applies on the modules that have checkstyle enabled. the packages in bookkeeper-server are not enforced yet.

Import Static and then Import. And sort the imports alphabetically. try to avoid start import.

```
        <module name="ImportOrder">
            <property name="severity" value="error"/>
            <!-- This ensures that static imports go first. -->
            <property name="option" value="top"/>
            <property name="sortStaticImportsAlphabetically" value="true"/>
            <property name="tokens" value="STATIC_IMPORT, IMPORT"/>
            <message key="import.ordering"
                     value="Import {0} appears after other imports that it should precede"/>
        </module>

        <module name="AvoidStarImport">
            <property name="severity" value="error"/>
        </module>
```

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RegionAwareEnsemblePlacementPolicy.java
```diff
@@ -500,19 +504,26 @@ protected BookieNode replaceFromRack(BookieNode bookieNodeToReplace,
...
-            List<Integer> finalList = new ArrayList<Integer>(writeSet.size());
-            List<Integer> localList = new ArrayList<Integer>(writeSet.size());
-            List<Long> localFailures = new ArrayList<Long>(writeSet.size());
-            List<Integer> remoteList = new ArrayList<Integer>(writeSet.size());
-            List<Long> remoteFailures = new ArrayList<Long>(writeSet.size());
-            List<Integer> readOnlyList = new ArrayList<Integer>(writeSet.size());
-            List<Integer> unAvailableList = new ArrayList<Integer>(writeSet.size());
-            for (Integer idx : writeSet) {
+
+            int localMask      = 0x01 << 24;
```
sijie: all these can be constants?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RegionAwareEnsemblePlacementPolicy.java
```diff
@@ -500,19 +504,26 @@ protected BookieNode replaceFromRack(BookieNode bookieNodeToReplace,
     }
 
     @Override
-    public final List<Integer> reorderReadSequence(ArrayList<BookieSocketAddress> ensemble, List<Integer> writeSet, Map<BookieSocketAddress, Long> bookieFailureHistory) {
```
sijie: It is worth adding a BC test comparing the logic here with the logic in 4.5.0. they should be the same.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerChecker.java
```diff
@@ -243,10 +243,12 @@ public void operationComplete(int rc, Boolean result) {
                                                   }
                                               });
 
-                for (int bi : lh.getDistributionSchedule().getWriteSet(entryToRead)) {
-                    BookieSocketAddress addr = curEnsemble.get(bi);
-                    bookieClient.readEntry(addr, lh.getId(),
-                                           entryToRead, eecb, null);
+                final ArrayList<BookieSocketAddress> curEnsembleFinal = curEnsemble;
```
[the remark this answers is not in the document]
ivankelly: Not needed. Got pulled in as part of cherry-pick, will remove.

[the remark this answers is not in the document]
ivankelly: Pulled in as part of cherry-pick. I guess there was a callback here at some point which required the final that isn't there now. removed.

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/collections/ArrayUtils2.java
```diff
@@ -0,0 +1,84 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.bookkeeper.util.collections;
+
+import java.util.Random;
+import org.apache.commons.lang3.ArrayUtils;
+
+public class ArrayUtils2 {
```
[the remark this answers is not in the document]
eolivelli: should we also move this code to bookkeeper common module ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/collections/ArrayUtils2.java
```diff
@@ -0,0 +1,84 @@
...
+    static Random r = new Random();
+
+    public static int[] addMissingIndices(int[] array, int maxIndex) {
```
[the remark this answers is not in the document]
eolivelli: if we call this DistributionScheduleUtils maybe it is better to keep them in bookkeeper-server module

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RoundRobinDistributionSchedule.java
```diff
@@ -34,56 +39,136 @@
...
  */
 class RoundRobinDistributionSchedule implements DistributionSchedule {
-    private int writeQuorumSize;
-    private int ackQuorumSize;
-    private int ensembleSize;
+    private final int writeQuorumSize;
+    private final int ackQuorumSize;
+    private final int ensembleSize;
 
+    private final int[][] writeSets;
```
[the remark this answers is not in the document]
ivankelly: Wrapping is a good idea. Doing that. Will have a knock on effect to the rest of the comments.

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/collections/ArrayUtils2.java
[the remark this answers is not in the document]
ivankelly: nah, noone of these are generic enough. Moving into the WriteSet wrapper class anyhow.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RegionAwareEnsemblePlacementPolicy.java
ivankelly: The tests that are already there effectively test B.C. as they put a write set through a reorder and check if it matches a set value.

asfgit:
https://builds.apache.org/job/bookkeeper-precommit-pullrequest-docker/157/

ivankelly:
Not ready for review yet. Still some comments to address.

ivankelly:
@sijie @eolivelli ready for review again. There's a lot of changes for the changing to a wrapped write set.  Multiple patches, though only the first one is big.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
```diff
@@ -323,7 +327,12 @@ private synchronized int getNextReplicaIndexToReadFrom() {
         }
 
         private int getReplicaIndex(int bookieIndex) {
-            return writeSet.indexOf(bookieIndex);
+            for (int i = 0; i < writeSet.size(); i++) {
```
sijie: make this as a method of `WriteSet`?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
```diff
@@ -192,9 +197,14 @@ void initiate(ByteBuf toSend, int entryLength) {
         // Retain the buffer until all writes are complete
         this.toSend.retain();
         this.entryLength = entryLength;
-        for (int bookieIndex : writeSet) {
-            sendWriteRequest(bookieIndex);
+
+        // Iterate over set and trigger the sendWriteRequests
+        DistributionSchedule.WriteSet writeSet
+            = lh.distributionSchedule.getWriteSet(entryId);
+        for (int i = 0; i < writeSet.size(); i++) {
```
sijie:
```
try {
  ...
} finally {
   writeSet.recycle();
}
```

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
ivankelly: writeSet already has indexOf. Removing the whole method.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
ivankelly: Done

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/EnsemblePlacementPolicy.java
```diff
@@ -294,33 +294,39 @@ public BookieSocketAddress replaceBookie(int ensembleSize,
...
+     *          Write quorum to read entries. This will be modified, rather than
+     *          allocating a new WriteSet.
+     * @return The read sequence. This will be the same object as the passed in
+     *         writeSet.
      * @since 4.5
      */
-    public List<Integer> reorderReadLACSequence(ArrayList<BookieSocketAddress> ensemble,
-                                                List<Integer> writeSet,
-                                                Map<BookieSocketAddress, Long> bookieFailureHistory);
+    public DistributionSchedule.WriteSet reorderReadLACSequence(
```
eolivelli: we could use 'void' as return type:
we are changing the signature and we do not want client code to allocate WriteSet but we would like the given WriteSet to be modified.

Anyway I am fine with this version.
But I would like to know @ivankelly  and @sijie options

ivankelly: @eolivelli I think void would make it a little harder to use.
For example, in a lot of places we do
```
WriteSet reordered = policy.reorderReadSequnce(ensemble, failed, schedule.getWriteSet(eid))
```
If reorder returned void, then we would need to first assign the write set returned from getWriteSet to a variable and then pass it in.

eolivelli: @ivankelly  I see, thank you for your clarification.

PR #663, opened 2017-10-25
Issue 662: Introduce Bookie Registration Manager for bookie server

zhaijack:
While in BOOKKEEPER-628, It is trying to improve/generalize the bookie registration process.
This PR is for bookie side, it tries to introduce Bookie registration manager for bookie server.

> ---
> quickly and easily:
> 
> 
> ---


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
```diff
@@ -261,18 +264,15 @@ int runCmd(CommandLine cmdLine) throws Exception {
             boolean result = Bookie.format(conf, interactive, force);
             // delete cookie
             if (cmdLine.hasOption("d")) {
-                ZooKeeperClient zkc =
-                        ZooKeeperClient.newBuilder()
-                                .connectString(conf.getZkServers())
-                                .sessionTimeoutMs(conf.getZkTimeout())
-                                .build();
+                RegistrationManager rm = new ZKRegistrationManager();
```
eolivelli: what about implementing AutoCloseable and use try-with-resources ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
```diff
@@ -99,6 +99,8 @@
     // Kluge for compatibility testing. Never set this outside tests.
     public final static String LEDGER_MANAGER_FACTORY_DISABLE_CLASS_CHECK = "ledgerManagerFactoryDisableClassCheck";
 
+
```
eolivelli: useless spaces?

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
```diff
@@ -485,4 +487,6 @@ public AbstractConfiguration setTLSEnabledProtocols(
     public String getTLSEnabledProtocols() {
         return getString(TLS_ENABLED_PROTOCOLS, null);
     }
+
```
eolivelli: nit: usespaces spaces

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
sijie: @eolivelli sorry for jumping in here. 

I have this `AutoCloseable` question for a long time. why you are preferring `AutoCloseable` over `Closeable`? I think `Closeable` extends `AutoCloseable`. I saw we were changing a few places that were using `Closeable` to `AutoCloseable` (e.g. BookKeeper, LedgerHandle). https://docs.oracle.com/javase/8/docs/api/java/io/Closeable.html technically all the classes that implements Closeable should have AutoCloseable ability, no?

regardless, I think `RegistrationManager` is already implemented using `AutoCloseable` in this pull request. the reason might be to keep code structure unchanged, so it is easier for review.

eolivelli: @sijie Closeable is from package java.io. its close() method wants only IOException. We are not dealing with java.io qe are at a higher level.

My comment is mostly about using try-with-resources blocks, I think it is more elegant and is more error prone.
The change is already big, using ARM blocks will make code cleaner IMHO

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieInitializationTest.java
```diff
@@ -66,18 +69,29 @@
...
-        String ledgersPath = "/" + runtime.getMethodName();
+        String ledgersPath = "/" + "ledgers" + runtime.getMethodName();
         baseClientConf.setZkLedgersRootPath(ledgersPath);
         baseConf.setZkLedgersRootPath(ledgersPath);
     }
 
     @Override
     public void setUp() throws Exception {
         super.setUp();
+        LOG.info("setUp");
```
sijie: remove this?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
sijie: @eolivelli my question about 'AutoCloseable' is not about this change, but just in general.

regarding try-with-resources blocks, I would say for new codes - yes. but for the legacy codes, I would suggest not changing at this pull request. because it would involve a lot of unnecessary code structure changes.

zhaijack: @eolivelli @sijie, Thanks for the comments. Yes, current code helps avoid un-necessary code changes, and make it more close to original code. 
Since this is a general issue, How about we open an new issue for "try-with-resources" changes?

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
zhaijack: Thanks, will remove it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
zhaijack: Thanks. will remove it.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieInitializationTest.java
zhaijack: Thanks, will remove it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
sijie: +1 for a separate issue for `try-with-resources`

asfgit:
https://builds.apache.org/job/bookkeeper-precommit-pullrequest-docker/152/

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
eolivelli: @sijie  @zhaijack here is the issue https://github.com/apache/bookkeeper/issues/668
Sorry for going off-topic.

@sijie  for "AutoCloseable" in general I think that AutoCloseable is an higher level construct instead of java.io.Closeable which is only for I/O subsystems and is somehow legacy

I understand that we should leave this patch only for the abstraction of the RegistrationManager and I am +1

PR #664, opened 2017-10-26
Recycle PendingAddOps

ivankelly:
Avoid creating a new PendingAddOp object for each entry added, thus
saving on garbage.

Originally commit 55ba4723 on the yahoo-4.3 branch.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
@@ -17,7 +17,10 @@
  */
 package org.apache.bookkeeper.client;
 
+import static com.google.common.base.Preconditions.checkNotNull;
 import io.netty.buffer.ByteBuf;
+import io.netty.util.Recycler;
+import io.netty.util.Recycler.Handle;
 import io.netty.util.ReferenceCountUtil;
 import io.netty.util.Timeout;
 import io.netty.util.TimerTask;
@@ -48,9 +51,10 @@ import java.util.concurrent.RejectedExecutionException;
  *
  *
  */
-class PendingAddOp implements WriteCallback, TimerTask {
+class PendingAddOp extends SafeRunnable implements WriteCallback, TimerTask {
     private final static Logger LOG = LoggerFactory.getLogger(PendingAddOp.class);
 
+    ByteBuf payload;
     ByteBuf toSend;
     AddCallback cb;
     Object ctx;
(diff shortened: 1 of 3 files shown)


ivankelly:
One change that may not pop out is that I now add PendingAddOps to the ordered executor rather than letting them go to any random executor. This makes reasoning about the threading of the recycling much easier.

@merlimat

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
```diff
@@ -48,15 +51,18 @@
...
+    ByteBuf payload;
     ByteBuf toSend;
     AddCallback cb;
     Object ctx;
     long entryId;
     int entryLength;
-    int[] writeSet;
+
+    int writeQuorumSize;
+    int[] writeSet = new int[0];
```
eolivelli: do we need this allocation of "new int[0]"?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
```diff
@@ -183,31 +211,42 @@ void unsetSuccessAndSendWriteRequest(int bookieIndex) {
...
-        for (int i = 0; i < writeSet.length; i++) {
+        for (int i = 0; i < writeQuorumSize; i++) {
             sendWriteRequest(writeSet[i]);
         }
     }
 
     @Override
     public void writeComplete(int rc, long ledgerId, long entryId, BookieSocketAddress addr, Object ctx) {
         int bookieIndex = (Integer) ctx;
+        --pendingWriteRequests;
```
eolivelli: @ivankelly we can do such things without "synchronized" because we are now using OrderedSafeExecution in a way that there is no concurrent operation ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
```diff
@@ -343,4 +384,55 @@ public boolean equals(Object o) {
...
+        //   sent.
+        // The object can only be recycled after all references are
+        // released, otherwise we could end up recycling twice and all
+        // joy that goes along with that.
+        if (hasRun && callbackTriggered && pendingWriteRequests == 0) {
+            recycle();
+        }
+    }
+
+    private void recycle() {
```
eolivelli: don't we reset writeSet ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
ivankelly: it avoids the need for a null check further down, and it only happens when the op is first created, so there's negligible cost for it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
ivankelly: I wouldn't remove the synchronized stuff yet, but at some point we should go through the ops, figure out exactly what thread stuff is running on, and put in assert statements to make sure they are running on those threads. At that point we can remove the synchronized blocks.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
ivankelly: It gets reset in create()

ivankelly: I can add it though for completeness.

asfgit:
https://builds.apache.org/job/bookkeeper-precommit-pullrequest-docker/150/

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
sijie: can you define a constant for `new int[0]`?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RoundRobinDistributionSchedule.java
```diff
@@ -62,10 +62,11 @@ public RoundRobinDistributionSchedule(int writeQuorumSize, int ackQuorumSize, in
 
     @Override
     public void getWriteSet(long entryId, int[] target) {
-        if (target.length != writeQuorumSize) {
```
sijie: any reason why you can this behavior? isn' t `!=` a stricter validation?

ivankelly: Because as PendingAddOps get reused, a PendingAddOp with a writeSet of size 5 may get used with a ledger that needs a writeSet of size 3.

sijie:
@ivankelly there are a few findbugs errors. can you address that?

ivankelly:
Fixing. I had fixed this before, must have lost the commit in the rebase.

PR #670, opened 2017-10-27
BP-14 Relax Durability - patch 1 Introducing LedgerType

eolivelli:
This patch introduces LedgerType.
We have two LedgerTypes:
- FORCE_ON_JOURNAL: this is the default LedgerType, BookKeeper works as before
- FORCE_DEFERRED_ON_JOURNAL, this is a new LedgerType

in next patches for FORCE_DEFERRED_ON_JOURNAL ledgers each write will receive acknowledge of success as soon as data has been written to disk and not "fsynced".

This patch only introduces LedgerType on metadata and on AddRequest protocol. 
Test cases will be added in the next patch.

LedgerType FORCE_DEFERRED_ON_JOURNAL is not available for v2 protocol clients.


asfgit:
https://builds.apache.org/job/bookkeeper-precommit-pullrequest-docker/155/

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerCreateOp.java
```diff
@@ -261,15 +269,36 @@ public CreateBuilder withDigestType(org.apache.bookkeeper.client.api.DigestType
...
+            if (bk.getConf().getUseV2WireProtocol() && builderLedgerType != LedgerType.FORCE_ON_JOURNAL) {
+                LOG.error("invalid builderLedgerType {} while using v2 wire protocol");
+                return false;
+            }
+
             if (builderWriteQuorumSize > builderEnsembleSize) {
                 LOG.error("invalid writeQuorumSize {} > ensembleSize {}", builderWriteQuorumSize, builderEnsembleSize);
                 return false;
             }
+            if (builderLedgerType.equals(LedgerType.FORCE_DEFERRED_ON_JOURNAL) && builderWriteQuorumSize != builderEnsembleSize) {
```
sijie: line 288 is using `!=` and line 297 is using `equals`. can we at least using same pattern?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerFragmentReplicator.java
```diff
@@ -317,7 +318,7 @@ public void writeComplete(int rc, long ledgerId,
                                 ledgerFragmentEntryMcb.processResult(rc, null,
                                         null);
                             }
-                        }, null, BookieProtocol.FLAG_RECOVERY_ADD);
+                        }, null, BookieProtocol.FLAG_RECOVERY_ADD, LedgerType.FORCE_ON_JOURNAL);
```
sijie: you need to use the type from ledger metadata, no?

bookkeeper-benchmark/src/main/java/org/apache/bookkeeper/benchmark/BenchBookie.java
```diff
@@ -175,7 +176,7 @@ public static void main(String[] args)
             toSend.writeLong(entry);
             toSend.writerIndex(toSend.capacity());
             bc.addEntry(new BookieSocketAddress(addr, port), ledger, new byte[20],
-                        entry, toSend, tc, null, BookieProtocol.FLAG_NONE);
+                        entry, toSend, tc, null, BookieProtocol.FLAG_NONE, LedgerType.FORCE_ON_JOURNAL);
```
sijie: line 179, 197, 217 need to use the ledger type bit in metadata, no?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerType.java
```diff
@@ -0,0 +1,46 @@
...
+ * LedgerTypes describes the behaviour of the ledger in respect to durability and provides
+ * hints to the storage of data on Bookies
+ *
+ * @since 4.6
+ */
+public enum LedgerType {
+    /**
+     * Persistent Durability, using Journal.<br>
+     * Each entry is persisted to the journal and every writes receives and acknowledgement only with the guarantee that
+     * it has been persisted durabily to it (data is fsync'd to the disk)
```
sijie: typo? durably

bookkeeper-benchmark/src/main/java/org/apache/bookkeeper/benchmark/BenchBookie.java
eolivelli: I think this is a tricky point. The replicator needs to have guarantee that entry is really persisted.
This was a point were I thought that we needed a per entry flag and not a ledger wise flag.
Do you think that replicator should not wait for fsync?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerCreateOp.java
eolivelli: I will use equals everywhere. Fixing

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerFragmentReplicator.java
eolivelli: I think this is a tricky point. The replicator needs to have guarantee that entry is really persisted.
This was a point were I thought that we needed a per entry flag and not a ledger wise flag.
Do you think that replicator should not wait for fsync?

bookkeeper-benchmark/src/main/java/org/apache/bookkeeper/benchmark/BenchBookie.java
eolivelli: Sorry bad line for comment. Here we should use but from metadata. Thanks

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerType.java
eolivelli: Will fix

bookkeeper-benchmark/src/main/java/org/apache/bookkeeper/benchmark/BenchBookie.java
jvrao: so what is the decision? Are we using bit from metadata or not?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerType.java
```diff
@@ -0,0 +1,46 @@
...
+     */
+    FORCE_ON_JOURNAL,
+    /**
+     * Volatile Durability, using Journal.<br>
+     * Each entry is persisted to the journal and writes receive acknowledgement without guarantees of persistence (data
+     * is eventually fsync'd to disk).<br>
+     * For this kind of ledgers the client MUST explicitly call {@link LedgerHandle#asyncSync(long, org.apache.bookkeeper.client.AsyncCallback.SyncCallback, java.lang.Object)
```
jvrao: asyncSync sounds little out of place. Any other names? like asyncPersist or asyncFlush or asyncCommit something like that?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerType.java
```diff
@@ -0,0 +1,46 @@
...
+     * }
+     * in order to have guarantees of the durability of writes and in order to advance the LastAddConfirmed entry id
+     */
+    FORCE_DEFERRED_ON_JOURNAL
```
jvrao: We have journal today, we may not have journal in next release. 
Why not call these by the function? ImmediatePersist and DelayedPersist?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -1406,7 +1407,7 @@ public void recoveryAddEntry(ByteBuf entry, WriteCallback cb, Object ctx, byte[]
             LedgerDescriptor handle = getLedgerForEntry(entry, masterKey);
             synchronized (handle) {
                 entrySize = entry.readableBytes();
-                addEntryInternal(handle, entry, cb, ctx);
+                addEntryInternal(handle, entry, LedgerType.FORCE_ON_JOURNAL, cb, ctx);
```
jvrao: We must not  play with leader type to instruct immediate sync. The caller must not change ledgertype just because the caller wanted different behavior. We can do an explicit sync call or have another write which can send optional sync/persist flag in addition to ledger type.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -1680,7 +1681,7 @@ public static void main(String[] args)
             buff.writeLong(1);
             buff.writeLong(i);
             cb.incCount();
-            b.addEntry(buff, cb, null, new byte[0]);
+            b.addEntry(buff, cb, null, new byte[0], LedgerType.FORCE_ON_JOURNAL);
```
jvrao: who uses this main()??

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -751,21 +754,17 @@ public boolean accept(long journalId) {
...
-    public void logAddEntry(ByteBuf entry, WriteCallback cb, Object ctx) {
+    public void logAddEntry(ByteBuf entry, LedgerType ledgerType, WriteCallback cb, Object ctx) {
         long ledgerId = entry.getLong(entry.readerIndex() + 0);
         long entryId = entry.getLong(entry.readerIndex() + 8);
         journalQueueSize.inc();
 
         //Retain entry until it gets written to journal
         entry.retain();
-        queue.add(new QueueEntry(entry, ledgerId, entryId, cb, ctx, MathUtils.nowInNano()));
+        queue.add(new QueueEntry(entry, ledgerId, entryId, cb, ctx, MathUtils.nowInNano(), ledgerType));
```
jvrao: We need to have uniform placement for legerType parameter. in logAddEntry it is next to the entry, but here the last argument. I think we should have it right next to entry as it makes sense for hierarchical ordering.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDescriptorImpl.java
```diff
@@ -122,7 +123,7 @@ ByteBuf getExplicitLac() {
             result = logFenceResult = SettableFuture.create();
         }
         ByteBuf entry = createLedgerFenceEntry(ledgerId);
-        journal.logAddEntry(entry, (rc, ledgerId, entryId, addr, ctx) -> {
+        journal.logAddEntry(entry, LedgerType.FORCE_ON_JOURNAL, (rc, ledgerId, entryId, addr, ctx) -> {
```
jvrao: Just repeating my comment. We must not overload ledger type to enforce immediate flush/persist. If needed we need to add another argument.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerCreateOp.java
```diff
@@ -261,15 +269,36 @@ public CreateBuilder withDigestType(org.apache.bookkeeper.client.api.DigestType
...
+                LOG.error("invalid writeQuorumSize {} != ensembleSize {} for FORCE_DEFERRED_ON_JOURNAL ledger. Striping is not allowed",
+                    builderWriteQuorumSize, builderEnsembleSize);
```
jvrao: I know we had some conversation on this. Did we decide striping is not allowed? Do we know it is because of any known consistency issues??

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtocol.java
```diff
@@ -45,11 +45,16 @@
     public static final byte CURRENT_PROTOCOL_VERSION = 2;
```
jvrao: Not changing protocol version? or is it another patch?

bookkeeper-benchmark/src/main/java/org/apache/bookkeeper/benchmark/BenchBookie.java
eolivelli: This is BenchBookie, it only creates LedgerType.FORCE_ON_JOURNAL ledgers.
I tried to do a refactor but this will be a big change in this class. I think it is not important to rewrite this class in this patch.
Mabye it would be better to create an issue on benchmarks in order to support LedgerType and force()

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerFragmentReplicator.java
eolivelli: @sijie @jvrao in this patch I will use the same LedgerType as it is in LedgerMetadata.
I am evaluating if LedgerFragmentReplicator needs an additional "force()" after the write.
This additional change will be in a further patch, in this patch we do not have force() yet

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerType.java
eolivelli: This was from the first patch. Now we have

```
CompletableFuture<Long> force()
```

I'm changing the javadoc in this patch

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerType.java
eolivelli: @sijie @jvrao @ivankelly
I think we should discuss it in the next meeting or on the mailing list.
This name is changing very often during reviews :-)

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
eolivelli: @jvrao. This makes sense. Changing.
I will review the recovery work in further patches

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
eolivelli: @jvrao
I don't know

@sijie can we drop this ? maybe it is another issue

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerFragmentReplicator.java
sijie: we already have a flag indicating an addEntry operation is for recovery add. every recovery add (both ledger recovery and auto recovery) can (and probably should) happen on journal.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
eolivelli: done

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerCreateOp.java
eolivelli: I think that in the latest discussions and the latest version of the document we decided to send force requests to all of the bookies. So actually there should be no problem with striping.

@sijie  @ivankelly

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDescriptorImpl.java
eolivelli: done

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtocol.java
eolivelli: I think we do not need to bump protocol version as far as new clients which use 'default' LedgerType are able to work with 4.5 bookies.
I will add compat tests, we must be sure that new clients which do not use new ledger types are able to work with old bookies

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerCreateOp.java
eolivelli: @sijie  I think I have addressed this one, the link is broken with a rebase

eolivelli:
@sijie  @jvrao patch is ready for second pass. Thank you

bookkeeper-benchmark/src/main/java/org/apache/bookkeeper/benchmark/BenchBookie.java
sijie: okay for me

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -1258,7 +1259,7 @@ private Journal getJournal(long ledgerId) {
     /**
      * Add an entry to a ledger as specified by handle.
      */
-    private void addEntryInternal(LedgerDescriptor handle, ByteBuf entry, WriteCallback cb, Object ctx)
+    private void addEntryInternal(LedgerDescriptor handle, ByteBuf entry, LedgerType ledgerType, WriteCallback cb, Object ctx)
```
sijie: - addEntryInternal takes a boolean `requiresForce` as parameter
- addEntry => calls addEntryInternal(ledgerType.equals(LedgerType.FORCE_ON_JOURNAL)
- recoveryAddEntry => callss addEmtryInternal(true)

all the recovery adds should force writes, no matter what ledger type it is.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -751,21 +753,17 @@ public boolean accept(long journalId) {
...
 
-    public void logAddEntry(ByteBuffer entry, WriteCallback cb, Object ctx) {
-        logAddEntry(Unpooled.wrappedBuffer(entry), cb, ctx);
-    }
-
     /**
      * record an add entry operation in journal.
      */
-    public void logAddEntry(ByteBuf entry, WriteCallback cb, Object ctx) {
+    public void logAddEntry(ByteBuf entry, boolean requiresForce, WriteCallback cb, Object ctx) {
```
sijie: I think `ackAfterForce` is better than `requireForce`.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerCreateOp.java
sijie: @jvrao @eolivelli -- I don't think we disallow striping.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtocol.java
sijie: this is V2 protocol, no?

sijie:
@eolivelli - I remembered that I asked for a BC test in previous comment. but I didn't see any BC related tests. thoughts?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerType.java
sijie: @jvrao I don't understand your comment here. as we discussed before, if we are going to support ledgers bypassing journal, we will have a new ledger type. what does 'ImmediatePersist' mean? does it use journal or not?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
sijie: @jvrao @eolivelli the existence of main function was used for testing, I think.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
ivankelly: ackAfterForce isn't clear either. we want to convey that we want to ack every time, and force some of the time. 
I think forceBeforeAck is better.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDescriptorImpl.java
```diff
@@ -122,7 +123,7 @@ ByteBuf getExplicitLac() {
             result = logFenceResult = SettableFuture.create();
         }
         ByteBuf entry = createLedgerFenceEntry(ledgerId);
-        journal.logAddEntry(entry, (rc, ledgerId, entryId, addr, ctx) -> {
+        journal.logAddEntry(entry, true, (rc, ledgerId, entryId, addr, ctx) -> {
```
ivankelly: Add a comment after true so its clear what true means.
```
journal.logAddEntry(entry, true /* forceBeforeAck */,
                    (rc ...
```

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerFragmentReplicator.java
```diff
@@ -319,7 +321,7 @@ public void readComplete(int rc, LedgerHandle lh,
                 for (BookieSocketAddress newBookie : newBookies) {
                     bkc.getBookieClient().addEntry(newBookie, lh.getId(),
                             lh.getLedgerKey(), entryId, toSend.retainedSlice(),
-                            multiWriteCallback, dataLength, BookieProtocol.FLAG_RECOVERY_ADD);
+                            multiWriteCallback, dataLength, BookieProtocol.FLAG_RECOVERY_ADD, ledgerType);
```
ivankelly: You're never doing a force for fragment replication, so I'd make this add a forcing write, regardless of ledger type.

eolivelli:
@sijie I am thinking about compatibility tests.
What is good to test, my idea is:
1) a new client with "deferred force" MUST NOT be able to write to an old bookie
2) an old client SHOULD able to read ledgers created with "deferred force" (for the reader we are not changing the LAC protocol, and the bit in metadata is not useful for readers)

apart from these cases, for which I can add test cases in next patches, but not in this (because in this patch nothing is "really" working) I think that existing tests are good tests: new clients which use default LedgerType are able to cooperate with old bookies/clients

Do you have other ideas ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerType.java
```diff
@@ -0,0 +1,46 @@
...
+     * Each entry is persisted to the journal and every writes receives and acknowledgement only with the guarantee that
+     * it has been persisted durably to it (data is fsync'd to the disk)
+     */
+    FORCE_ON_JOURNAL,
+    /**
+     * Volatile Durability, using Journal.<br>
+     * Each entry is persisted to the journal and writes receive acknowledgement without guarantees of persistence (data
+     * is eventually fsync'd to disk).<br>
+     * For this kind of ledgers the client MUST explicitly call {@link ForceSupported#force()}
+     * }
```
ivankelly: stray }

pom.xml
```diff
@@ -204,7 +204,8 @@
           <target>1.8</target>
           <compilerArgs>
             <compilerArg>-Werror</compilerArg>
-            <compilerArg>-Xlint:deprecation</compilerArg>
+            <!-- protobuf generates deprecated code for top level enums -->
+            <!--<compilerArg>-Xlint:deprecation</compilerArg>-->
```
ivankelly: Isn't it possible to exclude this code somehow without turning off deprecation checking for the whole code base? We can't be the first who came across this issue.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
eolivelli: done

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
eolivelli: fixed as @sijie suggested. recovery must always force fsync on journal it does not matter LedgerType

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerCreateOp.java
eolivelli: enabled striping

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerFragmentReplicator.java
eolivelli: with FLAG_RECOVERY_ADD on the bookie we will have ackAfterForce=true

pom.xml
eolivelli: I did not find any recipe for maven, see
https://github.com/google/protobuf/issues/2054

eolivelli:
@ivankelly I think I have addresses all of your comments.

I am missing the BC tests requested from @sijie but I am thinking about what could be committed in this patch

pom.xml
ivankelly: There's a couple of approaches possible; None is pretty, but it's prettier than turning off all warnings. I would actually go for c), because at some point we should break the bookkeeper client out of the server package, and in this case, the protocol will be a shared dependency.

a) use maven-replacer to add SuppressWarnings to all the generated classes
b) use a different compile phase for generated code.
c) put all generated code in a separate module

https://stackoverflow.com/questions/39994647/maven-3-how-to-exclude-generated-sources-from-xlint-check
https://github.com/google/protobuf/issues/1596

asfgit:
FAILURE
 
--none--

pom.xml
eolivelli: @ivankelly I prefer c) too, I have created the issue. I will change this after the issue is checked in.
https://github.com/apache/bookkeeper/issues/691

The weird think is that for @sijie the warning does not appear. Maybe he did not run the protoc compiler on this version. The bug is only for top-level enums, not for inner classes

asfgit:
FAILURE
 
--none--

asfgit:
SUCCESS 
 
--none--

asfgit:
FAILURE
 
--none--

pom.xml
sijie: +1 for 3)

sijie:
@eolivelli I would prefer having a BC before committing anything. we need to make sure at any given commit in the master, it doesn't break BC if necessary.

eolivelli:
Totally agree @sijie.
Which kind of compat test can be committed for this part? 
Writing ledger metadata from 4.6  and read with 4.5?
Current tests already cover the case of new client with default ledger type

sijie:
you are changing both metadata and requests, no?

- new client should be able to read metadata produced by old client
- old client should be able to read metadata produced by new client with default ledger type
- old client should not able to read metadata produced by new client with other ledger type

we need similar tests for requests

sijie:
As I pointed out in the master ticket of BP-14 #471, we need documentation for this pull request to be able to be merged, or at least an initial draft of how to add documentation for BP-14.

Issue #279, opened 2017-07-24
Provide LSN in Journal for better checkpointing


sijie:
**FEATURE REQUEST**

1. Please describe the feature you are requesting.

Address the TODO item in **InterleavedLedgerStorage** 

    @Override
    public void onRotateEntryLog() {
        // for interleaved ledger storage, we request a checkpoint when rotating a entry log file.
        // the checkpoint represent the point that all the entries added before this point are already
        // in ledger storage and ready to be synced to disk.
        // TODO: we could consider remove checkpointSource and checkpointSouce#newCheckpoint
        // later if we provide kind of LSN (Log/Journal Squeuence Number)
        // mechanism when adding entry.
        checkpointHolder.setNextCheckpoint(checkpointSource.newCheckpoint());
    }

2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have). Are you currently using any workarounds to address this issue?

*nice-to-have*

3. Provide any additional detail on your proposed use case for this feature.

N/A

ivankelly:
@sijie to confirm what the comment says, it proposes that when we write to the journal for an entry, we take a mark of the point of the journal(called LSN here) at which the entry was persisted, and pass this along with the entry to the ledger storage. The ledger storage can then make it's own decision on when to sync, and when it does, it can just inform the journal of the LSN up to which all data has been persisted. This can be either push or pull from the journal PoV. Correct?

I think this is a good change, and wouldn't be particularly big. Now is a good time to do it too, or maybe once the per ledger entrylog stuff is in, which should be fairly soon.

PR #677, opened 2017-10-27
Issue 659: Fix Checkpoint logic in SortedLedgerStorage

sijie:

- add a test case to reproduce the behavior


jvrao:
@reddycharan  fyi.

asfgit:
FAILURE
 
--none--

asfgit:
SUCCESS 
 
--none--

sijie:
this is ready for review now. @jvrao @ivankelly @reddycharan 

this change is using the original checkpoint implementation which has been used at twitter production for years.

sijie:
@jvrao @ivankelly @reddycharan please take a look when you have time.

ivankelly:
Could you update the PR description to include:
- A description of the original problem
- How this patch addresses that problem

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/SortedLedgerStorageCheckpointTest.java
```diff
@@ -0,0 +1,231 @@
...
+        checkpointSrc.advanceOffset(100);
+
+        // memory table holds the first checkpoint, but it is not completed yet.
+        memtableCp = storage.memTable.kvmap.cp;
+        assertEquals(new TestCheckpoint(0), memtableCp);
+
+        // trigger a memtable flush
+        storage.onSizeLimitReached(checkpointSrc.newCheckpoint());
+        // wait for checkpoint to complete
+        checkpoints.poll(Long.MAX_VALUE, TimeUnit.MILLISECONDS);
```
ivankelly: timeout values here are meaningless

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/SortedLedgerStorageCheckpointTest.java
```diff
@@ -0,0 +1,231 @@
...
+        }
+        // simulate journal persists the entries in journal;
+        checkpointSrc.advanceOffset(100);
+
+        // memory table holds the first checkpoint, but it is not completed yet.
+        memtableCp = storage.memTable.kvmap.cp;
+        assertEquals(new TestCheckpoint(0), memtableCp);
+        assertEquals(20, storage.memTable.kvmap.size());
+
+        final CountDownLatch readyLatch = new CountDownLatch(1);
```
ivankelly: What is the purpose of this?

sijie:
@ivankelly I agree. there is a open issue for that, to provide a better checkpoint mechanism : #279 we can improve the mechanism.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/SortedLedgerStorageCheckpointTest.java
sijie: yes. the test case just waits for checkpoint to complete.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/SortedLedgerStorageCheckpointTest.java
sijie: the purpose is to make sure we roll the log before checkpointing kicks in. it ensures the test case can be executed in a deterministic way.

ivankelly: How does this work though? Is there already a runnable on the executor, so by adding another, we know when the runnable has completed? A comment around this would be very useful for anyone else coming back to this test in a couple of months.

sijie: - rollLog simulates a entry log file rolling (for the use case that Charan reported)
- after this change, an underneath entry log rolling will not trigger any checkpoint, to ensure that there is not checkpoint happening, this test case insert a blocking runnable. so we can collect the right log id at line 209 and line 210 for the assertion at the end.

will add comments.

ivankelly: 👍

ivankelly: @sijie This is ready to merge once that comment is added.

sijie:
ping @jvrao

jvrao:
@sijie I will review if @reddycharan is not going to; I will ping him again tomorrow.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/SortedLedgerStorageCheckpointTest.java
ivankelly: @sijie can you add this comment?

reddycharan:
reviewing it now. Thanks

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/SortedLedgerStorageCheckpointTest.java
sijie: will add the comments after @reddycharan or @jvrao reviewed this.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/InterleavedLedgerStorage.java
```diff
@@ -465,6 +437,9 @@ public void onRotateEntryLog() {
         // TODO: we could consider remove checkpointSource and checkpointSouce#newCheckpoint
         // later if we provide kind of LSN (Log/Journal Squeuence Number)
         // mechanism when adding entry. {@link https://github.com/apache/bookkeeper/issues/279}
-        checkpointHolder.setNextCheckpoint(checkpointSource.newCheckpoint());
+        if (null != checkpointer) {
```
reddycharan: in which case it can be null? and is it ok to be null?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SyncThread.java
```diff
@@ -78,47 +76,43 @@ public SyncThread(ServerConfiguration conf,
...
                     }
-                    if (!disableCheckpoint) {
-                        checkpoint(checkpointSource.newCheckpoint());
-                    }
-                } catch (Throwable t) {
-                    LOG.error("Exception in SyncThread", t);
-                    dirsListener.fatalError();
                 }
+                if (!disableCheckpoint) {
+                    checkpoint(checkpoint);
```
reddycharan: i think somewhere in the call hierarchy, there should be null check for this 'checkpoint' arg. and appropriate logic to handle it

for eg - EntryMemTable.snapshot can return 'null' for checkpoint

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/InterleavedLedgerStorage.java
sijie: SortedLedgerStorage is setting it to null. The trigger of checkpoint for sortedledger storage should happen after flushing memtables, not when entrylog is rolled.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SyncThread.java
sijie: will add that

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/InterleavedLedgerStorage.java
reddycharan: 1) it is very confusing that you added 'checkpointer'  variable both in InterleavedLedgerStorage and its subclass - SortedLedgerStorage. It is easy to get lost or misunderstand.

2.a) I'm not convinced if it is right approach to have different behavior depending on the type of LedgerStorage. In the case of InterleavedLedgerStorage, in the event of 'onRotateEntryLog' you trigger checkpoint but in the case of SortedLedgerStorage it is noop (you swallow it). 
2.b) If you dont agree with what I mentioned in 2.a), then I would like to ask is it necessary to have EntryLogListener.onRotateEntryLog? cannt we have something similar to CacheCallback.onSizeLimitReached in InterleavedLedgerStorage also and get rid of EntryLogListener?

reddycharan: it is very confusing that you added 'checkpointer' variable both in InterleavedLedgerStorage and its subclass - SortedLedgerStorage. It is easy to get lost or misunderstand.

2.a) I'm not convinced if it is right approach to have different behavior depending on the type of LedgerStorage. In the case of InterleavedLedgerStorage, in the event of 'onRotateEntryLog' you trigger checkpoint but in the case of SortedLedgerStorage it is noop (you swallow it).

2.b) If you dont agree with what I mentioned in 2.a), then I would like to ask is it necessary to have EntryLogListener.onRotateEntryLog? cannt we have something similar to CacheCallback.onSizeLimitReached in InterleavedLedgerStorage also and get rid of EntryLogListener?

reddycharan: 1) it is very confusing that you added 'checkpointer' variable both in InterleavedLedgerStorage and its subclass - SortedLedgerStorage. It is easy to get lost or misunderstand.

2.a) I'm not convinced if it is right approach to have different behavior depending on the type of LedgerStorage. In the case of InterleavedLedgerStorage, in the event of 'onRotateEntryLog' you trigger checkpoint but in the case of SortedLedgerStorage it is noop (you swallow it).

2.b) If you dont agree with what I mentioned in 2.a), then I would like to ask is it necessary to have EntryLogListener.onRotateEntryLog? cannt we have something similar to CacheCallback.onSizeLimitReached in InterleavedLedgerStorage also and get rid of EntryLogListener?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SortedLedgerStorage.java
```diff
@@ -146,14 +165,9 @@ public ByteBuf getEntry(long ledgerId, long entryId) throws IOException {
     }
 
     @Override
-    public Checkpoint checkpoint(final Checkpoint checkpoint) throws IOException {
-        Checkpoint lastCheckpoint = checkpointHolder.getLastCheckpoint();
-        // if checkpoint is less than last checkpoint, we don't need to do checkpoint again.
-        if (lastCheckpoint.compareTo(checkpoint) > 0) {
-            return lastCheckpoint;
-        }
+    public void checkpoint(final Checkpoint checkpoint) throws IOException {
```
reddycharan: 1) do we really need this method? the actual caller of this method is SortedLedgerStorage.onSizeLimitReached..in that method "memTable.flush(SortedLedgerStorage.this);" is already called after getting 'checkpoint'. I dont see the need of calling memTable.flush(this, checkpoint); again here. 
2) we can remove this method from SortedLedgerStorage class and also we can remove the argument - 'checkpoint' for LedgerStorage.checkpoint method

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SyncThread.java
```diff
@@ -78,47 +76,43 @@ public SyncThread(ServerConfiguration conf,
             .setNameFormat("SyncThread-" + conf.getBookiePort() + "-%d");
         this.executor = Executors.newSingleThreadScheduledExecutor(tfb.build());
         flushInterval = conf.getFlushInterval();
```
reddycharan: this 'flushInterval' conf is not used anywhere anymore. This conf can be removed and correspondingly all its callers in testcases.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/InterleavedLedgerStorage.java
sijie: @reddycharan 

the logic here is to have each ledger storage control when they want to checkpoint. when ledger storage wants to do checkpoint, it issues a checkpoint request to checkpointer to ask it starts checkpointing. the idea was originated when we introduced checkpoint mechanism to avoid performance penalty. This change was just reverting the behavior the original checkpoint mechanism I introduced 3 years ago, which is proven to be correct and has been running at Twitter production for 3 years. Hope this clarifies 2. a)

the question 1 is unfortunately to be avoided (without changing a lot of code), if we follows the principle letting each ledger storage control when they want to checkpoint, because SortedLedgerStorage inherits from InterleavedLedgerStorage.

regarding 2. b), the idea for checkpointing sorted ledger storage is we only checkpoint sorted ledger storage, when the interleave ledger storage moving on to a new entry log file (which is effectively entry log file rolling). We don't want to checkpoint happens on a file that memtable will potentially flushing entries to. That hurts performance. so in this sense, the checkpoint for sorted ledger storage only happens after a memtable is flushed to entry log file and the entry log file is rotated to a new file.

The implementation here is not elegant as Ivan has raised the concerns 3 years ago. However, it turns out this implementation provides much efficient and correct behavior, which has been verified in twitter production for 3 years. I would suggest us going with this change and doing a refactor later. It is much safer from my experience.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SyncThread.java
sijie: I will mark it as deprecated and remove it the release after next one.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SortedLedgerStorage.java
sijie: 1. we need this method. onSizeLimitReached only flush the memory to entry log files. but it doesn't guarantee data to be flushed to disk. Indeed, it will issue a checkpoint request to checkpointer (which is SyncThread). SyncThread will do the actual checkpoint job. The checkpoint job will ensure the data added before this `checkpoint` will be persisted in entry log files before it update the marker in journal.

2. the argument `checkpoint` is used by memtable to compare if there are data still in memtable that are not flushed to entry log files, when checkpoint starts. it will be used for avoiding unnecessary flush during checkpointing.

reddycharan:
@sijie to summarize your change, SyncThread is not driving checkpoint timing logic anymore, instead LedgerStorage decides when to do checkpoint. 

(just FYI) This works for the existing implementation. But for the multiple entrylogs I'm working on I have to bring back flushInterval and then do force checkpoints for every flushInterval period. Since doing checkpoint on entrylog rotation is not an option when we have multiple entrylogs. Ofcourse whatever I do, I'll make sure existing behavior is not changed anyway for single active entrylog scenario and things are controlled by configs.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SortedLedgerStorage.java
reddycharan: 1) @sijie you misunderstood me. I mean to say do we need to have this overridden method in SortedLedgerStorage? Yes, ofcourse for persistence InterleavedLedgerStorage's checkpoint should be called and I'm not questioning that. My question is why do we need to call "memTable.flush(this, checkpoint);" again? memtable.flush is already called in onSizeLimitReached method, which is the caller for this method.

2) If you agree with my previous point, then you would get why I'm saying this 'checkpoint' argument is not needed for checkpoint method and hence we can remove the argument all together from the method declaration in LedgerStorage interface.

sijie:
@reddycharan yes. with this change, SyncThread is not driving the checkpoint time. The decision is made by ledger storage, so it controls when and how to checkpoint. 

okay for the multiple entrylogs to bring back flushInterval. in that sense, I will hold on removing flushInterval at this moment.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SortedLedgerStorage.java
sijie: @reddycharan - gotcha. I think my 2nd point at previous comment answered your question here :)

the `checkpoint` argument is used by the memtable to compare if there are data is need to flush to entry log files and make it durably stored on entry log files. it is a safeguard here to ensure we got things right. Also `Checkpoint.MAX` can be used for triggering fully flush and checkpoint. Hope this explain your comment.

reddycharan: hmm..you still didn't explain me why we need to call memtable.flush(this,checkpoint) here. I've couple of issues with this

1) SortLS.onSizeLimitReached -> SyncThread.checkpoint -> SortLS.checkpoint is the only call hierarchy in production code for this method. In SLS.onSizeLimitReached "memTable.flush(SortedLedgerStorage.this);" is already called, so what is the point in calling it again?

2) Ok, lets assume there is relevant data in EntryMemTable for this checkpoint, which is added after the last memtable.flush call (line 205) (actually that's not the case), now if we call "memTable.flush(this, checkpoint);" here, it would flush the data from memtable to current active log, but the "super.checkpoint(checkpoint);" will flush only rotatedlogs but not active current log, which defies your assumption/purpose

In summary, there is no point in calling "memTable.flush(this, checkpoint);" here and there is no point in having 'checkpoint' argument to the checkpoint method in LedgerStorage interface itself.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/InterleavedLedgerStorage.java
reddycharan: "the logic here is to have each ledger storage control when they want to checkpoint." Yes, I understood that and I agree in principle what you are doing.
1) I'm merely questioning the choice of variable names. Unless there is a reason, why do we have to get into variable hiding/shadowing. Why cannt we just name them differently say ilsCheckpointer in InterleavedLedgerStorage and slsCheckpointer in SortedLedgerStorage. Anyhow those variables are already private.

sijie: oh, regarding variable -- I don't think there is any specific reasons about renaming. the code was added 3 years ago. I was just following the original path to revert the checkpoint logic back. since they are private, I don't think they will conflict with each other, so I didn't rename them. I am fine with renaming them if you feel strong about that.

ivankelly: Just to add, we shouldn't go into a big refactor until the yahoo changes are in, because the yahoo changes will pull in a whole other ledger storage.

reddycharan: @sijie I think it would be much better and cleaner if you have protected 'checkpointer' variable in InterleavedLedgerStorage and remove 'checkpointer' variable from SortedLedgerStorage. Then override onRotateEntryLog method in SLS and make it noop. So this way it would be easier to get what each class is doing and when they are calling startCheckpoint. The current handling is sneaky and one would wonder why we are having this null check and why we are doing variable hiding/shadowing.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SortedLedgerStorage.java
sijie: @reddycharan nice catch here. I have been looking back to all the historic changes we made towards SortedLedgerStorage. I think `checkpoint` (originally known as logMark) as added as a parameter for optimizing skiplist flush, was originally introduced in twitter/boookkeeper#159be8dcb2079e30291aa877f3436126f0579a0b . the point 2 you are raised here is correct for current checkpoint logic. 

So for this logic here, I am going to stick to having a `checkpoint` as the argument to be a safeguard and address the issue at your 2nd point, to be attributing to original design. I will push a change for it.

sijie:
@reddycharan I push a change to address your comment at bf12164

reddycharan:
@sijie LGTM. Thanks.

PR #682, opened 2017-10-30
Issue 681: Deploying bookkeeper in k8s using StatefulSets

sijie:

Provide a yaml file to deploy bookkeeper in k8s using [StatefulSets](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/)

### DIFF
--- a/deploy/kubernetes/gke/bookkeeper.statefulset.yml
+++ b/deploy/kubernetes/gke/bookkeeper.statefulset.yml
@@ -95,8 +95,6 @@ spec:
   volumeClaimTemplates:
     - metadata:
         name: journaldisk
-        annotations:
-          volume.alpha.kubernetes.io/storage-class: default
         labels:
           component: bookkeeper
       spec:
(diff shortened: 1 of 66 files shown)


deploy/kubernetes/gke/bookkeeper.statefulset.yml
```diff
@@ -0,0 +1,139 @@
...
+            - name: journaldisk
+              mountPath: /bookkeeper/data/journal
+            - name: ledgersdisk
+              mountPath: /bookkeeper/data/ledgers
+
+  volumeClaimTemplates:
+    - metadata:
+        name: journaldisk
+        annotations:
+          volume.alpha.kubernetes.io/storage-class: default
```
merlimat: Annotation here should be not needed anymore, right?

sijie: you are right. going to remove it.

sijie: done

site/docs/latest/deployment/kubernetes.md
```diff
@@ -97,10 +97,14 @@ $ zk-shell localhost 2181
 
 ### Deploy Bookies
 
-Once ZooKeeper cluster is Running, you can then deploy the bookies.
+Once ZooKeeper cluster is Running, you can then deploy the bookies. You can deploy the bookies either using a [DaemonSet](https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/) or a [StatefulSet](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/).
```
merlimat: We could mention that StatefulSet is only practical when deploying in a cloud environment (or any K8S that has persistent volumes available) while DaemonSet is using the local disks

deploy/kubernetes/gke/bookkeeper.statefulset.yml
```diff
@@ -0,0 +1,164 @@
...
+# under the License.
+#
+
+## Credits to Franck Cuny : https://github.com/fcuny/distributedlog-on-k8s/blob/master/bookkeeper.statefulset.yaml
+
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: ssd
+provisioner: kubernetes.io/gce-pd
```
merlimat: I think this will only work on GCE, for AWS it might be different, though I don't know how it could be made portable (other than defining different storage classes).

sijie:
the file is under `gke` directory - that is for deployment in gke specific. we can create a `general` one for general k8s deployment, and have an `aws` directory for deployment in aws specific.

asfgit:
FAILURE
 
--none--

Issue #683, opened 2017-10-31
Ledger length is not updated correctly in LedgerHandleAdv


sijie:
```
ivankelly [4:37 AM] 
@eolivelli the adv ledger stuff, does it take the ledger length into account?


[4:37] 
when adding entries


eolivelli [8:30 AM] 
I hope so


[8:30] 
I am using LEdgerHAndleAdv and it is working


[8:30] 
I have to check code


[8:31] 
yes


[8:31] 
see doAsyncAddEntry in LedgerHandleAdv


[8:31] 
https://github.com/apache/bookkeeper/blob/c6ca4ecbf76f6dac7ebcd843a429dbbaaba62420/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandleAdv.java#L182
GitHub
apache/bookkeeper
bookkeeper - Apache Bookkeeper
 


ivankelly [10:25 AM] 
yes, that's what I'm asking about


[10:25] 
It's basically doubling the length of the ledger each time, without taking the data length into consideration


sijie [10:49 AM] 
@ivankelly I think this is a problem introduced by  netty 4 upgrade


[10:50] 
before netty 4 upgrade, the `length` is the entry length.


[10:52] 
the problem was introduced at 7b1eec47092d0de6776c5a89575dbfc678165ee7
```

sijie:
The problem was introduced when we use ByteBuf in 7b1eec47092d0de6776c5a89575dbfc678165ee7

sijie:
@eolivelli since you haven't cut the 4.5.1 tag, it would be great that we include this in `release/4.5.1` as well.

eolivelli:
Ok

sijie:
@ivankelly has fixed this issue at #664. that will be for master. 

@ivankelly it would be great if you can do a similar fix (without the recycler stuffs) for branch `4.5.1`.

sijie:
removed this from `release/4.6.0` and tag it for `release/4.5.1` only.

eolivelli:
Patch merged to branch-4.5

PR #684, opened 2017-11-02
ISSUE #683: Correctly update length in LedgerHandleAdv

ivankelly:
Commit 7b1eec introduced a change which removed the length parameter
to doAsyncAddEntry. This resulted in the length of the ledger being
added to itself each time for a LedgerHandleAdv, which ultimately
meant that that length would be 0, as it never had anything non-zero
added to it.

This change corrects this, by adding the length of the data parameter
to the ledger length.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandleAdv.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandleAdv.java
@@ -179,7 +179,7 @@ public class LedgerHandleAdv extends LedgerHandle {
                 wasClosed = true;
                 currentLength = 0;
             } else {
-                currentLength = addToLength(length);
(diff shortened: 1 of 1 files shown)


eolivelli:
@ivankelly I think this has been merged in branch-4.5, please check it.
Pull request has not been closed. GitHub integration sometimes lags

sijie:
@eolivelli interesting. how do you merge the pull request? I don't think #683 #684 are updated with correct labels.

eolivelli:
@sijie labels are not updated because I used the merge script of branch 4.5. I know it was an error, but running the script I did not notice that it was not asking my for github stuff.

The very strange fact is that github did not close the pr and the issue. In the commit message there are good references. Maybe such special commit log messages work only on master branch

sijie:
@eolivelli oh, then you should manually update the labels. you should use the script from master.

let's try to update all the labels and close the pr and issue manually.

PR #686, opened 2017-11-02
Issue-685 Merge script: make DEV_BRANCH_NAME configurable

eolivelli:


eolivelli:
@sijie @ivankelly we could take this from the PR but I think that having a config variable is more explicit

PR #687, opened 2017-11-03
Issue-409 BookKeeper Server distribution contains useless jar guava

eolivelli:
Exclude Guava Jar from assembly configuration. Guava is still present by relocated

### DIFF
--- a/src/assemble/bin.xml
+++ b/src/assemble/bin.xml
@@ -71,6 +71,9 @@
       <unpack>false</unpack>
       <scope>runtime</scope>
       <useProjectArtifact>false</useProjectArtifact>
+      <excludes>
+         <exclude>com.google.guava:guava</exclude>
+      </excludes>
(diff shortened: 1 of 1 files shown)


eolivelli:
This change affects bookkeeper-server and bookkeeper-dist

asfgit:
SUCCESS 
 
--none--

sijie:
@eolivelli if you run `mvn dependency:tree`, you will find guava is used at multiple modules (e.g. bookkeeper-benchmark, almost all the stats provider implementations). so your change doesn't work for `bookkeeper-dist` package.

eolivelli:
after all I think that we can close this issue as won't fix.
it is not worth to complicate the build. Guava is still on classpath but shaded on minimal bookie distribution.
ok @sijie  ?

sijie:
sure let me close it. mark it as "won't fix"

Issue #501, opened 2017-09-07
CompactionByEntriesTest>CompactionTest.testMinorCompactionWithNoWritableLedgerDirsButIsForceGCAllowWhenNoSpaceIsSet is flaky


sijie:
**BUG REPORT**

1. Please describe the issue you observed:

- What did you do?

run jenkins build

- What did you expect to see?

all tests should pass

- What did you see instead?

`CompactionByEntriesTest.testMinorCompactionWithNoWritableLedgerDirs` failed

```
CompactionByEntriesTest>CompactionTest.testMinorCompactionWithNoWritableLedgerDirsButIsForceGCAllowWhenNoSpaceIsSet
```

https://builds.apache.org/job/bookkeeper-precommit-pullrequest-docker/10/console

PR #690, opened 2017-11-03
ISSUE #501: Fix race in CompactionTest

ivankelly:
testMinorCompactionWithNoWritableLedgerDirsButIsForceGCAllowWhenNoSpaceIsSet
was failing regularly on jenkins. The issue was that it was trying to
kick off a forced garbage collection by setting the force flag and
triggering. However, in some cases another GC would be running and
would clear the flag after it had been set, but before the trigger was
called.

This patch adds a new form of triggering which allows the caller to
explicitly specify which flags will be active for the GC run.


asfgit:
SUCCESS 
 
--none--

PR #703, opened 2017-11-06
add .travis.yml to skip build asf-site branch

sijie:

although we exclude `asf-site` branch in travis.yml, but the file only exists in master and other release branches. copy the `travis.yml` to `asf-site` branch.

https://travis-ci.org/apache/bookkeeper/builds/298028474

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,43 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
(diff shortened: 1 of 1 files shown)


asfgit:
FAILURE
 
--none--

.travis.yml
```diff
@@ -0,0 +1,43 @@
...
+# limitations under the License.
+language: java
+
+# blacklist
+branches:
+  except:
+    - asf-site
+    - gh-pages
+
+matrix:
```
ivankelly: do we need all the extra junk? Surely the "branches: except" should be enough? why do we have gh-pages there?

sijie: we don't need them. I was just copying the file from master. I will only keep branches:except.

I think the yml file was from `distributedlog`, which it has a `gh-pages`. we can remove it from here.

asfgit:
FAILURE
 
--none--

.travis.yml
sijie: removed unused content, only keep `branches`

PR #704, opened 2017-11-06
BOOKKEEPER-1040: Use separate log for compaction and add transaction support

yzang:
Problem:
Compaction might keep generating duplicated data which would cause disk full.
This is because we don't have transactional operation for compaction. So if compaction
keeps failing in the middle, bookie would end up with a lot of garbage data.

Change:
1. Introduce abstract class AbstractLogCompactor with an interface compact().
2. Move the existing compaction logic to a separate class called EntryLogCompactor.
3. Introduce transactional compaction, we can recover an incomplete compaction or rollback a compaction failure.
4. Add an configuration to enable transactional compaction

Potential Risk:
1. No risk if we keep using the default compactor.
2. If we choose to enable transactional compaction with separate log file, we need to be careful about it, since we use a separate log for compaction, if we have a lot of old small ledgers in the bookie, we will end up with a lot of small entry log files. And this will potentially cause bookkeeper "Too many open files" because GC will scan all the entry log files and keep them open.

> ---
> quickly and easily:
> 
> 
> ---

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
@@ -247,9 +247,8 @@ public class GarbageCollectorThread extends SafeRunnable {
             LOG.info("Garbage collector thread forced to perform GC before expiry of wait time.");
         }
         // Recover and clean up previous state if using transactional compaction
-        if (compactor instanceof TransactionalEntryLogCompactor) {
-            ((TransactionalEntryLogCompactor) compactor).cleanUpAndRecover();
-        }
+        compactor.cleanUpAndRecover();
+
         // Extract all of the ledger ID's that comprise all of the entry logs
         // (except for the current new one which is still being written to).
         entryLogMetaMap = extractMetaFromEntryLogs(entryLogMetaMap);
@@ -338,16 +337,11 @@ public class GarbageCollectorThread extends SafeRunnable {
     @VisibleForTesting
     void doCompactEntryLogs(double threshold) {
         LOG.info("Do compaction to compact those files lower than {}", threshold);
-        // sort the ledger meta by occupied unused space
-        // sort the ledger meta by occupied unused space
-        Comparator<EntryLogMetadata> usageComparator = (m1, m2) -> {
-            double usage1 = m1.getUsage();
-            double usage2 = m2.getUsage();
-            return Double.compare(usage1, usage2);
-        };
+
+        // sort the ledger meta by usage in ascending order.
         List<EntryLogMetadata> logsToCompact = new ArrayList<EntryLogMetadata>();
         logsToCompact.addAll(entryLogMetaMap.values());
-        Collections.sort(logsToCompact, usageComparator);
+        logsToCompact.sort(Comparator.comparing(EntryLogMetadata::getUsage));
 
         for (EntryLogMetadata meta : logsToCompact) {
             if (meta.getUsage() >= threshold) {
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
@@ -22,10 +22,8 @@
 package org.apache.bookkeeper.bookie;
 
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.util.concurrent.RateLimiter;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import java.io.IOException;
-import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
@@ -37,7 +35,7 @@ import java.util.concurrent.Future;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
-import org.apache.bookkeeper.bookie.EntryLogger.EntryLogScanner;
+
 import org.apache.bookkeeper.bookie.GarbageCollector.GarbageCleaner;
 import org.apache.bookkeeper.conf.ServerConfiguration;
 import org.apache.bookkeeper.meta.LedgerManager;
@@ -76,15 +74,9 @@ public class GarbageCollectorThread extends SafeRunnable {
 
     final boolean isForceGCAllowWhenNoSpace;
 
-    final boolean isThrottleByBytes;
-    final int maxOutstandingRequests;
-    final int compactionRateByEntries;
-    final int compactionRateByBytes;
-    final CompactionScannerFactory scannerFactory;
-
     // Entry Logger Handle
     final EntryLogger entryLogger;
-
+    final AbstractLogCompactor compactor;
     final CompactableLedgerStorage ledgerStorage;
 
     // flag to ensure gc thread will not be interrupted during compaction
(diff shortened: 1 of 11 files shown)


asfgit:
FAILURE
 
--none--

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/TransactionalEntryLogCompactor.java
```diff
@@ -0,0 +1,413 @@
...
+    final CompactableLedgerStorage ledgerStorage;
+    final Throttler throttler;
+    final List<EntryLocation> offsets = new ArrayList<>();
+
+    // compaction log file suffix
+    static final String COMPACTING_SUFFIX = ".log.compacting";
+    // flushed compaction log file suffix
+    static final String COMPACTED_SUFFIX = ".compacted";
+
+    private class Throttler {
```
sijie: do we need to a separate class? I think we have a `Throttler` in `AbstractLogCompactor`?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/TransactionalEntryLogCompactor.java
```diff
@@ -0,0 +1,413 @@
...
+     * The first phase is to scan entries in 1.log and copy them to compaction log file "3.log.compacting".
+     * We'll try to allocate a new compaction log before scanning to make sure we have a log file to write.
+     * If after scanning, there's no data written, it means there's no valid entries to be compacted,
+     * so we can remove 1.log directly, clear the offsets and end the compaction.
+     * Otherwise, we should move on to the next phase.
+     * <p>
+     * If anything failed in this phase, we should delete the compaction log and clean the offsets.
+     */
+    class ScanEntryLogPhase extends CompactionPhase {
+        private EntryLogMetadata metadata;
```
sijie: make this final

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/TransactionalEntryLogCompactor.java
```diff
@@ -0,0 +1,413 @@
...
+            entryLogger.scanEntryLog(metadata.getEntryLogId(), new EntryLogScanner() {
+                @Override
+                public boolean accept(long ledgerId) {
+                    return metadata.containsLedger(ledgerId);
+                }
+
+                @Override
+                public void process(long ledgerId, long offset, ByteBuffer entry) throws IOException {
+                    throttler.acquire(entry.remaining());
+                    synchronized (this) {
```
sijie: I think this synchronization should be `synchronized (TransactionalEntryLogCompactor.this)`, no?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/TransactionalEntryLogCompactor.java
```diff
@@ -0,0 +1,413 @@
...
+
+    /**
+     * Assume we're compacting log 1 to log 3.
+     * This phase is to flush the compaction log.
+     * When this phase starts, there should be a compaction log file like "3.log.compacting"
+     * When compaction log is flushed, in order to indicate this phase is completed,
+     * a hardlink file "3.log.1.compacted" should be created, and "3.log.compacting" should be deleted.
+     */
+    class FlushCompactionLogPhase extends CompactionPhase {
+        long compactingLogId;
```
sijie: final

sijie:
@yzang did a quick review, have some minor comments. will go through the pull request again later today or tomorrow

sijie:
mark this feature for release `4.7.0`.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/TransactionalEntryLogCompactor.java
yzang: I'm thinking maybe we don't need synchronization here? Since we only have one thread doing GC and compaction anyway..

asfgit:
FAILURE
 
--none--

asfgit:
FAILURE
 
--none--

asfgit:
FAILURE
 
--none--

asfgit:
FAILURE
 
--none--

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -335,7 +246,10 @@ public void safeRun() {
         if (force) {
             LOG.info("Garbage collector thread forced to perform GC before expiry of wait time.");
         }
-
+        // Recover and clean up previous state if using transactional compaction
+        if (compactor instanceof TransactionalEntryLogCompactor) {
+            ((TransactionalEntryLogCompactor) compactor).cleanUpAndRecover();
```
eolivelli: It is better to add an empty default method _cleanUpAndRecover_ to AbstractLogCompactor and override it in TransactionalEntryLogCompactor .  It would be a more "object oriented"

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -423,51 +337,26 @@ private void doGcEntryLogs() {
...
+        // sort the ledger meta by occupied unused space
+        Comparator<EntryLogMetadata> usageComparator = (m1, m2) -> {
+            double usage1 = m1.getUsage();
+            double usage2 = m2.getUsage();
+            return Double.compare(usage1, usage2);
         };
         List<EntryLogMetadata> logsToCompact = new ArrayList<EntryLogMetadata>();
         logsToCompact.addAll(entryLogMetaMap.values());
-        Collections.sort(logsToCompact, sizeComparator);
+        Collections.sort(logsToCompact, usageComparator);
```
eolivelli: what about something like this ?

`logsToCompact.sort(Comparator.comparing(EntryLogMetadata::getUsage)`

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/TransactionalEntryLogCompactor.java
```diff
@@ -0,0 +1,387 @@
...
+     */
+    public void cleanUpAndRecover() {
+        // clean up compacting logs and recover index for already compacted logs
+        List<File> ledgerDirs = entryLogger.getLedgerDirsManager().getAllLedgerDirs();
+        for (File dir : ledgerDirs) {
+            File[] compactingPhaseFiles = dir.listFiles(file -> file.getName().endsWith(COMPACTING_SUFFIX));
+            if (compactingPhaseFiles != null) {
+                for (File file : compactingPhaseFiles) {
+                    if (file.delete()) {
+                        LOG.info("Deleted failed compaction file {}", file);
```
eolivelli: what happens if delete fails ?

bookkeeper-server/src/test/java/org/apache/bookkeeper/test/BookKeeperClusterTestCase.java
```diff
@@ -241,7 +241,12 @@ protected void stopAllBookies() throws Exception {
         for (BookieServer server : bs) {
             server.shutdown();
         }
+        bsConfs.clear();
         bs.clear();
+        if (bkc != null) {
```
eolivelli: this seems unrelated to the change, does this impact the compaction tests ?
I am fine with this change anyway

yzang: Yeah, this is changed for compaction tests.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
yzang: Didn't know we can do that, that's a good suggestion. Fixed.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
yzang: That's a good idea, fixed.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/TransactionalEntryLogCompactor.java
yzang: I'm not sure if there's anything to do when delete fails, it could happen when either the file doesn't exist or we have an IO issue.

asfgit:
FAILURE
 
--none--

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/TransactionalEntryLogCompactor.java
sijie: it doesn't really matter here. because next time when compaction runs, it will eventually clean up all the `compacting` files.

PR #706, opened 2017-11-07
Bookies should not queue read request indefinitely

agresch:
Integrating some changes made to Yahoo bookkeeper by Matteo Merli.  

The apache bookkeeper thread pool in use by OrderedScheduler is different, and does not have access to setting a queue size limit, so I created QueueAssessibleExecutorService, and check the queue size instead.  Other than that, this code change is pretty similar to Yahoo's.

Original bug description for this change:

Since we are using a thread pool to handle read requests in bookies, we have seen that when the Bookie read IO is maxed out, the requests are being accumulated in the bookie.

Essentially, the bookie is busy serving read requests that have already been timed out on the client side (end to end read latency can reach hours..). The queue keeps growing indefinitely, leading to OOM errors and bookie restart.

All unit tests pass for me locally with this change.


asfgit:
FAILURE
 
--none--

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/TestMaxSizeWorkersQueue.java
```diff
@@ -0,0 +1,106 @@
...
+        final CountDownLatch counter = new CountDownLatch(N);
+        final AtomicBoolean receivedTooManyRequestsException = new AtomicBoolean();
+
+        // Write few entries
+        for (int i = 0; i < N; i++) {
+            lh.asyncAddEntry(content, new AddCallback() {
+                @Override
+                public void addComplete(int rc, LedgerHandle lh, long entryId, Object ctx) {
+                    if (rc == BKException.Code.NotEnoughBookiesException) {
+                        receivedTooManyRequestsException.set(true);
```
ivankelly: Why does it get this error? Shouldn't it be TooManyRequestException?

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/QueueAssessibleExecutorService.java
```diff
@@ -0,0 +1,156 @@
...
+import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.TimeoutException;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Implements {@link ListeningScheduledExecutorService} and allows limiting the number
+ * of tasks to be scheduled in the thread's queue.
+ *
+ */
+public class QueueAssessibleExecutorService implements ListeningScheduledExecutorService {
```
ivankelly: The name isn't very good. Call it something like BoundedScheduledExecutorService

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/OrderedScheduler.java
```diff
@@ -88,6 +89,7 @@ public static SchedulerBuilder newSchedulerBuilder() {
         protected StatsLogger statsLogger = NullStatsLogger.INSTANCE;
         protected boolean traceTaskExecution = false;
         protected long warnTimeMicroSec = WARN_TIME_MICRO_SEC_DEFAULT;
+        protected int maxTasksInQueue = -1;
```
ivankelly: define -1 as a constant (NO_LIMIT or something).

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/QueueAssessibleExecutorService.java
ivankelly: You can reduce the number of useless implemented methods, like shutdown, by extending from ForwardingListeningExecutorService.

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/QueueAssessibleExecutorService.java
```diff
@@ -0,0 +1,156 @@
...
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
+import java.util.concurrent.RejectedExecutionException;
+import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.TimeoutException;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Implements {@link ListeningScheduledExecutorService} and allows limiting the number
```
ivankelly: There's something I don't like about extending scheduled executor service to be bounded, but I can't quite put my finger on it. Surely we're not the first people to need something like this, but JDK never implemented it (according to the code to make execution handling easier[1]), guava doesn't implement it. 

As a review comment, this is neither here nor there, but I'd like someone else to review to reassure me it's ok (@sijie ?).

[1] https://github.com/frohoff/jdk8u-jdk/blob/master/src/share/classes/java/util/concurrent/ScheduledThreadPoolExecutor.java#L134

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/TestMaxSizeWorkersQueue.java
agresch: See the comment on line 82.  Unfortunately, I'm not the author, and I'm new to the code base.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/TestMaxSizeWorkersQueue.java
```diff
@@ -0,0 +1,106 @@
...
+            @Override
+            public void readComplete(int rc, LedgerHandle lh, Enumeration<LedgerEntry> seq, Object ctx) {
+                rcSecondReadOperation.set(rc);
+                counter.countDown();
+            }
+        }, lh);
+
+        counter.await();
+
+        // assertEquals(BKException.Code.OK, rcFirstReadOperation.get());
```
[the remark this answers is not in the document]
agresch: Thanks, it's commented out in our code base, but seems to run with the assert fine.

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/QueueAssessibleExecutorService.java
agresch: Good tip.  Thanks.

asfgit:
FAILURE
 
--none--

asfgit:
SUCCESS 
 
--none--

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/BKException.java
```diff
@@ -214,6 +214,7 @@ public static String getMessage(int code) {
          * Write operations failed due to bookies are readonly.
          */
         int WriteOnReadOnlyBookieException = -104;
+        int TooManyRequestsException = -105;
```
jiazhai: It may need to add a comments for this exception as other exception here.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
```diff
@@ -288,7 +292,24 @@ private void processAddRequestV3(final BookkeeperProtocol.Request r, final Chann
...
         } else {
-            writeThreadPool.submitOrdered(r.getAddRequest().getLedgerId(), write);
+            try {
+                writeThreadPool.submitOrdered(r.getAddRequest().getLedgerId(), write);
+            } catch (RejectedExecutionException e) {
+                if (LOG.isDebugEnabled()) {
+                    LOG.debug("Failed to process request to add entry at {}:{}. Too many pending requests",
+                            r.getAddRequest().getLedgerId(), r.getAddRequest().getEntryId());
+                }
+                BookkeeperProtocol.AddResponse.Builder addResponse = BookkeeperProtocol.AddResponse.newBuilder() //
```
sijie: nit: removing the trailing "//"

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
```diff
@@ -316,7 +337,25 @@ private void processReadRequestV3(final BookkeeperProtocol.Request r, final Chan
...
-                readThreadPool.submitOrdered(r.getReadRequest().getLedgerId(), read);
+                try {
+                    readThreadPool.submitOrdered(r.getReadRequest().getLedgerId(), read);
+                } catch (RejectedExecutionException e) {
+                    if (LOG.isDebugEnabled()) {
+                        LOG.debug("Failed to process request to read entry at {}:{}. Too many pending requests",
+                                r.getReadRequest().getLedgerId(), r.getReadRequest().getEntryId());
+                    }
+                    BookkeeperProtocol.ReadResponse.Builder readResponse =
+                            BookkeeperProtocol.ReadResponse.newBuilder() //
```
sijie: nit: removing trailing "//"

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/TestMaxSizeWorkersQueue.java
```diff
@@ -0,0 +1,106 @@
+package org.apache.bookkeeper.client;
```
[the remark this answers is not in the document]
sijie: it is actually very interesting... this license header is missed in this file, but the travis build passes...

asfgit:
FAILURE
 
--none--

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/TestMaxSizeWorkersQueue.java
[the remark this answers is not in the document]
ivankelly: Maybe because it's a test?

jiazhai:
@agresch , Would you please solve the conflicting file? then we can ship it.

Issue #691, opened 2017-11-03
Create separate module for Protobuf files


eolivelli:
Protobuf compiler has some bug and created codes with warning
Seen in https://github.com/apache/bookkeeper/pull/670

We can move protobuf related stuff to a separate module.

Thinking about separating client-side code from bookie-side code this will be a necessary step

PR #711, opened 2017-11-09
ISSUE #691: Move generated files into their own module

ivankelly:
Upcoming changes are using GRpc in conjunction with protobuf. GRpc
generates code that created deprecation warnings when compiled with
java 8, so this change moves all generated code out to another module,
so that we don't have to turn off -Werror for all code.

In any case, at some point we should split the bookkeeper client out
from the server module, at which point we would need the definitions
on a common place.


bookkeeper-server/pom.xml
```diff
@@ -35,10 +35,9 @@
       <version>${project.parent.version}</version>
     </dependency>
     <dependency>
-      <groupId>com.google.protobuf</groupId>
-      <artifactId>protobuf-java</artifactId>
-      <version>${protobuf.version}</version>
-      <scope>compile</scope>
+      <groupId>org.apache.bookkeeper</groupId>
```
eolivelli: What about project.parent.grouiId?

bookkeeper-proto/pom.xml
```diff
@@ -0,0 +1,70 @@
...
+  <parent>
+    <artifactId>bookkeeper</artifactId>
+    <groupId>org.apache.bookkeeper</groupId>
+    <version>4.6.0-SNAPSHOT</version>
+  </parent>
+  <artifactId>bookkeeper-proto</artifactId>
+  <name>Apache BookKeeper :: Wire Protocols</name>
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <project.libdir>${basedir}/lib</project.libdir>
```
eolivelli: Unused?

ivankelly: will remove.

bookkeeper-server/pom.xml
ivankelly: We don't use it for common. We should be consistent IMO.

asfgit:
FAILURE
 
--none--

bookkeeper-proto/pom.xml
```diff
@@ -0,0 +1,70 @@
...
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <artifactId>bookkeeper</artifactId>
+    <groupId>org.apache.bookkeeper</groupId>
+    <version>4.6.0-SNAPSHOT</version>
+  </parent>
+  <artifactId>bookkeeper-proto</artifactId>
+  <name>Apache BookKeeper :: Wire Protocols</name>
```
sijie: it is not just wire protocols. it also contain metadata protobuf definitions. we need to change it to "Protocols".

ivankelly:
@eolivelli Addressed comments in latest push

asfgit:
FAILURE
 
--none--

PR #714, opened 2017-11-10
Issue-612 BOOKKEEPER-818: Make bookie recovery work with recovering multiple bookies

sijie:

This is a change ported from twitter branch.

- Enable recovering multiple bookies in bookie shell
- Add a unit test for BookieShell.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperAdmin.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperAdmin.java
@@ -42,10 +42,13 @@ import java.util.NoSuchElementException;
 import java.util.Optional;
 import java.util.Random;
 import java.util.Set;
+import java.util.SortedMap;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentSkipListMap;
 import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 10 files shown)


asfgit:
FAILURE
 
--none--

asfgit:
FAILURE
 
--none--

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperAdmin.java
```diff
@@ -442,6 +449,69 @@ public SyncObject() {
         }
     }
 
+    public SortedMap<Long, LedgerMetadata> getLedgersContainBookies(Set<BookieSocketAddress> bookies)
+            throws InterruptedException, BKException {
+        final SyncObject sync = new SyncObject();
```
eolivelli: I know this is kind of cherry pick.
Can you consider dropping this and use CompletableFuture. This legacy way of coding can be affected by spurious wake ups

sijie: I would defer this to a separate task. BookKeeperAdmin needs a cleanup. I don't want to mix two approaches in the same file. how does that sound?

eolivelli: sure, it is best to make one step at a time

Issue #490, opened 2017-09-04
Add a flag to not serialize `ctime` field


sijie:
**FEATURE REQUEST**

1. Please describe the feature you are requesting.

Add a flag to not serialize `ctime` field.

This request is for DL bookkeeper version upgrade. Because twitter's bk version doesn't include this change, so any ledgers written by new bk version will fail to be read by old version. To simplify the upgrade story, it is good to allow not serialize `ctime` field.

2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have). Are you currently using any workarounds to address this issue?

*nice-to-have*

3. Provide any additional detail on your proposed use case for this feature.

N/A

eolivelli:
Dropping the 4.5.1 label from this issue

// @sijie already agreed by email

PR #718, opened 2017-11-11
Issue 490: Add a flag to not serialize `ctime` field

sijie:

This request is for DL bookkeeper version upgrade. Because twitter's bk version doesn't include this change, so any ledgers written by new bk version will fail to be read by old version. To simplify the upgrade story, it is good to allow not serialize ctime field.


sijie:
This is the pull request for #490

sijie:
@eolivelli since the `ctime` feature was added by you, I am adding a flag for it. I turn it off to simplify the upgrade story. it should not impact your logic, because it will fall back to use the ctime in metadata store. please check and let me know if it is okay.

asfgit:
FAILURE
 
--none--

sijie:
@merlimat +1 on switching to binary format.

eolivelli:
+1 for me to for binary format

PR #720, opened 2017-11-11
Fixed simultaneus reads on same ledger/entry with v2 protocol

merlimat:
This change is coming from yahoo branch at yahoo/bookkeeper@ce7b0f7

With v2 protocol there's no way to disambiguate the read request since the completion key is made from (ledgerId, entryId).
This PR introduced an additional multimap to handle the conflicts (multiple simultaneus read requests for the same entry).

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
@@ -20,46 +20,6 @@ package org.apache.bookkeeper.proto;
 
 import static org.apache.bookkeeper.client.LedgerHandle.INVALID_ENTRY_ID;
 
-import com.google.common.base.Joiner;
-import com.google.common.collect.Sets;
-import com.google.protobuf.ByteString;
-import com.google.protobuf.ExtensionRegistry;
-
-import io.netty.bootstrap.Bootstrap;
-import io.netty.buffer.ByteBuf;
-import io.netty.buffer.ByteBufAllocator;
-import io.netty.buffer.PooledByteBufAllocator;
-import io.netty.buffer.Unpooled;
-import io.netty.buffer.UnpooledByteBufAllocator;
-import io.netty.channel.Channel;
-import io.netty.channel.ChannelFuture;
-import io.netty.channel.ChannelFutureListener;
-import io.netty.channel.ChannelHandler.Sharable;
-import io.netty.channel.ChannelHandlerContext;
-import io.netty.channel.ChannelInboundHandlerAdapter;
-import io.netty.channel.ChannelInitializer;
-import io.netty.channel.ChannelOption;
-import io.netty.channel.ChannelPipeline;
-import io.netty.channel.DefaultEventLoopGroup;
-import io.netty.channel.EventLoopGroup;
-import io.netty.channel.WriteBufferWaterMark;
-import io.netty.channel.epoll.EpollEventLoopGroup;
-import io.netty.channel.epoll.EpollSocketChannel;
-import io.netty.channel.local.LocalChannel;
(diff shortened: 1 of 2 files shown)


asfgit:
FAILURE
 
--none--

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookKeeperTest.java
```diff
@@ -818,4 +818,68 @@ public void testReadEntryReleaseByteBufs() throws Exception {
             }
         }
     }
+
+    /**
+     * Tests that issuing multiple reads for the same entry at the same time works as expected.
+     *
+     * @throws Exception
+     */
+    @Test(timeout = 20000)
```
jiazhai: Remove timemout? 
I recall there was a require to remove all the "timeout", since we have a global timeout set in surefire configuration.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookKeeperTest.java
```diff
@@ -818,4 +818,68 @@ public void testReadEntryReleaseByteBufs() throws Exception {
...
+
+        latch.await();
+    }
+
+    /**
+     * Tests that issuing multiple reads for the same entry at the same time works as expected.
+     *
+     * @throws Exception
+     */
+    @Test(timeout = 20000)
```
jiazhai: same here

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
```diff
@@ -845,6 +846,56 @@ private void writeAndFlush(final Channel channel,
             errorOut(key);
         }
     }
+    
+    void errorOutReadKey(final CompletionKey key) {
```
ivankelly: This code isn't valid anymore. Erroring out is now part of the completionValue, so this should be handled there.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
```diff
@@ -732,14 +729,18 @@ private void readEntryInternal(final long ledgerId,
...
-            // There's a pending read request on same ledger/entry. This is not supported in V2 protocol
-            LOG.warn("Failing concurrent request to read at ledger: {} entry: {}", ledgerId, entryId);
-            completion.errorOut(BKException.Code.UnexpectedConditionException);
+            // There's a pending read request on same ledger/entry. Use the multimap to track all of them
+            synchronized (this) {
+                completionObjectsV2Conflicts.put(completionKey, readCompletion);
+            }
+        }
+
+        final Channel c = channel;
```
ivankelly: This is handled in writeAndFlush now. You'll need to add a check for the conflicts map into the #errorOut method.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookKeeperTest.java
merlimat: Sure, will update

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
merlimat: Ok, I've missed that part. I'll fix that

sijie:
@merlimat @ivankelly : any updates on this PR? I was lost who is picking up pr.

ivankelly:
I made a PR into matteo's branch. When he merges that, this will be ready
for review again

merlimat:
Rebased and applied fixes from @ivankelly

Issue #612, opened 2017-10-17
Make bookie recovery work with recovering multiple bookies


ivankelly:
JIRA: https://issues.apache.org/jira/browse/BOOKKEEPER-818

Reporter: Sijie Guo @sijie

Make recovery tool work with multiple bookies so we could recover bookies inside ensemble at one time call. so we don't need to run the tool multiple times when we want to recover multiple bookies

### Comments from JIRA

---
*Sijie Guo* 2017-07-28T00:03:08.816+0000

move replication related changes to 4.6.0

sijie:
#641 address the core issue. I need to integrate this in the tool. I will do this for `4.6.0`.

PR #721, opened 2017-11-13
Issue 674: Documentation for HTTP endpoints

zhaijack:
Add a document for http endpoints.

### DIFF
--- a/site/_data/sidebar.yaml
+++ b/site/_data/sidebar.yaml
@@ -28,7 +28,7 @@ groups:
     endpoint: metrics
   - name: Upgrade
     endpoint: upgrade
-  - name: BookKeeper http endpoint
+  - name: BookKeeper Admin REST API
     endpoint: http
   #- name: Geo-replication
   #  endpoint: geo-replication
--- a/site/_data/sidebar.yaml
+++ b/site/_data/sidebar.yaml
@@ -28,6 +28,8 @@ groups:
     endpoint: metrics
   - name: Upgrade
     endpoint: upgrade
+  - name: BookKeeper http endpoint
+    endpoint: http
   #- name: Geo-replication
   #  endpoint: geo-replication
   #- name: Customized placement policies
--- a/site/docs/latest/admin/http.md
+++ b/site/docs/latest/admin/http.md
@@ -1,0 +1,317 @@
+---
+title: BookKeeper http endpoints
+---
+
+This document is to introducing BookKeeper http endpoints that could be used for BookKeeper administration.
+If user want to use this feature, parameter "httpServerEnabled" need to be set to "true" in file conf/bk_server.conf.
+
+## All the endpoints
+
+Currently all the http endpoints could be divided into these 4 components:
+1. Heartbeat: heartbeat for a specific bookie.
+1. Config: doing the server configuration for a specific bookie.
+1. Ledger: http endpoints related to ledgers.
+1. Bookie: http endpoints related to bookies.
+1. AutoRecovery: http endpoints related to auto recovery.
+
+## Heartbeat
+
+### Endpoint: /heartbeat
+* Method: GET
+* Description: Get heartbeat status for a specific bookie
+* Response: 
+
+| Code   | Description |
+|:-------|:------------|
+|200 | Successful operation |
+
+## Config
+
+### Endpoint: /api/v1/config/server_config
+1. Method: GET
+    * Description: Get value of all configured values overridden on local server config
+    * Response:  
+    
+        | Code   | Description |
+        |:-------|:------------|
+        |200 | Successful operation |
+        |403 | Don't have permission |
+        |404 | Error handling |
+1. Method: PUT
+    * Description: Update a local server config
+    * Parameters: 
+    
+        | Name | Type | Required | Description |
+        |:-----|:-----|:---------|:------------|
+        |configName  | String | Yes | Configuration name(key) |
+        |configValue | String | Yes | Configuration value(value) |
+    * Body: 
+        ```json
+        {
+           "configName1": "configValue1",
+           "configName2": "configValue2"
+        }
+        ```
+    * Response:  
+    
+        | Code   | Description |
+        |:-------|:------------|
+        |200 | Successful operation |
+        |403 | Don't have permission |
+        |404 | Error handling |
+
+## Ledger
+
+### Endpoint: /api/v1/ledger/delete/?ledger_id=<ledger_id>
+1. Method: DELETE
+    * Description: Delete a ledger.
+    * Parameters: 
+    
+        | Name | Type | Required | Description |
+        |:-----|:-----|:---------|:------------|
+        |ledger_id  | Long | Yes | ledger id of the ledger.  |
+    * Response:  
+    
+        | Code   | Description |
+        |:-------|:------------|
+        |200 | Successful operation |
+        |403 | Don't have permission |
+        |404 | Error handling |
+
+### Endpoint: /api/v1/ledger/list/?print_metadata=<metadata>
+1. Method: GET
+    * Description: List all the ledgers.
+    * Parameters: 
+    
+        | Name | Type | Required | Description |
+        |:-----|:-----|:---------|:------------|
+        |metadata  | Boolean | No |  whether print out metadata  |
+    * Response:  
+    
+        | Code   | Description |
+        |:-------|:------------|
+        |200 | Successful operation |
+        |403 | Don't have permission |
+        |404 | Error handling |
+
+### Endpoint: /api/v1/ledger/metadata/?ledger_id=<ledger_id>
+1. Method: GET
+    * Description: Get the metadata of a ledger.
+    * Parameters: 
+    
+        | Name | Type | Required | Description |
+        |:-----|:-----|:---------|:------------|
+        |ledger_id  | Long | Yes | ledger id of the ledger.  |
+    * Response:  
+    
+        | Code   | Description |
+        |:-------|:------------|
+        |200 | Successful operation |
+        |403 | Don't have permission |
+        |404 | Error handling |
+
+### Endpoint: /api/v1/ledger/read/?ledger_id=<ledger_id>&start_entry_id=<start_entry_id>&end_entry_id=<end_entry_id>
+1. Method: GET
+    * Description: Read a range of entries from ledger.
+    * Parameters: 
+    
+        | Name | Type | Required | Description |
+        |:-----|:-----|:---------|:------------|
+        |ledger_id       | Long | Yes| ledger id of the ledger.  |
+        |start_entry_id  | Long | No | start entry id of read range. |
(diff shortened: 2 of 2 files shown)


site/docs/latest/admin/http.md
```diff
@@ -0,0 +1,317 @@
+---
+title: BookKeeper http endpoints
+---
+
+This document is to introducing BookKeeper http endpoints that could be used for BookKeeper administration.
```
ivankelly: This document introduces BookKeeper HTTP endpoints, which can be used ...

site/docs/latest/admin/http.md
```diff
@@ -0,0 +1,317 @@
...
+If user want to use this feature, parameter "httpServerEnabled" need to be set to "true" in file conf/bk_server.conf.
```
ivankelly: To use this feature, set "httpServerEnabled" to "true" in file conf/bk_server.conf.

site/docs/latest/admin/http.md
```diff
@@ -0,0 +1,317 @@
...
+
+## All the endpoints
+
+Currently all the http endpoints could be divided into these 4 components:
```
ivankelly: The HTTP endpoints are divided into the following groups:

ivankelly: Perhaps they should be grouped by endpoints which are bookie specific and those which are for the whole cluster.

site/docs/latest/admin/http.md
```diff
@@ -0,0 +1,317 @@
...
+1. Method: GET
+    * Description: Get value of all configured values overridden on local server config
+    * Response:  
+    
+        | Code   | Description |
+        |:-------|:------------|
+        |200 | Successful operation |
+        |403 | Don't have permission |
+        |404 | Error handling |
+1. Method: PUT
```
ivankelly: Is there already an admin command to do this? Does this access a config in zookeeper or on the local disk?

site/docs/latest/admin/http.md
```diff
@@ -0,0 +1,317 @@
...
+        |404 | Error handling |
+
+### Endpoint: /api/v1/ledger/list/?print_metadata=<metadata>
+1. Method: GET
+    * Description: List all the ledgers.
+    * Parameters: 
+    
+        | Name | Type | Required | Description |
+        |:-----|:-----|:---------|:------------|
+        |metadata  | Boolean | No |  whether print out metadata  |
```
ivankelly: api doesn't print.

site/docs/latest/admin/http.md
```diff
@@ -0,0 +1,317 @@
...
+### Endpoint: /api/v1/config/server_config
+1. Method: GET
+    * Description: Get value of all configured values overridden on local server config
+    * Response:  
+    
+        | Code   | Description |
+        |:-------|:------------|
+        |200 | Successful operation |
+        |403 | Don't have permission |
+        |404 | Error handling |
```
ivankelly: What does "Error handling" mean? Shouldn't this be "Not found"?

site/docs/latest/admin/http.md
```diff
@@ -0,0 +1,317 @@
...
+
+### Endpoint: /api/v1/config/server_config
+1. Method: GET
+    * Description: Get value of all configured values overridden on local server config
+    * Response:  
+    
+        | Code   | Description |
+        |:-------|:------------|
+        |200 | Successful operation |
+        |403 | Don't have permission |
```
ivankelly: "Permission denied"

site/docs/latest/admin/http.md
```diff
@@ -0,0 +1,317 @@
...
+        |404 | Error handling |
+
+### Endpoint: /api/v1/ledger/read/?ledger_id=<ledger_id>&start_entry_id=<start_entry_id>&end_entry_id=<end_entry_id>
+1. Method: GET
```
ivankelly: What format are these entries returned in?

site/docs/latest/admin/http.md
```diff
@@ -0,0 +1,317 @@
...
+* Description: Get heartbeat status for a specific bookie
+* Response: 
+
+| Code   | Description |
+|:-------|:------------|
+|200 | Successful operation |
+
+## Config
+
+### Endpoint: /api/v1/config/server_config
```
ivankelly: The format of the response should be included in each call.

site/docs/latest/admin/http.md
```diff
@@ -0,0 +1,317 @@
...
+          "bookie_src": [ "bookie_src1", "bookie_src2"... ],
+          "bookie_dest": [ "bookie_dest1", "bookie_dest2"... ],
+          "delete_cookie": <bool_value>
+        }
+        ```
+    * Parameters:
+     
+        | Name | Type | Required | Description |
+        |:-----|:-----|:---------|:------------|
+        |bookie_src  | String | Yes | bookie source to recovery |
```
ivankelly: String or array of strings?

ivankelly: Should the strings have a specific format?

site/docs/latest/admin/http.md
```diff
@@ -0,0 +1,317 @@
...
+        |404 | Error handling |
+
+### Endpoint: /api/v1/autorecovery/list_under_replicated_ledger/?missingreplica=<bookie_address>&excludingmissingreplica=<bookie_address>
+1. Method: GET
+    * Description:  Get all under replicated ledgers.
+    * Parameters: 
+    
+        | Name | Type | Required | Description |
+        |:-----|:-----|:---------|:------------|
+        |missingreplica  | String | No | missing replica bookieId |
```
ivankelly: These parameters should have underscores between words to be consistent.

site/docs/latest/admin/http.md
```diff
@@ -0,0 +1,317 @@
...
+## Bookie
+
+### Endpoint: /api/v1/bookie/list_bookies/?type=<type>&print_hostnames=<hostnames>
+1. Method: GET
+    * Description:  Get all the available bookies.
+    * Parameters: 
+    
+        | Name | Type | Required | Description |
+        |:-----|:-----|:---------|:------------|
+        |type      | String | Yes |  value: rw/ro , list read-write/read-only bookies. |
```
ivankelly: rw/ro is the value? or rw or ro?

asfgit:
FAILURE
 
--none--

site/docs/latest/admin/http.md
zhaijack: Thanks, will change it.

site/docs/latest/admin/http.md
zhaijack: Thanks, wll change it.

site/docs/latest/admin/http.md
zhaijack: Thanks. will change the words. 
And it is a good idea to do a new grouping following your suggestion. will consider to do it, and opened a new issue #722 to tracking this, but as the requirements goes on, maybe 2 big groups need futher dividing.

site/docs/latest/admin/http.md
zhaijack: Thanks. the response is construct of code and body, and most of the command the response body not contains too much info. Will add the usefule respone body format.

site/docs/latest/admin/http.md
zhaijack: thanks. will add the format

site/docs/latest/admin/http.md
zhaijack: strings of host names or IP:Port

site/docs/latest/admin/http.md
zhaijack: will change to "rw or ro"

site/docs/latest/admin/http.md
zhaijack: seems not command for this. it only update the inmemory value, which could be used for later ops.

site/docs/latest/admin/http.md
sijie: @zhaijack with my latest change in #612, it should be a list of strings.

site/docs/latest/admin/http.md
ivankelly: I'm not sure it's wise to expose functionality like this at all. It's just asking for people to change configuration in an uncontrolled and unpredictable manner. Was there a driving use case to add this to the http endpoints?

lucperkins:
For the sake of maintainability, I think it'd be best to generate these docs from a JSON or YAML definition using a template, but I can do that work in a follow-up PR

site/docs/latest/admin/http.md
sijie: @ivankelly I think http endpoints (and configuration service) was introduced by twitter initially and Jia extended it to cover all admin commands. I can see this is useful on dynamic configurations.  but Twitter folks can chime in more on the use cases.

BTW If there is a question on a feature whether this should be included or not, we should probably be discussing a bit earlier on the coding pull request, rather than later on the documentation pull request. (for future improvements)

sijie:
@lucperkins a nice point! +1

site/_data/sidebar.yaml
```diff
@@ -28,6 +28,8 @@ groups:
     endpoint: metrics
   - name: Upgrade
     endpoint: upgrade
+  - name: BookKeeper http endpoint
```
sijie: I would suggest changing "http endpoint" to "Admin REST API", which would be more accurate.

site/docs/latest/admin/http.md
```diff
@@ -0,0 +1,394 @@
+---
+title: BookKeeper http endpoints
```
sijie: I would suggest changing "http endpoint" to "Admin REST API", which would be more accurate.

jiazhai: thanks

site/_data/sidebar.yaml
jiazhai: thanks. will change it.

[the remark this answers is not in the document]
jiazhai: thanks. will follow @sijie 's comments

site/docs/latest/admin/http.md
```diff
@@ -0,0 +1,394 @@
...
+---
+
+This document introduces BookKeeper HTTP endpoints, which can be used for BookKeeper administration.
+To use this feature, set "httpServerEnabled" to "true" in file conf/bk_server.conf.
```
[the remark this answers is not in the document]
jiazhai: Thanks. will change it.

site/docs/latest/admin/http.md
```diff
@@ -0,0 +1,394 @@
...
+
+## All the endpoints
+
+Currently all the http endpoints could be divided into these 4 components:
```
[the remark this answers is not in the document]
jiazhai: Thanks. will change it.

asfgit:
FAILURE
 
--none--

Issue #506, opened 2017-09-12
BP-15 New CreateLedger API


eolivelli:
Refer to BookKeeper Proposal 15 for a detailed description of this new feature
https://cwiki.apache.org/confluence/display/BOOKKEEPER/BP-15+New+CreateLedger+API

We are going to add a brand new API to manage ledgers using the Builder pattern.
Additionally we will introduce new interfaces to make it clear operations on ledgers, like WriteHandler and ReadHandler.
The new API will leverage CompletableFuture API

sijie:
can you create an issue for adding documentation for the new API and mark it as a blocker? we need documentation to claim BP-15 is completed.

PR #727, opened 2017-11-15
Issue 693: add interface and implementation of LedgerEntries

zhaijack:
In the new API proposed in #506 , we return Iterable<LedgerEntry> for read entries. It is a bit problematic when returning multiple iterators and have no ability to release the entry buffers hold by ledger entries.
In this change, the implementation of retainIterator() increased the reference of the buffers when creating an iterator, it will be easy for byteBuf's management; it also make LedgerEntries a recycle object, and it is returned to the pool when #close() is called. on #close(), it also releases all the references to release resources.

1. add interface LedgerEntries;
2. add implemtation of LedgerEntries;
3. fix related build and test errors.

> ---
> quickly and easily:
> 
> 
> ---

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
@@ -63,6 +63,7 @@ import org.apache.bookkeeper.client.SyncCallbackUtils.SyncReadCallback;
 import org.apache.bookkeeper.client.SyncCallbackUtils.SyncReadLastConfirmedCallback;
 import org.apache.bookkeeper.client.api.BKException.Code;
 import org.apache.bookkeeper.client.api.LastConfirmedAndEntry;
+import org.apache.bookkeeper.client.api.LedgerEntries;
 import org.apache.bookkeeper.client.api.WriteHandle;
 import org.apache.bookkeeper.client.impl.LedgerEntryImpl;
 import org.apache.bookkeeper.common.concurrent.FutureEventListener;
@@ -634,7 +635,7 @@ public class LedgerHandle implements WriteHandle {
      *          id of last entry of sequence
      */
     @Override
-    public CompletableFuture<org.apache.bookkeeper.client.api.LedgerEntries> read(long firstEntry, long lastEntry) {
+    public CompletableFuture<LedgerEntries> read(long firstEntry, long lastEntry) {
         // Little sanity check
         if (firstEntry < 0 || firstEntry > lastEntry) {
             LOG.error("IncorrectParameterException on ledgerId:{} firstEntry:{} lastEntry:{}",
@@ -675,8 +676,8 @@ public class LedgerHandle implements WriteHandle {
      * @see #readUnconfirmedEntries(long, long)
      */
     @Override
-    public CompletableFuture<org.apache.bookkeeper.client.api.LedgerEntries> readUnconfirmed(long firstEntry,
-                                                                                             long lastEntry) {        // Little sanity check
+    public CompletableFuture<LedgerEntries> readUnconfirmed(long firstEntry, long lastEntry) {
+        // Little sanity check
         if (firstEntry < 0 || firstEntry > lastEntry) {
             LOG.error("IncorrectParameterException on ledgerId:{} firstEntry:{} lastEntry:{}",
                     new Object[] { ledgerId, firstEntry, lastEntry });
@@ -689,14 +690,14 @@ public class LedgerHandle implements WriteHandle {
     void asyncReadEntriesInternal(long firstEntry, long lastEntry, ReadCallback cb, Object ctx) {
         if(!bk.isClosed()) {
             readEntriesInternalAsync(firstEntry, lastEntry)
-                .whenCompleteAsync(new FutureEventListener<org.apache.bookkeeper.client.api.LedgerEntries>() {
+                .whenCompleteAsync(new FutureEventListener<LedgerEntries>() {
                     @Override
-                    public void onSuccess(org.apache.bookkeeper.client.api.LedgerEntries iterable) {
+                    public void onSuccess(LedgerEntries entries) {
                         cb.readComplete(
                             Code.OK,
                             LedgerHandle.this,
                             IteratorUtils.asEnumeration(
-                                Iterators.transform(iterable.iterator(), le -> {
+                                Iterators.transform(entries.iterator(), le -> {
                                     LedgerEntry entry = new LedgerEntry((LedgerEntryImpl) le);
                                     le.close();
                                     return entry;
@@ -719,8 +720,8 @@ public class LedgerHandle implements WriteHandle {
         }
     }
 
-    CompletableFuture<org.apache.bookkeeper.client.api.LedgerEntries> readEntriesInternalAsync(long firstEntry,
-                                                                                               long lastEntry) {
+    CompletableFuture<LedgerEntries> readEntriesInternalAsync(long firstEntry,
+                                                              long lastEntry) {
         PendingReadOp op = new PendingReadOp(this, bk.getScheduler(), firstEntry, lastEntry);
         if(!bk.isClosed()) {
             bk.getMainWorkerPool().submitOrdered(ledgerId, op);
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
@@ -634,7 +634,7 @@ public class LedgerHandle implements WriteHandle {
      *          id of last entry of sequence
      */
     @Override
-    public CompletableFuture<Iterable<org.apache.bookkeeper.client.api.LedgerEntry>> read(long firstEntry, long lastEntry) {
+    public CompletableFuture<org.apache.bookkeeper.client.api.LedgerEntries> read(long firstEntry, long lastEntry) {
         // Little sanity check
         if (firstEntry < 0 || firstEntry > lastEntry) {
             LOG.error("IncorrectParameterException on ledgerId:{} firstEntry:{} lastEntry:{}",
@@ -675,8 +675,8 @@ public class LedgerHandle implements WriteHandle {
      * @see #readUnconfirmedEntries(long, long)
      */
     @Override
-    public CompletableFuture<Iterable<org.apache.bookkeeper.client.api.LedgerEntry>> readUnconfirmed(long firstEntry, long lastEntry) {
-        // Little sanity check
+    public CompletableFuture<org.apache.bookkeeper.client.api.LedgerEntries> readUnconfirmed(long firstEntry,
+                                                                                             long lastEntry) {        // Little sanity check
         if (firstEntry < 0 || firstEntry > lastEntry) {
             LOG.error("IncorrectParameterException on ledgerId:{} firstEntry:{} lastEntry:{}",
                     new Object[] { ledgerId, firstEntry, lastEntry });
@@ -689,9 +689,9 @@ public class LedgerHandle implements WriteHandle {
     void asyncReadEntriesInternal(long firstEntry, long lastEntry, ReadCallback cb, Object ctx) {
         if(!bk.isClosed()) {
             readEntriesInternalAsync(firstEntry, lastEntry)
-                .whenCompleteAsync(new FutureEventListener<Iterable<org.apache.bookkeeper.client.api.LedgerEntry>>() {
+                .whenCompleteAsync(new FutureEventListener<org.apache.bookkeeper.client.api.LedgerEntries>() {
                     @Override
-                    public void onSuccess(Iterable<org.apache.bookkeeper.client.api.LedgerEntry> iterable) {
+                    public void onSuccess(org.apache.bookkeeper.client.api.LedgerEntries iterable) {
                         cb.readComplete(
                             Code.OK,
                             LedgerHandle.this,
@@ -719,8 +719,8 @@ public class LedgerHandle implements WriteHandle {
         }
     }
 
-    CompletableFuture<Iterable<org.apache.bookkeeper.client.api.LedgerEntry>> readEntriesInternalAsync(long firstEntry,
-                                                                                                       long lastEntry) {
+    CompletableFuture<org.apache.bookkeeper.client.api.LedgerEntries> readEntriesInternalAsync(long firstEntry,
+                                                                                               long lastEntry) {
         PendingReadOp op = new PendingReadOp(this, bk.getScheduler(), firstEntry, lastEntry);
         if(!bk.isClosed()) {
             bk.getMainWorkerPool().submitOrdered(ledgerId, op);
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
@@ -36,7 +36,7 @@ public interface LedgerEntries extends AutoCloseable {
 
     /**
      * In this, It does not retain the ByteBuf references for the entries in this LedgerEntries.
-     * The caller who calls #iterator() should be careful for not releasing the references.
+     * The caller who calls {@link #iterator()} should be careful for not releasing the references.
      *
      * @return the iterator of type LedgerEntry
      */
@@ -44,7 +44,7 @@ public interface LedgerEntries extends AutoCloseable {
 
     /**
      * In this, It retains the ByteBuf references for the entries in this LedgerEntries.
-     * The caller who calls #iterator() is responsible for releasing the retained references.
+     * The caller who calls {@link #retainIterator()} is responsible for releasing the retained references.
      *
      * @return the iterator of type LedgerEntry that has been retained
      */
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
@@ -35,10 +35,15 @@ public interface LedgerEntries extends AutoCloseable {
     LedgerEntry getEntry(long entryId);
 
     /**
-     * In this, It does not retain the ByteBuf references for the entries in this LedgerEntries.
+     * In this method, It does not increment the reference counts of ByteBuf for the entries in this LedgerEntries.
      * The caller who calls {@link #iterator()} should be careful for not releasing the references.
      *
-     * @return the iterator of type LedgerEntry
+     * when iterator is called, you are handing out the entries, you may not know when the caller will
+     * complete iterating the entries. In this case, please use {@link #retainIterator()}, it will increment
+     * the refCnt, and the application who is holding iterator should be responsible for releasing one refCnt
+     * that {@link #retainIterator()} retains for entries.
+     *
+     *  @return the iterator of type LedgerEntry
      */
     Iterator<LedgerEntry> iterator();
 
@@ -48,5 +53,6 @@ public interface LedgerEntries extends AutoCloseable {
      *
      * @return the iterator of type LedgerEntry that has been retained
      */
+
     Iterator<LedgerEntry> retainIterator();
 }
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
@@ -35,24 +35,11 @@ public interface LedgerEntries extends AutoCloseable {
     LedgerEntry getEntry(long entryId);
 
     /**
-     * In this method, It does not increment the reference counts of ByteBuf for the entries in this LedgerEntries.
+     * This method does not increment the reference count of ByteBuf for the entries in this LedgerEntries.
      * The caller who calls {@link #iterator()} should be careful for not releasing the references.
-     *
-     * when iterator is called, you are handing out the entries, you may not know when the caller will
-     * complete iterating the entries. In this case, please use {@link #retainedIterator()}, it will increment
-     * the refCnt, and the application who is holding iterator should be responsible for releasing one refCnt
-     * that {@link #retainedIterator()} retains for entries.
+     * The implementation of this interface will release all the Entries ByteBuf reference when close.
      *
      *  @return the iterator of type LedgerEntry
      */
     Iterator<LedgerEntry> iterator();
-
-    /**
-     * In this, It retains the ByteBuf references for the entries in this LedgerEntries.
-     * The caller who calls {@link #retainedIterator()} is responsible for releasing the retained references.
-     *
-     * @return the iterator of type LedgerEntry that has been retained
-     */
-
-    Iterator<LedgerEntry> retainedIterator();
 }
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
@@ -39,9 +39,9 @@ public interface LedgerEntries extends AutoCloseable {
      * The caller who calls {@link #iterator()} should be careful for not releasing the references.
      *
      * when iterator is called, you are handing out the entries, you may not know when the caller will
-     * complete iterating the entries. In this case, please use {@link #retainIterator()}, it will increment
+     * complete iterating the entries. In this case, please use {@link #retainedIterator()}, it will increment
      * the refCnt, and the application who is holding iterator should be responsible for releasing one refCnt
-     * that {@link #retainIterator()} retains for entries.
+     * that {@link #retainedIterator()} retains for entries.
      *
      *  @return the iterator of type LedgerEntry
      */
@@ -49,10 +49,10 @@ public interface LedgerEntries extends AutoCloseable {
 
     /**
      * In this, It retains the ByteBuf references for the entries in this LedgerEntries.
-     * The caller who calls {@link #retainIterator()} is responsible for releasing the retained references.
+     * The caller who calls {@link #retainedIterator()} is responsible for releasing the retained references.
      *
      * @return the iterator of type LedgerEntry that has been retained
      */
 
-    Iterator<LedgerEntry> retainIterator();
+    Iterator<LedgerEntry> retainedIterator();
 }
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
@@ -1,0 +1,52 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.bookkeeper.client.api;
+
+import java.util.Iterator;
+
+/**
+ * Interface to wrap the entries.
+ *
+ * @since 4.6
+ */
+public interface LedgerEntries extends AutoCloseable {
+
+    /**
+     * Gets a specific LedgerEntry by entryId.
+     *
+     * @param entryId the LedgerEntry id
+     * @return the LedgerEntry, null if no LedgerEntry with such entryId.
+     */
+    LedgerEntry getEntry(long entryId);
+
+    /**
+     * In this, It does not retain the ByteBuf references for the entries in this LedgerEntries.
+     * The caller who calls #iterator() should be careful for not releasing the references.
+     *
+     * @return the iterator of type LedgerEntry
+     */
+    Iterator<LedgerEntry> iterator();
+
+    /**
+     * In this, It retains the ByteBuf references for the entries in this LedgerEntries.
+     * The caller who calls #iterator() is responsible for releasing the retained references.
+     *
+     * @return the iterator of type LedgerEntry that has been retained
+     */
+    Iterator<LedgerEntry> retainIterator();
+}
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
@@ -35,11 +35,12 @@ public interface LedgerEntries extends AutoCloseable {
     LedgerEntry getEntry(long entryId);
 
     /**
-     * This method does not increment the reference count of ByteBuf for the entries in this LedgerEntries.
-     * The caller who calls {@link #iterator()} should be careful for not releasing the references.
-     * The implementation of this interface will release all the Entries ByteBuf reference when close.
+     * Calling this method does not modify the reference count of the ByteBuf in the returned LedgerEntry objects.
+     * The caller who calls {@link #iterator()} should make sure that they do not call ByteBuf.release() on the
+     * LedgerEntry objects to avoid a double free.
+     * All reference counts will be decremented when the containing LedgerEntries object is closed.
      *
-     *  @return the iterator of type LedgerEntry
+     * @return the iterator of type LedgerEntry.
      */
     Iterator<LedgerEntry> iterator();
 }
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/impl/LedgerEntriesImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/impl/LedgerEntriesImpl.java
@@ -101,7 +101,7 @@ public class LedgerEntriesImpl implements LedgerEntries {
      * {@inheritDoc}
      */
     @Override
-    public Iterator<LedgerEntry> retainIterator() {
+    public Iterator<LedgerEntry> retainedIterator() {
         checkNotNull(entries, "entries has been recycled");
         // Retains the references for all the entries, caller is responsible for releasing.
         entries.forEach(ledgerEntry -> ledgerEntry.getEntryBuffer().retain());
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/impl/LedgerEntriesImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/impl/LedgerEntriesImpl.java
@@ -1,0 +1,104 @@
+/*
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+package org.apache.bookkeeper.client.impl;
+
+import io.netty.util.Recycler;
+import java.util.Iterator;
+import java.util.List;
+import org.apache.bookkeeper.client.api.LedgerEntry;
+
+public class LedgerEntriesImpl implements org.apache.bookkeeper.client.api.LedgerEntries {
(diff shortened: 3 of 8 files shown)


asfgit:
FAILURE
 
--none--

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
```diff
@@ -0,0 +1,44 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.bookkeeper.client.api;
+
+import java.util.Iterator;
+
+/**
+ * Interface to wrap the entries.
```
eolivelli: we should state the need to close this.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * @since 4.6
+ */
+public interface LedgerEntries {
```
eolivelli: we should extend AutoCloseable

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
```diff
@@ -0,0 +1,44 @@
...
+
+    /**
+     * Gets a specific LedgerEntry by entryId.
```
eolivelli: what happens if there is no such entry ?
I this that we should throw an IllegalArgumentException

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
```diff
@@ -0,0 +1,44 @@
...
+     *
+     * @param entryId the LedgerEntry id
+     * @return the LedgerEntry
+     */
+    LedgerEntry getEntry(long entryId);
+
+    /**
+     * In this, It retains the references for the entries in this LedgerEntries.
+     * The caller who calls #iterator() is responsible for releasing the retained references.
```
eolivelli: we are anyhow retaining a reference.
maybe we can do some cleanup on close().

maybe we could enforce and state that this method can be called only once

/cc @sijie @merlimat @ivankelly

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
jiazhai: Thanks, The extend of AutoCloseable was put in the implementation.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
jiazhai: Thanks, Following the convention in LedgerHandle, It returned null in the implementation. 
I will add comments for this.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
jiazhai: Thanks, @eolivelli  What is other cleanup work needed?
It can be called not only once, each time it retains a reference.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
jiazhai: Thanks, I C, will move extend AutoCloseable in to this file.

jiazhai:
@eolivelli, Regarding the release/retain story, Here is the understanding:

Before this change, read() return result as Iterable<Ledgers>, and if user call Iterable.iterator() multiple times, the references of ByteBuf have potential to be broken, and hard to manage. 
e.g. one of the iterator accessed some entry and then released the reference; while another iterator still want to access the same entry. 

By this change, each call of iterator() will inc the references, and after usage, caller will dec the reference, it is manage-able and more safe.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerEntries.java
```diff
@@ -0,0 +1,101 @@
...
+        long lastId = entries.get(entries.size() - 1).getEntryId();
+        if (entryId < firstId || entryId > lastId) {
+            return null;
+        }
+        return entries.get((int)(entryId - firstId));
+    }
+
+    @Override
+    public Iterator<org.apache.bookkeeper.client.api.LedgerEntry> iterator() {
+        // Retains the references for all the entries, caller is responsible for releasing.
```
mgodave: Extend Iterable for ergonomics?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
sijie: @eolivelli - my take here is: when iterator is called, you are handing out the entries, you don't know when the caller will complete iterating the entries. in this case, you have to increment the refCnt. The application who is holding iterator should be responsible for releasing one refcnt that it retains for entries. 

I think getting a more consistent behavior, we can follow the pattern that ByteBuf has, to have two methods: one is `iterator()`, the other one is `retainIterator()`. in `iterator` we don't need to increment the reference counts, while `retainIterator()`, it increment the reference counts. so user have a deterministic behavior - depends on how do they want to keep the iterators.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerEntries.java
```diff
@@ -0,0 +1,102 @@
...
+        long firstId = entries.get(0).getEntryId();
+        long lastId = entries.get(entries.size() - 1).getEntryId();
+        if (entryId < firstId || entryId > lastId) {
+            return null;
+        }
+        return entries.get((int)(entryId - firstId));
+    }
+
+    @Override
+    public Iterator<org.apache.bookkeeper.client.api.LedgerEntry> iterator() {
```
sijie: I think we have to rethink on the `LedgerEntry` interface as well. The `LedgerEntry` interface is completely a mess. `getEntry()` and `getEntryBuffer()` have completely different behaviors. `getEntry()` releases bytebuf automatically, while `getEntryBuffer()` returns the bytebuf (if you don't call getEntry, you are responsible for releasing the entry buffer.

I would suggest:

1) removing `getEntry()` from LedgerEntry, or make it reenterable. 
2) make LedgerEntry implement AutoCloseable. LedgerEntry should be responsible for releasing the bytebuf it is holding.
3) LedgerEntry should implement duplicate method. `iterator` here should have a separate copy of LedgerEntry and make a slice of the underlying bytebuf.

@jiazhai @eolivelli

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/SyncCallbackUtils.java
```diff
@@ -17,12 +17,11 @@
  */
 package org.apache.bookkeeper.client;
 
-import com.google.common.collect.Iterators;
+import static org.apache.bookkeeper.client.LedgerHandle.LOG;
```
sijie: this is not needed.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerEntries.java
eolivelli: Good ideas

eolivelli: Good ideas

jiazhai: +1, opened #731 for this idea.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/SyncCallbackUtils.java
jiazhai: Thanks. It is originally there, this change adjust the order of importing, will handle it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerEntries.java
```diff
@@ -0,0 +1,102 @@
...
+ */
+package org.apache.bookkeeper.client;
+
+import io.netty.util.Recycler;
+import java.util.Collections;
+import java.util.Enumeration;
+import java.util.Iterator;
+import java.util.List;
+
+public class LedgerEntries implements org.apache.bookkeeper.client.api.LedgerEntries {
```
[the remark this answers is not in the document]
jiazhai: Thanks. will do it. The originally intention is to follow LedgerEntry 's way.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
jiazhai: Thanks. will change it.

asfgit:
FAILURE
 
--none--

zhaijack:
This PR need update after PR #738  is merged.

jiazhai:
Rebased master, since the LedgerEntry changes. 
@sijie @eolivelli Would you please help check this again.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -634,7 +634,7 @@ public void asyncReadUnconfirmedEntries(long firstEntry, long lastEntry, ReadCal
      *          id of last entry of sequence
      */
     @Override
-    public CompletableFuture<Iterable<org.apache.bookkeeper.client.api.LedgerEntry>> read(long firstEntry, long lastEntry) {
+    public CompletableFuture<org.apache.bookkeeper.client.api.LedgerEntries> read(long firstEntry, long lastEntry) {
```
sijie: nit: I don't think you need fully qualified path name. it can just be `CompletableFuture<LedgerEntries>`.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -675,8 +675,8 @@ public void asyncReadUnconfirmedEntries(long firstEntry, long lastEntry, ReadCal
      * @see #readUnconfirmedEntries(long, long)
      */
     @Override
-    public CompletableFuture<Iterable<org.apache.bookkeeper.client.api.LedgerEntry>> readUnconfirmed(long firstEntry, long lastEntry) {
-        // Little sanity check
+    public CompletableFuture<org.apache.bookkeeper.client.api.LedgerEntries> readUnconfirmed(long firstEntry,
+                                                                                             long lastEntry) {        // Little sanity check
```
sijie: need put the comment in a new line.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
```diff
@@ -0,0 +1,52 @@
...
+     * In this, It does not retain the ByteBuf references for the entries in this LedgerEntries.
+     * The caller who calls #iterator() should be careful for not releasing the references.
+     *
+     * @return the iterator of type LedgerEntry
+     */
+    Iterator<LedgerEntry> iterator();
+
+    /**
+     * In this, It retains the ByteBuf references for the entries in this LedgerEntries.
+     * The caller who calls #iterator() is responsible for releasing the retained references.
```
sijie: change `#iterator()` to `{@link #retainIterator()}`

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/impl/LedgerEntriesImpl.java
```diff
@@ -0,0 +1,104 @@
...
+ *
+ */
+package org.apache.bookkeeper.client.impl;
+
+import io.netty.util.Recycler;
+import java.util.Iterator;
+import java.util.List;
+import org.apache.bookkeeper.client.api.LedgerEntry;
+
+public class LedgerEntriesImpl implements org.apache.bookkeeper.client.api.LedgerEntries {
```
sijie: you don't need use fully qualified class name. just change it to `LedgerEntries`.

sijie: please add a class comment

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/impl/LedgerEntriesImpl.java
```diff
@@ -0,0 +1,104 @@
...
+    };
+
+    private void recycle() {
+        releaseByteBuf();
+        recyclerHandle.recycle(this);
+    }
+
+    private void releaseByteBuf() {
+        if (entries != null) {
+            for (LedgerEntry entry : entries) {
```
sijie: entries.forEach(LedgerEntry::close());
entries.clear();
entries = null;

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/impl/LedgerEntriesImpl.java
```diff
@@ -0,0 +1,104 @@
...
+        }
+    }
+
+    /**
+     * Create ledger entries.
+     *
+     * @param entries the entries with ordering
+     * @return the LedgerEntriesImpl
+     */
+    public static LedgerEntriesImpl create(List<LedgerEntry> entries) {
```
sijie: checkArgument(!entries.isEmpty(), "...")

line 73 assumes entries is not empty.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/impl/LedgerEntriesImpl.java
```diff
@@ -0,0 +1,104 @@
...
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public LedgerEntry getEntry(long entryId) {
+        long firstId = entries.get(0).getEntryId();
+        long lastId = entries.get(entries.size() - 1).getEntryId();
+        if (entryId < firstId || entryId > lastId) {
+            return null;
```
sijie: we might just follow the pattern how ArrayList works. throw `ArrayIndexOutOfBoundsException` when requesting an entry is out of the range.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/impl/LedgerEntriesImpl.java
```diff
@@ -0,0 +1,104 @@
...
+        }
+        return entries.get((int)(entryId - firstId));
```
sijie: nit: space between ')('

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/api/BookKeeperApiTest.java
```diff
@@ -275,11 +276,13 @@ public void testCannotDeleteLedgerTwice() throws Exception {
         result(newDeleteLedgerOp().withLedgerId(lId).execute());
     }
 
-    private static void checkEntries(Iterable<LedgerEntry> entries, byte[] data)
+    private static void checkEntries(LedgerEntries entries, byte[] data)
         throws InterruptedException, BKException {
-        for (LedgerEntry le : entries) {
-            assertArrayEquals(data, le.getEntry());
+        Iterator<LedgerEntry> iterator = entries.retainIterator();
```
sijie: I don't think you need retainIterator here.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/impl/LedgerEntriesImpl.java
```diff
@@ -0,0 +1,104 @@
+/*
```
sijie: Can you add a test case `LedgerEntriesImplTest`? Like what I did for `LedgerEntryImplTest`?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
jiazhai: Thanks, will change it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
jiazhai: Thanks. will change it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
jiazhai: Thanks. will change it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/impl/LedgerEntriesImpl.java
jiazhai: Thanks. will add it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/impl/LedgerEntriesImpl.java
jiazhai: Thanks. will change it.

jiazhai: Thanks. will add it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/impl/LedgerEntriesImpl.java
jiazhai: Thanks. will change it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/impl/LedgerEntriesImpl.java
jiazhai: Thanks. will change it.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/api/BookKeeperApiTest.java
jiazhai: Thanks. right, it is now underneath LedgerEntryImpl, so below line entry.getEntry() will not release the reference now. will change it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/impl/LedgerEntriesImpl.java
jiazhai: Thanks. will change it

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/impl/LedgerEntriesImpl.java
jiazhai: Thanks. will change it.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/impl/LedgerEntriesImplTest.java
```diff
@@ -0,0 +1,130 @@
...
+
+    @After
+    public void tearDown() {
+        ledgerEntriesImpl.close();
+
+        try {
+            ledgerEntriesImpl.getEntry(entryId);
+            fail("should fail getEntry after close");
+        } catch (NullPointerException e) {
+
```
sijie: I would suggest not leaving an empty line. you can add some comments like "expected behavior"

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/impl/LedgerEntriesImplTest.java
```diff
@@ -0,0 +1,130 @@
...
+    public void testIterator() {
+        Iterator<LedgerEntry> entryIterator = ledgerEntriesImpl.iterator();
+        entryIterator.forEachRemaining(ledgerEntry -> assertEquals(1, ledgerEntry.getEntryBuffer().refCnt()));
+    }
+
+    @Test
+    public void testRetainIterator() {
+        Iterator<LedgerEntry> entryIterator = ledgerEntriesImpl.retainIterator();
+        entryIterator.forEachRemaining(ledgerEntry -> assertEquals(2, ledgerEntry.getEntryBuffer().refCnt()));
+        entryIterator.forEachRemaining(ledgerEntry -> ledgerEntry.getEntryBuffer().release());
```
sijie: I think this will not be tested. `forEachRemaining` will consume the iterator.

you might need change this to 

```
entryIterator.forEachRemaining(ledgerEntry -> {
    assertEquals(2, ledgerEntry.getEntryBuffer().refCnt());
    ledgerEntry.getEntryBuffer().release();
    assertEquals(1, ledgerEntry.getEntryBuffer().refCnt());
})
```

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
```diff
@@ -0,0 +1,52 @@
...
+     */
+    Iterator<LedgerEntry> iterator();
+
+    /**
+     * In this, It retains the ByteBuf references for the entries in this LedgerEntries.
+     * The caller who calls {@link #retainIterator()} is responsible for releasing the retained references.
+     *
+     * @return the iterator of type LedgerEntry that has been retained
+     */
+    Iterator<LedgerEntry> retainIterator();
```
ivankelly: Why? Why not just state that that for iterator(), the LedgerEntrys are owned by the LedgerEntries instance, and if you want to retain the entry beyond the lifetime of the LedgerEntries instance, you should call retain on the entry manually.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
```diff
@@ -0,0 +1,52 @@
...
+     * Gets a specific LedgerEntry by entryId.
+     *
+     * @param entryId the LedgerEntry id
+     * @return the LedgerEntry, null if no LedgerEntry with such entryId.
+     */
+    LedgerEntry getEntry(long entryId);
+
+    /**
+     * In this, It does not retain the ByteBuf references for the entries in this LedgerEntries.
+     * The caller who calls {@link #iterator()} should be careful for not releasing the references.
```
ivankelly: I think the documentation should view this from a different angle. The LedgerEntries are owned by the LedgerEntries instance. When returned by the iterator, they are still owned by the LedgerEntries instance, so noone else should release them. Only if they are first retained, should they be released by the caller.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/impl/LedgerEntriesImplTest.java
ivankelly: +1

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/impl/LedgerEntriesImplTest.java
jiazhai: Thanks. will change this

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
sijie: @ivankelly it was me requesting adding this method. it is following the pattern that ByteBuf has for creating a `slice` of the buffer. slice: https://netty.io/4.0/api/io/netty/buffer/ByteBuf.html#slice-- and retainedSlice https://netty.io/4.1/api/io/netty/buffer/ByteBuf.html#retainedSlice--.  This provides a shortcut for callers that want to retain refcnt for all the entries.

sijie: that was my review comment - https://github.com/apache/bookkeeper/pull/727#discussion_r151218118

jiazhai:
@ivan, Thanks for the comments, the reason is described  in  the issue, will move it here.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
jiazhai: thanks, will change it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
jiazhai: Thanks, @ivankelly @sijie , will put some of your comments here.

jiazhai:
@ivankelly @sijie  updated.

ivankelly:
@zhaijack it's good to have it here too, so that when we look at the git log, we don't have to jump back to the issue to figure stuff out. It's often the case also that the final patch diverges significantly from the original issue, so it's good to capture that and summarize at the end.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/impl/LedgerEntriesImplTest.java
jiazhai: oh. not noticed this one before.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
ivankelly: @sijie Ok. I'm ok with it, but I don't think it's needed. Your original comment seems to imply that the iterator itself owned the ref count of the entries, but I think it's better to view it as the LedgerEntries owning it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
```diff
@@ -35,10 +35,15 @@
     LedgerEntry getEntry(long entryId);
 
     /**
-     * In this, It does not retain the ByteBuf references for the entries in this LedgerEntries.
+     * In this method, It does not increment the reference counts of ByteBuf for the entries in this LedgerEntries.
```
ivankelly: This method does not increment the reference count of ByteBuf ...

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
```diff
@@ -35,10 +35,15 @@
...
      * The caller who calls {@link #iterator()} should be careful for not releasing the references.
      *
-     * @return the iterator of type LedgerEntry
+     * when iterator is called, you are handing out the entries, you may not know when the caller will
```
ivankelly: I don't think this is correct. You're not handing out the entries. You are handing out an iterator which will iterate over the set of entries held by the LedgerEntries. I don't think refcounting should come into the iterator part of it at all.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
sijie: sure. if we make a clear comment on `iterator()`, I don't think retainIterator is needed. However if a caller wants to keep an iterator beyond the lifecycle of `LedgerEntries`, then a retainedIterator() is required. because there is no other simpler way to do that. So I would suggest us following the ByteBuf pattern to provide convenient methods for caller to uses. However I am not going to make a strong case for it. I am fine with removing it if you feel strong about that.

ivankelly: I think keeping the iterator beyond the life of a LedgerEntries will be an uncommon usecase, and not something we should account for in the main api. Instead we could provide a utility method which takes a LedgerEntries and returns a Collection<LedgerEntry> with the refcounts bumped. I'm not sure why you would do this though, and not just keep the LedgerEntries.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
sijie: @ivankelly If you think this as `ByteBuf.slice()`, you are handing out a slice of the bytebuffer, you need to declare what would be the behavior for referencing count. Personally I feel having `iterator` and `retainIterator` will make a clear behavior what callers would expect from the interfaces. This is not a `correctness` debate here, it is more about what should be the reference counting semantic for the callers.

ivankelly: I didn't mean "correct" in the same way we usually talk about "correctness" :).

Rather, I was arguing semantics. I think the reference count stays with the LedgerEntries object.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
sijie: why people need retainSlice for a bytebuf? A slice is essentially an iterator for the shared buffer. 

There are use cases for this - for example, if you are building a service, which would cache LedgerEntries, there might be multiple readers will read those cached LedgerEntries, they would need different iterators on iterating the entries. The readers would retain the iterator, and release after they finish iteration. The cache will release its held reference when a LedgerEntries is evicted from the cache. A correct reference counting behavior should be provided when the caller attempts to create iterators.

If you just only provide iterator(), you have to do two rounds of iterations. First round, you get an iterator to retrieve ledger entries, retain the reference count and add those entries to a list, then return a brand new iterator. Second round, your application will iterate over the new iterator. This introduces unnecessary complexity.

again, if you are just viewing this from a single user application perspective, you don't need `retainIterator()`. but if you are thinking this from a shared use case, an iterator for a LedgerEntries is essentially a `slice` for a ByteBuf. The reasons why ByteBuf needs a `retainedSlice()` apply here.

again, I see this can simplify a lot of things for applications. However I am not going to make a strong case here. I don't see it is a correctness debate, it is more a convenient util function. If you feel strong about that, I am fine with removing it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
sijie: sure. "reference count" should goes with how caller uses it, not simply stays with LedgerEntries object. that is the semantic problem, you need to take account on how applications use it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
ivankelly: > If you just only provide iterator(), you have to do two rounds of iterations. First round, you get an iterator to retrieve ledger entries, retain the reference count and add those entries to a list, then return a brand new iterator. Second round, your application will iterate over the new iterator. This introduces unnecessary complexity.

Doesn't retainIterator need to iterate over the LedgerEntries anyhow to increase the ref counts.

> it is more a convenient util function

Which is why I think it would fit better in a Util class. 

```
class LedgerEntriesUtils {
    public static Iterator<LedgerEntry> retainIterator(Iterator<LedgerEntry> iterator);
}
```

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
ivankelly: The reference count that a LedgerEntry has before iterator() is called belongs to LedgerEntries, because that is referring to the fact that LedgerEntries own those objects.

With retainedIterator there is two options. a) the ownership of the reference is moved to the Iterator (like move semantics in c++) or b) a new Iterator object with a new reference count on the LedgerEntry is created.

a) is a move, b) is a copy (albeit shallow).

If retainedIterator is part of the API I would prefer the a) semantic, but this would imply that retainedIterator can only be called once. If it's b) I would prefer it be done with a Utility method. 

And if retainedIterator is kept, rename it to retainingIterator. It should be an adjective.

sijie: @ivankelly it is not a `move`. `retain` is to increment the reference count. there is no move here, as the caller don't really know how many reference counts are referencing the underneath buffer. 

The whole idea behind `retainedIterator` is following what `ByteBuf` has and provide exactly same behavior as what people would expect from `ByteBuf`. There is already some sort of reference counting mechanisms that bookkeeper is using. We can just follow what it already has and have a consistent naming, semantic and behavior as `ByteBuf`. Consistent naming and semantic is more important to me.

retain - https://netty.io/4.1/api/io/netty/buffer/ByteBuf.html#retain--
retainedSlice - https://netty.io/4.1/api/io/netty/buffer/ByteBuf.html#retainedSlice-int-int-

Anyway, let's move forward without `retainedIterator`. It is not a blocker for 4.6.0. We can circle back in future if it is needed.

ivankelly: > Anyway, let's move forward without retainedIterator. It is not a blocker for 4.6.0. We can circle back in future if it is needed.

Let's leave it out for now then. I think with APIs it's better to leave it out if in doubt. It's always easier to add calls to an API than remove them.

Regarding retain vs retained vs retaining, in the case of ByteBuf, it is the bytebuf itself that has the reference count, so it makes sense to "retain" a bytebuf and have a "retained" bytebuf. However, in the case of an iterator, it's not the iterator that is being retained. Rather the iterator is just a container for objects that are retained, so in some sense the iterator is retaining the objects, though that's not perfect either.

Another thing I thought about with the retainedIterator is that error handling becomes awkward. If while iterating an error occurs, you still need to iterate to the end to free all refcounts, which isn't immediately obvious from the api.

In summary, there seems to be a good few sharp corners on this call, so best to leave it out for now.

jiazhai: Thanks @ivankelly and @sijie for your comments. 
I will remove it now, and think of it later when it is wanted.

eolivelli:
I am taking last look

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
```diff
@@ -35,24 +35,11 @@
     LedgerEntry getEntry(long entryId);
 
     /**
-     * In this method, It does not increment the reference counts of ByteBuf for the entries in this LedgerEntries.
+     * This method does not increment the reference count of ByteBuf for the entries in this LedgerEntries.
```
ivankelly: Calling this methods does not modify the reference count of the ByteBuf in the returned LedgerEntry objects.
The caller who calls {@link #iterator()} should make sure that they do not call release on the LedgerEntry objects to avoid a double free. All reference counts will be decremented when the containing LedgerEntries object is closed.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
```diff
@@ -35,24 +35,11 @@
...
      * The caller who calls {@link #iterator()} should be careful for not releasing the references.
-     *
-     * when iterator is called, you are handing out the entries, you may not know when the caller will
-     * complete iterating the entries. In this case, please use {@link #retainedIterator()}, it will increment
-     * the refCnt, and the application who is holding iterator should be responsible for releasing one refCnt
-     * that {@link #retainedIterator()} retains for entries.
+     * The implementation of this interface will release all the Entries ByteBuf reference when close.
```
ivankelly: Remove this sentence if you add what I put above.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/api/BookKeeperApiTest.java
```diff
@@ -275,11 +276,13 @@ public void testCannotDeleteLedgerTwice() throws Exception {
...
 
-    private static void checkEntries(Iterable<LedgerEntry> entries, byte[] data)
+    private static void checkEntries(LedgerEntries entries, byte[] data)
         throws InterruptedException, BKException {
-        for (LedgerEntry le : entries) {
-            assertArrayEquals(data, le.getEntry());
+        Iterator<LedgerEntry> iterator = entries.iterator();
+        while(iterator.hasNext()) {
+            LedgerEntry entry = iterator.next();
+            // since getEntry() will call ByteBuf release, so no need release here.
```
eolivelli: this is clear but maybe it is better to call always call "close".
two motivations:
- close lets the reclycler to work, maybe it is not a real issue
- please use simple test cases as this as a code example

please take into account this comment but it is not a real problem

jiazhai: Thanks. will change it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
jiazhai: will change it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
jiazhai: will change it.

jiazhai:
seems, it is not able to merge with merge script, it reported:
```
bookkeeper git:(master) ✗ python dev/bk-merge-pr.py
git rev-parse --abbrev-ref HEAD
Which pull request would you like to merge? (e.g. 34): 727
No approvals found in this pull request
git rev-parse --abbrev-ref HEAD
git branch
``` 

the below message is to test again:
+1, lgtm.

Issue #726, opened 2017-11-15
Unit Tests failure with BKMetadataVersionException


jiazhai:
**BUG REPORT**

1. Please describe the issue you observed:

- What did you do?
Run mvn test
- What did you expect to see?
Expect all tests passed
- What did you see instead?
some tests failed, with Exception: BKMetadataVersionException
Failure cases:
```
  BookieRecoveryTest.testMetadataConflictWhenDelayingEnsembleChange » BKMetadataVersion
  BookieRecoveryTest.testMetadataConflictWithRecovery » BKMetadataVersion Bad le...
  BookieRecoveryTest.testSyncBookieRecoveryToRandomBookiesCheckForDupes » BKMetadataVersion
  TestDelayEnsembleChange.testChangeEnsembleSecondBookieReadOnly » BKMetadataVersion
  TestWatchEnsembleChange.testWatchEnsembleChange[1] » BKMetadataVersion Bad led...
  UpdateLedgerCmdTest.testUpdateLedgersToHostname » BKMetadataVersion Bad ledger...
  UpdateLedgerOpTest.testChangeEnsembleAfterRenaming:202 » BKMetadataVersion Bad...
  UpdateLedgerOpTest.testLimitLessThanTotalLedgers » BKMetadataVersion Bad ledge...
  UpdateLedgerOpTest.testManyLedgers » BKMetadataVersion Bad ledger metadata ver...
  UpdateLedgerOpTest.testRenameWhenAddEntryInProgress » BKMetadataVersion Bad le...
```

get one stack example from TestWatchEnsembleChange.testWatchEnsembleChange here:
```
 2017-11-15 13:00:11,838 - ERROR - [BookKeeperClientWorker-OrderedSafeExecutor-0-0:PendingAddOp@354] - Write of ledger entry to quorum failed: L0 E10
 org.apache.bookkeeper.client.BKException$BKMetadataVersionException: Bad ledger metadata version
     at org.apache.bookkeeper.client.SyncCallbackUtils.finish(SyncCallbackUtils.java:68)
     at org.apache.bookkeeper.client.SyncCallbackUtils$SyncAddCallback.addComplete(SyncCallbackUtils.java:264)
     at org.apache.bookkeeper.client.PendingAddOp.submitCallback(PendingAddOp.java:359)
     at org.apache.bookkeeper.client.LedgerHandle.errorOutPendingAdds(LedgerHandle.java:1300)
[...]

PR #730, opened 2017-11-15
Issue 726: Unit Tests failure with BKMetadataVersionException

sijie:

Problem:

The problem is introduced after making storing ctime optional. the verification become inconsistent because the way how ctime was assigned and compared. It causes the test cases failing in #726 

Solution:

This change is to change `isConflict` to ignore comparing ctime if client disables storing system time as ctime.


asfgit:
FAILURE
 
--none--

asfgit:
FAILURE
 
--none--

PR #735, opened 2017-11-16
Issue 734: bookkeeper-benchmark is broken after protobuf files are moved to bookkeeper-proto

sijie:

Problem:

The issue was introduced after protobuf files are moved to bookkeeper-proto. so the protobuf is not shaded correctly.
So in bookkeeper-benchmark, it is conflicted with the protobuf version introduced by hadoop.

Solution:

- update the bookkeeper-server pom.xml to shade protobuf correctly
- remove hadoop dependency from bookkeeper-benchmark.


asfgit:
FAILURE
 
--none--

PR #738, opened 2017-11-16
Issue 731: refine LedgerEntry interface and implementation

zhaijack:
In LedgerEntry interface, getEntry() and getEntryBuffer() have completely different behaviours. getEntry() releases bytebuf automatically, while getEntryBuffer() returns the bytebuf (if you don't call getEntry, you are responsible for releasing the entry buffer.

In this change:
make getEntry() reenter-able; make LedgerEntry implement AutoCloseable; LedgerEntry is responsible for releasing the bytebuf it is holding.

1.  add `close` and `duplicate` in `api.LedgerEntry`,
1.  LedgerEntryImpl implements `api.LedgerEntry`,
1.  client.LedgerEntry doesn't implement the new api. it wraps an `api.LedgerEntry`,  
1.  add `close` in api.LastConfirmedAndEntry."
1.  fix build and test errors.

> ---
> quickly and easily:
> 
> 
> ---

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerEntry.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerEntry.java
@@ -23,8 +23,8 @@ package org.apache.bookkeeper.client;
 import com.google.common.base.Preconditions;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufInputStream;
-
 import java.io.InputStream;
+import org.apache.bookkeeper.client.impl.LedgerEntryImpl;
 import org.apache.bookkeeper.conf.ClientConfiguration;
 
 /**
@@ -32,30 +32,28 @@ import org.apache.bookkeeper.conf.ClientConfiguration;
  * the entry content.
  *
  */
-public class LedgerEntry
-    implements org.apache.bookkeeper.client.api.LedgerEntry {
+public class LedgerEntry {
 
     final long ledgerId;
-    long entryId;
-    long length;
+    final long entryId;
+    final long length;
     ByteBuf data;
 
-    LedgerEntry(long lId, long eId) {
-        this.ledgerId = lId;
-        this.entryId = eId;
+    LedgerEntry(LedgerEntryImpl entry) {
+        this.ledgerId = entry.getLedgerId();
+        this.entryId = entry.getEntryId();
+        this.length = entry.getLength();
+        this.data = entry.getEntryBuffer().retain();
     }
 
-    @Override
     public long getLedgerId() {
         return ledgerId;
     }
 
-    @Override
     public long getEntryId() {
         return entryId;
     }
 
-    @Override
     public long getLength() {
         return length;
     }
@@ -68,7 +66,6 @@ public class LedgerEntry
      * @return the content of the entry
      * @throws IllegalStateException if this method is called twice
      */
-    @Override
     public byte[] getEntry() {
         Preconditions.checkState(null != data, "entry content can be accessed only once");
         byte[] entry = new byte[data.readableBytes()];
@@ -105,7 +102,6 @@ public class LedgerEntry
      * @throws IllegalStateException if the entry has been retrieved by {@link #getEntry()}
      * or {@link #getEntryInputStream()}.
      */
-    @Override
     public ByteBuf getEntryBuffer() {
         Preconditions.checkState(null != data, "entry content has been retrieved" +
             " by #getEntry or #getEntryInputStream");
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntry.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntry.java
@@ -23,7 +23,6 @@ package org.apache.bookkeeper.client.api;
 import io.netty.buffer.ByteBuf;
 import org.apache.bookkeeper.common.annotation.InterfaceAudience.Public;
 import org.apache.bookkeeper.common.annotation.InterfaceStability.Unstable;
-import org.apache.bookkeeper.conf.ClientConfiguration;
 
 /**
  * An entry.
@@ -32,7 +31,7 @@ import org.apache.bookkeeper.conf.ClientConfiguration;
  */
 @Public
 @Unstable
-public interface LedgerEntry {
+public interface LedgerEntry extends AutoCloseable {
 
     /**
      * The id of the ledger which contains the entry.
@@ -56,25 +55,31 @@ public interface LedgerEntry {
     long getLength();
 
     /**
-     * Returns the content of the entry. This method can be called only once. While using v2 wire protocol this method
-     * will automatically release the internal ByteBuf.
+     * Returns the content of the entry.
      *
      * @return the content of the entry
-     * @throws IllegalStateException if this method is called twice
      */
     byte[] getEntry();
 
     /**
      * Return the internal buffer that contains the entry payload.
      *
-     * <p>Note: Using v2 wire protocol it is responsibility of the caller
-     * to ensure to release the buffer after usage.
-     *
      * @return a ByteBuf which contains the data
-     *
-     * @see ClientConfiguration#setNettyUsePooledBuffers(boolean)
-     * @throws IllegalStateException if the entry has been retrieved by {@link #getEntry()}
      */
     ByteBuf getEntryBuffer();
+
+    /**
+     * Returns a duplicate of this entry.
+     *
+     * <p>This call will retain a slice of the underneath byte buffer.
+     *
+     * @return a duplicated ledger entry.
+     */
+    LedgerEntry duplicate();
+
+    /**
+     * {@inheritDoc}
+     */
+    void close();
 
 }
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/impl/LedgerEntryImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/impl/LedgerEntryImpl.java
@@ -1,0 +1,155 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.bookkeeper.client.impl;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.ByteBufUtil;
+import io.netty.util.Recycler;
+import io.netty.util.Recycler.Handle;
+import io.netty.util.ReferenceCountUtil;
+import org.apache.bookkeeper.client.api.LedgerEntry;
+
+/**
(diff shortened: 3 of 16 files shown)


asfgit:
FAILURE
 
--none--

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/SyncCounter.java
```diff
@@ -22,6 +22,7 @@
 package org.apache.bookkeeper.client;
 
 import java.util.Enumeration;
+import org.apache.bookkeeper.client.api.LedgerEntry;
```
eolivelli: this seems not necessary

asfgit:
FAILURE
 
--none--

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/SyncCounter.java
eolivelli: Sync Counter is not used by current code, it is here only for backward compatibility.
Did you run all rests, even compatibility?

eolivelli: Sync Counter is not used by current code, it is here only for backward compatibility.
Did you run all rests, even compatibility?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/impl/LedgerEntryImpl.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    @Override
+    public ByteBuf getEntryBuffer() {
+        return entryBuf;
+    }
+
+    @Override
+    public LedgerEntryImpl duplicate() {
+        LedgerEntryImpl other = RECYCLER.get();
```
sijie: this can be simplified with 

```
return duplicate(this);
```

bookkeeper-server/src/test/java/org/apache/bookkeeper/sasl/GSSAPIBookKeeperTest.java
```diff
@@ -1,13 +1,39 @@
...
+import org.junit.After;
+import org.junit.AfterClass;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /*
```
sijie: since you are touching this file, you should move the LICENSE header to the beginning of this file.

bookkeeper-server/src/test/java/org/apache/bookkeeper/sasl/MD5DigestBookKeeperTest.java
```diff
@@ -1,9 +1,27 @@
...
+import org.apache.bookkeeper.conf.ClientConfiguration;
+import org.apache.bookkeeper.conf.ServerConfiguration;
+import org.apache.bookkeeper.proto.BookieServer;
+import org.apache.bookkeeper.test.BookKeeperClusterTestCase;
+import org.junit.AfterClass;
+import org.junit.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /*
```
sijie: same as above

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerEntry.java
```diff
@@ -1,114 +0,0 @@
...
-
-import java.io.InputStream;
-import org.apache.bookkeeper.conf.ClientConfiguration;
-
-/**
- * Ledger entry. Its a simple tuple containing the ledger id, the entry-id, and
- * the entry content.
- *
- */
-public class LedgerEntry
```
sijie: I think you still need to keep this class for BC issue.

you can extend it from LedgerEntryImpl and override the getEntry() and keep getEntryInputStream(). because getEntry and getEntryInputStream should stay same semantics as legacy semantic in this class.

```
public class LedgerEntry extends LedgerEntryImpl {

     

}
```

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/SyncCounter.java
jiazhai: Thanks. will re-run the compatibility tests.

bookkeeper-server/src/test/java/org/apache/bookkeeper/sasl/GSSAPIBookKeeperTest.java
jiazhai: thanks, will change it. Seems the comments of header has some problem, and make the ide not recognise it.

bookkeeper-server/src/test/java/org/apache/bookkeeper/sasl/MD5DigestBookKeeperTest.java
jiazhai: Got it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/impl/LedgerEntryImpl.java
jiazhai: Thanks. will change it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/SyncCounter.java
jiazhai: re-checked, the BC tests all passed.

asfgit:
FAILURE
 
--none--

asfgit:
SUCCESS 
 
--none--

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerEntry.java
```diff
@@ -23,41 +23,37 @@
...
 
 /**
  * Ledger entry. Its a simple tuple containing the ledger id, the entry-id, and
  * the entry content.
  *
+ * @Deprecated since 4.6.0, in favor of using {@link org.apache.bookkeeper.client.impl.LedgerEntryImpl}.
  */
-public class LedgerEntry
-    implements org.apache.bookkeeper.client.api.LedgerEntry {
+@Deprecated
```
sijie: don't add `@Deprecated` at this release.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/AsyncCallback.java
```diff
@@ -17,6 +17,7 @@
 package org.apache.bookkeeper.client;
 
 import java.util.Enumeration;
+import org.apache.bookkeeper.client.api.LedgerEntry;
```
sijie: Oh I think there is a problem here - this is the legacy api, you can't change this to `api.LedgerEntry`. try remove this line and recompile, you will find it will not compile. I think `client.LedgerEntry` should extend `impl.LedgerEntryImpl`. The current approach for `client.LedgerEntry` doesn't really work.

jiazhai: I C. need changes

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerEntry.java
jiazhai: get it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntry.java
```diff
@@ -56,11 +56,9 @@
...
 
     /**
-     * Returns the content of the entry. This method can be called only once. While using v2 wire protocol this method
-     * will automatically release the internal ByteBuf.
+     * Returns the content of the entry.
      *
      * @return the content of the entry
-     * @throws IllegalStateException if this method is called twice
      */
     byte[] getEntry();
```
ivankelly: Returning a ByteBuf should be consistent whether using the V2 wire protocol or not. As it is now, the documentation for getEntryBuffer is super confusing.

ivankelly: As in, are we returning ownership of the object? Or does the LedgerEntry own the buffer, and is this why it requires closing?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntry.java
```diff
@@ -32,7 +32,7 @@
  */
 @Public
 @Unstable
-public interface LedgerEntry {
+public interface LedgerEntry extends AutoCloseable {
```
ivankelly: Should probably be a different commit, but there is an inconsistency in the data types used in read and write.

WriteHandle is takes ByteBuf and ByteBuffer
ReadHandle/LedgerEntry returns ByteBuf and byte[]

This should be resolved.

jiazhai:
@ivankelly , thanks for the comments, as @sijie pointed out, we may need keep old behavior for LedgerEntry, so may need to change this PR. will re-think and do the change later.

ivankelly:
What is the old behaviour for LedgerEntry? The javadoc leads me to believe it's different depending on whether v2 protocol is used or not.

sijie:
@ivankelly that is the 4.5 behavior of LedgerEntry. in the new api, we should change LedgerEntry to be a more consistent behavior.

jiazhai:
updated this PR to keep old behavior of LedgerEntry; the tests that use client.LedgerEntry was un -touched.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntry.java
jiazhai: #750 opened for it.

sijie: @ivankelly : a good point. it was inconsistent there because I asked @eolivelli to not use `byte[]`. but @eolivelli wants to have a `byte[]` array. that was probably what caused confusion in the initial commit. we can support all - `byte[]`, `ByteBuffer` and `ByteBuf`.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntry.java
jiazhai: In LedgerEntryImpl, LedgerEntry is reenterable,  closing is more for the ByteBuf release.

sijie:
@ivankelly @eolivelli  can you guys review this one again? since you reviewed that before.

jiazhai:
Oh, sorry, not noticed your comments @sijie , already did the merge. But if there is any comment, I could resolved it in later PRs.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntry.java
```diff
@@ -56,25 +55,31 @@
...
-     * Returns the content of the entry. This method can be called only once. While using v2 wire protocol this method
-     * will automatically release the internal ByteBuf.
+     * Returns the content of the entry.
      *
      * @return the content of the entry
-     * @throws IllegalStateException if this method is called twice
      */
     byte[] getEntry();
 
     /**
```
ivankelly: What happens to the ownership of the returned ByteBuf? Is the caller responsible for releasing it? @sijie @zhaijack

sijie: @ivankelly - This call doesn't retain any reference of the underneath bytebuf. if a caller releases the entry by #close(), it will release the underneath bytebuf. If the caller want to use the bytebuf beyond the lifecycle of this entry, it has to retain the buf by itself and then be responsible for releasing its retained refcnt. I have improved the comment in my subsequent change at #755 . 

https://github.com/apache/bookkeeper/pull/755/files

ivankelly: ah, excellent, will check it out there.

PR #740, opened 2017-11-16
Issue 739: Move the implementation of http services to be under server package

sijie:

bookkeeper-http module provides the framework for http admin rest api. it is using 'org.apache.bookkeeper.http' and 'org.apache.bookkeeper.http.service. the http package inbookkeeper-server` is reusing these two packages. it is not good for two modules sharing same package name, it is going to be confusing javadoc.


eolivelli:
And it won't work on java9 after modularization: a package can live only inside one jar

asfgit:
FAILURE
 
--none--

sijie:
@eolivelli I made this for 4.6. because http server is the new feature, I would like it get fixed before release, rather than letting this structure layout get out in 4.6.

sijie:
please retest this

sijie:
retest this please

asfgit:
FAILURE
 
--none--

PR #746, opened 2017-11-19
Issue-732 Update Dockerfile for 4.5.1 version

eolivelli:
Bump BK version in Docker file and set release manager's PGP key used to sign the artifacts


eolivelli:
@jiazhai could please test the Docker file in your env ?

asfgit:
FAILURE
 
--none--

jiazhai:
sure, I will do it.

jiazhai:
Seems there was an issue: the default statsProviderClass is org.apache.bookkeeper.stats.prometheus.PrometheusMetricsProvider, but PrometheusMetricsProvider is not included in server-bin.tar.gz, so when bookie start it will report error like this -- 
```
ERROR - [main:BookieServer@435] - Exception running bookie server : 
org.apache.commons.configuration.ConfigurationException: org.apache.commons.configuration.ConfigurationException: java.lang.ClassNotFoundException: org.apache.bookkeeper.stats.PrometheusMetricsProvider
```
have changed the stats to use NullStatsProvider in above commit.

Now it passed the tests in my env.

Since we have # 499 creating a -all package, we will use -all package for 4.6.0 docker image.

Issue #753, opened 2017-11-22
Allow option to disable data sync on journal


merlimat:
For deployments where a RAID battery-backed cache is not available and where durability is not a requirements, we should allow bookies to rely on page cache and relax durability.

This change adds a system-wide options. All bookies are expected to have the same setting. 
With relaxed durability, a user should expect that some amount of data can be loss in case of power failures or kernel panics. 

This change is kind of similar to the proposal for BP-14, though not in contrast with it. In BP-14, there is a fine grained approach to durability with the option to flush the ledger. Here we're adding the option to turn it off cluster-wide.

The rational around this setting is that: 
 * Fsync performance is thighly related to the hardware choice. BookKeeper can perform very well with strong durability on commodity hardware, provided the hardware setup is good. 
 * Typically, choosing the right hardware boils down to either:
    * Use a RAID controller with battery-backed write cache (this is very common in most RAID controllers). In this case the write cache is able to group together multiple IOps and radically increase the throughput and lower the latency, while maintaining the same degree of durability (in case of power loss, the write cache gets flushed on disk using the battery power). 
    * Use a "fast" SSD for bookie journal. There's a huge variance on SSDs performance. Some guy from Ceph community put together a list of drives with perf numbers (http://www.sebastien-han.fr/blog/2014/10/10/ceph-how-to-test-if-your-ssd-is-suitable-as-a-journal-device/). Even for "enterprise" models, performance on fsync heavy loads is all over the place.
 * In some cases, a user might not have the luxury of choosing the right hardware setup. This is typically the case in cloud environments where the hardware options are fixed.
 * If a user doesn't require strong durability for its use case, and can afford to lose data in some scenario, we should allow him to configure the bookies in this way.

PR #754, opened 2017-11-22
Issue #753: Allow option to disable data sync on journal

merlimat:
Context in #753 

For deployments where a RAID battery-backed cache is not available and where durability is not a requirements, we should allow bookies to rely on page cache and relax durability.

### DIFF
--- a/bookkeeper-server/conf/bk_server.conf
+++ b/bookkeeper-server/conf/bk_server.conf
@@ -296,6 +296,10 @@ journalDirectory=/tmp/bk-txn
 # Should we remove pages from page cache after force write
 # journalRemoveFromPageCache=false
 
+# Should the data be fsynced on journal before acknowledgment
+# Default is 'true'
+# journalSyncData=true
+
 # Should we group journal force writes, which optimize group commit
 # for higher throughput
 # journalAdaptiveGroupWrites=true
--- a/bookkeeper-server/conf/bk_server.conf
+++ b/bookkeeper-server/conf/bk_server.conf
@@ -296,8 +296,12 @@ journalDirectory=/tmp/bk-txn
 # Should we remove pages from page cache after force write
 # journalRemoveFromPageCache=false
 
-# Should the data be fsynced on journal before acknowledgment
-# Default is 'true'
+# Should the data be fsynced on journal before acknowledgment.
+# By default, data sync is enabled to guarantee durability of writes.
+# Beware: while disabling data sync in the Bookie journal might improve the bookie write performance, it will also
+# introduce the possibility of data loss. With no sync, the journal entries are written in the OS page cache but
+# not flushed to disk. In case of power failure, the affected bookie might lose the unflushed data. If the ledger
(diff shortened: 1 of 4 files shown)


sijie:
@jvrao @eolivelli please take a look about this since it is related to BP-14.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -914,9 +918,25 @@ public void run() {
...
+                                    cbThreadPool.execute((QueueEntry) toFlush.get(i));
+                                }
+
+                                toFlush.clear();
+                                if (shouldRolloverJournal) {
+                                    forceWriteRequests.put(new ForceWriteRequest(logFile, logId, lastFlushPosition,
+                                            new LinkedList<>(), shouldRolloverJournal, false));
+                                }
+                            }
+
```
jvrao: Why not simply move the logic into JournalChnnel:forceWrite()?
I haven't really thought through any issues that may come up with the durability guarantees that were built into the system.

eolivelli: I think this impl is good as it does not change other logics.
**We MUST add a strong advice that bookies with this flag are very dangerous as they does not guarantee a repeatable behaviour upon restart.**
They will eventually lose "fence bits" and will not guarantee the correctes of the LAC protocol, as writes are not guaranteed to have been really persisted to storage.

bookkeeper-server/conf/bk_server.conf
```diff
@@ -296,6 +296,10 @@ journalDirectory=/tmp/bk-txn
 # Should we remove pages from page cache after force write
 # journalRemoveFromPageCache=false
 
+# Should the data be fsynced on journal before acknowledgment
```
eolivelli: we need a warning here

sijie: sure. I would suggest also marking it as `@experimental`

sijie: fyi, I created an issue #756 (we need some guidelines on providing guidelines for a feature introduced in a release).

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
merlimat: > Why not simply move the logic into JournalChnnel:forceWrite()?

One reason is simplicity (the logic in the force-sync thread and request is already complicated, and I didn't want to add more in there). 

The other is to avoid one additional context switch before sending the ack. If we say that the write is "good" (for whatever definition of "good" is configured in the bookie), then we should acknowledge immediately.

bookkeeper-server/conf/bk_server.conf
merlimat: Sure, I'll add more explicit docs on the risks of enabling the flag.

merlimat:
@eolivelli Updated the docs for the flag. PTAL

PR #755, opened 2017-11-22
Issue 750: support ByteBuf, ByteBuffer, byte[] in both WriteHandle, WriteAdvHandle, ReadHandle

sijie:

- WriteHandle supports append byte[], ByteBuf and ByteBuffer
- WriteAdvHandle supports write byte[], ByteBuf and ByteBuffer
- LedgerEntry returns byte[], ByteBuffer and ByteBuf

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntry.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntry.java
@@ -26,11 +26,11 @@ import org.apache.bookkeeper.common.annotation.InterfaceAudience.Public;
 import org.apache.bookkeeper.common.annotation.InterfaceStability.Unstable;
 
 /**
- * An entry.
+ * An entry in a ledger.
  *
- * <p>The entry implementation might be holding references to byte buffers under the hood. The users holding the
+ * <p>The entry implementation may hold references to byte buffers under the hood. The users holding the
  * references to the instances of this class, are responsible for calling {@link LedgerEntry#close()} to release
- * resources held by the instances.
+ * resources held by the entry instances.
  *
  * @since 4.6
  */
@@ -60,26 +60,26 @@ public interface LedgerEntry extends AutoCloseable {
     long getLength();
 
     /**
-     * Returns the content of the entry into a byte array.
+     * Returns the content of the entry as a byte array.
      *
      * @return the content of the entry
      */
-    byte[] getEntry();
+    byte[] getEntryBytes();
 
     /**
      * Exposes this entry's data as an NIO {@link ByteBuffer}. The returned buffer
-     * shares the content with this underneath bytebuf (which you can get it by {@link #getEntryBuffer()}),
-     * while changing the position and limit of the returned NIO buffer does not affect the indexes and
+     * shares the content with this underneath bytebuf (which you can get it by {@link #getEntryBuffer()}).
+     * Changing the position and limit of the returned NIO buffer does not affect the indexes and
      * marks of this underneath buffer.  This method is identical
      * to {@code entry.getEntryBuffer().nioBuffer()}. This method does not
-     * modify {@code readerIndex} or {@code writerIndex} of the underneath bytebuf.
+     * modify {@code readerIndex} or {@code writerIndex} of the underlying bytebuf.
      */
-    ByteBuffer getNioBuffer();
+    ByteBuffer getEntryNioBuffer();
 
     /**
-     * Return the internal buffer that contains the entry payload.
+     * Return the internal {@link ByteBuf} that contains the entry payload.
      *
-     * <p>This call doesn't retain any reference on the underneath bytebuf. If you want to use the bytebuf
+     * <p>This call doesn't change the reference count on the returned bytebuf. If you want to use the bytebuf
      * after the entry is released (via {@link #close()}, the caller must retain the references of the bytebuf.
      *
      * @return a ByteBuf which contains the data
@@ -91,7 +91,7 @@ public interface LedgerEntry extends AutoCloseable {
      *
      * <p>This call will retain a slice of the underneath byte buffer.
      *
-     * @return a duplicated ledger entry.
+     * @return a duplicated ledger entry
      */
     LedgerEntry duplicate();
 
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntry.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntry.java
@@ -80,7 +80,7 @@ public interface LedgerEntry extends AutoCloseable {
      * Return the internal buffer that contains the entry payload.
      *
      * <p>This call doesn't retain any reference on the underneath bytebuf. If you want to use the bytebuf
-     * after the entry is released (via {@link #close()}, please retain the references of the bytebuf.
+     * after the entry is released (via {@link #close()}, the caller must retain the references of the bytebuf.
      *
      * @return a ByteBuf which contains the data
      */
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntry.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntry.java
@@ -21,11 +21,16 @@
 package org.apache.bookkeeper.client.api;
 
 import io.netty.buffer.ByteBuf;
+import java.nio.ByteBuffer;
 import org.apache.bookkeeper.common.annotation.InterfaceAudience.Public;
 import org.apache.bookkeeper.common.annotation.InterfaceStability.Unstable;
 
 /**
  * An entry.
+ *
+ * <p>The entry implementation might be holding references to byte buffers under the hood. The users holding the
+ * references to the instances of this class, are responsible for calling {@link LedgerEntry#close()} to release
+ * resources held by the instances.
  *
  * @since 4.6
  */
@@ -55,14 +60,27 @@ public interface LedgerEntry extends AutoCloseable {
     long getLength();
 
     /**
-     * Returns the content of the entry.
+     * Returns the content of the entry into a byte array.
      *
      * @return the content of the entry
      */
     byte[] getEntry();
 
     /**
+     * Exposes this entry's data as an NIO {@link ByteBuffer}. The returned buffer
+     * shares the content with this underneath bytebuf (which you can get it by {@link #getEntryBuffer()}),
+     * while changing the position and limit of the returned NIO buffer does not affect the indexes and
+     * marks of this underneath buffer.  This method is identical
+     * to {@code entry.getEntryBuffer().nioBuffer()}. This method does not
+     * modify {@code readerIndex} or {@code writerIndex} of the underneath bytebuf.
+     */
+    ByteBuffer getNioBuffer();
+
+    /**
      * Return the internal buffer that contains the entry payload.
+     *
+     * <p>This call doesn't retain any reference on the underneath bytebuf. If you want to use the bytebuf
+     * after the entry is released (via {@link #close()}, please retain the references of the bytebuf.
      *
      * @return a ByteBuf which contains the data
      */
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/WriteAdvHandle.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/WriteAdvHandle.java
@@ -54,7 +54,7 @@ public interface WriteAdvHandle extends ReadHandle {
     /**
      * Add entry asynchronously to an open ledger.
      *
-     * @param entryId entryId to  be added.
+     * @param entryId entryId to be added.
      * @param data array of bytes to be written
      * @return an handle to the result, in case of success it will return the same value of param {@code entryId}.
      */
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/WriteAdvHandle.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/WriteAdvHandle.java
@@ -45,10 +45,34 @@ public interface WriteAdvHandle extends ReadHandle {
      *
      * @param entryId entryId to be added
      * @param data array of bytes to be written
-     * @return an handle to the result, in case of success it will return the same value of param entryId
+     * @return an handle to the result, in case of success it will return the same value of param entryId.
      */
     default CompletableFuture<Long> write(final long entryId, final ByteBuffer data) {
         return write(entryId, Unpooled.wrappedBuffer(data));
+    }
+
+    /**
+     * Add entry asynchronously to an open ledger.
+     *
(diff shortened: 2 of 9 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntry.java
```diff
@@ -55,15 +60,28 @@
...
+     * to {@code entry.getEntryBuffer().nioBuffer()}. This method does not
+     * modify {@code readerIndex} or {@code writerIndex} of the underneath bytebuf.
+     */
+    ByteBuffer getNioBuffer();
+
     /**
      * Return the internal buffer that contains the entry payload.
      *
+     * <p>This call doesn't retain any reference on the underneath bytebuf. If you want to use the bytebuf
+     * after the entry is released (via {@link #close()}, please retain the references of the bytebuf.
```
jvrao: s/please/caller must/

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/WriteAdvHandle.java
```diff
@@ -45,12 +45,36 @@
...
+     * @return an handle to the result, in case of success it will return the same value of param {@code entryId}.
+     */
+    default CompletableFuture<Long> write(final long entryId, final byte[] data) {
+        return write(entryId, Unpooled.wrappedBuffer(data));
+    }
+
+    /**
+     * Add entry asynchronously to an open ledger.
+     *
+     * @param entryId entryId to  be added.
```
jvrao: nit space between "to  be"

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntry.java
sijie: done

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/WriteAdvHandle.java
sijie: done

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/WriteAdvHandle.java
```diff
@@ -45,12 +45,36 @@
...
      */
     default CompletableFuture<Long> write(final long entryId, final ByteBuffer data) {
         return write(entryId, Unpooled.wrappedBuffer(data));
     }
 
+    /**
+     * Add entry asynchronously to an open ledger.
+     *
+     * @param entryId entryId to be added.
+     * @param data array of bytes to be written
```
eolivelli: should we tell that the array will be processed async and so the caller MUST not chnge the contents of the array ?
It is not so obvious, a user could think that data is copied or is not used anymore once the method returns.
Users of BK are advanced developers, but a line of warning maybe would be good

sijie: @eolivelli I am not sure if we need that. that basically means we need to add same comment for every methods that take input from users. it applies almost every methods. If you think we need that, let's do it in a separate task and add comments to all the methods.

eolivelli: @sijie no problem, just asking your opinion. already gave my +1 on this patch

sijie: created #757 for future improvements.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntry.java
```diff
@@ -55,15 +60,28 @@
     long getLength();
 
     /**
-     * Returns the content of the entry.
+     * Returns the content of the entry into a byte array.
```
ivankelly: as a byte array.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntry.java
```diff
@@ -55,15 +60,28 @@
...
      *
      * @return the content of the entry
      */
     byte[] getEntry();
 
+    /**
+     * Exposes this entry's data as an NIO {@link ByteBuffer}. The returned buffer
+     * shares the content with this underneath bytebuf (which you can get it by {@link #getEntryBuffer()}),
```
ivankelly: underlying ByteBuf

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntry.java
```diff
@@ -55,15 +60,28 @@
...
+     * while changing the position and limit of the returned NIO buffer does not affect the indexes and
```
ivankelly: Remove while, and have "Changing" start a new sentence.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntry.java
```diff
@@ -55,15 +60,28 @@
...
+     * marks of this underneath buffer.  This method is identical
+     * to {@code entry.getEntryBuffer().nioBuffer()}. This method does not
+     * modify {@code readerIndex} or {@code writerIndex} of the underneath bytebuf.
```
ivankelly: underlying ByteBuf

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntry.java
ivankelly: Maybe add a @link for ByteBuf

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntry.java
```diff
@@ -55,15 +60,28 @@
...
+     */
+    ByteBuffer getNioBuffer();
+
     /**
      * Return the internal buffer that contains the entry payload.
      *
+     * <p>This call doesn't retain any reference on the underneath bytebuf. If you want to use the bytebuf
```
ivankelly: This call doesn't change the reference count on the returned ByteBuf.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntry.java
```diff
@@ -55,15 +60,28 @@
...
+     * after the entry is released (via {@link #close()}, the caller must retain the references of the bytebuf.
```
ivankelly: ..., the caller must call retain on the ByteBuf.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntry.java
```diff
@@ -55,15 +60,28 @@
     long getLength();
 
     /**
-     * Returns the content of the entry.
+     * Returns the content of the entry into a byte array.
      *
      * @return the content of the entry
      */
     byte[] getEntry();
```
ivankelly: Maybe rename to getEntryBytes()

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntry.java
```diff
@@ -55,15 +60,28 @@
...
 
+    /**
+     * Exposes this entry's data as an NIO {@link ByteBuffer}. The returned buffer
+     * shares the content with this underneath bytebuf (which you can get it by {@link #getEntryBuffer()}),
+     * while changing the position and limit of the returned NIO buffer does not affect the indexes and
+     * marks of this underneath buffer.  This method is identical
+     * to {@code entry.getEntryBuffer().nioBuffer()}. This method does not
+     * modify {@code readerIndex} or {@code writerIndex} of the underneath bytebuf.
+     */
+    ByteBuffer getNioBuffer();
```
ivankelly: getEntryNioBuffer to be consistent?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntry.java
```diff
@@ -21,12 +21,17 @@
...
 
 import io.netty.buffer.ByteBuf;
+import java.nio.ByteBuffer;
 import org.apache.bookkeeper.common.annotation.InterfaceAudience.Public;
 import org.apache.bookkeeper.common.annotation.InterfaceStability.Unstable;
 
 /**
  * An entry.
  *
+ * <p>The entry implementation might be holding references to byte buffers under the hood. The users holding the
```
ivankelly: The entry implementation may hold references to byte buffers under the hood.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntry.java
```diff
@@ -21,12 +21,17 @@
...
+ * references to the instances of this class, are responsible for calling {@link LedgerEntry#close()} to release
+ * resources held by the instances.
```
ivankelly: held by the entry instances.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntry.java
```diff
@@ -21,12 +21,17 @@
 package org.apache.bookkeeper.client.api;
 
 import io.netty.buffer.ByteBuf;
+import java.nio.ByteBuffer;
 import org.apache.bookkeeper.common.annotation.InterfaceAudience.Public;
 import org.apache.bookkeeper.common.annotation.InterfaceStability.Unstable;
 
 /**
  * An entry.
```
ivankelly: An entry in a Ledger.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/WriteAdvHandle.java
```diff
@@ -45,12 +45,36 @@
      *
      * @param entryId entryId to be added
      * @param data array of bytes to be written
-     * @return an handle to the result, in case of success it will return the same value of param entryId
+     * @return an handle to the result, in case of success it will return the same value of param entryId.
```
ivankelly: It's recommended to have no full stop at the end of @param, @return and @throws

http://www.oracle.com/technetwork/articles/java/index-137868.html

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/WriteHandle.java
```diff
@@ -41,7 +41,8 @@
     /**
      * Add entry asynchronously to an open ledger.
      *
-     * @param data array of bytes to be written
+     * @param data a bytebuf to be written. The bytebuf's reference count will be decremented by 1 after the
```
ivankelly: I think it would be better to restate this, or to also state, that the append call passes ownership to of the bytebuf to the handle.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/WriteHandle.java
```diff
@@ -56,6 +57,30 @@
         return append(Unpooled.wrappedBuffer(data));
     }
 
+    /**
+     * Add an entry asynchronously to an open ledger.
+     *
+     * @param data array of bytes to be written.
+     * @return a completable future represents the add result, in case of success the future returns the entry id
+     * of this newly appended entry.
```
ivankelly: Indent

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/WriteHandle.java
```diff
@@ -56,6 +57,30 @@
...
+    }
+
+    /**
+     * Add an entry asynchronously to an open ledger.
+     *
+     * @param data array of bytes to be written.
+     * @param offset the offset in the bytes array.
+     * @param length the length of the bytes to be appended.
+     * @return a completable future represents the add result, in case of success the future returns the entry id
+     * of this newly appended entry.
```
ivankelly: Intent

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/impl/LedgerEntryImpl.java
```diff
@@ -85,6 +86,7 @@ public void setLength(long length) {
     }
 
     public void setEntryBuf(ByteBuf buf) {
+        ReferenceCountUtil.release(entryBuf);
```
ivankelly: is this null safe?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntry.java
jiazhai: For backward compat, we may not able to change this?

ivankelly: There shouldn't be backward compat issues with anything under the api package. client LedgerEntry doesn't inherit from this.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/impl/LedgerEntryImpl.java
sijie: yes it is null safe.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/WriteHandle.java
sijie: I am not sure what do you expect to see about this. There is no ownership thing for this. The accurate statement is the bytebuf is decremented by 1 when the returned completablefuture is done. Because if the caller still has reference counts, the bytebuf is still 'owned' by the caller after the completable future is returned.

sijie:
@ivankelly I addressed your comments. please take a look.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/WriteHandle.java
ivankelly: In the normal case, the bytebuf will have a refcount of 1. It will be passed to append, and then the caller can forget about the bytebuf. In this case, the caller is passing ownership of that 1 refcount to the callee.
If the caller wants to keep a reference, it increments the refcount again. In this case, the callee will still own one refcount, and the caller will own another.

Not too fussed about this in any case. What's there is fine. Though the "completable future is returned" bit is unclear. Is it when the completable future completes, or when the append method returns?

sijie: okay I will change it to "when the returned completablefuture is completed"

ivankelly:
+1

PR #770, opened 2017-11-23
Make LedgerEntries an Iterable<LedgerEntry>

ivankelly:
This seems to have been overlooked in the previous changes. As the title
says, this change just adds the Iterable<LedgerEntry> to the interfaces
from which LedgerEntries extends.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
@@ -20,27 +20,31 @@ package org.apache.bookkeeper.client.api;
 import java.util.Iterator;
 
 /**
- * Interface to wrap the entries.
+ * Interface to wrap a sequence of entries.
  *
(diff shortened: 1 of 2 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
```diff
@@ -20,27 +20,31 @@
...
+     *
      * Calling this method does not modify the reference count of the ByteBuf in the returned LedgerEntry objects.
      * The caller who calls {@link #iterator()} should make sure that they do not call ByteBuf.release() on the
      * LedgerEntry objects to avoid a double free.
      * All reference counts will be decremented when the containing LedgerEntries object is closed.
      *
-     * @return the iterator of type LedgerEntry.
+     * @return an iterator of LedgerEntry objects
      */
     Iterator<LedgerEntry> iterator();
```
eolivelli: maybe add  @Override annotation

PR #771, opened 2017-11-23
Don't require password to create or open ledgers

ivankelly:
The new fluent api shouldn't require a password to create or open a
ledger. The password is only an extra check to avoid accidentally
overwriting data, and the scenarios where this would happen are very
rare, so it's of limited utility. It's not security.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerCreateOp.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerCreateOp.java
@@ -222,7 +222,7 @@ class LedgerCreateOp implements GenericCallback<Void> {
         private int builderEnsembleSize = 3;
         private int builderAckQuorumSize = 2;
         private int builderWriteQuorumSize = 2;
-        private byte[] builderPassword;
+        private byte[] builderPassword = new byte[0];
         private org.apache.bookkeeper.client.api.DigestType builderDigestType
             = org.apache.bookkeeper.client.api.DigestType.CRC32;
         private Map<String, byte[]> builderCustomMetadata = Collections.emptyMap();
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerOpenOp.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerOpenOp.java
@@ -220,7 +220,7 @@ class LedgerOpenOp implements GenericCallback<LedgerMetadata> {
 
         private boolean builderRecovery = false;
         private Long builderLedgerId;
-        private byte[] builderPassword;
+        private byte[] builderPassword = new byte[0];
         private org.apache.bookkeeper.client.api.DigestType builderDigestType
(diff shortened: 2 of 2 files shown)


eolivelli:
This is what I did initially, but @sijie required it.
I prefer to have a default empty password.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerCreateOp.java
```diff
@@ -222,7 +222,7 @@ private void createComplete(int rc, LedgerHandle lh) {
         private int builderEnsembleSize = 3;
         private int builderAckQuorumSize = 2;
         private int builderWriteQuorumSize = 2;
-        private byte[] builderPassword;
+        private byte[] builderPassword = new byte[0];
```
eolivelli: Better to have a constant, so we save an allocation

ivankelly: Sure, will update tomorrow, and check the tests too.

sijie:
password is not the security mechanism. but it is there in the bookies. bookie does password checking. all the applications that written using old api are using some password. we should enforce the check rather than using a default value. because both create and open have to use the same password. I am -1 on providing a default value, given current bookie implementation has password checking.

I don't think it is worth discussing whether password should be default or not. A broader question is whether we still need password, given we already have all the security mechanism in place. instead, I would like to see a proposal or a discussion on removing password completely. before that, I would prefer making password not default.

ivankelly:
I'm not too fussed about this, I just found it annoying that it needed to be added.

Note that if we do get rid of password, the first thing we'll need to do is to get people to stop using it explicitly in their code, which will require a default valid. Also, if we get rid of password, digest type MAC  has to go too (which is no harm).

sijie:
@ivankelly 

> if we do get rid of password, the first thing we'll need to do is to get people to stop using it explicitly in their code, which will require a default valid.

using it explicitly doesn't prevent us removing it. most of existing application already have used some passwords there. if we are going to deprecate password, we remove it from server side first, mark `withPassword` unused or deprecated in one release, in next release, we remove `withPassword` from the builder. so there is no need to make password optional at this release. when we want to change the behavior or remove/deprecate feature, we have to start from server, deprecate from one release and remove in next release.

at this moment, password is required because of bookie implementation. so lets make it required.

jvrao:
We don't use / depend on the password. I am absolutely fine to remove it completely.

eolivelli:
Ok for me to draw a roadmap towards dropping password.
I think that having a default will let new applications and users be able to not deal with it.
So I prefer having a default, but I am not against having it required

sijie:
I would prefer deferring making this default until we have a plan to remove password. we don't have to rush this change in the api. you can always make a required field optional at any time. but if we make this field default but we never remove password, then that's a bad experience to users.

eolivelli:
As I  have already said I am fine with keeping it required, without default. As I committed in the first version on the API

ivankelly:
I don't feel strongly about this one, so closing.

PR #772, opened 2017-11-23
Explicitly search for bookkeeper-server jar

ivankelly:
Rather than listing all jars, and excluding some exceptions, search for
jars which match a specific pattern, namely:

  bookkeeper-server-[version].jar

Where [version] may include numbers, dots and optionally -SNAPSHOT.

### DIFF
--- a/bookkeeper-server/bin/bookkeeper
+++ b/bookkeeper-server/bin/bookkeeper
@@ -71,18 +71,18 @@ else
 fi
 
 # exclude tests jar
-RELEASE_JAR=$(ls ${BK_HOME}/*bookkeeper-server-*.jar 2> /dev/null | grep -v tests | tail -1)
+RELEASE_JAR=$(ls ${BK_HOME}/*bookkeeper-server-*.jar 2> /dev/null | grep -v tests | grep -v sources | tail -1)
 if [ -n "${RELEASE_JAR}" ]; then
   BOOKIE_JAR=${RELEASE_JAR}
 else
-  RELEASE_JAR=$(ls ${BK_HOME}/lib/*bookkeeper-server-*.jar 2> /dev/null | grep -v tests | tail -1)
+  RELEASE_JAR=$(ls ${BK_HOME}/lib/*bookkeeper-server-*.jar 2> /dev/null | grep -v tests | grep -v sources | tail -1)
   if [ -n "${RELEASE_JAR}" ]; then
     BOOKIE_JAR=${RELEASE_JAR}
   fi
 fi
 
 # exclude tests jar
-BUILT_JAR=$(ls ${BK_HOME}/target/*bookkeeper-server-*.jar 2> /dev/null | grep -v tests | tail -1)
+BUILT_JAR=$(ls ${BK_HOME}/target/*bookkeeper-server-*.jar 2> /dev/null | grep -v tests | grep -v sources | tail -1)
 
 if [ -z "${BUILT_JAR}" ] && [ -z "${BOOKIE_JAR}" ]; then
   echo "Couldn't find bookkeeper jar."
--- a/bookkeeper-server/bin/bookkeeper
+++ b/bookkeeper-server/bin/bookkeeper
@@ -70,19 +70,28 @@ else
   JAVA=$JAVA_HOME/bin/java
 fi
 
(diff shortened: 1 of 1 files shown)


sijie:
from programming perspective, this change looks good.

However, I didn't encounter this problem in current master:

I tried two ways:

1. 
```
$ mvn clean install -DskipTests

$ cd bookkeeper-server

$ ls target/
apidocs                                           checkstyle-checker.xml                            javadoc-bundle-options
bookkeeper-server-4.6.0-SNAPSHOT-javadoc.jar      checkstyle-result.xml                             maven-archiver
bookkeeper-server-4.6.0-SNAPSHOT-shaded-tests.jar checkstyle-suppressions.xml                       maven-shared-archive-resources
bookkeeper-server-4.6.0-SNAPSHOT-shaded.jar       classes                                           maven-status
bookkeeper-server-4.6.0-SNAPSHOT-sources.jar      generated-sources                                 rat.txt
bookkeeper-server-4.6.0-SNAPSHOT-tests.jar        generated-test-sources                            test-classes
bookkeeper-server-4.6.0-SNAPSHOT.jar              invoker

$ bin/bookkeeper shell simpletest
JMX enabled by default
Ledger ID: 0
396 entries written
1000 entries written to ledger 0
```

2. I removed lib/ and target/ from bookkeeper-server. and run steps at 1. all the tests passed.

so I don't think this behavior has been broken since we added sources package. I might be a problem that command would produce different results when running on different shells.

bookkeeper-server/bin/bookkeeper
```diff
@@ -71,18 +71,18 @@ else
 fi
 
 # exclude tests jar
-RELEASE_JAR=$(ls ${BK_HOME}/*bookkeeper-server-*.jar 2> /dev/null | grep -v tests | tail -1)
+RELEASE_JAR=$(ls ${BK_HOME}/*bookkeeper-server-*.jar 2> /dev/null | grep -v tests | grep -v sources | tail -1)
```
sijie: if the problem comes from multiple package, or your shell doesn't emit the right result. You need to exclude following packages:

```
target/bookkeeper-server-4.6.0-SNAPSHOT-javadoc.jar
target/bookkeeper-server-4.6.0-SNAPSHOT-shaded.jar
target/bookkeeper-server-4.6.0-SNAPSHOT-sources.jar
```

However I would suggest digging into this issue a bit more and come up a command that can produce determistic results.

ivankelly: Strange. Yes, we should make this more deterministic. 

My shell(debian stable, bash) collates differently:
```~/src/bookkeeper $ ls bookkeeper-server/target/
apidocs                                            bookkeeper-server-4.7.0-SNAPSHOT-sources.jar  classes                 maven-archiver
bookkeeper-server-4.7.0-SNAPSHOT.jar               bookkeeper-server-4.7.0-SNAPSHOT-tests.jar    generated-sources       maven-shared-archive-resources
bookkeeper-server-4.7.0-SNAPSHOT-javadoc.jar       checkstyle-checker.xml                        generated-test-sources  maven-status
bookkeeper-server-4.7.0-SNAPSHOT-shaded.jar        checkstyle-result.xml                         invoker                 test-classes
bookkeeper-server-4.7.0-SNAPSHOT-shaded-tests.jar  checkstyle-suppressions.xml                   javadoc-bundle-options
```

ivankelly: Maybe this is why I was seeing the shading problem also.

sijie: the result returned by `ls` is unsorted. that means it will have different behaviors on different bash (mac vs linux).

a deterministic way is to `ls ... | sort | tail -1`

ivankelly: I assumed ls would lexically sort by default, but I guess that would make ls very expensive on large directories. In any case, I think I'll either check that there's a version before the ".jar" or maybe directly use the shaded. This may fix #773 also.

ivankelly:
@jiazhai ya, I think it was filesystem specific.

@sijie I've updated to search explicitly for the correct jar. This fixes the shading issue also.

ivankelly:
@merlimat I'll create another PR

PR #773, opened 2017-11-23
Shade bookkeeper-common into bookkeeper-server

ivankelly:
bookkeeper-server extends some interfaces from bookkeeper-common which
in turn use guava. As guava is shaded and relocated, we need to shade
the common module also, so that it can find the guava classes.

I hit this problem running localbookie from a locally build source directory. The write path worked fine, but it triggered a NoSuchMethod exception when trying to read a ledger.

### DIFF
--- a/bookkeeper-server/pom.xml
+++ b/bookkeeper-server/pom.xml
@@ -210,6 +210,7 @@
               <shadeTestJar>true</shadeTestJar>
               <artifactSet>
                 <includes>
+                  <include>org.apache.bookkeeper:bookkeeper-common</include>
                   <include>org.apache.bookkeeper:bookkeeper-proto</include>
                   <include>com.google.protobuf:protobuf-java</include>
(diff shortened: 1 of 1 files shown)


sijie:
can you describe more about the issue here, and how to reproduce it?

in general, for bugs, it would be good to create an issue first and describe the bug and how to reproduce it. the code change looks safe and good to me. but I have no idea where does the issue come from.

ivankelly:
@sijie The issue was blocking me, so by the time I understood what was happening, I already had a fix. Will add repro info to the PR description.

sijie:
I think you encountered this issue because of #772, which you might take `shaded` jar into your classpath. but in general, this change looks good to me.

ivankelly:
#772 also solves this now, so closing. We should revisit shading when we break the bookkeeper client out from the server.

PR #776, opened 2017-11-24
Issue 709: Add Slow Bookkeeper Servers to Placement Policy for read ordering

philipsu522:

Maintain a list of slow bookkeeper hosts, which are tried only after readonly bookkeeper hosts are tried. Bookkeeper hosts can be categorized as "slow" when a speculative timeout occurs on that bookkeeper host.


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
```diff
@@ -79,7 +85,8 @@
     static final int REMOTE_MASK      = 0x04 << 24;
     static final int REMOTE_FAIL_MASK = 0x08 << 24;
     static final int READ_ONLY_MASK   = 0x10 << 24;
-    static final int UNAVAIL_MASK     = 0x20 << 24;
+    static final int SLOW_MASK        = 0x20 << 24;
+    static final int UNAVAIL_MASK     = 0x40 << 24;
```
eolivelli: can't we preserve old value for constant ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
```diff
@@ -779,37 +796,107 @@ protected BookieNode selectRandomFromRack(String netPath, Set<Node> excludeBooki
         throw new BKNotEnoughBookiesException();
     }
 
+    @Override
+    public void registerSlowBookie(BookieSocketAddress bookieSocketAddress, long entryId) {
+        slowBookies.put(bookieSocketAddress, entryId);
+
```
eolivelli: nit: remove empty line

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/TestRegionAwareEnsemblePlacementPolicy.java
```diff
@@ -219,14 +220,80 @@ public void testNodeReadOnly() throws Exception {
...
+                ensemble, new BookKeeperServerHealthInfo(), writeSet);
         DistributionSchedule.WriteSet expectedSet
             = writeSetFromValues(3, 1, 2, 0);
         LOG.info("reorder set : {}", reorderSet);
         assertFalse(reorderSet.equals(origWriteSet));
         assertEquals(expectedSet, reorderSet);
     }
 
+    @Test
+    public void testnodeSlow() throws Exception {
```
eolivelli: nit: testNodeSlow

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperServerHealthInfo.java
```diff
@@ -0,0 +1,65 @@
...
+    this.bookiePendingRequests = bookiePendingRequests;
+  }
+
+  /**
+   * Return the failure history for a bookie.
+   *
+   * @param bookieSocketAddress
+   * @return failed entries on a bookie, -1 if there has been no failures.
+   */
+  long getBookieFailureHistory(BookieSocketAddress bookieSocketAddress) {
```
eolivelli: please make this and other getters public, placement policies are pluggable, so if you pass an object to a policy it should be able to "use" it

@sijie @ivankelly

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperServerHealthInfo.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+import org.apache.bookkeeper.net.BookieSocketAddress;
+
+/**
+ * This class encompasses heuristics used to determine the health of a Bookkeeper server for read
+ * ordering. Currently we use the history of failed entries and length of the pending requests queue.
+ */
+public class BookKeeperServerHealthInfo {
```
eolivelli: this class provides info about multiple bookies.
what about "BookiesHealthInfo" ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
ivankelly: No, the order is important, since candidate bookies are sorted by these masks. You want to select a slow bookie before an unavailable one. You probably want to select a slow bookie before a readonly bookie too.

eolivelli: Ok. I understand

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperServerHealthInfo.java
sijie: @eolivelli If naming is not a big concern, let's skip it or rename it later, since it is a port back from twitter branch.

eolivelli: Ok I didn't know it is a port. Maybe it was not clear from the description.
+1

eolivelli: Ok I didn't know it is a port. Maybe it was not clear from the description.
+1

jiazhai:
This is a great idea. 
A question in my mind is that, How it comes when a bookie is first slow, but then after a fix or something it becomes fast?

ivankelly:
@philipsu522 @sijie 

> Maintain a list of slow bookkeeper hosts, which are tried only after readonly bookkeeper hosts are tried.

What is the rationale for trying read-only before slow? Surely a write to the read-only is likely to fail.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
```diff
@@ -420,6 +426,21 @@ synchronized void logErrorAndReattemptRead(int bookieIndex, BookieSocketAddress
             }
         }
 
+        @Override
+        boolean complete(int bookieIndex, BookieSocketAddress host, ByteBuf buffer) {
+            boolean completed = super.complete(bookieIndex, host, buffer);
+            if (completed && lh.bk.getConf().getEnsemblePlacementPolicySlowBookies()) {
+                int numReplicasTried = getNextReplicaIndexToReadFrom();
+                    // Check if any speculative reads were issued and mark any slow bookies before
```
ivankelly: broken indent

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
```diff
@@ -420,6 +426,21 @@ synchronized void logErrorAndReattemptRead(int bookieIndex, BookieSocketAddress
             }
         }
 
+        @Override
+        boolean complete(int bookieIndex, BookieSocketAddress host, ByteBuf buffer) {
+            boolean completed = super.complete(bookieIndex, host, buffer);
+            if (completed && lh.bk.getConf().getEnsemblePlacementPolicySlowBookies()) {
```
ivankelly: i don't think we need to check if this is enabled here. if disabled, the placement policy should just ignore the extra info.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
```diff
@@ -779,37 +796,107 @@ protected BookieNode selectRandomFromRack(String netPath, Set<Node> excludeBooki
...
+        BookKeeperServerHealthInfo bookKeeperServerHealthInfo,
+        boolean regionAware,
+        String myRegion,
+        int remoteNodeInReorderSequence) {
+        boolean useRegionAware = regionAware && (!myRegion.equals(UNKNOWN_REGION));
         int ensembleSize = ensemble.size();
 
         for (int i = 0; i < writeSet.size(); i++) {
-            int idx = writeSet.get(i);
+            Integer idx = writeSet.get(i);
```
ivankelly: We can probably do this without autoboxing, but not something to worry about for this change.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
```diff
@@ -779,37 +796,107 @@ protected BookieNode selectRandomFromRack(String netPath, Set<Node> excludeBooki
...
-                // from readonly bookie in prior to writable
-                // bookies.
+                // from readonly bookie prior to writable bookies.
                 if ((null == readOnlyBookies)
-                        || !readOnlyBookies.contains(address)) {
+                    || !readOnlyBookies.contains(address)) {
                     writeSet.set(i, idx | UNAVAIL_MASK);
                 } else {
-                    writeSet.set(i, idx | READ_ONLY_MASK);
+                    if (slowBookies.asMap().containsKey(address)) {
```
ivankelly: yikes, this is going to run for each entry and allocates a map. Why not use getIfPresent?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
```diff
@@ -779,37 +796,107 @@ protected BookieNode selectRandomFromRack(String netPath, Set<Node> excludeBooki
...
     public DistributionSchedule.WriteSet reorderReadSequence(
             ArrayList<BookieSocketAddress> ensemble,
-            Map<BookieSocketAddress, Long> bookieFailureHistory,
+            BookKeeperServerHealthInfo bookKeeperServerHealthInfo,
             DistributionSchedule.WriteSet writeSet) {
+        Map<Integer, String> writeSetWithRegion = new HashMap<>();
+        for (int i = 0; i < writeSet.size(); i++) {
+            writeSetWithRegion.put(writeSet.get(i), "");
+        }
+        return reorderReadSequenceWithRegion(ensemble, writeSet, writeSetWithRegion, bookKeeperServerHealthInfo, false, "", writeSet.size());
```
ivankelly: why is all this region stuff going into rack aware placement policy? Shouldn't it be region aware?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
ivankelly: hmm, maybe not. Still though, getIfPresent is more direct

sijie: I don't think there is `getIfPresent` in LoadingCache. If I am correct, the only way to get a key from loading cache without loading, is `asMap().containsKey(address))`.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperServerHealthInfo.java
philipsu522: I could make this change - it wasn't a clean port so I had to reimplement a lot of stuff anyways.

sijie:
> What is the rationale for trying read-only before slow? Surely a write to the read-only is likely to fail.

I am not sure about. I think slow should be before read-only. @philipsu522 @yzang can you guys answer this?

philipsu522:
@jiazhai, the loading cache will evict the slow bookie after a certain amount of time based on last write , so if it becomes fast then there won't be any speculative reads. No more speculative reads means that the cache won't be updated for the bookie, and it will be evicted after a certain time.

philipsu522:
> I am not sure about. I think slow should be before read-only. @philipsu522 @yzang can you guys answer this?

I think that we put slow bookies behind read-only bookies thinking that slow bookies are more likely to hurt performance, while bookies could be in RO for any number of reasons. However, based on experience of everyone else, if RO bookies seem to be less reliable than slow bookies, I don't mind changing the order so that it is available bookies, slow bookies, RO bookies...

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
philipsu522: @sijie is correct - there isn't a getIfPresent() function for LoadingCaches. Another option is to call get() and if it returns -1 (defined by the LoadingCache constructor), we can assume it doesn't exist. Thoughts on this?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
```diff
@@ -92,11 +92,17 @@
...
+                    ensemble,
+                    lh
+                );
                 writeSet = lh.bk.getPlacementPolicy()
                     .reorderReadSequence(
                             ensemble,
-                            lh.bookieFailureHistory.asMap(),
-                            lh.distributionSchedule.getWriteSet(eId));
+                            bookKeeperServerHealthInfo,
+                            lh.distributionSchedule.getWriteSet(eid));
```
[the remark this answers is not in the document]
philipsu522: Not sure why it's showing "eId" when the change looks like it says "entryId" now. In any case, this value defined in the super class LedgerEntry:
```
    abstract class LedgerEntryRequest extends LedgerEntry implements SpeculativeRequestExecutor {
```
and
```
public class LedgerEntry
    implements org.apache.bookkeeper.client.api.LedgerEntry {

    final long ledgerId;
    long entryId;
    long length;
    ByteBuf data;
```

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
philipsu522: That's a good question. The reasoning is that I wanted to try to avoid duplication in RackAware and RegionAware placement policies, so I put the shared code in RackAwarePlacementPolicyImpl so that both of the policies can access it since RegionAware is a sublcass of RackAware.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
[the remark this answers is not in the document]
philipsu522: Actually looks like the LedgerEntry interface was changed in a recent commit - I'll need a bit more time to incorporate that change. Will submit changes soon.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
sijie: I think both `get` and `getUnchecked` do loading the key. we don't need loading a key here, so I think we should stick to `asMap().containsKey(address)` here.

ivankelly: getIfPresent is part of Cache, which LoadingCache inherits from.
https://google.github.io/guava/releases/20.0/api/docs/com/google/common/cache/LoadingCache.html

ivankelly:
@philipsu522 re: slow vs read only, makes sense. I had it in my head that this was used for writes, but reordering only happens for reads.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
ivankelly: Ok. We should probably refactor this at some point. methods with loads of parameters and boolean switches become very hard to follow and maintain.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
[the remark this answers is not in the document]
ivankelly: 👍

ivankelly:
@philipsu522 could you merge master into this, and address any comments when you have a chance. If it's left too long it'll be a nightmare to get in.

philipsu522:
Yep, sorry about that - I'll get this done today

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
philipsu522: fixed in latest commit

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
philipsu522: fixed

philipsu522:
@sijie I just rebased master

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -310,6 +313,29 @@ DistributionSchedule getDistributionSchedule() {
         return distributionSchedule;
     }
 
+    /**
+     * Generate the health info for bookies in the write set.
+     *
+     * @return BookiesHealthInfo for every bookie in the write set.
+     */
+    public BookiesHealthInfo generateHealthInfoForWriteSet(
```
ivankelly: This is very allocatey for something that runs for every read request. It allocates a new map for bookiePendingMap, and converts bookieFailureHistory into a map, and allocates a new BookiesHealthInfo object, but all of these are just references to info elsewhere in the client. 

Instead of this, we can have BookiesHealthInfo as a interface, and have a member field in LedgerHandle or BookKeeper.

```
final BookiesHealthInfo bookieHealthInfo = new BookiesHealthInfo() {
    @Override
    public long getBookieFailureHistory(BookieSocketAddress bookieSocketAddress) {
        Long lastFailure = bookieFailureHistory.getIfPresent(bookieSocketAddress);
        return lastFailure == null ? -1L : lastFailure;
    }

    @Override
    public int getBookiePendingRequests(BookieSocketAddress bookieSocketAddress) {
        PerChannelBookieClientPool pcbcPool = client.lookupClient(address);
        return pcbcPool == null ? 0 : pcbcPool.getNumPendingCompletionRequests();
   }
}
```

philipsu522:
@ivankelly makes sense - push a change to convert BookiesHealthInfo into an interface

sijie:
@ivankelly can you review this PR again after @philipsu522 addressed your comments?

ivankelly:
@sijie @philipsu522 I'll take a look first thing tomorrow. Got caught up on other things today.

@philipsu522 Seems this needs another merge from master. I guess it's checkstyle changes, so the easiest thing to do would be take your changes in the conflict resolution, compile and fix any issues again.

Issue #689, opened 2017-11-03
Fix release guide for minor releases


eolivelli:
This issue is for fixing the release guide.
The description of this issue is still WORK-IN-PROGRESS

Problems found:
- no milestone is to be created for a minor release
- no new branch is to be created for a minor release, only a tag
- clarification is needed for the parameters of the maven release plugin
- link "Source and Binary distribution deployed to the dev repository" links to "incubator"
- missing command to bump the branch to the next version (from 4.5.1-SNAPSHOT to 4.5.2-SNAPSHOT for instance)
- fix typos in email template (like "Please down this packages" instead of "please download")

eolivelli:
@sijie I am noting here all the problems I will upgrade the release guide after the release of 4.5.1

PR #777, opened 2017-11-25
Issue-689 improve release guide for minor versions

eolivelli:


eolivelli:
see #689 for description of changes

site/community/release_guide.md
```diff
@@ -234,21 +244,14 @@ Use Maven release plugin to create the release branch and update the current bra
 > $ git reset --hard apache/<master branch OR release tag>
 > $ git branch -D ${BRANCH_NAME}
 
-However, if you are doing an incremental/hotfix release, please run the following command after checking out the release tag of the release being patched.
```
sijie: what to do for minor release?

site/community/release_guide.md
```diff
@@ -378,10 +381,10 @@ Start the review-and-vote thread on the dev@ mailing list. Here’s an email tem
     * All artifacts to be deployed to the Maven Central Repository [3]
     * Source code tag "release-4.5.0" [4]
 
-    BookKeeper's KEY file contains PGP keys we use to sign this release:
+    BookKeeper's KEYS file contains PGP keys we used to sign this release:
```
sijie: why 'used'?

site/community/release_guide.md
```diff
@@ -447,6 +450,18 @@ Once all issues have been resolved, you should go back and build a new release c
 
 Once the release candidate has been reviewed and approved by the community, the release should be finalized. This involves the final deployment of the release candidate to the release repositories, merging of the website changes, etc.
 
+### Advance version on release branch
```
sijie: can't maven:release plugin handle this?

jiazhai:
Overall LGTM. Since we have turn into github only, It is OK to remove all Jira related things in the documentation?  @eolivelli @sijie

site/community/release_guide.md
eolivelli: @sijie I thought that "incremental/hotfix" is a synonym for "minor"

site/community/release_guide.md
eolivelli: Just because while sending the email you have already signed the artifacts, it is on the past, maybe my English is not very good.
Maybe I can reword as
"BookKeeper's KEYS file contains PGP keys we use to sign this releases:"

site/community/release_guide.md
eolivelli: AFAIK this is done done, in fact Jia for 4.6 release left the version to "4.6.0-SNAPSHOT" and I left it to "4.5.1-SNAPSHOT" and JV to "4.5.0-SNAPSHOT"
so some command is missing, or maybe there is another better way to bump version

eolivelli:
@jiazhai @ivankelly has another task to clean up the website from JIRA entirely

site/community/release_guide.md
sijie: yes. but my question is why do you removing the command below. They should be different commands, no?

site/community/release_guide.md
sijie: I think mvn:release does bump the major version when creating the branch. I need to check out the commands.

site/community/release_guide.md
eolivelli: for a minor (4.5.2) release you do not have to use that command, it will create a new branch, but it this case the branch is already here (branch-4.5 in my case)

site/community/release_guide.md
eolivelli: anyway in my example I am using the mvn release plugin with the "update-versions" mojo

eolivelli: and after releasing 4.6.0 (major version) you should have 4.6.1-SNAPSHOT, maybe Jia ans JV used the wrong command

jiazhai: Right, it is create when we creating the branch.

site/community/release_guide.md
sijie: okay make sense

Issue #732, opened 2017-11-16
Release 4.5.1: update docker image file after 4.5.1 is released


jiazhai:
update docker image file after 4.5.1 is released

sijie:
@eolivelli just a reminder you need to update docker image file in branch-4.5 for release 4.5.1. so dockerhub can build the right docker image.

PR #781, opened 2017-11-28
Issue 780:   change BK_statsProviderClass in to NullStatsProvider in Docker file

jiazhai:
In issue #732, we changed Makefile for Docker:
BK_statsProviderClass=org.apache.bookkeeper.stats.NullStatsProvider
to avoid error:
java.lang.ClassNotFoundException: org.apache.bookkeeper.stats.PrometheusMetricsProvider

The reason is that 4.5.1 doesn't ship the providers, so we hard code it to the null providers.

It is more safe to change the default env in Dockerfile instead of Makefile. So that user use the default 4.5.1 Docker will not meet error.

> ---
> quickly and easily:
> 
> 
> ---

### DIFF
--- a/docker/scripts/entrypoint.sh
+++ b/docker/scripts/entrypoint.sh
@@ -37,10 +37,13 @@ export BK_journalDirectory=${BK_journalDirectory:-${BK_DATA_DIR}/journal}
 export BK_ledgerDirectories=${BK_ledgerDirectories:-${BK_DATA_DIR}/ledgers}
 export BK_indexDirectories=${BK_indexDirectories:-${BK_DATA_DIR}/index}
 
+export BK_statsProviderClass=org.apache.bookkeeper.stats.NullStatsProvider
+
 echo "BK_bookiePort bookie service port is $BK_bookiePort"
 echo "BK_zkServers is $BK_zkServers"
 echo "BK_DATA_DIR is $BK_DATA_DIR"
 echo "BK_CLUSTER_ROOT_PATH is $BK_CLUSTER_ROOT_PATH"
+echo "BK_statsProviderClass is $BK_statsProviderClass"
 
 
 mkdir -p "${BK_journalDirectory}" "${BK_ledgerDirectories}" "${BK_indexDirectories}"


docker/scripts/entrypoint.sh
```diff
@@ -37,10 +37,13 @@ export BK_journalDirectory=${BK_journalDirectory:-${BK_DATA_DIR}/journal}
 export BK_ledgerDirectories=${BK_ledgerDirectories:-${BK_DATA_DIR}/ledgers}
 export BK_indexDirectories=${BK_indexDirectories:-${BK_DATA_DIR}/index}
 
+export BK_statsProviderClass=org.apache.bookkeeper.stats.NullStatsProvider
```
ivankelly: This should be
```
export BK_statsProviderClass=${BK_statsProviderClass:-org.apache.bookkeeper.stats.NullStatsProvider}
```

Otherwise, the user can never override this.

sijie: @ivankelly I think you misunderstood what the change is doing.

I believe that the change is here the correct behavior. If I understand correctly, in 4.5.1, none of the stats providers are packaged into the release binary and docker image. setting this to `NullStatsProvider` is to avoid people overriding this to other stats providers, and cause `ClassNotFoundException` exception.

@eolivelli changed that in the makefile only #746 , which only applies to dockers run by the Makefile. but that change doesn't prevent the error when people use the docker image in other environment, e.g. gke. I think the change here is overriding this stats provider and avoid user trying to customize other stats providers. 

I also think this change is only need in 4.5, but it isn't needed in 4.6 and onwards. because 4.6 switch to bookkeeper-dist for release package and docker image, it will package the stats providers in the package.

jiazhai: @sijie Thanks for the comments, it is only for 4.5.1.  and sorry for the un-complete change in 746.

ivankelly: @sijie @jiazhai 
Ah ok, PR description should be updated to state that then. Why have a variable for stats provider at all if it's going to be constant?

Do we have a 4.5.2 release scheduled?

sijie: I think the PR is filed against 4.5 and it is labelled as `release/4.5.1`, no?

the variable is used the `docker/scripts/apply-config-from-env.py`. because the script will pick up environment variables for applying the settings to the config file.

this change is patching docker files. it doesn't require generating a new package. we just need to push this change, so the docker hub can automatically rebuild the image. so, no 4.5.2 release scheduled yet.

ivankelly: @sijie I mean, the PR description should state that 4.5.1 doesn't ship the providers, so we hard code it to the null providers.

sijie: oh okay.

jiazhai:
added the comments in description, Thanks. @ivankelly @sijie

eolivelli:
So we are going to re-tag 4.5.1 ?

sijie:
yes. unfortunately we have to retag it for this release. let's bring the discussion on how to handle docker images back.

PR #782, opened 2017-11-28
Don't print out redundant JMX notice

ivankelly:
Users don't really care that JMX is enabled by default, so don't print
it out.


ivankelly:
Oh, I didn't know that. Ok that case, this while if else should be blocked

ivankelly:
*removed

ivankelly:
Ignore previous two comments. JMX flags are still valid. Will merge this.

PR #785, opened 2017-11-29
Issue 784: LedgerEntries#close should not throw checked exception

sijie:

override `close()` in LedgerEntries to make it not throw checked exception.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
@@ -52,5 +52,6 @@ public interface LedgerEntries
     /**
      * Close to release the resources held by this instance.
      */
+    @Override
     void close();
 }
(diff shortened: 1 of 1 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerEntries.java
```diff
@@ -42,10 +42,15 @@
...
      *
      * @return an iterator of LedgerEntry objects
      */
     @Override
     Iterator<LedgerEntry> iterator();
+
+    /**
+     * Close to release the resources held by this instance.
+     */
+    void close();
```
ivankelly: @Override

ivankelly: haha, there's a github user called override. poor person must have a mailbox full of spam.

sijie: empty repo though :)

sijie:
add `@override`

ivankelly:
lgtm, will merge.

PR #786, opened 2017-11-29
Issue 765: Add `isClosed()` to ReadHandle

sijie:

add `isClosed` to ReadHandle, so distributedlog can use this for implementing readers.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/ReadHandle.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/ReadHandle.java
@@ -127,7 +127,7 @@ public interface ReadHandle extends Handle {
      *
      * @return true if the ledger is sealed, otherwise false.
      */
-    boolean isSealed();
+    boolean isClosed();
 
     /**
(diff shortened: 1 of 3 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/ReadHandle.java
```diff
@@ -115,6 +115,20 @@
...
+     * <p>A ledger is sealed when either the client explicitly closes it ({@link WriteHandle#close()} or
+     * {@link WriteAdvHandle#close()}) or another client explicitly open and recovery it
+     * {@link OpenBuilder#withRecovery(boolean)}.
+     *
+     * <p>This method only checks the metadata cached locally. The metadata can be not update-to-date because
+     * the metadata notification is delayed.
+     *
+     * @return true if the ledger is sealed, otherwise false.
+     */
+    boolean isSealed();
```
ivankelly: If we don't rename the operation that seals the ledger, we shouldn't leave this as isClosed().

sijie:
revert `isSealed` to `isClosed` for now. defer the change to #790

PR #788, opened 2017-11-29
Issue 787: Client should be able to get ledger metadata from Handles.

sijie:

- create a `LedgerMetadata` interface for exposing public information
- add `getLedgerMetadata` to Handle, so both WriteHandle and ReadHandle are able to access the metadata.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
@@ -251,13 +251,9 @@ public class LedgerMetadata implements org.apache.bookkeeper.client.api.LedgerMe
         this.length = length;
     }
 
+    @Override
     public boolean isClosed() {
         return state == LedgerMetadataFormat.State.CLOSED;
-    }
-
-    @Override
-    public boolean isSealed() {
-        return isClosed();
     }
 
     public boolean isInRecovery() {
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
@@ -35,9 +35,11 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.NavigableMap;
 import java.util.Set;
 import java.util.SortedMap;
 import java.util.TreeMap;
+import org.apache.bookkeeper.client.api.DigestType;
 import org.apache.bookkeeper.net.BookieSocketAddress;
 import org.apache.bookkeeper.proto.DataFormats.LedgerMetadataFormat;
 import org.apache.bookkeeper.versioning.Version;
@@ -50,7 +52,7 @@ import org.slf4j.LoggerFactory;
  *
  * <p>It provides parsing and serialization methods of such metadata.
  */
-public class LedgerMetadata {
+public class LedgerMetadata implements org.apache.bookkeeper.client.api.LedgerMetadata {
     static final Logger LOG = LoggerFactory.getLogger(LedgerMetadata.class);
 
     private static final String closed = "CLOSED";
@@ -77,7 +79,7 @@ public class LedgerMetadata {
     private boolean storeSystemtimeAsLedgerCreationTime;
 
     private LedgerMetadataFormat.State state;
-    private SortedMap<Long, ArrayList<BookieSocketAddress>> ensembles =
+    private TreeMap<Long, ArrayList<BookieSocketAddress>> ensembles =
         new TreeMap<Long, ArrayList<BookieSocketAddress>>();
     ArrayList<BookieSocketAddress> currentEnsemble;
     volatile Version version = Version.NEW;
@@ -173,30 +175,35 @@ public class LedgerMetadata {
      * @return SortedMap of Ledger Fragments and the corresponding
      * bookie ensembles that store the entries.
      */
-    public SortedMap<Long, ArrayList<BookieSocketAddress>> getEnsembles() {
+    public TreeMap<Long, ArrayList<BookieSocketAddress>> getEnsembles() {
         return ensembles;
     }
 
-    void setEnsembles(SortedMap<Long, ArrayList<BookieSocketAddress>> ensembles) {
+    @Override
+    public NavigableMap<Long, ? extends List<BookieSocketAddress>> getAllEnsembles() {
+        return ensembles;
+    }
+
+    void setEnsembles(TreeMap<Long, ArrayList<BookieSocketAddress>> ensembles) {
         this.ensembles = ensembles;
     }
 
+    @Override
     public int getEnsembleSize() {
         return ensembleSize;
     }
 
+    @Override
     public int getWriteQuorumSize() {
         return writeQuorumSize;
     }
 
+    @Override
     public int getAckQuorumSize() {
         return ackQuorumSize;
     }
 
-    /**
-     * Get the creation timestamp of the ledger
-     * @return
-     */
+    @Override
     public long getCtime() {
         return ctime;
     }
@@ -221,18 +228,21 @@ public class LedgerMetadata {
         return Arrays.copyOf(password, password.length);
     }
 
-    BookKeeper.DigestType getDigestType() {
+    @Override
+    public DigestType getDigestType() {
         if (digestType.equals(LedgerMetadataFormat.DigestType.HMAC)) {
-            return BookKeeper.DigestType.MAC;
+            return DigestType.MAC;
         } else {
-            return BookKeeper.DigestType.CRC32;
+            return DigestType.CRC32;
         }
     }
 
+    @Override
     public long getLastEntryId() {
         return lastEntryId;
     }
 
+    @Override
     public long getLength() {
         return length;
     }
@@ -243,6 +253,11 @@ public class LedgerMetadata {
 
     public boolean isClosed() {
         return state == LedgerMetadataFormat.State.CLOSED;
+    }
+
+    @Override
+    public boolean isSealed() {
+        return isClosed();
     }
 
     public boolean isInRecovery() {
@@ -279,6 +294,11 @@ public class LedgerMetadata {
         return ensembles.get(ensembles.headMap(entryId + 1).lastKey());
     }
 
+    @Override
+    public List<BookieSocketAddress> getEnsembleAt(long entryId) {
+        return getEnsemble(entryId);
+    }
+
     /**
      * the entry id greater than the given entry-id at which the next ensemble change takes
      * place
@@ -296,6 +316,7 @@ public class LedgerMetadata {
         }
     }
 
+    @Override
     public Map<String, byte[]> getCustomMetadata() {
         return this.customMetadata;
     }
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/Handle.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/Handle.java
@@ -56,6 +56,17 @@ public interface Handle extends AutoCloseable {
     }
 
     /**
+     * Returns the metadata of this ledger.
+     *
+     * <p>This call only retrieves the metadata cached locally. If there is any metadata updated, the read
+     * handle will receive the metadata updates and update the metadata locally. The metadata notification
+     * can be deplayed, so it is possible you can receive a stale copy of ledger metadata from this call.
+     *
+     * @return the metadata of this ledger.
+     */
+    LedgerMetadata getLedgerMetadata();
+
+    /**
      * Asynchronous close, any adds in flight will return errors.
      *
      * <p>Closing a ledger will ensure that all clients agree on what the last
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerMetadata.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerMetadata.java
@@ -1,0 +1,122 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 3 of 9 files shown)


ivankelly:
Why? What's the usecase?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/Handle.java
```diff
@@ -55,6 +55,17 @@ default void close() throws BKException, InterruptedException {
...
+    /**
+     * Returns the metadata of this ledger.
+     *
+     * <p>This call only retrieves the metadata cached locally. If there is any metadata updated, the read
+     * handle will receive the metadata updates and update the metadata locally. The metadata notification
+     * can be deplayed, so it is possible you can receive a stale copy of ledger metadata from this call.
+     *
+     * @return the metadata of this ledger.
+     */
+    LedgerMetadata getLedgerMetadata();
```
ivankelly: rename to getMetadata()? The ledger bit is redundant.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerMetadata.java
```diff
@@ -0,0 +1,122 @@
...
+    long getLastEntryId();
+
+    /**
+     * Returns the length of this ledger.
+     *
+     * <p>If this ledger is not sealed {@link #isSealed()}, it returns {@code 0}.
+     *
+     * @return the length of this ledger if it is sealed, otherwise 0.
+     */
+    long getLength();
```
ivankelly: Better to return Optional<Long> for this and getLastEntryId. These won't be called a lot, so garbage isn't an issue, and it'll be clearer that the value may not exist.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
```diff
@@ -279,6 +294,11 @@ public void addEnsemble(long startEntryId, ArrayList<BookieSocketAddress> ensemb
         return ensembles.get(ensembles.headMap(entryId + 1).lastKey());
     }
 
+    @Override
+    public List<BookieSocketAddress> getEnsembleAt(long entryId) {
+        return getEnsemble(entryId);
```
ivankelly: return a deep copy of the ensemble. We don't want people to be able to change behaviour by calling getLedgerMetadata.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
```diff
@@ -173,30 +175,35 @@ private LedgerMetadata() {
...
      */
-    public SortedMap<Long, ArrayList<BookieSocketAddress>> getEnsembles() {
+    public TreeMap<Long, ArrayList<BookieSocketAddress>> getEnsembles() {
         return ensembles;
     }
 
-    void setEnsembles(SortedMap<Long, ArrayList<BookieSocketAddress>> ensembles) {
+    @Override
+    public NavigableMap<Long, ? extends List<BookieSocketAddress>> getAllEnsembles() {
+        return ensembles;
```
ivankelly: Make a copy, so that people cannot modify behaviour.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerMetadata.java
```diff
@@ -0,0 +1,122 @@
...
+import java.util.Map;
+import java.util.NavigableMap;
+import org.apache.bookkeeper.common.annotation.InterfaceAudience.LimitedPrivate;
+import org.apache.bookkeeper.common.annotation.InterfaceStability.Unstable;
+import org.apache.bookkeeper.net.BookieSocketAddress;
+
+/**
+ * Represents the client-side metadata of a ledger. It is immutable.
+ *
+ * @since 4.6
```
ivankelly: Is this going into 4.6? I thought we had a feature freeze?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerMetadata.java
sijie: I am not sure this should be Optional. there is no point for Optional. If no data is written to a ledger, the length is 0 and last entry id is -1L.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerMetadata.java
sijie: I am running through the requirements for distributedlog to move the new API. The current new API is not in a complete form. I would like 4.6 to include a new API as complete as possible. That's why I've tried to work on many github issues to cover all the corners that were missed in BP-15. 

so it is not a completely new feature, it is for the completeness of BP-15.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerMetadata.java
ivankelly: But the length may not be 0 and last entry Id may not be -1L in the case that the ledger has been written to, just not sealed. The Optional is to indicate that the returned values are not valid. 0 and 1L are valid values, in the case of a empty ledger being sealed.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerMetadata.java
ivankelly: Sure, I'm hitting the same stuff updating the tutorial. Probably should have done these things earlier in the cycle.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/Handle.java
sijie: there is already a "getLedgerMetadata()" in LedgerHandle. it is easier to just provide the same method name.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
sijie: this should be done in a separate task to make LedgerMetadata implementation immutable. This change is more about interfacing existing class. I would prefer not changing any implementation.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
sijie: this should be done in a separate task to make LedgerMetadata implementation immutable. This change is more about interfacing existing class. I would prefer not changing any implementation.

sijie:
> Why? What's the usecase?

There is already a `getLedgerMetadata` in existing LedgerHandle. distributedlog is using this method. The ensemble information is also used by distributedlog for providing mapreduce splits. so I need this method exposed in the new API, otherwise distributedlog can't use the new API. 

This isn't about any new use case. It is more about interfacing existing class and make a "complete" api set for meeting the features exposed by old api.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerMetadata.java
sijie: It is not actually an optional field. I am not sure we really need. I would prefer sticking to current interface, because both methods are already there in the old implementation and used by applications.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
ivankelly: This is ortogonal to the LedgerMetadata immutability. This PR is exposing bookkeeper client internals, and if we do so, we must do it in a way that people cannot abuse them. It's a simple change, just add: new ArrayList(getEnsemble(entryId); BookieSocketAddress is already immutable. The NavigableSet will require a little more work, but not a huge amount.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerMetadata.java
ivankelly: It is optional if you can have a LedgerMetadata object with them unset, which is the case if the ledger is unsealed. The new api gives us a chance to break old interfaces, so we should take that opportunity to make things cleaner.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerMetadata.java
```diff
@@ -0,0 +1,122 @@
...
+     */
+    Map<String, byte[]> getCustomMetadata();
+
+    /**
+     * Returns the ensemble at the given {@code entryId}.
+     *
+     * @param entryId the entry id to retrieve its ensemble information
+     * @return the ensemble which contains the given {@code entryId}.
+     */
+    List<BookieSocketAddress> getEnsembleAt(long entryId);
```
eolivelli: is BookieSocketAddress in the new API ?

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/LedgerMetadataTest.java
```diff
@@ -35,6 +38,36 @@
 
     private static final byte[] passwd = "testPasswd".getBytes(UTF_8);
 
+    @Test
+    public void testGetters() {
```
eolivelli: there are tests (BookKeeperBuildersTest) in which I had to use a cast to LedgerHandle to access LedgerMetadata, maybe it would be good to update them as well

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
sijie: it is not exposing *new* client internals. the exposure is already there in LedgerHandle. This interface doesn't change anything we have been exposed.

The reason I don't want constructing a new ArrayList is because this list should be constructed using ImmutableList during we construct the LedgerMetadata. it should not be done on individual gets. so this makes sense to be part of making ledger metadata immutable, not be part of an API abstraction. If you feel strong about this, I can spend time on haveing a follow up issue to make ledger metadata immutable.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerMetadata.java
sijie: - I am not sure whether optional is more cleaner and returning 0 is more cleaner. looking into storage apis, returning an integer value is very common and well understood for length, last value.

- I don't think it is Optional. Optional means this field is not optional or can be missing. In this case, it just mean the length is not updated yet.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerMetadata.java
sijie: `BookieSocketAddress` is under `net` package not `client` package. it is a common class.

I think we need to move `net` package to bookkeeper-common at some point. If you prefer me doing this movement in this pr, I can do that.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/LedgerMetadataTest.java
sijie: @eolivelli I would prefer cleaning them up when we deprecate old api. otherwise, it might be unnecessarily I am going to touch a lot of places.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerMetadata.java
eolivelli: Let's keep it here now. The module is not important, until we modularize bk in distinct  java 9 modules.
It is important that now that package is in the new public API and we won't change/hide it in the future.
An alternative would be introduce an interface and we are doing for Ledger Metadata. In the future the full .client.api package will have its own module.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/LedgerMetadataTest.java
eolivelli: Fine for me

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
ivankelly: https://github.com/apache/bookkeeper/commit/95373234a273920fadc06742479ddcc4e603fa1b

It's only been exposed this release cycle AFAICT. 

Please don't start on the immutable ledger metadata stuff yet. I already have a partial patch for that from 3+ years ago that does most of the work, but have been holding it off. Big changes like that should be avoided until all yahoo changes are in.

sijie: yes. in distributedlog, we wrote some classes under `org.apache.bookkeeper.client` to access this package private field. One of the reasons that I am making this public as part of ReadHandle is to be able to access some metadata. so I don't need to keep any hacky code in distributedlog.

Also I marked this class as `LimitedPrivate` and `Unstable` at this moment. It is subject to a change in future. because I want to make it immutable, but not at this moment. when we make it immutable, we convert this as `Public`. Hope this make sense to you.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/LedgerMetadata.java
sijie: @eolivelli BookieSocketAddress is shared by both client and server. I feel it is a common class, instead of an interface part of `client.api` package.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
ivankelly: Making this safe now is a trivial change. I don't see a good reason for not doing so. I can do it myself if you want.

sijie: again my concern is allocation. It is nothing about trivial. also again this is `LimitedPrivate`, I don't see any reason why you want to push a change that is going to change in future.

ivankelly: Is this going to be called so much that allocation is a problem? We should only worry about allocation for stuff that gets called with every add or read. 

Anyhow, I'm not going to block this any longer just for this change.

PR #790, opened 2017-11-29
BP-21:  New API close inconsistencies

ivankelly:
Proposal to clean up inconsistencies in Handle#close.

### DIFF
--- a/site/bps/BP-21-new-api-close-inconsistencies.md
+++ b/site/bps/BP-21-new-api-close-inconsistencies.md
@@ -1,9 +1,11 @@
 ---
 title: "BP-21: New API close inconsistencies"
 issue: https://github.com/apache/bookkeeper/issues/789
-state: "Under Discussion"
+state: "Rejected"
 release: "4.6.0"
 ---
+
+Rejected due to lack of agreement that the issues raised in the motivation are valid.
 
 ### Motivation
 
--- a/site/bps/BP-21-new-api-close-inconsistencies.md
+++ b/site/bps/BP-21-new-api-close-inconsistencies.md
@@ -1,0 +1,63 @@
+---
+title: "BP-21: New API close inconsistencies"
+issue: https://github.com/apache/bookkeeper/issues/789
+state: "Under Discussion"
+release: "4.6.0"
+---
+
+### Motivation
+
+The [Handle](http://bookkeeper.apache.org/docs/latest/api/javadoc/org/apache/bookkeeper/client/api/Handle.html) interface provides two methods, #asyncClose and #close (overriding AutoCloseable). 
+
+#close is implemented in both [ReadHandle](http://bookkeeper.apache.org/docs/latest/api/javadoc/org/apache/bookkeeper/client/api/ReadHandle.html) and [WriteHandle](http://bookkeeper.apache.org/docs/latest/api/javadoc/org/apache/bookkeeper/client/api/WriteHandle.html). 
+
+1. The implementations in ReadHandle and WriteHandle do vastly different things. In ReadHandle, #close unregisters listeners from the ledger manager. This is local resource cleanup, which is in line with what AutoCloseable is designed for. In WriteHandle, #close calls #asyncClose which writes the lastAddConfirmed to the LedgerMetadata. This violates the principle of separation of concerns, and overloads the meaning of the term "close".
+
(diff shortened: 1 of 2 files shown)


ivankelly:
https://github.com/ivankelly/bookkeeper/blob/bp-21/site/bps/BP-21-new-api-close-inconsistencies.md for easy to read page.

sijie:
a few thoughts here:

- regarding `seal` and `close`, I am not sure how much it would buy us. because the problem I see this with separated `seal` call, what does it mean if seal failed, shall users call close or not? it is very huge change to the bookkeeper, we will have to change all the documentation and have a guide about what is the difference between `seal` and `close` if we are going to make this change.

- regarding `asyncClose` and `close`, it is not true in ReadHandle. it might be not very meaningful in current implementation. However at Twitter, we have been discussing about unwatch from zookeeper server when closing ledger. It is an action to clean up resources at the server side. We haven't done the change. But there is a possibility we need to clean up resources at the server side for different metadata backends in future. so from this point, I would not say 'asyncClose' is not meaningful. `asyncClose` is still need for readonly handle.

- I am also concerned on removing `asyncClose`, because leaving a `close` method there is not good for async programming. because it is a bit unknown whether close would be blocking or unblocking or not. 

I think if the concern is about the inconsistency about "asyncClose" and "close". I would suggest we changing the pattern to have both sync and async version in the interface, just like what pulsar client api does. it should address your inconsistent concern, and we stay with existing semantic regarding close.

```
CompletableFuture<T> methodAsync();

T method() throws BKException;

CompletableFuture<Void> closeAsync();
void close() ...;
```

ivankelly:
> because the problem I see this with separated seal call, what does it mean if seal failed, shall users call close or not?

With this change, seal is completely divorced from the lifecycle of the local object. Once a WriteHandle has been sealed, the user can still call read() and all the methods of ReadHandle. As such, close would still need to be called. seal mutates the state of the remote object. 

I would consider it to be analogous to #commit and #close in JDBC connection. They do different things, so they are different calls.
https://docs.oracle.com/javase/8/docs/api/java/sql/Connection.html

>  it is very huge change to the bookkeeper, we will have to change all the documentation and have a guide about what is the difference between seal and close if we are going to make this change.

```
~/src/bookkeeper $ find site -name \*.md -exec grep -i Clos[ei] {} \; | wc -l
128
```

I'd say about half a day to change documentation. Keep in mind that this will only be for new api, so most documentation would not have to change.

> regarding asyncClose and close, it is not true in ReadHandle. it might be not very meaningful in current implementation. However at Twitter, we have been discussing about unwatch from zookeeper server when closing ledger. It is an action to clean up resources at the server side. We haven't done the change. But there is a possibility we need to clean up resources at the server side for different metadata backends in future. so from this point, I would not say 'asyncClose' is not meaningful. asyncClose is still need for readonly handle.

Why would the client need to wait for the response in this case? #close is traditionally for cleaning up resources. Is it meaningful for a client if close fails while cleaning up watchers? As in, can it do anything with that result? #seal failing communicates something to the client (i.e. that it probably shouldn't try to move to the next ledger). Cleaning up resources, not so much. Remote resources need to be able to clean themselves up eventually anyhow, given that clients can fail.

> I am also concerned on removing asyncClose, because leaving a close method there is not good for async programming. because it is a bit unknown whether close would be blocking or unblocking or not.

The javadoc for autocloseable says that #close shouldn't throw InterrupttedException which suggests to me that it should never block.

https://docs.oracle.com/javase/8/docs/api/java/lang/AutoCloseable.html#close

> I think if the concern is about the inconsistency about "asyncClose" and "close".
This is one of the concerns, but the overloaded meaning is also a problem.

sijie:
> Why would the client need to wait for the response in this case? #close is traditionally for cleaning up resources. Is it meaningful for a client if close fails while cleaning up watchers? As in, can it do anything with that result? #seal failing communicates something to the client (i.e. that it probably shouldn't try to move to the next ledger). Cleaning up resources, not so much. Remote resources need to be able to clean themselves up eventually anyhow, given that clients can fail.

it is not about wait for response. it is about having an async method for that. whether you want to wait until close to complete or not, it is up to applications. 

```
The javadoc for autocloseable says that #close shouldn't throw InterrupttedException which suggests to me that it should never block.
```

there is not any indication about whether it is a blocking call or non-blocking call. but in an interface that having async and sync methods, #close() typically means a blocking call. that's why people usually have a asyncClose or closeAsync call.


I am not sure about `seal` and `close`, I would defer to see if other people's opinions.

regarding `close`, I would insist us need both sync and async version of close. so -1 on removing async version of close. if we want to make methods consistency, we can do something like below.

```
CompletableFuture<T> methodAsync();
T method() throws BKException;

CompletableFuture<Void> closeAsync();
void close() ...;
```

ivankelly:
Closing for now. See https://github.com/apache/bookkeeper/pull/795#issuecomment-348708093

Issue #230, opened 2017-07-04
Add Checkstyle to the build process


sigee:
**Add Checkstyle to the build process**
[Ruleset](https://github.com/apache/incubator-distributedlog/blob/master/distributedlog-build-tools/src/main/resources/distributedlog/checkstyle.xml)

- [X] bookkeeper-benchmark #231
- [X] Stats API for bookkeeper #231
- [x] bookkeeper-stats-providers
  - [X] Stats provider for codahale metrics #231
  - [X] Stats provider for Finagle stats #231
  - [X] Stats provider for Prometheus #231
  - [X] Stats provider for twitter-ostrich package #231
  - [X] Stats provider for twitter-stats package #231
- [ ] compability dependencies
  - [ ] bookkeeper-server-compat400
  - [ ] bookkeeper-server-compat410
  - [ ] bookkeeper-server-compat420
- [ ] bookkeeper-server
  - [X] org.apache.bookkeeper.auth #234
  - [X] org.apache.bookkeeper.bookie #234
  - [ ] org.apache.bookkeeper.client
  - [ ] org.apache.bookkeeper.conf
  - [x] org.apache.bookkeeper.feature #675 
  - [ ] org.apache.bookkeeper.meta
  - [ ] org.apache.bookkeeper.metastore
  - [ ] org.apache.bookkeeper.net
  - [x] org.apache.bookkeeper.processor #675 
  - [ ] org.apache.bookkeeper.proto
  - [ ] org.apache.bookkeeper.replication
  - [ ] org.apache.bookkeeper.sasl
  - [x] org.apache.bookkeeper.shims.zk #675 
  - [x] org.apache.bookkeeper.stats #675 
  - [x] org.apache.bookkeeper.streaming #675 
  - [ ] org.apache.bookkeeper.util
  - [x] org.apache.bookkeeper.versioning #675 
  - [x] org.apache.bookkeeper.zookeeper #675

jiazhai:
👍, It is clear to have them tracked here.

sijie:
#231 is merged.

sijie:
reopened this. it is a master ticket.

sijie:
@sigee are you still working on this? shall other contributors help pick up this as well?

sijie:
ping @sigee

sigee:
@sijie,
I couldn't finish it. I've got a new project at my job, so I'm very busy, but I'd like to finish this issue as soon as possible.

sijie:
[...]

PR #792, opened 2017-11-29
ISSUE #230: Enable Checkstyle on the proto package

acoburn:
This is part of #230 and adds Checkstyle verification to the proto
package in bookeeper-server.

The changes here should generally be non-controversial, but I would like to highlight one change for particular review. In the `PerChannelBookieClient` class, there is a `public static final` field called `txnIdGenerator`. The checkstyle rules require that such public fields are ALL_CAPS. So one option would be to change that field to `TX_ID_GENERATOR` (this would be a change to the public API).

The relevant diff is here: https://github.com/apache/bookkeeper/compare/master...acoburn:checkstyle_proto?expand=1#diff-e50ee2c1aec1539ea185a94605b0e550R143

However, that field is never used outside this class, so the other option is to make the field private and keep the name as-is. (Making the field private is also a change to the public API)

I chose the second option because the field seemed more like an internal field. If, however, the field should be left public, I can add a commit that changes the name to `TX_ID_GENERATOR` so that it passes the checkstyle rules.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/AuthHandler.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/AuthHandler.java
@@ -20,6 +20,8 @@
  */
 package org.apache.bookkeeper.proto;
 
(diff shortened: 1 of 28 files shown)


sijie:
@acoburn awesome work! the second option about `txnIdGenerator` looks good to me. +1

PR #796, opened 2017-11-30
ISSUE 230: Enable Checkstyle on the conf package

acoburn:
This is part of #230 and adds Checkstyle verification to the conf
package in bookkeeper-server.

As with #792, most of the changes here should be non-controversial. However, as with #792, there is a constant field whose name needed to be renamed. `AbstractConfiguration` has a `protected static final` field called `defaultLoader`. This field is used in the two implementing classes, so keeping `protected` visibility seems right. So in this case, the field has been renamed to `DEFAULT_LOADER` while keeping the `protected static final` modifiers.

The changed line is: https://github.com/apache/bookkeeper/compare/master...acoburn:checkstyle_conf?expand=1#diff-b4f821552b13a688d62015e96092e08eR45


eolivelli:
merging now

PR #798, opened 2017-11-30
Removed all per-entry allocations in Journal operations

ivankelly:
This change was originally b2cc158a from the yahoo-4.3 branch.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
@@ -40,7 +40,6 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import org.apache.bookkeeper.bookie.LedgerDirsManager.NoWritableLedgerDirException;
-
 import org.apache.bookkeeper.conf.ServerConfiguration;
 import org.apache.bookkeeper.proto.BookkeeperInternalCallbacks.WriteCallback;
 import org.apache.bookkeeper.stats.Counter;
@@ -400,7 +399,7 @@ public class Journal extends BookieCriticalThread implements CheckpointSource {
                           RecyclableArrayList<QueueEntry> forceWriteWaiters,
                           boolean shouldClose,
                           boolean isMarker) {
-        ForceWriteRequest req = ForceWriteRequestsRecycler.get();
+        ForceWriteRequest req = forceWriteRequestsRecycler.get();
         req.forceWriteWaiters = forceWriteWaiters;
         req.logFile = logFile;
         req.logId = logId;
@@ -411,7 +410,7 @@ public class Journal extends BookieCriticalThread implements CheckpointSource {
         return req;
     }
 
-    private final Recycler<ForceWriteRequest> ForceWriteRequestsRecycler
+    private final Recycler<ForceWriteRequest> forceWriteRequestsRecycler
         = new Recycler<ForceWriteRequest>() {
                 protected ForceWriteRequest newObject(
                         Recycler.Handle<ForceWriteRequest> handle) {
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
@@ -24,6 +24,10 @@ package org.apache.bookkeeper.bookie;
 import com.google.common.base.Stopwatch;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
+import io.netty.util.Recycler;
+import io.netty.util.Recycler.Handle;
+import io.netty.util.internal.RecyclableArrayList;
+
 import java.io.File;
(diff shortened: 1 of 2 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -24,20 +24,24 @@
...
-import java.util.LinkedList;
 import java.util.List;
+import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
-import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
 import org.apache.bookkeeper.bookie.LedgerDirsManager.NoWritableLedgerDirException;
+
 import org.apache.bookkeeper.conf.ServerConfiguration;
```
sijie: nit: remove empty spaces

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -359,8 +375,48 @@ public void closeFileIfNecessary() {
...
+        req.logFile = logFile;
+        req.logId = logId;
+        req.lastFlushedPosition = lastFlushedPosition;
+        req.shouldClose = shouldClose;
+        req.isMarker = isMarker;
+        forceWriteQueueSize.inc();
+        return req;
+    }
+
+    private final Recycler<ForceWriteRequest> ForceWriteRequestsRecycler
```
sijie: "forceWriteRequestRecycler" - lowercase "f"

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -24,20 +24,24 @@
 import com.google.common.base.Stopwatch;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
+import io.netty.util.Recycler;
+import io.netty.util.Recycler.Handle;
+import io.netty.util.internal.RecyclableArrayList;
```
sijie: It doesn't sound a good idea to use netty internal classes. can we avoid this? if we can avoid, I would prefer we have a similar util class. so we are not subject to netty changes.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -335,11 +351,11 @@ public int process(boolean shouldForceWrite) throws IOException {
                 lastLogMark.setCurLogMark(this.logId, this.lastFlushedPosition);
 
                 // Notify the waiters that the force write succeeded
-                for (QueueEntry e : this.forceWriteWaiters) {
-                    cbThreadPool.execute(e);
+                for (int i = 0; i < forceWriteWaiters.size(); i++) {
+                    cbThreadPool.execute((QueueEntry) forceWriteWaiters.get(i));
```
sijie: it is not very good for type cast. but I think this question is related to the comment above - can we avoid using netty internal class. if we did avoid, can we make it type safe.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
ivankelly: It's a little nasty, but any time we change netty versions we have a bunch of changes to make anyhow, right @merlimat

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
ivankelly: Will do

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
ivankelly: Ya, the casts suck. I don't understand why the linter isn't complaining.

ivankelly:
I'll look at copying in the recyclable array list and adding generics to it.

ivankelly:
@sijie @eolivelli added our own RecycleableArrayList and added generics. Take another look.

ivankelly:
retest

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/collections/RecyclableArrayList.java
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * A simple list which is recyclable.
+ */
+public final class RecyclableArrayList<T> extends ArrayList<T> {
+
+    private static final int DEFAULT_INITIAL_CAPACITY = 8;
+
+    public static class Recycler<X>
+        extends io.netty.util.Recycler<RecyclableArrayList<X>> {
```
sijie: nit: Recycler is already imported. so you can simplify this with `Handler`.

ivankelly: will update

PR #802, opened 2017-12-04
Issue 801: remove assembly plugin in root pom

jiazhai:
mvn release:prepare cmd will fail, for the wrong assembly plugin in root pom. This change remove it, for it has already in bookkeeper-dist/pom.


jiazhai:
Thanks for the reminder

Issue #761, opened 2017-11-22
BP-20: Github workflow for bookkeeper proposals


sijie:
**BP**

This is the master ticket for tracking BP-20 :

Implementing a github workflow for bookkeeper proposals, so all the bookkeeper proposals can be reviewed as code changes, and they can be maintained in the same way as how we maintain documentation.

Proposal PR - #760

ivankelly:
How should this deal with failed/rejected proposals? These proposals will have BP numbers allocated, but they'll not be merged (or will they)?

sijie:
@ivankelly good point. for the failed/rejected proposals, in the BP PR, update the proposal list page to add them to the "Discard" section. so we have all the history about the proposal, which we can revisit them later and know all the context. I will try to send out a PR to add to the proposal workflow.

sijie:
@ivankelly I sent a PR update the workflow. please take a look.

PR #810, opened 2017-12-05
Update the BP workflow

sijie:

- incorporate the feedback from JV last week: allow people to use google doc for writing a BP. Google doc is great for collaboration in general.
- incorporate Ivan's comment on workflow #761  : how to handle failed BPs. The principle here is BPs are in general good ideas, we should keep those records for future references. so add an instruction on merging the failed BPs.


sijie:
@ivankelly this is merged.

PR #813, opened 2017-12-05
Handle unexpected throwable in Bookie V2 request processors

merlimat:
Merging from https://github.com/yahoo/bookkeeper/commit/c54eafbe

When getting any exception, the request processor for V2 protocol must make sure to get some response back to client. In the specific case, it was some NPE that was thrown and not handled properly.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadEntryProcessor.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadEntryProcessor.java
@@ -130,8 +130,8 @@ class ReadEntryProcessor extends PacketProcessorBase {
             LOG.error("Unauthorized access to ledger " + read.getLedgerId(), e);
             errorCode = BookieProtocol.EUA;
         } catch (Throwable t) {
-            LOG.error("Unexpected exception reading at {}:{} : {}",
-                    new Object[] { read.getLedgerId(), read.getEntryId(), t.getMessage(), t });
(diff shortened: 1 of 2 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/WriteEntryProcessor.java
```diff
@@ -87,6 +87,10 @@ protected void processPacket() {
         } catch (BookieException e) {
             LOG.error("Unauthorized access to ledger " + add.getLedgerId(), e);
             rc = BookieProtocol.EUA;
+        } catch (Throwable t) {
+            LOG.error("Unexpected exception while writing {}@{} : {}", new Object[] { add.ledgerId, add.entryId, t.getMessage(), t });
```
sijie: I think this is a too-long line, it might fail checkstyle.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadEntryProcessor.java
```diff
@@ -129,6 +129,10 @@ protected void processPacket() {
         } catch (BookieException e) {
             LOG.error("Unauthorized access to ledger " + read.getLedgerId(), e);
             errorCode = BookieProtocol.EUA;
+        } catch (Throwable t) {
+            LOG.error("Unexpected exception reading at {}:{} : {}",
+                    new Object[] { read.getLedgerId(), read.getEntryId(), t.getMessage(), t });
```
sijie: I think in current version of slf4j, you don't need to use `new Object[] {}` any more.

Issue #568, opened 2017-10-17
Issue with 'preallocation' feature in EntryLogger


ivankelly:
JIRA: https://issues.apache.org/jira/browse/BOOKKEEPER-1082

Reporter: Charan Reddy Guttapalem @reddycharan

With BOOKKEEPER-643 (https://github.com/apache/bookkeeper/commit/694568b0ff0d048c284c8d5db0c9455d30dfa3ce) feature, 'entryLogFilePreallocationEnabled' is introduced. But by the way it is handled,  it looks like it can never be used. 

In EntryLoggerAllocator.createNewLog, even if entryLogPreAllocationEnabled is configured to true, else block is unreachable because ‘preallocation’ will always be null and it would end up with ‘if’ block. So effectively ‘entryLogFilePreallocationEnabled’ logic is broken.

        synchronized BufferedLogChannel createNewLog() throws IOException {
            BufferedLogChannel bc;
            if (!entryLogPreAllocationEnabled || null == preallocation) {
                // initialization time to create a new log
                bc = allocateNewLog();
            } else {
                // has a preallocated entry log
                ......
                ......
                preallocation = allocatorExecutor.submit(new Callable<BufferedLogChannel>() {   <-------- this is the only place where 'preallocation' is set and it is not possible to get into the else block in this method  ------------>

                    @Override
                    public BufferedLogChannel call() throws IOException {
                        return allocateNewLog();
                    }
                });
            }
            LOG.info("Created new entry logger {}.", bc.getLogId());
            return bc;
        }

PR #814, opened 2017-12-05
Issue 568: Refactor EntryLoggerAllocator to enable preallocation

ArvinDevel:

Enable 'entryLogFilePreallocationEnabled' which is  never be used.
> ---
> quickly and easily:
>
>
>
> Otherwise:
> 
> 
> ---

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
@@ -128,6 +128,7 @@ public class EntryLogger {
     private List<BufferedLogChannel> logChannelsToFlush;
     private volatile BufferedLogChannel logChannel;
     private volatile BufferedLogChannel compactionLogChannel;
+
     private final EntryLoggerAllocator entryLoggerAllocator;
     private final boolean entryLogPreAllocationEnabled;
     private final CopyOnWriteArrayList<EntryLogListener> listeners = new CopyOnWriteArrayList<EntryLogListener>();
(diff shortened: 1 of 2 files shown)


sijie:
@eolivelli this is not a port back. please read #568 for the context -- it is a bug reported by Charan a while ago. I think this feature will be exercised by all the test cases by default. Although I agree a more explicit test case would be better.

@ArvinDevel can you try to see if you can add an explicit test case for this?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -648,6 +655,13 @@ void stop() {
             allocatorExecutor.shutdown();
             LOG.info("Stopped entry logger preallocator.");
         }
+
+        /**
+         * get the preallocation for tests.
+         */
+        protected Future<BufferedLogChannel> getPreallocation(){
```
sijie: nit: I think `getPreallocationFuture()` is better. You can remove `protected` to make the method package private, because the test and class are in same package.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -493,6 +494,12 @@ void createNewLog() throws IOException {
         currentDir = logChannel.getLogFile().getParentFile();
     }
 
+    /**
+     * get EntryLoggerAllocator, Just for tests.
+     */
+    protected EntryLoggerAllocator getEntryLoggerAllocator() {
```
sijie: nit: you can remove `protected` to make this method package private, because the test and class are in the same package.

PR #817, opened 2017-12-05
Client times out requests in batch rather than individually

ivankelly:
This patch is a squash of f59b597f, 00fc5cd2, 1658eb61 & 833dc307 from the yahoo-4.3 branch. The change removes the HashWheelTimer from PerChannelBookieClient and instead runs through all outstanding ops periodically, timing them out as necessary.

The motivation for this change is to reduce the number of objects created per add/read request. With HashWheelTimer a TimerTask needs to be added to the Timer, which allocates and returns a Timeout object. We had to be careful to cancel this timeout on success.

The new approach is simpler, doesn't require any allocation per request, and doesn't require any cancellation. Functionally the end result is the same - very slow requests are timed out.

The patch also makes some additions to the original changes (due to the codebase moving since it was originally written).
- PendingAddOps are changed to use the same mechanism for quorum timeouts.
- The checkTimeout method is moved to the PCBCPool so we can run the check on non-connected PCBC instances. This is necessary for TLS.


ivankelly:
This change tickled some tests in a way that made their flakes more prominent. I've fixed the one in TestPCBC in #816 . There's another in TestDisableEnsembleChange which I still haven't tracked down yet. I'm 99% sure this patch isn't causing the root issue though.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
```diff
@@ -768,6 +763,30 @@ public void getBookieInfo(final long requested, GetBookieInfoCallback cb, Object
...
+                if (value.maybeTimeout()) {
+                    ++timedOutOperations;
+                    iterator.remove();
+                }
+            }
+        }
+
+        if (timedOutOperations > 0) {
+            LOG.info("Timed-out {} operations to channel {} for {}",
+                    new Object[] { timedOutOperations, channel, addr });
```
sijie: nit: you don't need `new Object[]` for the current version of slf4j

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ClientConfiguration.java
```diff
@@ -678,11 +679,28 @@ public ClientConfiguration setReadEntryTimeout(int timeout) {
...
-     * TimeoutTask. This value is in milliseconds. Every X milliseconds, the timeout task
-     * will be executed and it will error out entries that have timed out.
+     * Get the interval between successive executions of the operation timeout monitor. This value is in seconds. Every
+     * X seconds, the timeout monitor will be executed and it will error out entries that have timed out.
+     *
+     * @return the interval at which request timeouts will be checked
+     */
+    public long getTimeoutMonitorIntervalSec() {
+        return getLong(TIMEOUT_MONITOR_INTERVAL_SEC,
+                Math.max(Math.min(getAddEntryTimeout(), getReadEntryTimeout()) / 2, 1));
```
sijie: This setting is also used for quorum timeout. so I think the calculation of the default value should be changed, no?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
```diff
@@ -91,14 +88,11 @@ static PendingAddOp create(LedgerHandle lh, ByteBuf payload, AddCallback cb, Obj
...
-        if (op.timeout != null) {
-            op.timeout.cancel();
-        }
-        op.timeout = null;
-        op.timeoutSec = lh.bk.getConf().getAddEntryQuorumTimeout();
+        op.timeoutNanos = TimeUnit.SECONDS.toNanos(lh.bk.getConf().getAddEntryQuorumTimeout());
         op.pendingWriteRequests = 0;
         op.callbackTriggered = false;
         op.hasRun = false;
+        op.requestTimeNanos = Long.MAX_VALUE;
```
sijie: did I miss something there? why the requestTimeNanos is `Long.MAX_VALUE`?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
```diff
@@ -91,14 +88,11 @@ static PendingAddOp create(LedgerHandle lh, ByteBuf payload, AddCallback cb, Obj
         op.completed = false;
         op.ackSet = lh.distributionSchedule.getAckSet();
         op.addOpLogger = lh.bk.getAddOpLogger();
-        if (op.timeout != null) {
-            op.timeout.cancel();
-        }
-        op.timeout = null;
-        op.timeoutSec = lh.bk.getConf().getAddEntryQuorumTimeout();
+        op.timeoutNanos = TimeUnit.SECONDS.toNanos(lh.bk.getConf().getAddEntryQuorumTimeout());
```
sijie: may be worth calculating the value in bookkeeper instance. so we don't need to calculate for every add op.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -182,6 +184,19 @@ public Integer getSample() {
                                               }
                                           });
         initializeExplicitLacFlushPolicy();
+
+        if (bk.getConf().getAddEntryQuorumTimeout() > 0) {
+            SafeRunnable monitor = new SafeRunnable() {
```
sijie: I think if you use this [SafeRunnable](https://github.com/apache/bookkeeper/blob/master/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/SafeRunnable.java), you can use lambda here.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
```diff
@@ -1641,19 +1656,29 @@ void reset(final CompletionKey key,
...
-            cancelTimeoutAndLogOp(rc);
+            logOpResult(rc);
             originalCallback.writeComplete(rc, ledgerId, entryId, addr, ctx);
             key.release();
             handle.recycle(this);
         }
 
+        @Override
+        boolean maybeTimeout() {
+            LOG.info("maybeTimeout");
```
sijie: remove this logging

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieClient.java
```diff
@@ -110,11 +119,29 @@ public BookieClient(ClientConfiguration conf, EventLoopGroup eventLoopGroup,
...
+                    }
+                };
+            this.timeoutFuture = this.scheduler.scheduleAtFixedRate(monitor,
+                                                                    conf.getTimeoutMonitorIntervalSec(),
+                                                                    conf.getTimeoutMonitorIntervalSec(),
+                                                                    TimeUnit.SECONDS);
+        }
+    }
+
+    public BookieClient(ClientConfiguration conf, EventLoopGroup eventLoopGroup, OrderedSafeExecutor executor,
```
sijie: Can this constructor be package private and add `@VisibleForTesting` annotation? I believe this constructor is only used for testing.

sijie: who is using this constructor?

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieClient.java
```diff
@@ -98,7 +101,13 @@ public BookieClient(ClientConfiguration conf, EventLoopGroup eventLoopGroup,
     }
 
     public BookieClient(ClientConfiguration conf, EventLoopGroup eventLoopGroup,
-                        OrderedSafeExecutor executor, StatsLogger statsLogger) throws IOException {
+                        OrderedSafeExecutor executor, ScheduledExecutorService scheduler) throws IOException {
```
sijie: who is using this constructor?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -182,6 +184,19 @@ public Integer getSample() {
                                               }
                                           });
         initializeExplicitLacFlushPolicy();
+
+        if (bk.getConf().getAddEntryQuorumTimeout() > 0) {
```
eolivelli: can AddEntryQuorumTimeout be disabled ? I think this would be an error. Maybe we can add some assertion and fail fast. We can do it in the future, not blocker

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -1371,6 +1389,16 @@ void handleUnrecoverableErrorDuringAdd(int rc) {
...
     }
 
+    void monitorPendingAddOps() {
+        int timedOut = 0;
+        for (PendingAddOp op : pendingAddOps) {
+            if (op.maybeTimeout()) {
+                timedOut++;
+            }
+        }
+        LOG.info("Timed out {} add ops", timedOut);
```
eolivelli: maybe it is better to log this only if timedOut > 0, otherwise we will spam logs at INFO level

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
eolivelli: +1

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieClient.java
eolivelli: +1 for dropping if not used in BK neither in DL

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieClient.java
```diff
@@ -511,12 +538,14 @@ public void safeRun() {
         }
     }
 
-    public boolean isClosed() {
-        return closed;
+    public void monitorPendingOperations() {
+        for (PerChannelBookieClientPool clientPool : channels.values()) {
```
eolivelli: channels.values().forEach ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
jiazhai: Not related to this review, Since we have already import lombok, how about refine all the LOG by lombok.extern.slf4j.Slf4j to make them more unify and cleaner?

ivankelly: @zhaijack Yes, but lets hold off until all the yahoo changes are in. Changes like that can cause a lot of churn and make merges a pain.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
ivankelly: Unfortunately not. SafeRunnable is not a functional interface (I guess because it has more than one overridable methods).

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -1371,6 +1389,16 @@ void handleUnrecoverableErrorDuringAdd(int rc) {
         asyncCloseInternal(NoopCloseCallback.instance, null, rc);
     }
 
+    void monitorPendingAddOps() {
```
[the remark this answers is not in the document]

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
ivankelly: done

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
ivankelly: It's defensively defaulting to disabling the timeout if we forget to set the start time of the request. Long.MAX_VALUE will always be in the future.

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ClientConfiguration.java
ivankelly: I'll add it, though the quorum timeout shouldn't be less than add entry timeout.

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ClientConfiguration.java
```diff
@@ -678,11 +679,28 @@ public ClientConfiguration setReadEntryTimeout(int timeout) {
...
+    }
+
+    public ClientConfiguration setTimeoutMonitorIntervalSec(long timeoutInterval) {
```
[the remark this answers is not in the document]
ivankelly: done

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieClient.java
ivankelly: killed all constructors except the main one.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
```diff
@@ -768,6 +763,30 @@ public void getBookieInfo(final long requested, GetBookieInfoCallback cb, Object
         writeAndFlush(channel, completionKey, getBookieInfoRequest);
     }
 
+    private static BiPredicate<CompletionKey, CompletionValue> timeoutCheck = (key, value) -> {
```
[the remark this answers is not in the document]
ivankelly: done

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
ivankelly: done

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
ivankelly: logging removed

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
ivankelly: @eolivelli how do you mean it would be an error? If add entry quorum timeout is disabled, the timeouts will only be on the individual add entry RPCs sent to bookies.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieClient.java
ivankelly: I don't think it would add anything to the clarity, and we'd be allocating a lamba each time.

ivankelly:
retest this please

ivankelly:
retest this please

ivankelly:
@sijie could you check this one again?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
merlimat: I had added something for that in MangedLedger code: 
https://github.com/apache/incubator-pulsar/blob/master/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/util/SafeRun.java

They way it's used is like: 

```java
executor.submit(safeRun(() -> {
 // ....
});
```

jvrao:
at a high level, we are trading GC for CPU. i.e instead of allocating objects/cleaning them we are doing a regular check of the queue. Also, timeouts may not be as accurate as before due to timeout checker period granularity.  I am not suggesting this is bad, but have we run any analysis on the gains/losses ?

ivankelly:
@jvrao , @merlimat would be the best one to answer that.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
eolivelli: @ivankelly got it, thank you. I had another (wrong) idea in my mind

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
sijie: The new `SafeRunnable` is a functional interface. The old `SafeRunnable` is going to be deprecated in future.

sijie:
@jvrao : I think:

for normal cases, most of requests will complete very quickly. so this change will save a lot of object allocations.

for timeout cases, the timeout checking runs at half of timeout periods. it is mostly *accurate*.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
ivankelly: Ah, I didn't know there were two SafeRunnables. Was in the middle of changing when @jiazhai merged :)

We can change these to use lambdas when we deprecate the old SafeRunnable.

PR #821, opened 2017-12-07
Use ConcurrentLongLongHashMap in EntryLogMetadata

merlimat:
Porting commits https://github.com/yahoo/bookkeeper/commit/74afc16d1c1e93739edf4a78801c5ccb6afd3883 and https://github.com/yahoo/bookkeeper/commit/eba3bbfbbaf58118c0c5736b56d3897a5928bc89 from Yahoo branch. 

Currently the `EntryLogger` is keeping a `ConcurrentHashMap<Long, Long>` to keep track of amount of data stored in a single entry log file for each of the ledgers. 

The amount of memory required is high because of the objects boxing. The `Long` used for key and value are 16 bytes each and the hash-map node will take another ~24 bytes. 

Additionally, all these objects are created, stored for hours and then GCed, after being tenured. 

By using `ConcurrentLongLongHashMap`, we eliminate the object boxing and the need for the "node" since it's a open hashmap. This helps reducing a lot the memory used by the bookie when storing a large amount of entries (54 --> 16 bytes).
Also since we store the items in `long[]`, this will not need objects and it will not affect GC.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogMetadata.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogMetadata.java
@@ -21,8 +21,9 @@
 
 package org.apache.bookkeeper.bookie;
 
-import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.LongPredicate;
+
+import org.apache.bookkeeper.util.collections.ConcurrentLongLongHashMap;
 
 /**
  * Records the total size, remaining size and the set of ledgers that comprise a entry log.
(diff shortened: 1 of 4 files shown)


merlimat:
retest this please

merlimat:
I've checked the test failures and they appear to be unrelated flaky. The `CreateNewLogTest` is failing for master as well.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogMetadata.java
```diff
@@ -86,10 +74,20 @@ public long getRemainingSize() {
...
-    Map<Long, Long> getLedgersMap() {
+    ConcurrentLongLongHashMap getLedgersMap() {
         return ledgersMap;
     }
 
+    public void removeLedgerIf(LongPredicate predicate) {
+        ledgersMap.removeIf((ledgerId, size) -> {
+            boolean shouldRemove = predicate.test(ledgerId);
+            if (shouldRemove) {
+                remainingSize -= size;
```
eolivelli: This is interesting, if we had a standard ConcurrentHashMap the count could be wrong.
But in our implementation the predicate is executed only once.

PR #826, opened 2017-12-09
DbLedgerStorage implementation

merlimat:
Porting multiple squashed commits from Yahoo branch. 

This change adds a new `LedgerStorage` implementation that uses RocksDB to store the indexes for (ledgerId, entryId) to the data that is still kept in the entry logs.

In addition, there are also a `WriteCache` and a `ReadCache` that are used to completely decouple the read & write paths.

Note: This still has some methods that are not implemented. They are around the features added in 4.5 release: 

```java
public long getLastAddConfirmed(long ledgerId) throws IOException;
public Observable waitForLastAddConfirmedUpdate(long ledgerId, long previoisLAC, Observer observer) throws IOException;
public void setExplicitlac(long ledgerId, ByteBuf lac) throws IOException;
public ByteBuf getExplicitLac(long ledgerId);
```

These will be implemented in a subsequent commit.

### DIFF
--- a/bookkeeper-proto/src/main/proto/DbLedgerStorageDataFormats.proto
+++ b/bookkeeper-proto/src/main/proto/DbLedgerStorageDataFormats.proto
@@ -24,7 +24,7 @@ option optimize_for = SPEED;
  * Ledger metadata stored in the bookie
  */
 message LedgerData {
-	required bool exists = 1;
+    required bool exists = 1;
     required bool fenced = 2;
     required bytes masterKey = 3;
 }
--- a/bookkeeper-proto/src/main/proto/DbLedgerStorageDataFormats.proto
+++ b/bookkeeper-proto/src/main/proto/DbLedgerStorageDataFormats.proto
@@ -1,0 +1,30 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+ 
+syntax = "proto2";
+
+option java_package = "org.apache.bookkeeper.bookie.storage.ldb";
+option optimize_for = SPEED;
+
+/**
+ * Ledger metadata stored in the bookie
+ */
+message LedgerData {
+	required bool exists = 1;
+    required bool fenced = 2;
+    required bytes masterKey = 3;
+}
--- a/bookkeeper-server/pom.xml
+++ b/bookkeeper-server/pom.xml
@@ -38,6 +38,11 @@
       <groupId>org.apache.bookkeeper</groupId>
       <artifactId>bookkeeper-proto</artifactId>
       <version>${project.parent.version}</version>
+	</dependency>
+    <dependency>
+      <groupId>org.rocksdb</groupId>
+      <artifactId>rocksdbjni</artifactId>
+      <version>5.8.6</version>
     </dependency>
     <dependency>
       <groupId>org.slf4j</groupId>
--- a/bookkeeper-server/pom.xml
+++ b/bookkeeper-server/pom.xml
@@ -38,7 +38,7 @@
       <groupId>org.apache.bookkeeper</groupId>
       <artifactId>bookkeeper-proto</artifactId>
       <version>${project.parent.version}</version>
-	</dependency>
+    </dependency>
     <dependency>
       <groupId>org.rocksdb</groupId>
       <artifactId>rocksdbjni</artifactId>
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
@@ -26,6 +26,7 @@ import static org.apache.bookkeeper.bookie.TransactionalEntryLogCompactor.COMPAC
 import static org.apache.bookkeeper.util.BookKeeperConstants.MAX_LOG_SIZE_LIMIT;
 
 import com.google.common.collect.MapMaker;
+import com.google.common.collect.Sets;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufAllocator;
@@ -40,6 +41,7 @@ import java.io.FileFilter;
 import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
+import java.io.FilenameFilter;
 import java.io.IOException;
 import java.io.InputStreamReader;
 import java.io.OutputStreamWriter;
@@ -52,6 +54,7 @@ import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
@@ -203,7 +206,7 @@ public class EntryLogger {
     /**
      * Scan entries in a entry log file.
      */
-    interface EntryLogScanner {
+    public interface EntryLogScanner {
         /**
          * Tests whether or not the entries belongs to the specified ledger
          * should be processed.
@@ -837,7 +840,7 @@ public class EntryLogger {
         leastUnflushedLogId = flushedLogId + 1;
     }
 
-    void flush() throws IOException {
+    public synchronized void flush() throws IOException {
         flushRotatedLogs();
         flushCurrentLog();
     }
@@ -867,7 +870,7 @@ public class EntryLogger {
         }
     };
 
-    synchronized long addEntry(long ledger, ByteBuf entry, boolean rollLog) throws IOException {
+    public synchronized long addEntry(long ledger, ByteBuf entry, boolean rollLog) throws IOException {
         int entrySize = entry.readableBytes() + 4; // Adding 4 bytes to prepend the size
         boolean reachEntryLogLimit =
             rollLog ? reachEntryLogLimit(entrySize) : readEntryLogHardLimit(entrySize);
@@ -963,7 +966,6 @@ public class EntryLogger {
         }
     }
 
-
     private void incrementBytesWrittenAndMaybeFlush(long bytesWritten) throws IOException {
         if (!doRegularFlushes) {
             return;
@@ -986,7 +988,8 @@ public class EntryLogger {
         return logChannel.position() + size > Integer.MAX_VALUE;
     }
 
-    ByteBuf readEntry(long ledgerId, long entryId, long location) throws IOException, Bookie.NoEntryException {
+    public ByteBuf internalReadEntry(long ledgerId, long entryId, long location)
+            throws IOException, Bookie.NoEntryException {
         long entryLogId = logIdForOffset(location);
         long pos = location & 0xffffffffL;
         ByteBuf sizeBuff = sizeBuffer.get();
@@ -1035,6 +1038,15 @@ public class EntryLogger {
                                               + pos + "(" + rc + "!=" + entrySize + ")", ledgerId, entryId);
         }
         data.writerIndex(entrySize);
+
+        return data;
+    }
+
+    public ByteBuf readEntry(long ledgerId, long entryId, long location) throws IOException, Bookie.NoEntryException {
+        long entryLogId = logIdForOffset(location);
+        long pos = location & 0xffffffffL;
+
+        ByteBuf data = internalReadEntry(ledgerId, entryId, location);
         long thisLedgerId = data.getLong(0);
         if (thisLedgerId != ledgerId) {
             data.release();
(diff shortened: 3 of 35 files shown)


bookkeeper-proto/src/main/proto/DbLedgerStorageDataFormats.proto
```diff
@@ -0,0 +1,30 @@
...
+syntax = "proto2";
+
+option java_package = "org.apache.bookkeeper.bookie.storage.ldb";
+option optimize_for = SPEED;
+
+/**
+ * Ledger metadata stored in the bookie
+ */
+message LedgerData {
+	required bool exists = 1;
```
sijie: tab => spaces

bookkeeper-server/pom.xml
```diff
@@ -38,6 +38,11 @@
       <groupId>org.apache.bookkeeper</groupId>
       <artifactId>bookkeeper-proto</artifactId>
       <version>${project.parent.version}</version>
+	</dependency>
```
sijie: tab => spaces

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -837,7 +840,7 @@ void flushRotatedLogs() throws IOException {
         leastUnflushedLogId = flushedLogId + 1;
     }
 
-    void flush() throws IOException {
+    public synchronized void flush() throws IOException {
```
sijie: any reason why `synchronized` the whole method?

sijie:
@merlimat I think there are findbugs and checkstyle errors. It is good if you can take a look and fix them. Since it is a new implementation, I only have a few comments on the places modifying existing code, otherwise, the change looks good to me.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/ArrayUtil.java
```diff
@@ -0,0 +1,77 @@
...
+            array[index + 5] = (byte) (value >>> 16);
+            array[index + 6] = (byte) (value >>> 8);
+            array[index + 7] = (byte) value;
+        }
+    }
+
+    public static final boolean isArrayAllZeros(final byte[] array) {
+        int sum = 0;
+        for (byte b : array) {
+            sum |= b;
```
eolivelli: Does this complete faster than early exiting with an if (b != 0) return false ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
```diff
@@ -0,0 +1,767 @@
...
+                // Insert entry in read cache
+                readCache.put(orginalLedgerId, currentEntryId, entry);
+
+                count++;
+                size += entry.readableBytes();
+
+                currentEntryLocation += 4 + entry.readableBytes();
+                currentEntryLogId = currentEntryLocation >> 32;
+
+                entry.release();
```
eolivelli: This should be in a finally block?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
```diff
@@ -0,0 +1,767 @@
...
+            }
+
+            readAheadBatchCountStats.registerSuccessfulValue(count);
+            readAheadBatchSizeStats.registerSuccessfulValue(size);
+        } catch (Exception e) {
```
eolivelli: Is there a motivation for catching Exception and not a more specific set of exception?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
```diff
@@ -0,0 +1,767 @@
...
+            throw new IOException(e);
+        } finally {
+            isFlushOngoing.set(false);
+        }
+
+        return;
+    }
+
+    @Override
+    public synchronized void flush() throws IOException {
```
eolivelli: There is a mix of synchronized blocks and usage of readwrite locks. I did not checked all the flows.
Just curious to know why don't you use only locks

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/EntryLocationIndex.java
```diff
@@ -0,0 +1,341 @@
...
+            }
+        } finally {
+            firstKeyWrapper.recycle();
+            lastKeyWrapper.recycle();
+            keyToDelete.recycle();
+            batch.close();
+        }
+    }
+
+    ///
```
eolivelli: Empty comment

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageRocksDB.java
```diff
@@ -0,0 +1,539 @@
...
+            try {
+                db.write(optionSync, writeBatch);
+            } catch (RocksDBException e) {
+                throw new IOException("Failed to flush RocksDB batch", e);
+            }
+        }
+    }
+
+    private static class EntryWrapper implements Entry<byte[], byte[]> {
+        private byte[] key;
```
eolivelli: Final?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/ArrayUtil.java
merlimat: The theory is that removing the branching should get the optimizer job much easier, since it can then unroll the loop. 

This is only used to check for empty master key, so it won't be a long array in any case.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
merlimat: good point

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
merlimat: I think the intention was to ignore failures in read ahead, since we can already complete the current read entry operation.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
merlimat: Good point. That's more of a reflection of how the code evolved. This is actually the condensed view of 10s of iterations over the time. 

I think we can change the `synchronized` into a `ReentrantLock` to make it easier to follow.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
merlimat: There can only be a single `flush()` happening at a point in time. That will not affect the `addEntry()` which is only stopped while the 2 `WriteCache` instances are swapped.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/storage/ldb/ArraySortGroupTest.java
```diff
@@ -0,0 +1,97 @@
...
+        };
+
+        ArrayGroupSort sorter = new ArrayGroupSort(2, 4);
+        sorter.sort(data);
+
+        assertArrayEquals(expectedSorted, data);
+    }
+
+    @Test
+    public void errors() {
```
eolivelli: Maybe we can split in smaller methods? To make it clearer what the test does?

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/storage/ldb/ArraySortGroupTest.java
```diff
@@ -0,0 +1,97 @@
...
+
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.fail;
+
+import org.junit.Test;
+
+public class ArraySortGroupTest {
+
+    @Test
+    public void simple() {
```
eolivelli: Don't know if it is better to add some test at least for most common corner cases like empty array, all equals values, odd/even number of elements.....

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/storage/ldb/ConversionRollbackTest.java
```diff
@@ -0,0 +1,138 @@
...
+
+    @Test
+    public void convertFromDbStorageToInterleaved() throws Exception {
+        File tmpDir = File.createTempFile("bkTest", ".dir");
+        tmpDir.delete();
+        tmpDir.mkdir();
+        File curDir = Bookie.getCurrentDirectory(tmpDir);
+        Bookie.checkDirectoryStructure(curDir);
+
+        System.out.println(tmpDir);
```
eolivelli: Logger?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/WriteCache.java
```diff
@@ -0,0 +1,261 @@
...
+        }
+    }
+
+    public void deleteLedger(long ledgerId) {
+        deletedLedgers.add(ledgerId);
+    }
+
+    private static final ArrayGroupSort groupSorter = new ArrayGroupSort(2, 4);
+
+    public synchronized void forEach(EntryConsumer consumer) {
```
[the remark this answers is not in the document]
merlimat: Even if only one method is `synchronized`, still we need to have only 1 single thread calling the method at a time. 

The reason is that the `forEach()` iterates over the entries in order (by ledgerId, entryId), thus it needs to convert the hashmap into a sorted array of pair of longs. The array used is a member variable of the instance (`long[] sortedEntries`).

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
```diff
@@ -0,0 +1,767 @@
...
+            WriteCache tmp = writeCacheBeingFlushed;
+            writeCacheBeingFlushed = writeCache;
+            writeCache = tmp;
+
+            // since the cache is switched, we can allow flush to be triggered
+            hasFlushBeenTriggered.set(false);
+            flushWriteCacheCondition.signalAll();
+        } finally {
+            isFlushOngoing.set(true);
+            writeCacheMutex.writeLock().unlock();
```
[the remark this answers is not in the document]
merlimat: Well, the only reason for `writeCacheMutex` to not be unlocked would be to have `isFlushOngoing == null`, but its a final Object assigned in the constructor.

For the method logic, it's critical to set the flag, to mark an ongoing flush before releasing the write lock. That is used when a flush is triggered by the write cache being full. In this case we want to trigger 1 (and only 1) flush and wait until the write cache is rotated to put the entry in the new empty write cache and return.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/ReadCache.java
```diff
@@ -0,0 +1,185 @@
...
+        }
+    }
+
+    @Override
+    public void close() {
+        cacheSegments.forEach(ByteBuf::release);
+    }
+
+    public void put(long ledgerId, long entryId, ByteBuf entry) {
+        lock.readLock().lock();
```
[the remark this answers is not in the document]
merlimat: Sure, the 2 lines there don't need to be in the read lock section.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/storage/ldb/ArraySortGroupTest.java
merlimat: Sure, added a bit more of these

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
```diff
@@ -0,0 +1,767 @@
...
+            readCache.put(ledgerId, entryId, entry);
+
+            // Try to read more entries
+            long nextEntryLocation = entryLocation + 4 /* size header */ + entry.readableBytes();
+            fillReadAheadCache(ledgerId, entryId + 1, nextEntryLocation);
+
+            recordSuccessfulEvent(readCacheMissStats, startTime);
+            recordSuccessfulEvent(readEntryStats, startTime);
+            return entry;
+        } catch (NoEntryException e) {
```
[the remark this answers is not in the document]
merlimat: Sure, makes sense. `NoEntryException` cannot be thrown after that point.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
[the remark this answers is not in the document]
merlimat: Actually, findbugs catches it. Just resolved to additional `try / finally` block to make findbugs happy :)

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
merlimat: Uhm, I thought this was `LedgerStorage.flush()`. Let me revisit this.

merlimat: Yes, that was not needed indeed.

merlimat:
Addressed comments & fixed checkstyle and findbugs warnings

merlimat:
retest this please

merlimat:
@jiazhai eventually I ended up doing similar change and that was merged in https://github.com/yahoo/bookkeeper/commit/c4f1f9aca8ea9a8d470626e220fb9d3d06cdcafb

This change is what the Yahoo branch was having and its included in this PR.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/ArrayUtil.java
eolivelli: FYI: I have double checked, PlatformDependant has a "isZero" method

jiazhai:
👍, I C, right it is included.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/ArrayUtil.java
merlimat: 👍  Nice. I think that was added recently. Will switch to that.

merlimat:
@ivankelly Sure, I can break it up that way.

merlimat:
I think I have addressed all comments so far. Closing it right now to send out individual patches.

PR #828, opened 2017-12-11
DbLedgerStorage -- Write cache

merlimat:
1st part of DbLedgerStorage related changes.

This PR introduces a `WriteCache` class that is used to store entries before they get flushed on the entryLogs. 

The key part is that it provides a way to iterate in order over the entries and it is is garbage-free at steady state.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/WriteCache.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/WriteCache.java
@@ -1,0 +1,278 @@
+/**
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+package org.apache.bookkeeper.bookie.storage.ldb;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.ByteBufAllocator;
+import io.netty.buffer.Unpooled;
+
+import java.io.Closeable;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.LongAdder;
+import java.util.concurrent.locks.ReentrantLock;
+
+import org.apache.bookkeeper.util.collections.ConcurrentLongHashSet;
+import org.apache.bookkeeper.util.collections.ConcurrentLongLongHashMap;
+import org.apache.bookkeeper.util.collections.ConcurrentLongLongPairHashMap;
+import org.apache.bookkeeper.util.collections.ConcurrentLongLongPairHashMap.LongPair;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Write cache implementation.
+ *
+ * <p>The write cache will allocate the requested size from direct memory
+ * and it will break it down into multiple segments.
+ *
+ * <p>The entries are appended in a common buffer and indexed though a
+ * hashmap, until the cache is cleared.
+ *
+ * <p>There is the possibility to iterate through the stored entries in
+ * an ordered way, by (ledgerId, entry).
+ */
+public class WriteCache implements Closeable {
+
+    /**
+     * Consumer that is used to scan the entire write cache.
+     */
+    public interface EntryConsumer {
+        void accept(long ledgerId, long entryId, ByteBuf entry);
+    }
+
+    private final ConcurrentLongLongPairHashMap index = new ConcurrentLongLongPairHashMap(4096,
+            2 * Runtime.getRuntime().availableProcessors());
+
+    private final ConcurrentLongLongHashMap lastEntryMap = new ConcurrentLongLongHashMap(4096,
+            2 * Runtime.getRuntime().availableProcessors());
+
+    private final ByteBuf[] cacheSegments;
+    private final int segmentsCount;
+
+    private static final int MaxSegmentSize = Integer.MAX_VALUE;
+    private static final long SegmentOffsetMask = (long) Integer.MAX_VALUE;
+
+    private final long maxCacheSize;
+    private final AtomicLong cacheSize = new AtomicLong(0);
+    private final LongAdder cacheCount = new LongAdder();
+
+    private final ConcurrentLongHashSet deletedLedgers = new ConcurrentLongHashSet();
+
+    public WriteCache(long maxCacheSize) {
+        this.maxCacheSize = maxCacheSize;
+        this.segmentsCount = 1 + (int) (maxCacheSize / MaxSegmentSize);
+
+        this.cacheSegments = new ByteBuf[segmentsCount];
+
+        for (int i = 0; i < segmentsCount - 1; i++) {
+            // All intermediate segments will be full-size
+            cacheSegments[i] = Unpooled.directBuffer(MaxSegmentSize, MaxSegmentSize);
+        }
+
+        int lastSegmentSize = (int) (maxCacheSize % MaxSegmentSize);
+        cacheSegments[segmentsCount - 1] = Unpooled.directBuffer(lastSegmentSize, lastSegmentSize);
+    }
+
+    public void clear() {
+        cacheSize.set(0L);
+        cacheCount.reset();
+        index.clear();
+        lastEntryMap.clear();
+        deletedLedgers.clear();
+    }
+
+    @Override
+    public void close() {
+        for (ByteBuf buf : cacheSegments) {
+            buf.release();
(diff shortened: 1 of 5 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/WriteCache.java
```diff
@@ -0,0 +1,278 @@
...
+        // Align to 64 bytes so that different threads will not contend the same L1 cache line
+        int alignedSize = align64(size);
+
+        long offset;
+        int localOffset;
+        int segmentIdx;
+
+        while (true) {
+            offset = cacheSize.getAndAdd(alignedSize);
+            localOffset = (int) (offset & SegmentOffsetMask);
```
ivankelly: The mixture of & and / are a bit weird here. & requires that SegmentOffsetMask be a power of 2, minus 1. This being the case, then getting the segment index should be a bitwise shift, not a division.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/storage/ldb/ArraySortGroupTest.java
```diff
@@ -0,0 +1,157 @@
...
+
+        assertArrayEquals(expectedSorted, data);
+    }
+
+    @Test
+    public void keySmallerThanTotalSize() {
+        try {
+            new ArrayGroupSort(3, 2);
+            fail("should have failed");
+        } catch (IllegalArgumentException e) {
```
eolivelli: Nit: This can be put in @Test(expected=IllegalArgumentException.class)

Same for below checks

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/WriteCache.java
```diff
@@ -0,0 +1,278 @@
...
+        sortedEntriesLock.lock();
+
+        try {
+            int entriesToSort = (int) index.size();
+            int arrayLen = entriesToSort * 4;
+            if (sortedEntries == null || sortedEntries.length < arrayLen) {
+                sortedEntries = new long[(int) (arrayLen * 2)];
+            }
+
+            long startTime = System.nanoTime();
```
ivankelly: Use MathUtils here and in the calculations further down.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/WriteCache.java
```diff
@@ -0,0 +1,278 @@
...
+            startTime = System.nanoTime();
+
+            // Sort entries by (ledgerId, entryId) maintaining the 4 items groups
+            groupSorter.sort(sortedEntries, 0, sortedEntriesIdx);
+            if (log.isDebugEnabled()) {
+                log.debug("sorting {} ms", (System.nanoTime() - startTime) / 1e6);
+            }
+            startTime = System.nanoTime();
+
+            ByteBuf[] entrySegments = new ByteBuf[segmentsCount];
```
ivankelly: i don't understand why you're making this copy here.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/WriteCache.java
```diff
@@ -0,0 +1,278 @@
...
+        if (lastEntryId == -1) {
+            // Ledger not found in write cache
+            return null;
+        } else {
+            return get(ledgerId, lastEntryId);
+        }
+    }
+
+    public void deleteLedger(long ledgerId) {
+        deletedLedgers.add(ledgerId);
```
eolivelli: We can delete a ledger during a foreach. This is handled gracefully using concurrent collections.
But isn't it bad to have a foreach to return only part of the entries in case of a delete happening during the scan?
Maybe it will not be possible I am just asking about this case just to understand

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/storage/ldb/ArraySortGroupTest.java
```diff
@@ -0,0 +1,157 @@
...
+        try {
+            sorter.sort(new long[] { 1, 2 });
+            fail("should have failed");
+        } catch (IllegalArgumentException e) {
+            // ok
+        }
+    }
+
+    @Test
+    public void emtpy() {
```
ivankelly: typo: empty

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/storage/ldb/WriteCacheTest.java
```diff
@@ -0,0 +1,113 @@
...
+        assertTrue(cache.isEmpty());
+        assertEquals(0, cache.count());
+        assertEquals(0, cache.size());
+
+        entry1.release();
+        cache.close();
+    }
+
+    @Test
+    public void cacheFull() throws Exception {
```
ivankelly: Could do with a few more tests.
- ledger deletion when there's more than one ledger.
- Edge cases on buffer/segment sizes
- Multiple threads accessing at the same time.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/WriteCache.java
merlimat: Very good point. I absolutely cannot remember why is this not a `>> 32`. Need to think about it.

merlimat: Actually, I think the result is the same. 

The logic is that the offset keeps increasing when adding each entry. 
If each segment is 2Gb (maxInt), and the current offset is 2.5Gb, the `segmentIdx` will be 1 after the division. The 31 bits shift will be faster though.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/WriteCache.java
merlimat: This is making a copy of the cache buffers for each segment (just the ByteBuf shell objects) so that we can play with the `readerIdx` and `writerIdx` and pass on always the same `ByteBuf` instance to the iteration consumer. See later: 

```java
ByteBuf entry = entrySegments[segmentIdx];
entry.setIndex(localOffset, localOffset + (int) length);
consumer.accept(ledgerId, entryId, entry);
```

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/WriteCache.java
```diff
@@ -0,0 +1,278 @@
...
+public class WriteCache implements Closeable {
+
+    /**
+     * Consumer that is used to scan the entire write cache.
+     */
+    public interface EntryConsumer {
+        void accept(long ledgerId, long entryId, ByteBuf entry);
+    }
+
+    private final ConcurrentLongLongPairHashMap index = new ConcurrentLongLongPairHashMap(4096,
```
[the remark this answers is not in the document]
merlimat: Fixed

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/WriteCache.java
```diff
@@ -0,0 +1,278 @@
...
+            2 * Runtime.getRuntime().availableProcessors());
+
+    private final ConcurrentLongLongHashMap lastEntryMap = new ConcurrentLongLongHashMap(4096,
+            2 * Runtime.getRuntime().availableProcessors());
+
+    private final ByteBuf[] cacheSegments;
+    private final int segmentsCount;
+
+    private static final int MaxSegmentSize = Integer.MAX_VALUE;
```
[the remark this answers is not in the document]
merlimat: Fixed

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/WriteCache.java
merlimat: Sure

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/storage/ldb/ArraySortGroupTest.java
merlimat: Fixed

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/storage/ldb/ArraySortGroupTest.java
merlimat: Fixed

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/storage/ldb/WriteCacheTest.java
merlimat: Done

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/WriteCache.java
merlimat: If the ledger is deleted during a foreach, some entries for that ledger may be flushed and some may not. That is handled anyway when the entryLogs are compacted.

sijie:
@eolivelli : @merlimat addressed your comments. please take a look again.

sijie:
@eolivelli sorry I missed your review :(

sijie:
@merlimat seems there are a few checkstyle errors

PR #829, opened 2017-12-11
DbLedgerStorage -- KeyValue storage interface with RocksDB implementation

merlimat:
Add an interface for a generic key-value storage local library, with the default implementation based on RocksDB.

### DIFF
--- a/bookkeeper-server/pom.xml
+++ b/bookkeeper-server/pom.xml
@@ -40,6 +40,11 @@
       <version>${project.parent.version}</version>
     </dependency>
     <dependency>
+      <groupId>org.rocksdb</groupId>
+      <artifactId>rocksdbjni</artifactId>
+      <version>5.8.6</version>
+    </dependency>
+    <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-log4j12</artifactId>
       <version>${slf4j.version}</version>
--- a/bookkeeper-server/pom.xml
+++ b/bookkeeper-server/pom.xml
@@ -42,7 +42,7 @@
     <dependency>
       <groupId>org.rocksdb</groupId>
       <artifactId>rocksdbjni</artifactId>
-      <version>5.8.6</version>
+      <version>${rocksdb.version}</version>
     </dependency>
     <dependency>
       <groupId>org.slf4j</groupId>
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorage.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorage.java
@@ -1,0 +1,162 @@
+/**
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+package org.apache.bookkeeper.bookie.storage.ldb;
+
+import java.io.Closeable;
+import java.io.IOException;
+import java.util.Map.Entry;
+
+/**
+ * Abstraction of a generic key-value local database.
+ */
+public interface KeyValueStorage extends Closeable {
+
+    void put(byte[] key, byte[] value) throws IOException;
+
+    /**
+     * Get the value associated with the given key.
+     *
+     * @param key
+     *            the key to lookup
+     * @return the value or null if the key was not found
+     */
+    byte[] get(byte[] key) throws IOException;
+
+    /**
+     * Get the value associated with the given key.
+     *
+     * <p>This method will use the provided array store the value
+     *
+     * @param key
+     *            the key to lookup
+     * @param value
+     *            an array where to store the result
+     * @return -1 if the entry was not found or the length of the value
+     * @throws IOException
+     *             if the value array could not hold the result
+     */
+    int get(byte[] key, byte[] value) throws IOException;
+
+    /**
+     * Get the entry whose key is the biggest and it's lesser than the supplied key.
+     *
+     * <p>For example if the db contains :
+     *
+     * <pre>
+     * {
+     *      1 : 'a',
+     *      2 : 'b',
+     *      3 : 'c'
+     * }
+     * </pre>
+     *
+     * <p>Then:
+     *
+     * <pre>
+     * getFloor(3) --> (2, 'b')
+     * </pre>
+     *
+     * @param key
+     *            the non-inclusive upper limit key
+     * @return the entry before or null if there's no entry before key
+     */
+    Entry<byte[], byte[]> getFloor(byte[] key) throws IOException;
+
+    /**
+     * Get the entry whose key is bigger or equal the supplied key.
+     *
+     * @param key
+     * @return
+     * @throws IOException
+     */
+    Entry<byte[], byte[]> getCeil(byte[] key) throws IOException;
+
+    /**
+     *
+     * @param key
+     * @throws IOException
+     */
+    void delete(byte[] key) throws IOException;
+
+    /**
+     * Get an iterator over to scan sequentially through all the keys in the
+     * database.
+     *
+     * @return
+     */
+    CloseableIterator<byte[]> keys();
+
+    /**
+     * Get an iterator over to scan sequentially through all the keys within a
+     * specified range.
+     *
+     * @param firstKey
+     *            the first key in the range (included)
+     * @param lastKey
+     *            the lastKey in the range (not included)
+     *
+     */
+    CloseableIterator<byte[]> keys(byte[] firstKey, byte[] lastKey);
+
+    /**
+     * Return an iterator object that can be used to sequentially scan through all
+     * the entries in the database.
+     */
+    CloseableIterator<Entry<byte[], byte[]>> iterator();
+
+    /**
+     * Commit all pending write to durable storage.
+     */
+    void sync() throws IOException;
+
+    /**
+     * @return the number of keys.
+     */
+    long count() throws IOException;
+
+    /**
+     * Iterator interface.
+     *
+     * @param <T>
+     */
+    interface CloseableIterator<T> extends Closeable {
+        boolean hasNext() throws IOException;
+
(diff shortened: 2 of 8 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageRocksDB.java
```diff
@@ -0,0 +1,541 @@
...
+
+        return new CloseableIterator<byte[]>() {
+            @Override
+            public boolean hasNext() {
+                return iterator.isValid();
+            }
+
+            @Override
+            public byte[] next() {
+                checkArgument(iterator.isValid());
```
eolivelli: checkState?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageRocksDB.java
```diff
@@ -0,0 +1,541 @@
...
+
+        return new CloseableIterator<byte[]>() {
+            @Override
+            public boolean hasNext() {
+                return iterator.isValid() && ByteComparator.compare(iterator.key(), lastKey) < 0;
+            }
+
+            @Override
+            public byte[] next() {
+                checkArgument(iterator.isValid());
```
eolivelli: checkState?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageRocksDB.java
```diff
@@ -0,0 +1,541 @@
...
+            } catch (RocksDBException e) {
+                throw new IOException("Failed to flush RocksDB batch", e);
+            }
+        }
+    }
+
+    private static final class EntryWrapper implements Entry<byte[], byte[]> {
+        // This is not final since the iterator will reuse the same EntryWrapper
+        // instance at each step
+        private byte[] key;
```
eolivelli: Final?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageRocksDB.java
```diff
@@ -0,0 +1,541 @@
...
+        long startTime = System.nanoTime();
+
+        String rocksDbPath = path + ".rocksdb";
+        KeyValueStorage source = new KeyValueStorageRocksDB(path, dbConfigType, conf, true /* read-only */);
+
+        log.info("Opened existing db, starting copy");
+        KeyValueStorage target = new KeyValueStorageRocksDB(rocksDbPath, dbConfigType, conf);
+
+        // Copy into new database. Write in batches to speed up the insertion
+        CloseableIterator<Entry<byte[], byte[]>> iterator = source.iterator();
```
eolivelli: This may fail to be closed?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageRocksDB.java
```diff
@@ -0,0 +1,541 @@
...
+        try {
+            Files.createFile(FileSystems.getDefault().getPath(path, ROCKSDB_MARKER));
+        } catch (FileAlreadyExistsException e) {
+            // Ignore
+        }
+    }
+
+    private static final Comparator<byte[]> ByteComparator = UnsignedBytes.lexicographicalComparator();
+
+    private static final Logger log = LoggerFactory.getLogger(KeyValueStorageRocksDB.class);
```
eolivelli: Does check style pass? I thought it wants constant at the beginning of the class

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageRocksDB.java
merlimat: I've added the comment to clarify why that's not possible in current code.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageRocksDB.java
merlimat: Makes sense

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageRocksDB.java
merlimat: the iterator is closed in the `finally` block. In any case, this part of code was only used in a old migration at Yahoo (older code was using a different format for indexes, or was using LevelDb back in the day, so the in-place conversion was happening as needed). This will not be needed anymore.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageRocksDB.java
merlimat: Yes, I ran checkstyle & findbugs locally.

bookkeeper-server/pom.xml
```diff
@@ -39,6 +39,11 @@
       <artifactId>bookkeeper-proto</artifactId>
       <version>${project.parent.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.rocksdb</groupId>
+      <artifactId>rocksdbjni</artifactId>
+      <version>5.8.6</version>
```
eolivelli: really sorry for missing this at first round.
I think we write versions in properties, like ${rocksdbjni.version}

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageRocksDB.java
```diff
@@ -0,0 +1,447 @@
...
+                tableOptions.setBlockSize(blockSize);
+                tableOptions.setBlockCacheSize(blockCacheSize);
+                tableOptions.setFormatVersion(2);
+                tableOptions.setChecksumType(ChecksumType.kxxHash);
+                if (bloomFilterBitsPerKey > 0) {
+                    tableOptions.setFilter(new BloomFilter(bloomFilterBitsPerKey, false));
+                }
+
+                // Options best suited for HDDs
+                tableOptions.setCacheIndexAndFilterBlocks(true);
```
ivankelly: since these are best suited to HDD, why aren't we giving the user an option to switch them on or not.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageRocksDB.java
```diff
@@ -0,0 +1,447 @@
...
+                iterator.seekToLast();
+                if (iterator.isValid()) {
+                    return new EntryWrapper(iterator.key(), iterator.value());
+                } else {
+                    // Db is empty
+                    return null;
+                }
+            }
+
+            iterator.prev();
```
ivankelly: I would expect floor to return the entry which is less than _or equal_ to the passed key. This always returns the entry which is less than, so it shouldn't really be called Floor.

https://docs.oracle.com/javase/8/docs/api/java/lang/Math.html#floor-double-

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorage.java
```diff
@@ -0,0 +1,162 @@
...
+package org.apache.bookkeeper.bookie.storage.ldb;
+
+import java.io.Closeable;
+import java.io.IOException;
+import java.util.Map.Entry;
+
+/**
+ * Abstraction of a generic key-value local database.
+ */
+public interface KeyValueStorage extends Closeable {
```
ivankelly: Should probably have a javadoc to state that all ops are non-syncing so sync needs to be called to ensure durability.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageRocksDB.java
```diff
@@ -0,0 +1,447 @@
...
+            return value;
+        }
+
+        @Override
+        public byte[] getKey() {
+            return key;
+        }
+    }
+
+    private static final Comparator<byte[]> ByteComparator = UnsignedBytes.lexicographicalComparator();
```
ivankelly: Should be CAPS instead of CamelCase. When you use CamelCase, and make a call to ByteComparator.compare(), it looks like you are calling a static method on the ByteComparator class, which is confusing.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorage.java
```diff
@@ -0,0 +1,162 @@
...
+
+    /**
+     * Iterator interface.
+     *
+     * @param <T>
+     */
+    interface CloseableIterator<T> extends Closeable {
+        boolean hasNext() throws IOException;
+
+        T next() throws IOException;
```
ivankelly: Add a javadoc that the returned value will be reused, so if the user wants to keep it after the subsequent call to next(), they should copy

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorage.java
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    Batch newBatch();
+
+    /**
+     * Interface for a batch to be written in the storage.
+     */
+    public interface Batch extends Closeable {
```
ivankelly: Does closing implicitly flush?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageRocksDB.java
ivankelly: Better than a comment would be to make access to key and value explicitly though a setter, setKeyAndValue

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageRocksDB.java
```diff
@@ -0,0 +1,447 @@
...
+import org.rocksdb.RocksIterator;
+import org.rocksdb.WriteBatch;
+import org.rocksdb.WriteOptions;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * RocksDB based implementation of the KeyValueStorage.
+ */
+public class KeyValueStorageRocksDB implements KeyValueStorage {
```
ivankelly: Are there other implementations also?

Issue #620, opened 2017-10-17
Close opened entry log files when they become idle


ivankelly:
JIRA: https://issues.apache.org/jira/browse/BOOKKEEPER-721

Reporter: Sijie Guo @sijie

we might consider closing opened entry log files when they become idle.

shoothzj:
closing for inactivity. feel free to reopen

PR #832, opened 2017-12-12
Issue 620: Close the fileChannels for read when they are idle

ArvinDevel:

use guava cache to replace concurrentMap for logid2fileChannel, as guava cache has a eviction way, so we can use it to close the idle fileChannel

Master Issue: #620 

> ---
> quickly and easily:
>
>
>
> Otherwise:
> 
> 
> ---

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
@@ -374,9 +374,11 @@ public class EntryLogger {
 
     static class ReferenceCountedFileChannel extends AbstractReferenceCounted {
         private final FileChannel fc;
+        AtomicBoolean dead;
 
         public ReferenceCountedFileChannel(FileChannel fileChannel) {
             this.fc = fileChannel;
+            dead = new AtomicBoolean(false);
         }
 
         @VisibleForTesting
@@ -392,6 +394,7 @@ public class EntryLogger {
         // when the refCnt decreased to 0 or force deallocate
         @Override
         protected void deallocate() {
+            dead.compareAndSet(false, true);
             try {
                 fc.close();
             } catch (IOException e) {
@@ -1172,10 +1175,14 @@ public class EntryLogger {
         ReferenceCountedFileChannel oldFc =
                 logid2FileChannel.putIfAbsent(entryLogId, new ReferenceCountedFileChannel(newFc));
         if (null != oldFc) {
-            newFc.close();
-            newFc = oldFc.fc;
             // increment the refCnt
-            oldFc.retain();
+            // double check to ensure the fileChannel is not closed due to refCnt down to 0.
+            if (oldFc.refCnt() > 0 && !oldFc.dead.get()) {
+                oldFc.retain();
+                newFc.close();
+                newFc = oldFc.fc;
+            }
+            // otherwise the fileChannel is being closed, use newFc
         }
 
         // We set the position of the write buffer of this buffered channel to Long.MAX_VALUE
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
@@ -25,21 +25,17 @@ import static com.google.common.base.Charsets.UTF_8;
 import static org.apache.bookkeeper.bookie.TransactionalEntryLogCompactor.COMPACTING_SUFFIX;
 import static org.apache.bookkeeper.util.BookKeeperConstants.MAX_LOG_SIZE_LIMIT;
 
+import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.CacheLoader;
-import com.google.common.cache.LoadingCache;
-import com.google.common.cache.RemovalListener;
-import com.google.common.cache.RemovalListeners;
-import com.google.common.cache.RemovalNotification;
-import com.google.common.collect.MapMaker;
 import com.google.common.collect.Sets;
-
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufAllocator;
 import io.netty.buffer.PooledByteBufAllocator;
 import io.netty.buffer.Unpooled;
+import io.netty.util.AbstractReferenceCounted;
+import io.netty.util.ReferenceCounted;
 import io.netty.util.concurrent.FastThreadLocal;
-
 import java.io.BufferedReader;
 import java.io.BufferedWriter;
 import java.io.File;
@@ -59,9 +55,10 @@ import java.util.Collections;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
-import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CancellationException;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
@@ -69,7 +66,6 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
-
 import org.apache.bookkeeper.bookie.LedgerDirsManager.LedgerDirsListener;
 import org.apache.bookkeeper.conf.ServerConfiguration;
 import org.apache.bookkeeper.util.IOUtils;
@@ -206,6 +202,7 @@ public class EntryLogger {
     private final boolean doRegularFlushes;
     private long bytesWrittenSinceLastFlush = 0;
     private final int maxSaneEntrySize;
+    private final int expireReadChannelCacheInHour = 1;
 
     final ServerConfiguration conf;
     /**
@@ -335,79 +332,99 @@ public class EntryLogger {
      * A thread-local variable that wraps a mapping of log ids to bufferedchannels
      * These channels should be used only for reading. logChannel is the one
      * that is used for writes.
+     * We use this Guava cache to store the BufferedReadChannel.
+     * When the BufferedReadChannel is removed, the underlying fileChannel's refCnt decrease 1,
+     * temporally use 1h to relax replace after reading.
      */
-    private final ThreadLocal<Map<Long, BufferedReadChannel>> logid2Channel =
-            new ThreadLocal<Map<Long, BufferedReadChannel>>() {
+    private final ThreadLocal<Cache<Long, BufferedReadChannel>> logid2Channel =
+            new ThreadLocal<Cache<Long, BufferedReadChannel>>() {
         @Override
-        public Map<Long, BufferedReadChannel> initialValue() {
+        public Cache<Long, BufferedReadChannel> initialValue() {
             // Since this is thread local there only one modifier
             // We dont really need the concurrency, but we need to use
             // the weak values. Therefore using the concurrency level of 1
-            return new MapMaker().concurrencyLevel(1)
-                .weakValues()
-                .makeMap();
+            return CacheBuilder.newBuilder().concurrencyLevel(1)
+                    .expireAfterAccess(expireReadChannelCacheInHour, TimeUnit.HOURS)
+                    //decrease the refCnt
+                    .removalListener(removal -> logid2FileChannel.get(removal.getKey()).release())
+                    .build(readChannelLoader);
         }
     };
 
-    private final  CacheLoader<Long, FileChannel> loader = new CacheLoader<Long, FileChannel> () {
-        public FileChannel load(Long entryLogId) throws Exception {
+    private final  CacheLoader<Long, BufferedReadChannel> readChannelLoader =
+            new CacheLoader<Long, BufferedReadChannel> () {
+        public BufferedReadChannel load(Long entryLogId) throws Exception {
 
-            File file = findFile(entryLogId);
-            // get channel is used to open an existing entry log file
-            // it would be better to open using read mode
-            return new RandomAccessFile(file, "r").getChannel();
+            return getChannelForLogId(entryLogId);
+
         }
     };
-    // close the file channel, when it was removed from cache
-    private final RemovalListener<Long, FileChannel> removalListener = new RemovalListener<Long, FileChannel>() {
-        public void onRemoval(RemovalNotification<Long, FileChannel> removal) {
-            FileChannel fc = removal.getValue();
-            if (null != fc) {
-                try {
-                    fc.close();
-                } catch (IOException e) {
-                    LOG.warn("Exception while closing channel for log file: {}", removal.getKey());
-                } finally {
-                    IOUtils.close(LOG, fc);
-                }
+
+
+    private static class ReferenceCountedFileChannel extends AbstractReferenceCounted {
+        private final FileChannel fc;
+
+        public ReferenceCountedFileChannel(FileChannel fileChannel) {
+            this.fc = fileChannel;
+        }
+
+
+        @Override
+        public ReferenceCounted touch(Object hint) {
+            return this;
+        }
+
+        // when the refCnt decreased to
+        @Override
+        protected void deallocate() {
+            try {
+                fc.close();
+            } catch (IOException e) {
+                LOG.warn("Exception occurred in ReferenceCountedFileChannel"
+                        + " while closing channel for log file: {}", fc);
+            } finally {
+                IOUtils.close(LOG, fc);
             }
         }
-    };
 
-    private final ExecutorService removeExecutor = Executors.newSingleThreadExecutor();
+    }
+
 
     /**
      * Each thread local buffered read channel can share the same file handle because reads are not relative
-     * and don't cause a change in the channel's position. We use this Guava cache to store the file channels. Each
-     * file channel is mapped to a log id which represents an open log file. When the file channel is removed/idle,
-     * it will be closed. temporally use 1h to relax replace after reading.
+     * and don't cause a change in the channel's position.
+     * Each file channel is mapped to a log id which represents an open log file.
+     * when ReferenceCountedFileChannel's refCnt decrease to 0, close the fileChannel.
      */
-    private LoadingCache<Long, FileChannel> logid2FileChannel =  CacheBuilder.newBuilder()
-                .expireAfterAccess(1, TimeUnit.HOURS)
-                .removalListener(RemovalListeners.asynchronous(removalListener, removeExecutor))
-                .build(loader);
+    private ConcurrentMap<Long, ReferenceCountedFileChannel>
+            logid2FileChannel = new ConcurrentHashMap<>();
+
 
     /**
      * Put the logId, bc pair in the map responsible for the current thread.
      * @param logId
      * @param bc
      */
-    public BufferedReadChannel putInReadChannels(long logId, BufferedReadChannel bc) {
-        Map<Long, BufferedReadChannel> threadMap = logid2Channel.get();
-        return threadMap.put(logId, bc);
+    public void putInReadChannels(long logId, BufferedReadChannel bc) {
+        Cache<Long, BufferedReadChannel> threadCahe = logid2Channel.get();
+        threadCahe.put(logId, bc);
     }
 
     /**
      * Remove all entries for this log file in each thread's cache.
-     * the guava cache will close the fileChannel automatically.
      * @param logId
      */
     public void removeFromChannelsAndClose(long logId) {
-        logid2FileChannel.invalidate(logId);
+        //remove the fileChannel from logId2Channel
+        Cache<Long, BufferedReadChannel> threadCahe = logid2Channel.get();
+        threadCahe.invalidate(logId);
+
+        //remove the fileChannel from logId2FileChannel
+        logid2FileChannel.remove(logId);
     }
 
     public BufferedReadChannel getFromChannels(long logId) {
-        return logid2Channel.get().get(logId);
+        return logid2Channel.get().getIfPresent(logId);
     }
 
     /**
@@ -1124,21 +1141,20 @@ public class EntryLogger {
         }
     }
 
+
     private BufferedReadChannel getChannelForLogId(long entryLogId) throws IOException {
         BufferedReadChannel brc = getFromChannels(entryLogId);
         if (brc != null) {
+            // increment the refCnt
+            logid2FileChannel.get(entryLogId).retain();
             return brc;
         }
 
-        // logid2FileChannel cache will load fileChannel automatically
-        FileChannel fc = null;
-        try {
-            fc = logid2FileChannel.get(entryLogId);
-        } catch (ExecutionException e){
-            LOG.error("ExecutionException found in get fileChannel for log {} in logid2FileChannel cache", entryLogId);
-            // throw exception to avoid pass null to BufferedReadChannel
-            throw new IOException(e);
-        }
+        File file = findFile(entryLogId);
+        // get channel is used to open an existing entry log file
+        // it would be better to open using read mode
+        FileChannel fc = new RandomAccessFile(file, "r").getChannel();
+        logid2FileChannel.put(entryLogId, new ReferenceCountedFileChannel(fc));
 
         // We set the position of the write buffer of this buffered channel to Long.MAX_VALUE
         // so that there are no overlaps with the write buffer while reading
@@ -1395,8 +1411,10 @@ public class EntryLogger {
         LOG.info("Stopping EntryLogger");
         try {
             flush();
-            //invalidateAll and close corresponding fileChannel
-            logid2FileChannel.invalidateAll();
+            //close corresponding fileChannel
+            for (ReferenceCountedFileChannel rfc : logid2FileChannel.values()) {
+                rfc.deallocate();
+            }
             // close current writing log file
             closeFileChannel(logChannel);
             synchronized (compactionLogLock) {
@@ -1408,9 +1426,8 @@ public class EntryLogger {
             // we have no idea how to avoid io exception during shutting down, so just ignore it
             LOG.error("Error flush entry log during shutting down, which may cause entry log corrupted.", ie);
         } finally {
-            removeExecutor.shutdown();
-            for (FileChannel fc : logid2FileChannel.asMap().values()) {
-                IOUtils.close(LOG, fc);
+            for (ReferenceCountedFileChannel rfc : logid2FileChannel.values()) {
+                IOUtils.close(LOG, rfc.fc);
             }
             forceCloseFileChannel(logChannel);
             synchronized (compactionLogLock) {
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
@@ -25,6 +25,7 @@ import static com.google.common.base.Charsets.UTF_8;
 import static org.apache.bookkeeper.bookie.TransactionalEntryLogCompactor.COMPACTING_SUFFIX;
 import static org.apache.bookkeeper.util.BookKeeperConstants.MAX_LOG_SIZE_LIMIT;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.CacheLoader;
@@ -368,16 +369,16 @@ public class EntryLogger {
 
 
     static class ReferenceCountedFileChannel extends AbstractReferenceCounted {
-        FileChannel getFc() {
-            return fc;
-        }
-
         private final FileChannel fc;
 
         public ReferenceCountedFileChannel(FileChannel fileChannel) {
             this.fc = fileChannel;
         }
 
+        @VisibleForTesting
+        FileChannel getFileChannel(){
+            return fc;
+        }
 
         @Override
         public ReferenceCounted touch(Object hint) {
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
@@ -54,10 +54,8 @@ import java.util.Collections;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
-import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CancellationException;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
@@ -65,7 +63,6 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.atomic.AtomicInteger;
 import org.apache.bookkeeper.bookie.LedgerDirsManager.LedgerDirsListener;
 import org.apache.bookkeeper.conf.ServerConfiguration;
 import org.apache.bookkeeper.util.IOUtils;
@@ -337,7 +334,7 @@ public class EntryLogger {
      * When the BufferedReadChannel is removed, the underlying fileChannel's refCnt decrease 1,
      * temporally use 1h to relax replace after reading.
      */
-    private final ThreadLocal<Cache<Long, BufferedReadChannel>> logid2Channel =
+    private final ThreadLocal<Cache<Long, BufferedReadChannel>> logid2ReadChannel =
             new ThreadLocal<Cache<Long, BufferedReadChannel>>() {
         @Override
         public Cache<Long, BufferedReadChannel> initialValue() {
@@ -347,7 +344,7 @@ public class EntryLogger {
             return CacheBuilder.newBuilder().concurrencyLevel(1)
                     .expireAfterAccess(readChannelCacheExpireTimeMs, TimeUnit.MILLISECONDS)
                     //decrease the refCnt
-                    .removalListener(removal -> logid2FileChannel.get((Long) removal.getKey()).release())
+                    .removalListener(removal -> fileChannelBackingCache.get((Long) removal.getKey()).release())
                     .build(readChannelLoader);
         }
     };
@@ -372,163 +369,14 @@ public class EntryLogger {
     };
 
     /**
-     * FileChannelBackingCache used to cache RefCntFileChannels for read.
-     * In order to avoid get released file, adopt design of FileInfoBackingCache.
-     * @see FileInfoBackingCache
-     */
-    class FileChannelBackingCache {
-        static final int DEAD_REF = -0xdead;
-
-        final ConcurrentHashMap<Long, CachedFileChannel> fileChannels = new ConcurrentHashMap<>();
-
-        CachedFileChannel loadFileChannel(long logId) throws IOException {
-            CachedFileChannel cachedFileChannel = fileChannels.get(logId);
-            if (cachedFileChannel != null) {
-                boolean retained = cachedFileChannel.tryRetain();
-                assert(retained);
-                return cachedFileChannel;
-            }
-            File file = findFile(logId);
-            // get channel is used to open an existing entry log file
-            // it would be better to open using read mode
-            FileChannel newFc = new RandomAccessFile(file, "r").getChannel();
-            cachedFileChannel = new CachedFileChannel(logId, newFc);
-            fileChannels.put(logId, cachedFileChannel);
-            boolean retained = cachedFileChannel.tryRetain();
-            assert(retained);
-            return cachedFileChannel;
-        }
-
-        /**
-         * close FileChannel and remove from cache when possible.
-         * @param logId
-         * @param fc
-         */
-        private void releaseFileChannel(long logId, CachedFileChannel fc) {
-            if (fc.markDead()) {
-                try {
-                    fc.fc.close();
-                } catch (IOException e) {
-                    LOG.warn("Exception occurred in ReferenceCountedFileChannel"
-                            + " while closing channel for log file: {}", fc);
-                } finally {
-                    IOUtils.close(LOG, fc.fc);
-                }
-                fileChannels.remove(logId);
-            }
-        }
-
-        /**
-         * Remove all entries for this log file in each thread's cache.
-         * @param logId
-         */
-        public void removeFromChannelsAndClose(long logId) {
-            //remove the fileChannel from FileChannelBackingCache and close it
-            CachedFileChannel fileChannel = fileChannels.remove(logId);
-            fileChannel.release();
-            try {
-                fileChannel.fc.close();
-            } catch (IOException e) {
-                LOG.warn("Exception occurred in CachedFileChannel"
-                        + " while closing channel for log file: {}", logId);
-            } finally {
-                IOUtils.close(LOG, fileChannel.fc);
-            }
-        }
-
-        void closeAllFileChannels() throws IOException {
-            for (Map.Entry<Long, CachedFileChannel> entry : fileChannels.entrySet()) {
-                entry.getValue().fc.close();
-            }
-        }
-
-        public CachedFileChannel get(Long logId) {
-            return fileChannels.get(logId);
-        }
-
-        class CachedFileChannel {
-            private final FileChannel fc;
-            private final long entryLogId;
-            final AtomicInteger refCount;
-
-            CachedFileChannel(long entryLogId, FileChannel fileChannel) {
-                this.entryLogId = entryLogId;
-                this.fc = fileChannel;
-                this.refCount = new AtomicInteger(0);
-            }
-
-            /**
-             * Mark this fileinfo as dead. We can only mark a fileinfo as
-             * dead if noone currently holds a reference to it.
-             *
-             * @return true if we marked as dead, false otherwise
-             */
-            private boolean markDead() {
-                return refCount.compareAndSet(0, DEAD_REF);
-            }
-
-            /**
-             * Attempt to retain the file info.
-             * When a client obtains a fileinfo from a container object,
-             * but that container object may release the fileinfo before
-             * the client has a chance to call retain. In this case, the
-             * file info could be released and the destroyed before we ever
-             * get a chance to use it.
-             *
-             * <p>tryRetain avoids this problem, by doing a compare-and-swap on
-             * the reference count. If the refCount is negative, it means that
-             * the fileinfo is being cleaned up, and this fileinfo object should
-             * not be used. This works in tandem with #markDead, which will only
-             * set the refCount to negative if noone currently has it retained
-             * (i.e. the refCount is 0).
-             *
-             * @return true if we managed to increment the refcount, false otherwise
-             */
-            boolean tryRetain() {
-                while (true) {
-                    int count = refCount.get();
-                    if (count < 0) {
-                        return false;
-                    } else if (refCount.compareAndSet(count, count + 1)) {
-                        return true;
-                    }
-                }
-            }
-
-            @VisibleForTesting
-            int getRefCount() {
-                return refCount.get();
-            }
-
-            @VisibleForTesting
-            FileChannel getFileChannel() {
-                return fc;
-            }
-
-            void release() {
-                if (refCount.decrementAndGet() == 0) {
-                    releaseFileChannel(entryLogId, this);
-                }
-            }
-
-            @Override
-            public String toString() {
-                return "CachedFileChannel(logId=" + entryLogId
-                        + ",refCount=" + refCount.get()
-                        + ",id=" + System.identityHashCode(this) + ")";
-            }
-        }
-    }
-
-    /**
      * Each thread local buffered read channel can share the same file handle because reads are not relative
      * and don't cause a change in the channel's position.
      * Each file channel is mapped to a log id which represents an open log file.
      */
-    private FileChannelBackingCache logid2FileChannel = new FileChannelBackingCache();
+    private FileChannelBackingCache fileChannelBackingCache = new FileChannelBackingCache(this::findFile);
     @VisibleForTesting
-    FileChannelBackingCache getLogid2FileChannel() {
-        return logid2FileChannel;
+    FileChannelBackingCache getFileChannelBackingCache() {
+        return fileChannelBackingCache;
     }
 
     /**
@@ -537,12 +385,12 @@ public class EntryLogger {
      * @param bc
      */
     public void putInReadChannels(long logId, BufferedReadChannel bc) {
-        Cache<Long, BufferedReadChannel> threadCache = logid2Channel.get();
+        Cache<Long, BufferedReadChannel> threadCache = logid2ReadChannel.get();
         threadCache.put(logId, bc);
     }
 
-    public BufferedReadChannel getFromChannels(long logId) {
-        return logid2Channel.get().getIfPresent(logId);
+    BufferedReadChannel getFromChannels(long logId) {
+        return logid2ReadChannel.get().getIfPresent(logId);
     }
 
     /**
@@ -869,7 +717,7 @@ public class EntryLogger {
      *          Entry Log File Id
      */
     protected boolean removeEntryLog(long entryLogId) {
-        logid2FileChannel.removeFromChannelsAndClose(entryLogId);
+        fileChannelBackingCache.removeFromChannelsAndClose(entryLogId);
         File entryLogFile;
         try {
             entryLogFile = findFile(entryLogId);
@@ -1267,7 +1115,7 @@ public class EntryLogger {
         FileChannelBackingCache.CachedFileChannel cachedFileChannel = null;
         try {
             do {
-                cachedFileChannel = logid2FileChannel.loadFileChannel(entryLogId);
+                cachedFileChannel = fileChannelBackingCache.loadFileChannel(entryLogId);
             } while (!cachedFileChannel.tryRetain());
         } finally {
             if (null != cachedFileChannel) {
@@ -1276,7 +1124,7 @@ public class EntryLogger {
         }
         // We set the position of the write buffer of this buffered channel to Long.MAX_VALUE
         // so that there are no overlaps with the write buffer while reading
-        brc = new BufferedReadChannel(cachedFileChannel.fc, conf.getReadBufferBytes());
+        brc = new BufferedReadChannel(cachedFileChannel.fileChannel, conf.getReadBufferBytes());
         putInReadChannels(entryLogId, brc);
         return brc;
     }
@@ -1529,7 +1377,7 @@ public class EntryLogger {
         LOG.info("Stopping EntryLogger");
         try {
             //close corresponding fileChannels for read
-            logid2FileChannel.closeAllFileChannels();
+            fileChannelBackingCache.closeAllFileChannels();
             flush();
             // close current writing log file
             closeFileChannel(logChannel);
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
@@ -419,8 +419,9 @@ public class EntryLogger {
         Cache<Long, BufferedReadChannel> threadCahe = logid2Channel.get();
         threadCahe.invalidate(logId);
 
-        //remove the fileChannel from logId2FileChannel
-        logid2FileChannel.remove(logId);
+        //remove the fileChannel from logId2FileChannel and close it
+        logid2FileChannel.remove(logId).deallocate();
+
     }
 
     public BufferedReadChannel getFromChannels(long logId) {
@@ -1415,6 +1416,8 @@ public class EntryLogger {
             for (ReferenceCountedFileChannel rfc : logid2FileChannel.values()) {
                 rfc.deallocate();
             }
+            // clear the mapping, so we don't need to go through the channels again in finally block in normal case.
+            logid2FileChannel.clear();
             // close current writing log file
             closeFileChannel(logChannel);
             synchronized (compactionLogLock) {
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
@@ -372,7 +372,8 @@ public class EntryLogger {
     };
 
     /**
-     * FileChannelBackingCache, to avoid get released file, adopt design of FileInfoBackingCache.
+     * FileChannelBackingCache used to cache RefCntFileChannels for read.
+     * In order to avoid get released file, adopt design of FileInfoBackingCache.
      * @see FileInfoBackingCache
      */
     class FileChannelBackingCache {
@@ -435,7 +436,7 @@ public class EntryLogger {
             }
         }
 
-        void closeAllWithoutFlushing() throws IOException {
+        void closeAllFileChannels() throws IOException {
             for (Map.Entry<Long, CachedFileChannel> entry : fileChannels.entrySet()) {
                 entry.getValue().fc.close();
             }
@@ -1263,11 +1264,16 @@ public class EntryLogger {
         if (brc != null) {
             return brc;
         }
-
-        FileChannelBackingCache.CachedFileChannel cachedFileChannel = logid2FileChannel.loadFileChannel(entryLogId);
-//        do {
-//            LOG.info("tryRetain to avoid cachedFileChannel invalidated by other thread.");
-//        } while (!cachedFileChannel.tryRetain());
+        FileChannelBackingCache.CachedFileChannel cachedFileChannel = null;
+        try {
+            do {
+                cachedFileChannel = logid2FileChannel.loadFileChannel(entryLogId);
+            } while (!cachedFileChannel.tryRetain());
+        } finally {
+            if (null != cachedFileChannel) {
+                cachedFileChannel.release();
+            }
+        }
         // We set the position of the write buffer of this buffered channel to Long.MAX_VALUE
         // so that there are no overlaps with the write buffer while reading
         brc = new BufferedReadChannel(cachedFileChannel.fc, conf.getReadBufferBytes());
@@ -1522,9 +1528,9 @@ public class EntryLogger {
         // since logChannel is buffered channel, do flush when shutting down
         LOG.info("Stopping EntryLogger");
         try {
+            //close corresponding fileChannels for read
+            logid2FileChannel.closeAllFileChannels();
             flush();
-            //close corresponding fileChannel
-            logid2FileChannel.closeAllWithoutFlushing();
             // close current writing log file
             closeFileChannel(logChannel);
             synchronized (compactionLogLock) {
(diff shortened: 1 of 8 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -341,12 +346,43 @@ private int readFromLogChannel(long entryLogId, BufferedReadChannel channel, Byt
...
      * Each thread local buffered read channel can share the same file handle because reads are not relative
-     * and don't cause a change in the channel's position. We use this map to store the file channels. Each
-     * file channel is mapped to a log id which represents an open log file.
+     * and don't cause a change in the channel's position. We use this Guava cache to store the file channels. Each
+     * file channel is mapped to a log id which represents an open log file. When the file channel is removed/idle,
+     * it will be closed. temporally use 1h to relax replace after reading.
      */
-    private final ConcurrentMap<Long, FileChannel> logid2FileChannel = new ConcurrentHashMap<Long, FileChannel>();
+    private LoadingCache<Long, FileChannel> logid2FileChannel =  CacheBuilder.newBuilder()
+                .expireAfterAccess(1, TimeUnit.HOURS)
```
eolivelli: @sijie  should be this value configurable ?
Maybe having a configuration value will let us write some test

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -1079,24 +1109,26 @@ private Header getHeaderForLogId(long entryLogId) throws IOException {
...
         }
+
+        // logid2FileChannel cache will load fileChannel automatically
+        FileChannel fc = null;
+        try {
+            fc = logid2FileChannel.get(entryLogId);
+        } catch (ExecutionException e){
+            LOG.error("ExecutionException found in get fileChannel for log {} in logid2FileChannel cache", entryLogId);
+            // throw exception to avoid pass null to BufferedReadChannel
+            throw new IOException(e);
```
sijie: ExecutionException wrap the actual cause on loading the file channel. so you need to unwrap this.
```
if (e.getCause() instanceof IOException) {
    throw (IOException) e.getCause();
} else {
    throw new IOException("Encountered unknown exception on opening read channel for entry log " + entryLogId, e.getCause());
}
```

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -341,12 +346,43 @@ private int readFromLogChannel(long entryLogId, BufferedReadChannel channel, Byt
         }
     };
 
+    private final  CacheLoader<Long, FileChannel> loader = new CacheLoader<Long, FileChannel> () {
+        public FileChannel load(Long entryLogId) throws Exception {
```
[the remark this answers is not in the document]
sijie: I am not sure if you can use lambda for `CacheLoader` here, it is not an interface. CacheLoader is an abstract class

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -341,12 +346,43 @@ private int readFromLogChannel(long entryLogId, BufferedReadChannel channel, Byt
         }
     };
 
+    private final  CacheLoader<Long, FileChannel> loader = new CacheLoader<Long, FileChannel> () {
```
sijie: I would suggest renaming this to a more meaningful name, like `readonlyFileChannelLoader`.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -341,12 +346,43 @@ private int readFromLogChannel(long entryLogId, BufferedReadChannel channel, Byt
...
+        public FileChannel load(Long entryLogId) throws Exception {
+
+            File file = findFile(entryLogId);
+            // get channel is used to open an existing entry log file
+            // it would be better to open using read mode
+            return new RandomAccessFile(file, "r").getChannel();
+        }
+    };
+    // close the file channel, when it was removed from cache
+    private final RemovalListener<Long, FileChannel> removalListener = new RemovalListener<Long, FileChannel>() {
```
sijie: rename this to 'readonlyFileChannelRemovalListener'

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
sijie: +1 for making this configurable.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -341,12 +346,43 @@ private int readFromLogChannel(long entryLogId, BufferedReadChannel channel, Byt
...
-     * and don't cause a change in the channel's position. We use this map to store the file channels. Each
-     * file channel is mapped to a log id which represents an open log file.
+     * and don't cause a change in the channel's position. We use this Guava cache to store the file channels. Each
+     * file channel is mapped to a log id which represents an open log file. When the file channel is removed/idle,
+     * it will be closed. temporally use 1h to relax replace after reading.
      */
-    private final ConcurrentMap<Long, FileChannel> logid2FileChannel = new ConcurrentHashMap<Long, FileChannel>();
+    private LoadingCache<Long, FileChannel> logid2FileChannel =  CacheBuilder.newBuilder()
+                .expireAfterAccess(1, TimeUnit.HOURS)
+                .removalListener(RemovalListeners.asynchronous(removalListener, removeExecutor))
```
sijie: I don't think you need to use asynchronous removal listener here. because this would somehow change the behavior of `close` and `removeFromChannelsAndClose`. I would suggest using removalListener here, so the behavior will be the same as what it was before when a filechannel is removed and closed.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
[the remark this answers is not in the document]
ArvinDevel: @eolivelli  Just as @sijie said, can't use lambda due to it's an abstract class

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -70,6 +72,9 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+
```
sijie: remove these 3 blank lines

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -327,54 +333,100 @@ private int readFromLogChannel(long entryLogId, BufferedReadChannel channel, Byt
...
             // Since this is thread local there only one modifier
             // We dont really need the concurrency, but we need to use
             // the weak values. Therefore using the concurrency level of 1
-            return new MapMaker().concurrencyLevel(1)
-                .weakValues()
-                .makeMap();
+            return CacheBuilder.newBuilder().concurrencyLevel(1)
+                    .expireAfterAccess(expireReadChannelCacheInHour, TimeUnit.HOURS)
+                    //decrease the refCnt
+                    .removalListener(removal -> logid2FileChannel.get(removal.getKey()).release())
```
sijie: it might be good to check if the returned value is null or not.

```
ReferenceCountUtils.release(logid2FileChannel.get(removal.getKey());
```

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -327,54 +333,100 @@ private int readFromLogChannel(long entryLogId, BufferedReadChannel channel, Byt
...
     public void removeFromChannelsAndClose(long logId) {
-        FileChannel fileChannel = logid2FileChannel.remove(logId);
-        if (null != fileChannel) {
-            try {
-                fileChannel.close();
-            } catch (IOException e) {
-                LOG.warn("Exception while closing channel for log file:" + logId);
-            }
-        }
+        //remove the fileChannel from logId2Channel
```
sijie: I don't think you need to invalidate the local thread cache. The original behavior is to close the file channel directly, the local thread cache might be still referencing this file channel, but it is okay since the file channel is anyway closed.

I would suggest

- add a method called `forceCloseFileChannel()` in ReferenceCountedFileChannel
- deallocate() call this `forceCloseFileChannel()`
- in here, you just remove it from `logid2FileChannel` and ``forceCloseFileChannel()`  it.

```
ReferenceCountedFileChannel fileChannel = logid2FileChannel.remove(logId);
if (null != fileChannel) {
    fileChannel.forceCloseFileChannel();
}
```

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -1078,25 +1130,26 @@ private Header getHeaderForLogId(long entryLogId) throws IOException {
         }
     }
 
+
```
sijie: remove blank line

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -1078,25 +1130,26 @@ private Header getHeaderForLogId(long entryLogId) throws IOException {
...
         // get channel is used to open an existing entry log file
         // it would be better to open using read mode
-        FileChannel newFc = new RandomAccessFile(file, "r").getChannel();
-        FileChannel oldFc = logid2FileChannel.putIfAbsent(entryLogId, newFc);
-        if (null != oldFc) {
-            newFc.close();
-            newFc = oldFc;
-        }
+        FileChannel fc = new RandomAccessFile(file, "r").getChannel();
+        logid2FileChannel.put(entryLogId, new ReferenceCountedFileChannel(fc));
```
sijie: the logic should be similar as before. you need to use `putIfAbsent` for concurrent operations.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -1316,8 +1369,9 @@ public void shutdown() {
         LOG.info("Stopping EntryLogger");
         try {
             flush();
-            for (FileChannel fc : logid2FileChannel.values()) {
-                fc.close();
+            //close corresponding fileChannel
+            for (ReferenceCountedFileChannel rfc : logid2FileChannel.values()) {
+                rfc.deallocate();
```
sijie: `forceClose`

sijie:
@ArvinDevel left a few more comments on your latest changes. can you also add a test case for it?

ArvinDevel:
@eolivelli @sijie  Any other suggestions?

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
```diff
@@ -315,6 +314,93 @@ public void testRecoverFromLedgersMapOnV0EntryLog() throws Exception {
...
+        assertNull(cacheThreadLocal.get().getIfPresent(1L));
+        // the cache hasn't readChannel for 2.log
+        assertNull(cacheThreadLocal.get().getIfPresent(2L));
+        // the corresponding file channel should be closed
+        LOG.info("map content is {}", logid2FileChannel.toString());
+        assertEquals(0, logid2FileChannel.get(1L).refCnt());
+        assertEquals(0, logid2FileChannel.get(2L).refCnt());
+        assertEquals(1, logid2FileChannel.get(3L).refCnt());
+        assertEquals(1, logid2FileChannel.get(4L).refCnt());
+//        assertNull(logid2FileChannel.get(2L).getFc());
```
sijie: remove this line if it is not needed

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
```diff
@@ -239,6 +240,27 @@ public ServerConfiguration setEntryLogFilePreAllocationEnabled(boolean enabled)
...
+    }
+
+    /**
+     * set ReadChannelCache expire time. Default value is 1h.
+     *
+     * @param millis
+     *          expire time.
+     * @return server configuration object.
+     */
+    public ServerConfiguration setExpireReadChannelCache(long millis) {
```
sijie: it is better to call out "time" and "ms" in the setting, like READ_CHANNEL_CACHE_EXPIRE_TIME_MS

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
sijie: You can use this one: https://netty.io/4.0/api/io/netty/util/ReferenceCountUtil.html

sijie:
> there is a question, how to verify the fileChannel is closed? currently I checked the refCnt in test.

A tricky solution is to attempt to write data to a file channel. If a file channel is closed, the write will be thrown with [ChannelClosedException](https://docs.oracle.com/javase/7/docs/api/java/nio/channels/ClosedChannelException.html)

ArvinDevel:
@sijie @eolivelli have fixed comments

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
```diff
@@ -47,7 +47,7 @@
     protected static final String COMPACTION_RATE = "compactionRate";
     protected static final String COMPACTION_RATE_BY_ENTRIES = "compactionRateByEntries";
     protected static final String COMPACTION_RATE_BY_BYTES = "compactionRateByBytes";
-    protected static final String EXPIRE_READ_CHANNEL_CACHE = "expireReadChannelCache";
+    protected static final String READ_CHANNEL_CACHE_EXPIRE_TIME_MS = "expireReadChannelCache";
```
sijie: you need to change the setting key "readChannelCacheExpireTimeMs"

sijie:
@eolivelli please also review it since you are involved in this pull request before.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -330,54 +334,111 @@ private int readFromLogChannel(long entryLogId, BufferedReadChannel channel, Byt
...
-     * file channel is mapped to a log id which represents an open log file.
+     * and don't cause a change in the channel's position.
+     * Each file channel is mapped to a log id which represents an open log file.
+     * when ReferenceCountedFileChannel's refCnt decrease to 0, close the fileChannel.
      */
-    private final ConcurrentMap<Long, FileChannel> logid2FileChannel = new ConcurrentHashMap<Long, FileChannel>();
+    private ConcurrentMap<Long, ReferenceCountedFileChannel>
+            logid2FileChannel = new ConcurrentHashMap<>();
+    // only for test.
+    ConcurrentMap<Long, ReferenceCountedFileChannel> getLogid2FileChannel() {
```
eolivelli: add VisibleForTesting

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -1091,24 +1152,30 @@ private Header getHeaderForLogId(long entryLogId) throws IOException {
...
         }
+
         // We set the position of the write buffer of this buffered channel to Long.MAX_VALUE
         // so that there are no overlaps with the write buffer while reading
-        fc = new BufferedReadChannel(newFc, conf.getReadBufferBytes());
-        putInReadChannels(entryLogId, fc);
-        return fc;
+        brc = new BufferedReadChannel(newFc, conf.getReadBufferBytes());
+        putInReadChannels(entryLogId, brc);
+        LOG.info("put readChannel: {}, corresponding to: {} ", brc, entryLogId);
```
eolivelli: drop debug ?

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
```diff
@@ -315,6 +316,104 @@ public void testRecoverFromLedgersMapOnV0EntryLog() throws Exception {
...
+        conf.setEntryLogFilePreAllocationEnabled(false);
+        // below one will cost logId 0
+        Bookie bookie = new Bookie(conf);
+        // create some entries
+        int numLogs = 4;
+        int numEntries = 10;
+        long[][] positions = new long[numLogs][];
+        for (int i = 0; i < numLogs; i++) {
+            positions[i] = new long[numEntries];
+            EntryLogger logger = new EntryLogger(conf,
```
eolivelli: should we shutdown this EntryLogger ?

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
```diff
@@ -315,6 +316,104 @@ public void testRecoverFromLedgersMapOnV0EntryLog() throws Exception {
...
+        assertNotNull(cacheThreadLocal.get().getIfPresent(1L));
+        for (int j = 0; j < numEntries; j++) {
+            logger.readEntry(1, j, positions[1][j]);
+        }
+        LOG.info("cache size is {}, content is {}", cacheThreadLocal.get().size(),
+                cacheThreadLocal.get().asMap().toString());
+        // the cache has readChannel for 2.log
+        assertNotNull(cacheThreadLocal.get().getIfPresent(2L));
+        // expire time
+        Thread.sleep(1000);
```
eolivelli: is there a way not to use a blind sleep ? but wait in a loop until a condition is met or a timeout fires ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
ArvinDevel: thanks,fixed

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
ArvinDevel: that would be good.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
ArvinDevel: ExpireTime semantic is about a period, I think timeout is just like sleep.

sijie: @ArvinDevel : guava cache has a mock ticker that allows you to advance time to trigger expriation.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -330,54 +334,111 @@ private int readFromLogChannel(long entryLogId, BufferedReadChannel channel, Byt
...
+            return CacheBuilder.newBuilder().concurrencyLevel(1)
+                    .expireAfterAccess(expireReadChannelCache, TimeUnit.MILLISECONDS)
+                    //decrease the refCnt
+                    .removalListener(removal -> logid2FileChannel.get(removal.getKey()).release())
+                    .build(readChannelLoader);
         }
     };
 
+    // only used for test.
+    ThreadLocal<Cache<Long, BufferedReadChannel>> getLogid2Channel() {
```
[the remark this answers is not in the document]
sijie: accessing internal state is okay if you are just testing the specific test case. I don't see a reason we will change it in future. 

@ArvinDevel can you add `@VisibleForTesting`?

ArvinDevel:
@sijie @eolivelli please review the latest changes.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -1091,24 +1156,29 @@ private Header getHeaderForLogId(long entryLogId) throws IOException {
...
         FileChannel newFc = new RandomAccessFile(file, "r").getChannel();
-        FileChannel oldFc = logid2FileChannel.putIfAbsent(entryLogId, newFc);
+        ReferenceCountedFileChannel oldFc =
+                logid2FileChannel.putIfAbsent(entryLogId, new ReferenceCountedFileChannel(newFc));
         if (null != oldFc) {
             newFc.close();
-            newFc = oldFc;
+            newFc = oldFc.fc;
+            // increment the refCnt
+            oldFc.retain();
```
ivankelly: It's possible that oldFc could have been released between getting it from logid2FileChannel and calling retain, so the channel returned could be closed.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -330,54 +334,115 @@ private int readFromLogChannel(long entryLogId, BufferedReadChannel channel, Byt
...
             // We dont really need the concurrency, but we need to use
             // the weak values. Therefore using the concurrency level of 1
-            return new MapMaker().concurrencyLevel(1)
-                .weakValues()
-                .makeMap();
+            return CacheBuilder.newBuilder().concurrencyLevel(1)
+                    .expireAfterAccess(readChannelCacheExpireTimeMs, TimeUnit.MILLISECONDS)
+                    //decrease the refCnt
+                    .removalListener(removal -> logid2FileChannel.get(removal.getKey()).release())
+                    .build(readChannelLoader);
```
ivankelly: build(this::getChannelForLogId) should work here, i think, so you don't have to define the loader at all.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -330,54 +334,115 @@ private int readFromLogChannel(long entryLogId, BufferedReadChannel channel, Byt
...
+        @Override
+        public ReferenceCounted touch(Object hint) {
+            return this;
+        }
+
+        // when the refCnt decreased to 0 or force deallocate
+        @Override
+        protected void deallocate() {
+            try {
+                fc.close();
```
ivankelly: we close when refcount hits 0. When is this removed from the map though?

ArvinDevel: currently the ReferenceCountedFileChannel was not removed from the map, use guava cache to finish this would be better.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
ArvinDevel: you're right, I'll consider this later, thanks.

sijie:
@ArvinDevel did you have time to address @ivankelly 's comments?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
ArvinDevel: @ivankelly I have fixed your comment. I'm not sure whether it still has race condition, please has a review. If it still has race condition, I'll try use your design in #913. And the last resort can be removing hashMap and use multiple fileChannel directly.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
ArvinDevel: this loader is not an interface, so I can't use this functional style.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
ArvinDevel: I describe the design briefly: `refCnt` in AbstractReferenceCounted is volatile and `dead` is AtomicBoolean. Assume current refCnt is one and thread A is release()ing and thread B is retain()ing. If thread A is executed before thread B, as long as dead is changed, thread B will use the new fileChannel in case of using closed one.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -1091,24 +1159,33 @@ private Header getHeaderForLogId(long entryLogId) throws IOException {
...
         FileChannel newFc = new RandomAccessFile(file, "r").getChannel();
-        FileChannel oldFc = logid2FileChannel.putIfAbsent(entryLogId, newFc);
+        ReferenceCountedFileChannel oldFc =
+                logid2FileChannel.putIfAbsent(entryLogId, new ReferenceCountedFileChannel(newFc));
         if (null != oldFc) {
-            newFc.close();
-            newFc = oldFc;
+            // increment the refCnt
+            // double check to ensure the fileChannel is not closed due to refCnt down to 0.
+            if (oldFc.refCnt() > 0 && !oldFc.dead.get()) {
```
ivankelly: A cache on another thread could have released between L1176 and L1177, invalidating the oldFc object.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -1091,24 +1159,33 @@ private Header getHeaderForLogId(long entryLogId) throws IOException {
...
+            return brc;
         }
+
         File file = findFile(entryLogId);
         // get channel is used to open an existing entry log file
         // it would be better to open using read mode
         FileChannel newFc = new RandomAccessFile(file, "r").getChannel();
-        FileChannel oldFc = logid2FileChannel.putIfAbsent(entryLogId, newFc);
+        ReferenceCountedFileChannel oldFc =
+                logid2FileChannel.putIfAbsent(entryLogId, new ReferenceCountedFileChannel(newFc));
```
ivankelly: I don't see anywhere that entries are removed from logid2FileChannel.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
sijie: @ivankelly I don't think it is worth on stucking here whether to use lambda or not. thoughts?

sijie:
@ArvinDevel can you address @ivankelly 's comments?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
ivankelly: it's a minor thing. the important thing is that there are no races.

ArvinDevel:
@sijie , I found @ivankelly 's FileInfoBackingCache design can address his comments, I'll adopt his elegant design.

ArvinDevel:
@ivankelly thanks for your design and I adopted it, please review it when you have time.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -330,54 +333,216 @@ private int readFromLogChannel(long entryLogId, BufferedReadChannel channel, Byt
...
+            if (fc.markDead()) {
+                try {
+                    fc.fc.close();
+                } catch (IOException e) {
+                    LOG.warn("Exception occurred in ReferenceCountedFileChannel"
+                            + " while closing channel for log file: {}", fc);
+                } finally {
+                    IOUtils.close(LOG, fc.fc);
+                }
+                fileChannels.remove(logId);
```
ivankelly: No guarantee that the CachedFileChannel being removed from fileChannels is actually the one referenced by fc.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -1094,24 +1259,25 @@ private Header getHeaderForLogId(long entryLogId) throws IOException {
...
+            return brc;
+        }
+        FileChannelBackingCache.CachedFileChannel cachedFileChannel = null;
+        try {
+            do {
+                cachedFileChannel = logid2FileChannel.loadFileChannel(entryLogId);
+            } while (!cachedFileChannel.tryRetain());
+        } finally {
+            if (null != cachedFileChannel) {
+                cachedFileChannel.release();
```
ivankelly: ?
You're releasing the reference straight away after getting it. By the time cachedFileChannel is used on L1278, the reference could easily be dead.

ArvinDevel: `loadFileChannel` and `tryRetain` both add one reference count, so releasing one is necessary.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -330,54 +333,216 @@ private int readFromLogChannel(long entryLogId, BufferedReadChannel channel, Byt
...
-    private final ConcurrentMap<Long, FileChannel> logid2FileChannel = new ConcurrentHashMap<Long, FileChannel>();
+    class FileChannelBackingCache {
+        static final int DEAD_REF = -0xdead;
+
+        final ConcurrentHashMap<Long, CachedFileChannel> fileChannels = new ConcurrentHashMap<>();
+
+        CachedFileChannel loadFileChannel(long logId) throws IOException {
+            CachedFileChannel cachedFileChannel = fileChannels.get(logId);
+            if (cachedFileChannel != null) {
+                boolean retained = cachedFileChannel.tryRetain();
```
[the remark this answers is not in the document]
ArvinDevel: um, that's right. The ReadWriteLock is necessary, I'll refactor again. I thought that lock is used to guarantee FileInfo's file open/write operations before now.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
ArvinDevel: I'll use lock to avoid multi write/delete operations to the fileChannels.

ivankelly: it shouldn't need locking. there's a remove method that takes the key and the value, and only removes if both match.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -330,54 +333,216 @@ private int readFromLogChannel(long entryLogId, BufferedReadChannel channel, Byt
...
             // Since this is thread local there only one modifier
             // We dont really need the concurrency, but we need to use
             // the weak values. Therefore using the concurrency level of 1
-            return new MapMaker().concurrencyLevel(1)
-                .weakValues()
-                .makeMap();
+            return CacheBuilder.newBuilder().concurrencyLevel(1)
+                    .expireAfterAccess(readChannelCacheExpireTimeMs, TimeUnit.MILLISECONDS)
+                    //decrease the refCnt
+                    .removalListener(removal -> logid2FileChannel.get((Long) removal.getKey()).release())
```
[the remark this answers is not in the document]
ArvinDevel: I think that adding a readLock when getting value from FileChannelBackingCache has the same effect, and that has little change to BufferedReadChannel. I prefer to this one, which one is better in your opinion?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
ArvinDevel: okay, then I wonder why you still use `writeLock` in `releaseFileInfo` method? see [code](https://github.com/apache/bookkeeper/blob/26c3379c9fdd812ef36da6176d83e26179705535/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileInfoBackingCache.java#L80-L90). Can we reduce lock in there? or the lock is a must when change the state of concurrentHashMap?

ivankelly: The write lock is needed because marking a fileinfo as dead, flushing its contents to disk and removing it from the map needs to be an atomic operation. If the lock wasn't there, then we could mark as dead, then before removing from map another thread could call loadFileInfo() and end up returning a dead fileinfo (tryRetain would fail). this would put the caller into a tight loop. We could, i think, remove the lock entirely from this class, but I'm very reluctant to mess with it, since this stuff is hard to get right (and it's often very hard to tell that you got it wrong).

asfgit:

ArvinDevel:
@ivankelly hi, could you review it again? thank you very much!

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileChannelBackingCache.java
```diff
@@ -0,0 +1,214 @@
...
+
+    final ConcurrentHashMap<Long, CachedFileChannel> fileChannels = new ConcurrentHashMap<>();
+
+    CachedFileChannel loadFileChannel(long logId) throws IOException {
+        lock.readLock().lock();
+        try {
+            CachedFileChannel cachedFileChannel = fileChannels.get(logId);
+            if (cachedFileChannel != null) {
+                boolean retained = cachedFileChannel.tryRetain();
+                assert(retained);
```
ivankelly: don't use assert. we're trying to get rid of them. Use a Precondition.check of some sort.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileChannelBackingCache.java
```diff
@@ -0,0 +1,214 @@
...
+        for (Map.Entry<Long, CachedFileChannel> entry : fileChannels.entrySet()) {
+            entry.getValue().fileChannel.close();
+        }
+    }
+
+    @VisibleForTesting
+    CachedFileChannel get(Long logId) {
+        lock.readLock().lock();
+        try {
+        return fileChannels.get(logId);
```
ivankelly: indent

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -993,56 +1006,58 @@ synchronized boolean readEntryLogHardLimit(long size) {
...
-        long entryLogId = logIdForOffset(location);
-        long pos = location & 0xffffffffL;
-        ByteBuf sizeBuff = sizeBuffer.get();
-        sizeBuff.clear();
-        pos -= 4; // we want to get the ledgerId and length to check
-        BufferedReadChannel fc;
+        EntryLogBufferedReadChannel bc = null;
         try {
-            fc = getChannelForLogId(entryLogId);
+            long entryLogId = logIdForOffset(location);
```
ivankelly: What is this change for? It looks pretty much the same as it was before, with the exception handling changed.

ivankelly: If you're going to change around the exception handling, you may as well make EntryLogBufferedReadChannel autocloseable, and use a try-with-resources block.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -703,7 +716,7 @@ void stop() {
      *          Entry Log File Id
      */
     protected boolean removeEntryLog(long entryLogId) {
-        removeFromChannelsAndClose(entryLogId);
+        fileChannelBackingCache.removeFromChannelsAndClose(entryLogId);
```
ivankelly: What about the entries still in the caches? I would suggest you get rid of ```removeFromChannelsAndClose``` completely, and invalidate the entry from all the caches. You could also do nothing, as the cache would release their refcount eventually, but better to help it along a little.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -1090,28 +1106,46 @@ private Header getHeaderForLogId(long entryLogId) throws IOException {
...
+            Callable<EntryLogBufferedReadChannel> loader = () -> {
+                CachedFileChannel cachedFileChannel = fileChannelBackingCache.loadFileChannel(entryLogId);
+                // We set the position of the write buffer of this buffered channel to Long.MAX_VALUE
+                // so that there are no overlaps with the write buffer while reading
+                return new EntryLogBufferedReadChannel(cachedFileChannel, conf.getReadBufferBytes());
+            };
+            do {
+                // Put the logId, bc pair in the cache responsible for the current thread.
+                brc = getThreadLocalCacheForReadChannel().get(entryLogId, loader);
+                if (!brc.cachedFileChannel.tryRetain()) {
```
ivankelly: better to do this in the loader:

```
() -> {
    while (true) {
        CachedFileChannel cachedFileChannel = fileChannelBackingCache.loadFileChannel(entryLogId);
        if (cachedFileChannel.tryRetain()) {
            return new EntryLogBufferedReadChannel(cachedFileChannel, conf.getReadBufferBytes());
        }
    }
}
```

You'll avoid having to add and remove from cache.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -1174,24 +1208,19 @@ private File findFile(long logId) throws FileNotFoundException {
      * @throws IOException
      */
     public void scanEntryLog(long entryLogId, EntryLogScanner scanner) throws IOException {
-        // Buffer where to read the entrySize (4 bytes) and the ledgerId (8 bytes)
-        ByteBuf headerBuffer = Unpooled.buffer(4 + 8);
-        BufferedReadChannel bc;
-        // Get the BufferedChannel for the current entry log file
+        EntryLogBufferedReadChannel bc = null;
```
ivankelly: There's a lot of code moved around here for what seems to be just exception. Couldn't you have just released the bc object in the final finally block.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -1361,25 +1401,20 @@ public void shutdown() {
         // since logChannel is buffered channel, do flush when shutting down
         LOG.info("Stopping EntryLogger");
         try {
+            //close corresponding fileChannels for read
+            fileChannelBackingCache.closeAllFileChannels();
```
ivankelly: Invalidating all the caches should have the same effect.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
```diff
@@ -391,4 +489,57 @@ public void testGetEntryLogsSet() throws Exception {
 
         assertEquals(Sets.newHashSet(0L, 1L, 2L, 3L), logger.getEntryLogsSet());
     }
+
+    /**
+     * Fake Ticker to advance ticker as expected.
+     */
+    class FakeTicker extends Ticker {
```
ivankelly: I'm surprised gauva itself doesn't supply this, since testing is its reason for being.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
```diff
@@ -322,6 +327,99 @@ public void testRecoverFromLedgersMapOnV0EntryLog() throws Exception {
...
+        int gcWaitTime = 1000;
+        int expireTime = 1000;
+        ServerConfiguration conf = TestBKConfiguration.newServerConfiguration();
+        conf.setGcWaitTime(gcWaitTime);
+        conf.setLedgerDirNames(new String[] {tmpDir.toString()});
+        //since last access, expire after 1s
+        conf.setReadChannelCacheExpireTimeMs(expireTime);
+        conf.setEntryLogFilePreAllocationEnabled(false);
+        // below one will cost logId 0
+        Bookie bookie = new Bookie(conf);
```
ivankelly: You don't need to create a full bookie.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
```diff
@@ -391,4 +489,57 @@ public void testGetEntryLogsSet() throws Exception {
...
+     */
+    class FakeEntryLogger extends EntryLogger {
+        private Ticker ticker;
+        FakeEntryLogger(ServerConfiguration conf, LedgerDirsManager ledgerDirsManager,
+                        Ticker ticker) throws IOException{
+            super(conf, ledgerDirsManager);
+            this.ticker = ticker;
+        }
+
+        private final ThreadLocal<Cache<Long, EntryLogBufferedReadChannel>> logid2ReadChannel =
```
ivankelly: Why are you overridding this? For the ticker? It would be better to always pass in a ticker, and have a #getTicker() method that you can override. This will cut down the amount of code you need to duplicate.

ivankelly: Obviously, the EntryLogger#getTicker() should just return Ticker.systemTicker

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
```diff
@@ -391,4 +489,57 @@ public void testGetEntryLogsSet() throws Exception {
...
+                            .removalListener(removal -> ((EntryLogBufferedReadChannel) removal.getValue()).release())
+                            .ticker(ticker).build();
+                    }
+                };
+
+        /**
+         * Override parend method to use mock logid2ReadChannel.
+         * @return
+         */
+        Cache<Long, EntryLogBufferedReadChannel> getThreadLocalCacheForReadChannel() {
```
ivankelly: If you do the ticker thing I said above, this isn't needed.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -330,54 +350,47 @@ private int readFromLogChannel(long entryLogId, BufferedReadChannel channel, Byt
...
-     * @param bc
+     * Each thread local buffered read channel can share the same file handle because reads are not relative
+     * and don't cause a change in the channel's position.
+     * Each file channel is mapped to a log id which represents an open log file.
      */
-    public BufferedReadChannel putInReadChannels(long logId, BufferedReadChannel bc) {
-        Map<Long, BufferedReadChannel> threadMap = logid2Channel.get();
-        return threadMap.put(logId, bc);
+    private FileChannelBackingCache fileChannelBackingCache = new FileChannelBackingCache(this::findFile);
+    @VisibleForTesting
```
ivankelly: Since you have a FakeEntryLogger in the test, why not just make fileChannelBackingCache package private, and use FakeEntryLogger from there?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -330,54 +350,47 @@ private int readFromLogChannel(long entryLogId, BufferedReadChannel channel, Byt
...
     };
 
-    /**
-     * Each thread local buffered read channel can share the same file handle because reads are not relative
-     * and don't cause a change in the channel's position. We use this map to store the file channels. Each
-     * file channel is mapped to a log id which represents an open log file.
-     */
-    private final ConcurrentMap<Long, FileChannel> logid2FileChannel = new ConcurrentHashMap<Long, FileChannel>();
+    @VisibleForTesting
+    long getReadChannelCacheExpireTimeMs() {
```
ivankelly: Should be able to get this from the configuraiton object in the test.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
ArvinDevel: We can't invalidate it from all the caches eg. logid2ReadChannel as it's a threadLocal variable. Do you think it's okay to delete  the entryLog file before close the corresponding fileChannel? I'm not sure if the JVM will throw exception if we do that. If it's fine then we can remove it at all.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
ArvinDevel: I just move previous exception handling code's location. I think keeping original's exception handling logic is necessary.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
ArvinDevel: If we want to invalidate the threadLocal cache, then we has 2 ways to implement which was discussed in here(<html>https://stackoverflow.com/questions/2795447/is-there-no-way-to-iterate-over-or-copy-all-the-values-of-a-java-threadlocal</html>).  I think invalidating method is complicated than simply closing in fileChannelBackingCache.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
ArvinDevel: There exists indeed. Now I import the guava test-lib dependency to solve this.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
ivankelly: I guess it would throw an exception when the file is eventually dropped from all caches and it tries to close it. This could be unpredictable, as removals happen when another action trigger it, like a put, and it's that put that would get the exception.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
ivankelly: Ah yes, sorry, you're right. This is the same as the defensive code in IndexPersistentMgr to avoid a bug that caused a tight loop.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
ivankelly: I'm questioning it, because it creates noise the in patch. Theres a bunch of lines added and deleted, which are likely exactly the same, but they need to be checked anyhow. If you need to refactor, it's better to do in a separate patch which doesn't change any logic.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
ivankelly: it doesn't add much complexity. You just need to create a concurrent list of some sort, and when initialValue is called, add the cache to the list before you return it. The threads in bookkeeper live for the lifetime of the process, so you don't need to worry about removing the caches from the list (otherwise a per thread cache would be pointless).

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -330,54 +350,47 @@ private int readFromLogChannel(long entryLogId, BufferedReadChannel channel, Byt
...
             // Since this is thread local there only one modifier
             // We dont really need the concurrency, but we need to use
             // the weak values. Therefore using the concurrency level of 1
-            return new MapMaker().concurrencyLevel(1)
-                .weakValues()
-                .makeMap();
+            return CacheBuilder.newBuilder().concurrencyLevel(1)
+                    .expireAfterAccess(readChannelCacheExpireTimeMs, TimeUnit.MILLISECONDS)
+                    //decrease the refCnt
+                    .removalListener(removal -> ((EntryLogBufferedReadChannel) removal.getValue()).release())
```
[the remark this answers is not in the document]
ArvinDevel: I found that the cast is must because the removal's type is not declared.

ArvinDevel:
@ivankelly I have fixed the nits, you can review it again when you have space time.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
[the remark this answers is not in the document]
sijie: @ArvinDevel 

a better way here is not to cast the object itself, but assigning the type for the lambda closure.

```
CacheBuilder.newBuilder()
            .removalListener((RemovalListener<String, EntryLogBufferedReadChannel>) notification -> notification.getValue.release)
            .build();
```

@ivankelly because it is not pure RemovalListener<K, V>. it is RemovalListener<K1 extends K, V1 extends V>, you need type to do the inference. I believe oracle jdk and open jdk have different ways on interpreting this, even different java versions as well. so a safer approach is to have the interface type for lambda function.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
ArvinDevel: After add a global readChannel concurrent list, we can let invalidating cache to trigger close operation. So is it still a need to keep the removeFromChannelsAndClose()? I guess it's not anymore.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileChannelBackingCache.java
```diff
@@ -0,0 +1,190 @@
...
+                checkArgument(retained);
+                return cachedFileChannel;
+            }
+        } finally {
+            lock.readLock().unlock();
+        }
+
+        lock.writeLock().lock();
+        try {
+            File file = fileLoader.load(logId);
```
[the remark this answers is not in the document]
sijie: If this is borrowed from FileInfoBackingCache, same question is applied there.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileChannelBackingCache.java
```diff
@@ -0,0 +1,190 @@
...
+            // get channel is used to open an existing entry log file
+            // it would be better to open using read mode
+            FileChannel newFc = new RandomAccessFile(file, "r").getChannel();
+            CachedFileChannel cachedFileChannel = new CachedFileChannel(logId, newFc);
+            fileChannels.put(logId, cachedFileChannel);
```
sijie: you need to check the return value. there can be race condition that a file channel is added after get.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileChannelBackingCache.java
```diff
@@ -0,0 +1,190 @@
...
+        }
+    }
+
+    /**
+     * close FileChannel and remove from cache when possible.
+     * @param logId
+     * @param cachedFileChannel
+     */
+    private void releaseFileChannel(long logId, CachedFileChannel cachedFileChannel) {
+        lock.writeLock().lock();
```
sijie: same here, I am not sure why we are closing a file channel under a write lock.

so the question is what does `lock` is actually locking.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileChannelBackingCache.java
```diff
@@ -0,0 +1,190 @@
...
+    CachedFileChannel get(Long logId) {
+        lock.readLock().lock();
+        try {
+            return fileChannels.get(logId);
+        } finally {
+            lock.readLock().unlock();
+        }
+    }
+
+    class CachedFileChannel {
```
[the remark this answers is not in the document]
ArvinDevel: Because it uses FileChannelBackingCache's non-static method `releaseFileChannel`, so it can't be static

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileChannelBackingCache.java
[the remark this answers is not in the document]
ArvinDevel: This guarantees the refCnt's correctness.
Imagine the scenario: thread A and thread B both want to load the same fileChannel without lock:
1.A and B both run to the line after created fileChannel respectively;
2.A execute: 
`fileChannels.put(logId, cachedFileChannel);
  boolean retained = cachedFileChannel.tryRetain();`;
3.A was switched and B execute these line again;
Then the FileChannelBacking cache will hold the last fileChannel with one refCnt, but the refCnt should be two actually.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileChannelBackingCache.java
ArvinDevel: Yes, just like mentioned above, the writeLock can guarantee it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -1094,25 +1104,40 @@ private Header getHeaderForLogId(long entryLogId) throws IOException {
...
+    /**
+     * Add one refCnt for BufferedReadChannel and return it, caller need to subtract one refCnt.
+     * @param entryLogId
+     * @return
+     * @throws IOException
+     */
+    private EntryLogBufferedReadChannel getChannelForLogId(long entryLogId) throws IOException {
+        try {
+            EntryLogBufferedReadChannel brc;
+            Callable<EntryLogBufferedReadChannel> loader = () -> {
```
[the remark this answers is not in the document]
ArvinDevel: A better improvement, I have fixed it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileChannelBackingCache.java
ArvinDevel: The lock can guarantee the fileChannel being loading not closed by another thread.
Below is a race condition.
Thread A holds one refCnt for the specific fileChannel and release it, so the refCnt is 0 and being closed.
Thread B load the fileChannel with ReadLock, so it can't continue if thread A call `releaseFileChannel ` firstly as the Thread A holds writeLock. This case is fine.
If thread B add refCnt before thread A get writeLock, then `releaseFileChannel `'s logic guarantee the fileChannel was not closed by check `markDead()` under writeLock. So it still works fine under this case.

ivankelly: Only the removal from the map needs to be under the lock, to prevent another thread from handing out file channel being closed. Once the channel has been removed from the map, it can be closed outside of the lock. This is different than with fileinfo, because fileinfo needs to flush it's contents out to disk, ensuring noone else reads them before they are fully flushed.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileChannelBackingCache.java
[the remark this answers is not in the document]
ivankelly: @sijie for fileinfobackingcache, we need to load under a lock to have mutal exclusion with another thread that is flushing. That isn't the case here, since there's no harm in having two channels to the same file open. 

So this can probably be changed to remove the locking completely. When you load a new entry, do a putIfAbsent, and if it fails, clean up what you just opened. Alternatively, computeIfAbsent could be used.

ivankelly:
> I think the original implementation is much simpler approach.

The original implementation had race conditions, where dead file channels would be handed out, which is why I suggested Arvin take a look at the refcounting stuff from the fileinfobackingcache, which dealt with a similar problem. The locking from the fileinfobackingcache is probably not needed though.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileChannelBackingCache.java
ArvinDevel: @ivankelly I think the `markDead()` needs to be under the lock to achieve preventing another thread from handing out file channel being closed. Close operation can be move out completely.

ivankelly: markDead doesn't need to be under a lock. The caller should call tryRetain() after receiving, which will either prevent markDead from having an effect, if it is called before it.

ArvinDevel: Then we need some do while check in loadFileChannel. 
I update again.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileChannelBackingCache.java
```diff
@@ -52,36 +53,33 @@
...
+            fileChannels.computeIfAbsent(logId, logFileId -> {
+                CachedFileChannel cfc = null;
+                try {
+                    File file = fileLoader.load(logFileId);
+                    // get channel is used to open an existing entry log file
+                    // it would be better to open using read mode
+                    FileChannel newFc = new RandomAccessFile(file, "r").getChannel();
+                    cfc = new CachedFileChannel(logFileId, newFc);
+                } catch (IOException ioe){
+                    throw new UncheckedIOException(ioe);
```
ivankelly: This will end up doing throwing a RuntimeException. Rather than using a call to computeIfAbsent(), do:
```
CachedFileChannel cachedFileChannel = null;
do {
    CachedFileChannel c = fileChannels.get(logId);
    if (c != null) {
        if (c.tryRetain()) {
            cachedFileChannel = c;
        } else {
            // this isn't strictly necessary, but it's good defensively to avoid infinite loop
            fileChannels.remove(logId, c);
        }
    } else {
        c = // the construction stuff
        CachedFileChannel existing = fileChannels.putIfAbsent(logId, c);
        if (existing != null) {
            // cleanup c
            if (existing.tryRetain()) {
                cachedFileChannel = existing;
            } else {
                fileChannels.remove(logId, existing);
            }
        }
    }
} while (cachedFileChannel == null);
```

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileChannelBackingCache.java
```diff
@@ -52,36 +53,33 @@
     final ConcurrentHashMap<Long, CachedFileChannel> fileChannels = new ConcurrentHashMap<>();
 
     CachedFileChannel loadFileChannel(long logId) throws IOException {
```
ivankelly: lock field should be no longer used, so it can be removed from the class.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileChannelBackingCache.java
```diff
@@ -90,22 +88,18 @@ CachedFileChannel loadFileChannel(long logId) throws IOException {
...
-                    IOUtils.close(LOG, cachedFileChannel.fileChannel);
-                }
-                // to guarantee the removed cachedFileChannel is what we want to remove.
-                fileChannels.remove(logId, cachedFileChannel);
+        if (cachedFileChannel.markDead()) {
+            // to guarantee the removed cachedFileChannel is what we want to remove.
+            fileChannels.remove(logId, cachedFileChannel);
+            // close corresponding fileChannel
+            try {
+                cachedFileChannel.fileChannel.close();
```
ivankelly: it would be better is cachedFileChannel had a close method rather than directly accessing members.

jvrao:
retest

eolivelli:
closing for inactivity. feel free to reopen

PR #834, opened 2017-12-13
Issue 741: Switch to Spotbugs

mino181295:
Moved from findbugs plugin to spotbugs.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -115,7 +115,7 @@
     <zookeeper.version>3.5.3-beta</zookeeper.version>
     <findbugs-jsr305.version>3.0.0</findbugs-jsr305.version>
     <!-- plugin dependencies -->
-    <findbugs-maven-plugin.version>3.0.5</findbugs-maven-plugin.version>
+    <spotbugs-maven-plugin.version>3.1.0-RC6</spotbugs-maven-plugin.version>
     <puppycrawl.checkstyle.version>6.19</puppycrawl.checkstyle.version>
     <maven-checkstyle-plugin.version>2.17</maven-checkstyle-plugin.version>
     <maven-compiler-plugin.version>3.7.0</maven-compiler-plugin.version>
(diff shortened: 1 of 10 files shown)


pom.xml
```diff
@@ -197,9 +197,9 @@
...
     <plugins>
       <plugin>
-        <groupId>org.codehaus.mojo</groupId>
-        <artifactId>findbugs-maven-plugin</artifactId>
-        <version>${findbugs-maven-plugin.version}</version>
+        <groupId>com.github.spotbugs</groupId>
+        <artifactId>spotbugs-maven-plugin</artifactId>
+        <version>${spotbugs-maven-plugin.version}</version>
         <configuration>
           <excludeFilterFile>${session.executionRootDirectory}/buildtools/src/main/resources/bookkeeper/findbugsExclude.xml</excludeFilterFile>
```
ivankelly: Should probably rename this file also

eolivelli: @ivankelly there is already an issue, I have already sent emails in the past months about this change.
I think we change keep findbugsExclude.xml as effectively SpotBugs is Findbugs (it just changed owner) and internally the format of the file is the same. The SportBugs maven plugin is a port of the Findbugs Maven plugin and all the configs are the same.

But if you feel strong about this we can change it as well.

sijie:
committers: when you merge this pull request, please update all the jenkins job with the new command

eolivelli:
@sijie sorry, I missed your comment before sending my approval

eolivelli:
@sijie I will take care of updating the jobs on Jenkins

pom.xml
ivankelly: @eolivelli it still needs to be announced when done, so people know they need to run spotbugs:check instead of findbugs:check

ivankelly: filename change isn't important, i'd just update it if other things need an update

eolivelli:
@sijie I think @mino181295 addressed all of your comments, I will wait you acknowledge

sijie:
@eolivelli I merged this PR and updated CI jobs.

PR #835, opened 2017-12-13
Allow subclasses of AbstractConfiguration to use fluent style

ivankelly:
And in doing so, get rid of ugly casts.


sijie:
retest this please

PR #836, opened 2017-12-13
Fix Flaky CreateNewLogTest

sijie:

CreateNewLogTest becomes flaky after fixing #568 because preallocation can happen in the background. Changing the sequence of creating entry logger and entry log file injection to make the test sequence deterministic.


sijie:
retest this please

PR #838, opened 2017-12-13
Issue-326:  Replace observer/observable with a simplified watcher/watchable implementation

sijie:

- long poll only need one-time notification on lac updates
- replace observer/observable with a simplified watcher/watchable implementation. watchers are removed after they are fired.
- add `RecycableHashSet` for `Watchable` keeping the list of watchers
- object pooling on `Watchable` and `LastAddConfirmedUpdateNotification`

### DIFF
--- a/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/collections/RecyclableHashSet.java
+++ b/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/collections/RecyclableHashSet.java
@@ -1,0 +1,60 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.bookkeeper.common.collections;
+
+import io.netty.util.Recycler.Handle;
+import java.util.HashSet;
+import org.apache.bookkeeper.common.util.Recyclable;
+
+/**
+ * A hash set which is recyclable.
+ */
+public final class RecyclableHashSet<T> extends HashSet<T> implements Recyclable {
+
+    /**
+     * An HashSet Recycler.
+     */
+    public static class Recycler<X>
+        extends io.netty.util.Recycler<RecyclableHashSet<X>> {
+
+        @Override
+        protected RecyclableHashSet<X> newObject(Handle<RecyclableHashSet<X>> handle) {
+            return new RecyclableHashSet<>(handle);
+        }
+
+        public RecyclableHashSet<X> newInstance() {
+            return get();
+        }
+    }
+
+    private final Handle<RecyclableHashSet<T>> handle;
+
+    private RecyclableHashSet(Handle<RecyclableHashSet<T>> handle) {
+        super();
+        this.handle = handle;
+    }
+
+    @Override
+    public void recycle() {
+        clear();
+        handle.recycle(this);
+    }
+
+}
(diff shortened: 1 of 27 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/InterleavedLedgerStorage.java
```diff
@@ -253,9 +249,11 @@ public long getLastAddConfirmed(long ledgerId) throws IOException {
     }
 
     @Override
-    public Observable waitForLastAddConfirmedUpdate(long ledgerId, long previoisLAC, Observer observer)
+    public boolean waitForLastAddConfirmedUpdate(long ledgerId,
+                                                 long previoisLAC,
+                                                 Watcher<LastAddConfirmedUpdateNotification> watcher)
             throws IOException {
-        return ledgerCache.waitForLastAddConfirmedUpdate(ledgerId, previoisLAC, observer);
+        return ledgerCache.waitForLastAddConfirmedUpdate(ledgerId, previoisLAC, watcher);
```
ivankelly: may as well fix typo in previousLAC while we're at it

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/Watchable.java
```diff
@@ -0,0 +1,120 @@
...
+ *
+ * <p>Note that this notification mechanism has nothing to do with threads
+ * and is completely separate from the <tt>wait</tt> and <tt>notify</tt>
+ * mechanism of class <tt>Object</tt>.
+ *
+ * <p>When an watchable object is newly created, its set of watchers is
+ * empty. Two watchers are considered the same if and only if the
+ * <tt>equals</tt> method returns true for them.
+ */
+public class Watchable<T> implements Recyclable {
```
ivankelly: Is the recyclable still necessary? It's used in FileInfo which isn't itself recyclable, and we only have one at a per ledger level. GC optimizations like this are only really necessary at the ledge entry level.

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/collections/RecyclableHashSet.java
```diff
@@ -0,0 +1,60 @@
...
+package org.apache.bookkeeper.common.collections;
+
+import io.netty.util.Recycler.Handle;
+import java.util.HashSet;
+import org.apache.bookkeeper.common.util.Recyclable;
+
+/**
+ * A hash set which is recyclable.
+ */
+public final class RecyclableHashSet<T> extends HashSet<T> implements Recyclable {
```
ivankelly: Making a HashSet recyclable wins you very little. A HashSet is really a HashMap<T, Boolean> underneath, so any lookups and puts will be generating non-recyled Cell objects. 

If we're going down this path, it would be better to use a ArrayList rather than a HashSet for Watcher, as it stores the objects directly. Insertion is O(n) if you want to ensure uniqueness, but the number of watchers is likely to be small enough that this would outperform a HashSet anyhow due to cache locality.

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/Watchable.java
sijie: yes. it is needed when it holds a reference to a recyclable object. FileInfo needs to recycle this when it is closed.

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/collections/RecyclableHashSet.java
sijie: it still saves you one allocation per lac update and long poll request. Based on current implementation, we can potentially use RecyclableArrayList, I will look into this.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/InterleavedLedgerStorage.java
sijie: will do

sijie:
@eolivelli it is on the read path. especially when the traffic of a single ledger is huge, all the lac notifications generate quite a lot of allocations.

eolivelli:
@sijie ok for me, it is clear. Thanks.

sijie:
retest this please

sijie:
@ivankelly addressed your comments

eolivelli:
There is a findbugs issue. I think recyclable object cannot be serializable. We do not use java serialization in BK

sijie:
@ivankelly I've addressed your comments. please review it.

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/Watchable.java
ivankelly: ah, got you now. The Watcher objects themselves are not recycled, but watchers is recreated each time a watch triggers.

PR #843, opened 2017-12-13
DbLedgerStorage -- ReadCache

merlimat:
Adds a `ReadCache` class to be used from LedgerStorage. 

The read cache is used when doing read-ahead caching after reading one entry. 

The idea is that entries for the same ledger as stored in order (for each flush cycle, eg: 1min). When reading 1 entry, we expect the client to read all subsequent entries. We can then skip the RocksDb `get()` to fetch the index of the entries, by just keep reading from the entryLog, and exploiting page cache and seqeuential reads. The entries are then added to this ReadCache, where they will looked when the request for next entry comes in. 

This implementation is based on dividing the memory into multiple segments. Whenever the read cache is full, the oldest segment gets recycled and overwritten with new entries.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/ReadCache.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/ReadCache.java
@@ -53,7 +53,7 @@ public class ReadCache implements Closeable {
     private final List<ConcurrentLongLongPairHashMap> cacheIndexes;
 
     private int currentSegmentIdx;
-    private AtomicInteger currentSegmentOffset = new AtomicInteger(0);
+    private final AtomicInteger currentSegmentOffset = new AtomicInteger(0);
 
     private final int segmentSize;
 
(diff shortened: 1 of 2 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/ReadCache.java
```diff
@@ -0,0 +1,197 @@
...
+ */
+public class ReadCache implements Closeable {
+
+    private static final int DEFAULT_MAX_SEGMENT_SIZE = 1 * 1024 * 1024 * 1024;
+
+    private final List<ByteBuf> cacheSegments;
+    private final List<ConcurrentLongLongPairHashMap> cacheIndexes;
+
+    private int currentSegmentIdx;
+    private AtomicInteger currentSegmentOffset = new AtomicInteger(0);
```
sijie: nit: final

merlimat:
> one question: I think this ReadCache is general can be shared with different ledger storage implementations, right? 

Yes, same for `WriteCache` or even for `KeyValueStorage` if that's needed in any other component.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/ReadCache.java
merlimat: yes, I thought it was being reassigned but it's not.

sijie:
@merlimat we can consider sharing the components with other implementations once the db ledger storage change is merged.

sijie:
merging this now to unblock subsequent db ledger storage changes.

Issue #668, opened 2017-10-27
Use try-with-resources blocks whenever is possible


eolivelli:
**FEATURE REQUEST**

1. Please describe the feature you are requesting.
We have several Closeable/AutoCloseable resources both on client-side and on server-side.
We may use java7 try-with-resources blocks in order to have error prone code and more elegant

2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have). Are you currently using any workarounds to address this issue?

nice-to-have

PR #845, opened 2017-12-14
ISSUE #668: Use try-with-resources blocks whenever is possible

pasha-kuznetsov:
Use java7 try-with-resources blocks in order to have error prone code and more elegant


sijie:
@jiazhai can you update `contribution guide`?

jiazhai:
@sijie  sure will do it.

Issue #846, opened 2017-12-14
Ledger Balancer


sijie:
**BP**

This is the master ticket for tracking BP-23:

In **Messaging/Streaming/Logging** oriented use case (where old ledgers/segments are most likely will be deleted at some point), we don't actually need to rebalance the ledgers stored on bookies. 

However,
In **Storage** oriented use cases (where data most likely will never be deleted), BookKeeper data might not always be placed uniformly across bookies. One common reason is addition of new bookies to an existing cluster. This proposal is proposing to provide a balancer mechanism (as an utility, also as part of `AutoRecovery` daemon), that analyzes ledger distributions and balances ledgers across bookies.

It replicated ledgers to new bookies (based on resource-aware placement policies) until the cluster is deemed to be balanced, which means that disk utilization of every bookie (ratio of used space on the node to the capacity of the node) differs from the utilization of the cluster (ratio of used space on the cluster to total capacity of the cluster) by no more than a given threshold percentage.

The balancer will replicate ledgers away from disk-filled bookies as first priority.

Proposal PR - [TBD]

sijie:
bump to 4.8.0

truong-hua:
Is it available in 4.16.0 yet?

PR #847, opened 2017-12-14
BP-23: Ledger Balancer (WIP)

sijie:

In Messaging/Streaming/Logging oriented use case (where old ledgers/segments are most likely will be deleted at some point), we don't actually need to rebalance the ledgers stored on bookies.

However,
In Storage oriented use cases (where data most likely will never be deleted), BookKeeper data might not always be placed uniformly across bookies. One common reason is addition of new bookies to an existing cluster. This proposal is proposing to provide a balancer mechanism (as an utility, also as part of AutoRecovery daemon), that analyzes ledger distributions and balances ledgers across bookies.

It replicated ledgers to new bookies (based on resource-aware placement policies) until the cluster is deemed to be balanced, which means that disk utilization of every bookie (ratio of used space on the node to the capacity of the node) differs from the utilization of the cluster (ratio of used space on the cluster to total capacity of the cluster) by no more than a given threshold percentage.

The balancer will replicate ledgers away from disk-filled bookies as first priority.

Master Issue: #846

### DIFF
--- a/site/bps/BP-23-ledger-rebalancer.md
+++ b/site/bps/BP-23-ledger-rebalancer.md
@@ -1,0 +1,50 @@
+---
+title: "BP-23: ledger balancer"
+issue: https://github.com/apache/bookkeeper/846
+state: "WIP" 
+release: "x.y.z"
+---
+
+### Motivation
+
+There are typical two use cases of _Apache BookKeeper_, one is *Messaging/Streaming/Logging* style use cases, the other one is *Storage* style use cases.
+
+In Messaging/Streaming/Logging oriented use case (where old ledgers/segments are most likely will be deleted at some point), we don't actually need to rebalance the ledgers stored on bookies.
+
+However,
+In Storage oriented use cases (where data most likely will never be deleted), BookKeeper data might not always be placed uniformly across bookies. One common reason is addition of new bookies to an existing cluster. This proposal is proposing to provide a balancer mechanism (as an utility, also as part of AutoRecovery daemon), that analyzes ledger distributions and balances ledgers across bookies.
+
+It replicated ledgers to new bookies (based on resource-aware placement policies) until the cluster is deemed to be balanced, which means that disk utilization of every bookie (ratio of used space on the node to the capacity of the node) differs from the utilization of the cluster (ratio of used space on the cluster to total capacity of the cluster) by no more than a given threshold percentage.
+
+The balancer will replicate ledgers away from disk-filled bookies as first priority.
+
+### Public Interfaces
+
+There is not public API changes.
+
+Potentially we might need a new command in `BookieShell` to run balancer.
+
+### Proposed Changes
+
+[TBD]
+
+A couple of thoughts:
+
+- it should be moving ledgers from filled-up bookies only.
+- it should enforce resource-awareness into the ledger placement policy.
+- it should provide capabilities to throttle bandwidth usage.
+
+
+### Compatibility, Deprecation, and Migration Plan
+
+N/A
+
+### Test Plan
+
+[TBD]
+
+### Rejected Alternatives
+
+Manual balancer using `Recovery` tools.
+
+[TBD]
--- a/site/community/bookkeeper_proposals.md
+++ b/site/community/bookkeeper_proposals.md
@@ -85,7 +85,7 @@ using Google Doc.
 
 This section lists all the _bookkeeper proposals_ made to BookKeeper.
 
-*Next Proposal Number: 23*
+*Next Proposal Number: 24*
 
 ### Inprogress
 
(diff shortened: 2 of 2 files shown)


sijie:
This is WIP, putting a placeholder here.

site/bps/BP-23-ledger-rebalancer.md
```diff
@@ -0,0 +1,50 @@
...
+### Motivation
+
+There are typical two use cases of _Apache BookKeeper_, one is *Messaging/Streaming/Logging* style use cases, the other one is *Storage* style use cases.
+
+In Messaging/Streaming/Logging oriented use case (where old ledgers/segments are most likely will be deleted at some point), we don't actually need to rebalance the ledgers stored on bookies.
+
+However,
+In Storage oriented use cases (where data most likely will never be deleted), BookKeeper data might not always be placed uniformly across bookies. One common reason is addition of new bookies to an existing cluster. This proposal is proposing to provide a balancer mechanism (as an utility, also as part of AutoRecovery daemon), that analyzes ledger distributions and balances ledgers across bookies.
+
+It replicated ledgers to new bookies (based on resource-aware placement policies) until the cluster is deemed to be balanced, which means that disk utilization of every bookie (ratio of used space on the node to the capacity of the node) differs from the utilization of the cluster (ratio of used space on the cluster to total capacity of the cluster) by no more than a given threshold percentage.
```
ivankelly: Currently in rereplication, we only replicate a fragment. I.e. if striping is used only one stripe is rereplicated. This doesn't make a huge amount of sense. We stripe to allow an increased throughput. At the rereplication stage, throughput isn't a concern. 

Perhaps for rereplication and for this balancing, instead of copying by fragment, we should just copy the whole ledger, 0->lac.

Also, maybe we shouldn't modify the ensemble when we do this. The ensemble is overloaded right now, as a list of who can vote on an entry, and a pointer to location of an entry. Once the entry has been acknowledged to a user, who voted on that entry can't be changed, yet we do change it. Perhaps we should have a shadow ensemble, which is updated only when a fragment is rereplicated or rebalanced. This would greatly simply ensemble change for the writer, as the ensemble would never change unless the ledger is being recovered, and in that case the ledger would be fenced anyhow, so the write could give up. Only recovery would have to deal with changing ensembles.

sijie: > Only recovery would have to deal with changing ensembles.

I am not sure I understand your comment here. But in current implementation, only write failures on normal writes and ledger recovery deal with ensemble changes. rereplication doesn't change ensemble. The ensemble is only updated when a fragment is fully replicated. 

> Perhaps for rereplication and for this balancing, instead of copying by fragment, we should just copy the whole ledger, 0->lac.

I am not sure it worths changing the copying mechanism here. because rereplication and balancing would only deal with sealed fragments. I don't think there are issues with current copying mechanism, unless I missed anything.

This BP here should be more focusing on the strategy on balancing data using whatever copying mechanism available in the system. Changing the copying mechanism should be a different task. Otherwise we are mixing two big changes here, which would make things a bit complicated to discuss, review and implement.

ivankelly: > I am not sure I understand your comment here. But in current implementation, only write failures on normal writes and ledger recovery deal with ensemble changes. rereplication doesn't change ensemble. The ensemble is only updated when a fragment is fully replicated.

With the current implementation, if a write needs to change the ensemble, and a previous fragment has been rereplicated, the change ensemble update will fail, the client needs to read back to check if the ledger is fenced and merge the ensembles. This merging is messy, which is why I'm suggesting we get rid of it.

With a shadow ensemble for replication, the writer changing ensemble can assume that the ensemble it is trying to write is correct, so if the write fails, it only checks if there is fencing, and if not retries the write with the assumed to be correct ensemble. Similarly for multiple recovers, if they have to change the ensemble [1], each one will be able to ensure that all necessary entries are on all nodes in their presumably correct ensemble, so if they fail to write, they can just try again (while checking the lastEntry in the metadata). Again, merging is unnecessary.

[1] Even then, we could get rid of ensemble change in recovery

> I am not sure it worths changing the copying mechanism here. because rereplication and balancing would only deal with sealed fragments. I don't think there are issues with current copying mechanism, unless I missed anything.

Ya, not worth changing now. If we were to start again though, I think I'd do it this way. That said, I'd probably get rid of ensemble changes too, and push that up to the higher level client

sijie: > Ya, not worth changing now. If we were to start again though, I think I'd do it this way. That said, I'd probably get rid of ensemble changes too, and push that up to the higher level client

Agreed. If you think from managed ledger / distributedlog, you can simply close a segment and open a new segment.

site/bps/BP-23-ledger-rebalancer.md
```diff
@@ -0,0 +1,50 @@
...
+
+Potentially we might need a new command in `BookieShell` to run balancer.
+
+### Proposed Changes
+
+[TBD]
+
+A couple of thoughts:
+
+- it should be moving ledgers from filled-up bookies only.
```
merlimat: Rather than filled-up, I would say that bookies that have passed a certain threshold, eg: 50%, because we should avoid to fill up bookies in order to avoid losing IO capacity.

site/bps/BP-23-ledger-rebalancer.md
jvrao: @sijie I would request this to be a separate doc, when we come out with a design. 
If we are *not* changing the code, and this proposal is to create a stand-alone tool, then we can surely
work with md file. On the other hand if we need to change replication logic, it makes great sense to have
a separate design doc.

ivankelly: @jvrao nah, we're not going to change the logic. Sorry for muddying the waters.

asfgit:

PR #851, opened 2017-12-14
Issue 578 : make MajorCompaction controlled by time of the day/day of the week

ArvinDevel:

Add time of day based GC policy to MajorCompaction, when the time is in specified zone(here is when system is in low load status), it will adjust the gc threshold to specified value.
In my opinion, after adding this policy to MajorCompaction, there is no need to add similar policy to MinorCompaction, because the MajorCompaction has done the deeper gc.
Next, I will add a test case and explore the system load based policy.

> ---
> quickly and easily:
>
>
>
> Otherwise:
> 
> 
> ---


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -183,6 +233,8 @@ public GarbageCollectorThread(ServerConfiguration conf,
                + majorCompactionThreshold + ", interval=" + majorCompactionInterval);
 
         lastMinorCompactionTime = lastMajorCompactionTime = MathUtils.now();
+        calendar = Calendar.getInstance();
```
eolivelli: You have to create a new calendar at every execution.

You have to explicitly configure a timezone for this calendar, otherwise the configuration is not deterministic

Can you consider using java time API and not legacy Calendar?

eolivelli:
+1 for the cron format

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
ArvinDevel: I think that using the default system timezone is ok, the user configure the specified time  period using their default system timezone usually.
What does you mean that explicitly configure a timezone?  Adding a new field to timezone in config file?

ArvinDevel:
@sijie good idea. 
So we need to add a new config file which contains the cron plan or add the cron plan to the bk_server.conf, I think adding a new line in  bk_server.conf is better.

ArvinDevel:
@eolivelli @sijie Now I keep the two alternative to parse cron string, one is just a class to which I put in common.util, another is org.quartz.CronExpression which need add a maven dependency, do you think which one is better?

eolivelli:
Please do not add a dependency if not really needed

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/CronExpression.java
```diff
@@ -0,0 +1,493 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Locale;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+
+/**
+ * This provides cron support for java8 using java-time.
```
sijie: we might need to add a copyright notice to NOTICE file, if this is modified based on https://github.com/frode-carlsen/cron

bookkeeper-common/src/test/java/org/apache/bookkeeper/common/util/CronExpressionTest.java
```diff
@@ -0,0 +1,848 @@
...
+import org.apache.bookkeeper.common.util.CronExpression.CronFieldType;
+import org.apache.bookkeeper.common.util.CronExpression.DayOfMonthField;
+import org.apache.bookkeeper.common.util.CronExpression.DayOfWeekField;
+import org.apache.bookkeeper.common.util.CronExpression.SimpleField;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+/**
+ * A test for CronExpression.
```
sijie: Add a comment this file is based on https://github.com/frode-carlsen/cron/blob/master/java8/src/test/java/fc/cron/CronExpressionTest.java

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
```diff
@@ -1101,6 +1105,101 @@ public ServerConfiguration setMajorCompactionThreshold(double threshold) {
...
+        setProperty(MAJOR_HIGH_COMPACTION_THRESHOLD, threshold);
+        return this;
+    }
+
+    /**
+     * Get median major compaction cron expression.
+     * default is 0-3 h every day.
+     */
+    public String getMedianMajorCompactionCron() {
+        return getString(MEDIAN_MAJOR_COMPACTION_CRON, "0 * 0-3 ? * *");
```
sijie: I think default value should be returning null. if the return value is null, the cron based compaction is disabled.

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
```diff
@@ -1101,6 +1105,101 @@ public ServerConfiguration setMajorCompactionThreshold(double threshold) {
...
+        setProperty(MEDIAN_MAJOR_COMPACTION_CRON, medianMajorCompactionCron);
+        return this;
+    }
+
+    /**
+     * Get high major compaction cron expression.
+     * default is 0-3 h in Saturday.
+     */
+    public String getHighMajorCompactionCron() {
+        return getString(HIGH_MAJOR_COMPACTION_CRON, "0 * 0-3 ? * 7");
```
sijie: return null as default. returning null means this feature is disabled.

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
```diff
@@ -42,6 +42,10 @@
     protected static final String MINOR_COMPACTION_THRESHOLD = "minorCompactionThreshold";
     protected static final String MAJOR_COMPACTION_INTERVAL = "majorCompactionInterval";
     protected static final String MAJOR_COMPACTION_THRESHOLD = "majorCompactionThreshold";
+    protected static final String MAJOR_MEDIAN_COMPACTION_THRESHOLD = "majorMedianCompactionThreshold";
```
sijie: add these new settings to https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/conf/bk_server.conf and https://github.com/apache/bookkeeper/blob/master/site/_data/config/bk_server.yaml

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -134,15 +142,20 @@ public GarbageCollectorThread(ServerConfiguration conf,
         // compaction parameters
         minorCompactionThreshold = conf.getMinorCompactionThreshold();
         minorCompactionInterval = conf.getMinorCompactionInterval() * SECOND;
-        majorCompactionThreshold = conf.getMajorCompactionThreshold();
+        originalMajorCompactionThreshold = majorCompactionThreshold = conf.getMajorCompactionThreshold();
         majorCompactionInterval = conf.getMajorCompactionInterval() * SECOND;
+        medianMajorCompactionThreshold = conf.getMedianMajorCompactionThreshold();
```
sijie: do we need to set any constraints between `MajorCompactionInterval` `MedianMajorCompactionInterval` and `HighMajorCompactionInterval`? E.g `MajorCompactionInterval` < `MedianMajorCompactionInterval` < `HighMajorCompactionInterval`?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -264,12 +278,53 @@ public void safeRun() {
...
+        majorCompactionThreshold = originalMajorCompactionThreshold;
+
     }
 
     public void runWithFlags(boolean force, boolean suspendMajor, boolean suspendMinor) {
         if (force) {
             LOG.info("Garbage collector thread forced to perform GC before expiry of wait time.");
         }
+
+        changeMajorCompactionThreshold();
```
sijie: Based on my understanding to this logic, the crons are used for adjusting major compaction thresholds. that means major compaction interval should be smaller than the interval that cron runs, otherwise the logic won't work here, right?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
ArvinDevel: that's right

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
ArvinDevel: There doesn't exists this constraint, as long as the next cron fire time fit in the next gc period, it will change the threshold. And the next cron fire time is calculated on `now()`, the `lastChangeThresholdTime` is equal to the lastGCTime.

dlg99:
@sijie I'll take a look at the change.

I did functionally similar change but it ended up being unused due to less than predictable schedule in our pre-prod env and work in progress on having multiple active ledgers to eliminate compaction completely. 
We are pretty much planning to drop time-based configuration but if you feel that this is helpful to community I can merge/do a pull req for it to the community code.

My approach was slightly different:

abstract config allowed one to specify time ranges and any parameter could be enabled to make use of it:
```text
    ## Some settings support configuration override for specific time range.
   ## For example one can set:
   ## gcWaitTime=1800000
   ## gcWaitTime@timerange0=900000
   ## 
   ## Time ranges defined as:
   ## <rangename>.start=<start time>
   ## <rangename>.end=<end time>
   ## Optional:
   ## <rangename>.day=<comma-separated list of days of the week>
   ## <rangename> consists of prefix "timerange" plus numeric id.
   ## Numeric id starts from 0 and incremented by 1. 
   ## Search for ranges stops on the first missing id and/or on the first range that misses 
   ## or has wrong value for the time or day.
   ## 
   ## Start and end time defined in ISO_LOCAL_TIME format (24 hr, local time zone)
   ## 
   ## Time ranges can overlap because different ranges can be applied to different groups of
   ## parameters.
   ## 
   ## Examples:
   ## 
   ## timerange0.start=09:00:00
   ## timerange0.end=17:59:59.999
   ## 
   ## timerange1.start=22:00
   ## timerange1.end=23:00
   ## timerange1.day=Saturday,Sunday
   ## 
   ## minorCompactionThreshold=0.2
   ## minorCompactionThreshold@timerange0=0.1
   ## 
   ## minorCompactionInterval=3600
   ## minorCompactionInterval@timerange0=1800
   ## 
   ## majorCompactionThreshold=0.3
   ## majorCompactionThreshold@timerange0=0.2
   ## majorCompactionThreshold@timerange1=0.5
   ## 
   ## majorCompactionInterval=86400 
   ## majorCompactionInterval@timerange1=1800 
```

```java
    public double getMinorCompactionThreshold(LocalTime now, DayOfWeek day) {
        return getDouble(this.getPropertyNameForFirstActiveTimerange(MINOR_COMPACTION_THRESHOLD, now, day), 0.2f);
    }
```

and in compaction thread I had to make sure that parameters are refreshed before each use like:

```java
    refreshGcConfigParams(conf.getLocalTime(), conf.getDayOfWeek());
    ....
    private void refreshGcConfigParams(LocalTime now, DayOfWeek day) {
        gcWaitTime = conf.getGcWaitTime(now, day);
        minorCompactionThreshold = conf.getMinorCompactionThreshold(now, day);
        minorCompactionInterval = conf.getMinorCompactionInterval(now, day) * SECOND;
        majorCompactionThreshold = conf.getMajorCompactionThreshold(now, day);
        majorCompactionInterval = conf.getMajorCompactionInterval(now, day) * SECOND;
       .....
```

sijie:
@dlg99 I see. I think the main reason you dropped the time-based compaction is because of multiple entry loggers feature that Charan is working on. But not every use cases will move to use the approach. so compaction can not be avoided for some of the use cases. it has the value to have time-based compaction, so applications can have a better control on when compaction happens.

If you have time to review this, we can try to consolidate the efforts here with your efforts, to find a good approach for it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
```diff
@@ -1084,7 +1088,7 @@ public ServerConfiguration setMinorCompactionThreshold(double threshold) {
      * @return threshold of major compaction
      */
     public double getMajorCompactionThreshold() {
-        return getDouble(MAJOR_COMPACTION_THRESHOLD, 0.8f);
+        return getDouble(MAJOR_COMPACTION_THRESHOLD, 0.6f);
```
dlg99: why did default change?

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
```diff
@@ -1101,6 +1105,101 @@ public ServerConfiguration setMajorCompactionThreshold(double threshold) {
...
+    /**
+     * Get threshold of high major compaction.
+     *
+     * <p>same with the @see #getMajorCompactionThreshold()
+     * the high threshold is used when the system is on a very low load status
+     *
+     * @return threshold of high major compaction
+     */
+    public double getHighMajorCompactionThreshold() {
+        return getDouble(MAJOR_HIGH_COMPACTION_THRESHOLD, 0.9f);
```
reddycharan: can you add sample values to bk_server.conf

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
```diff
@@ -1101,6 +1105,101 @@ public ServerConfiguration setMajorCompactionThreshold(double threshold) {
         return this;
     }
 
+    /**
+     * Get threshold of median major compaction.
+     *
+     * <p>same with the @see #getMajorCompactionThreshold()
+     * the high threshold is used when the system is on a low load status
```
dlg99: The change has nothing to do with adjusting compaction parameters based on actual system/bookie load as one can think after reading this, it adjusts them on schedule (which supposed to reflect one's best guess for the system load). I'd adjust the doc text to reflect this.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -68,9 +69,16 @@
     long lastMinorCompactionTime;
 
     boolean enableMajorCompaction = false;
-    final double majorCompactionThreshold;
+    final double originalMajorCompactionThreshold;
+    double majorCompactionThreshold;
```
athanatos: I don't think you want to maintain both majorCompactionThreshold and originalMajorCompactionThreshold.  majorCompactionThreshold is mainly used in two places:
1. In the constructor, you do some validation of the config, but you don't need to change it to do that.
2. In runWithFlags, you set it based on the current time, pass it into doCompactEntryLogs, and restore it back to original before returning.

Instead, rename originalMajorCompactionThreshold to majorCompactionThreshold (and leave it final) and get rid of the mutable majorCompactionThreshold.  In runWithFlags, modify changeMajorCompactionThreshold to getCurrentMajorCompactionThreshold and let it return a value based on the current time which you can keep in a local in runWithFlags.  You don't need restoreThreshold at all.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -264,12 +278,53 @@ public void safeRun() {
...
+     * increase the threshold when the system in a very low load status.
+     */
+    private void activateHighThreshold(){
+        majorCompactionThreshold = highMajorCompactionThreshold;
+    }
+
+    /**
+     * Check whether the configured cron is met and activate threshold.
+     */
+    private void changeMajorCompactionThreshold(){
```
athanatos: This strikes me as quite brittle.  It seems to me that you'll end up selecting the median (or high) threshhold several times in a row (approximately majorCompactionInterval / gcWaitTIme times).  It sort of works out because at line 350 all but 1 (or possibly 0 or 2, I'm not sure that the time between calls to now() couldn't cause a race condition where you skip a high/median compaction threshold or do it twice) will fail the curTime - lastMajorCompactionTime > majorCompactionInterval test.

It would be considerably simpler if you changed changeMajorCompactionThreshold to return the current threshold, move the call under the test at line 350, and maintain two pieces of state recording the last cron time where you returned a high or median value.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -264,12 +278,53 @@ public void safeRun() {
...
+        // next fire time is in the interval
+        if (medianMajorCron.nextTimeAfter(ZonedDateTime.now()).toInstant().toEpochMilli()
```
dlg99: Rather general notes after looking at this and the change overall. I can be wrong and often am.

1. reusing cron/time-based parameter change for other parameters looks like a royal pain in the bottom. It is not transparent part of the configuration.
Got to add extra parameter for the cron, one parameter for each cron. 
Want to add extra interval? do the code change or get really creative with cron definitions.

Want to do something similar for i.e. auditor configuration (i.e. to disable it for maintenance windows), change compaction throttling, change journal's buffer size or something else? Got to add configs for crons + replicate the same logic of using the crons.

2. Cron is great for defining scheduled tasks/timers. It is confusing when applied to interval ranges. 
Imagine you have a prod issue at midnight of Friday potentially related to compaction/disk load and you are trying to figure out which values these config parameters had at exactly 23:42 GMT this Friday? 

3. if we want cron, maybe make it actual cron. schedule a task that will run and change any config parameter value. All that code has to do after this is to read config value before using it rather than using the final value read once in constructor.
 
This can be internal cron or simply external that changes config via rest api (insert security considerations here)
internal cron can be configured like 
```text
cron1 = 00 09-18 * * * isAuditorPeriodicCheckOkToRunNow=true, auditorPeriodicBookieCheckInterval=86400, majorCompactionThreshold=0.6
cron1 = 00 11,16 * * * isAuditorPeriodicCheckOkToRunNow=false, majorCompactionThreshold=0.8
```
it does not address p.2 but maybe logging cron events and parameters will suffice.

Overall yes, it does cover that one usecase but I'd love to be able to easily reuse cron functionality for other cases since we are adding code to deal with crons. I think this will push the change from "ok" to "awesome".

sijie: > Cron is great for defining scheduled tasks/timers. It is confusing when applied to interval ranges.

+1 I think one of my comments before was : not changing the threshold, only use cron for triggering compaction. I think that is the most valuable thing to have cron. The cron can also used for triggering other timer tasks, like when to schedule auditing task.

> Overall yes, it does cover that one usecase but I'd love to be able to easily reuse cron functionality for other cases since we are adding code to deal with crons. I think this will push the change from "ok" to "awesome".

+1

I think we can simply this PR with two functionalities:

1) introduce cron, and be able to schedule a task based on cron.
2) don't adjust compaction thresholds, only use cron to schedule when to do compaction.

jvrao: >  introduce cron, and be able to schedule a task based on cron.
    > don't adjust compaction thresholds, only use cron to schedule when to do compaction.

This sums up very well. I  agree with @sijie that time-of-the-day feature can be helpful when entrylog-per-ledger feature is not enabled.

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
ArvinDevel: in this time-based compaction policy, I split the compaction to 3 level, original normal compaction's default value becomes un-distinguish from median and high thresholds.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
ArvinDevel: good idea, thanks.

ArvinDevel:
Thanks for all of you, I'll refactor the logic.

ivankelly:
This looks like a lot of change for something that should really be very simple to implement. 

We have a http endpoint in the bookieserver now. Just create a new endpoint which when called calls GarbageCollectorThread.triggerGC(true, false, false).

Then, this can be triggered via a cronjob locally, or a cronjob in a central location that calls out to all bookies. As a bonus, we don't have to implement our own cron, and have fewer configuration knobs to test and maintain.

sijie:
> We have a http endpoint in the bookieserver now. Just create a new endpoint which when called calls GarbageCollectorThread.triggerGC(true, false, false).

> Then, this can be triggered via a cronjob locally, or a cronjob in a central location that calls out to all bookies. As a bonus, we don't have to implement our own cron, and have fewer configuration knobs to test and maintain.

@ivankelly this can be a solution for advanced users. 

However IMO using an external cron job triggering things increases ops complexity: 1) you have multiple locations to manage things (bookie config, cron configs) 2) you need to debug multiple locations when issues occur (external cron, http service, and the bookie).

I think if there is a robust cron library that we can use, having a cron based scheduler to control when to trigger timer tasks (e.g. compaction, auditor) can be an advanced improvement.

> have fewer configuration knobs to test and maintain

I think the agreement here is trending to a cron expression/mechanism on controlling when to schedule compactions. that says, ideally if we have a robust cron, you don't need to define intervals, we can just write a cron expression in the config file.

ivankelly:
> However IMO using an external cron job triggering things increases ops complexity: 1) you have multiple locations to manage things (bookie config, cron configs) 2) you need to debug multiple locations when issues occur (external cron, http service, and the bookie).

I disagree. Every ops person knows how to use cron, they should be able to tap out a cron schedule from memory. Even if we're cron compatible, the user would have to dig into our docs too see what the format is, and find out where is incompatible with crontab(5). Debugging cron itself is easy. You can even make it send email on failures. 

But it wouldn't even need to be cron, it would be up to the ops to decide what to use to trigger the GC.

The underlying goal of this change is to allow the user more control over when GC runs. A HTTP endpoint, leave the entire decision to the user. 
- They can decide based on time or some other criteria. 
- They can schedule without rebooting the bookies
- They can control the schedule centrally, so all bookies don't end up running GC at the same time
- They can kick off adhoc GC when they see they need to.

And I guess this would take 100-150 LOC, 10% of the current patch. What's more, it wouldn't change anything in the core GC mechanism, it just calls triggerGC that we already test in a bunch of places.

merlimat:
There is a Java library for cron like job scheduling, Quartz. 

My only concern with the scheduling is to avoid having all bookies doing compaction at the same time. Ideally, if prefer all bookies doing it at different times.

sijie:
@ivankelly it is understandable that different people have different opinions regarding ops. We already have http, it doesn't affect people who want to use http. If people doesn't want to use an external cron, they can configure intervals or a cron expression to when they want to schedule compaction.

> And I guess this would take 100-150 LOC, 10% of the current patch. What's more, it wouldn't change anything in the core GC mechanism, it just calls triggerGC that we already test in a bunch of places

If you follow the comments, what we are discussing at the end of this PR is, not change any GC mechanisms, only provided a mechanism of when to trigger the GC tasks.

eolivelli:
@merlimat we already dropped the idea if using quartz, not wanting to add another dependency.

sijie:
@merlimat 

> There is a Java library for cron like job scheduling, Quartz.

Yes, I think that is the best option. Mesos and Aurora use that for cron scheduling as well. But I think Enrico rejects introducing a new library.

> My only concern with the scheduling is to avoid having all bookies doing compaction at the same time. Ideally, if prefer all bookies doing it at different times.

Yes. What people usually do in different systems, is to inject gitters to scheduling, so they won't be started around the same time. Currently we don't even have any control in interval based scheduling (unless yahoo has introduced one).

merlimat:
> Yes. What people usually do in different systems, is to inject gitters to scheduling, so they won't be started around the same time. Currently we don't even have any control in interval based scheduling (unless yahoo has introduced one).

With the current bookies, the compaction is times 1h or 24h after the restart, so after a rolling restart, the compactions are at least a bit skewed by the sequential restart. 

Another approach would be to configure something like that at most 20% of bookies can be doing compaction at the same time, though it gets a bit more involved as a change, and also it could be problematic if compaction is taking very long times.

sijie:
@eolivelli it is server dependency, if it is useful (depends on whether we agreed on to do or not) it is still worth bringing in. I don't think there is a restriction for server dependency. You are paranoid about this because currently client and server are at the same module. it is making things to not introduce a new dependency for clients. but servers shouldn't have such restriction. Things would be better and clearer after we split client and server in future.

sijie:
@merlimat 

> With the current bookies, the compaction is times 1h or 24h after the restart, so after a rolling restart, the compactions are at least a bit skewed by the sequential restart.

yes agreed. they can be started at different time, but if the compaction time is longer than your restart duration, you potentially still have all bookies compaction at the same time.

randomized jitters have same effects as time-skewed introduced in restart sequence.

> Another approach would be to configure something like that at most 20% of bookies can be doing compaction at the same time, though it gets a bit more involved as a change, and also it could be problematic if compaction is taking very long times.

yes that requires some sort of coordination. it is kind of getting things complicated. I think a randomized jitter approach is probably 'good' enough.

so ideally people can just configure compaction, auditor checks either at rate or at cron, with potentially applied jitters.

rate expression: `rate(x minutes)`
cron expression: `cron(...)`
with jitter: `jitter(x minutes)`

so a compaction setting:

compaction: `rate(60 minutes) withJitter(10 minutes)` => run  compaction every 60 minutes, with a randomized jitter of 10 minutes.
compaction: `cron(...) withJitter(10 minutes)` => run  compaction in a cron expression, injecting a randomized jitter of 10 minutes.
auditorCheck: `cron(..)` => schedule auditor check at a given time.

dlg99:
@merlimat @sijie If we are happy with current distribution of compaction across the cluster (or simply do not want to change it right now) instead of triggering compaction on cron we can trigger property value change. Compaction triggers as usual but will have to refresh config properties before doing actual work.

eolivelli:
I can accept to include quartz I have used it production for many years as distributed scheduler,  using a shared jdbc datasource.
I have abandoned it in favour of more scalable approaches and for simple cases I switched to simple zk based scheduler.
We talked about quartz only for parsing cron style config and running a local scheduler.

You are now talking about coordinating bookies and I think this is out of the scope of the patch. I think it is worth working on this aspect

sijie:
> You are now talking about coordinating bookies and I think this is out of the scope of the patch. I think it is worth working on this aspect

I never talked about coordinating bookies. All my discussions in the thread is only about when to trigger compaction.

eolivelli:
@sijie sorry, I didn't want to make confusion. After reading the discussion it seemed to me that we started to talk about a global problem to coordinate heavy tasks, like compaction, in order to achieve stable performance during the day. I think it was my only own thought. I will stick to the     original problem

ivankelly:
@sijie 
> If you follow the comments, what we are discussing at the end of this PR is, not change any GC mechanisms, only provided a mechanism of when to trigger the GC tasks.

Ah, had only looked at latest version of patch, assuming that was what was up for submission.

I'd definitely prefer using something like quartz rather than any implementing our own parsing and scheduling. It looks like it's pretty well tested, system tests and all.

sijie:
@ArvinDevel : 

please check the meeting notes about this issue here: https://cwiki.apache.org/confluence/display/BOOKKEEPER/2018-01-11+Meeting+Notes

You might consider to simply this pull request

asfgit:

sijie:
It seems @ArvinDevel is not working on this PR any more and there is no other motivation for this feature. Closing it now.

PR #853, opened 2017-12-14
BP-14 WriteFlag DEFERRED_SYNC Client Side Implementation

eolivelli:
Introduce implementation of WriteFlag#DEFERRED_SYNC on LedgerHandle (purely client-side).

Tests are only on the client-side (with Mockito), there is no implementation on the bookie side. Most significant tests will come with the force() API, at this point we can only test that we have not broken the LAC advance mechanism.

There is a new class LastAddSyncedManager which will be really used on the next patch, when we will introduce the force() API and the Bookie will be able to return to the client the id of the last entry really sync'd to the journal.


.gitignore
```diff
@@ -20,3 +20,4 @@ lib/
 log/
 target/
 dependency-reduced-pom.xml
+/bookkeeper-server/nbproject/
```
sijie: what is this?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LastAddSyncedManager.java
```diff
@@ -0,0 +1,76 @@
...
+        lastAddSyncedMap.put(bookieIndexHeardFrom, lastAddSynced);
+    }
+
+    /**
+     * Estimates the LastAddConfirmed entry.
+     *
+     * @return the estimated value, considering the lastAddSynced piggybacked value from each bookie
+     */
+    public long calculateCurrentLastAddSynced() {
+        /*
```
sijie: change `/* */` to `//`

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
```diff
@@ -121,7 +121,7 @@ void sendWriteRequest(int bookieIndex) {
         int flags = isRecoveryAdd ? BookieProtocol.FLAG_RECOVERY_ADD : BookieProtocol.FLAG_NONE;
 
         lh.bk.getBookieClient().addEntry(lh.metadata.currentEnsemble.get(bookieIndex), lh.ledgerId, lh.ledgerKey,
-                entryId, toSend, this, bookieIndex, flags);
+                entryId, toSend, this, bookieIndex, flags, lh.writeFlags);
```
sijie: isn't `writeFlags` a thing of `PendingAddOp`? if so, it should be passed in when constructing `PendingAddOps`s.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
```diff
@@ -524,10 +526,15 @@ void writeLac(final long ledgerId, final byte[] masterKey, final long lac, ByteB
...
     void addEntry(final long ledgerId, byte[] masterKey, final long entryId, ByteBuf toSend, WriteCallback cb,
-                  Object ctx, final int options) {
+                  Object ctx, final int options, final EnumSet<WriteFlag> writeFlags) {
         Object request = null;
         CompletionKey completionKey = null;
         if (useV2WireProtocol) {
+            if (writeFlags.contains(WriteFlag.DEFERRED_SYNC)) {
+                LOG.error("invalid writeflags {} for v2 protocol", writeFlags);
+                toSend.release();
+                cb.writeComplete(BKException.Code.IllegalOpException, ledgerId, entryId, addr, ctx);
```
sijie: don't you need to return here?

need a test case to cover this path.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/DeferredSyncClientSideTest.java
```diff
@@ -0,0 +1,58 @@
...
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.Unpooled;
+import org.apache.bookkeeper.client.api.WriteFlag;
+import org.apache.bookkeeper.client.api.WriteHandle;
+import org.junit.Test;
+
+/**
+ * Client side tests on deferred sync write flag.
+ */
+public class DeferredSyncClientSideTest extends MockBookKeeperTestCase {
```
sijie: why do you need a separate test case here? isn't this part of BookKeeperApiTest?

tests/backward/src/test/java/org/apache/bookkeeper/tests/backward/TestBackwardCompat.java
```diff
@@ -314,22 +318,34 @@ void stop() throws Exception {
     }
 
     static class LedgerCurrent {
```
sijie: please bring `4.6` server into testing scope also. you need to test the compatibility between 4.7 and 4.6/4.5.

After 4.6 is out, we need to enforce EOL to clean up backward compat tests.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -1444,17 +1460,23 @@ void sendAddSuccessCallbacks() {
...
+                               pendingAddsSequenceHead);
                 }
                 return;
             }
 
             pendingAddOps.remove();
             explicitLacFlushPolicy.updatePiggyBackedLac(lastAddConfirmed);
-            lastAddConfirmed = pendingAddOp.entryId;
+            if (writeFlags.contains(WriteFlag.DEFERRED_SYNC)) {
+                this.lastAddConfirmed = lastAddSyncedManager.calculateCurrentLastAddSynced();
```
sijie: I didn't see anyone is updating `lastAddSyncedManager`. I would prefer introducing this class until it is used.

for now, you can make a no-op here for `if (writeFlags.contains(WriteFlag.DEFERRED_SYNC))`

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -1444,17 +1460,23 @@ void sendAddSuccessCallbacks() {
                 return;
             }
             // Check if it is the next entry in the sequence.
-            if (pendingAddOp.entryId != 0 && pendingAddOp.entryId != lastAddConfirmed + 1) {
+            if (pendingAddOp.entryId != 0 && pendingAddOp.entryId != pendingAddsSequenceHead) {
```
sijie: I would defer introducing `pendingAddsSequenceHead` until `LastAddSynced` is introduced. At this moment, there is no `LastAddSynced`, so there is no reason to make change at this moment.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LastAddSyncedManager.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Handles LastAddSynced on the client side.
+ */
+class LastAddSyncedManager {
```
sijie: I would suggest taking this class out of this pr, introducing it only after `LastAddSynced` is introduced.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LastAddSyncedManager.java
```diff
@@ -0,0 +1,76 @@
...
+
+    private final Map<Integer, Long> lastAddSyncedMap = new HashMap<>();
```
sijie: I think using `Map<Integer, Long>` here is not efficient. 

- it has a lot of allocations, boxing and unboxing.
- it can be simplified with `long[]`
- most of the ensembles are pretty small (e.g. 3 or 5)

.gitignore
eolivelli: Sorry, will drop. It is a netbeans empty folder

eolivelli: Sorry, will drop. It is a netbeans empty folder

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LastAddSyncedManager.java
eolivelli: Ok

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LastAddSyncedManager.java
eolivelli: Will do in next patch

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
eolivelli: *pendingAddsSequenceHead*  is the core of this  change. We have to handle the sequence of adds without advancing lastAddConfirmed

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
eolivelli: Ok I will introduce it in next patch

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
eolivelli: We can save  a field, we have lh, there is no need to have a field in *PendingAddOp*. 
If you think it would be clearer I can add the field anyway

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
eolivelli: I forgot it while doing the cut and paste from original patch.
Sure I will add a test case

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/DeferredSyncClientSideTest.java
eolivelli: No, this class will contain all the relevant client side tests. In this patch I have only  this one which is worthy, because it states that LAC must not advance even if add ops succeed.
For instance this class will contain all force() API tests for the client side

tests/backward/src/test/java/org/apache/bookkeeper/tests/backward/TestBackwardCompat.java
eolivelli: Yes but I think we need a separate patch to refactor backward compatibility tests.
And with Jia and Ivan today we saw that it is not possible to run the request kind of tests with the actual framework,please see the conversation on slack

sijie: yes, I read the conversation. for this change to be checked in, please test backward between 4.7 and 4.6/4.5. if there is a need for refactor backward compatibility tests, lets refactor them now rather than later. that says I can image more and more changes checked in, we should have something to guarantee we don't break backward compatibility. that says I would mark backward test between 4.7 and 4.6/4.5 as a blocker for this pull request.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/DeferredSyncClientSideTest.java
sijie: okay then I will rename it to "DeferredSyncTest". because it is already under `client` package, there is no need to say "client side".

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
sijie: doesn't we allow different flags for recovery_adds?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
sijie: sure. but my concern is no one is actually using this. I was thinking this pull request only for pushing the flag to the wire. to implementing the actual logic, you should do the server side first and client side later. Otherwise, you are pushing a lot of `unused` logic into the client, and we don't know when they are going to be enabled (given current pull requests are merged very quickly)

eolivelli:
marked this PR as WIP, it has to be reworked, but I will send bookie-side changes first
@sijie

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LastAddSyncedManager.java
eolivelli: done

.gitignore
eolivelli: done

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
eolivelli: as we have pushed server-side changes I will keep *pendingAddsSequenceHead*

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
eolivelli: done

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/DeferredSyncClientSideTest.java
eolivelli: done

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
eolivelli: done

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
eolivelli: done

eolivelli:
@sijie I have addressed all of your review comments, except from dropping *pendingAddsSequenceHead*,  because now we have server side implementation.

Do you think we should add BC tests in this commit ? or it is better to have a new commit/pr ? now each new BC tests requires a new maven module

sijie:
@eolivelli is this still "WIP" or ready for review?

eolivelli:
It is ready

eolivelli:
@sijie do you have cycles for review ?

sijie:
Will review it. Just one question, is server side change done?

eolivelli:
yes.We merged journal and bookie side protocol changes.
This commit does not introduce full cycle tests, I can add

eolivelli:
See https://github.com/apache/bookkeeper/commit/a226dbedff03102291648df3a60e78a2c85ae65f

sijie:
I know the journal change was merged. My question is is all server side changes are done?

eolivelli:
Oh I see. Actually we have to introduce the force() API on the server side and the LastAddSynced piggy back.
IMHO we can consider them separate changes. Do you think we need at least force() on bookie side?

sijie:
@eolivelli it is always good to complete server side changes before adding logics/functionality at client side. otherwise you are adding functionality which is not *usable* at the client, which we don't know when the next change will come in. so typically when we add a new feature, server side first, client side later, and then BC.

eolivelli:
Okay it makes sense.no hurry

I don't know when I will have time to create patch for force() API and the other change. I hope that someone could pick it up

sijie:
Will try to see if I have cycles to help

eolivelli:
@sijie thank you so much

asfgit:

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
eolivelli: we decided to drop the concept of server side LastAddSynced, so this *pendingAddsSequenceHead* is fine in my opinion. IT is like what @jvrao is calling "nonDurableLAC"

eolivelli:
retest this please

eolivelli:
Gently pinging @jvrao and @sijie for review.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
```diff
@@ -307,6 +309,10 @@ public void writeComplete(int rc, long ledgerId, long entryId, BookieSocketAddre
             // bookie client is closed.
             lh.errorOutPendingAdds(rc);
             return;
+        case BKException.Code.IllegalOpException:
```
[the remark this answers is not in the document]
eolivelli: The test case about using v2protocol covers this case

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerFragmentReplicator.java
```diff
@@ -322,7 +324,8 @@ public void readComplete(int rc, LedgerHandle lh,
                 for (BookieSocketAddress newBookie : newBookies) {
                     bkc.getBookieClient().addEntry(newBookie, lh.getId(),
                             lh.getLedgerKey(), entryId, ByteBufList.clone(toSend),
-                            multiWriteCallback, dataLength, BookieProtocol.FLAG_RECOVERY_ADD);
+                            multiWriteCallback, dataLength, BookieProtocol.FLAG_RECOVERY_ADD,
+                            false, EnumSet.noneOf(WriteFlag.class));
```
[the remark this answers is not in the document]
eolivelli: done

.gitignore
```diff
@@ -24,5 +24,6 @@ dependency-reduced-pom.xml
 # Logs
 logs/
 
+=======
```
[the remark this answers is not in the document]
eolivelli: it came with merge master fix/conflicts. Will remove

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
```diff
@@ -584,14 +586,19 @@ void writeLac(final long ledgerId, final byte[] masterKey, final long lac, ByteB
...
 
     void addEntry(final long ledgerId, byte[] masterKey, final long entryId, ByteBufList toSend, WriteCallback cb,
-                  Object ctx, final int options, boolean allowFastFail) {
+                  Object ctx, final int options, boolean allowFastFail, final EnumSet<WriteFlag> writeFlags) {
         Object request = null;
         CompletionKey completionKey = null;
         if (useV2WireProtocol) {
+            if (writeFlags.contains(WriteFlag.DEFERRED_SYNC)) {
+                LOG.error("invalid writeflags {} for v2 protocol", writeFlags);
+                cb.writeComplete(BKException.Code.IllegalOpException, ledgerId, entryId, addr, ctx);
```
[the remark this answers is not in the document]
eolivelli: good catch !. Fixed

eolivelli:
@sijie I have addressed your comments.

About the test case I think it is already covered by the new test about v2 protocol. I can add a more narrow test if you prefer.

eolivelli:
@sijie @jvrao 


My point about LAC:
1) having LAC that is not advancing without guarantees is good to preserve BK feature that "if is it read once it will be always readable" (if you respect LAC). This is current implementation. It will require force() for the LAC to advance.
2) having LAC which advances is like having a per-ledger syncData = false. 

Option 2) also makes things simpler, and we can implement 1) with an additional WriteFlag. We are talking only about client-side behaviour.

So I am leaning towards option 2, I think in BP14 documents we were talking about 1, if @jvrao is OK I will change the implementation.

The force() API will enable the client to sync the ledger and get guarantees about persistence

sijie:
@eolivelli 

> My point about LAC:

I think @jvrao and me discussed before, we agreed on 2). that's why we don't need maintain cursor on bookie side. but I can be wrong, will wait for @jvrao 's comment.

eolivelli:
@sijie great. I will wait for JV ack and simplify the patch. I hope will be able to complete this BP within the end of May

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
jvrao: @sijie  Proposal to have nonDurableLAC is indeed to keep the LAC protocol unchanged. If we update LAC with non-durable LAC value, recovery becomes complicated because recovery assumes that up to LAC things are persisted.  As the piggy-backed LAC gets persisted on the disk, and if the client dies before closing the ledger, Current recovery open logic hunts for writes beyond LAC. But if LAC doesn't reflect persisted entry, then Recovery needs to go all the way to entry-id-0. If we update LAC only when persisted we keep the overall understanding of LAC intact.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -1678,17 +1692,23 @@ void sendAddSuccessCallbacks() {
...
             pendingAddOps.remove();
             explicitLacFlushPolicy.updatePiggyBackedLac(lastAddConfirmed);
-            lastAddConfirmed = pendingAddOp.entryId;
+            if (writeFlags.contains(WriteFlag.DEFERRED_SYNC)) {
+                // cannot update LastAddConfirmed, we are not sure bookie has stored the entry durably at this point
+            } else {
+                this.lastAddConfirmed = Math.max(lastAddConfirmed, pendingAddOp.entryId);
+            }
+
+            pendingAddsSequenceHead = pendingAddOp.entryId + 1;
```
jvrao: I would do:
pendingAddsSequenceHead = pendingAddOp.entryId;
if (!writeFlags.contains(WriteFlag.DEFERRED_SYNC)) {
    lastAddConfirmed = pendingAddsSequenceHead;
}

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerRecoveryOp.java
```diff
@@ -112,6 +112,7 @@ public void readLastConfirmedDataComplete(int rc, RecoveryData data) {
                             synchronized (lh) {
                                 lh.lastAddPushed = lh.lastAddConfirmed = data.getLastAddConfirmed();
                                 lh.length = data.getLength();
+                                lh.pendingAddsSequenceHead = lh.lastAddConfirmed + 1;
```
jvrao: I would like to avoid this logic change; I won't comment on the +1 part for the rest of the code.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -170,6 +181,8 @@
             length = 0;
         }
 
+        this.pendingAddsSequenceHead = lastAddConfirmed + 1;
```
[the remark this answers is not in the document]
eolivelli: Makes sense. Good idea

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -1678,17 +1692,23 @@ void sendAddSuccessCallbacks() {
...
             }
 
             pendingAddOps.remove();
             explicitLacFlushPolicy.updatePiggyBackedLac(lastAddConfirmed);
-            lastAddConfirmed = pendingAddOp.entryId;
+            if (writeFlags.contains(WriteFlag.DEFERRED_SYNC)) {
+                // cannot update LastAddConfirmed, we are not sure bookie has stored the entry durably at this point
+            } else {
+                this.lastAddConfirmed = Math.max(lastAddConfirmed, pendingAddOp.entryId);
+            }
```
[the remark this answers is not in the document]
eolivelli: I will revert, it is mostly a leftover of the very first patch, now things are different

jvrao:
> we agreed on 2)

If we advance LAC in non-durable write response, recovery becomes complicated.
Unless we are willing to recognize the durability in the recovery case, and do one of the following.
1. Recover starting from entryid-0 
2. Never attempt to recover this type of the ledger.

Problem is we can't start from LAC as with the current logic.

jvrao:
Maybe another idea is to have a special case for piggyBackLac and ExplicitLac for non-Durable writes.
Basically, bookie's LAC will get updated by the client only when they are durable. i.e on sync operation.

eolivelli:
@jvrao current idea is to update LAC only using an explicit force() (sync), this way explicit LAC/piggy backed LAC to the bookie is updated only in case of durability guarantees.

This is content of next patch (the client side force() API), I can send the new patch but I wanted to get agreement on this part.

I think that the good direction is to go on with the current approach (option 1) and have 'LastAddConfirmed" advanced only in front of durable guarantees (explicit sync).
This is preserving the overall LAC protocol and the guarantees from the reader and replicator point of view.

If you guys prefer I can add the force() API to this patch so that we can have a global view of the changes on the client side in a single patch. 

cc @sijie

eolivelli:
@jvrao  @sijie another idea to implement both of the two behaviours

we can do that DEFERRED_SYNC writeflag, acts like in this patch, that is LAC does not advance without explicit force()

we can add an additional write flag, like OPTIMISTIC_LAC, to be set togheder with DEFERRED_SYNC , which makes lastAddConfirmed advance in sendAddSuccessCallbacks even in presence of DEFERRED_SYNC

like:

```
if (!writeFlags.contains(WriteFlag.DEFERRED_SYNC)  
          || writeFlags.contains(WriteFlag.OPTIMISTIC_LAC)) {

                this.lastAddConfirmed = pendingAddsSequenceHead;
}
 ```

**Comment by eolivelli** [comment]
@sijie  and @jvrao as we are opting for option 1 this patch is ready for review.

I will add tests for "real" client to bookie communication in the next patch.
I can do it here if you prefer

File: bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java [hunk 19]
```diff
@@ -1678,17 +1692,23 @@ void sendAddSuccessCallbacks() {
                 return;
             }
             // Check if it is the next entry in the sequence.
-            if (pendingAddOp.entryId != 0 && pendingAddOp.entryId != lastAddConfirmed + 1) {
```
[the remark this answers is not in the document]
eolivelli: @jvrao done

eolivelli:
@sijie do you have other comments about this change ?

sijie:
@jvrao please review this PR again

jvrao:
Looks good to me. Thanks for iterating over it @eolivelli  and @sijie

eolivelli:
@jvrao  @sijie thank you. 
I am merging this one. I will send next patch hopefully tomorrow.

PR #854, opened 2017-12-14
Add attribution for binary jars covered by New BSD License

ivankelly:
This means a separate LICENSE file for server and all, as all pulls in
more stuff.


bookkeeper-dist/src/main/resources/LICENSE-all.bin.txt
```diff
@@ -0,0 +1,707 @@
...
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+Code generated by the Protocol Buffer compiler is owned by the owner
+of the input file used when generating it.  This code is not
+standalone and requires a support library to be linked with it.  This
+support library is itself covered by the above license.
+
+------------------------------------------------------------------------------------
+For lib/javax.servlet-javax.servlet-api-3.1.0.jar
```
sijie: please remove version. otherwise, it becomes really *painful* to maintain this.

bookkeeper-dist/src/main/resources/LICENSE-all.bin.txt
```diff
@@ -0,0 +1,707 @@
...
+CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+------------------------------------------------------------------------------------
+For lib/org.scala-lang.modules-scala-parser-combinators_2.11-1.0.4.jar
```
sijie: why this one is separated from `lib/org.scala-lang*.jar`

bookkeeper-dist/src/main/resources/LICENSE-all.bin.txt
```diff
@@ -0,0 +1,707 @@
...
+
+   As between Initial Developer and the Contributors, each party is
+   responsible for claims and damages arising, directly or indirectly,
+   out of its utilization of rights under this License and You agree
+   to work with Initial Developer and Contributors to distribute such
+   responsibility on an equitable basis. Nothing herein is intended or
+   shall be deemed to constitute any admission of liability.
+
+------------------------------------------------------------------------------------
+For lib/com.thoughtworks.paranamer-paranamer-2.8.jar
```
sijie: please remove version as well.

bookkeeper-dist/src/main/resources/LICENSE-server.bin.txt
```diff
@@ -0,0 +1,613 @@
...
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+Code generated by the Protocol Buffer compiler is owned by the owner
+of the input file used when generating it.  This code is not
+standalone and requires a support library to be linked with it.  This
+support library is itself covered by the above license.
+
+------------------------------------------------------------------------------------
+For lib/javax.servlet-javax.servlet-api-3.1.0.jar
```
sijie: remove version

bookkeeper-dist/src/main/resources/LICENSE-all.bin.txt
ivankelly: licensing can change between versions. a case in point is protobuf. We ship with protobuf 3.4.0. The license for protobuf in the this file doesn't match https://github.com/google/protobuf/blob/v3.4.0/LICENSE. There are notices for 2 c++ headers, which doesn't apply to us, but also the copyright year has changed. There are 6 years of changes we are not attributing to google.

That said, we shouldn't expect devs to remember to update notice and license each time they update a dependency. We should have a script that actually checks this stuff, which I plan on writing next week.

bookkeeper-dist/src/main/resources/LICENSE-all.bin.txt
ivankelly: It's a different project, different copyright years.

bookkeeper-dist/src/main/resources/LICENSE-all.bin.txt
sijie: I agree with matching versions and automating this. but assume if automation is not done, removing version is the most convenient way for maintenance. We can't assume that someone is going to write automate in future. we have to assume no one is doing the work, what is the best for the project to maintain.

At the minimal, I would prefer us being consistent here, either we all use specific versions, or we remove versions. That's why I prefer removing versions here.

bookkeeper-dist/src/main/resources/LICENSE-all.bin.txt
sijie: okay

bookkeeper-dist/src/main/resources/LICENSE-all.bin.txt
ivankelly: @sijie I've added the version to them all now. The filenames were wrong as well.

A little pain is not a bad thing. It's an incentive to actually automate this. In any case, it's much more painful to have to -1 a release candidate when everything else looks good, just because of licences. I've -1d 3 RCs (2 BK, 1 Pulsar) in as many weeks, due to bundled jars not being handled correctly.

I hope to have a script ready early next week. To be clear, the script will only verify that the LICENSE and NOTICE files reference all bundled jars. Updating the files will still be manual, as it should be.

sijie:
merged to branch-4.6

sijie:
cherry-picked back to master as well.

sijie:
thanks @ivankelly for fixing this.

@jiazhai you should be unblocked for next release candidate.

Issue #590, opened 2017-10-17
CRC32C for bookkeeper


ivankelly:
JIRA: https://issues.apache.org/jira/browse/BOOKKEEPER-951

Reporter: Venkateswararao Jujjuri (JV) @jvrao

CRC32C is hardware based CRC32 and it substantially improves performance and practically offloads CPU. 

[~merlimat] at yahoo implemented this for Pulsar, opening this item to port that change.

https://github.com/yahoo/pulsar/tree/master/pulsar-checksum

Java 9 has CRC32C native support but we need this JNI patch to achieve it on Java 8.

PR #856, opened 2017-12-15
ISSUE #590 (@bug W-4556980@) Reduce excessive CPU usage on client side: add crc3…

dlg99:
…2c with SSE processor instruction support, benchmark to make sure it is better than crc32

CRC32C for bookkeeper

Master Issue: #590

### DIFF
--- a/bookkeeper-server/pom.xml
+++ b/bookkeeper-server/pom.xml
@@ -182,6 +182,11 @@
       <artifactId>http-server</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.apache.pulsar</groupId>
+      <artifactId>pulsar-checksum</artifactId>
+      <version>1.20.0-incubating</version>
+    </dependency>
   </dependencies>
   <build>
     <plugins>
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/CRC32CDigestManager.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/CRC32CDigestManager.java
@@ -1,0 +1,65 @@
+package org.apache.bookkeeper.client;
+
+/*
+* Licensed to the Apache Software Foundation (ASF) under one
+* or more contributor license agreements.  See the NOTICE file
+* distributed with this work for additional information
+* regarding copyright ownership.  The ASF licenses this file
+* to you under the Apache License, Version 2.0 (the
+* "License"); you may not use this file except in compliance
+* with the License.  You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+import com.scurrilous.circe.crc.Sse42Crc32C;
+import io.netty.buffer.ByteBuf;
+import org.apache.commons.lang3.mutable.MutableBoolean;
+import org.apache.commons.lang3.mutable.MutableInt;
+import org.apache.pulsar.checksum.utils.Crc32cChecksum;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+class CRC32CDigestManager extends DigestManager {
+    static final Logger LOG = LoggerFactory.getLogger(CRC32CDigestManager.class);
+
+    private final ThreadLocal<MutableInt> currentCrc = ThreadLocal
+            .withInitial(() -> new MutableInt(0));
+    private final ThreadLocal<MutableBoolean> isNewCrc = ThreadLocal
+            .withInitial(() -> new MutableBoolean(true));
+
+    public CRC32CDigestManager(long ledgerId) {
+        super(ledgerId);
+        if (!Sse42Crc32C.isSupported()) {
+            LOG.error("Sse42Crc32C is not supported, will use less slower CRC32C implementation.");
+        }
+    }
+
+    @Override
+    int getMacCodeLength() {
+        return 4;
+    }
+
+    @Override
+    void populateValueAndReset(ByteBuf buf) {
+        buf.writeInt(currentCrc.get().intValue());
+        isNewCrc.get().setTrue();
+    }
+
+    @Override
+    void update(ByteBuf data) {
+        if (isNewCrc.get().isTrue()) {
+            isNewCrc.get().setFalse();
+            currentCrc.get().setValue(Crc32cChecksum.computeChecksum(data));
+        } else {
+            final int lastCrc = currentCrc.get().intValue();
+            currentCrc.get().setValue(Crc32cChecksum.resumeChecksum(lastCrc, data));
+        }
+    }
+}
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
@@ -119,7 +119,7 @@ public class LedgerMetadata implements org.apache.bookkeeper.client.api.LedgerMe
         this.metadataFormatVersion = CURRENT_METADATA_FORMAT_VERSION;
 
         this.digestType = digestType.equals(BookKeeper.DigestType.MAC)
-            ? LedgerMetadataFormat.DigestType.HMAC : LedgerMetadataFormat.DigestType.CRC32;
+            ? LedgerMetadataFormat.DigestType.HMAC : LedgerMetadataFormat.DigestType.valueOf(digestType.toString());
         this.password = Arrays.copyOf(password, password.length);
         this.hasPassword = true;
         if (customMetadata != null) {
@@ -230,10 +230,15 @@ public class LedgerMetadata implements org.apache.bookkeeper.client.api.LedgerMe
 
     @Override
     public DigestType getDigestType() {
-        if (digestType.equals(LedgerMetadataFormat.DigestType.HMAC)) {
-            return DigestType.MAC;
-        } else {
-            return DigestType.CRC32;
+        switch (digestType) {
+            case HMAC:
+                return DigestType.MAC;
+            case CRC32:
+                return DigestType.CRC32;
+            case CRC32C:
+                return DigestType.CRC32C;
+            default:
+                throw new IllegalArgumentException("Unable to convert digest type " + digestType);
         }
     }
 
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieWriteLedgersWithDifferentDigestsTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieWriteLedgersWithDifferentDigestsTest.java
@@ -54,7 +54,7 @@ public class BookieWriteLedgersWithDifferentDigestsTest extends
             .getLogger(BookieWriteLedgersWithDifferentDigestsTest.class);
 
     byte[] ledgerPassword = "aaa".getBytes();
-    LedgerHandle lh, lh2;
+    LedgerHandle lh;
     Enumeration<LedgerEntry> ls;
 
     // test related variables
@@ -65,6 +65,7 @@ public class BookieWriteLedgersWithDifferentDigestsTest extends
     ArrayList<byte[]> entries2; // generated entries
 
     private final DigestType digestType;
+    private final DigestType otherDigestType;
 
     private static class SyncObj {
         volatile int counter;
@@ -77,7 +78,7 @@ public class BookieWriteLedgersWithDifferentDigestsTest extends
 
     @Parameterized.Parameters
     public static Collection<Object[]> configs() {
-        return Arrays.asList(new Object[][] { {DigestType.MAC }, {DigestType.CRC32}});
+        return Arrays.asList(new Object[][] { {DigestType.MAC }, {DigestType.CRC32}, {DigestType.CRC32C} });
     }
 
     @Override
@@ -93,6 +94,7 @@ public class BookieWriteLedgersWithDifferentDigestsTest extends
     public BookieWriteLedgersWithDifferentDigestsTest(DigestType digestType) {
         super(3);
         this.digestType = digestType;
+        this.otherDigestType = digestType == DigestType.CRC32 ? DigestType.MAC : DigestType.CRC32;
         String ledgerManagerFactory = "org.apache.bookkeeper.meta.HierarchicalLedgerManagerFactory";
         // set ledger manager
         baseConf.setLedgerManagerFactoryClassName(ledgerManagerFactory);
@@ -103,11 +105,11 @@ public class BookieWriteLedgersWithDifferentDigestsTest extends
     public void testLedgersWithDifferentDigestTypesNoAutodetection() throws Exception {
         bkc.conf.setEnableDigestTypeAutodetection(false);
         // Create ledgers
-        lh = bkc.createLedgerAdv(3, 2, 2, DigestType.MAC, ledgerPassword);
+        lh = bkc.createLedgerAdv(3, 2, 2, digestType, ledgerPassword);
 
         final long id = lh.ledgerId;
 
-        LOG.info("Ledger ID-1: " + lh.getId());
+        LOG.info("Ledger ID: {}, digestType: {}", lh.getId(), digestType);
         SyncObj syncObj1 = new SyncObj();
         for (int i = numEntriesToWrite - 1; i >= 0; i--) {
             ByteBuffer entry = ByteBuffer.allocate(4);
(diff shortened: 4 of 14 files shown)


microbenchmarks/pom.xml
```diff
@@ -0,0 +1,151 @@
+<!-- Licensed to the Apache Software Foundation (ASF) under one or more contributor
```
merlimat: AFS header appears to be misformatted.

microbenchmarks/results.txt
```diff
@@ -0,0 +1,89 @@
+# on macbook pro
```
merlimat: @dlg99 One thing to note is that the SSE4.2 instructions are using a JNI module and that's only compiled for linux in the released artifacts for `pulsar-checksum`.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
```diff
@@ -119,7 +119,7 @@ public LedgerMetadata(int ensembleSize,
         this.metadataFormatVersion = CURRENT_METADATA_FORMAT_VERSION;
 
         this.digestType = digestType.equals(BookKeeper.DigestType.MAC)
-            ? LedgerMetadataFormat.DigestType.HMAC : LedgerMetadataFormat.DigestType.CRC32;
+            ? LedgerMetadataFormat.DigestType.HMAC : LedgerMetadataFormat.DigestType.valueOf(digestType.toString());
```
sijie: Can this be simplified with `this.digestType = LedgerMetadataFormat.DigestType.valueOf(digestType.toString())`?

microbenchmarks/pom.xml
```diff
@@ -0,0 +1,151 @@
...
+    language governing permissions and limitations under the License. -->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.apache.bookkeeper</groupId>
+        <artifactId>bookkeeper</artifactId>
+        <version>4.7.0-SNAPSHOT</version>
+    </parent>
+    <artifactId>microbenchmarks</artifactId>
```
sijie: There is actually a module `jmh` under `tests`: https://github.com/apache/bookkeeper/tree/master/tests/jmh

We can push this jmh benchmark in that module.

microbenchmarks/results.txt
sijie: we probably don't need this file to be checked in.

microbenchmarks/run.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
```
sijie: add license header?

bookkeeper-server/pom.xml
```diff
@@ -182,6 +182,11 @@
       <artifactId>http-server</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
```
sijie: nit: I think we probably want to move the whole `checksum` package from pulsar to bookkeeper. but it is okay for this pull request. I will discuss with @merlimat on how to handle that.

microbenchmarks/src/main/java/org/apache/bookkeeper/client/DigestTypeBenchmark.java
```diff
@@ -0,0 +1,172 @@
...
+
+            crc32c = DigestManager.instantiate(ThreadLocalRandom.current().nextLong(0, Long.MAX_VALUE),
+                    password, DigestType.CRC32C);
+
+            mac = DigestManager.instantiate(ThreadLocalRandom.current().nextLong(0, Long.MAX_VALUE),
+                    password, DigestType.MAC);
+
+            digestBuf = Unpooled.buffer(getDigestManager(digest).getMacCodeLength());
+
+            final int entrySizeKb = 64;
```
merlimat: Might be worth comparing also with different entry sizes.

microbenchmarks/pom.xml
dlg99: I added this one for a few reasons:
- I already had it like that ;)
- jmh-bench did not build even after I fixed version in pom. "ProtocolBenchmark.java:[88,41] constructor AddRequest in class org.apache.bookkeeper.proto.BookieProtocol.AddRequest cannot be applied to given types". 
- I was not sure what the plans fro jmh-bench are, given its current state.

microbenchmarks/results.txt
```diff
@@ -0,0 +1,89 @@
...
+DigestTypeBenchmark.digestManager:·gc.churn.G1_Old_Gen                    NOT_ARRAY_BACKED     CRC32  thrpt    5      0.158 ±    0.078  MB/sec
+DigestTypeBenchmark.digestManager:·gc.churn.G1_Old_Gen.norm               NOT_ARRAY_BACKED     CRC32  thrpt    5      1.439 ±    0.769    B/op
+DigestTypeBenchmark.digestManager:·gc.count                               NOT_ARRAY_BACKED     CRC32  thrpt    5    784.000             counts
+DigestTypeBenchmark.digestManager:·gc.time                                NOT_ARRAY_BACKED     CRC32  thrpt    5   2949.000                 ms
+DigestTypeBenchmark.digestManager:·stack                                  NOT_ARRAY_BACKED     CRC32  thrpt             NaN                ---
+DigestTypeBenchmark.digestManager                                         NOT_ARRAY_BACKED   CRC32_C  thrpt    5    130.444 ±    3.441  ops/ms
+DigestTypeBenchmark.digestManager:·gc.alloc.rate                          NOT_ARRAY_BACKED   CRC32_C  thrpt    5   7824.727 ±  212.028  MB/sec
+DigestTypeBenchmark.digestManager:·gc.alloc.rate.norm                     NOT_ARRAY_BACKED   CRC32_C  thrpt    5  65552.001 ±    0.001    B/op
```
merlimat: Having 64Kb allocated (and GCed) per each buffer seems to indicate that something is off. Prob, that's because of the pure java implementation fallback due to not having the JNI lib on MacOS.

microbenchmarks/results.txt
dlg99: @sijie It takes awhile to get results, I assume we don't want to run these on each build.
OTOH having benchmark results easily accessible may help people decide on digest type to use. 
I can remove it, of course, if you insist.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
dlg99: @sijie I wish, there is "HMAC" vs "MAC" case.

microbenchmarks/pom.xml
sijie: don't worry at this patch. I think jmh module was forgot due to some merge issues before. I had the change to fix that. I can incorporate your module in my change.

microbenchmarks/results.txt
sijie: @dlg99 okay for me.

merlimat: I think it's good to have it somewhere. To help people decide, probably we should also mention a summary of the differences in the Javadoc and website.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
sijie: gotcha

microbenchmarks/results.txt
dlg99: @merlimat in unit tests I did not see this log: 
        if (!Sse42Crc32C.isSupported()) {
            LOG.error("Sse42Crc32C is not supported, will use less slower CRC32C implementation.");
        }
so I assume that it won't use pure java implementation.

This is because DoubleByteBuf returns  false == hasMemoryAddress() and false == hasArray()

    public static int computeChecksum(ByteBuf payload) {
        if (payload.hasMemoryAddress() && (CRC32C_HASH instanceof Sse42Crc32C)) {
            return CRC32C_HASH.calculate(payload.memoryAddress() + payload.readerIndex(), payload.readableBytes());
        } else if (payload.hasArray()) {
            return CRC32C_HASH.calculate(payload.array(), payload.arrayOffset() + payload.readerIndex(),
                    payload.readableBytes());
        } else {
            return CRC32C_HASH.calculate(payload.nioBuffer());
        }
    }

ends up in .nioBuffer() which ends up doing the following:

      >>>>>   ByteBuffer dst = ByteBuffer.allocate(length);
        ByteBuf b = Unpooled.wrappedBuffer(dst);

merlimat: Oh, that's true. That's because it's doing that on a `DoubleByteBuf`. We should make sure we don't do that in the real code :)

bookkeeper-server/pom.xml
dlg99: @sijie I'll address the comments later/tomorrow-ish.

On the other note I also tried/compared adler32 (from jdk) and crc32c from jdk9 (ported it to run with jdk8), neither were better than regular crc32.
oh, and I tried DUMMY digest (aka no digest). This one crazy fast and garbage free. Unfortunately, it is useless.

sijie: @dlg99 did you run with jdk9? I've seem very good number with jdk9. in java 9, the crc32c with sse42 is intrinsic on jvm. I will send my results as well.

merlimat: In the checksum module we should be able to check for the JVM runtime version and select the fastest available implementation.

dlg99: @sijie I ran with jdk8. I tweaked http://hg.openjdk.java.net/jdk9/dev/jdk/file/feef46bbf70b/src/java.base/share/classes/java/util/zip/CRC32C.java to work on jdk8 and, to my surprise, saw no performance difference compared with jdk8's crc32.

microbenchmarks/run.sh
dlg99: will do

microbenchmarks/src/main/java/org/apache/bookkeeper/client/DigestTypeBenchmark.java
dlg99: will do

microbenchmarks/pom.xml
dlg99: will fix

dlg99:
@sijie @merlimat I've addressed the CR feedback. 
Please let me know if you want me to squash commits together, I assumed this will be done as part of pull request's merge.

bookkeeper-server/pom.xml
sijie: @dlg99 the approach you took didn't work because the assemble code of using intel crc32c instruction is a hotspot code which only available in jdk9. you can port the code from java9 to java8 and it can compile and work, but jdk8 doesn't this assemble code, so why you won't see any performance difference.

You can checkout the hotspot code here http://hg.openjdk.java.net/jdk9/jdk9/hotspot/rev/a60e232aa8f2#l11.20

sijie: Also check my benchmark at #858

dlg99:
Build failure is unrelated to my changes:

"[ERROR] /home/jenkins/jenkins-slave/workspace/bookkeeper-precommit-pullrequest/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java:[73,8] org.apache.bookkeeper.bookie.storage.ldb.DbLedgerStorage is not abstract and does not override abstract method "

sijie:
@dlg99 that was due two concurrent merges. the first one change the method, the second one was introducing a new implmenetation based on the old interface. it should be fixed now. I am triggering ci builds.

sijie:
retest this please

sijie:
close/reopen the pull request to trigger travis ci build as well.

dlg99:
@sijie Jenkins succeeded. 
Travis CI was runs rat plugin which was not happy to see results.txt without apache header.
I added header and squashed commits. Hopefully this takes care of things.

What is the right way to run maven locally to hit all these checks? "mvn clean install" is not enough to mimic travis's behavior.

sijie:
cool.
```
maven clean apache-rat:check install spotbugs:check
```

`apache-rat:check` checks the license headers
`spotbugs:check` checks the bugs

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/CRC32CDigestManager.java
```diff
@@ -0,0 +1,65 @@
...
+    public CRC32CDigestManager(long ledgerId) {
+        super(ledgerId);
+        if (!Sse42Crc32C.isSupported()) {
+            LOG.error("Sse42Crc32C is not supported, will use less slower CRC32C implementation.");
+        }
+    }
+
+    @Override
+    int getMacCodeLength() {
+        return 4;
```
sijie: @merlimat @dlg99 shall we stick to 8 like crc32 for consistency? assume we might use jdk9 crc32c in future.

https://docs.oracle.com/javase/9/docs/api/java/util/zip/CRC32C.html#getValue--

dlg99: @sijie Is there any guarantee that these will be interchangeable? i.e. jdk's one may  use actual full long and pulsar's one will have int + another int (i.e. 0) as a filler. So we won't be able to upgrade bookie if it has ledgers created with one crc32c implementation and bookie just switched to another. Safer way would be to have yet another digest type (CRC32C_J) and set it explicitly.

merlimat:
> close/reopen the pull request to trigger travis ci build as well.

@sijie I think for Travis you can trigger the rebuild in the Travis page directly (and it will update the PR, unlike Jenkins)

bookkeeper-server/pom.xml
jvrao: crc32c will have its fullest potential only if it is using hardware instruction set. (obviously) :)

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieWriteLedgersWithDifferentDigestsTest.java
```diff
@@ -93,21 +94,22 @@ public void setUp() throws Exception {
...
         this.digestType = digestType;
+        this.otherDigestType = digestType == DigestType.CRC32 ? DigestType.MAC : DigestType.CRC32;
         String ledgerManagerFactory = "org.apache.bookkeeper.meta.HierarchicalLedgerManagerFactory";
         // set ledger manager
         baseConf.setLedgerManagerFactoryClassName(ledgerManagerFactory);
         baseClientConf.setLedgerManagerFactoryClassName(ledgerManagerFactory);
     }
 
-    @Test
+    @Test()
```
eolivelli: nit: remove parentheses

dlg99: done.

sijie:
@merlimat what is your thought about making the crc32c value exchangeable between pulsar implementation and java 9 implementation?

sijie:
I think this change is almost ready to go. I can incorporate into my PR #858 and move pulsar-checksum module into bookkeeper and make the crc32c value exchangeable between different implementations.

sijie:
@merlimat any thoughts here?

sijie:
ping @merlimat ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/CRC32CDigestManager.java
sijie: @dlg99 I think it should be exchangeable. 

CRC32C is defined as exactly 4 bytes. But I think most of java versions returns a long, is to 1) have compatibility with java.util.zip.Checksum and 2) to avoid signedness issues with putting a 4-byte quantity in a signed int. I think we can stick to 4 bytes or 8 bytes, it doesn't really matter here.

sijie:
Can any other committers take a look at this? If we agree with CRC32C and 4 bytes, let's move forward with this.

bookkeeper-server/pom.xml
sijie: Created #915 for tracking importing the pulsar-checksum module to bookkeeper as discussed with @merlimat before. 

It *doesn't* block this PR. This PR can go in first if we agreed with 4 bytes code length with crc32c.

eolivelli:
+1  for crc32c and 4 bytes from me

sijie:
My comment is merge this pull request if we agreed on crc32c and have the other pull requests to import pulsar-checksum.

eolivelli:
Ok for merging this
+1

sijie:
@merlimat can you review this when you have time?

sijie:
@merlimat when you have time, can you review this again? If you are good with this, let move forward with this patch.

PR #857, opened 2017-12-15
Issue 778: Make build pass using JDK9

sijie:

- make bookkeeper can be compiled with java 9
- add java9 to travis ci to make sure it can be built for both java8 and java9


sijie:
This change was for testing crc32c on java9.

PR #860, opened 2017-12-15
DbLedgerStorage -- Added config options in bk_server.conf

merlimat:
Added commented config options with default values.

### DIFF
--- a/bookkeeper-server/conf/bk_server.conf
+++ b/bookkeeper-server/conf/bk_server.conf
@@ -574,3 +574,42 @@ zkEnableSecurity=false
 # This provides the plugin run extra services along with a bookie server.
 #
 # extraServerComponents=
+
+
+#############################################################################
+## DB Ledger storage configuration
+#############################################################################
+
+# These configs are used when the selected 'ledgerStorageClass' is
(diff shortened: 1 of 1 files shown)


bookkeeper-server/conf/bk_server.conf
```diff
@@ -574,3 +574,42 @@ zkEnableSecurity=false
 # This provides the plugin run extra services along with a bookie server.
 #
 # extraServerComponents=
+
+
+#############################################################################
```
sijie: Can you add those settings to  documentation? https://github.com/apache/bookkeeper/blob/master/site/_data/config/bk_server.yaml

eolivelli:
It seems that master is not compilable on travis, but it is not due to this patch

sijie:
retest this please

sijie:
close/reopen to trigger travis ci

bookkeeper-server/conf/bk_server.conf
merlimat: Sure, will do that before merging

merlimat:
> +1 for me. I think that it will be very important to write at least one page of docs about this great feature 

Yes, will do that before 4.7 is ready

Issue #849, opened 2017-12-14
Mult. formats for LedgerId in BKShell


reddycharan:
**FEATURE REQUEST**

1. Please describe the feature you are requesting.

- Added LedgerIdFormatter, so it can be used for
configuring format of LedgerId (Long/hex/UUID) in input/output
of BKShell command 
- provide config options for both ledgeridformattet and entryformatter and java options for the same for BKShell commands
- enhance ReadLedgerEntriesCmd command to have options rather than as arguments
- fix checkstyle errors - "Redundant 'final' modifier." and for invalid variable name format

2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have). Are you currently using any workarounds to address this issue?

it is kind-of must-have for our use case, since in all other components of our system LedgerId, is in UUID format.

PR #864, opened 2017-12-15
ISSUE #849: LedgerIdFormatter and EntryFormatter Config options

reddycharan:

- LedgerIdFormatter, so it can be used for configuring format of LedgerId (Long/hex/UUID) in input/output of BKShell command
- provide config options for both ledgeridformattet and entryformatter and java options for the same for BKShell commands
- enhance ReadLedgerEntriesCmd command to have options rather than as arguments
- fix checkstyle errors - "Redundant 'final' modifier." and for invalid variable name format

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
@@ -6,9 +6,9 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
- * <p>
+ *
  * http://www.apache.org/licenses/LICENSE-2.0
- * <p>
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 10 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
```diff
@@ -6,9 +6,9 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
- *
- *     http://www.apache.org/licenses/LICENSE-2.0
- *
+ * <p>
```
sijie: I don't think we can have this in the license header.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
```diff
@@ -7,7 +7,7 @@
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
  *
- *     http://www.apache.org/licenses/LICENSE-2.0
+ * http://www.apache.org/licenses/LICENSE-2.0
```
sijie: @reddycharan - can you add spaces for this?

reddycharan:
these test failures are not related to this change

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
reddycharan: done

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
reddycharan: hmm..i didnt add this tag manually, format/checksum must have added this tag. Anyhow I removed them manually.

sijie:
@reddycharan thank you for the change. merged it.

PR #867, opened 2017-12-15
During journal replay, force to re-insert the master key in ledger storage

merlimat:
When the bookie crashes and restart, it will replay the entries from the journal. In the case of ledgers that were created, but their metadata was not flushed on disk, the ledger metadata might be lost. 

This is more evident in `DbLedgerStorage`, since the ledger metadata is cached in memory and it's only inserted in the db at the checkpoint time, though it applies as well for `InterleavedLedgerStorage` where the ledger metadata was inserted in the cache but not fsynced on disk. 

If the page cache content is gone, when replaying the journal content, the ledger will be tracked in memory but the bookie won't have the metadata on disk anymore.

Merging from yahoo branch: https://github.com/yahoo/bookkeeper/commit/d3626396ecf1a2b98c782bb30a7b9c00c75c6cdf


eolivelli:
+1

Issue #863, opened 2017-12-15
BK client error messages needs to be more descriptive


pasha-kuznetsov:
BK Client prints error messages as follows:
```
20171212011311.000426 [ERROR] [clientWorker-orderedsafeexecutor-0-0] [LedgerHandle.java:1055] - 
Closing ledger 4099501146997097674 due to error -101
```
it'd be more user-friendly if it gave a description of what `-101` meant, for example:
```
20171212011311.000426 [ERROR] [clientWorker-orderedsafeexecutor-0-0] [LedgerHandle.java:1055] - 
Closing ledger 4099501146997097674 due to error -101:
Ledger has been fenced off. Some other client must have opened it to read
```

merlimat:
👍 Nice suggestion. Also it doesn't make sense to print `-101`. We should rather use the enum name for that error.

PR #868, opened 2017-12-16
ISSUE #863: BK client error messages need to be more descriptive

pasha-kuznetsov:
This improves logging of error codes as described in the issue #863. I only addressed the most glaring cases where bare `rc` was logged, by introducing a new logging utility which can be used as follows:
```
LOG.error("Closing ledger {} due to {}", ledgerId, BKException.codeLogger(rc));
```
producing messages similar to the following:
> 2017-12-15 17:01:11,602 - ERROR - [Builder-OrderedSafeExecutor-1-0:LedgerHandle@1398] - Closing ledger 1 due to **LedgerFencedException: Ledger has been fenced off. Some other client must have opened it to read**

We can also discuss if the cases similar to the following should be brought in line with the above changes, which produces more information compared to the new approach, including stack trace, although not the original one:
```
    LOG.error("BK error opening ledger: " + lId, BKException.create(rc)); // <------------------
    finalLedgerIterCb.processResult(rc, null, null);
    return;
}
```

P.S. Please also note that we (Salesforce) are starting to add commit tags simplifying our merge process, e.g. `(@bug W-3843968@)` -- please let us know if that is significant enough to be discussed separately.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDescriptorImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDescriptorImpl.java
@@ -26,6 +26,7 @@ import io.netty.buffer.ByteBuf;
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.concurrent.atomic.AtomicBoolean;
+import org.apache.bookkeeper.client.api.BKException;
 import org.apache.bookkeeper.common.util.Watcher;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -122,7 +123,7 @@ public class LedgerDescriptorImpl extends LedgerDescriptor {
         }
         ByteBuf entry = createLedgerFenceEntry(ledgerId);
         journal.logAddEntry(entry, (rc, ledgerId, entryId, addr, ctx) -> {
-            LOG.debug("Record fenced state for ledger {} in journal with rc {}", ledgerId, rc);
+            LOG.debug("Record fenced state for ledger {} in journal with rc {}", ledgerId, BKException.codeLogger(rc));
             if (rc == 0) {
                 fenceEntryPersisted.compareAndSet(false, true);
                 result.set(true);
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDescriptorImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDescriptorImpl.java
@@ -124,7 +124,8 @@ public class LedgerDescriptorImpl extends LedgerDescriptor {
         ByteBuf entry = createLedgerFenceEntry(ledgerId);
         journal.logAddEntry(entry, (rc, ledgerId, entryId, addr, ctx) -> {
             if (LOG.isDebugEnabled()) {
-                LOG.debug("Record fenced state for ledger {} in journal with rc {}", ledgerId, BKException.codeLogger(rc));
+                LOG.debug("Record fenced state for ledger {} in journal with rc {}",
+                        ledgerId, BKException.codeLogger(rc));
             }
             if (rc == 0) {
                 fenceEntryPersisted.compareAndSet(false, true);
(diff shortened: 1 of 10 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDescriptorImpl.java
```diff
@@ -122,7 +123,7 @@ ByteBuf getExplicitLac() {
         }
         ByteBuf entry = createLedgerFenceEntry(ledgerId);
         journal.logAddEntry(entry, (rc, ledgerId, entryId, addr, ctx) -> {
-            LOG.debug("Record fenced state for ledger {} in journal with rc {}", ledgerId, rc);
+            LOG.debug("Record fenced state for ledger {} in journal with rc {}", ledgerId, BKException.codeLogger(rc));
```
dlg99: wrap this with

   if (LOG.isDebugEnabled()) {
       LOG.debug ...
   }

codeLogger() creates new objects and in 99.9% of cases debug logging is off. This lets us bypass extra object creation.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/api/BookKeeperApiTest.java
```diff
@@ -38,6 +43,9 @@
 import org.apache.bookkeeper.client.BKException.BKNoSuchLedgerExistsException;
 import org.apache.bookkeeper.client.BKException.BKUnauthorizedAccessException;
 import org.apache.bookkeeper.client.MockBookKeeperTestCase;
+import org.apache.bookkeeper.util.LoggerOutput;
+import org.apache.log4j.spi.LoggingEvent;
```
dlg99: don't use log4j's API directly, only slf4j

pasha-kuznetsov: It's pretty tricky here, as `slf4j` doesn't have appenders, and the test is done via capturing the appended `LoggingEvents`, at the `log4j` level...

pasha-kuznetsov: @dlg99: pushed an update bfe22f3 addressing this as described above

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDescriptorImpl.java
pasha-kuznetsov: @dlg99: see a1aea61

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/BKException.java
```diff
@@ -52,10 +67,33 @@ public final int getCode() {
...
+
+    /**
+     * Creates a lazy error code formatter suitable to pass to log functions.
+     *
+     * @param code the error code value
+     *
+     * @return lazy error code log formatter
+     */
+    public static Object codeLogger(int code) {
+        return new CodeLogFormatter(code);
```
eolivelli: As the number of rc is limited can't we have a pool and save memory allocations? this code may be executed inside the write path or read path

pasha-kuznetsov: Sure! Please see d142c71

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/BKException.java
```diff
@@ -240,4 +255,60 @@ public static String getMessage(int code) {
         int UnexpectedConditionException = -999;
     }
 
+    /**
+     * Code log message pool.
+     */
+    private static class LogMessagePool {
+        private static final int lastKnown = 200;
```
jiazhai: Currently it seems OK. It would be better to have this connected with size of Code above, and not hard coded.

pasha-kuznetsov: @jiazhai sure, please see 7676dab

eolivelli:
+1 still very good

Issue #280, opened 2017-07-24
Turn bookie into readonly when SortedLedgerStorage fails to flush data


sijie:
**FEATURE REQUEST**

1. Please describe the feature you are requesting.

Address the TODO item in `SortedLedgerStorage`

    scheduler.execute(new Runnable() {
            @Override
            public void run() {
                try {
                    LOG.info("Started flushing mem table.");
                    long logIdBeforeFlush = entryLogger.getCurrentLogId();
                    memTable.flush(SortedLedgerStorage.this);
                    long logIdAfterFlush = entryLogger.getCurrentLogId();
                    // in any case that an entry log reaches the limit, we roll the log and start checkpointing.
                    // if a memory table is flushed spanning over two entry log files, we also roll log. this is
                    // for performance consideration: since we don't wanna checkpoint a new log file that ledger
                    // storage is writing to.
                    if (entryLogger.reachEntryLogLimit(0) || logIdAfterFlush != logIdBeforeFlush) {
                        entryLogger.rollLog();
                        LOG.info("Rolling entry logger since it reached size limitation");
                    }
                } catch (IOException e) {
                    // TODO: if we failed to flush data, we should switch the bookie back to readonly mode
                    //       or shutdown it.
                    LOG.error("Exception thrown while flushing skip list cache.", e);
                }
            }
        });

2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have). Are you currently using any workarounds to address this issue?

*nice-to-have*

3. Provide any additional detail on your proposed use case for this feature.

N/A

sijie:
This has been implemented by #873

PR #873, opened 2017-12-18
Issue280: Add StateManager to manage Bookie's state, and use it to turn bookie into readonly when sortedLedgerStorage failed to flush data

ArvinDevel:

Add SortedLedgerStorageListener to implement, and add a test case to verify it.

Master Issue: #280 

> ---
> quickly and easily:
>
>
>
> Otherwise:
> 
> 
> ---

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
@@ -72,7 +72,6 @@ import org.apache.bookkeeper.bookie.BookieException.UnknownBookieIdException;
 import org.apache.bookkeeper.bookie.Journal.JournalScanner;
 import org.apache.bookkeeper.bookie.LedgerDirsManager.LedgerDirsListener;
 import org.apache.bookkeeper.bookie.LedgerDirsManager.NoWritableLedgerDirException;
-import org.apache.bookkeeper.bookie.SortedLedgerStorage.SortedLedgerStorageListener;
 import org.apache.bookkeeper.common.util.Watcher;
 import org.apache.bookkeeper.conf.ServerConfiguration;
 import org.apache.bookkeeper.discover.RegistrationManager;
@@ -139,7 +138,7 @@ public class Bookie extends BookieCriticalThread {
     // Flag identify whether it is in shutting down progress
     private volatile boolean shuttingdown = false;
     // Bookie status
-    private final BookieStatus bookieStatus = new BookieStatus();
+    final BookieStatus bookieStatus = new BookieStatus();
 
     private int exitCode = ExitCode.OK;
 
@@ -149,9 +148,7 @@ public class Bookie extends BookieCriticalThread {
 
     private final AtomicBoolean rmRegistered = new AtomicBoolean(false);
     protected final AtomicBoolean forceReadOnly = new AtomicBoolean(false);
-    // executor to manage the state changes for a bookie.
-    final ExecutorService stateService = Executors.newSingleThreadExecutor(
-            new ThreadFactoryBuilder().setNameFormat("BookieStateService-%d").build());
+    StateManager stateManager;
 
     // Expose Stats
     private final StatsLogger statsLogger;
@@ -668,6 +665,7 @@ public class Bookie extends BookieCriticalThread {
         } catch (KeeperException e) {
             throw new MetadataStoreException("Failed to initialize ledger manager", e);
         }
+        stateManager = new BookieStateManager(bookieStatus, registrationManager);
 
         // Initialise ledgerDirMonitor. This would look through all the
         // configured directories. When disk errors or all the ledger
@@ -680,7 +678,7 @@ public class Bookie extends BookieCriticalThread {
             if (!conf.isReadOnlyModeEnabled()) {
                 throw nle;
             } else {
-                this.transitionToReadOnlyMode();
+                this.stateManager.transitionToReadOnlyMode();
             }
         }
 
@@ -695,7 +693,7 @@ public class Bookie extends BookieCriticalThread {
                 if (!conf.isReadOnlyModeEnabled()) {
                     throw nle;
                 } else {
-                    this.transitionToReadOnlyMode();
+                    this.stateManager.transitionToReadOnlyMode();
                 }
             }
         }
@@ -723,6 +721,7 @@ public class Bookie extends BookieCriticalThread {
             ledgerManager,
             ledgerDirsManager,
             indexDirsManager,
+                stateManager,
             checkpointSource,
             syncThread,
             statsLogger);
@@ -878,11 +877,6 @@ public class Bookie extends BookieCriticalThread {
 
         ledgerStorage.start();
 
-        // if ledgerStorage is SortedLedgerStorage, then add SortedLedgerStorageListener
-        if (ledgerStorage instanceof SortedLedgerStorage){
-            ((SortedLedgerStorage) ledgerStorage).addSortedLedgerStorageListener(getSortedLedgerStorageListener());
-        }
-
         // check the bookie status to start with
         if (forceReadOnly.get()) {
             this.bookieStatus.setToReadOnlyMode();
@@ -895,7 +889,7 @@ public class Bookie extends BookieCriticalThread {
         // if setting it in bookie thread, the watcher might run before bookie thread.
         running = true;
         try {
-            registerBookie(true).get();
+            stateManager.registerBookie(true).get();
         } catch (Exception e) {
             LOG.error("Couldn't register bookie with zookeeper, shutting down : ", e);
             shutdown(ExitCode.ZK_REG_FAIL);
@@ -928,7 +922,7 @@ public class Bookie extends BookieCriticalThread {
             @Override
             public void allDisksFull() {
                 // Transition to readOnly mode on all disks full
-                transitionToReadOnlyMode();
+                stateManager.transitionToReadOnlyMode();
             }
 
             @Override
@@ -940,23 +934,13 @@ public class Bookie extends BookieCriticalThread {
             @Override
             public void diskWritable(File disk) {
                 // Transition to writable mode when a disk becomes writable again.
-                transitionToWritableMode();
+                stateManager.transitionToWritableMode();
             }
 
             @Override
             public void diskJustWritable(File disk) {
                 // Transition to writable mode when a disk becomes writable again.
-                transitionToWritableMode();
-            }
-        };
-    }
-
-    private SortedLedgerStorageListener getSortedLedgerStorageListener(){
-        return new SortedLedgerStorageListener() {
-            @Override
-            public void flushFailed() {
-                // turn the bookie to readonly mode when SortedLedgerStroage flush fails.
-                transitionToReadOnlyMode();
+                stateManager.transitionToWritableMode();
             }
         };
     }
@@ -977,150 +961,8 @@ public class Bookie extends BookieCriticalThread {
         return manager.initialize(conf, () -> {
             rmRegistered.set(false);
             // schedule a re-register operation
-            registerBookie(false);
+            stateManager.registerBookie(false);
         }, statsLogger);
-    }
-
-    /**
-     * Register as an available bookie.
-     */
-    protected Future<Void> registerBookie(final boolean throwException) {
-        return stateService.submit(new Callable<Void>() {
-            @Override
-            public Void call() throws IOException {
-                try {
-                    doRegisterBookie();
-                } catch (IOException ioe) {
-                    if (throwException) {
-                        throw ioe;
-                    } else {
-                        LOG.error("Couldn't register bookie with zookeeper, shutting down : ", ioe);
-                        triggerBookieShutdown(ExitCode.ZK_REG_FAIL);
-                    }
-                }
-                return (Void) null;
-            }
-        });
-    }
-
-    protected void doRegisterBookie() throws IOException {
-        doRegisterBookie(forceReadOnly.get() || bookieStatus.isInReadOnlyMode());
-    }
-
-    private void doRegisterBookie(boolean isReadOnly) throws IOException {
-        if (null == registrationManager || ((ZKRegistrationManager) this.registrationManager).getZk() == null) {
-            // registration manager is null, means not register itself to zk.
-            // ZooKeeper is null existing only for testing.
-            LOG.info("null zk while do register");
-            return;
-        }
-
-        rmRegistered.set(false);
-        try {
-            registrationManager.registerBookie(bookieId, isReadOnly);
-            rmRegistered.set(true);
-        } catch (BookieException e) {
-            throw new IOException(e);
-        }
-    }
-
-
-    /**
-     * Transition the bookie from readOnly mode to writable.
-     */
-    private Future<Void> transitionToWritableMode() {
-        return stateService.submit(new Callable<Void>() {
-            @Override
-            public Void call() throws Exception {
-                doTransitionToWritableMode();
-                return null;
-            }
-        });
-    }
-
-    @VisibleForTesting
-    public void doTransitionToWritableMode() {
-        if (shuttingdown || forceReadOnly.get()) {
-            return;
-        }
-
-        if (!bookieStatus.setToWritableMode()) {
-            // do nothing if already in writable mode
-            return;
-        }
-        LOG.info("Transitioning Bookie to Writable mode and will serve read/write requests.");
-        if (conf.isPersistBookieStatusEnabled()) {
-            bookieStatus.writeToDirectories(ledgerDirsManager.getAllLedgerDirs());
-        }
-        // change zookeeper state only when using zookeeper
-        if (null == registrationManager) {
-            return;
-        }
-        try {
-            doRegisterBookie(false);
-        } catch (IOException e) {
-            LOG.warn("Error in transitioning back to writable mode : ", e);
-            transitionToReadOnlyMode();
-            return;
-        }
-        // clear the readonly state
-        try {
-            registrationManager.unregisterBookie(bookieId, true);
-        } catch (BookieException e) {
-            // if we failed when deleting the readonly flag in zookeeper, it is OK since client would
-            // already see the bookie in writable list. so just log the exception
-            LOG.warn("Failed to delete bookie readonly state in zookeeper : ", e);
-            return;
-        }
-    }
-
-    /**
-     * Transition the bookie to readOnly mode.
-     */
-    private Future<Void> transitionToReadOnlyMode() {
-        return stateService.submit(new Callable<Void>() {
-            @Override
-            public Void call() {
-                doTransitionToReadOnlyMode();
-                return (Void) null;
-            }
-        });
-    }
-
-    @VisibleForTesting
-    public void doTransitionToReadOnlyMode() {
-        if (shuttingdown) {
-            return;
-        }
-        if (!bookieStatus.setToReadOnlyMode()) {
-            return;
-        }
-        if (!conf.isReadOnlyModeEnabled()) {
-            LOG.warn("ReadOnly mode is not enabled. "
-                    + "Can be enabled by configuring "
-                    + "'readOnlyModeEnabled=true' in configuration."
-                    + "Shutting down bookie");
-            triggerBookieShutdown(ExitCode.BOOKIE_EXCEPTION);
-            return;
-        }
-        LOG.info("Transitioning Bookie to ReadOnly mode,"
-                + " and will serve only read requests from clients!");
-        // persist the bookie status if we enable this
-        if (conf.isPersistBookieStatusEnabled()) {
-            this.bookieStatus.writeToDirectories(ledgerDirsManager.getAllLedgerDirs());
-        }
-        // change zookeeper state only when using zookeeper
-        if (null == registrationManager) {
-            return;
-        }
-        try {
-            registrationManager.registerBookie(bookieId, true);
-        } catch (BookieException e) {
-            LOG.error("Error in transition to ReadOnly Mode."
-                    + " Shutting down", e);
-            triggerBookieShutdown(ExitCode.BOOKIE_EXCEPTION);
-            return;
-        }
     }
 
     /*
@@ -1235,8 +1077,8 @@ public class Bookie extends BookieCriticalThread {
                     idxMonitor.shutdown();
                 }
 
-                // Shutdown the state service
-                stateService.shutdown();
+                // close the stateManager
+                stateManager.close();
             }
             // Shutdown the ZK client
             if (registrationManager != null) {
@@ -1322,7 +1164,7 @@ public class Bookie extends BookieCriticalThread {
             }
             success = true;
         } catch (NoWritableLedgerDirException e) {
-            transitionToReadOnlyMode();
+            stateManager.transitionToReadOnlyMode();
             throw new IOException(e);
         } finally {
             long elapsedNanos = MathUtils.elapsedNanos(requestNanos);
@@ -1347,7 +1189,7 @@ public class Bookie extends BookieCriticalThread {
                 handle.setExplicitLac(entry);
             }
         } catch (NoWritableLedgerDirException e) {
-            transitionToReadOnlyMode();
+            stateManager.transitionToReadOnlyMode();
             throw new IOException(e);
         }
     }
@@ -1382,7 +1224,7 @@ public class Bookie extends BookieCriticalThread {
             }
             success = true;
         } catch (NoWritableLedgerDirException e) {
-            transitionToReadOnlyMode();
+            stateManager.transitionToReadOnlyMode();
             throw new IOException(e);
         } finally {
             long elapsedNanos = MathUtils.elapsedNanos(requestNanos);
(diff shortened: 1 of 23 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -862,6 +863,11 @@ public synchronized void start() {
 
         ledgerStorage.start();
 
+        // if ledgerStorage is SortedLedgerStorage, then add SortedLedgerStorageListener
+        if (ledgerStorage instanceof SortedLedgerStorage){
+            ((SortedLedgerStorage) ledgerStorage).addSortedLedgerStorageListener(getSortedLedgerStorageListener());
```
eolivelli: This is not very good, it is bad practice to use instanceof in this case. We should have a common interface

sijie: @ArvinDevel any reason you can't make this a general storage listener?

ArvinDevel: Because flush failure is only relative to SortedLedgerStorage, so I just changed the SortedLedgerStorage to avoid much changes. Do you think that adding a common listener interface to LedgerStorage is ok? @sijie @eolivelli

sijie: @ArvinDevel an alternative:

- make a StateManager interface, move the state management in Bookie into an implementation of `StateManager`. The state manager is taking care of switching states of a bookie.
- pass the StateManager into the ledger storage. different ledger storage implementations use this state manager to decide when to switch the state.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieInitializationTest.java
```diff
@@ -781,7 +786,7 @@ public void testPersistBookieStatus() throws Exception {
         bookie = bookieServer.getBookie();
         assertTrue(bookie.isReadOnly());
         // transition to writable mode
-        bookie.doTransitionToWritableMode();
+        ((Bookie.BookieStateManager) bookie.getStateManager()).doTransitionToWritableMode();
```
eolivelli: why do we need this kind of cast ?

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/CompactionTest.java
```diff
@@ -636,7 +636,8 @@ public void checkpointComplete(CheckpointSource.Checkpoint checkpoint, boolean c
             manager,
             dirs,
             dirs,
-            checkpointSource,
+                new SimpleStateManger(),
```
eolivelli: typo: SimpleStateManger -> SimpleStateManager

ArvinDevel:
Current StateManager has not include the shutdown action of bookie, I think that a more specific PR is more better to refactor the StateManager.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieInitializationTest.java
eolivelli: What about making "transitionToWritableMode" return a Future<Void> and than call 
bookie.getStateManager()).transitionToWritableMode().get()
so you won't need to access BookieStateManager ?
//cc @sijie  @ivankelly

ArvinDevel: @eolivelli The `doTransitionToWritableMode()` executed the transition directly and `transitionToWritableMode` will use one `ExecutorService` to run it, I'm trying to keep the original behavior.

eolivelli: @ArvinDevel my suggestion will preserve the same behavior without the need of these casts.
Anyway we are in tests some some shortcut is allowed.
I am fine with keeping the casts

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/CompactionTest.java
ArvinDevel: fixed

eolivelli:
There are spotbugs errors, please fix them

ArvinDevel:
the spotbugs is about Inconsistent synchronization which was caused by accessing registrationManager unsynchronized from unit test, I have fixed it by adding synchronized keyword.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -147,9 +147,7 @@
 
     private final AtomicBoolean rmRegistered = new AtomicBoolean(false);
     protected final AtomicBoolean forceReadOnly = new AtomicBoolean(false);
-    // executor to manage the state changes for a bookie.
-    final ExecutorService stateService = Executors.newSingleThreadExecutor(
-            new ThreadFactoryBuilder().setNameFormat("BookieStateService-%d").build());
+    StateManager stateManager;
```
sijie: make this a private final field?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -245,12 +243,12 @@ public boolean accept(File dir, String name) {
     }
 
     @VisibleForTesting
-    public void setRegistrationManager(RegistrationManager rm) {
-        this.registrationManager = rm;
+    public synchronized void setRegistrationManager(RegistrationManager rm) {
```
sijie: I don't think you need `synchronized` if you make `stateManager` final

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -655,6 +653,7 @@ public Bookie(ServerConfiguration conf, StatsLogger statsLogger)
         } catch (KeeperException e) {
             throw new MetadataStoreException("Failed to initialize ledger manager", e);
         }
+        stateManager = new BookieStateManager(bookieStatus, registrationManager);
```
sijie: moved this to constructor if you make this field final

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -1446,6 +1304,153 @@ public LedgerStorage getLedgerStorage() {
         return ledgerStorage;
     }
 
+    /**
+     * A implementation of StateManager.
+     */
+    public class BookieStateManager implements StateManager{
```
sijie: - I think you can move this class out of Bookie. The current `Bookie` class is too huge. The tricky part is StateManager need to trigger bookie shutdown if it can't register to registration service. What you can do is adding a shutdown handler, bookie register a shutdown handler to state manager. if state manager can't not register, then it will trigger shutdown. An example code:

```

interface StateManager {

    interface ShutdownHandler {
         void shutdown();
    }

    void setShutdownHandler(ShutdownHandler handler);

}

```

- remove `public` to make it a package protected class
- need space before '{'

Once you moved this class out of `Bookie` class, you can write a unit test for this class. you can mock RegistrationManager. so we will have a good code coverage of testing bookie state transitions.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -1446,6 +1304,153 @@ public LedgerStorage getLedgerStorage() {
...
+        // use an executor to execute the state changes task
+        final ExecutorService stateService = Executors.newSingleThreadExecutor(
+                new ThreadFactoryBuilder().setNameFormat("BookieStateManagerService-%d").build());
+        final BookieStatus bookieStatus;
+        RegistrationManager registrationManager;
```
sijie: make this final?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -1446,6 +1304,153 @@ public LedgerStorage getLedgerStorage() {
...
+                        + " Shutting down", e);
+                triggerBookieShutdown(ExitCode.BOOKIE_EXCEPTION);
+                return;
+            }
+        }
+
+    }
+
+    @VisibleForTesting
+    public StateManager getStateManager() {
```
sijie: do you need to make it public?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -1446,6 +1304,153 @@ public LedgerStorage getLedgerStorage() {
...
+        return this.stateManager;
+    }
```
sijie: add a blank line after '}'

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/ReadOnlyBookie.java
```diff
@@ -42,6 +41,20 @@
     public ReadOnlyBookie(ServerConfiguration conf, StatsLogger statsLogger)
             throws IOException, KeeperException, InterruptedException, BookieException {
         super(conf, statsLogger);
+        stateManager = new BookieStateManager(bookieStatus, registrationManager){
```
sijie: space before '{'

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
sijie: you can return BookieStateManager here? so you don't need to cast in the test cases.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieInitializationTest.java
sijie: @ArvinDevel if you make `Bookie#getStateManager()` return `BookieStateManager`, you don't need to cast here.

bookkeeper-server/src/test/java/org/apache/bookkeeper/test/BookKeeperClusterTestCase.java
```diff
@@ -723,4 +724,31 @@ public Auditor getAuditor(int timeout, TimeUnit unit) throws Exception {
...
+    public static class SimpleStateManager implements StateManager {
+
+        @Override
+        public void close() {
+
+        }
+
+        @Override
+        public Future<Void> registerBookie(boolean throwException) {
+            return null;
```
sijie: `return Future.Void()`?

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/CompactionTest.java
```diff
@@ -248,6 +247,7 @@ public void checkpointComplete(Checkpoint checkPoint, boolean compact)
             LedgerManagerFactory.newLedgerManagerFactory(conf, zkc).newLedgerManager(),
             dirManager,
             dirManager,
+                new SimpleStateManager(),
```
sijie: wrong indents

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieInitializationTest.java
```diff
@@ -108,8 +109,14 @@ public void tearDown() throws Exception {
         }
 
         void testRegisterBookie(ServerConfiguration conf) throws IOException {
-            super.doRegisterBookie();
+            ((BookieStateManager) super.getStateManager()).doRegisterBookie();
         }
+
+        void updateRegistration(RegistrationManager newRm){
```
sijie: change to setRegistrationManager

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -147,9 +147,7 @@
 
     private final AtomicBoolean rmRegistered = new AtomicBoolean(false);
```
sijie: I think `rmRegistered` and `forceReadOnly` should be part of state manager as well.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -137,7 +137,7 @@
     // Flag identify whether it is in shutting down progress
     private volatile boolean shuttingdown = false;
     // Bookie status
-    private final BookieStatus bookieStatus = new BookieStatus();
+    final BookieStatus bookieStatus = new BookieStatus();
```
sijie: I think `bookieStatus` should also belong to state manager.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
ArvinDevel: because tests in `org.apache.bookkeeper.test` may use stateManager, for example ReadOnlyBookieTest, so `getStateManager ` should use public

ArvinDevel:
In this version, what a little unreasonable is that BookieId was moved from Bookie to BookieStateManager to reduce constructing complex, but bookieId is indeed an property of Bookie.

ArvinDevel:
Inconsistent synchronization still exists, so add `synchronized` to avoid it.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/LedgerCacheTest.java
```diff
@@ -552,6 +555,53 @@ public void testEntryMemTableFlushFailure() throws Exception {
...
+        bookie.addEntry(generateEntry(1, 1), new Bookie.NopWriteCallback(), null, "passwd".getBytes());
+        flushTestSortedLedgerStorage.addEntry(generateEntry(1, 2));
+        assertFalse("Bookie is expected to be in ReadWrite mode", bookie.isReadOnly());
+        assertTrue("EntryMemTable SnapShot is expected to be empty", memTable.snapshot.isEmpty());
+
+        // set flags, so that FlushTestSortedLedgerStorage simulates FlushFailure scenario
+        flushTestSortedLedgerStorage.setInjectMemTableSizeLimitReached(true);
+        flushTestSortedLedgerStorage.setInjectFlushException(true);
+        flushTestSortedLedgerStorage.addEntry(generateEntry(1, 2));
+        Thread.sleep(1000);
```
sijie: do we need to `sleep(1000)` here? can we make this test sequence more deterministic rather than depending on time, to reduce flakiness？

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/StateManagerTest.java
```diff
@@ -0,0 +1,207 @@
...
+
+    private static class MockZKRegistrationManager extends ZKRegistrationManager {
+        boolean registerFailed = false;
+        void setRegisterFail(boolean failOrNot){
+            registerFailed = failOrNot;
+        }
+        @Override
+        public void registerBookie(String bookieId, boolean readOnly) throws BookieException {
+            if (registerFailed) {
+                throw BookieException.create(-100);
```
sijie: can you use a meaning BookieException code constant rather than a number `-100`?

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/StateManagerTest.java
```diff
@@ -0,0 +1,207 @@
...
+            .setZkServers(zkUtil.getZooKeeperConnectString());
+
+        BookieServer bkServer = new BookieServer(conf) {
+            protected Bookie newBookie(ServerConfiguration conf)
+                    throws IOException, KeeperException, InterruptedException,
+                    BookieException {
+                Bookie bookie = new Bookie(conf);
+                rm.setRegisterFail(true);
+                rm.initialize(conf, () -> {}, NullStatsLogger.INSTANCE);
+                LOG.info(" is {} ", zkc == null);
```
sijie: it seems this logging statement is missing some info?

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/StateManagerTest.java
```diff
@@ -0,0 +1,207 @@
...
+     */
+    @Test
+    public void testBookieTransitions() throws Exception {
+        File tmpDir = createTempDir("stateManger", "test");
+
+        final ServerConfiguration conf = TestBKConfiguration.newServerConfiguration();
+        conf.setJournalDirName(tmpDir.getPath())
+                .setLedgerDirNames(new String[] { tmpDir.getPath() })
+                .setZkServers(zkUtil.getZooKeeperConnectString());
+        Bookie bookie = new Bookie(conf);
```
sijie: I don't think you need to start a bookie. can you just construct a BookieStateManager and use it for the remaining test?

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/StateManagerTest.java
```diff
@@ -0,0 +1,207 @@
...
+        bookie.start();
+        assertTrue(bookie.isRunning());
+
+        bookie.getStateManager().transitionToReadOnlyMode();
+        //sleep a little to wait transition finish
+        Thread.sleep(1000);
```
sijie: This is not a good idea. It is not good to use `sleep(1000)`. Can you wait on the future returned by transitionToReadOnlyMode?

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/StateManagerTest.java
```diff
@@ -0,0 +1,207 @@
...
+        assertTrue(bookie.isRunning());
+        assertTrue(bookie.isReadOnly());
+
+        bookie.getStateManager().transitionToWritableMode();
+        Thread.sleep(1000);
```
sijie: same as above

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/StateManagerTest.java
```diff
@@ -0,0 +1,207 @@
...
+
+        // readOnly disabled bk
+        conf.setReadOnlyModeEnabled(false);
+        Bookie bookie2 = new Bookie(conf);
+        bookie2.start();
+        assertTrue(bookie2.isRunning());
+
+        bookie2.getStateManager().transitionToReadOnlyMode();
+        //sleep a little to wait transition finish
+        Thread.sleep(1000);
```
sijie: same as above

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/StateManagerTest.java
```diff
@@ -0,0 +1,207 @@
...
+                .setZkServers(zkUtil.getZooKeeperConnectString())
+                .setForceReadOnlyBookie(true);
+        ReadOnlyBookie readOnlyBookie = new ReadOnlyBookie(readOnlyConf, NullStatsLogger.INSTANCE);
+        readOnlyBookie.start();
+        assertTrue(readOnlyBookie.isRunning());
+        assertTrue(readOnlyBookie.isReadOnly());
+
+        // transition has no effect if bookie start with readOnly mode
+        readOnlyBookie.getStateManager().transitionToWritableMode();
+        Thread.sleep(1000);
```
sijie: same as above

sijie:
@eolivelli since you reviewed this pr before, can you review the latest changes from @ArvinDevel ? Let move this change forward.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -132,24 +127,12 @@
...
     private final ConcurrentLongHashMap<byte[]> masterKeyCache = new ConcurrentLongHashMap<>();
 
-    protected final String bookieId;
-
-    private final AtomicBoolean rmRegistered = new AtomicBoolean(false);
-    protected final AtomicBoolean forceReadOnly = new AtomicBoolean(false);
-    // executor to manage the state changes for a bookie.
-    final ExecutorService stateService = Executors.newSingleThreadExecutor(
-            new ThreadFactoryBuilder().setNameFormat("BookieStateService-%d").build());
+    StateManager stateManager;
```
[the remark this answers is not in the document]
ArvinDevel: ReadOnlyBookie will override stateManager, so private is too strict

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStateManager.java
```diff
@@ -0,0 +1,271 @@
...
+import org.apache.bookkeeper.conf.ServerConfiguration;
+import org.apache.bookkeeper.discover.RegistrationManager;
+import org.apache.bookkeeper.discover.ZKRegistrationManager;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A implementation of StateManager.
+ */
+public class BookieStateManager implements StateManager {
```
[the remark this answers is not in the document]
ArvinDevel: BookKeeperClusterTestCase will use it in `setBookieToReadOnly`, which is in test package, so package private is not enough.

ArvinDevel:
@eolivelli @sijie have fixed the comments

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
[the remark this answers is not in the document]
sijie: okay make it protected if there are classes extended this class will use this variable.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStateManager.java
```diff
@@ -61,15 +67,47 @@
...
+                }
+            }
+        });
+    }
+
+    @VisibleForTesting
+    BookieStateManager(ServerConfiguration conf, RegistrationManager registrationManager) throws IOException {
+        this.conf = conf;
+        this.registrationManager = registrationManager;
+        this.statsLogger = null;
```
sijie: use NullStatsLogger.INSTANCE rather than plain `null`

Also I think this constructor use this other constructor

```
this(conf, NullStatsLogger.INSTANCE, registrationManager, new LedgerDirsManager(...));
```

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStateManager.java
```diff
@@ -128,9 +149,14 @@ public boolean isRunning(){
     public boolean isShuttingDown(){
         return shuttingdown;
     }
+    @VisibleForTesting
```
sijie: blank line before `@VisibleForTesting`

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/StateManager.java
```diff
@@ -0,0 +1,107 @@
...
+ */
+
+package org.apache.bookkeeper.bookie;
+
+import java.util.concurrent.Future;
+
+/**
+ * State management of Bookie, including register, turn bookie to w/r mode.
+ */
+public interface StateManager extends AutoCloseable {
```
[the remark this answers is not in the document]
ArvinDevel: LedgerStorage will use it, so it should be public

sijie:
@eolivelli can you review @ArvinDevel 's latest change?

sijie:
ping @eolivelli

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/CompactionTest.java
```diff
@@ -67,6 +64,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
```
eolivelli: Nit: remove new line

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/StateManagerTest.java
```diff
@@ -0,0 +1,258 @@
...
+        stateManager.initState();
+        // up
+        assertTrue(stateManager.isRunning());
+        // unregistered
+        assertFalse(stateManager.isRegistered());
+
+        stateManager.registerBookie(true).get();
+        // registered
+        assertTrue(stateManager.isRegistered());
+        stateManager.getShutdownHandler().shutdown(ExitCode.OK);
```
eolivelli: This looks strange to me.
You are calling directly the callback, as far as I can see from this test you can simply run the code in the callback.
If you expect the callback to be called more than once please add some logic to test it, like an AtomicInteger counter

ArvinDevel: I used apis the stateManager exposes to outside, where it will call the actual action methods. And the  returned Future<Void> of public api methods is just an indicator for progress, where I can't pass callback to it. I thought the test cover the logic of the class. Sorry for that I couldn't understand your concern.

sijie:
> Will wait for seeing Sijie's comments addressed

@eolivelli which comments?

eolivelli:
It seemed to me that there were some of your comments left unadressed. Like the one about *synchronized*

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/StateManagerTest.java
eolivelli: I can't understand why you have to set the shutdown handler. The code you are passing is not called by the bookie but by the test code itself.
It seems that is testing the shutdown handler but it is not doing it actually

eolivelli: I can't understand why you have to set the shutdown handler. The code you are passing is not called by the bookie but by the test code itself.
It seems that is testing the shutdown handler but it is not doing it actually

ArvinDevel: yeah, you're right. The logic of the shutdownHandler is same with the bookie. After we extract the bookie from the tests, I think we can only test the logic by this way although it seems a bit superfluous.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStateManager.java
```diff
@@ -0,0 +1,310 @@
...
+    private final ServerConfiguration conf;
+    private final LedgerDirsManager ledgerDirsManager;
+
+
+    // use an executor to execute the state changes task
+    final ExecutorService stateService = Executors.newSingleThreadExecutor(
+            new ThreadFactoryBuilder().setNameFormat("BookieStateManagerService-%d").build());
+
+    // Running flag
+    private volatile boolean running = false;
```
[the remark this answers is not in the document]
ArvinDevel: After I investigated the original code, I found that the state of a bookie has several dimensions, which means the states has overlap, such as the read/writable and running/shuttingdown, for example, a bookie which is shuttingdown will has the readOnly state. The legacy state expression is confusing, a more clear state is needed. So what potential impact of the new state should be investigated first.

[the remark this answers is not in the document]
sijie: I don't like the idea to couple logic refactoring with code refactoring. I would suggest not changing anything when doing an abstraction refactor. It is much clearer to reason about when we move code around. 

If there is a need to refactor the logic, this should be a separate PR in a separate discussion after code refactor. 

@ivankelly

.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SortedLedgerStorage.java
```diff
@@ -233,4 +236,10 @@ public void onRotateEntryLog() {
         // can happen because compaction. in a sorted ledger storage, checkpoint should happen after the data is
         // flushed to the entry log file.
     }
+
+    @VisibleForTesting
```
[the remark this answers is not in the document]
ArvinDevel: fixed

sijie:
@eolivelli 

> It seemed to me that there were some of your comments left unadressed. Like the one about synchronized

I think @ArvinDevel left a comment about `synchronized`. please check the comment.

ArvinDevel:
currently `synchronized` keyword must exists, it's more about a legacy from RegistrationManager, which was exposed to test without `final`, I think the spot check is imported after importing RegistrationManager, maybe we can refactor RegistrationManager to remove this un-graceful usage.

eolivelli:
retest this please

sijie:
@ivankelly can you review my last comment and let's move forward?

@eolivelli I think you mentioned the wrong person

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStateManager.java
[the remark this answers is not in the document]
ivankelly: @ArvinDevel 
>  a bookie which is shuttingdown will has the readOnly state

What is the readOnly state in this case, though? It's a flag to check whether you can write to the bookie, i.e. that it's not in the WRITABLE state, which is what we really should be checking.

@sijie 
> If there is a need to refactor the logic, this should be a separate PR in a separate discussion after code refactor.

Sure, this can go in as is and the states can be cleaned up later.

ivankelly:
@sijie ship it

sijie:
retest this please

Issue #865, opened 2017-12-15
LongHierarchicalLedgerManager: fix race with concurrent updates in initialize()


athanatos:
LongHierarchicalLedgerRangeIterator.initialize() can erroneously exit with iteratorDone if the lexicographically first path in zk has fewer than 4 levels.  This can happen for a few reasons including a case where a client creating a ledger on that path crashed during the zk updates or the iterator.hasNext() call simply raced with an in-progress node creation or removal.  ScanAndCompareGarbageCollector is hasNext() returns false will delete all ledgers on the bookie, so this is a fairly serious bug.

athanatos:
I've got a patch based on an earlier version I'm porting forward, should have a PR up pretty quickly.

PR #875, opened 2017-12-18
Issue 865: LongHierarchicalLedgerManager: refactor to fix races

athanatos:
LongHierarchicalLedgerRangeIterator.initialize() could erroneously exit
with iteratorDone if the lexicographically first path in zk had fewer
than 4 levels.  This can happen for a few reasons including a case where
a client creating a ledger on that path crashed during the zk updates or
the iterator.hasNext() call simply raced with an in-progress node
creation or removal.  ScanAndCompareGarbageCollector is hasNext()
returns false will delete all ledgers on the bookie, so this is a fairly
serious bug.

Ruling out other such bugs was fairly dificult with the structure of the
code as written, so instead use a simpler recursive iterator design with
simpler pre/post conditions.

Also, surface KeeperException.NoNodeException from
ZkUtils.getChildrenInSingleNode so that we can actually handle it in
LHLM.

Master Issue: #865


sijie:
retest this please

Issue #870, opened 2017-12-16
ScanAndCompareGarbageCollector.gc: harden against LedgerManager bugs


athanatos:
There is special logic in gc() for handling the case where ledgerRangeIterator.hasNext() returns false immediately (the case where the LedgerManager is empty).  The logic immediately deletes all of the indices which were present prior to the iterator call.  Strictly speaking, this is correct, but segregating the handling of this case from the handling of (for instance) the case where there is 1 ledger means that we unnecessarily duplicate the removal logic and, in this case, skip the part where we double check that the metadata really is absent.  Refactor this method to use the same logic in both cases.

More generally, add a config to force the bookie to use both the iterator and the readLedgerMetadata paths to prove a ledger is really dead.

I'll have a PR up for this in the next few days.

PR #876, opened 2017-12-18
Issue 870: ScanAndCompareGarbageCollector: harden against LM bugs

athanatos:
The idea behind this patch is to make it more likely that more than one
LedgerManager bug would be required to erroneously believe a ledger to be
deleted.
1) Remove the special handling for the completely empty LM.  It's not
really a common case nor is it much of an optimization even when it
happens.
2) Add a config variable to cause the bookie to also check the
LedgerManager.readLedgerMetadata path before actually removing a ledger.
The implementations of readLedgerMetadata and the iterator are
sufficiently different that it seems worth it to have the option of
checking both to guard somewhat against a bug in either.

(@bug W-4292747@)
(@bug W-3027938@)
Signed-off-by: Charan Reddy Guttapalem <cguttapalem@salesforce.com>

Master Issue: #870

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/ScanAndCompareGarbageCollector.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/ScanAndCompareGarbageCollector.java
@@ -28,8 +28,10 @@ import java.util.List;
 import java.util.NavigableSet;
 import java.util.Set;
 import java.util.SortedMap;
+import java.util.TreeSet;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Semaphore;
+import java.util.concurrent.atomic.AtomicInteger;
 import org.apache.bookkeeper.client.BKException;
 import org.apache.bookkeeper.client.LedgerMetadata;
 import org.apache.bookkeeper.conf.ServerConfiguration;
@@ -76,6 +78,7 @@ public class ScanAndCompareGarbageCollector implements GarbageCollector{
     private final long gcOverReplicatedLedgerIntervalMillis;
     private long lastOverReplicatedLedgerGcTimeMillis;
     private final String zkLedgersRootPath;
+    private final boolean verifyMetadataOnGc;
 
     public ScanAndCompareGarbageCollector(LedgerManager ledgerManager, CompactableLedgerStorage ledgerStorage,
             ServerConfiguration conf) throws IOException {
@@ -91,6 +94,7 @@ public class ScanAndCompareGarbageCollector implements GarbageCollector{
         this.zkLedgersRootPath = conf.getZkLedgersRootPath();
         LOG.info("Over Replicated Ledger Deletion : enabled=" + enableGcOverReplicatedLedger + ", interval="
                 + gcOverReplicatedLedgerIntervalMillis);
+        verifyMetadataOnGc = conf.getVerifyMetadataOnGC();
     }
 
     @Override
@@ -105,18 +109,6 @@ public class ScanAndCompareGarbageCollector implements GarbageCollector{
             // Get a set of all ledgers on the bookie
             NavigableSet<Long> bkActiveLedgers = Sets.newTreeSet(ledgerStorage.getActiveLedgersInRange(0,
                     Long.MAX_VALUE));
-
-            // Iterate over all the ledger on the metadata store
-            LedgerRangeIterator ledgerRangeIterator = ledgerManager.getLedgerRanges();
-
-            if (!ledgerRangeIterator.hasNext()) {
-                // Empty global active ledgers, need to remove all local active ledgers.
-                for (long ledgerId : bkActiveLedgers) {
-                    garbageCleaner.clean(ledgerId);
-                }
-            }
-
-            long lastEnd = -1;
 
             long curTime = MathUtils.now();
             boolean checkOverreplicatedLedgers = (enableGcOverReplicatedLedger && curTime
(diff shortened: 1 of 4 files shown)


athanatos:
@jvrao  You might want to glance at how I handled the double check.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/ScanAndCompareGarbageCollector.java
```diff
@@ -134,27 +124,50 @@ public void gc(GarbageCleaner garbageCleaner) {
...
                     if (!ledgersInMetadata.contains(bkLid)) {
+                        if (conf.getVerifyMetadataOnGC()) {
+                            CountDownLatch latch = new CountDownLatch(1);
+                            final AtomicInteger metaRC = new AtomicInteger(0);
+                            ledgerManager.readLedgerMetadata(bkLid, (int rc, LedgerMetadata x) -> {
+                                metaRC.set(rc);
+                                latch.countDown();
+                            });
+                            latch.await();
+                            if (metaRC.get() != BKException.Code.NoBookieAvailableException) {
```
sijie: wrong return code here? it should be NoLedgerExistsException?

bookkeeper-server/src/test/java/org/apache/bookkeeper/meta/GcLedgersTest.java
```diff
@@ -431,4 +657,59 @@ public boolean waitForLastAddConfirmedUpdate(long ledgerId,
             return false;
         }
     }
+
+    class LedgerManagerWrapper implements LedgerManager {
```
sijie: I think you can use `CleanupLedgerManager` for this. You don't need to recreate a similar class here.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/ScanAndCompareGarbageCollector.java
sijie: I am not sure how the below test cases work, if the return code is wrong.

bookkeeper-server/src/test/java/org/apache/bookkeeper/meta/GcLedgersTest.java
```diff
@@ -315,6 +319,228 @@ public void clean(long ledgerId) {
...
+     * false and next to be null. So even in this case it is expected not to clean any ledger's data.
+     *
+     * This testcase is needed for validating fix of bug - W-4292747.
+     *
+     * ScanAndCompareGarbageCollector/GC should clean data of ledger only if both the LedgerManager.getLedgerRanges says
+     * that ledger is not existing and also ledgerManager.readLedgerMetadata fails with error
+     * NoSuchLedgerExistsException.
+     *
+     */
+    @Test(timeout = 60000)
```
sijie: remove timeout value. currently we use a global timeout setting to control. If this test case doesn't require special timeout value, we can remove it here.

bookkeeper-server/src/test/java/org/apache/bookkeeper/meta/GcLedgersTest.java
```diff
@@ -315,6 +319,228 @@ public void clean(long ledgerId) {
         assertEquals("Should have cleaned first ledger" + first, (long) first, (long) cleaned.get(0));
     }
 
+    /*
+     * in this scenario no ledger is created, so ledgeriterator's hasNext call would return false and next would be
+     * null. GarbageCollector.gc is expected to behave normally
+     */
+    @Test(timeout = 60000)
```
eolivelli: drop timeout

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/ScanAndCompareGarbageCollector.java
athanatos: Yeah, sorry.  I needed to adapt that bit somewhat from our branch and evidently introduced a typo.  Let me figure out why the tests worked and correct it.

athanatos: Ah, it passed because there were no "normal" gc tests where either the iterator or the ledger manager weren't corrupted with that option enabled.  Enabling it for testGcLedgersWithLedgersInSameLedgerRange catches this bug.

bookkeeper-server/src/test/java/org/apache/bookkeeper/meta/GcLedgersTest.java
athanatos: ok

bookkeeper-server/src/test/java/org/apache/bookkeeper/meta/GcLedgersTest.java
athanatos: ok

athanatos:
@sijie Updated.

sijie:
jdk9 passed, jdk8 failed with TestBenchmark (but it is unrelated, known to be a flaky test)

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
```diff
@@ -309,6 +310,25 @@ public ServerConfiguration setUseTransactionalCompaction(boolean useTransactiona
...
+     * Get whether the bookie is configured to double check prior to gc.
+     *
+     * @return use transactional compaction
+     */
+    public boolean getVerifyMetadataOnGC() {
+        return this.getBoolean(VERIFY_METADATA_ON_GC, false);
+    }
+
+    /**
+     * Set whether to use transactional compaction and using a separate log for compaction or not.
```
ArvinDevel: the description is not correct yet

sijie: nice catch @ArvinDevel !

athanatos: Thanks, will update.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/ScanAndCompareGarbageCollector.java
```diff
@@ -134,27 +124,50 @@ public void gc(GarbageCleaner garbageCleaner) {
...
 
                 Iterable<Long> subBkActiveLedgers = bkActiveLedgers.subSet(start, true, end, true);
 
-                Set<Long> ledgersInMetadata = lRange.getLedgers();
                 if (LOG.isDebugEnabled()) {
                     LOG.debug("Active in metadata {}, Active in bookie {}", ledgersInMetadata, subBkActiveLedgers);
                 }
                 for (Long bkLid : subBkActiveLedgers) {
                     if (!ledgersInMetadata.contains(bkLid)) {
+                        if (conf.getVerifyMetadataOnGC()) {
```
reddycharan: nit: why does this config variable has to be read every single time..probably a variable would be sufficient

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/ScanAndCompareGarbageCollector.java
```diff
@@ -134,27 +124,50 @@ public void gc(GarbageCleaner garbageCleaner) {
...
+                            CountDownLatch latch = new CountDownLatch(1);
+                            final AtomicInteger metaRC = new AtomicInteger(0);
+                            ledgerManager.readLedgerMetadata(bkLid, (int rc, LedgerMetadata x) -> {
+                                metaRC.set(rc);
+                                latch.countDown();
+                            });
+                            latch.await();
+                            if (metaRC.get() != BKException.Code.NoSuchLedgerExistsException) {
+                                LOG.info(
```
reddycharan: nit: warn or error?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/ScanAndCompareGarbageCollector.java
athanatos: The price of the config read seems to me to be trivial compared to the cost of querying zk.  I'm inclined to keep it simple.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/ScanAndCompareGarbageCollector.java
athanatos: Yeah, good point, will fix.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/ScanAndCompareGarbageCollector.java
reddycharan: I'm just saying instead of reading config everytime that config value can be assigned to a final variable and use it here..anyhow that config value is not going to change just like rest of the configs

athanatos: Ok, I'll switch it.

ArvinDevel:
+1 LGTM

PR #880, opened 2017-12-19
Issue 879: TestTLS fails when project path contains spaces

sijie:

Use `URI` rather than `URL`, so it won't use url-encoded path as the file path.


sijie:
TestTLS passes now: https://builds.apache.org/job/bookkeeper_precommit_pullrequest/34/jdk=JDK%201.8%20(latest),label=ubuntu/console

PR #887, opened 2017-12-19
make replaceBookie() exclude ensemble bookie racks

sijie:

This is a change ported back from yahoo/bookkeeper@0589d0d58d61364be1a897a5c1478d3109f4fc17. It excludes the ensemble bookie racks when replacing bookies. since there is already a new weight-based placement strategy introduced, so this change only applies when `weight-based` placement policy is not enabled.

This is a change contributed from Siddharth Boobna.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
@@ -27,7 +27,6 @@ import io.netty.util.HashedWheelTimer;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.util.ArrayList;
-import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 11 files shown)


sijie:
@merlimat can you eyeball this change to make sure I ported this correctly? since this change is conflicting with weight-based placement policy. so this feature is only *ON* when weight-based placement policy is disabled.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
```diff
@@ -419,6 +418,14 @@ public void handleBookiesThatJoined(Set<BookieSocketAddress> joinedBookies) {
         return nodes;
     }
 
+    private Set<String> getNetworkLocations(Set<Node> bookieNodes) {
```
eolivelli: static ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
```diff
@@ -594,8 +608,57 @@ public BookieNode selectFromNetworkLocation(
...
+                                                   Predicate<BookieNode> predicate,
+                                                   Ensemble<BookieNode> ensemble)
+            throws BKNotEnoughBookiesException {
+        // first attempt to select one from local rack
+        try {
+            return selectRandomFromRack(networkLoc, excludeBookies, predicate, ensemble);
+        } catch (BKNotEnoughBookiesException e) {
+            if (isWeighted) {
+                // if weight based selection is enabled, randomly select one from the whole cluster
+                // based on weights and ignore the provided predicate.
```
eolivelli: "ignore the provided predicate" -> maybe the comment is wrong ?  it seems to me that the predicate is applied in selectRandom

sijie:
@eolivelli fixed your comments.

PR #888, opened 2017-12-19
Rework of binary distribution licenses

ivankelly:
All redistributed jars are mentioned in the LICENSE file. If the jar
is ASLv2, it is simply listed under the ASLv2 text. If these ASLv2
jars have their own NOTICE files, then the relevant portions are added
to our NOTICE file.

All jar under license other than ASLv2 get their own section in
LICENSE, which links to the full text of their license. Non-ASLv2
generally do not require an update to NOTICE.

This change also includes a script, check-binary-license, which checks
that the contents on the LICENSE and NOTICE actually matches what
bundled in the distribution tarball.


ivankelly:
I've only covered -all for now. I will do -server after.

Also, check-binary-license will fail for now, as rocksdb isn't accounted for in the LICENSE file, and the jsr305 artifact version changed since 4.6.0.

eolivelli:
@ivankelly this is a huge work, thank you very much

Did you take a look at maven license plugin ?
It has the ability to download all third party licences
http://www.mojohaus.org/license-maven-plugin/add-third-party-mojo.html

Maybe the very interesting part of the maven plugin is that it can FAIL the build is a non compliant license is discovered

Apart from this comment we need to review all the licenses manually as you did, it will take some time but I think it is worth

ivankelly:
@eolivelli that plugin looks interesting, but it doesn't do the notice files. Also, I guess it depends on the license stated in the pom, which can be wrong in some cases (jsr305 for example). Do you have any example of it being used with BSD style licenses (where the copyright notice needs to be preserved)?

sijie:
also it would be good to create an issue for this and track the discussion there.

ivankelly:
> I don't think it is a good idea to pull in all the license and notice from all the dependencies.

This isn't the licenses and notices from all the dependencies, only from the dependencies whose license requires it.

> It will make maintenance and dependency upgrade complicated,

It shouldn't. The check-binary-license script should flag if the license needs to be updated in CI. Updating the license is quick, I checked every dependency we have in one morning.

> the license files will quickly get out of sync

Again, the check-binary-license script should prevent this.

> I think we should look for better solutions than this.

I don't think there is a better solution than doing it manually. A script can't parse a NOTICE file to see what to bubble up.

> also it would be good to create an issue for this and track the discussion there.

Will create an issue.

ivankelly:
Issue is #891

sijie:
first, I think it is a good idea to have a check script to do the things. I am not against it.

my main concern is about the approach how we maintain the source-of-truth of notice files and how this script does the verification. basically you maintain some notice/license files as the source-of-truth, and have a script verifying the LICENSE file against the _maintained_ notice/license files. How do you guarantee the notice/license files can always be the correct? That introduces extra complexities than maintaining a LICENSE file. 

when I am saying "look for better solutions", I am not saying "doing it manually". There can always be better solutions. I am not sure how it is feasible, just thinking here, you can get artifactory (version, name) from pom files or assembled package, and fetch their licenses/notice from their website and verify the LICENSE against their notice.

At the minimal, I don't think we should be maintaining any LICENSE and NOTICE files from dependencies, and we shouldn't use our maintained versions as the source-of-truth for verification.

ivankelly:
> How do you guarantee the notice/license files can always be the correct?

We assume that, for a given version, the notice and license files will not change, which is a safe assumption to make.

I'm dubious as to whether it will be possible to pull in NOTICE files automatically. We would have to pull in every NOTICE file, which isn't really necessary. And then someone would have to check the contents of the pulled in NOTICE file to ensure everything is ok.

If we make the pulling of licenses automatic, then they will only ever be checked at release time. At release time, all dependencies need to be checked, and when there's so much to check, people are likely to just give it a quick glance, and +1 it, without actually checking each dependency.

I would prefer that the work in manually checking dependencies occurs as part of the development process, each time we update a dependency. At this time, there will be a smaller subset of the dependencies changing, so it can be reviewed more carefully. The submitter will be able to take their time with it, and the reviewer will be able to give each dependency their full attention. Once a license/notice has been updated for a version of the dependency, it shouldn't need to be looked at again (as licenses/notices don't change within a single version).

sijie:
@ivankelly so my question is if NOTICE is already included in LICENSE file, why do we need to maintain them twice in different places? can we just keep one copy in LICENSE file and the check script should use something else (e.g. fetching LICENSE and NOTICE from their sources) to verify and make sure LICENSE file is correct?

ivankelly:
Notice isn't included in the license file. Where are you seeing that?

sijie:
> Notice isn't included in the license file. Where are you seeing that?

sorry I mean the copyrights included in the license file.

ivankelly:
In the submitted change, there are no copyrights in the LICENSE file, though for all non-ASLv2 dependencies, we link to a LICENSE file which does include the copyright for those dependencies. In MIT and BSD the copyright notice is part of the license.

ASLv2 has to be handled differently. The copyright is part of the notice file, so it has to be bubbled up there. In a lot of cases, the dependency is another ASF project, so it's already covered by the copyright at the top of the project NOTICE file.

sijie:
@ivankelly I think it is a bit confusing here back and forth. I will just comment on the pull request.

bookkeeper-dist/src/assemble/bin-all.xml
```diff
@@ -50,6 +50,11 @@
         <include>${basedir}/*.txt</include>
       </includes>
     </fileSet>
+    <fileSet>
+      <directory>../src/main/resources/deps</directory>
```
sijie: we should not maintain the 3rdparty license. we should only attach the needed notice and license that is required into NOTICE or LICENSE file. maintaining this directory makes things complicated. we should avoid it.

The check script should parse pom files or the assemble tarballs to see what dependencies are included, and fetch their corresponding notices and verify if the NOTICE file includes all dependencies or not and if their licenses are matched and notices are attached.

bookkeeper-dist/src/main/resources/LICENSE-all.bin.txt
```diff
@@ -202,507 +202,170 @@
...
-DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
-THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
-(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
-OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
-
-Code generated by the Protocol Buffer compiler is owned by the owner
-of the input file used when generating it.  This code is not
-standalone and requires a support library to be linked with it.  This
-support library is itself covered by the above license.
+lib/io.netty-netty-3.10.1.Final.jar bundles some 3rd party dependencies.
```
sijie: if we are not use the 3rd party dependencies that netty bundled with, we shouldn't include their license and notices. because those dependencies might already be excluded.

bookkeeper-dist/src/main/resources/LICENSE-all.bin.txt
```diff
@@ -202,507 +202,170 @@
...
 
+For details see deps/netty-3.19.1.Final/
 ------------------------------------------------------------------------------------
-For lib/javax.servlet-javax.servlet-api-3.1.0.jar
+lib/io.netty-netty-all-4.1.12.Final.jar bundles some 3rd party dependencies
```
sijie: same as above

bookkeeper-dist/src/assemble/bin-all.xml
ivankelly: I've moved all the NOTICE stuff into the NOTICE, so that doesn't link anywhere.

The ASF licensing recommendations (http://www.apache.org/dev/licensing-howto.html#permissive-deps) actually say to bundle the license file, instead of putting directly in the LICENSE file unless the license is very short. I actually prefer it like this. 
https://github.com/ivankelly/bookkeeper/blob/license-rework/bookkeeper-dist/src/main/resources/LICENSE-all.bin.txt is easier to read than https://github.com/apache/bookkeeper/blob/master/bookkeeper-dist/src/main/resources/LICENSE-all.bin.txt. 

LICENSE-all.bin.txt would get huge if we flattened them all. I'm particularly eager to keep the CDDL out of it, that license is huge.

> The check script should parse pom files or the assemble tarballs to see what dependencies are included, and fetch their corresponding notices and 

We should check on the final output (i.e. the assemble tarball) as that is what we distribute. 

> verify if the NOTICE file includes all dependencies or not and if their licenses are matched and notices are attached.

This verification is very hard to do in an automated fashion. How will a machine know that the protobuf license contains stuff that isn't relevant? How will it work out which part of the netty NOTICE needs to be pulled in and which doesn't?

It's not hard for a human, but the human will need guidelines which we should put in the wiki.

I've added a check to the script to check if the bundled license files are linked, and ensure all linked files exist.

bookkeeper-dist/src/main/resources/LICENSE-all.bin.txt
ivankelly: Removed those that aren't relevant. There's a bunch of the notices which refer to code which netty has copied into their own source repo, so I've retained those in the LICENSE file (they're all BSD/MIT).

ivankelly:
> I think it is hard for me to comment at the pull request because this pull request includes so main texts, my browser barely move.

I've found chrome to suck at github. Moved to the new firefox a couple of weeks ago, and it works much better.

bookkeeper-dist/src/main/resources/LICENSE-all.bin.txt
sijie: can you point me links on how those licenses should handle "copied code"? 

My interpretation here is those "copied code" has already been compiled and redistributed as part of the jar, we should just retain netty's notice, but we don't need to their license, no?

bookkeeper-dist/src/main/resources/LICENSE-all.bin.txt
```diff
@@ -202,507 +202,265 @@
...
-
-Code generated by the Protocol Buffer compiler is owned by the owner
-of the input file used when generating it.  This code is not
-standalone and requires a support library to be linked with it.  This
-support library is itself covered by the above license.
+lib/io.netty-netty-3.10.1.Final.jar contains the extensions to Java Collections Framework which has
+been derived from the works by JSR-166 EG, Doug Lea, and Jason T. Greene:
+
+  * LICENSE:
+    * deps/netty-3.10.1.Final/LICENSE.jsr166y.txt (Public Domain)
```
sijie: I am not sure we need to retain these LICENSE files. I would remove them and just reference back to netty. 

Also I think netty is used by multiple ASF projects, it would be good if you can point me any ASF projects that do this. Otherwise it is going to be painful to maintain LICENSE of those "compiled code" in our dependencies.

bookkeeper-dist/src/main/resources/LICENSE-all.bin.txt
```diff
@@ -202,507 +202,265 @@
...
-
-10. RESPONSIBILITY FOR CLAIMS.
-
-   As between Initial Developer and the Contributors, each party is
-   responsible for claims and damages arising, directly or indirectly,
-   out of its utilization of rights under this License and You agree
-   to work with Initial Developer and Contributors to distribute such
-   responsibility on an equitable basis. Nothing herein is intended or
-   shall be deemed to constitute any admission of liability.
+lib/io.netty-netty-all-4.1.12.Final.jar contains a modified portion of and optionally depends on 'Protocol Buffers', Google's data
```
sijie: This is optionally dependency. why do we need this?

bookkeeper-dist/src/main/resources/LICENSE-all.bin.txt
```diff
@@ -202,507 +202,265 @@
...
+interchange format, which can be obtained at:
 
+  * LICENSE:
+    * deps/netty-4.1.12.Final/LICENSE.protobuf.txt (New BSD License)
+  * HOMEPAGE:
+    * https://github.com/google/protobuf
+
+------------------------------------------------------------------------------------
+This product bundles the JSR-305 reference implementation, which is available under
```
sijie: we should exclude this jar since it is compliation dependency not runtime dependency.

bookkeeper-dist/src/main/resources/LICENSE-all.bin.txt
ivankelly: "contains a modified portion of" implies that they are pulling code directly in.

bookkeeper-dist/src/main/resources/LICENSE-all.bin.txt
ivankelly: If we include the code, we should include the licenses. The public domain ones are not so important, but the BSD style ones need to be either here, or in a license file linked from here. Dependencies of dependencies should be treated the same as dependencies [1], which makes sense because if netty is distributing it, then we are distributing it.

[1] http://www.apache.org/dev/licensing-howto.html#deps-of-deps

bookkeeper-dist/src/main/resources/LICENSE-all.bin.txt
ivankelly: Are you sure? I was under the same impression, but when I looked at the source, the annotations have runtime retention. If we can get away with removing it, I'm all for that.

bookkeeper-dist/src/main/resources/LICENSE-all.bin.txt
ivankelly: Retaining the notice is enough if the code is ASLv2. With BSD/MIT, the notice is the license, so the whole thing has to be redistributed.

bookkeeper-dist/src/main/resources/LICENSE-all.bin.txt
sijie: if a dependency of dependency is bundled, it makes sense to include its license. otherwise it doesn't make sense to me. 

I am not sure about "code copied" into a netty and we bundle netty as a dependency. It is a very complicated case to me. I would suggest following what other ASF projects that use Netty do. I am pretty sure there are tons of projects depend on netty. 

If you really feel strong about using this approach you proposed here, I would suggest reaching legal-discuss@ to get a clarification on how to update LICENSE and NOTICE when including a bundle that contains copied-and-modified code. I am not qualified for reviewing if this is the right or not.

From my personal view, I would follow what other ASF projects do, which they are more or less reviewed and accepted by the ASF. The approach here suddenly pulls in a lot of licenses from dependencies of dependencies, sounds too complicated to me and it is going to be very hard for the community to maintain and keep them in sync.

bookkeeper-dist/src/main/resources/LICENSE-all.bin.txt
sijie: I think it follows into the above comment. We need to address the comment above if you feel strong on using this approach here.

This also a problem I can see here. We are unclear about what our dependencies do and what they changed. We should just include our dependencies' license and notice, we should not import their licenses, unless that is the dependency of dependency is explicitly included in our bundle.

Last, protobuf is already in the notice/license in our bundle. I am not sure why we still need a separate one.

bookkeeper-dist/src/main/resources/LICENSE-all.bin.txt
ivankelly: Code copied is actually the simplest case. You are distributing binaries derived from from copied code, and if it is BSD/MIT licensed, then you must include the license. Since I have the hadoop NOTICE open, I'll use that as the example:
https://github.com/apache/hadoop/blob/trunk/NOTICE.txt
Netty is there at the top (though the BSD/MIT stuff should be in LICENSE I believe).

I've emailed legal-discuss@, you're cc'd.

Yes, this is a pain, but it's less of a pain than reimplementing all out dependencies ourselves. And it's not that complex. I'll write it up if this goes in, and I don't expect it to be more than 2 or 3 paragraphs.

> sounds too complicated to me and it is going to be very hard for the community to maintain and keep them in sync.

The other option is stalled release processes if we leave it up to then to check licenses. That's what I'm trying to get away from with this.

bookkeeper-dist/src/main/resources/LICENSE-all.bin.txt
ivankelly: Again, this is code copied in, since it is modified. As it is copied in, we are distributing it, we should distribute the license for it.

bookkeeper-dist/src/main/resources/LICENSE-all.bin.txt
sijie: since you point out the hadoop's NOTICE file as the example, can you please tell me where does hadoop keep the those license files of netty's dependencies? (e.g. license/LICENSE.jsr166y.txt, license/LICENSE.base64.txt) I can't find those license files in hadoop repo or their binary package.

If I understand this correctly, hadoop only adds netty's notice file in its notice file, it doesn't copy the licenses of netty's dependencies into its repo and redistributed with its binary package. https://github.com/netty/netty/blob/4.1/NOTICE.txt The approach is exactly what I said here, we don't need to copy the licenses of netty's dependencies, we only need to attach the notices/licenses of the dependencies we bundled.

If you think hadoop is the right example to follow, let's follow what hadoop does (remove all the deps licenses, only include the notice/license of the bundles we distribute). If you think hadoop's notice/license also has problems, let's wait for responses from the legal team.

ivankelly: I do think hadoop has problems in this regard. The jsr166y and base64 being missing aren't a problem, since they're public domain, but the others BSD stuff should have the license distributed.

ivankelly: Let's wait to see what legal say

sijie: sure. let's see

sijie: Any updates from legal?

ivankelly: pinged the list again.

ivankelly: http://mail-archives.apache.org/mod_mbox/www-legal-discuss/201801.mbox/ajax/%3CCAJdLeK0euKouMx2h2%2BK6EC2-HJSL7XZnHcA7qvDtejN-D4k0QA%40mail.gmail.com%3E

ivankelly: @sijie legal's position on this is "if it's bundled, it needs the license"

sijie: @ivankelly I left one comment in the email thread, will wait for a clarification of my understanding there.

sijie:
Copied a few notes from the legal-discussion email thread there:

(Justin Mclean)**All bundle software licenses needs to be added to LICENSE [1] you can include ALv2 license software if you want. If an bundled ALv2 software includes a NOTICE file you need to look at that and move parts of it to your NOTICE file [3] but try to keep it as simple as needed and not add anything that s not required.**

>  do we need to include `Webbit` license (since it is BSD) in Hadoop's distribution? or we don't have to include `Webbit` license, since it is bundled in Netty and mentioned in Netty's notice file.

(Justin Mclean)**Yep you need to include anything that is bundled. [1] If any of the course code from Webbit ends up in a release then yes the BSD license should be added to the release’s LICENSE file. [2] There is no need to have it mentioned in NOTICE as Hadoop have done. [3]**



> Using Hadoop as a model for constructing your own LICENSE and NOTICE files is not a bad idea.

(Justin Mclean)**It looks to me that the Hadoop NOTICE file contains far too much information, and while this is not a licensing error, it best to keep the NOTICE as short as possible [1] as it can effect downstream projects. Permissive licenses such as MIT and BSD should not be mentioned in NOTICE. [2] it also looks like too many copyright lines have been added there (a common mistake), only relocated copyright lines should be mentioned or lines from other ALv2 NOTICE files. [3]**


(Justin Mclean)**Generally NOTICE doesn’t contain copyrights as they are part of the license text and should go in (or be pointed at) by LICENSE. Attributions also do not belong in a NOTICE file and also go in LICENSE. Only relocated headers [4], contents of upstream ALv2 NOTICE files [5], and required 3rd party notices that don’t belong in license [6] need to be in NOTICE.**

**The third inclusion is uncommon and would apply to licenses that have clauses that state you must link to the original source code or state that changes has been made to the original source code need to be listed. An advertising clause (if it wasn’t Category X) would also be placed in NOTICE.**

**I covered this in detail a few times in a talk I’ve given at ApacheCon. It's was recorded at ApacheCon Miami and may help people in this thread. [7]**

1. http://www.apache.org/dev/licensing-howto.html#guiding-principle
2. http://www.apache.org/dev/licensing-howto.html#permissive-deps
3. http://www.apache.org/dev/licensing-howto.html#mod-notice
4. https://www.apache.org/legal/src-headers.html#headers
5. http://www.apache.org/dev/licensing-howto.html#alv2-dep
6. https://www.apache.org/legal/resolved.html#required-third-party-notices
7. https://www.youtube.com/watch?v=I0-lp1t9ee0

sijie:
@ivankelly based on the discussion, I think the NOTICE file should be simplified. most of the content should go to LICENSE file. 

- "If an ALv2 dependency supplies a NOTICE file, its contents must be analyzed and the relevant portions bubbled up into the top-level NOTICE file." [[1]](http://www.apache.org/dev/licensing-howto.html#alv2-dep)

- from Justin Clean: Generally NOTICE doesn’t contain copyrights as they are part of the license text and should go in (or be pointed at) by LICENSE. Attributions also do not belong in a NOTICE file and also go in LICENSE. Only relocated headers [4], contents of upstream ALv2 NOTICE files [5], and required 3rd party notices that don’t belong in license [6] need to be in NOTICE.

ivankelly:
>  if the LICENSE and NOTICE file can not be generated automatically and still need reviews and eyes, I would suggest removing removing versions from the jars. This increases inconvenience for contributors who sends PRs to bump the version. We should make things easier for contributors.

LICENSEs and NOTICEs can change between versions. Dependencies can be added or removed. They can pull in code from thirdparties into their own jar. Without the version, we go back to not having a clue what we are actually shipping.

If you don't want to add a burden on contributors when bumping versions, then we shouldn't ship binary distributions. It's as simple as that.

>  line 205-305 in LICENSE-all.bin.txt isn't really needed.

Yes, they're not absolutely needed. However, they're there so that the tool can automatically check that everything we ship in the tarball mentioned in the LICENSE. Otherwise, this check is 100% manual. There are 73 jars there. That's a lot of room for omissions and mistakes. And people won't check it, they haven't been checking it so far.

The URLs are their to make it easier to bump the versions. If a dependency version changes, it should be trivial to find the NOTICE/LICENSE of the newer version.

> please move deps to all module or rename it to deps/all/ rather than in the root module. because server module has different deps than all module. it is confusing and unclear deps is used by what module. and we should not include licenses that we don't bundle.

Server module deps are a subset of all module deps. If I move it to the all module it will mean having duplicate copies of some of the licenses in the source tree. I'll move if you're ok with the dupes.

ivankelly:
To add to the point about versions.

From the license howto (http://www.apache.org/dev/licensing-howto.html)
- Section: "Bundling Permissively-Licensed Dependencies"
"NOTE: It's also possible to include the text of the 3rd party license within the LICENSE file. This is best reserved for short licenses. It's important to specify the version of the dependency as licenses are sometimes changed."

- Section: "Bundling an Apache-2.0-licensed Dependency"
"However, for completeness it is useful to list the products and their versions, as is done for products under other licenses."

sijie:
Okay fine for me about the versions. Please write a doc on how to update license file when bumping dependencies. Lets try it out, if other contributors feel it creates barriers, we can revert it back. Since you are adding versions in license file, to prevent it getting out of sync quickly, please add the execution of the script to both Travis and Jenkins CI.

Regarding deps, as the policy says, we should not ship license we don't bundle. So unfortunately we have to duplicate.

ivankelly:
> Regarding deps, as the policy says, we should not ship license we don't bundle. So unfortunately we have to duplicate.

For this we can do specific includes. The checker script also checked that deps referenced from the LICENSE match those in the actual tarball.

sijie:
Either duplication or specific includes works for me. Whatever makes things clear to maintain.

ivankelly:
Will do specific includes.

ivankelly:
@sijie
- updated for server package
- explicit includes for license files
- documentation for the website (under developers section)

ivankelly:
@sijie how do you test changes to CI?

bookkeeper-dist/src/assemble/bin-all.xml
```diff
@@ -50,6 +50,33 @@
...
+      <outputDirectory>/deps</outputDirectory>
+      <includes>
+        <include>javax.servlet-api-3.1.0/CDDL+GPL-1.1</include>
+        <include>jsr-305/LICENSE</include>
+        <include>netty-3.10.1.Final/LICENSE.base64.txt</include>
```
sijie: nit: can this simplified with `netty-3.10.1.Final/*`?

bookkeeper-dist/src/assemble/bin-all.xml
```diff
@@ -50,6 +50,33 @@
...
+        <include>netty-3.10.1.Final/LICENSE.jsr166y.txt</include>
+        <include>netty-3.10.1.Final/LICENSE.jzlib.txt</include>
+        <include>netty-3.10.1.Final/LICENSE.webbit.txt</include>
+        <include>netty-4.1.12.Final/LICENSE.base64.txt</include>
```
sijie: nit: can this simplified with `netty-4.1.12.Final/*`?

bookkeeper-dist/src/assemble/bin-server.xml
```diff
@@ -45,6 +45,24 @@
         <include>${basedir}/*.txt</include>
       </includes>
     </fileSet>
+    <fileSet>
+      <directory>../src/main/resources/deps</directory>
+      <outputDirectory>/deps</outputDirectory>
+      <includes>
+        <include>javax.servlet-api-3.1.0/CDDL+GPL-1.1</include>
+        <include>netty-4.1.12.Final/LICENSE.base64.txt</include>
```
sijie: nit: can this simplified with netty-4.1.12.Final/*?

bookkeeper-dist/src/main/resources/NOTICE-all.bin.txt
```diff
@@ -1,55 +1,162 @@
...
 
-Licensed under the Apache License, Version 2.0 (the
-"License"); you may not use this file except in compliance
+This product includes software developed at
+The Apache Software Foundation (http://www.apache.org/).
+
+This product contains a modified version of 'Circe', a high-performance
+hash algorithm framework & library from Trevor Robinson:
+
+  * LICENSE: Apache License 2.0
```
sijie: I don't think we need any copy rights in NOTICE per Justin MClean. All these below are already in LICENSE. NOTICE is informative and should be as brief as possible.

bookkeeper-dist/src/main/resources/NOTICE-server.bin.txt
```diff
@@ -1,43 +1,122 @@
...
 
-Licensed under the Apache License, Version 2.0 (the
-"License"); you may not use this file except in compliance
+This product includes software developed at
+The Apache Software Foundation (http://www.apache.org/).
+
+This product contains a modified version of 'Circe', a high-performance
+hash algorithm framework & library from Trevor Robinson:
+
+  * LICENSE: Apache License 2.0
```
sijie: I don't think we need any copy rights in NOTICE per Justin MClean. All these below are already in LICENSE. NOTICE is informative and should be as brief as possible.

site/docs/latest/development/licensing.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Apache Software License, Version 2 dependencies
+
+1. Add the jar under "The following bundled 3rd party jars are distributed under the Apache Software License, Version 2."
+2. Add a link to the source code of this dependency if available. This will help anyone updating the license in the future.
+3. Check the LICENSE file of the dependency.
+  - If it only contains the Apache Software License, Version 2, nothing needs to be copied into our LICENSE file.
+  - If there is something other than the ASLv2, then you must figure out if it refers to code that is actually shipped in the jar. If it is shipped in the jar, then check the license of that code and apply the rules to it as you would if it was a first order dependency.
+4. Check the NOTICE file of the dependency, if there is one.
+  - Copy any copyright notices to our NOTICE, unless they are for Apache Software Foundation (already covered by our own copyright notice), or they are refer to code covered by a BSD/MIT style license (some projects mistakenly put these in the NOTICE file, but these should be noted in the _LICENSE_ file).
```
sijie: the principle to callout: NOTICE is informative, and should be as brief as possible. typically should only include relocated copyrights (e.g. sources code copied / imported)

site/docs/latest/development/licensing.md
```diff
@@ -0,0 +1,66 @@
+---
+title: Third party dependencies and licensing
+---
+
+The bookkeeper project ships two binary distributions.
```
sijie: better to call out LICENSE/NOTICE in source package vs LICENSE/NOTICE in binary package, when to update them accordingly.

bookkeeper-dist/src/assemble/bin-all.xml
ivankelly: done

bookkeeper-dist/src/assemble/bin-all.xml
ivankelly: done

bookkeeper-dist/src/assemble/bin-server.xml
ivankelly: done

bookkeeper-dist/src/main/resources/NOTICE-all.bin.txt
ivankelly: Removed from all NOTICE files.

bookkeeper-dist/src/main/resources/NOTICE-server.bin.txt
ivankelly: done

site/docs/latest/development/licensing.md
ivankelly: Added a section for source dependencies.

site/docs/latest/development/licensing.md
ivankelly: Added a lists of rules of thumb at the top which covers this.

ivankelly:
> Jenkins is a bit tricky to test. but I think it is okay to just use travis to test the licenses, which is enough.

Added to travis. It would be good to have some docs on how to modify jenkins too.

sijie:
@ivankelly there is a page: http://bookkeeper.apache.org/community/testing/

ivankelly:
Where is that linked from? It would be good to get a sidebar on the community docs, as there's a lot in there now. This licensing stuff should probably go their too.

sijie:
@ivankelly 

regarding NOTICE: my comment is about removing the content below circe, because all of them are covered by LICENSE. see Justin's comment:

> do we need to include Webbit license (since it is BSD) in Hadoop's distribution? or we don't have to include Webbit license, since it is bundled in Netty and mentioned in Netty's notice file.

> (Justin Mclean)Yep you need to include anything that is bundled. [1] If any of the course code from Webbit ends up in a release then yes the BSD license should be added to the release’s LICENSE file. [2] There is no need to have it mentioned in NOTICE as Hadoop have done. [3]

I don't think you should remove `circe` from the notice. because it is a relocated copyright, which should appear in NOTICE, according to https://www.apache.org/legal/src-headers.html#headers

sijie:
> Where is that linked from? It would be good to get a sidebar on the community docs, as there's a lot in there now. This licensing stuff should probably go their too.

I added this link when rewrite the jenkins using dsl. but I probably forgot to add it to the sidebar. but it should be a testing guide under community (I will send a PR for it). licensing part can probably go to coding guide.

ivankelly:
@sijie 

Regarding circe, it's code that ASF does not own the copyright on, and it's not the copyright holder submitting it, so there's no relocated copyright.

So it's not https://www.apache.org/legal/src-headers.html#headers, but https://www.apache.org/legal/src-headers.html#3party

ivankelly:
Moving to community/, there's no sidebar at all for community

sijie:
@ivankelly sidebar is for documentation, which contains the content varies between versions/releases. the content don't belong to release specific should go to other pages.

ivankelly:
yes, I'm saying it would be useful to have one for community docs as well. not a concern for this PR in any case.

sijie:
@ivankelly you are right with circe, it is not a relocated copyright. fine with that.

sijie:
@ivankelly can you check my [comment](#issuecomment-356579707)? I think we don't need those copyrights in binary NOTICE files, based on Justin's comments.

ivankelly:
@sijie ah, yes, will remove

ivankelly:
oh, actually I misunderstood your comments, looking again

ivankelly:
I've removed some of the unneeded stuff. The Intel stuff is required, as it's relocated copyrights. For the other projects, I've taken their NOTICE files in complete form.

bookkeeper-dist/src/main/resources/NOTICE-all.bin.txt
```diff
@@ -1,55 +1,136 @@
...
-  Twitter Util under The Apache Software License, Version 2.0
-  Vertx under The Apache Software License, Version 2.0
-  ZooKeeper under The Apache Software License, Version 2.0
-
-This product contains a modified version of 'Circe', a high-performance
-hash algorithm framework & library from Trevor Robinson:
-
-  * LICENSE: Apache License 2.0
-  * HOMEPAGE: https://github.com/trevorr/circe
+lib/io.netty-netty-all-4.1.12.Final.jar contains a modified portion of 'Apache Harmony', an open source
```
sijie: where does this come from? 

1) I didn't see it appears in netty's notice: https://github.com/netty/netty/blob/netty-4.1.12.Final/NOTICE.txt#L61
2) it doesn't belong to relocated copyright. relocated copyright only applies for src files. http://www.apache.org/legal/src-headers.html#headers

bookkeeper-dist/src/main/resources/NOTICE-all.bin.txt
```diff
@@ -1,55 +1,136 @@
...
-Copyright 2011-2017 The Apache Software Foundation
+Copyright 2011-2018 The Apache Software Foundation
 
-Licensed under the Apache License, Version 2.0 (the
-"License"); you may not use this file except in compliance
+This product includes software developed at
+The Apache Software Foundation (http://www.apache.org/).
+
+------------------------------------------------------------------------------------
+- lib/io.dropwizard.metrics-metrics-core-3.1.0.jar
```
sijie: It is okay and enough to keep versioning in LICENSE. However I would suggest removing versioning from NOTICE file: 1) versioning is not legally required in NOTICE. 2)  NOTICE as short as possible as it will affect downstream projects. this is going to cause NOTICE updates everytime we bumped version.

 http://www.apache.org/dev/licensing-howto.html#mod-notice

ivankelly: the versions are there to facilitate automated checking. if we remove the versions we can't ensure that what is in the notice file exists in the shipped tarball. It will need to be checked manually, so more chance of a -1 on a release candidate.

Also, notices can change between version, so the rationale for having them in the LICENSE holds for the NOTICE also.

bookkeeper-dist/src/main/resources/NOTICE-all.bin.txt
ivankelly: From the harmony NOTICE.
http://svn.apache.org/viewvc/harmony/enhanced/java/trunk/NOTICE?revision=929253&view=markup

The bit they pull from harmony is parts of java.lang.String, which is from intel.
http://svn.apache.org/viewvc/harmony/enhanced/classlibadapter/trunk/modules/kernel/src/main/java/java/lang/String.java?view=markup#l88

netty should be bubbling this up.

ivankelly: https://github.com/netty/netty/commit/681d4609382f9741325b0b2743eb19b85352f425

sijie: @ivankelly netty is not an ASF project and it is not our responsibility for modifying a NOTICE file beyond projects. we are bundled netty 4.1.12, which should use the NOTICE in 4.1.12, if netty is going to fix that in future releases and we bump the version, we include the updated NOTICE.

ivankelly: Apache harmony is an ASF project. netty doesn't include any NOTICE for harmony, and this is wrong. We ship code which belongs to harmony, so we should also ship it's notice. That it come in via netty is largely irrelevant. We can ask about this to legal-discuss@ if you like.

bookkeeper-dist/src/main/resources/NOTICE-all.bin.txt
ivankelly: People will do the bare minimum to make their builds pass. If there are no versions in the NOTICE, they won't touch the notice, and we'll end up with -1 on release candidates.

This notice should have no effect on downstream projects. Downstream depend on our maven jar, not our binary tarballs, which is covered by the top level NOTICE.

sijie: @ivankelly fine with me.

bookkeeper-dist/src/main/resources/NOTICE-all.bin.txt
sijie: please ask. we are not directly bundling harmony, we bundled netty. harmony is not the dependency of netty, it is the source code copied by netty. I don't think we need to include harmony NOTICE, because as what you said NOTICE can be changed between versions, you can't use the latest NOTICE in harmony as the NOTICE when netty introduced the modified code. so we have to stick to whatever NOTICE shipped by netty.

ivankelly: I'll email them tomorrow

ivankelly: I dug down deeper. The String class that netty copies from is originally from IBM,
https://issues.apache.org/jira/browse/HARMONY-14
http://mail-archives.apache.org/mod_mbox/harmony-dev/200511.mbox/browser

They've assigned ASF copyright in the initial contrib though, except for ICU stuff, which isn't part of String, so it turns out there's no mention needed in the notice. Will remove from this patch. Netty NOTICE is still wrong though, will ping them about it.

ivankelly:
@sijie could we get some movement on this? I've removed the harmony stuff in the end, as I tracked down the original harmony submission and copyright was assigned to ASF

sijie:
@ivankelly +1

sijie:
merged. thanks @ivankelly for driving this and make this happen in 4.7.0 release.

PR #889, opened 2017-12-19
BP-14 Implementation of WriteFlag.DEFERRED_SYNC on Journal

eolivelli:
Implement WriteFlags.DERERRED_SYNC on Bookie side.
In case of DEFERRED_SYNC write Journal will early acknowledge the write, just after flushing the buffer to the OS cache, but before waiting for an fsync.


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -283,30 +283,43 @@ public boolean accept(long journalId) {
         WriteCallback cb;
         Object ctx;
         long enqueueTime;
+        boolean ackBeforeForce;
+        boolean persisted;
```
eolivelli: maybe this should be 'volatile'

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -960,6 +976,16 @@ public void run() {
                             }
                             journalFlushWatcher.reset().start();
                             bc.flush(false);
+
```
sijie: why this logic here can't merge with line 1015-1018?

entry.persisted = !entry.ackBeforeForce?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -283,30 +283,43 @@ public boolean accept(long journalId) {
...
+                          ackBeforeForce, persisted);
+            }
+
+            if ((!ackBeforeForce) || (ackBeforeForce && !persisted)) {
+                // for entries with ackBeforeForce=true this callback will be executed twice
+                // we will update the LastAddSynced cursor on the second call
+                journalAddEntryStats.registerSuccessfulEvent(MathUtils.elapsedNanos(enqueueTime), TimeUnit.NANOSECONDS);
+                cb.writeComplete(0, ledgerId, entryId, null, ctx);
+            }
+            if (persisted){
```
sijie: why does it only recycle on `persisted=true`? how do you recycle objects for `persisted=false? I can't find any references in this pull request.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/DeferredSyncBookieTest.java
```diff
@@ -0,0 +1,89 @@
...
+import org.apache.bookkeeper.proto.BookkeeperInternalCallbacks.WriteCallback;
+import org.apache.bookkeeper.test.BookKeeperClusterTestCase;
+import org.junit.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Tests about bookie-side implementation of {@link WriteFlag#DEFERRED_SYNC} write flag.
+ */
+public class DeferredSyncBookieTest extends BookKeeperClusterTestCase {
```
sijie: if this class is only for testing bookie-side implementation, it should be in `bookie` package rather than `client` package. also you should just test bookie side changes, you don't need to involve client.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/DeferredSyncBookieTest.java
```diff
@@ -0,0 +1,89 @@
+/*
+ * Copyright 2017 The Apache Software Foundation.
```
sijie: remove this line. it is an apache project, you shouldn't include "Copyright" line here.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -1248,7 +1248,7 @@ private LedgerDescriptor getLedgerForEntry(ByteBuf entry, final byte[] masterKey
                 bb.put(masterKey);
                 bb.flip();
 
-                getJournal(ledgerId).logAddEntry(bb, new NopWriteCallback(), null);
+                getJournal(ledgerId).logAddEntry(bb, false /* ackBeforeForce */, new NopWriteCallback(), null);
```
sijie: we stick to `deferred_sync`. I would suggest us using `ackBeforeSync` rather than `ackBeforeForce`. we should make sure naming consistent across the places.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -283,30 +283,43 @@ public boolean accept(long journalId) {
         WriteCallback cb;
         Object ctx;
         long enqueueTime;
+        boolean ackBeforeForce;
```
sijie: I would suggest defer adding these two variables until you really need them. it isn't clear how callback will be executed and the object will be recycled in some branches. this pull request should just take the flags and acknowledge before sync.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/DeferredSyncBookieTest.java
eolivelli: I would like so, but CRC32DigestManager is not available in that package

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/DeferredSyncBookieTest.java
eolivelli: sorry, wrong license, done

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
eolivelli: No, maybe the name 'persisted' is confusing or it is not clear the meaning.
We have two kind of QueueEntry for writes:
- normal (ackBeforeForce=false): the *run* method is called once, only when we are sure that data has been *persisted* (FileChanned forced) and after calling the callback we can "recycle" the object
- deferred_sync (ackBeforeForce=true): the callback is called once, but the *run* method is called twice, the first time is after the *flush(false)*, the second time is after the entry as been *persisted* (FileChanned forced) and we can recycle the object after the second call.


We are setting *persisted* = true when we are sure that the entry as really been persisted (forced/fsync'd), the QueueEntry object will never be used for this 'entry' and so we can put it in the recycler.
Every entry always starts with *persisted* = false and then the journal sets persisted = true after the force().

*ackBeforeForce* is a "constant" for the entry and it marks the type of write (deferred_sync vs normal)

We have to call the *run* method twice for deferred_sync entries because at the first time we are firing the callback (acknowledging the write to the client), and at the second time we will update the cursor on the Bookie side which tracks which entries have been persisted durably (this is not present in this patch).
If I remove the entry (or set it to *null*) from the toFlush, I call the callback and then recycle the entry I will never have a way to track that the entry has been really persisted (forced/fsync'd)

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
eolivelli: see other comment

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
eolivelli: I think this is needed, because the lifecycle of the QueueEntry is different, we can't simply fire the callback and forget about the entry, see my comments below

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
eolivelli: ok for "ackBeforeSync"

eolivelli:
@sijie addressed some of the comments, see my explanations for the rest

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -857,6 +870,7 @@ public void run() {
         LOG.info("Starting journal on {}", journalDirectory);
 
         RecyclableArrayList<QueueEntry> toFlush = entryListRecycler.newInstance();
+        int waitingFlush = 0;
```
sijie: nit: numEntriesToFlush?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -960,19 +974,31 @@ public void run() {
                             }
                             journalFlushWatcher.reset().start();
                             bc.flush(false);
+
+                            for (int i = 0; i < toFlush.size(); i++) {
```
sijie: any reason why you still keep the for-loop here? There should be no code added between line 975 and line 988, it is registering the time journal thread spends on flushing.

I suggested you merge this for-loop with line 1005-1038, to keep branches handling close to each other. any reason why you didn't do that?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -939,7 +953,7 @@ public void run() {
                             shouldFlush = true;
                             flushMaxWaitCounter.inc();
                         } else if (qe != null
-                                && ((bufferedEntriesThreshold > 0 && toFlush.size() > bufferedEntriesThreshold)
```
sijie: you should use `toFlush.size()` here. otherwise you might potentially have a list accumulating a lot of nulls

sijie: make sure you have a test case cover here

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieJournalForceTest.java
```diff
@@ -0,0 +1,166 @@
...
+import org.powermock.core.classloader.annotations.PrepareForTest;
+import org.powermock.modules.junit4.PowerMockRunner;
+
+/**
+ * Test the bookie journal.
+ */
+@RunWith(PowerMockRunner.class)
+@PrepareForTest({JournalChannel.class})
+@Slf4j
+public class BookieJournalForceTest {
```
sijie: Add test cases:

- interleaved requests (ackBeforeSync=true and ackBeforeSync=false)
- the branch I mentioned above.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieJournalForceTest.java
```diff
@@ -0,0 +1,166 @@
...
+
+    private static final ByteBuf DATA = Unpooled.wrappedBuffer(new byte[]{});
+
+    @Rule
+    public TemporaryFolder tempDir = new TemporaryFolder();
+
+    @Test
+    public void testAckBeforeSync() throws Exception {
```
sijie: the logic in testAckBeforeSync and testAckAfterSync does sounds good to me.

I think for:

- for testAckBeforeSync, you need to ensure callback come before sync and eventually sync will still advance the file and the log marker
- for testAckAfterSync, you need to ensure callback come back only after sync and the log markers are advanced after that.

One solution for this:

- you need to be able to control force write thread, e.g. suspend forcewrite thread and make sure callbacks can trigger even forcewrite thread is suspended and verify the forcewrite request enqueued to forcewrite queue contains null items.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieWriteToJournalTest.java
```diff
@@ -0,0 +1,112 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+package org.apache.bookkeeper.bookie;
+
+
```
sijie: remove blank line

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieWriteToJournalTest.java
```diff
@@ -0,0 +1,112 @@
...
+        File ledgerDir = tempDir.newFolder();
+        Bookie.checkDirectoryStructure(Bookie.getCurrentDirectory(ledgerDir));
+        ServerConfiguration conf = TestBKConfiguration.newServerConfiguration();
+        conf.setJournalDirName(journalDir.getPath())
+            .setLedgerDirNames(new String[]{ledgerDir.getPath()})
+            .setZkServers(null);
+        Bookie b = new Bookie(conf);
+        b.start();
+        Boolean[] effectiveAckBeforeSync = new Boolean[1];
+        replace(method(Journal.class, "logAddEntry", ByteBuf.class, Boolean.TYPE, WriteCallback.class, Object.class))
```
sijie: why not use PowerMock to mock journal constructor? so you will have mocked journal object. so that you can assert journal method calls with expected flags.

eolivelli: @sijie  @ivankelly @yzang 

I have updated this test and now I am using a full mock Journal, now the test is super "flaky" I think due to the caching of FileInfo, if I run the test with a debugger it passes.

I wonder if I have to spend time in fixing the flakiness or revert to previous test or wait for fixes on FileInfo cache flakyness

```
org.apache.bookkeeper.bookie.Bookie$NoLedgerException: Ledger 1 not found
	at org.apache.bookkeeper.bookie.IndexPersistenceMgr$FileInfoLoader.call(IndexPersistenceMgr.java:226)
	at org.apache.bookkeeper.bookie.IndexPersistenceMgr$FileInfoLoader.call(IndexPersistenceMgr.java:211)
	at com.google.common.cache.LocalCache$LocalManualCache$1.load(LocalCache.java:4904)
	at com.google.common.cache.LocalCache$LoadingValueReference.loadFuture(LocalCache.java:3628)
	at com.google.common.cache.LocalCache$Segment.loadSync(LocalCache.java:2336)
	at com.google.common.cache.LocalCache$Segment.lockedGetOrLoad(LocalCache.java:2295)
	at com.google.common.cache.LocalCache$Segment.get(LocalCache.java:2208)
	at com.google.common.cache.LocalCache.get(LocalCache.java:4053)
	at com.google.common.cache.LocalCache$LocalManualCache.get(LocalCache.java:4899)
	at org.apache.bookkeeper.bookie.IndexPersistenceMgr.getFileInfo(IndexPersistenceMgr.java:257)
	at org.apache.bookkeeper.bookie.IndexPersistenceMgr.isFenced(IndexPersistenceMgr.java:499)
	at org.apache.bookkeeper.bookie.LedgerCacheImpl.isFenced(LedgerCacheImpl.java:147)
	at org.apache.bookkeeper.bookie.InterleavedLedgerStorage.isFenced(InterleavedLedgerStorage.java:208)
	at org.apache.bookkeeper.bookie.LedgerDescriptorImpl.isFenced(LedgerDescriptorImpl.java:76)
	at org.apache.bookkeeper.bookie.Bookie.addEntry(Bookie.java:1361)
	at org.apache.bookkeeper.bookie.BookieWriteToJournalTest.testJournalLogAddEntryCalledCorrectly(BookieWriteToJournalTest.java:110)
```

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
eolivelli: done

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
eolivelli: changed, will add tests ASAP

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
eolivelli: This is because I have to call the callback "before" line 1007, that is requesting the fsync the file and firing the callbacks.

```
forceWriteRequests.put(createForceWriteRequest(logFile, logId, lastFlushPosition,
                                                                               toFlush, shouldRolloverJournal, false));
```

If I fire the callbacks "after" that line the callback could be fired by the ForceWriteThread as well.

So I enqueue the callbacks (QueueEntry) into the threadpool and remove them from the toFlush list.
The ForceWriteThread will find nulls for callbacks already fired

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieJournalForceTest.java
eolivelli: done

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieWriteToJournalTest.java
eolivelli: @ivankelly found the problem, the Bookie exists with the mock journal, I have to mock the Journal#join method

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieJournalForceTest.java
eolivelli: @sijie I have pushed a new testcase (testAckBeforeSyncSuspendingJournal) which suspends the ForceWriteThread, please take a look if this approach sounds good to you I will clean up the suite and create the missing tests

eolivelli:
@sijie when you have time can you please take a look at my answers and at the new test case which suspends the ForceWriteThread ?

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieJournalForceTest.java
```diff
@@ -0,0 +1,405 @@
...
+        long ledgerId = 1;
+        long entryId = 0;
+        journal.logAddEntry(ledgerId, entryId, DATA, true /* ackBeforeSync */, new WriteCallback() {
+            @Override
+            public void writeComplete(int rc, long ledgerId, long entryId, BookieSocketAddress addr, Object ctx) {
+                latch.countDown();
+            }
+        }, null);
+        assertTrue(latch.await(20, TimeUnit.SECONDS));
+        assertTrue(supportQueue.isEmpty());
```
sijie: better assert that `forceWrite` is not called.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieJournalForceTest.java
```diff
@@ -0,0 +1,405 @@
...
+
+        doAnswer((Answer) (InvocationOnMock iom) -> {
+            supportQueue.put(iom.getArgument(0));
+            return null;
+        }).when(forceWriteRequests).put(any(ForceWriteRequest.class));
+
+        when(forceWriteRequests.take()).thenAnswer(new Answer<ForceWriteRequest>() {
+            @Override
+            public ForceWriteRequest answer(InvocationOnMock iom) throws Throwable {
+                forceWriteThreadSuspendedIsWaiting.countDown();
```
sijie: you don't really need `forceWriteThreadSuspendedIsWaiting` here. because what you want to test here is `forceWrite` is never called, when ackBeforeSync is completed. and you don't really need another supportQueue.

```
CountDownLatch forceWriteThreadSuspendedLatch = new CountDownLatch(1);
BlockingQueue<ForceWriteRequest> forceWriteRequests = mock(BlockingQueue.class);
when(forceWriteRequests.take()).thenAnswer(invocationOnMock -> {
    // suspend the force write thread
    forceWriteThreadSuspendedLatch.await();

   return invocationOnMock.callRealMethod();
});

Whitebox.setInternalState(journal, "forceWriteRequests", forceWriteRequests);

// mock journal channel ...

journal.start();

journal.logAddEntry(...); // add an entry with `ackBeforeSync = true`

// wait for addentry to complete
// verify forceWrite is never called.
verify(jc, times(0)).forceWrite(...);

// resume the forcewrite queue
forceWriteThreadSuspendedLatch.countDown();
```

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieJournalForceTest.java
```diff
@@ -0,0 +1,405 @@
...
+        for (long entryId = 0; entryId < numEntries; entryId++) {
+            journal.logAddEntry(ledgerId, entryId, DATA, true /* ackBeforeSync */, new WriteCallback() {
+                @Override
+                public void writeComplete(int rc, long ledgerId, long entryId, BookieSocketAddress addr, Object ctx) {
+                    latch.countDown();
+                }
+            }, null);
+        }
+        assertTrue(latch.await(20, TimeUnit.SECONDS));
+        assertTrue(flushMaxOutstandingBytesCounter.get() > 1);
```
sijie: also assert forceWrite called times.

eolivelli:
@sijie I did a push now but patch is not ready yet.
Will continue to work on this tomorrow

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieJournalForceTest.java
eolivelli: @sijie your approach is not working because a mock BlockingQueue is not enoufh, the ForceWriteThread needs a real queue, which retains entries and finally returns them.
And we cannot use callRealMethod() on a "interface"

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieJournalForceTest.java
eolivelli: done

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieJournalForceTest.java
```diff
@@ -0,0 +1,405 @@
...
+                }
+            }, null);
+        }
+        assertTrue(latchAckBeforeSynch.await(20, TimeUnit.SECONDS));
+        assertTrue(latchAckAfterSynch.await(20, TimeUnit.SECONDS));
+
+        journal.shutdown();
+    }
+
+    private static class CounterImpl implements Counter {
```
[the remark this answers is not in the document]
eolivelli: done

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieWriteToJournalTest.java
```diff
@@ -0,0 +1,130 @@
...
+        File ledgerDir = tempDir.newFolder();
+        Bookie.checkDirectoryStructure(Bookie.getCurrentDirectory(ledgerDir));
+        ServerConfiguration conf = TestBKConfiguration.newServerConfiguration();
+        conf.setJournalDirName(journalDir.getPath())
+            .setLedgerDirNames(new String[]{ledgerDir.getPath()})
+            .setZkServers(null);
+        BookieSocketAddress bookieAddress = Bookie.getBookieAddress(conf);
+        CountDownLatch journalJoinLatch = new CountDownLatch(1);
+        Journal journal = mock(Journal.class);
+        Boolean[] effectiveAckBeforeSync = new Boolean[1];
```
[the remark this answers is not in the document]
eolivelli: done

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieWriteToJournalTest.java
```diff
@@ -0,0 +1,130 @@
...
+            data.writeLong(entryId);
+            final long expectedEntryId = entryId;
+            b.addEntry(data, ackBeforeSync, (int rc, long ledgerId1, long entryId1,
+                                             BookieSocketAddress addr, Object ctx) -> {
+                assertSame(expectedCtx, ctx);
+                assertEquals(ledgerId, ledgerId1);
+                assertEquals(expectedEntryId, entryId1);
+                latch.countDown();
+            }, expectedCtx, masterKey);
+            assertTrue(latch.await(30, TimeUnit.SECONDS));
```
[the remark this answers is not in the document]
eolivelli: done

eolivelli:
@sijie I know this patch needs rebase, but I think I have addressed all of your comments, so it is ready for review. I will rebase before merge

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieJournalForceTest.java
sijie: @eolivelli sorry it was my typo. you should be using `spy`.

`BlockingQueue<ForceWriteRequest> forceWriteRequests = spy(new LinkedBlockingQueue())`

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -901,7 +921,7 @@ public void run() {
                                 TimeUnit.NANOSECONDS);
                     }
 
-                    if (toFlush.isEmpty()) {
+                    if (numEntriesToFlush == 0) {
```
sijie: I don't think we need to change this line here. It is better to use toFlush here, because toFlush will never decrease on each run, while `numEntriesToFlush` can be changed. it is good to use toFlush as the condition, because all the group commit logic is based on toFlush.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -959,19 +979,31 @@ public void run() {
...
-                                              e.ledgerId, e.entryId);
+                                    if (e != null) {
+                                        LOG.debug("Written and queuing for flush Ledger: {}  Entry: {}",
+                                                  e.ledgerId, e.entryId);
+                                    }
                                 }
                             }
 
-                            forceWriteBatchEntriesStats.registerSuccessfulValue(toFlush.size());
+                            forceWriteBatchEntriesStats.registerSuccessfulValue(numEntriesToFlush);
```
sijie: if you are changing to report `numEntriesToFlush`, you should also change `batchSize`. otherwise, you will have inconsistent metrics.

However I don't think you need to change this here. because `forceWriteBatchEntries` and `forceWriteBatchBytes` are measuring how many entries/bytes per batch that submitted to force write queue. ackBeforeSync is only changing the ack behavior, it doesn't change the flushing or force write behavior. so we shouldn't change the metric here.

In that sense, I don't see a need of `numEntriesToFlush`.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -959,19 +979,31 @@ public void run() {
                             }
                             journalFlushWatcher.reset().start();
                             bc.flush(false);
+
+                            for (int i = 0; i < toFlush.size(); i++) {
```
sijie: also I think I had a comment before : it's better not to put this loop in between
line 981: `bc.flush(false);` and
line 992: `lastFlushPosition = bc.position();`.

because there is a timer `journalFlushWatcher` that calculating the time spending on flushing. consider moving down to just before `if (syncData) {`. so all the stats/timer are correct as expected and easier to understand what metrics are measuring what logic.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -346,11 +354,16 @@ public int process(boolean shouldForceWrite) throws IOException {
                     this.logFile.forceWrite(false);
                     journalSyncStats.registerSuccessfulEvent(MathUtils.elapsedNanos(startTime), TimeUnit.NANOSECONDS);
                 }
+                LOG.info("setCurLogMark at {}", lastLogMark.getCurMark());
```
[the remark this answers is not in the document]
eolivelli: that was my debug, dropping

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
eolivelli: ok, fine

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
eolivelli: it is better, will revert, thanks

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
eolivelli: oaky, let me try to revert and do another proposal

eolivelli: @sijie I have reverted the change and move the logic to a different branch. Tests are passing.
Is this what you are thinking ?

I wonder if this approach really reduces the number of fsyncs as anyway we are issueing a "ForceWriteRequest".
I would like that in the case all the entries written to the bookie are DEREFFED_SYNC the behaviour should be like having syncData = false

sijie: @eolivelli 

This is different from using 'syncData = false'. because the goal here is add entries can complete before fsync, while the whole assumption of lac still remain. the `syncData = fase` is disabling fsync completely. These two cases have different durability guarantees.

Also I don't think the goal is to reduce the number of fsyncs. But if you want to reduce the number of fsyncs, you can anyway increase the flush commit interval or flush buffer size.

sijie:
@eolivelli do you mind fixing the CI? otherwise, the change is ready to go.

eolivelli:
Thank you @sijie will fix as soon as possible, likely tomorrow

eolivelli:
@sijie resolved conflicts against current master and fixed compilation issue,let's wait for tests on CI

eolivelli:
retest this please

PR #892, opened 2017-12-19
Misc LedgerHandle changes

athanatos:
Adds logging for new ensembles, a getCtime call, and a check for ReadOnlyLedgerHandle in initializeExplicitLacInterval.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
@@ -1633,6 +1633,8 @@ public class LedgerHandle implements WriteHandle {
 
             // We've successfully changed an ensemble
             ensembleChangeCounter.inc();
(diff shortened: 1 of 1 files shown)


athanatos:
Let me know if it would be better for me to submit 3 different PRs.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -204,7 +204,7 @@ public void safeRun() {
     }
 
     protected void initializeExplicitLacFlushPolicy() {
-        if (!metadata.isClosed() && bk.getExplicitLacInterval() > 0) {
+        if (!metadata.isClosed() && !(this instanceof ReadOnlyLedgerHandle) && bk.getExplicitLacInterval() > 0) {
```
reddycharan: yep, I missed this fix. Caught this while porting changes from community  to our codebase, but forgot to push it back to the community.

Issue #893, opened 2017-12-20
Delete & List Cookie BookieShell Command Options


reddycharan:
1. Please describe the feature you are requesting.

Delete & List Cookie BookieShell Command Options

    - add ListCookie Options to UpdateCookie command
    - add DeleteCookie Option to UpdateCookie command
    - move expandstorage command to UpdateCookie command as option, since functionality wise they are similar, and it would be easier for user to choose which command to use for his need
    - in listfilesondisc command use canonical path (full path)

2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have). Are you currently using any workarounds to address this issue?

should have / nice to have

PR #894, opened 2017-12-20
ISSUE #893: more options for UpdateCookieCmd

reddycharan:

- add ListCookie Options to UpdateCookie command
- add DeleteCookie Option to UpdateCookie command
- move expandstorage command to UpdateCookie command as option,
since expandstorage is just a cookie update operation
- in listfilesondisc command use canonical path (full path)

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
@@ -49,6 +49,7 @@ import java.util.Formatter;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -86,6 +87,7 @@ import org.apache.bookkeeper.net.BookieSocketAddress;
 import org.apache.bookkeeper.proto.BookkeeperInternalCallbacks.GenericCallback;
 import org.apache.bookkeeper.replication.AuditorElector;
 import org.apache.bookkeeper.stats.NullStatsLogger;
+import org.apache.bookkeeper.util.BookKeeperConstants;
 import org.apache.bookkeeper.util.DiskChecker;
 import org.apache.bookkeeper.util.EntryFormatter;
 import org.apache.bookkeeper.util.IOUtils;
@@ -99,6 +101,8 @@ import org.apache.commons.cli.BasicParser;
 import org.apache.commons.cli.CommandLine;
 import org.apache.commons.cli.HelpFormatter;
 import org.apache.commons.cli.MissingArgumentException;
+import org.apache.commons.cli.Option;
+import org.apache.commons.cli.OptionBuilder;
 import org.apache.commons.cli.Options;
 import org.apache.commons.cli.ParseException;
 import org.apache.commons.configuration.CompositeConfiguration;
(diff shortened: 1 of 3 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
```diff
@@ -1724,63 +1797,85 @@ private boolean verifyCookie(Cookie oldCookie, File dir) throws IOException {
...
-                    allLedgerDirs.addAll(Arrays.asList(indexDirectories));
-                }
+                if (delete) {
+                    boolean confirm = force;
+                    if (!confirm) {
+                        confirm = IOUtils.confirmPrompt("Are you sure you want to delete Cookies from Zookeeper?");
+                    }
 
-                try {
-                    conf.setAllowStorageExpansion(true);
```
reddycharan: I removed this line, not sure why this is added in this commit -https://github.com/sijie/bookkeeper/commit/0db0cf6181f4949fe4684770bb81442694a24b27

sijie: @reddycharan I added this because this command is 'expandStorage'. so I set this flag explicitly, don't rely on the configuration file.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
```diff
@@ -1724,63 +1797,85 @@ private boolean verifyCookie(Cookie oldCookie, File dir) throws IOException {
...
-                    return -1;
+                    if (confirm) {
+                        cookie.getValue().deleteFromRegistrationManager(rm, bkConf, cookie.getVersion());
+                        LOG.info("Deleted Cookie from Zookeeper for Bookie: {}", bookieAddress);
+                    } else {
+                        LOG.info("Skipping deleting cookie from ZooKeeper for Bookie: {}", bookieAddress);
+                    }
+                } else {
+                    LOG.info("{} bookie's Cookie path in Zookeeper: {} ", bookieAddress,
+                            rm.getCookiePath(bookieAddress.toString()));
```
sijie: do we really need line 1874-1875? I am not sure if it is worth making a getCookiePath interface in RegistrationManager interface just for this purpose.

reddycharan: in this particular command, we are intending to print cookiePath, which depends on RegistrationManager. 

If we don't want to have this method at RegistrationManager interface level, then we can remove it from interface and typecast RegistrationManager instance to ZKRegistrationManager and call method of ZKRegistrationManager.

If you have any other thoughts, I'm open.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
reddycharan: hmm..I'm not sure if overriding conf value is the expected behavior.

sijie: Are you guys the original contributor of `expandstorage`? If the intend is to use whatever value in the configuration, I am fine with removing overrides here. but it would be worth calling out in the help description of this command, because it relies on whatever value provided in the configuration.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
sijie: I would suggest leave this as part of ZKRegistraitonManager for now. Because it is internal to zookeeper implementation.

You don't need to do cast. You can change `try (RegistrationManager rm = new ZKRegistrationManager())` to `try (ZKRegistrationManager rm = new ZKRegistrationManager())`.

reddycharan: done

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
reddycharan: yes we are the original contributors.

Anyhow I reverted the removal..i added it back.

sijie:
link #914

Issue #283, opened 2017-07-24
V2 ReadEntryProcessor should handle asynchronous fence like what V3 does


sijie:
**FEATURE REQUEST**

1. Please describe the feature you are requesting.

Address TODO item in *ReadEntryProcessor*:

                // TODO:
                // currently we don't have readCallback to run in separated read
                // threads. after BOOKKEEPER-429 is complete, we could improve
                // following code to make it not wait here
                //
                // For now, since we only try to wait after read entry. so writing
                // to journal and read entry are executed in different thread
                // it would be fine.

2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have). Are you currently using any workarounds to address this issue?

*nice-to-have*

3. Provide any additional detail on your proposed use case for this feature.

N/A

ivankelly:
In general ReadEntryProcessor#processPacket needs a bit of a refactor to use java 8 features. This change should be pretty straightforward though.

liketic:
Hi @sijie , I opened #2021 , could you please take a look?

PR #898, opened 2017-12-20
Bookies should prioritize recovery reads/writes

ivankelly:
The change adds a new flag for reads, RECOVERY_READ. This flag is set
when a client is retrying to recover a ledger. Recovery reads should
be higher priority than normal reads, as recovery usually happens when
a writer has failed, and another node needs to take over the stream
that that ledger belonged to. All writes on that stream are blocked
until the ledger the ledger has been recovered, so giving these reads
priority will mean the stream is out for action for the shortest
amount of time possible.

Priority is given to these reads by executing the reads directly in
the BookieRequestProcessor, rather than adding them to the
readThreadPool where they could end up queuing.

This change was originally 372a99db in the yahoo.


bookkeeper-proto/src/main/proto/BookkeeperProtocol.proto
```diff
@@ -90,6 +90,7 @@ message ReadRequest {
     enum Flag {
         FENCE_LEDGER = 1;
         ENTRY_PIGGYBACK = 2;
+        RECOVERY_READ = 3;
```
ivankelly: @merlimat @sijie 
There's a potential thorny BC issue here. On yahoo branch, RECOVERY_READ is 2.https://github.com/yahoo/bookkeeper/blob/165971209f933e0bcefaac997db8e75c8776f942/bookkeeper-server/src/main/proto/BookkeeperProtocol.proto#L83

@merlimat are any any the yahoo clients using protobufs? Or are they all on V2? If so it may not be a problem.

eolivelli:
This is an important protocol change how can we do BC tests?

bookkeeper-proto/src/main/proto/BookkeeperProtocol.proto
merlimat: I think the only clients using the v3 protocols are the auto-recovery daemons which are using the default settings. What would be the worst case if they do a read request with `ENTRY_PIGGYBACK` flag? 

The `RECOVERY_READ` flag is anyway optional, in the sense that is currently just used to signal urgency to the bookie

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtocol.java
```diff
@@ -265,7 +265,7 @@ ByteBuf getData() {
         }
 
         boolean isRecoveryAdd() {
```
sijie: the name should be changed here

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
```diff
@@ -152,6 +159,14 @@ public BookieRequestProcessor(ServerConfiguration serverCfg, Bookie bookie,
             createExecutor(
                 this.serverCfg.getNumLongPollWorkerThreads(),
                 "BookieLongPollThread-" + serverCfg.getBookiePort(), OrderedScheduler.NO_TASK_LIMIT);
+        this.fenceThreadPool = Executors.newCachedThreadPool(
```
sijie: I don't think you need a separate thread pool if you fix #283. I would suggest fixing #283 first.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
```diff
@@ -342,7 +361,15 @@ private void processReadRequestV3(final BookkeeperProtocol.Request r, final Chan
             }
         } else {
             read = new ReadEntryProcessorV3(r, c, this, fenceThreadPool);
-            if (null == readThreadPool) {
+
+            // If it's a recovery read, we want to make sure it gets submitted right away,
+            // bypassing the eventual executor queue,
+            // to reduce the chances of ledger recovery exceptions.
+            boolean isRecoveryRead = r.getReadRequest().hasFlag()
```
sijie: Can we make this into some until function and have a test case cover it?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -794,6 +796,19 @@ public void onFailure(Throwable cause) {
...
 
+    /*
+     * Read the last entry in the ledger
+     *
+     * @param cb
+     *            object implementing read callback interface
+     * @param ctx
+     *            control object
+     */
+    public void asyncReadLastEntry(ReadCallback cb, Object ctx) {
```
sijie: Can you remove this from this pull request and create a seperate one for it?

- we need make this as part of new api as well
- we need a test case to cover it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
```diff
@@ -725,6 +689,15 @@ private void readEntryInternal(final long ledgerId,
                 readBuilder = readBuilder.setFlag(ReadRequest.Flag.ENTRY_PIGGYBACK);
             }
 
+            // Only one flag can be set on the read requests
+            if (((short) flags & BookieProtocol.FLAG_DO_FENCING) == BookieProtocol.FLAG_DO_FENCING) {
+                readBuilder.setFlag(ReadRequest.Flag.FENCE_LEDGER);
+                checkArgument(masterKey != null);
```
sijie: dont do this. fail the callback with IncorrectParameterException

bookkeeper-proto/src/main/proto/BookkeeperProtocol.proto
sijie: a read request with `ENTRY_PIGGYBACK` is a normal read request. so it only means you don't assign priorities for recovery read, which I don't think it is a big deal.

ivankelly: I think this definitely needs to be BC tested. The question is whether we block this change to do it, or just make it a blocker on the 4.7.0 release.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
ivankelly: #283 is to submit a runnable to the read processor thread for checking the fencing? I don't think this applies here. This fenceThreadPool is for V3. The problem was that writeThreadPool and readThreadPool can reject executions, if they hit a limit. This was causing failures in on of the tests for fencing (testRecoveryNotRejected). I would think that fencing should be outside of these execution limits, as it is part of recovery.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
ivankelly: Will do.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
ivankelly: +1

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
ivankelly: It's unused it seems. Removing completely for now.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtocol.java
ivankelly: done

bookkeeper-proto/src/main/proto/BookkeeperProtocol.proto
sijie: my feeling is we need to do this as early as possible.  I am not sure if there is other similar changes coming up or not. I understand this is a port from yahoo branch, it probably doesn't make sense to block this change to be BC tested. I am fine with creating a blocker issue for adding BC tests for this change here.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
sijie: ah sorry I read as it is for v2 process. I thought that you move fence read to a separate thread.

then I have a few questions here /cc @merlimat 

1) this change will move fence read to netty I/O thread, which it would be blocking subsequent requests. If read threads is accumulating requests, that typically means I/O is slow. fenceRead can potentially be very slow due to overwhelmed I/O, then it will subsequently block all the network I/O. I am not sure how much we can gain from moving fence read to netty I/O thread. It is against the purposes we having thread pools, is to avoid blocking network I/O. 

@merlimat any metrics or experiences show this change improved things a lot?

2) If this change is to prioritize recovery related ops, isn't that better to have its own recovery read thread pools? rather than blocking network I/O.

3) even worse at V2, the fence read is waiting for both read I/O and journal I/O (v3 makes fence request a non-blocking op on waiting fence result). If the bookie is overwhelmed (e.g. hitting bandwidth limit), journal can become a bottleneck and hence block netty I/O thread. (I assume this would potentially block sending responses back)

All the questions are related to the goal of not blocking network thread. If I read this change correctly, this change is against this goal, which can potentially have very bad impact.


then shall v2 fencing be executed in this fencing thread or not? because v2 fencing is actually a blocking call, while v3 fencing is not a blocking call. If you run v2 fencing in netty I/O thread, isn't it blocking other I/O operations?

bookkeeper-proto/src/main/proto/BookkeeperProtocol.proto
ivankelly: SGTM. Created issue #903

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
merlimat: @sijie This change is a companion of #706. The story here is that in BK 4.6 if the read rate is very high, maxing out the bookie IO capacity what starts to happen is that read requests are piled up in the read executor task queue. 

The read thread is processing the read request, but since they spent a lot of time in the queue, the client will have already timed them out and re-sent other read requests. 

In #706, we've introduced the max-limit for the size of the executor queue, in order to limit the amount of time spent in the queue and rejecting additional requests in order to fail-fast. 

The rationale is to avoid to read ops that are going to be timed-out and that will not make any progress. 

Another problem with the read operations is that when the timeout/rejections happens, it impacts also fencing-reads which are used in recovery. 

This change is to make sure the recovery reads can make progress even though there is a very high rate of regular read requests. As Ivan pointed out, the recovery reads are critical because they typically prevent a writer from being operative, while regular reads are "just" reading data and they can thus be deferred to lower priority. 

The original change was doing the fence-read directly in IO thread as the way to skip the executor thread queue. I guess that using a dedicated thread pool for fence reads is probably a much better option in order not to block the IO threads.

ivankelly: +1 on dedicated recovery thread pool. I'm currently working on BC stuff first though, but will come back once that's done.

sijie: @merlimat +1 on a separate recovery thread pool.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/TestMaxSizeWorkersQueue.java
```diff
@@ -122,4 +131,60 @@ public void addComplete(int rc, LedgerHandle lh, long entryId, Object ctx) {
 
         assertTrue(receivedTooManyRequestsException.get());
     }
+
+    @Test(timeout = 60000)
```
eolivelli: Nit: Remember to drop these timeouts

bookkeeper-proto/src/main/proto/BookkeeperProtocol.proto
jvrao: In the case of pure bookkeeper this feature may be much useful as the ledger is boundary of the writer. Is this the usecase for Pulsar? In general, for Salesforce usecase regular reads are more important( or atleast not less imprtant) than the background recovery reads.  Being said that, there won't be many recovery reads as compared to regular reads, so may be it is OK. This is just first comment, and I haven't read through the patch fully.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
```diff
@@ -725,6 +689,15 @@ private void readEntryInternal(final long ledgerId,
...
+                readBuilder.setMasterKey(ByteString.copyFrom(masterKey));
+            } else if (((short) flags & BookieProtocol.FLAG_RECOVERY) == BookieProtocol.FLAG_RECOVERY) {
+                readBuilder.setFlag(ReadRequest.Flag.RECOVERY_READ);
```
jvrao: Hmm, so the ledger replicator, which could be part of recovering under replicated ledgers gets higher priority than the the actual client reads? That doesn't sound right to me.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
jvrao: Separate thread pool does sound like a good idea, but I would really restrict recovery reads to real fencing reads only, not under replication reads.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerFragmentReplicator.java
```diff
@@ -320,7 +320,7 @@ public void readComplete(int rc, LedgerHandle lh,
                 for (BookieSocketAddress newBookie : newBookies) {
                     bkc.getBookieClient().addEntry(newBookie, lh.getId(),
                             lh.getLedgerKey(), entryId, toSend.retainedSlice(),
-                            multiWriteCallback, dataLength, BookieProtocol.FLAG_RECOVERY_ADD);
+                            multiWriteCallback, dataLength, BookieProtocol.FLAG_RECOVERY);
```
jvrao: Are we giving high priority to replicator reads too? It should not...

asfgit:

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerFragmentReplicator.java
ivankelly: This is an addEntry. The recovery flag has a different meaning for adds. It means that if the ledger has been fenced, still store the entry (as it's not a new entry and the client will not see an entry acknowledged as successful after the point of the fencing event).

ivankelly: oh, seems we're shortcutting those too. hmm.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
ivankelly: @sijie @jvrao @merlimat Trying to get back the context on this again. Any objection to reusing the fencing thread pool for the recovery reads/adds?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerFragmentReplicator.java
ivankelly: I think I need to create a new explicit PRIORITY flag for these operations, rather than using the semantics from RECOVERY, as RECOVERY already means something for adds.

ivankelly:
Pushing to see how it does on CI, was a pretty big merge from master after 2 months dormant. Still need to change the flags, and fixup the executor stuff.

ivankelly:
Rebased because it was getting messy to see what had changed after merging master.

This version is the same as the what was pushed 2 months ago except:
- The flag has been renamed to HIGH_PRIORITY, which is separate to RECOVERY_ADD, so that we can have recovery adds which are not high priority (for rereplication) // @jvrao 
- The fencing thread pool has been renamed the high priority thread. It handles all requests with the high priority flag or the fencing flag.

@sijie @merlimat @jiazhai @jvrao @eolivelli This is ready for review again

ivankelly:
Pushing to see how it does on CI, was a pretty big merge from master after 2 months dormant. Still need to change the flags, and fixup the executor stuff.

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
```diff
@@ -679,27 +674,4 @@ public String getTLSEnabledProtocols() {
...
      */
     protected abstract T getThis();
-
-    /**
-     * returns the string representation of json format of this config.
-     *
-     * @return
-     * @throws ParseJsonException
-     */
-    public String asJson() throws ParseJsonException {
```
eolivelli: This change seems unrelated

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieServer.java
```diff
@@ -82,14 +81,6 @@ public BookieServer(ServerConfiguration conf, StatsLogger statsLogger)
             BookieException, UnavailableException, CompatibilityException, SecurityException {
         this.conf = conf;
         validateUser(conf);
-        String configAsString;
-        try {
-            configAsString = conf.asJson();
-            LOG.info(configAsString);
-        } catch (ParseJsonException pe) {
```
eolivelli: Yes you are reverting a change from @charanreddy

ivankelly: Ah crap, must have been in flight as I was turning the merge into a rebase. Will fix.

ivankelly: updated

bookkeeper-proto/src/main/proto/BookkeeperProtocol.proto
```diff
@@ -90,8 +90,9 @@ message ReadRequest {
     enum Flag {
         FENCE_LEDGER = 1;
         ENTRY_PIGGYBACK = 2;
+        HIGH_PRIORITY = 3;
     }
-    optional Flag flag = 100;
+    repeated Flag flag = 100;
```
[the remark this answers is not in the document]
ivankelly: > hmm I think this changes too much for a portback. i know repeated field is compatible with optional. However I am not sure how it would impact other existing code logic. I don't feel comfortable with change, especially this is coupled with a portback change.

The code changes required are forced by the change because the api for accessing repeated is completely different than that for accessing an optional.

I agree with what you're saying though, I was actually expecting this to fail BC in CI as enums in protobuf don't take kindly to new additions, so I was surprised when it past. My contingency was to add something like you suggest, though a boolean. An int is a good idea though. I'll leave everything else the same with regards to BookieClient.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
```diff
@@ -276,10 +291,12 @@ public void processRequest(Object msg, Channel c) {
             // process packet
             switch (r.getOpCode()) {
                 case BookieProtocol.ADDENTRY:
-                    processAddRequest(r, c);
+                    assert(r instanceof BookieProtocol.ParsedAddRequest);
```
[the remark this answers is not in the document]
ivankelly: Will change.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
```diff
@@ -154,9 +162,15 @@ public BookieRequestProcessor(ServerConfiguration serverCfg, Bookie bookie,
                 statsLogger);
         this.longPollThreadPool = createExecutor(
                 this.serverCfg.getNumLongPollWorkerThreads(),
-                "BookieLongPollThread",
-                OrderedScheduler.NO_TASK_LIMIT,
-                statsLogger);
+                "BookieLongPollThread-" + serverCfg.getBookiePort(),
+                OrderedScheduler.NO_TASK_LIMIT, statsLogger);
+        this.highPriorityThreadPool = Executors.newCachedThreadPool(
```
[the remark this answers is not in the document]
ivankelly: no idea. I just reused what was there for fencedThreadPool. probably something about newCachedThreadPool being harder to fail, because of unboundedness, but this introduces problems of it's own. Will change to something bounded.

On a side note, with these thread pools, the architecture is starting to look more like something which could benefit from an actor architecture using something like lmax disruptor. Way in the future though, if at all.

ivankelly:
Yahoo uses the V2 protocol, so the change should be compatible (the whole BC test changes are to ensure this, there's a test with the yahoo custom version now).

bookkeeper-proto/src/main/proto/BookkeeperProtocol.proto
```diff
@@ -101,6 +101,7 @@ message ReadRequest {
     optional int64 previousLAC = 4;
     // Used as a timeout (in milliseconds) for the long polling request
     optional int64 timeOut = 5;
+    optional uint32 priority = 6 [default = 0];
```
sijie: Can you make priority as part of PacketHeader? so this would be available to all type of requests.

As what I kind see here, it can be reused by `GetBookieInfoRequest`, `WriteLacRequest` and `ReadLacRequest` potentially.

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
```diff
@@ -127,6 +127,7 @@
     protected static final String MAX_PENDING_READ_REQUESTS_PER_THREAD = "maxPendingReadRequestsPerThread";
     protected static final String MAX_PENDING_ADD_REQUESTS_PER_THREAD = "maxPendingAddRequestsPerThread";
     protected static final String NUM_LONG_POLL_WORKER_THREADS = "numLongPollWorkerThreads";
+    protected static final String NUM_HIGH_PRIORITY_WORKER_THREADS = "numHighPriorityWorkerThreads";
```
sijie: add this to bookkeeper-server/conf/bk_server.conf and site/_data/config/bk_server.yaml

sijie:
@jvrao please review as well since you were in the conversation of this PR.

bookkeeper-proto/src/main/proto/BookkeeperProtocol.proto
ivankelly: sure

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
ivankelly: will add it

ivankelly:
Will merge this tomorrow if there are no more comment (cc: @jvrao)

PR #899, opened 2017-12-20
BookKeeperBuilderTest has recovery backwards

ivankelly:
NoRecovery test was specifying withRecovery(true).
Recovery test was specifying withRecovery(false).

This patch swaps them to how they should be.


ivankelly:
I think the fact that we got it wrong in the very first tests we wrote for it highlights the fact that these API calls are not very clear in their meaning.

Issue #904, opened 2017-12-21
Support to use short hostname(instead of FQDN) when UseHostForBookieId=true is set 


reddycharan:
**FEATURE REQUEST**

1. Please describe the feature you are requesting.

- Support to use short hostname(instead of FQDN) when UseHostForBookieId=true is set
- provide useShortHostName config to use short hostName for bookieId
- print appropriate string representation, both hostname & ip for bkshell cmds

2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have). Are you currently using any workarounds to address this issue?

must-have for one of our usecase scenario

PR #905, opened 2017-12-21
ISSUE #904: short hostname for bookieid

reddycharan:

- provide useShortHostName config to use short hostName for bookieId
- corresponding testcases
- For BKShell commands print appropriate string representation (hostname and ipaddress)
and port number of bookies

Master Issue: #904

> ---
> quickly and easily:
>
>
>
> Otherwise:
> 
> 
> ---

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
@@ -23,6 +23,7 @@ import static com.google.common.base.Charsets.UTF_8;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
+import com.google.common.net.InetAddresses;
 import com.google.common.util.concurrent.AbstractFuture;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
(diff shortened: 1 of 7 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
```diff
@@ -1244,8 +1244,6 @@ Options getOptions() {
             super(CMD_LISTBOOKIES);
             opts.addOption("rw", "readwrite", false, "Print readwrite bookies");
             opts.addOption("ro", "readonly", false, "Print readonly bookies");
-            opts.addOption("h", "hostnames", false,
```
sijie: I think it is not a good idea to remove an option directly. I will suggest marking this as deprecated and remove this in 4.8.

reddycharan: ok..I'll add it back but it will effectively be ignored. BTW I dont see any particular format for adding deprecated tag for command line option. I'll just add following comment for the 'hostnames' option -

"//@deprecated 'rw'/'ro' option prints both hostname and ip, so this option is not needed anymore"

sijie: oh there is not particular format. I was suggesting adding this to the description

"Also print hostname of the bookie" =>

"Also print hostname of the bookie (deprecated, 'rw'/'ro' option print both hostname and ip)"

sijie:
added this pull request to #914 so we remember to write documentation for this.

PR #912, opened 2017-12-22
Journal: Fix journal write queue so it decrements properly.

athanatos:
(@bug W-3169683@)
Signed-off-by: Dustin <dcastor@salesforce.com>

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
@@ -440,7 +440,6 @@ public class Journal extends BookieCriticalThread implements CheckpointSource {
                 ForceWriteRequest req = null;
                 try {
                     req = forceWriteRequests.take();
-
                     // Force write the file and then notify the write completions
                     //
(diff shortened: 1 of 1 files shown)


eolivelli:
@sijie I know this is a backport from SF fork. I am ok with this patch.

IMHO this patch needed at least one a +1 from at least two committers before merging.

What about adding a test case so that we are sure that we do not fall into regressions in the future ?
Maybe I am missing some piece of the story

sijie:
@eolivelli the pull requests that I merged are very trivial changes, either it is related to logging or stats. None of these changes related to logic or correctness. For these logging or stats fixes, I don't think it is worth holding them up. For other changes, I am following the procedure with two +1 from committers from merging.

sijie:
@eolivelli also this change is a stat fix.

eolivelli:
@sijie makes sense to me and I agree.

From the diff it seemed to me the this change moved the execution of ForceWriteRequest out of a branch and I was afraid that this can affect how the thread is working. So it seemed to me not just a trivial change like the other ones you recently committed

sijie:
we might consider cherry-pick this change to 4.6.

PR #913, opened 2017-12-22
Refactor FileInfo locking and refcounting out IndexPersistenceMgr

ivankelly:
There is a number of bugs in index persistence mgr related to how
refcounts are handled and how FileInfos are freed, due to the locking
mechanism. These can result in the fencing bit being lost in ledgers,
which is a serious issue. They also cause flakes in
IndexPersistenceMgrTest#testEvictFileInfoWhenUnderlyingFileExists.

For details see the discussion at:
https://github.com/apache/bookkeeper/pull/513/files/8075b0#r156676238

There are two key problems.

1. FileInfos are flushed asynchronously on eviction. If another thread
tries to read the FileInfo after the flush has been scheduled, but
before it runs, it will read stale date (and we'll possibly lose the
fence bit).
2. FileInfos can be closed while it is still references in the
IndexPersistenceMgr. This means that it can be fenced, but this will
very be flushed to disk because the FileInfo is already closed.

The patch solves this by moving FileInfo locking and refcount into a
separate class, FileInfoBackingCache. When either the write cache or
read cache load a file info, it tries to load it from this backing
cache. All writes to the the backing cache are done under a write
lock. This class also takes care of reference counting. It hands out
CachedFileInfo objects, which will be flushed to disk when all
references are released.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileInfoBackingCache.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileInfoBackingCache.java
@@ -1,0 +1,114 @@
+/**
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+package org.apache.bookkeeper.bookie;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+class FileInfoBackingCache {
+    final ReentrantReadWriteLock lock = new ReentrantReadWriteLock();
+    final ConcurrentHashMap<Long, CachedFileInfo> fileInfos = new ConcurrentHashMap<>();
+    final FileLoader fileLoader;
+
+    FileInfoBackingCache(FileLoader fileLoader) {
+        this.fileLoader = fileLoader;
+    }
+
+    CachedFileInfo loadFileInfo(long ledgerId, byte[] masterKey) throws IOException {
+        lock.readLock().lock();
+        try {
+            CachedFileInfo fi = fileInfos.get(ledgerId);
+            if (fi != null) {
+                fi.retain();  // caller of loadFileInfo owns this reference
+                return fi;
+            }
+        } finally {
+            lock.readLock().unlock();
+        }
+
+        // else FileInfo not found, create it under write lock
+        lock.writeLock().lock();
+        try {
+            File backingFile = fileLoader.load(ledgerId, masterKey != null);
+            CachedFileInfo fi = new CachedFileInfo(ledgerId, backingFile, masterKey);
+            fileInfos.put(ledgerId, fi);
+
+            fi.retain(); // caller of loadFileInfo owns this reference
+            return fi;
+        } finally {
+            lock.writeLock().unlock();
+        }
+    }
+
+    void releaseFileInfo(long ledgerId, FileInfo fileInfo) {
+        lock.writeLock().lock();
+        try {
+            fileInfo.close(true);
+            fileInfos.remove(ledgerId, fileInfo);
+        } catch (IOException ioe) {
+            log.error("Error evicting file info({}) for ledger {} from backing cache",
+                      fileInfo, ledgerId, ioe);
+        } finally {
+            lock.writeLock().unlock();
+        }
+    }
+
+    void closeAllWithoutFlushing() throws IOException {
+        for (Map.Entry<Long, CachedFileInfo> entry : fileInfos.entrySet()) {
+            entry.getValue().close(false);
+        }
+    }
+
+    class CachedFileInfo extends FileInfo {
+        final long ledgerId;
+        final AtomicInteger refCount;
+
+        CachedFileInfo(long ledgerId, File lf, byte[] masterKey) throws IOException {
+            super(lf, masterKey);
+            this.ledgerId = ledgerId;
+            this.refCount = new AtomicInteger(0);
+        }
+
+        void retain() {
+            refCount.incrementAndGet();
+        }
+
+        int getRefCount() {
+            return refCount.get();
+        }
+
+        void release() {
+            if (refCount.decrementAndGet() == 0) {
+                releaseFileInfo(ledgerId, this);
+            }
+        }
+    }
+
+    interface FileLoader {
+        File load(long ledgerId, boolean createIfMissing) throws IOException;
+    }
+}
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileInfoBackingCache.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileInfoBackingCache.java
@@ -64,7 +64,7 @@ class FileInfoBackingCache {
         }
     }
 
-    void releaseFileInfo(long ledgerId, FileInfo fileInfo) {
+    private void releaseFileInfo(long ledgerId, FileInfo fileInfo) {
         lock.writeLock().lock();
         try {
             fileInfo.close(true);
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileInfoBackingCache.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileInfoBackingCache.java
@@ -120,7 +120,7 @@ class FileInfoBackingCache {
          * When a client obtains a fileinfo from a container object,
          * but that container object may release the fileinfo before
          * the client has a chance to call retain. In this case, the
-         * file info could be releases and the destroyed before we ever
+         * file info could be released and the destroyed before we ever
          * get a chance to use it.
          *
          * <p>tryRetain avoids this problem, by doing a compare-and-swap on
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileInfoBackingCache.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileInfoBackingCache.java
@@ -64,11 +64,22 @@ class FileInfoBackingCache {
         }
     }
 
-    private void releaseFileInfo(long ledgerId, FileInfo fileInfo) {
+    private void releaseFileInfo(long ledgerId, CachedFileInfo fileInfo) {
         lock.writeLock().lock();
         try {
-            fileInfo.close(true);
-            fileInfos.remove(ledgerId, fileInfo);
+            /* RefCount may have been incremented between the call
+             * to decrementAndGet in CachedFileInfo#release and getting
+             * to this point in the code, so check it again.
+             * If the refCount for a fileInfo is 0, it means that the only
+             * object referencing it is the fileInfos map. Acquiring the
+             * fileInfo from the map is done under the lock, so if the
+             * refcount is 0 here, we'll be able to remove it from the
+             * map before anyone has a chance to increment it.
+             */
+            if (fileInfo.getRefCount() == 0) {
+                fileInfo.close(true);
+                fileInfos.remove(ledgerId, fileInfo);
+            }
         } catch (IOException ioe) {
             log.error("Error evicting file info({}) for ledger {} from backing cache",
                       fileInfo, ledgerId, ioe);
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileInfoBackingCache.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileInfoBackingCache.java
@@ -30,6 +30,8 @@ import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
 class FileInfoBackingCache {
+    static final int DEAD_REF = -0xdead;
+
     final ReentrantReadWriteLock lock = new ReentrantReadWriteLock();
     final ConcurrentHashMap<Long, CachedFileInfo> fileInfos = new ConcurrentHashMap<>();
     final FileLoader fileLoader;
@@ -43,7 +45,14 @@ class FileInfoBackingCache {
         try {
             CachedFileInfo fi = fileInfos.get(ledgerId);
             if (fi != null) {
-                fi.retain();  // caller of loadFileInfo owns this reference
+                // tryRetain only fails if #markDead() has been called
+                // on fi. This is only called from within the write lock,
+                // and if it is called (and succeeds) the fi will have been
+                // removed from fileInfos at the same time, so we should not
+                // have been able to get a reference to it here.
+                // The caller of loadFileInfo owns the refence, and is
+                // responsible for calling the corresponding #release().
+                assert(fi.tryRetain());
                 return fi;
             }
         } finally {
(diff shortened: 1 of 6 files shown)


ivankelly:
WIP because some tests are failing. But would be good to get some feedback.

sijie:
> These can result in the fencing bit being lost in ledgers,
which is a serious issue.

I think it is better to make a clarification here: it is not "physically lost". it is on disk, but you might have a concurrent fileinfo loaded into the cache with no fencing bit.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/IndexPersistenceMgr.java
```diff
@@ -80,15 +77,13 @@ public static final String getLedgerName(long ledgerId) {
     }
 
     // use two separate cache for write and read
-    final Cache<Long, FileInfo> writeFileInfoCache;
-    final Cache<Long, FileInfo> readFileInfoCache;
+    final Cache<Long, CachedFileInfo> writeFileInfoCache;
+    final Cache<Long, CachedFileInfo> readFileInfoCache;
+    final FileInfoBackingCache fileInfoBackingCache;
```
sijie: The change here is same as the original change I made using guava cache at Twitter. I did it using two guava cache with one backing concurrent map. However when I left twitter, Yiming found some bugs using this approach, which it made him remove the backing concurrent hashmap and use guava cache only.

ivankelly:
> I think it is better to make a clarification here: it is not "physically lost". it is on disk, but you might have a concurrent fileinfo loaded into the cache with no fencing bit.

It's no even on disk in the second refcounting case. And in the original case, it could be overwritten again when the FileInfo is evicted.

sijie:
I don't think it will never be overwritten. Because, header will not be flushed if there is no modification. If I remember the code correctly.

ivankelly:
Ya, i checked the code for the second case again. The only reason to flush the header is fence bit being set, so that one won't happen now in reality, but it could happen if there's more things in the header (perhaps writer id in a paxos impl for example).

ivankelly:
Tests pass now, and I added a new test for the backing cache

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/IndexPersistenceMgr.java
```diff
@@ -174,90 +170,57 @@ public Number getSample() {
...
-                    }
-                } catch (IOException e) {
-                    LOG.error("Exception closing file info when ledger {} is evicted from file info cache.",
-                        ledgerId, e);
-                } finally {
-                    fileInfoLock.writeLock().unlock();
-                }
-            });
         }
         fileInfo.release();
```
yzang: Here handleEviction is not synchronized with getFileInfo which will cause race condition.

Race Condition:
1. Thread A just successfully get a FileInfo from cache but haven't increment the refCount yet at line 221 "fi = readFileInfoCache.get(ledger, loader);"
2. Thread B is trying to get another FileInfo for read and cache is full, so it triggers eviction to evict the fileInfo which is referenced by Thread A.
3. Thread B continues to evict FileInfo and find out the refCount=0, so it close the FileInfo.
4. Thread A now returns the FileInfo which is already closed by Thread B.
5. java.nio.channels.ClosedChannelException will be thrown and cause serious issues

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/IndexPersistenceMgr.java
```diff
@@ -354,8 +316,7 @@ private void getActiveLedgers() throws IOException {
      */
     void removeLedger(Long ledgerId) throws IOException {
         // Delete the ledger's index file and close the FileInfo
-        FileInfo fi = null;
-        fileInfoLock.writeLock().lock();
```
yzang: Remove ledger is not synchronized with getFileInfo, so it will introduce race conditions.

Race Condition:
1. Thread A is on line 221 and get a FileInfo from cache successfully, but haven't increment the refCount.
2. Thread B is on line 335 and after it invalidate the FileInfo from cache, refCount will becomes 0 so the underline FileChannel will be closed.
3. Now Thread A continues to increment the refCount, and returns the closed FileInfo to application.
4. java.nio.channels.ClosedChannelException will be thrown when we starts to use this FileInfo.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/IndexPersistenceMgr.java
yzang: Yeah, this will cause issues based on the assumption that we put fileInfo into both write cache and read cache when it's first created for write. Please see my above comments for details.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileInfoBackingCache.java
```diff
@@ -0,0 +1,125 @@
...
+    FileInfoBackingCache(FileLoader fileLoader) {
+        this.fileLoader = fileLoader;
+    }
+
+    CachedFileInfo loadFileInfo(long ledgerId, byte[] masterKey) throws IOException {
+        lock.readLock().lock();
+        try {
+            CachedFileInfo fi = fileInfos.get(ledgerId);
+            if (fi != null) {
+                fi.retain();  // caller of loadFileInfo owns this reference
```
yzang: If we use another ConcurrentHashMap to store FileInfo, it will cause issues. This is based on the assumption that you fix the comment for line 219 in IndexPersistenceMgr.java, because we need to put FileInfo into both read cache and write cache when it's first created for write.

Now let's talk about the issue:
1. We created a new ledger, so we create a new FileInfo and put it into both read cache and write cache. 
2. However, here since it's already in the hashmap, we only increment the refCount by one although we store the ref in 2 different caches.
3. Now the fileInfo is evicted from write cache and refCount becomes 0, so we decide to close the file.
4. After that, we try to get that fileInfo for read, and we can find it from cache, however, when we get the file from cache, the file is already closed.

ivankelly: Each time you load it into one of the caches you will increment the refcount (in loadFileInfo)

yzang:
@ivankelly Could you also explain a little bit about how the below thing could happen?

> 2. FileInfos can be closed while it is still references in the
> IndexPersistenceMgr. This means that it can be fenced, but this will
> very be flushed to disk because the FileInfo is already closed.

The only place where we close FileInfo without flushing header is when we remove the ledger and when we shutdown bookkeeper.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/IndexPersistenceMgr.java
ivankelly: hmm, yes, another race. unlikely to happen because the get from 1., will put it to the top of the access queue, so it's unlikely to be expired, but still not acceptable. really we need the retain to occur while the guava cache segment is still locked internally. There's no easy way to do this though.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/IndexPersistenceMgr.java
```diff
@@ -174,90 +170,57 @@ public Number getSample() {
...
+                activeLedgers.put(ledger, true);
+                return fileInfo;
+            };
             if (null != masterKey) {
-                fi = writeFileInfoCache.get(ledger,
-                    new FileInfoLoader(ledger, masterKey));
-                if (null == readFileInfoCache.asMap().putIfAbsent(ledger, fi)) {
-                    fi.use();
-                }
+                fi = writeFileInfoCache.get(ledger, loader);
```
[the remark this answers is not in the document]
ivankelly: It does increment refCount by 2. The slowness of reading a fileinfo is due to disk I/O, which the concurrent hash map mitigates.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileInfoBackingCache.java
yzang: You wouldn't if the fileInfo is already in the map, right?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/IndexPersistenceMgr.java
ivankelly: This is same race as https://github.com/apache/bookkeeper/pull/913#discussion_r158849128, no?

ivankelly:
@yzang this race is demonstracted by the following test case.
https://github.com/ivankelly/bookkeeper/commit/15609e571c5e7c166544ed2ca54f1d214518e7ff

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileInfoBackingCache.java
ivankelly: loading into readFileInfoCache and writeFileInfoCache will each invoke a separate call to loadFileInfo, so retain will be called twice.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/IndexPersistenceMgr.java
yzang: This is the same extreme situation that could happen in https://github.com/apache/bookkeeper/pull/513, if you think it's unlikely to happen then we can ignore it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/IndexPersistenceMgr.java
yzang: Yeah, similar but these race conditions caused a lot of issues to our production cluster under high load, we've been testing these for months.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileInfoBackingCache.java
yzang: @ivankelly Could you point me to the cod where the retain would be called twice?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/IndexPersistenceMgr.java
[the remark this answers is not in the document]
yzang: Why would it increment refCount by 2? The function who call getFileInfo will release the fileInfo immediately in the finally block.

yzang:
@ivankelly If the fence and release is the only race condition issue here, why not fix the fence logic directly? For example, when we call release and decide to close the file, we also check if we need to flush the header or not. Or can we flush the fence bit header right in the setFence function?

ivankelly:
there are two race conditions here. the original one I pointed out, and the "fence and release" one.

> For example, when we call release and decide to close the file, we also check if we need to flush the header or not.
> Or can we flush the fence bit header right in the setFence function?

This may not be a bad idea, but there would still be races. If the FileInfo which setFenced is called on has been evicted just before setFenced is called, and loaded in again, then the FileInfo for that ledger would not have the fenced bit, even though it exists on disk.

We can get around the race you highlighted on this PR by doing a check on the refcount when we do a retain. 

On eviction from the backing cache, we do a compareAndSet(0, -0xdeadbeef). On getting from the cache, retain does a increment and get. If the refcount is positive, it will block it being deleted from the backing cache. This doesn't fill me with confidence though.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileInfoBackingCache.java
ivankelly: readFileInfoCache.get(ledgerId, loader) calls loadFileInfo if ledgerId isn't the cache. this is one call to retain()
writeFileInfoCache.get(ledgerId, loader) calls loadFileInfo if ledgerId isn't in that cache. This is the second call to retain.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/IndexPersistenceMgr.java
ivankelly: I don't think these can be ignored. As I've said previously, a guarantee is a guarantee.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/IndexPersistenceMgr.java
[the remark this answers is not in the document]
ivankelly: I meant increment to 2, if it's in both write and read caches.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/IndexPersistenceMgr.java
ivankelly: Aside from the race you identified, what is the actual issue with this?

ivankelly:
@yzang @sijie 

Pushed a change with the compare and set approach mentioned in https://github.com/apache/bookkeeper/pull/913#issuecomment-354166855.

With this change, retaining and releasing a fileinfo uses a lock free type mechanism, while inserting and deleting from the cache uses a ReadWriteLock. 

Another option would be to dig into guava cache itself to increase the refcount when retrieving, but I'm not even sure that's possible, because guava doesn't use locks in the read path (most of the time).

ivankelly:
retest this please

sijie:
Ping @yzang - can you review @ivankelly 's latest change?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileInfoBackingCache.java
```diff
@@ -104,8 +105,42 @@ void closeAllWithoutFlushing() throws IOException {
...
+        private boolean markDead() {
+            return refCount.compareAndSet(0, DEAD_REF);
+        }
+
+        /**
+         * Attempt to retain the file info.
+         * When a client obtains a fileinfo from a container object,
+         * but that container object may release the fileinfo before
+         * the client has a chance to call retain. In this case, the
+         * file info could be releases and the destroyed before we ever
```
yzang: NIT: could be "released"?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/IndexPersistenceMgr.java
yzang: @ivankelly The previous issue at Twitter is mostly because refCount isn't increased when we get a FileInfo from concurrentHashMap, which I believe should be already addressed in your change.

yzang: @ivankelly The previous issue with concurrent hashmap in Twitter is because the refCount is not correctly incremented when we get the FileInfo from the map. I think that issue is already addressed in your change.

yzang:
@ivankelly After taking a look at your latest change, I still have some questions. Could you help me understand how your latest change resolve the race condition?

Based on my understanding, the following race condition could still happen:
1. Current status: FileInfo f is cached in both write cache and read cache, so the refCount=2.
2. Somehow FileInfo f is evicted from write cache, now refCount = 1.
3. Thread A call getFileInfo() for read, so it get f from read cache at line 221, it hasn't been able to do tryRetain() yet. Now refCount = 1
4. Thread B at the same time is evicting f from read cache, in handleLedgerEviction it will call fi.release, in the release() function "if (refCount.decrementAndGet() == 0) {" will decrement the refCount, so now refCount = 0, and "releaseFileInfo(ledgerId, this)" will be called. 
5. Thread B call fileInfo.markDead(), since refCount ==0 is true, it returns true. So now thread B close the FileInfo f.
6. Thread A continues, it reach tryRetain with refCount =0, and successfully return true.
7. Thread A returns the FileInfo which is already closed.

ivankelly:
@yzang thanks for looking again. It's really good to have someone thinking deeply about this stuff to shake out the issues.

Regarding the proposed race, in step 5, markDead will compareAndSet on the refCount. If refCount is 0, it sets to -0xDEAD. In step 6, when thread A calls tryRetain, it will see refCount is less than 0, since -0xDEAD is negative, and so return false.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileInfoBackingCache.java
ivankelly: done

yzang:
@ivankelly I see what you mean, after a double check, I think the fix LGTM. One last thing is since we add a refCount layer above FileInfo, shall we completely clean up the useCount logic in FileInfo.java? Otherwise will it make people a little bit confused?

ivankelly:
retest this please

ivankelly:
retest this please

sijie:
retest this please

ivankelly:
java 9 job completed with SUCCESS, but for some reason it's getting set to FAILURE :/

sijie:
@ivankelly it probably was timed out.

PR #916, opened 2017-12-23
Issue 915: Part 1 - Copy pulsar-checksum module as circe-checksum

sijie:

- this change copies the files under `com.scurrilous.circe` in pulsar-checksum module into bookkeeper as `circe-checksum`.
- this change doesn't copy files under `org.apache.pulsar` except NativeUtils. will have a subsequent change to add it.
- Move NativeUtils to `com.scurrilous.circe.utils`.
- Update notice files to respect to Trevor Robinson's change


sijie:
@merlimat : I am not sure how much code that pulsar has modified circe's code. so I only include copyright to Trevor Robinson. If there is additional notice that I need to include for this module, let me know, I will update.

sijie:
The RAT failure is actually fixed in #917 here. so #916 and #917 should probably be reviewed and merged at the same time.

eolivelli:
I suggest not using package name we don't 'own'.
We should at least prepend org.apache.bookkeeper.
This has always been a good principle and when we will step into a modularized world we won't hit issues.

eolivelli:
Another approach would be to handle this module externally from BookKeeper and not to import it as source code

sijie:
1) maintain and setup an external module required a lot of infrastructure wise work., E.g setting up to CI and publishing jars. I don't think we can do that. 2) maintain and setup an external module is effectively same as just using pulsar-checksum. The original author doesn't have a published maven artifacts, that was the reason that pulsar imports the code. We can't import pulsar-checksum directly, because of the recursive dependency will make things a bit hard to maintain if pulsar uses the official bk.


Regarding modularize, I assume we only consider modularize when we bump the minimal Java support to Java 9. If we bump the minimal Java support to 9, then we should consider just using jdk crc32c. So I don't think this is a problem for modularize in future. 

But since this module is imported from pulsar, I would defer this to Matteo and see what is best solution for both bookkeeper and pulsar.

eolivelli:
I am fine with keeping this module as you are proposing, if it will help pulsar project expecially

Issue #915, opened 2017-12-23
Import pulsar-checksum (circe checksum) module


sijie:
**FEATURE REQUEST**

1. Please describe the feature you are requesting.

In #856, we are introducing CRC32C as the new digest type. It is using `pulsar-checksum`. To avoid recursive dependency, it makes sense to push this module down to bookkeeper itself.

2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have). Are you currently using any workarounds to address this issue?

*must-have* if #856 is accepted.

3. Provide any additional detail on your proposed use case for this feature.

N/A

PR #917, opened 2017-12-23
Issue 915: Part 2 - Add crc32c integer/long checksum utils under `com.scurrilous.circe.checksum`

sijie:

This is the part 2 change of #915. It is based on #916, adding crc32c int/long checksum util classes under `com.scurrilous.circe.checksum`. It also fixes a incremental checksum issue on long crc32c. I added bunch of test cases to ensure the both long/int crc32 checksum is correct on incremental cases.

a7a26e1  contains the changes to be reviewed in this pull request.


sijie:
@merlimat : 

I didn't add the util functions in `org.apache.bookkeeper` or `org.apache.pulsar` package, because if we upgrade the minimal support of java to jdk9 in future, we can just simply remove this package without touching other modules. so I put the until functions into `com.scurrilous.circe.checksum`. Is that okay with you?

sijie:
Also this module is expected to be fail with jdk9, because it is using a pretty old version of jmockit. will have a subsequent change to address that.

merlimat:
> so I put the until functions into com.scurrilous.circe.checksum. Is that okay with you? 

Sure, sounds good.

circe-checksum/pom.xml
```diff
@@ -48,6 +48,12 @@
       <version>${guava.version}</version>
     </dependency>
 
+    <dependency>
+      <groupId>io.netty</groupId>
+      <artifactId>netty-buffer</artifactId>
```
merlimat: For this artifact we need to be careful. Sure, the checksum only uses the `ByteBuf`, though in the other components we're already using the `netty-all` dependency which already includes the `netty-buffer` but maven will pull in both since it doesn't know it.

sijie: @merlimat I exclude `netty-buffer` in bookkeeper-server when switching it from `pulsar-checksum` to use this imported module. (see the last commit in #937 )

PR #921, opened 2017-12-27
Add descriptive message for TooManyRequestsException

eolivelli:
After upgrading from 4.5 to 4.6 I found a lot of "Unexpected condition" in some benchmarks.

TooManyRequestsException is a new error introduced in 4.6, there is not description, so the default description is "Unexpected condition", without a debugger it is not possible to understand the cause of the error.

### DIFF
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/api/BKExceptionTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/api/BKExceptionTest.java
@@ -37,9 +37,9 @@ public class BKExceptionTest {
                 int code = f.getInt(null);
                 String msg = BKException.getMessage(code);
                 if (code == UnexpectedConditionException) {
-                    assertEquals(msg, "Unexpected condition");
+                    assertEquals("Unexpected condition", msg);
                 } else {
-                    assertNotEquals("failure on code " + f.getName(), msg, "Unexpected condition");
+                    assertNotEquals("failure on code " + f.getName(), "Unexpected condition", msg);
                 }
                 count++;
             }
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/api/BKExceptionTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/api/BKExceptionTest.java
@@ -17,6 +17,7 @@ package org.apache.bookkeeper.client.api;
 
 import static org.apache.bookkeeper.client.api.BKException.Code.UnexpectedConditionException;
 import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.greaterThan;
 import static org.hamcrest.Matchers.not;
 import static org.junit.Assert.assertThat;
 
@@ -31,6 +32,7 @@ public class BKExceptionTest {
     @Test
     public void testGetMessage() throws Exception {
         Field[] fields = BKException.Code.class.getFields();
+        int count = 0;
         for (Field f : fields) {
             if (f.getType() == Integer.TYPE && !f.getName().equals("UNINITIALIZED")) {
                 int code = f.getInt(null);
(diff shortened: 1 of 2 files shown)


eolivelli:
@sijie I think this should go to 4.6 branch, I got into this after upgrading.
Maybe it would be worth to add a note in the "release notes" of 4.6

//cc @jiazhai

sijie:
+1 for 4.6.1

sijie:
can you add a test case for this, to prevent "we add a new code but forget to add descriptions for it"?

eolivelli:
@sijie sure will do as soon as I have time

eolivelli:
@sijie added a test case, please review I would like to merge it soon, thanks

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/api/BKExceptionTest.java
```diff
@@ -0,0 +1,52 @@
...
+    @Test
+    public void testGetMessage() throws Exception {
+        Field[] fields = BKException.Code.class.getFields();
+        int count = 0;
+        for (Field f : fields) {
+            if (f.getType() == Integer.TYPE && !f.getName().equals("UNINITIALIZED")) {
+                int code = f.getInt(null);
+                String msg = BKException.getMessage(code);
+                if (code == UnexpectedConditionException) {
+                    assertThat(msg, equalTo("Unexpected condition"));
```
sijie: any reason why not use "assertEquals"?

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/api/BKExceptionTest.java
```diff
@@ -0,0 +1,52 @@
...
+                } else {
+                    assertThat("failure on code " + f.getName(), msg,
```
sijie: any reason why not use `assertNotEquals`?

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/api/BKExceptionTest.java
```diff
@@ -0,0 +1,52 @@
...
+                               not(equalTo("Unexpected condition")));
+                }
+                count++;
+            }
+        }
+        // assert that we found at least 1 code other than UnexpectedConditionException
+        assertThat(count, greaterThan(2));
```
sijie: any reason not use assertTrue(count > 2)?

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/api/BKExceptionTest.java
```diff
@@ -0,0 +1,52 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.bookkeeper.client.api;
+
+import static org.apache.bookkeeper.client.api.BKException.Code.UnexpectedConditionException;
+import static org.hamcrest.Matchers.equalTo;
```
sijie: any reason why not use junit Asserts? does it not meet your requirements?

eolivelli: @sijie 
I saw that in another commit we started to use *assertThat*, recently I started to use it even for "equals", usually error messages generated automatically are more readable (like the one for greaterThan)
I have changed this, now using basic assertEquals/assertNotEquals

eolivelli:
@sijie addressed your comments

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/api/BKExceptionTest.java
```diff
@@ -38,15 +37,14 @@ public void testGetMessage() throws Exception {
                 int code = f.getInt(null);
                 String msg = BKException.getMessage(code);
                 if (code == UnexpectedConditionException) {
-                    assertThat(msg, equalTo("Unexpected condition"));
+                    assertEquals(msg, "Unexpected condition");
```
sijie: nit: expected message should be the first parameter: assertEquals("Unexpected condition", msg);

Issue #922, opened 2017-12-27
Bookie and BookKeeperAdmin and BookKeeper client does not use shaded Protobuf JAR


eolivelli:
**BUG REPORT**

1. Please describe the issue you observed:

Bookie depends on non-shaded Google Protobuf 3.4.0

- What did you do?
Start a Bookie (embedded) without Protobuf on the classpath

- What did you expect to see?
The Bookie should start, because we are shading and relocating Protobuf

- What did you see instead?

This is the error on BookKeeper client:
```
java.lang.NoClassDefFoundError: com/google/protobuf/GeneratedMessageV3
        at java.lang.ClassLoader.defineClass1(Native Method)
        at java.lang.ClassLoader.defineClass(ClassLoader.java:763)
        at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
        at java.net.URLClassLoader.defineClass(URLClassLoader.java:467)
        at java.net.URLClassLoader.access$100(URLClassLoader.java:73)
        at java.net.URLClassLoader$1.run(URLClassLoader.java:368)
        at java.net.URLClassLoader$1.run(URLClassLoader.java:362)
        at java.security.AccessController.doPrivileged(Native Method)
        at java.net.URLClassLoader.findClass(URLClassLoader.java:361)
        at magnews.server.ServiceClassLoader.findClass(ServiceClassLoader.java:105)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
        at org.apache.bookkeeper.client.LedgerMetadata.parseConfig(LedgerMetadata.java:459)
        at org.apache.bookkeeper.meta.AbstractZkLedgerManager$3.processResult(AbstractZkLedgerManager.java:389)
```

This is the error on BookKeeperAdmin:
```

Caused by: java.lang.NoClassDefFoundError: com/google/protobuf/ExtensionRegistry
	at org.apache.bookkeeper.proto.BookieClient.<init>(BookieClient.java:108)
	at org.apache.bookkeeper.client.BookKeeper.<init>(BookKeeper.java:492)
	at org.apache.bookkeeper.client.BookKeeper.<init>(BookKeeper.java:365)
	at org.apache.bookkeeper.client.BookKeeperAdmin.format(BookKeeperAdmin.java:1181)
```

This is the error on Bookie:

```
[...]

PR #923, opened 2017-12-27
Issue-922 Fix shade plugin configuration

eolivelli:
Apply the shade plugin to the bookkeeper-proto module, see #922

### DIFF
--- a/bookkeeper-proto/pom.xml
+++ b/bookkeeper-proto/pom.xml
@@ -88,7 +88,6 @@
               <shadeTestJar>true</shadeTestJar>
               <artifactSet>
                 <includes>
-                  <include>org.apache.bookkeeper:bookkeeper-proto</include>
                   <include>com.google.protobuf:protobuf-java</include>
                 </includes>
               </artifactSet>
--- a/bookkeeper-proto/pom.xml
+++ b/bookkeeper-proto/pom.xml
@@ -66,41 +66,6 @@
           </execution>
         </executions>
       </plugin>
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-shade-plugin</artifactId>
-        <version>${maven-shade-plugin.version}</version>
-        <configuration>
-          <!-- put your configurations here -->
-        </configuration>
-        <executions>
-          <execution>
-            <phase>package</phase>
-            <goals>
-              <goal>shade</goal>
-            </goals>
-            <configuration>
-              <createDependencyReducedPom>true</createDependencyReducedPom>
-              <promoteTransitiveDependencies>true</promoteTransitiveDependencies>
-              <shadedArtifactAttached>true</shadedArtifactAttached>
-              <shadedClassifierName>shaded</shadedClassifierName>
-              <minimizeJar>true</minimizeJar>
-              <shadeTestJar>true</shadeTestJar>
-              <artifactSet>
-                <includes>
-                  <include>com.google.protobuf:protobuf-java</include>
-                </includes>
-              </artifactSet>
-              <relocations>
-                <relocation>
-                  <pattern>com.google.protobuf</pattern>
-                  <shadedPattern>bk-shade.com.google.proto_${protobuf.version}</shadedPattern>
-                </relocation>
-              </relocations>
-            </configuration>
-          </execution>
-        </executions>
-      </plugin>
     </plugins>
   </build>
 </project>
--- a/bookkeeper-proto/pom.xml
+++ b/bookkeeper-proto/pom.xml
@@ -66,6 +66,42 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-shade-plugin</artifactId>
+        <version>${maven-shade-plugin.version}</version>
+        <configuration>
+          <!-- put your configurations here -->
+        </configuration>
+        <executions>
+          <execution>
+            <phase>package</phase>
+            <goals>
+              <goal>shade</goal>
+            </goals>
+            <configuration>
+              <createDependencyReducedPom>true</createDependencyReducedPom>
+              <promoteTransitiveDependencies>true</promoteTransitiveDependencies>
+              <shadedArtifactAttached>true</shadedArtifactAttached>
+              <shadedClassifierName>shaded</shadedClassifierName>
+              <minimizeJar>true</minimizeJar>
+              <shadeTestJar>true</shadeTestJar>
+              <artifactSet>
+                <includes>
+                  <include>org.apache.bookkeeper:bookkeeper-proto</include>
+                  <include>com.google.protobuf:protobuf-java</include>
+                </includes>
+              </artifactSet>
+              <relocations>
+                <relocation>
+                  <pattern>com.google.protobuf</pattern>
+                  <shadedPattern>bk-shade.com.google.proto_${protobuf.version}</shadedPattern>
+                </relocation>
+              </relocations>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
     </plugins>
   </build>
 </project>
--- a/bookkeeper-server/pom.xml
+++ b/bookkeeper-server/pom.xml
@@ -317,6 +317,12 @@
               <directory>${project.libdir}</directory>
               <followSymlinks>false</followSymlinks>
             </fileset>
+            <fileset>
+              <directory>${project.basedir}</directory>
+              <includes>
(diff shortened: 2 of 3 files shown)


eolivelli:
@sijie this is blocker for upgrades from 4.5 to 4.6

I am really sorry I did not test 4.6 against this kind of problems, but they have been reported only by downstream integration tests (out of my scope actually).

I will work add to all of downstream integration tests in my company to run them against SNAPSHOTS and RC candidates (it will take time)

bookkeeper-proto/pom.xml
```diff
@@ -66,6 +66,42 @@
...
+            <configuration>
+              <createDependencyReducedPom>true</createDependencyReducedPom>
+              <promoteTransitiveDependencies>true</promoteTransitiveDependencies>
+              <shadedArtifactAttached>true</shadedArtifactAttached>
+              <shadedClassifierName>shaded</shadedClassifierName>
+              <minimizeJar>true</minimizeJar>
+              <shadeTestJar>true</shadeTestJar>
+              <artifactSet>
+                <includes>
+                  <include>org.apache.bookkeeper:bookkeeper-proto</include>
```
ivankelly: why you shading bookkeeper-proto ? there should be no conflict here.

eolivelli: ops, it is a cut and paste from bookkeeper-server pom, will drop, thanks

eolivelli:
@ivankelly addressed you comment

eolivelli:
This is not the right approach, the shade plugin must run only on bookkeeper-server, but actually is not working very well.
We need to tweak bookkeeper-common module as it is using guava

sijie:
@eolivelli it would be good to remove any release labels and milestone, if an issue is closed due to 'won't fix'. we don't need to include this pull requests in any releases.

eolivelli:
@sijie @ivankelly @jiazhai all tests are passing on master
can we merge this into master and cherry pick to 4.6 branch ?

this is the new dependency:tree of a client project:
```

org.apache.bookkeeper:bookkeeper-server:jar:4.7.0-SNAPSHOT:compile
   +- org.rocksdb:rocksdbjni:jar:5.8.6:compile
   +- org.slf4j:slf4j-log4j12:jar:1.7.25:compile
   +- org.apache.zookeeper:zookeeper:jar:3.5.3-beta:compile
   +- commons-cli:commons-cli:jar:1.2:compile
   +- commons-codec:commons-codec:jar:1.6:compile
   +- commons-io:commons-io:jar:2.4:compile
   +- org.apache.commons:commons-lang3:jar:3.3.2:compile
   +- org.apache.commons:commons-collections4:jar:4.1:compile
   +- net.java.dev.jna:jna:jar:3.2.7:compile
   +- log4j:log4j:jar:1.2.15:compile
   +- io.netty:netty-all:jar:4.1.12.Final:compile
   +- io.netty:netty-tcnative-boringssl-static:jar:2.0.3.Final:compile
   +- org.apache.bookkeeper.http:http-server:jar:4.7.0-SNAPSHOT:compile
   |  +- com.fasterxml.jackson.core:jackson-core:jar:2.8.9:compile
   |  +- com.fasterxml.jackson.core:jackson-databind:jar:2.8.9:compile
   |  \- com.fasterxml.jackson.core:jackson-annotations:jar:2.8.9:compile
   +- org.slf4j:slf4j-api:jar:1.7.25:compile
   \- commons-configuration:commons-configuration:jar:1.10:compile
      +- commons-lang:commons-lang:jar:2.6:compile
      \- commons-logging:commons-logging:jar:1.1.1:compile
```

sijie:
I am not sure why do you need to remove the classifier. We should provide unshaded jar by default, and if users want to use a shaded jar, he can use a shaded classifier jar. For me the shade doesn't have to enforce by default, we should use classifier. So -1 on removing shaded classifier.

eolivelli:
@sijie I understand your point and I want to do things as best as possible.

But what can we do ?
The artifact with the 'shaded' classifier is wrong and it is not usable.

Option 1: forget about the "shaded" jar, do not distribute it (I don't like it)
Option 2: revert to 4.5 bookkeeper-server artifact style and then think about this problem without hurry for 4.7 (this proposal)

The other theory is that am not able to make the plugin work properly. Maybe this is the case, I will be very happy if someone could pick this issue up and fix the configuration.

sijie:
I introduced this, I will take a look.

eolivelli:
@sijie thank you so much

bookkeeper-server/pom.xml
```diff
@@ -202,19 +202,16 @@
...
         <executions>
           <execution>
+            <id>create-shaded-jar</id>
             <phase>package</phase>
             <goals>
               <goal>shade</goal>
             </goals>
             <configuration>
               <createDependencyReducedPom>true</createDependencyReducedPom>
-              <promoteTransitiveDependencies>true</promoteTransitiveDependencies>
```
jiazhai: with these lines, the shaded jar will not work?

jiazhai: seems we should also config the <relocations> also?

jiazhai: @eolivelli have you tried this before?

jiazhai:
Agree with @sijie , We should provide unshaded jar by default.

bookkeeper-server/pom.xml
eolivelli: No, but actually we don't need to change the package name for our own module.

The problem is about the dependencies declared in the pom if the  shaded artifact.
It seems that the plugin does not publish the reduced pom but the original one

eolivelli: No, but actually we don't need to change the package name for our own module.

The problem is about the dependencies declared in the pom if the  shaded artifact.
It seems that the plugin does not publish the reduced pom but the original one

eolivelli: It is the createDependencyReducedPom that it is not working.
Maybe it is a bug in the plugin. Le'ts see @sijie findings.

I think the base jar should be non-shaded but we should also provide the jar without original Guava and Protobuf. These jar are evil in the Hadoop ecosystem look for "Hbase guava" or "hadoop guava", similar is protobuf

jiazhai: I see, It is createDependencyReducedPom that is required.
Seems other guy have met the same [problem](http://maven.40175.n5.nabble.com/shade-plugin-doesn-t-create-dependency-reduced-pom-td4331135.html).  

In my test, after turn `shadedArtifactAttached` into false, it will create a `./bookkeeper-server/dependency-reduced-pom.xml`, but this change will meet another error
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-shade-plugin:2.4.3:shade (default) on project bookkeeper-server: Error creating shaded jar: Could not resolve following dependencies: [javax.jms:jms:jar:1.1 (compile), com.sun.jdmk:jmxtools:jar:1.2.1 (compile), com.sun.jmx:jmxri:jar:1.2.1 (compile)]: Could not resolve dependencies for project org.apache.bookkeeper:bookkeeper-server:jar:4.7.0-SNAPSHOT: The following artifacts could not be resolved: javax.jms:jms:jar:1.1, com.sun.jdmk:jmxtools:jar:1.2.1, com.sun.jmx:jmxri:jar:1.2.1: Could not transfer artifact javax.jms:jms:jar:1.1 from/to java.net (https://maven-repository.dev.java.net/nonav/repository): Cannot access https://maven-repository.dev.java.net/nonav/repository with type legacy using the available connector factories: BasicRepositoryConnectorFactory: Cannot access https://maven-repository.dev.java.net/nonav/repository with type legacy using the available layout factories: Maven2RepositoryLayoutFactory: Unsupported repository layout legacy -> [Help 1]
```

eolivelli:
@jiazhai that is my same finding, with sadedArtifactAttached the plugin is not working properly

I totally agree with @jiazhai  and @sijie that we have to distribute the plain artifact, but at this moment we are not ready to do so, so I would prefer to stick to old layout and maybe work with Maven guys to fix the plugin, the fix won't be so hard to do, I can work on it.

@sijie did you find other options ?

sijie:
@eolivelli sorry for late response. was stuck with something else today. only get the chance to look at this for now. I don't like the old approach we are using. We only shipped a jar with shaded protobuf and guava. If a project is using bookkeeper and also protobuf and guava, and it is also going to produce a shaded jar, it is going to shade protobuf and guava again. The same set of dependencies are shaded again and again in the dependency chain. At the mimial, we should provide plain artifact and a shaded artifact.

@jiazhai @eolivelli please check #928. I have summarized the approaches there.

jiazhai:
Since #928 merged, would like to close this one.

PR #928, opened 2017-12-29
Issue 922: changes to shading bookkeeper-server jar

sijie:

**Problem:**

`shadedArtifactAttached` was introduced in 4.6.0 to generate plain and shaded artifact for `bookkeeper-server`. Because for projects (e.g. distributedlog) that would use same protobuf and guava version as bookkeeper, they will use the plain artifact (to avoid shading protobuf/guava multiple times). However there are a few problems with that:

1) bookkeeper-common is not included in the shading include set.
2) `shadedArtifactAttached` only generate a shaded artifact (with relocated classes), but doesn't generate a dependency-reduced pom file. so you have to manually exclude relocated dependencies.

**Solutions:**

- for shaded artifact of `bookkeeper-server`, include all the bookkeeper submodules that bookkeeper-server depends on. so the shaded artifact includes all the classes that need to be relocated.
- introduce a `shaded` module for keeping new shaded libraries. Currently `bookkeeper-server-shaded` is a shaded module of `bookkeeper-server`, while `bookkeeper-server-tests-shaded` is a shaded moulde of the tests jar of `bookkeeper-server`.
- update the client api documentation about the instructions on how to use shaded artifact of `bookkeeper-server` and the new shaded module `bookkeeper-server-shaded`.
- add test cases to verify the shaded artifact and the new shaded modules generate the right artifacts.

**This change doesn't address:**

Currently both client and server are bundled in one `bookkeeper-server`. this change will not introduce any shaded client-only jar. that would be deferred until we split `bookkeeper-server` module.

### DIFF
--- a/shaded/bookkeeper-server-tests-shaded/pom.xml
+++ b/shaded/bookkeeper-server-tests-shaded/pom.xml
@@ -1,0 +1,147 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+   Licensed to the Apache Software Foundation (ASF) under one or more
+   contributor license agreements.  See the NOTICE file distributed with
+   this work for additional information regarding copyright ownership.
+   The ASF licenses this file to You under the Apache License, Version 2.0
+   (the "License"); you may not use this file except in compliance with
+   the License.  You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+-->
+<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.apache.bookkeeper</groupId>
+    <artifactId>shaded-parent</artifactId>
+    <version>4.7.0-SNAPSHOT</version>
+    <relativePath>..</relativePath>
+  </parent>
+  <artifactId>bookkeeper-server-tests-shaded</artifactId>
+  <name>Apache BookKeeper :: Shaded :: bookkeeper-server-tests-shaded</name>
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <project.libdir>${basedir}/lib</project.libdir>
+  </properties>
+  <dependencies>
+    <dependency>
+      <groupId>org.apache.bookkeeper</groupId>
+      <artifactId>bookkeeper-server-shaded</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.bookkeeper</groupId>
+      <artifactId>bookkeeper-server</artifactId>
+      <type>test-jar</type>
+      <version>${project.version}</version>
+      <exclusions>
+        <exclusion>
+          <groupId>org.apache.bookkeeper</groupId>
+          <artifactId>bookkeeper-common</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.apache.bookkeeper</groupId>
+          <artifactId>bookkeeper-proto</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.apache.bookkeeper</groupId>
+          <artifactId>bookkeeper-server</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.apache.bookkeeper.stats</groupId>
+          <artifactId>bookkeeper-stats-api</artifactId>
+        </exclusion>
+        <exclusion>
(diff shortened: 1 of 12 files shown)


shaded/bookkeeper-server-tests-shaded/pom.xml
```diff
@@ -0,0 +1,147 @@
...
+    <groupId>org.apache.bookkeeper</groupId>
+    <artifactId>shaded-parent</artifactId>
+    <version>4.7.0-SNAPSHOT</version>
+    <relativePath>..</relativePath>
+  </parent>
+  <artifactId>bookkeeper-server-tests-shaded</artifactId>
+  <name>Apache BookKeeper :: Shaded :: bookkeeper-server-tests-shaded</name>
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <project.libdir>${basedir}/lib</project.libdir>
```
eolivelli: maybe this is not really used (the clean plugin will not find this directory)

sijie:
Also a few other things to call out

1) currently I keep two options open. a shaded artifact on `bookkeeper-server` and a separate artifact `bookkeeper-server-shaded`. If it is going to confuse users, I am open to remove one.

2) `shadeTestJar` can also produce a shaded artifact over the tests jar of `bookkeeper-server`. however shade plugin doesn't install/deploy the shaded tests jar to mvn repository. so if a project wanna to use a shaded bookkeeper-server jar, but actually it can't use the tests jar of bookkeeper-server. the `bookkeeper-server-tests-shaded` is the shaded tests jar of `bookkeeper-server-shaded`, so when `bookkeeper-server-shaded` is used, the corresponding tests jar can be used.

tests/bookkeeper-server-shaded-artifact-test/src/test/java/org/apache/bookkeeper/tests/shaded/BookKeeperServerShadedArtifactTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Test(expected = ClassNotFoundException.class)
+    public void testGuavaIsShaded() throws Exception {
+        Class.forName("com.google.common.cache.Cache");
+    }
+
+    @Test
+    public void testGuavaShadedPath() throws Exception {
+        Class.forName("org.apache.bookkeeper.shaded.com.google.common.cache.Cache");
+        assertTrue(true);
```
eolivelli: nit: remove this line and similar below

shaded/bookkeeper-server-tests-shaded/pom.xml
sijie: good catch. this is a cut-paste from bookkeeper-server. I fix it.

tests/bookkeeper-server-shaded-artifact-test/src/test/java/org/apache/bookkeeper/tests/shaded/BookKeeperServerShadedArtifactTest.java
sijie: I explicitly add this there for asserting no exception is thrown after `Class.forName`

eolivelli:
@sijie I like this approach.

I think we can drop bookkeeper-server:shaded it is not useful and it is very complicated

I don't see much value for the shaded "tests" jar but I am not an user of "tests" jar in downstream project

tests/bookkeeper-server-shaded-artifact-test/src/test/java/org/apache/bookkeeper/tests/shaded/BookKeeperServerShadedArtifactTest.java
eolivelli: any exception thrown by the method will fail the test

but if you think this it is clearer OK for me

sijie:
@eolivelli please check my comment [here](https://github.com/apache/bookkeeper/pull/928#issuecomment-354430211) regarding the tests jar.

eolivelli:
@sijie ok for keeping the shaded tests

shaded/bookkeeper-server-tests-shaded/pom.xml
sijie: done

tests/bookkeeper-server-shaded-artifact-test/src/test/java/org/apache/bookkeeper/tests/shaded/BookKeeperServerShadedArtifactTest.java
sijie: not going to remove them unless people feel strong about this.

eolivelli: ok

eolivelli:
@sijie are you going to drop the bookkeeper-server with classifier "shaded" artifact ?

As soon as you think the patch is ready I will check it out locally and give it a try on real downstream projects.
It is almost new stuff so I think it would be safe to cherry pick to 4.6 branch

sijie:
This change is ready to go. Dropping shaded artifact or not doesn't block you verify. I will keep the shaded artifact for now.

We can't simply cherry pick because of the versions. I will cherry pick, modify and send a separate pull request for branch 4.6. once this one is merged to master.

eolivelli:
@sijie  OK I will checkout this branch and perform testing

jiazhai:
did a cherry-pick for this fix to branch-4.6: https://github.com/apache/bookkeeper/pull/929

Issue #903, opened 2017-12-21
Integration/compatibility tests based on docker


ivankelly:
With the changes coming in from yahoo, we need to ensure that we're not breaking compatibility. The current compatibility tests are horrible to work with. This issue proposes setting up an integration test framework, based on docker, with different versions of bookkeeper running in different JVMs.

This is required to test #898 correctly.

eolivelli:
@ivankelly  @sijie I think the approach based on Arquillian maybe be better, if I understand it correctly. I never used Arquillian but I am afraid that spawing containers could be too complex for this case.
We can use isolated classloaders (and I think Arquillian uses this approach)

Anyway every idea and approach is welcome, I am fine with the docker approach too.
Maybe it would limit tests execution from contributors as it is not always possible to have docker on developer machines (for instance in my company we have nasty rules of this kind)

sijie:
> We can use isolated classloaders (and I think Arquillian uses this approach)

Not sure if Arquillian uses the approach you are suggesting. But If it does, we should just go with Arquillian, no? Reinventing a testing framework is not fun.

ivankelly:
I don't think we should go with a classloader based approach. Classloader
based would allow current jars versions, but that's all. These tests will
be used to test failure cases like network partitions, which we can't test
with junit. Each node needs to appear as its own machine for this, which is
why I'm suggesting docker. The underlying assumption is that it will be
with arquillian too, but I'd like to see how that works too.

I would suspect that these tests will be too heavy to run on developer
machines in any case, and I wouldn't expect a dev to run them before
submitting a patch. Rather, they would be run as part of the pre commit.

eolivelli:
I have some idea for the classloader approch we can try with both and maybe they could be complementary. I think the docker approach is well suited for running bookies but it will be trickier for clients, dually it will be simpler to create tests for the client with the CL approch.
I will create another issue with my proposal

sijie:
@eolivelli you can use this issue here but just send a PR for your proposal. we don't need to have two different issues for tracking changes for same purpose.

ivankelly:
#930

ivankelly:
This has been implemented in all the linked PRs.

PR #930, opened 2018-01-02
[WIP] Issue #903: Bc tests based on docker

ivankelly:
This is a work in progress, but I'd like some feedback/opinions before cleaning it up. Don't do a detailed review, I'd just like feedback on the general approach and whether you think it's worth pursuing. It'll be easier to look at if you go patch by patch.

The basis idea is to generate a proxy server for each client version. Each proxy runs in its own container, and has a completely isolated classpath. The client is compiled against a single version of BK (4.2.0), so if we have a binary break, we should catch it by just using the proxy. This should be tested separately though. The proxy server/client is GRpc based.

I've converted one test to show what it looks like in usage. Really, this test needs to be broken down into smaller tests, it's a beast right now. This all needs a lot of cleanup.

I'm not sure we should use arquillian. It doesn't seem to add much. For all the interesting stuff I find myself dropping out to using the docker client directly. Arquillian basically just provides something like docker-compose, but even then, it's a bit limited.  The documentation is not very good (there's a lot of it, but it reads like a novel, and there's no javadoc). It doesn't seem to allow you to dump the logs from the containers at the end of a run. I think we'd be more productive just using docker-client directly.

#903


sijie:
the overall framework looks good to me. However I am not sure if using a backward compat proxy is the right approach here. 

1) I guess the reason it exists is because you want to simplify writing test cases and be able to adopt to multiple versions. But this approach only works for very basic bc tests, because you can't really have a common proxy over multiple versions. 

2) I am afraid adding a bc proxy will introduce pains on maintaining the proxy for bc tests. It doesn't sound actually reduce the maintenance overhead here.

3) I would prefer testing BC between client and bookies without extra proxy if possible. You can actually test and assert on the client behaviors that actually talks to bookies and zookeeper, instead of proxying or delegation.

Other comments:

- I think we need to stop support versions prior to 4.3.0. We need to enforce EOL as we planned in https://cwiki.apache.org/confluence/display/BOOKKEEPER/BP-13+-+Time+Based+Release+Plan
- I am not sure whether arquillian is good or not. but the original idea is using arquillian is not only for BC tests, but also for integration/failure injection tests. It is good for use one framework, rather than inventing our own testing wheel for different purposes.

ivankelly:
@sijie @eolivelli 
How about this approach? It uses a library called MavenClassLoader to pull all dependencies for each version into ClassLoader. Then the methods are called via reflection (I don't think there any other way).
The backward proxy stuff is no longer needed.

This is only needed for interop tests, where we need to check how ledgers from one version interact with ledgers from another version. Testing all client versions with the latest bookies should each have their own simpler tests.

Do we actually state anywhere what our overarching compatibility guarantees are? 

Re arquillian, I'm aware we want to use for integration tests and failure injection too. However, I'd like to see a PoC of how this would be done before settling with it, but so far it's been a pain to work with. Docker-java is wordy, but it follows the docker rest api, which is very well documented.

ivankelly:
maven-classloader is available at https://github.com/yahro/maven-classloader

It's got no license right now. I've messaged to author to see if that could be changed.

eolivelli:
@ivankelly I was doing the same, see my email

ivankelly:
@eolivelli @sijie Added groovy version based on enrico's work. Looks nice. I recommend we go down this path.

sijie:
the maven resolver + groovy approach looks good to me. +1

one comment: regarding docker image - we publish official docker images since 4.5.0. it would be good to use the official docker image. so this would also cover testing docker images as well.

@eolivelli I would prefer using docker if possible. docker image is pretty much a standard container now. 1) using docker images provide better isolation, which can increase testing parallelism. 2) we can also test docker images as well.

eolivelli:
I am happy that you like this approach @sijie

If you feel strong about keeping docker I can accept it.

I don't think it can increase parallelism, you are not using multiple machines, it will make the test slower.

I think that tests on the official docker images can be a good thing but it will be a separate task/system.
In order to support the current version bookie we have to create a dummy image, maybe with the spotify plugin. With a pure maven resolver approach you can use simply the artifacts of the current build

sijie:
> I don't think it can increase parallelism, you are not using multiple machines, it will make the test slower.

why? in a single machine, you can run parallel junit test, if each unit test has its own isolated docker servers.

sijie:
just to clarify what I understand about the approach we are taking here:

- we are still using Arquilian + docker for setting up junit tests (with docker images)
- we use MavenResolver for downloading clients and write unit tests using groovy

ivankelly:
I don't have a strong opinion on whether we use docker for these tests or not, but for sure we will use docker for integration tests soon, so we may as well take the plunge now. Also, the isolation is very nice to have.

@sijie regarding the official image, we can't use that because that only ships with one version of bookkeeper. We need all versions, so that we can verify upgrades between versions. Separate docker images for each version doesn't work, even if sharing volumes, because the IP addresses will change.

@eolivelli You mentioned before that your corp policy doesn't allow docker. Is this just locally? Can you run on a server? I wouldn't expect all users to run these tests every time before they push. They will be heavy (with should be and will be testing far more BC stuff than we do now), so you probably don't even want to run them on a laptop.

sijie:
> regarding the official image, we can't use that because that only ships with one version of bookkeeper. We need all versions, so that we can verify upgrades between versions. Separate docker images for each version doesn't work, even if sharing volumes, because the IP addresses will change.

sure. but not all tests requires testing bookie upgrade, no? e.g. for network BC tests, why can't you use the official images?

ivankelly:
@sijie Most likely. The one test I chose for the PoC did need upgrade though. We should use the official image where ever possible. However, we will need to change some things to allow for the master to go into the image.

sijie:
@ivankelly okay. it is okay to use all-versions image for that. we can have a separate PR for having pure client BC tests for a given docker image to verify official docker image as well.

ivankelly:
There's going to be a bunch of PRs for this.

If we agree on approach (which I think we do), MavenResolver + groovy + docker + arquillian, then I'll start getting PRs up today.

The endgoal is to remove all the shading rubbish from BC tests. After that we can start adding new BC tests.

sijie:
+1

eolivelli:
+1 for docker and the current approach. I image you will clean up this patch.
I wonder if checkstyle can work on groovy files.

I am using a lot docker + Jenkins to run tests and integration tests in my company. We have strict rules because technically docker does not provide a strong sandbox. 
For it won't be  a real problem maybe it would for new contributors. I propose to move forward and if contributors will have problems we will see how to help.
It is important for a contributor which needs to write a BC test to be able to work efficiently.

I think that accesssing the logs for such ephemeral containers could be nasty, for instance

ivankelly:
> It is important for a contributor which needs to write a BC test to be able to work efficiently.
The plan is for it to work out of the box with mvn test -Pintegration if you have docker installed.

> I think that accesssing the logs for such ephemeral containers could be nasty, for instance

This is one of my gripes with Arquillian. I think I can do something with stop actions though.

ivankelly:
This is ready to be broken up

sijie:
@ivankelly is this PR still needed?

ivankelly:
Nope, last of it went in just now

Issue #931, opened 2018-01-02
Add option for tracking Bookie task execution time


athanatos:
Fix bug in OrderedSafeExecutor which prevents time tracking in some cases and add an option to enable on the Bookie.

Will have PR shortly.  See also #907

Issue #907, opened 2017-12-22
Add tests to validate Bookie read/write stats


athanatos:
It's pretty easy at the moment to inadvertently break stats necessary for service monitoring.  Add some unit tests to validate some basic bookie read/write stats along with machinery for expanding coverage later.

PR #932, opened 2018-01-02
ISSUE #931,#907: Add option to track task execution time

athanatos:
Fixes a bug in OrderedScheduler introduced in
e33ec10aa400f32c2e0278c15ea80a0f624e5919 which failed to track execution
time with some calls and adds an option to enable it in the bookie.  Also
fixes a bug with task_queued duration.

Add a simple mock for remembering stats long enough to verify that
counters are actually used and sensible in unit tests and bake it into
BookKeeperClusterTestCase so that we can write tests to ensure that the
stats are actually counted and make sense.  Use said mock to add simple
tests for top level read and write stats validating this fix.

(@bug W-4276826@)
(@bug W-4268290@)

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
@@ -142,16 +142,21 @@ public class BookieRequestProcessor implements RequestProcessor {
             StatsLogger statsLogger, SecurityHandlerFactory shFactory) throws SecurityException {
         this.serverCfg = serverCfg;
         this.bookie = bookie;
-        this.readThreadPool = createExecutor(this.serverCfg.getNumReadWorkerThreads(),
-                "BookieReadThread-" + serverCfg.getBookiePort(),
-                serverCfg.getMaxPendingReadRequestPerThread());
-        this.writeThreadPool = createExecutor(this.serverCfg.getNumAddWorkerThreads(),
-                "BookieWriteThread-" + serverCfg.getBookiePort(),
-                serverCfg.getMaxPendingAddRequestPerThread());
-        this.longPollThreadPool =
-            createExecutor(
+        this.readThreadPool = createExecutor(
+                this.serverCfg.getNumReadWorkerThreads(),
+                "BookieReadThreadPool",
+                serverCfg.getMaxPendingReadRequestPerThread(),
+                statsLogger);
+        this.writeThreadPool = createExecutor(
+                this.serverCfg.getNumAddWorkerThreads(),
+                "BookieWriteThreadPool",
+                serverCfg.getMaxPendingAddRequestPerThread(),
+                statsLogger);
+        this.longPollThreadPool = createExecutor(
                 this.serverCfg.getNumLongPollWorkerThreads(),
-                "BookieLongPollThread-" + serverCfg.getBookiePort(), OrderedScheduler.NO_TASK_LIMIT);
+                "BookieLongPollThread",
+                OrderedScheduler.NO_TASK_LIMIT,
+                statsLogger);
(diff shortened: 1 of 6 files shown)


bookkeeper-server/src/test/java/org/apache/bookkeeper/test/OpStatTest.java
```diff
@@ -0,0 +1,125 @@
...
+        f.accept(logger.getSuccessCount(), logger.getSuccessAverage());
+    }
+
+    private void validateOpStat(TestStatsProvider stats, String paths[], BiConsumer<Long, Double> f) {
+        for (String path : paths) {
+            validateOpStat(stats, path, f);
+        }
+    }
+
+    @Test(timeout = 6000)
```
eolivelli: Nit: Drop timeouts we have now in surefire config

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
```diff
@@ -142,16 +143,21 @@ public BookieRequestProcessor(ServerConfiguration serverCfg, Bookie bookie,
...
+                this.serverCfg.getNumAddWorkerThreads(),
+                "BookieWriteThreadPool",
+                serverCfg.getMaxPendingAddRequestPerThread(),
+                statsLogger);
+        this.longPollThreadPool = createExecutor(
                 this.serverCfg.getNumLongPollWorkerThreads(),
-                "BookieLongPollThread-" + serverCfg.getBookiePort(), OrderedScheduler.NO_TASK_LIMIT);
+                "BookieLongPollThread",
+                OrderedScheduler.NO_TASK_LIMIT,
+                NullStatsLogger.INSTANCE);
```
eolivelli: Can't we pass the statslogger here as well?

bookkeeper-server/src/test/java/org/apache/bookkeeper/test/BookKeeperClusterTestCase.java
```diff
@@ -480,16 +485,14 @@ public void restartBookie(BookieSocketAddress addr) throws Exception {
...
         for (BookieServer server : bs) {
             if (server.getLocalAddress().equals(addr)) {
-                server.shutdown();
                 toRemove = server;
                 break;
             }
             ++toRemoveIndex;
         }
         if (toRemove != null) {
-            stopAutoRecoveryService(toRemove);
```
eolivelli: This line seems to related to the issue

eolivelli: Sorry for typo, it does not seem to be related to the issue

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
athanatos: Yeah, no reason not to.  Will fix.

bookkeeper-server/src/test/java/org/apache/bookkeeper/test/BookKeeperClusterTestCase.java
athanatos: stopAutoRecoveryService gets called as part of of killBookie.  I refactored it a little to make the bs/bsLoggers state maintenance slightly less error prone.

bookkeeper-server/src/test/java/org/apache/bookkeeper/test/OpStatTest.java
athanatos: Yep.

athanatos:
@eolivelli I think I've addressed your other comments.

PR #939, opened 2018-01-03
BookieRequestProcessor: submit bookie info v3 ordered

athanatos:
This addition came in after 7673febb6519fedb96320ec98db71601826a63b1
and wasn't caught.


sijie:
@eolivelli you are right. nice catch.

@athanatos is this change in salesforce's production?

athanatos:
No, sorry about that.  It's something I noticed when was merging some other changes, but apparently I didn't look hard enough!

athanatos:
Actually, it is running in production.  I wonder why that seems to work...

athanatos:
Ok, getAddRequest() there is going to return a default instance with ledger set to 0, so it would work, but all such requests would wind up on a wq single thread.  Will fix in our code base, thanks!

PR #940, opened 2018-01-03
BookieProtoEncoding: use retainedSlice() to reduce garbage

athanatos:

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtoEncoding.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtoEncoding.java
@@ -289,8 +289,8 @@ public class BookieProtoEncoding {
                 entryId = buffer.readLong();
 
                 if (rc == BookieProtocol.EOK) {
-                    ByteBuf content = buffer.slice();
-                    return new BookieProtocol.ReadResponse(version, rc, ledgerId, entryId, content.retain());
+                    return new BookieProtocol.ReadResponse(version, rc,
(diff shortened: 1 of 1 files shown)


sijie:
@merlimat can you review this bytebuf change? since you might have more context about it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtoEncoding.java
```diff
@@ -289,8 +289,8 @@ public Object decode(ByteBuf buffer)
                 entryId = buffer.readLong();
 
                 if (rc == BookieProtocol.EOK) {
-                    ByteBuf content = buffer.slice();
-                    return new BookieProtocol.ReadResponse(version, rc, ledgerId, entryId, content.retain());
+                    return new BookieProtocol.ReadResponse(version, rc,
```
merlimat: I'm not 100% sure the `slice()` can be avoided, because we would have to double check all the places that are touching that buffer.

I think a better option here is to use `buffer.retainedSlice()` which achieves both goals (and also lets the resulting `ByteBuf` slice object to be pooled.

merlimat: `retainedSlices()` was not used originally because it was introduced in netty-4.1 while the original patch was based on netty-4.0.

athanatos: Alright, I'll switch it to retainedSlice()

sijie:
@merlimat can you review @athanatos 's latest change?

PR #941, opened 2018-01-03
BookieNettyServer: add contextHandler to the local transport pipeline

athanatos:
Left off of 8e0bd2c3d81b522e97434d8646915f36422a104b.  In fact,
authentication is already enabled on LocalTransport.  This extra line is
needed for the machinery which allows auth plugins to access ssl state.
Currently, the only plugin which uses that machinery is in TestTLS.
Adds test cases to validate that functionality with Local Transport
enabled.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieNettyServer.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieNettyServer.java
@@ -372,6 +372,7 @@ class BookieNettyServer {
                             : new RejectRequestHandler();
                     pipeline.addLast("bookieRequestHandler", requestHandler);
 
+                    pipeline.addLast("contextHandler", contextHandler);
                 }
             });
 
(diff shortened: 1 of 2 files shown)


athanatos:
@kishorekasi Is this right?

eolivelli:
In theory  local transport does not need TLS.
I am an user of local transport but not tls.
It would be good to add a testcase to the tls suite which certifies that bookie is working properly in case of tls enabled + local transport enabled.
Just pick a tls test and enable local transport on the bookie, on client side it is always enabled

kishorekasi:
I agree with @eolivelli. While it is generally true that tls is not required for in-vm-communication, when I initially coded this, I added contextHandler not knowing how there is an a usecase with tls.

athanatos:
@kishorekasi Ok, so you'd suggest dropping it?

kishorekasi:
There is not harm in keeping it. ContextHandler will only be used when TLS is enabled.

athanatos:
@kishorekasi Ok, I'll find a test on which to enable tls.

athanatos:
Ok, it's really there to enable passing ssl cert state into the auth plugin.  Probably not strictly *necessary* with a LocalTransport, but it should probably be symmetric.  I added tests to validate the behavior.

athanatos:
@kishorekasi How about this?

kishorekasi:
LGTM

Issue #943, opened 2018-01-04
Cleanup `Thread.sleep` in unit tests 


sijie:
*Problem*

Currently we are using a lot of `Thread.sleep` in the unit tests, waiting for some tasks to complete execution. This pattern is very problematic, which it makes the tests unstable and non-deterministic. 

*Solution*

We should try to clean up the usages of `Thread.sleep` in the unit tests. This is a master ticket for tracking this effort.

eolivelli:
This is related to #751

PR #944, opened 2018-01-04
Introduce MockClock and MockExecutorController to improve timer-based testing

sijie:

- introduce MockClock and MockExecutorController to control time advancing to trigger execution on tasks
- convert LedgerDirsMonitor from using Thread to use ScheduledExecutorService (moved the logic in `run` method to `check` without code changes)
- change TestLedgerDirsManager to use MockClock and MockExecutorController to remove the usage of `Thread.sleep`

Master Issue:  #943

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDirsMonitor.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDirsMonitor.java
@@ -21,12 +21,17 @@
 
 package org.apache.bookkeeper.bookie;
 
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.ConcurrentMap;
 
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
 import org.apache.bookkeeper.bookie.LedgerDirsManager.LedgerDirsListener;
 import org.apache.bookkeeper.bookie.LedgerDirsManager.NoWritableLedgerDirException;
 import org.apache.bookkeeper.conf.ServerConfiguration;
@@ -40,7 +45,7 @@ import org.slf4j.LoggerFactory;
 /**
  * Thread to monitor the disk space periodically.
  */
-class LedgerDirsMonitor extends BookieThread {
+class LedgerDirsMonitor {
     private static final Logger LOG = LoggerFactory.getLogger(LedgerDirsMonitor.class);
 
     private final int interval;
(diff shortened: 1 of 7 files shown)


sijie:
This change is to demonstrate how we can avoid `Thread.sleep` using a mock executor and a mock clock on controlling time advancing for timer tasks.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDirsMonitor.java
```diff
@@ -40,117 +45,109 @@
 /**
  * Thread to monitor the disk space periodically.
  */
-class LedgerDirsMonitor extends BookieThread {
+class LedgerDirsMonitor {
```
jvrao: I understand you are trying to use an executor for this, which is nice, but does it have to do anything with introducing mock executor? If not, this being a different commit will help reviewer.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDirsMonitor.java
```diff
@@ -40,117 +45,109 @@
...
     }
 
-    @Override
-    public void run() {
-        while (true) {
-            try {
-                List<File> writableDirs = ldm.getWritableLedgerDirs();
-                // Check all writable dirs disk space usage.
-                for (File dir : writableDirs) {
+    private void check() {
```
jvrao: Looks like a majority of the changes are because of removing the run() around the code and making it a check() method. I believe there is no logic change. If so, a comment will help the reviewer. I wish these diff tools are intelligent enough to identify things like that.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDirsMonitor.java
sijie: @jvrao : 

the PR is more about introducing mock executor to improve testing ability. 

I am using changing LedgerDirsMonitor from a thread based approach to an executor based approach as an example, to demonstrate how an executor based approach together with a mock executor can improve testing ability.

I was thinking of separating this PR into two, but that means the reviewers don't have the direct impression on how an executor based approach together with a mock executor can improve testing ability. I choose LedgerDirsMonitor as the example to demonstrate is becase it is a real self-contained class that can be easily converted from a thread based approach to an executor based approach.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDirsMonitor.java
sijie: yes there is no code change there. I will add a note to description and leave a comment so people will see it.

sijie: for all reviewers: `check` method is the logic moved from `run` method, no code change here.

sijie:
@eolivelli it is a very common pattern used in quite a lot of places. for example, all the twitter util libraries using mock time and mock timer for testing timer oriented tasks. apache aurora (another twitter open source project) does similar thing. even in guava, you have a mock ticker that controls expiration.

sijie:
@jvrao ping? do you want me to break this into two prs or not?

Issue #947, opened 2018-01-05
In LocalBookkeeper, make zkDir configurable and get ledgersrootpath from conf


reddycharan:
**FEATURE REQUEST**

1. Please describe the feature you are requesting.
- make ZK's data directory configurable for local bookie
- Localbookie to use ledgersrootpath from conf file

2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have). Are you currently using any workarounds to address this issue?

must-have for our testing envs.

PR #948, opened 2018-01-05
ISSUE #947: ZK configs for LocalBookkeeper

reddycharan:

Master Issue: #947 

- make ZK's data directory configurable for local bookie
- Localbookie to use ledgersrootpath from conf file

e032218a910113b0e4de561a76da8bdf61ba3db7
(@bug W-2887104@) make ZK's data directory configurable for local bookie

Author: Andrey Yegorov <ayegorov@salesforce.com>

> ---
> quickly and easily:
>
>
>
> Otherwise:
> 
> 
> ---

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/util/LocalBookKeeper.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/util/LocalBookKeeper.java
@@ -31,6 +31,7 @@ import java.util.List;
 
 import org.apache.bookkeeper.bookie.BookieException;
 import org.apache.bookkeeper.client.BKException;
+import org.apache.bookkeeper.conf.AbstractConfiguration;
 import org.apache.bookkeeper.conf.ServerConfiguration;
 import org.apache.bookkeeper.proto.BookieServer;
 import org.apache.bookkeeper.replication.ReplicationException.CompatibilityException;
@@ -104,7 +105,7 @@ public class LocalBookKeeper {
         return server;
     }
 
-    private void initializeZookeeper(String zkHost, int zkPort) throws IOException {
+    private void initializeZookeeper(AbstractConfiguration conf, String zkHost, int zkPort) throws IOException {
         LOG.info("Instantiate ZK Client");
         //initialize the zk client with values
         ZooKeeperClient zkc = null;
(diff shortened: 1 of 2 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/util/LocalBookKeeper.java
```diff
@@ -280,11 +291,19 @@ static void startLocalBookiesInternal(ServerConfiguration conf,
         List<File> bkTmpDirs = null;
         try {
             if (shouldStartZK) {
-                zkTmpDir = IOUtils.createTempDir("zookeeper", dirSuffix);
+                if (null == zkDataDir) {
+                    zkTmpDir = IOUtils.createTempDir("zookeeper", dirSuffix);
```
sijie: any reason why you are using two different createTempDir method here? is `IOUtils.createTempDir` not enough?

reddycharan: IOUtils.createTempDir creates dir in temp folder and Files.createTempDirectory creates dir in the specified directory, which we are intending to do with this change

sijie: IOUtils.createTempDir is using File.createTempFile(...), which create a temp dir under default temporary-file directory.

You can use [`File.createTempFile(prefix, suffix, directory)`](https://docs.oracle.com/javase/7/docs/api/java/io/File.html#createTempFile(java.lang.String,%20java.lang.String,%20java.io.File)) to create temp dir at a given directory.

I think the question I am having here: we are having two different classes for creating temp directory. It is better to add `createTempDir(prefix, suffix, directory)` in IOUtils and use this method here, to make things consistent and easier to maintain.

```
class IOUtils {

        public static File createTempDir(String prefix, String suffix) {
             return createTempDir(prefix, suffix, null);
        }

         public static File createTempDir(String prefix, String suffix, File dir)
            throws IOException {
        File tmpDir = File.createTempFile(prefix, suffixi, dir);
        if (!tmpDir.delete()) {
            throw new IOException("Couldn't delete directory " + tmpDir);
        }
        if (!tmpDir.mkdir()) {
            throw new IOException("Couldn't create directory " + tmpDir);
        }
        return tmpDir;
    }

}
```

then you can use it here:

```
zkTempDir = IOUtils.createTempDir("zookeeper", "dirSuffix", zkDataDir);
```

reddycharan: made corresponding changes

PR #949, opened 2018-01-05
Build Docker image from maven

ivankelly:
The docker image is build on the install phase, using the tarball
generated by the bookkeeper-dist-server module.

It is only built if the docker profile is enabled.
```
mvn install -Pdocker
```

This removes the need to update the docker file after each release, as
the image is automatically generated as part of the release process.

### DIFF
--- a/bookkeeper-dist/server/pom.xml
+++ b/bookkeeper-dist/server/pom.xml
@@ -78,7 +78,7 @@
         <version>${maven-assembly-plugin.version}</version>
         <configuration>
           <finalName>bookkeeper-server-${project.version}</finalName>
-          <attach>false</attach>
+          <attach>true</attach>
           <descriptors>
             <descriptor>../src/assemble/bin-server.xml</descriptor>
           </descriptors>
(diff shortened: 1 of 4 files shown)


sijie:
@ivankelly I don't think we should couple changing how current docker image is built together with BC tests. Because the workflow of publishing docker image is based on dockerhub autobuild, your change will break the workflow completely. 

> This removes the need to update the docker file after each release, as
the image is automatically generated as part of the release process.

Currently the docker image is autobuild under `apache` dockerhub account, which you can't make the image generation as part of the release process. If we want to change the docker image generation and release process, this has to go to a different topic, which there was already an email thread discussing that, but lack of efforts pushing forward.

I would suggest not touching `docker` directory, if you want to generate a docker image for current master, make it part of `tests` for now. We should have a different story/BP discussing docker image generation, release and making it an official image.

eolivelli:
I agree with @sijie . I missed that this is a change in the docker directory, I thought it was a new Dockerfile.
I think it is fine to have a separate Dockerfile for this purpose

ivankelly:
Sure. Will scrap this change for now. However, it means we can't use the official image in bc/integ testing (except in a very narrow case). I'll continue with the all-versions image.

Issue #950, opened 2018-01-05
EntryLogger.extractEntryLogMetadataByScanning.accept check is wrong


athanatos:
Should be >= 0 (0 is a valid ledger id).  Seems to be causing org.apache.bookkeeper.replication.TestReplicationWorker.test2RWsShouldCompeteForReplicationOf2FragmentsAndCompleteReplication to fail in cases where the gc runs before the recovery reads happen.  The gc winds up believing the entry log to be empty and deletes it erroneously.

athanatos:
Have patch, will post shortly.

PR #951, opened 2018-01-05
ISSUE #950: EntryLogger.extractEntryLogMetadataByScanning: fix check

athanatos:
0 is a valid ledger id. Seems to be causing
test2RWsShouldCompeteForReplicationOf2FragmentsAndCompleteReplication to
fail in cases where the gc runs before the recovery reads happen. The gc
winds up believing the entry log to be empty and deletes it erroneously.

Introduced in 143bd1952851ac70ace9d16c3708ae30dad2a0aa.

(@bug W-4596080@)

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
@@ -1339,7 +1339,7 @@ public class EntryLogger {
 
             @Override
             public boolean accept(long ledgerId) {
-                return ledgerId > 0;
+                return ledgerId >= 0;
             }
         });
 


eolivelli:
@sijie this has not been cherry picked to 4.6 branch, shall I do it ?

eolivelli:
@sijie  cherry pick fails, will send another PR for inclusion of this fix in 4.6

eolivelli:
@sijie  branch 4.6 is not affected by this problem

```
private EntryLogMetadata extractEntryLogMetadataByScanning(long entryLogId) throws IOException {
        final EntryLogMetadata meta = new EntryLogMetadata(entryLogId);

        // Read through the entry log file and extract the entry log meta
        scanEntryLog(entryLogId, new EntryLogScanner() {
            @Override
            public void process(long ledgerId, long offset, ByteBuffer entry) throws IOException {
                // add new entry size of a ledger to entry log meta
                meta.addLedgerSize(ledgerId, entry.limit() + 4);
            }

            @Override
            public boolean accept(long ledgerId) {
                return true;
            }
        });

        if (LOG.isDebugEnabled()) {
            LOG.debug("Retrieved entry log meta data entryLogId: {}, meta: {}", entryLogId, meta);
        }
        return meta;
    }
```

can you confirm ?
If so I will drop the 4.6.1 label

eolivelli:
It was I bug introduced on master with
```
143bd1952851ac70ace9d16c3708ae30dad2a0aa	7-dic-2017 23.42.46	Matteo Merli <mmerli@apache.org>
Use ConcurrentLongLongHashMap in EntryLogMetadata

Porting commits https://github.com/yahoo/bookkeeper/commit/74afc16d1c1e93739edf4a78801c5ccb6afd3883 and https://github.com/yahoo/bookkeeper/commit
```/eba3bbfbbaf58118c0c5736b56d3897a5928bc89 from Yahoo branch.

sijie:
Yes that's my bad . We need to remove the 4.6 label

Issue #935, opened 2018-01-03
org.apache.bookkeeper.tests.shaded.BookKeeperServerShadedArtifactTest.testProtobufIsShaded failed


sijie:
**BUG REPORT**

1. Please describe the issue you observed:

- What did you do?

run `mvn install` in current master

- What did you expect to see?

it should complete successfully.

- What did you see instead?

The command failed with 

```
Error Message
Expected exception: java.lang.ClassNotFoundException
Stacktrace
java.lang.AssertionError: Expected exception: java.lang.ClassNotFoundException
```

The problem is introduced due to merging #856 after merging #928. Because `pulsar-checksum` depends on guava, which is not included in the changes of #928.

athanatos:
I pushed a PR for this one.

sijie:
@athanatos this is fixed in #937 , because the problem was introduced by `pulsar-checksum` which will be substituted to imported `circe-checksum` in #937. I didn't merge #937 because of current CI job doesn't handle well with paths containing spaces, which I have a fix in #952 .

PR #937, opened 2018-01-03
Issue 935: Issue 915: Part 3 - Make circe-checksum module work for both java 8 and java 9

sijie:
Descriptions of the changes in this PR:

    - switch from jmockit to mockito, and from testng to junit. make testing and mocking framework consistent across modules.
    - skip `-Werror` check in circe-checksum since `finalize` is deprecated at java 9
    - change CRC32CDigestManager to use `circe-checksum` module
    - fix shading to include `circe-checksum`

This change is based #917 . Gitsha 73049f8 contains the changes to make circe-checksum module work for both java8 and java9.

PR #955, opened 2018-01-05
ISSUE #935: tests/bookkeeeper-server-shaded-*test/pom.xml: exclude pulsar

athanatos:
This prevents the test artifact from pulling in guava via pulsar.


athanatos:
This seems to fix the test issue, but I'm not really sure that it's correct.

sijie:
@athanatos the correct fix is to include `pulsar-checksum` in the shading rule to make guava shaded correctly. see #937 

also see my comment at #935

athanatos:
@sijie Gotcha, I'll leave it to you then.

Issue #954, opened 2018-01-05
dir_*_usage stats are always 0


pasha-kuznetsov:
**BUG REPORT**

1. Please describe the issue you observed:

- What did you do?

Normal BK operation

- What did you expect to see?

`dir_*_usage` gauges to contain dir usage %

- What did you see instead?

`dir_*_usage` gauges are always `0` due to mismatch between the way they are initialized in `LedgerDirsManager`:

```
    LedgerDirsManager(ServerConfiguration conf, File[] dirs, DiskChecker diskChecker, StatsLogger statsLogger) {
        this.ledgerDirectories = Arrays.asList(Bookie
                .getCurrentDirectories(dirs)); // <------------- NOTE adding "/current" here
        this.writableLedgerDirectories = new ArrayList<File>(ledgerDirectories);
        this.filledDirs = new ArrayList<File>();
        ...
        for (File dir : dirs) { // <------------- NOTE using the original dirs
            diskUsages.put(dir, 0f);
            String statName = "dir_" + dir.getPath().replace('/', '_') + "_usage";
            final File targetDir = dir;
            statsLogger.registerGauge(statName, new Gauge<Number>() {
                ...
                @Override
                public Number getSample() {
                    return diskUsages.get(targetDir) * 100;
                }
            });
        }
```
and the way the `diskUsages` is populated subsequently in `LedgerDirsMonitor`:
```
    @Override
    public void run() {
        while (true) {
            try {
                List<File> writableDirs = ldm.getWritableLedgerDirs();
                // Check all writable dirs disk space usage.
                for (File dir : writableDirs) {
                    try {
                        diskUsages.put(dir, diskChecker.checkDir(dir)); // <------------- NOTE that writableDirs has "/current" appended
                ...
            try {
                if (hasWritableLedgerDirs
                        || (totalDiskUsage = diskChecker.getTotalDiskUsage(ldm.getAllLedgerDirs())) < conf
                                .getDiskLowWaterMarkUsageThreshold()) {
                    // Check all full-filled disk space usage
                    for (File dir : fullfilledDirs) {
                        try {
                            diskUsages.put(dir, diskChecker.checkDir(dir)); // <------------- NOTE that fullfilledDirs has "/current" appended
```

jvrao:
Good Catch @pasha-kuznetsov

PR #957, opened 2018-01-06
ISSUE #954: dir_*_usage stats are always 0

pasha-kuznetsov:
due to mismatch between the way they are initialized in `LedgerDirsManager` and
the way the `diskUsages` is populated subsequently in `LedgerDirsMonitor`

Decisions potentially worth some attention:
- stat names are kept `dir_{dir}_usage` while the actual dirs being checked have been changed to `{dir}/current`;
- the stats aren't being updated if the bookie is in a readonly mode -- uncovered while testing, not sure if it's worth addressing at all, kept the fix minimalistic for now.

(@bug W-4594204@)


sijie:
retest this please

PR #959, opened 2018-01-08
Docker image for testing

ivankelly:
Contains all released versioned and the current version (from the
current source tree). The image only gets built if the docker maven
profile is enabled:

```
mvn package -Pdocker
```

Supervisord is used to start and stop bookkeeper. Each version has its
own supervisord config. The same journal and ledger directories are
used for all versions. Logs are written to per version log files in
/var/log/bookkeeper.


tests/docker-all-version-image/Dockerfile
```diff
@@ -0,0 +1,58 @@
...
+RUN wget -nv https://archive.apache.org/dist/zookeeper/bookkeeper/bookkeeper-4.3.0/bookkeeper-server-4.3.0-bin.tar.gz{,.sha1,.md5,.asc}
+RUN wget -nv https://archive.apache.org/dist/bookkeeper/bookkeeper-4.2.4/bookkeeper-server-4.2.4-bin.tar.gz{,.sha1,.md5,.asc}
+RUN wget -nv https://archive.apache.org/dist/bookkeeper/bookkeeper-4.3.0/bookkeeper-server-4.3.0-bin.tar.gz{,.sha1,.md5,.asc}
+RUN wget -nv https://archive.apache.org/dist/bookkeeper/bookkeeper-4.3.1/bookkeeper-server-4.3.1-bin.tar.gz{,.sha1,.md5,.asc}
+RUN wget -nv https://archive.apache.org/dist/bookkeeper/bookkeeper-4.3.2/bookkeeper-server-4.3.2-bin.tar.gz{,.sha1,.md5,.asc}
+RUN wget -nv https://archive.apache.org/dist/bookkeeper/bookkeeper-4.4.0/bookkeeper-server-4.4.0-bin.tar.gz{,.sha1,.md5,.asc}
+RUN wget -nv https://archive.apache.org/dist/bookkeeper/bookkeeper-4.5.0/bookkeeper-server-4.5.0-bin.tar.gz{,.sha1,.md5,.asc}
+RUN wget -nv https://archive.apache.org/dist/bookkeeper/bookkeeper-4.5.1/bookkeeper-server-4.5.1-bin.tar.gz{,.sha1,.md5,.asc}
+RUN wget -nv https://archive.apache.org/dist/bookkeeper/bookkeeper-4.6.0/bookkeeper-server-4.6.0-bin.tar.gz{,.sha1,.md5,.asc}
```
eolivelli: @sijie we will have to update this file at every release

eolivelli: @ivankelly  @sijie 
maybe we should have somewhere a list of all released versions, don't know where

ivankelly: You mean in the source tree somewhere?

eolivelli: Sonething that is usable to have a list of version in machine readable format.
Maybe on the the git repo on master branch, like the website.
Script like this will do some kind of loop and we won't have to update them manually at very each release

ivankelly: well, i've put these on individual lines on purpose, so that it doesn't download them each time. docker should cache them.

We could just download every bookkeeper-server-*-bin.tar.gz under dist/bookkeeper/* every time, but I don't think the apache folks would be too happy about that.

eolivelli: Very good idea, docker creates and caches layers

tests/docker-all-version-image/pom.xml
```diff
@@ -0,0 +1,88 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <groupId>org.apache.bookkeeper.tests</groupId>
+    <artifactId>tests-parent</artifactId>
+    <version>4.7.0-SNAPSHOT</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>org.apache.bookkeeper.tests</groupId>
+  <artifactId>docker-all-version-image</artifactId>
+  <name>Apache BookKeeper :: Tests :: Docker All Versions Image</name>
```
sijie: change "docker-all-version-image" to "docker-all-versions-image" to make naming consistent.

tests/docker-all-version-image/pom.xml
```diff
@@ -0,0 +1,88 @@
...
+              <execution>
+                <id>default</id>
+                <goals>
+                  <goal>build</goal>
+                  <goal>push</goal>
+                </goals>
+              </execution>
+            </executions>
+            <configuration>
+              <repository>apache/bookkeeper-all-versions</repository>
```
sijie: I would suggest not using `apache` namespace, because `apache` is a docker hub namespace managed by the ASF: https://hub.docker.com/u/apache/. I would suggest using `apachebookkeeper/bookkeeper-all-versions`, the `apachebookkeeper` namespace is at least owned by our community.

`apachebookkeeper` was created for the purpose of making an official docker repo.

bookkeeper-dist/server/pom.xml
```diff
@@ -78,7 +78,7 @@
         <version>${maven-assembly-plugin.version}</version>
         <configuration>
           <finalName>bookkeeper-server-${project.version}</finalName>
-          <attach>false</attach>
+          <attach>true</attach>
```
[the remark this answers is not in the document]
ivankelly: Then we wouldn't be testing what we actually ship.

How about we just skip the deploy phase for this module.

```
<plugin>
  <artifactId>maven-deploy-plugin</artifactId>
  <version>X.Y</version>
  <configuration>
    <skip>true</skip>
  </configuration>
</plugin>
```

This will stop it being sent remotely, but sibling modules will be able to depend on it.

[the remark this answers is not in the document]
sijie: That is fine to me as well.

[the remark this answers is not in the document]
ivankelly: Just checked and this this configuration is already there

tests/docker-all-version-image/pom.xml
ivankelly: Changed

tests/docker-all-version-image/pom.xml
ivankelly: changed

Issue #933, opened 2018-01-03
Enable PEM format X.509 certs


kishorekasi:
**FEATURE REQUEST**

1. Please describe the feature you are requesting.
Currently Bookkeeper only supports JKS key store and trust stores. Another widely used file format is PEM for certificates and certificate chains. Without in-build support in bookkeeper, an external tool is necessary to convert PEM files to key stores.

2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have). Are you currently using any workarounds to address this issue?
should-have.

3. Provide any additional detail on your proposed use case for this feature.
Support for PEM file formats in addition to JKS file formats.

PR #965, opened 2018-01-10
Issue 933: Add support for PEM Key file formats

kishorekasi:
(@bug W-4203319@) TLS authentication to support PEM format X.509 
certificates along with JKS and PKCS12 formats.

Added another KeyStore Type to the list of formats (JKS, PKCS12)
currently supported during SSLContext creation.

@rev ayegorov@

Signed-off-by: Kishore Kasi Udayashankar <kudayashankar@salesforce.com>

Master Issue: #933


bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ClientConfiguration.java
```diff
@@ -1414,118 +1414,137 @@ public String getClientRole() {
...
 
     /**
-     * Get the keystore type for client. Default is JKS.
+     * Get the keyfile type for client. Default is JKS.
      *
      * @return
      */
-    public String getTLSKeyStoreType() {
-        return getString(TLS_KEYSTORE_TYPE, "JKS");
+    public String getTLSKeyFileType() {
```
eolivelli: I think we cannot change all of these names it is public API.
I would prefer to keep old names

bookkeeper-server/src/main/java/org/apache/bookkeeper/tls/TLSContextFactory.java
```diff
@@ -70,11 +80,14 @@ private String getPasswordFromFile(String path) throws IOException {
...
         value = "OBL_UNSATISFIED_OBLIGATION",
         justification = "work around for java 9: https://github.com/spotbugs/spotbugs/issues/493")
-    private KeyStore loadKeyStore(String keyStoreType, String keyStoreLocation, String keyStorePassword)
+    private KeyStore loadKeyFile(String keyFileType, String keyFileLocation, String keyFilePassword)
             throws KeyStoreException, NoSuchAlgorithmException, CertificateException, IOException {
-        KeyStore ks = KeyStore.getInstance(keyStoreType);
-        try (FileInputStream ksin = new FileInputStream(keyStoreLocation)) {
-            ks.load(ksin, keyStorePassword.trim().toCharArray());
+        KeyStore ks = KeyStore.getInstance(keyFileType);
+        FileInputStream ksin = new FileInputStream(keyFileLocation);
```
eolivelli: Nit: use try-with-resources

bookkeeper-server/src/main/java/org/apache/bookkeeper/tls/TLSContextFactory.java
```diff
@@ -282,10 +389,12 @@ public SslHandler newTLSHandler() {
         if (protocols != null && protocols.length != 0) {
             sslHandler.engine().setEnabledProtocols(protocols);
         }
+        LOG.info("Enabled cipher protocols: {} ", Arrays.toString(sslHandler.engine().getEnabledProtocols()));
```
eolivelli: Do you need this at info level?

bookkeeper-server/src/test/java/org/apache/bookkeeper/tls/TestTLS.java
```diff
@@ -82,30 +105,113 @@ public void setUp() throws Exception {
...
-        baseConf.setTLSTrustStoreType("JKS");
-        baseConf.setTLSTrustStore(
-            this.getClass().getClassLoader().getResource("cacerts").toURI().getPath());
-        baseConf.setTLSTrustStorePasswordPath(
-            this.getClass().getClassLoader().getResource("trustStorePassword.txt").toURI().getPath());
+
+        switch (serverKeyFileFormat) {
+        case PEM:
+            baseConf.setTLSKeyFileType("PEM");
+            baseConf.setTLSKeyFilePath(this.getClass().getClassLoader().getResource("server-key.pem").getPath());
```
eolivelli: Do you think it is possible to generate these files at build time instead of using static files?
Certificates will expire one day.
If not so simple I am fine with keeping files, but having a readme files which explains how to rebuild the files in each format

bookkeeper-server/src/main/java/org/apache/bookkeeper/tls/TLSContextFactory.java
kishorekasi: ok

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ClientConfiguration.java
kishorekasi: will revert to old names

bookkeeper-server/src/main/java/org/apache/bookkeeper/tls/TLSContextFactory.java
kishorekasi: Not necessarily; this can move to debug

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ClientConfiguration.java
```diff
@@ -145,14 +145,6 @@
     // Client auth provider factory class name. It must be configured on Bookies to for the Auditor
     protected static final String CLIENT_AUTH_PROVIDER_FACTORY_CLASS = "clientAuthProviderFactoryClass";
 
-    // Client TLS
-    protected static final String TLS_KEYSTORE_TYPE = "clientKeyStoreType";
```
sijie: this breaks BC. 

for each field you are change here, I think you have to do:

- in this file: change "TLS_XYZ" to "CLIENT_XYZ", but not change the value.
- change getTLSxyz() to : getProperty(TLS_XYZ, getProperty(CLIENT_XYZ, ...))

I am not what AutoRecovery is using, is it using client settings or server settings? if it is using client settings, you need also to update the configuration settings in the bookie conf.

kishorekasi: I am moving TLS_XYZ from ServerConfiguration and ClientConfiguration to the common class AbstractConfiguration and just use one set of config params (no clientXYZ params anymore). So, effectively this is a dead code. What I am doing here is to keep just one set of config params so both bookie and client config params names are same.

AutoRecovery can still use the same set of params as bookies to configure the bookkeeper client.

Only config param names are changing, but the variable name still remains the same as TLS_XYZ. Do you still think its a BC issue?

sijie: yeah param name changing is considered as breaking BC. because if a client upgrades from 4.6.0 to 4.7.0, without changing the configuration, it won't work.

sijie:
fyi - had a conversation with @kishorekasi offline. we agreed on the approach for addressing the BC concerns here. @kishorekasi will update this PR.

The example PR on showing how to deal with the BC concerns is here: https://github.com/kishorekasi/bookkeeper/pull/1

bookkeeper-server/pom.xml
```diff
@@ -284,16 +284,37 @@
...
+        <executions>
+            <execution>
+                <id>Generate Self-Signed Certificates</id>
+                <phase>generate-test-resources</phase>
+                <goals>
+                    <goal>exec</goal>
+                </goals>
+                <configuration>
+                    <workingDirectory>${basedir}/src/test/resources</workingDirectory>
+                    <executable>${basedir}/src/test/resources/generateKeysAndCerts.sh</executable>
```
eolivelli: This is going to create new certificates at every build, two problems:
1) this is slow
2) it requires some jind of setup of openssl

I suggest to add a profile to build the certificates and commit a working version of certificates.
When they will expire we will do another commit

sijie: How slow is this?

kishorekasi:
2-3 seconds on my Ubuntu server

bookkeeper-server/pom.xml
eolivelli: With this patch at every "mvn test" we are going to re-create new certificates.
I think we have two options:
1) commit certificates and rebuild them when expired
2) make the script rebuild the certificate only if it does not exist

I prefer 1)as it makes tests reproducible at 100% and it does not impact build and dev time

kishorekasi: @eolivelli If it makes sense, I can commit certificates with a very long expiry so that we may never need to generate the certificates again. Also move certification generation to a new profile, just in case we need it.

eolivelli: It would be fine for me, thank you

sijie: +1 on @kishorekasi 's proposed approach.

PR #968, opened 2018-01-10
Journal: add stat for cb queue size

athanatos:
Also adds a bit of functionality to TestStatsProvider to facilitate
testing.

(@bug W-3019451@)
Signed-off-by: Dustin Castor <dcastor@salesforce.com>
[Adapted to current branch, added simple test]


bookkeeper-server/src/test/java/org/apache/bookkeeper/test/TestStatsProvider.java
```diff
@@ -62,12 +64,31 @@ public void dec() {
...
         }
+
+        public void updateMax() {
+            while (true) {
+                long curMax = max.get();
+                long curVal = val.get();
+                if (curVal <= curMax) {
+                    break;
+                } else {
+                    if (max.compareAndSet(curMax, curVal)) {
```
eolivelli: I think this can be done better with only one call to accumulateAndGet

athanatos: @eolivelli Ok, how about this version?

Issue #966, opened 2018-01-10
Expose quorum write commit latency to the client


athanatos:
Will have patch shortly.

PR #970, opened 2018-01-11
ISSUE #966: Expose quorum write complete latency to the client

athanatos:
Add quorum write complete latency to the bookkeeper client API. This
allows callers to find out exactly how long the quorum write took for
diagnostic purposes.

(@bug W-3616193@)
[Fixed merge conflicts, added testing]

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/AsyncCallback.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/AsyncCallback.java
@@ -26,6 +26,30 @@ import org.apache.bookkeeper.common.annotation.InterfaceStability;
 @InterfaceAudience.Public
 @InterfaceStability.Stable
 public interface AsyncCallback {
+    /**
+     * Async Callback for adding entries to ledgers with latency information.
+     *
+     * @since 4.7
+     */
+    @InterfaceAudience.Public
+    @InterfaceStability.Evolving
+    interface AddCallbackWithLatency {
+        /**
+         * Callback declaration which additionally passes quorum write complete latency.
+         *
+         * @param rc
+         *          return code
+         * @param lh
+         *          ledger handle
+         * @param entryId
+         *          entry identifier
+         * @param qwcLatency
+         *          QuorumWriteComplete Latency
+         * @param ctx
+         *          context object
+         */
+        void addCompleteWithLatency(int rc, LedgerHandle lh, long entryId, long qwcLatency, Object ctx);
+    }
 
     /**
      * Async Callback for adding entries to ledgers.
@@ -34,9 +58,9 @@ public interface AsyncCallback {
      */
     @InterfaceAudience.Public
     @InterfaceStability.Stable
-    interface AddCallback {
+    interface AddCallback extends AddCallbackWithLatency {
         /**
-         * Callback declaration.
+         * Callback to implement if latency information is not desired.
          *
          * @param rc
          *          return code
@@ -48,6 +72,24 @@ public interface AsyncCallback {
          *          context object
          */
         void addComplete(int rc, LedgerHandle lh, long entryId, Object ctx);
+
+        /**
+         * Callback declaration which additionally passes quorum write complete latency.
+         *
+         * @param rc
+         *          return code
+         * @param lh
+         *          ledger handle
+         * @param entryId
+         *          entry identifier
+         * @param qwcLatency
+         *          QuorumWriteComplete Latency
+         * @param ctx
+         *          context object
+         */
+        default void addCompleteWithLatency(int rc, LedgerHandle lh, long entryId, long qwcLatency, Object ctx) {
+            addComplete(rc, lh, entryId, ctx);
+        }
     }
 
     /**
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandleAdv.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandleAdv.java
@@ -32,6 +32,7 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.PriorityBlockingQueue;
 import java.util.concurrent.RejectedExecutionException;
 import org.apache.bookkeeper.client.AsyncCallback.AddCallback;
+import org.apache.bookkeeper.client.AsyncCallback.AddCallbackWithLatency;
 import org.apache.bookkeeper.client.SyncCallbackUtils.SyncAddCallback;
 import org.apache.bookkeeper.client.api.WriteAdvHandle;
 import org.apache.bookkeeper.client.api.WriteFlag;
@@ -151,8 +152,33 @@ public class LedgerHandleAdv extends LedgerHandle implements WriteAdvHandle {
         asyncAddEntry(entryId, Unpooled.wrappedBuffer(data, offset, length), cb, ctx);
     }
 
+    /**
+     * Add entry asynchronously to an open ledger, using an offset and range.
+     *
+     * @param entryId
+     *            entryId of the entry to add
+     * @param data
+     *            array of bytes to be written
+     * @param offset
+     *            offset from which to take bytes from data
+     * @param length
+     *            number of bytes to take from data
+     * @param cb
+     *            object implementing callbackinterface
+     * @param ctx
+     *            some control object
+     * @throws ArrayIndexOutOfBoundsException
+     *             if offset or length is negative or offset and length sum to a
+     *             value higher than the length of data.
+     */
+    @Override
+    public void asyncAddEntry(final long entryId, final byte[] data, final int offset, final int length,
+                              final AddCallbackWithLatency cb, final Object ctx) {
+        asyncAddEntry(entryId, Unpooled.wrappedBuffer(data, offset, length), cb, ctx);
+    }
+
     private void asyncAddEntry(final long entryId, ByteBuf data,
-            final AddCallback cb, final Object ctx) {
+            final AddCallbackWithLatency cb, final Object ctx) {
         PendingAddOp op = PendingAddOp.create(this, data, cb, ctx);
         op.setEntryId(entryId);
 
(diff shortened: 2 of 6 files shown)


athanatos:
For a bit of context, I chose to extend the AddCallback interface this way so that code in LedgerHandle/PendingAddOp can't inadvertently call the old addComplete method while still preserving compatibility with existing users.  I wouldn't be surprised if there's a better way to do this, though.

athanatos:
@jvrao Here's the qwcLatency patch.

athanatos:
@sijie Good question.  It would be easy enough to add a write() variant to WriteAdvHandle which returns a CompletableFuture<WriteResult> which has getResult() and getQWC() methods.  If we go in that direction, would it be better to modify the above patch to pass the same WriteResult object rather than adding an argument?  I think the main disadvantage would be yet another allocation?

sijie:
@athanatos 

> If we go in that direction, would it be better to modify the above patch to pass the same WriteResult object rather than adding an argument? 

I think it is fine to have current approach, since you guys might still need this old client for BC purpose. We can create an issue for adding this feature to the new client api. 

> It would be a natural place to add any future information we want to pass out of the write pipeline. 

yes.

> I think the main disadvantage would be yet another allocation?

I think this can be addressed with object pooling. for example, https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/impl/LedgerEntriesImpl.java

sijie:
retest this please

athanatos:
BC isn't really a concern in this case since we'll update the user call site when we switch to this version.  On the other hand, while I'd generally rather take the time to make it future proof (or at least future resistant), I really don't have a good candidate for anything else we'd want to put into a WriteResult object, so simply passing a long would seem to be an ok answer if you're ok with it.

sijie:
@athanatos okay for me

athanatos:
@sijie Added an issue for adding it to the new api.

athanatos:
@sijie Anything else we need to do here?  Are we blocked on implementing the new interface variant?

sijie:
sorry I missed this one. I approved this one. so I need another +1 from other committer to get this merged.

@jiazhai @merlimat @ivankelly @eolivelli @jvrao can anyone take a look and ship this ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandleAdv.java
```diff
@@ -152,8 +153,33 @@ public void asyncAddEntry(final long entryId, final byte[] data, final int offse
...
+     * @param cb
+     *            object implementing callbackinterface
+     * @param ctx
+     *            some control object
+     * @throws ArrayIndexOutOfBoundsException
+     *             if offset or length is negative or offset and length sum to a
+     *             value higher than the length of data.
+     */
+    @Override
+    public void asyncAddEntry(final long entryId, final byte[] data, final int offset, final int length,
```
ivankelly: is it possible to annotate this method with Evolving interface stability?

sijie: this is old api, I don't think we have any annotation on the old api.

ivankelly: Yet we do provide BC guarantees on the old api, and it's also not going away any time soon. Also, the new api isn't stable yet, so for production use people should be using the old API. For this reason we should be careful about what we add to it, and mark new/unstable stuff as such.

sijie: yes. we do provide BC guarantees and we are carefully about any methods adding to the old api. the problem is the old api doesn't any interfaces, those are direct implementation, so we don't have any annotations on that.

ivankelly: There are some interfaces annotated in the supporting classes for old api. Is it possible to use ```@Evolving``` for a single method? Not sure how to generate the javadoc anymore to check, the obvious candidates don't work.

sijie: We only annotated some interfaces that are not gone in future releases, for example ensemble placement policy. For old api, there are no clean interfaces defined, it doesn’t make sense to spend time on annotating them since 1) they are not interfaces, 2) they are going to be deprecated in next release after the yahoo and sales force merge. That was the contract when we introduced annotations. I don’t think it is worth changing that and blocking this change at this moment.

sijie:
@athanatos do you mind rebasing your PR to latest master?  I think it is ready to go once @ivankelly approves it.

athanatos:
@sijie Repushed.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
```diff
@@ -344,7 +347,11 @@ void submitCallback(final int rc) {
         } else {
             addOpLogger.registerSuccessfulEvent(latencyNanos, TimeUnit.NANOSECONDS);
         }
-        cb.addComplete(rc, lh, entryId, ctx);
+        if (lh instanceof LedgerHandleAdv) {
```
ivankelly: This instanceof is unnecessary. addCompleteAdv calls addComplete by default anyhow.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/AsyncCallback.java
```diff
@@ -48,6 +48,24 @@
          *          context object
          */
         void addComplete(int rc, LedgerHandle lh, long entryId, Object ctx);
```
ivankelly: What was saying in the meeting,but wasn't very clear, is that if you add a default to addComplete, then all users of asyncAddEntry can take advantage of this latency reporting (not just the LedgerHandleAdv users).

As it is now, a user would have to override addComplete and addCompleteAdv to use the a latency stuff. If addComplete has a default, then existing code will continue to work, but a user who wants latency information can override only addCompleteAdv, ignoring addComplete.

The addComplete default can be empty, as the bookkeeper code only calls addCompleteAdv. addComplete is _only_ ever called by addCompleteAdv, in the case that addCompleteAdv isn't implemented.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/AsyncCallback.java
```diff
@@ -48,6 +48,24 @@
...
+         * @param lh
+         *          ledger handle
+         * @param entryId
+         *          entry identifier
+         * @param qwcLatency
+         *          QuorumWriteComplete Latency
+         * @param ctx
+         *          context object
+         */
+        default void addCompleteAdv(int rc, LedgerHandle lh, long entryId, long qwcLatency, Object ctx) {
```
ivankelly: I wonder if it's possible to put the Stability.Evolving annotation on a single method.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/AsyncCallback.java
athanatos: Yeah, that's better.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
athanatos: Yeah, will fix.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/AsyncCallback.java
athanatos: @ivankelly Ah, right.  Doing that means that there is no longer an abstract method, so it no longer works as a FunctionalInterface, so lambda expressions don't work.  That would break some existing users.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/AsyncCallback.java
athanatos: Will do

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/AsyncCallback.java
athanatos: @ivankelly I wasn't able to find a good way to do this that didn't break existing users without this particular quirk that you have to implement both.

ivankelly: Oops, meant to reply yesterday. Yes, since it would break the functional stuff, we have to add a new interface. In that case, I'd keep both callbacks completely separate, so that both can be functional.

athanatos:
@ivankelly I've pushed another somewhat more complicated variant I worked up.  The gist of it is that the new functionality is added as a new interface from which AddCallback inherits.  Both remain functional interfaces.  There's a nice upside in that the callback is passed around internally as an AddCallbackWithLatency preventing future changes in the area from accidentally calling the old callback method.

bookkeeper-server/src/test/java/org/apache/bookkeeper/test/BookKeeperClusterTestCase.java
```diff
@@ -92,6 +95,16 @@
 
     private boolean isAutoRecoveryEnabled;
 
+    SynchronousQueue<Throwable> asyncExceptions = new SynchronousQueue<>();
+    protected void captureThrowable(Runnable c) {
+        try {
+            c.run();
+        } catch (Throwable e) {
+            e.printStackTrace();
```
eolivelli: Use logger?

athanatos:
@eolivelli Pushed a fix

sijie:
@eolivelli can you review this again? @athanatos addressed your comment.

sijie:
ping @eolivelli

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/AsyncCallback.java
```diff
@@ -26,6 +26,30 @@
 @InterfaceAudience.Public
 @InterfaceStability.Stable
 public interface AsyncCallback {
+    /**
+     * Async Callback for adding entries to ledgers with latency information.
+     *
+     * @since 4.0
```
eolivelli: Nit: 4.7.0

athanatos:
@eolivelli repushed with fix

asfgit:

sijie:
retest this please

sijie:
all CI job passed except integration tests. the integration tests seems to fail due to can't connect to zookeeper. it seems irrelevant to this change. so ignore integration test ci to merge this change.

sijie:
IGNORE CI

Issue #934, opened 2018-01-03
Additional stats in perChannelBookieClient


kishorekasi:
**FEATURE REQUEST**

1. Please describe the feature you are requesting.
Additional stats to track outstanding WRITE, READ or ALL operations, time taken to connect on a PCBC.

2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have). Are you currently using any workarounds to address this issue?
should-have

PR #971, opened 2018-01-11
Issue 934: Additional stats to track netty latencies in PCBC

kishorekasi:

(@bug W-4058645@) add metrics to measure netty stack latency
- Counters to track outstanding reads/writes operations
- Collect Netty IO rate and latency histograms

@rev ayegorov@

PerChannelBookieClient: add additional stats

Adds:
- exceptionCounter
- connectTimer
- addEntryOutstanding
- readEntryOutstanding
- nettyOpLogger

Signed-off-by: Kishore Kasi Udayashankar<kudayashankar@salesforce.com>

Master Issue: #934


bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
```diff
@@ -861,7 +875,21 @@ private void writeAndFlush(final Channel channel,
         }
 
         try {
-            channel.writeAndFlush(request, channel.voidPromise());
+            final long startTime = MathUtils.nowInNano();
+            ChannelFuture future = channel.writeAndFlush(request, channel.voidPromise());
```
eolivelli: Is this working? With voidPromise?

@merlimat @ivankelly I think not using voidPromise we are going to add an allocation on the hot path

ivankelly: yes, the void promise bit is there to avoid an allocation. it was part of yahoo's GC changes.

merlimat: with `voidPromise` the returned future will be the one from the `voidPromise`, so not very useful :)

kishorekasi:
@eolivelli Fixed the issue with voidPromise(). I see two test failures -
org.apache.bookkeeper.replication.AuditorPeriodicCheckTest.testPeriodicCheckWhenLedgerDeleted
org.apache.bookkeeper.replication.AuditorPeriodicCheckTest.testFailedWriteRecovery

I don't think they are related to my changes. Let me know if you think otherwise.

sijie:
retest this please

kishorekasi:
These tests pass on my local machine. The previous Jenkins run (java 9) is successful too.

PR #974, opened 2018-01-12
Utilities for working with Arquillian and Docker

ivankelly:
This patch contains the following utilities.
- An arquillian StopAction which copy logs from /var/log/bookkeeper to
  the target/ directory.
- An arquillian StopAction which dumps the docker log to the target/
  directory.
- An arquillian AwaitStrategy which checks whether a zookeeper cluster
  is running.
- An arquillian AwaitStrategy which returns immediately (surprising
  arquillian didn't already have this).
- Utilities for working with a cluster of zookeeper/bookkeeper running
  on docker.

Master Issue: #903

PR #975, opened 2018-01-12
Maven classloader

ivankelly:
This will allow us to test different versions of the client.

Also includes some extra methods to make it easier to work with the
bookkeeper client.

Master Issue: #903


tests/integration-tests-utils/pom.xml
```diff
@@ -0,0 +1,95 @@
...
+  <properties>
+    <shrinkwrap.version>3.0.1</shrinkwrap.version>
+    <arquillian-cube.version>1.12.1-SNAPSHOT</arquillian-cube.version>
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-compress</artifactId>
+      <version>1.15</version>
```
eolivelli: maybe it is better to use a property to define the version, like for other deps

tests/integration-tests-utils/pom.xml
```diff
@@ -0,0 +1,95 @@
...
+      <version>${shrinkwrap.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.jboss.shrinkwrap.resolver</groupId>
+      <artifactId>shrinkwrap-resolver-api</artifactId>
+      <version>${shrinkwrap.version}</version>
+    </dependency>
+
+    <dependency>
+      <groupId>org.apache.zookeeper</groupId>
```
eolivelli: I would like to point out that having zookeeper jar on the main classloader will make every client use "this" version of ZK client (or a mix of this an of the one downloaded from Maven).

This may not a problem, because ZK client is quite stable. So for me we can keep this setup if it makes things simpler.
We are already setting up a complex environment, I don't know it if it worth to load ZooKeeper client in BookKeeperClusterUtils.java using our brand new MavenClassLoader and drop this dependency at all here

tests/integration-tests-utils/src/main/java/org/apache/bookkeeper/tests/BookKeeperClusterUtils.java
```diff
@@ -0,0 +1,142 @@
...
+
+    public static String zookeeperConnectString(DockerClient docker) {
+        return DockerUtils.cubeIdsMatching("zookeeper").stream()
+            .map((id) -> DockerUtils.getContainerIP(docker, id)).collect(Collectors.joining(":"));
+    }
+
+    public static ZooKeeper zookeeperClient(DockerClient docker) throws Exception {
+        String connectString = BookKeeperClusterUtils.zookeeperConnectString(docker);
+        CompletableFuture<Void> future = new CompletableFuture<>();
+        ZooKeeper zk = new ZooKeeper(connectString, 10000,
```
eolivelli: maybe it would be difficult to implement a Watcher in Java using MavenClassLoader, so dropping the dependency is not so simple

tests/integration-tests-utils/pom.xml
ivankelly: I ran into this from another angle earlier. It can be solved with a isolated classloader for the surefire test, and passing the systemclassloader as the parent classloader. messes up log4j though, so looking into it more.

sijie:
This should wait #974 to be merged, since the change here is based on #974

tests/integration-tests-utils/pom.xml
ivankelly: done

tests/integration-tests-utils/pom.xml
ivankelly: Latest patch solves this. Tests need to run with forkCount=0 though to get isolated classloader (the other methods for doing so don't seem to work).

eolivelli: great!
what about adding a test that certifies that a ZooKeeper handle created using a MavenClassLoader is not loaded from the boot classloader ?

another issue: please check tests with Java9 too, the hierarchy of main classloaders is not "changed" but the "boot" classloader now does not load every class in the JDK, but only a minimal subset (like java.lang, java.io....)

ivankelly: Added test for zookeeper (for KeeperException, because having to create a watcher would suck).

Jenkins should flag any java9 issue, but I wouldn't expect any. system != boot. system is the classloader that apps usually use. The isolated classloaders are something that surefire does (http://maven.apache.org/surefire/maven-surefire-plugin/examples/class-loading.html).

ivankelly:
Added comment

tests/integration-tests-utils/pom.xml
eolivelli: sounds good to me, thank you

sijie:
@ivankelly I try to resolve conflicts from PR, so CI can kick in and test this. but it doesn't seem to work. can you rebase this? so we can move this forward.

ivankelly:
@sijie whatever you did, it looks good. I would say merge if you're good with the change.

ivankelly:
@sijie @eolivelli going to merge this, all checks have passed, and it already has two reviews.

PR #977, opened 2018-01-12
LedgerHandle: add counters to track ledger->bookie counts

athanatos:
We can use this to detect distribution issues.

(@bug W-4188471@)
[Ported to master, cleaned up a bit, added test]


athanatos:
@jvrao I refactored it slightly to reduce duplication, you might want to glance over it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerCreateOp.java
```diff
@@ -209,6 +215,15 @@ public void operationComplete(int rc, Void result) {
             createComplete(BKException.Code.IncorrectParameterException, null);
             return;
         }
+
+        List<BookieSocketAddress> curEns = lh.getLedgerMetadata().getEnsemble(0L);
+        LOG.info("Ensemble: {} for ledger: {}", curEns, lh.getId());
```
eolivelli: Change log level?

athanatos: We actually like it at info.  @jvrao thoughts?

athanatos: @eolivelli Would it be possible to leave it at info, or are there use cases you are worried about where the volume of ledger creation would be too high?

eolivelli: Ok

Issue #978, opened 2018-01-13
decouple metaformat cmd


reddycharan:
**FEATURE REQUEST**

1. Please describe the feature you are requesting.

'metaformat' command does both nuking and initializing new Bookkeeper cluster in ZK. But having a omnipotent administrative commands with multi-purpose functionality is not desirable and too risky. Moreover there aren't enough checks in 'metaformat' to make sure that we are nuking the right cluster data in ZK which is even more dangerous. Also the current 'metaformat' command doesn't complains if the bookies are connected to the ZK while the 'metaformat' command is being executed, which leads to mismatched instanceids and consequently failure of bookie startup when it is restarted later. Hence we added new set of commands with decoupled functionality and more checks to not to make it error-prone.

- introduce new commands - initnewcluster, nukeexistingcluster, whatisinstanceid
- add testcases initnewcluster and nukeexistingcluster at BookKeeperAdmin level

2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have). Are you currently using any workarounds to address this issue?

must-have feature since in our env. a ZK cluster can be used for multiple Bookie clusters and by other services also. Hence, these BookieShell commands should be more robust and granular.

3. Provide any additional detail on your proposed use case for this feature.

# **Bookkeeper Cluster and Bookie - format / initialization commands**

Previously we were using 'metaformat' command for both nuking and initializing new Bookkeeper cluster in ZK and 'bookieformat' command for nuking the file system data in bookie node and initializing bookie.

But having a omnipotent administrative commands with multi-purpose functionality is not desirable and too risky. Moreover there aren't enough checks in 'metaformat' to make sure that we are nuking the right cluster data in ZK which is even more dangerous. Also the current 'metaformat' command doesn't complains if the bookies are connected to the ZK while the 'metaformat' command is being executed, which leads to mismatched instanceids and consequently failure of bookie startup when it is restarted later. Hence we added new set of commands with decoupled functionality and more checks to not to make it error-prone.

Following commands are introduced and going to be used for nuking and initializing bookkeeper cluster and bookie

* whatisinstanceid
* nukeexistingcluster
* initnewcluster
* bookieformat
* initbookie

The order of the commands to execute to setup a new Bookkeeper cluster is the following

1. Shutdown all the running bookies in the cluster. If we are starting from a fresh state of cluster then we can ignore this step. If any of the bookie process is alive and connected to the ZK then nukeexistingcluster command would fail.
[...]

PR #979, opened 2018-01-13
ISSUE #978: decouple metaformat cmd

reddycharan:

- introduce new commands - initnewcluster, nukeexistingcluster, whatisinstanceid
- add testcases initnewcluster and nukeexistingcluster at BookKeeperAdmin level

Master Issue: #978

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/discover/ZKRegistrationManager.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/discover/ZKRegistrationManager.java
@@ -46,7 +46,6 @@ import org.apache.bookkeeper.bookie.BookieException.MetadataStoreException;
 import org.apache.bookkeeper.client.BookieWatcher;
 import org.apache.bookkeeper.conf.ClientConfiguration;
 import org.apache.bookkeeper.conf.ServerConfiguration;
-import org.apache.bookkeeper.meta.AbstractZkLedgerManager;
 import org.apache.bookkeeper.meta.LayoutManager;
 import org.apache.bookkeeper.meta.LedgerManagerFactory;
 import org.apache.bookkeeper.meta.ZkLayoutManager;
@@ -74,9 +73,6 @@ import org.apache.zookeeper.ZooDefs.Ids;
 import org.apache.zookeeper.ZooKeeper;
 import org.apache.zookeeper.data.ACL;
 import org.apache.zookeeper.data.Stat;
-
-
-
 
 /**
  * ZooKeeper Based {@link RegistrationManager}.
@@ -491,7 +487,6 @@ public class ZKRegistrationManager implements RegistrationManager {
         return true;
     }
 
-
     @Override
     public boolean nukeExistingCluster() throws Exception {
         String zkLedgersRootPath = conf.getZkLedgersRootPath();
@@ -530,45 +525,8 @@ public class ZKRegistrationManager implements RegistrationManager {
             }
         }
 
-        // Format all ledger metadata layout
         LedgerManagerFactory ledgerManagerFactory = LedgerManagerFactory.newLedgerManagerFactory(conf, layoutManager);
-        AbstractZkLedgerManager zkLedgerManager = (AbstractZkLedgerManager) ledgerManagerFactory.newLedgerManager();
-
-        /*
-         * before proceeding with nuking existing cluster, make sure there
-         * are no unexpected znodes under ledgersRootPath
-         */
-        List<String> ledgersRootPathChildrenList = zk.getChildren(zkLedgersRootPath, false);
-        for (String ledgersRootPathChildren : ledgersRootPathChildrenList) {
-            if ((!zkLedgerManager.isSpecialZnode(ledgersRootPathChildren))
-                    && (!zkLedgerManager.isLedgerParentNode(ledgersRootPathChildren))) {
-                log.error("Found unexpected znode : {} under ledgersRootPath : {} so exiting nuke operation",
-                        ledgersRootPathChildren, zkLedgersRootPath);
-                return false;
-            }
-        }
-
-        // formatting ledgermanager deletes ledger znodes
-        ledgerManagerFactory.format(conf, layoutManager);
-
-        // now delete all the special nodes recursively
-        ledgersRootPathChildrenList = zk.getChildren(zkLedgersRootPath, false);
-        for (String ledgersRootPathChildren : ledgersRootPathChildrenList) {
-            if (zkLedgerManager.isSpecialZnode(ledgersRootPathChildren)) {
-                ZKUtil.deleteRecursive(zk, zkLedgersRootPath + "/" + ledgersRootPathChildren);
-            } else {
-                log.error("Found unexpected znode : {} under ledgersRootPath : {} so exiting nuke operation",
-                        ledgersRootPathChildren, zkLedgersRootPath);
-                return false;
-            }
-        }
-
-        // finally deleting the ledgers rootpath
-        zk.delete(zkLedgersRootPath, -1);
-
-        log.info("Successfully nuked existing cluster, ZKServers: {} ledger root path: {}",
-                zkServers, zkLedgersRootPath);
-        return true;
+        return ledgerManagerFactory.validateAndNukeExistingCluster(conf, layoutManager);
     }
 
     @Override
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/discover/ZKRegistrationManager.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/discover/ZKRegistrationManager.java
@@ -36,6 +36,7 @@ import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
 
 import lombok.extern.slf4j.Slf4j;
 
@@ -43,7 +44,10 @@ import org.apache.bookkeeper.bookie.BookieException;
 import org.apache.bookkeeper.bookie.BookieException.BookieIllegalOpException;
 import org.apache.bookkeeper.bookie.BookieException.CookieNotFoundException;
 import org.apache.bookkeeper.bookie.BookieException.MetadataStoreException;
-import org.apache.bookkeeper.client.BookieWatcher;
+import org.apache.bookkeeper.client.BKException;
+import org.apache.bookkeeper.client.BKException.BKInterruptedException;
+import org.apache.bookkeeper.client.BKException.MetaStoreException;
+import org.apache.bookkeeper.common.concurrent.FutureUtils;
 import org.apache.bookkeeper.conf.ClientConfiguration;
 import org.apache.bookkeeper.conf.ServerConfiguration;
 import org.apache.bookkeeper.meta.LayoutManager;
@@ -79,6 +83,18 @@ import org.apache.zookeeper.data.Stat;
  */
 @Slf4j
 public class ZKRegistrationManager implements RegistrationManager {
+
+    private static final Function<Throwable, BKException> EXCEPTION_FUNC = cause -> {
+        if (cause instanceof BKException) {
+            log.error("Failed to get bookie list : ", cause);
+            return (BKException) cause;
+        } else if (cause instanceof InterruptedException) {
+            log.error("Interrupted reading bookie list : ", cause);
+            return new BKInterruptedException();
+        } else {
+            return new MetaStoreException();
+        }
+    };
 
     private ServerConfiguration conf;
     private ZooKeeper zk;
@@ -506,16 +522,19 @@ public class ZKRegistrationManager implements RegistrationManager {
         try (RegistrationClient regClient = new ZKRegistrationClient()) {
             regClient.initialize(new ClientConfiguration(conf), null, NullStatsLogger.INSTANCE, Optional.empty());
             if (availableNodeExists) {
-                Collection<BookieSocketAddress> rwBookies = BookieWatcher.getBookies(regClient);
+                Collection<BookieSocketAddress> rwBookies = FutureUtils
+                        .result(regClient.getWritableBookies(), EXCEPTION_FUNC).getValue();
                 if (rwBookies != null && !rwBookies.isEmpty()) {
                     log.error("Bookies are still up and connected to this cluster, "
                             + "stop all bookies before nuking the cluster");
                     return false;
                 }
+
                 String readOnlyBookieRegPath = availableBookiesPath + "/" + BookKeeperConstants.READONLY;
                 boolean readonlyNodeExists = null != zk.exists(readOnlyBookieRegPath, false);
                 if (readonlyNodeExists) {
-                    Collection<BookieSocketAddress> roBookies = BookieWatcher.getReadOnlyBookies(regClient);
+                    Collection<BookieSocketAddress> roBookies = FutureUtils
+                            .result(regClient.getReadOnlyBookies(), EXCEPTION_FUNC).getValue();
                     if (roBookies != null && !roBookies.isEmpty()) {
                         log.error("Readonly Bookies are still up and connected to this cluster, "
                                 + "stop all bookies before nuking the cluster");
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/discover/ZKRegistrationManager.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/discover/ZKRegistrationManager.java
@@ -26,22 +26,33 @@ import static org.apache.bookkeeper.util.BookKeeperConstants.READONLY;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Charsets;
+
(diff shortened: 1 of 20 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/discover/ZKRegistrationManager.java
```diff
@@ -421,94 +436,187 @@ public LayoutManager getLayoutManager(){
...
-            if (!availableNodeExists) {
-                zk.create(conf.getZkAvailableBookiesPath(), "".getBytes(Charsets.UTF_8),
-                    zkAcls, CreateMode.PERSISTENT);
+    public boolean prepareFormat() throws Exception {
+        boolean ledgerRootExists = null != zk.exists(conf.getZkLedgersRootPath(), false);
+        boolean availableNodeExists = null != zk.exists(conf.getZkAvailableBookiesPath(), false);
+        List<ACL> zkAcls = ZkUtils.getACLs(conf);
+        // Create ledgers root node if not exists
+        if (!ledgerRootExists) {
+            zk.create(conf.getZkLedgersRootPath(), "".getBytes(Charsets.UTF_8), zkAcls, CreateMode.PERSISTENT);
```
athanatos: Why not just call create() and ignore the EEXIST exception?  This variant has almost the same effect, but suffers from a race if the ledgerRoot is created between the check and here.  Same with available bookies node and the READONLY node.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/FlatLedgerManagerFactory.java
```diff
@@ -88,13 +88,14 @@ public LedgerUnderreplicationManager newLedgerUnderreplicationManager()
     @Override
     public void format(AbstractConfiguration conf, LayoutManager layoutManager)
             throws InterruptedException, KeeperException, IOException {
-        String ledgersRootPath = conf.getZkLedgersRootPath();
-        List<String> children = zk.getChildren(ledgersRootPath, false);
-        for (String child : children) {
-            if (FlatLedgerManager.isSpecialZnode(child)) {
-                continue;
+        try (FlatLedgerManager ledgerManager = (FlatLedgerManager) newLedgerManager()) {
```
athanatos: Why not make isLedgerParentNode static?

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/HierarchicalLedgerManager.java
```diff
@@ -86,6 +86,12 @@ protected long getLedgerId(String ledgerPath) throws IOException {
         return StringUtils.stringToLongHierarchicalLedgerId(hierarchicalPath);
     }
 
+    @Override
+    public boolean isSpecialZnode(String znode) {
```
athanatos: This would probably have to be legacyLM.isSpecialZnode(znode) || longLM.isSpecialZnode(znode) || super.isSpecialZnode(znode);

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookKeeperAdminTest.java
```diff
@@ -230,4 +241,200 @@ public void testDecommissionForLedgersWithMultipleSegmentsAndNotWriteClosed() th
         }
         bkAdmin.close();
     }
+
+    @Test(timeout = 60000)
```
jiazhai: Since we have a default timeout setting for all tests, it is suggested to remove `timeout` for individual test case.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/LegacyHierarchicalLedgerManagerFactory.java
```diff
@@ -79,7 +79,7 @@ public LedgerIdGenerator newLedgerIdGenerator() {
 
     @Override
     public LedgerManager newLedgerManager() {
-        return new HierarchicalLedgerManager(conf, zk);
```
jiazhai: oh, the original code seems weird, why it return HierarchicalLedgerManager?

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookKeeperAdminTest.java
```diff
@@ -230,4 +241,200 @@ public void testDecommissionForLedgersWithMultipleSegmentsAndNotWriteClosed() th
...
+        lh.close();
+
+        /*
+         * read instanceId
+         */
+        byte[] data = zkc.getData(newConfig.getZkLedgersRootPath() + "/" + BookKeeperConstants.INSTANCEID, false, null);
+        String readInstanceId = new String(data, UTF_8);
+
+        /*
+         * registrate a RO bookie
```
jiazhai: typo? register

bookkeeper-server/src/main/java/org/apache/bookkeeper/discover/ZKRegistrationManager.java
reddycharan: I didnt touch the implementation logic of this method, I just removed method arg for this method and try block.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/HierarchicalLedgerManager.java
reddycharan: as we discussed, I'm merging all of these isSpecialZnode methods into one method, for the sake of simplicity and clarity.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookKeeperAdminTest.java
reddycharan: fixed it

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookKeeperAdminTest.java
reddycharan: what is the default time? and what if we want more time?

bookkeeper-server/src/main/java/org/apache/bookkeeper/discover/ZKRegistrationManager.java
```diff
@@ -421,94 +436,187 @@ public LayoutManager getLayoutManager(){
...
+        // create readonly bookies node if not exists
+        if (null == zk.exists(conf.getZkAvailableBookiesPath() + "/" + READONLY, false)) {
+            zk.create(conf.getZkAvailableBookiesPath() + "/" + READONLY, new byte[0], zkAcls, CreateMode.PERSISTENT);
+        }
+
+        return ledgerRootExists;
+    }
+
+    @Override
+    public boolean initNewCluster() throws Exception {
```
sijie: the logic here is problematic than the regular format. 

in regular format, it is crashing at the middle of a formatting op, you can re-run format to get it eventually done. in your implementation here, you can't. If I am crashing in the middle of a `init` op, I can't simply re-run `init`.

You might want to document this behavior on those two newly introduced command, to make sure people how to run the commands correctly.

sijie: an alternative way to use `zk.txn` to make sure all the path creations are done in a multi op. but this probably is out of this scope, can be done in a separate task.

bookkeeper-server/src/main/java/org/apache/bookkeeper/discover/ZKRegistrationManager.java
```diff
@@ -421,94 +436,187 @@ public LayoutManager getLayoutManager(){
...
+                zkServers, zkLedgersRootPath);
+
+        boolean ledgerRootExists = null != zk.exists(conf.getZkLedgersRootPath(), false);
+        if (!ledgerRootExists) {
+            log.info("There is no existing cluster with ledgersRootPath: {} in ZKServers: {}, "
+                    + "so exiting nuke operation", zkLedgersRootPath, conf.getZkServers());
+            return true;
+        }
+
+        String availableBookiesPath = conf.getZkAvailableBookiesPath();
```
sijie: The logic here is way too complicated. You don't need to read the list the znodes and check if they are still exists. You can just issue a zk txn (don't remember that exactly method, but the following code is trying to demo how it works), if there are still bookies alive, the txn will fail.

```
Multi multi = zk.multi();
multi.delete(zkLedgersRootPath + "/available/readonly")
multi.delete(zkLedgersRootPath + "/available");

multi.execute();
```

sijie: if you do as what I suggested, you don't need to change BookieWatcher.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookKeeperAdminTest.java
sijie: - default time is more than 60 seconds
- typically if a test is required more than 60 seconds, then we really need to think about how to improve the test. A lot of frakiness in current bookkeeper tests come from unnecessarily `Thread.sleep` or wait for something to trigger at a given time. we should try to avoid that pattern.

bookkeeper-server/src/main/java/org/apache/bookkeeper/discover/ZKRegistrationManager.java
```diff
@@ -59,10 +70,14 @@
...
 import org.apache.zookeeper.Watcher.Event.KeeperState;
 import org.apache.zookeeper.ZKUtil;
+import org.apache.zookeeper.ZooDefs.Ids;
 import org.apache.zookeeper.ZooKeeper;
 import org.apache.zookeeper.data.ACL;
 import org.apache.zookeeper.data.Stat;
 
+
+
+
```
athanatos: Stray whitespace?

reddycharan: fixed it

bookkeeper-server/src/main/java/org/apache/bookkeeper/discover/ZKRegistrationManager.java
reddycharan: for this commit, I would leave as it is

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/LegacyHierarchicalLedgerManagerFactory.java
reddycharan: yes, I do see few other issues with 057af8dbce6c08794eb8b46ca52ca13f222d9bbb . Tried to address in this commit and we noticed other issues as well - Issue #1014

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookKeeperAdminTest.java
reddycharan: removed timeouts

bookkeeper-server/src/main/java/org/apache/bookkeeper/discover/ZKRegistrationManager.java
reddycharan: Here before proceeding with nuking cluster data, I would like to make sure that there are no connected bookies to the cluster and print informative log message for each step/issue. Also, I moved deletions of the ZK nodes to  LedgerManagerFactory class, for the sake better abstraction. Anyhow the changes I made to BookieWatcher are minimal. For now, I feel comfortable to make independent zk operations, because check/delete operations are made at multiple classes with proper validation.

bookkeeper-server/src/main/java/org/apache/bookkeeper/discover/ZKRegistrationManager.java
reddycharan: I added enough logging when these operations fail, for the admin users to know why exactly it failed. Anyhow I'll add little more information to the issue description, regarding usage of these commands.

bookkeeper-server/src/test/java/org/apache/bookkeeper/meta/LedgerManagerIteratorTest.java
```diff
@@ -517,4 +557,20 @@ public void testLedgerManagerFormat() throws Throwable {
         Assert.assertTrue("ChildrenOfLedgersRootPathAfterFormat should contain all the invalid znodes created",
                 childrenOfLedgersRootPathAfterFormat.containsAll(invalidZnodes));
     }
+
+    @Test
+    public void hierarchicalLedgerManagerAsyncProcessLedgersTest() throws Throwable {
```
sijie: nit: better to follow the naming as other methods here, so it is consistent.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractHierarchicalLedgerManager.java
```diff
@@ -63,7 +63,11 @@ void asyncProcessLevelNodes(
         zk.sync(path, new AsyncCallback.VoidCallback() {
             @Override
             public void processResult(int rc, String path, Object ctx) {
-                if (rc != Code.OK.intValue()) {
+                if (rc == Code.NONODE.intValue()) {
```
sijie: These commits here seem not related to "decouple metaformat cmd". I would suggest they should go to a separate issue.

bookkeeper-server/src/main/java/org/apache/bookkeeper/discover/ZKRegistrationManager.java
sijie: > Here before proceeding with nuking cluster data, I would like to make sure that there are no connected bookies to the cluster and print informative log message for each step/issue.

the operation that I proposed is doing exactly same purpose as you hope, with much simpler approach and less error prone.

if you really feel the need to read and delete, you should not use BookieWatch, you should just use registration client api, rather than api in bookie watch. because bookie watch is the user of metadata interfaces, the metadata interface implementation should not use those api, this is a cycle dependency. 

These are the apis you should use: 

https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/discover/RegistrationClient.java#L75
https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/discover/RegistrationClient.java#L82

bookkeeper-server/src/main/java/org/apache/bookkeeper/discover/ZKRegistrationManager.java
sijie: yes. we need enough instructions for people to run the command when they fail at the middle of running this command.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractHierarchicalLedgerManager.java
reddycharan: ok..would move the 3rd commit to a new PR. But Bug1 in Issue #1014, would be fixed in this PR, since it is getting fixed as part of my changes for this PR. But Bug2 in Issue #1014 and testcases for both the bugs will be in new PR (which is currently 3rd commit).

bookkeeper-server/src/main/java/org/apache/bookkeeper/discover/ZKRegistrationManager.java
reddycharan: reverting BookieWatcher changes and going to use RegistrationClient API

reddycharan:
@sijie 

- I removed commit related to Issue #1014 from this PR
- reverted changes made to BookieWatcher, used RegistrationClient instead for getting RW and RO bookies in ZKRegistrationManager
- Will add new commands description/usage to the issue

reddycharan:
@sijie you want me to provide more details to the issue description - https://github.com/apache/bookkeeper/issues/978 ?

sijie:
@reddycharan no. I mean adding more descriptions in the commands. For example, if I failed when running initNewCluster, what I should do next? For users who don’t write the code, they will never not how to do. Because it is unlike meta format , you can run many times. You need to make instructions to ask people to nuke the cluster and reinit the cluster. Either you add the message at help message, or print this when failure happened. But we need something in the tool to tell users.

reddycharan:
done adding suggested description and comments to the command. I think it should be good to go now.

sijie:
@reddycharan can you rebase your branch? it should be ready to go.

reddycharan:
rebased and checks are looking good

reddycharan:
@sijie can you push this PR. Since this change touches quite a few files, I might run into rebasing issue again if it is delayed. Also, I've a bug fix which goes on top of this PR and I'm waiting for this PR to be closed.

I think other PRs of mine are also good to go.

sijie:
@reddycharan sure. I would do it now. sorry I was busy yesterday, so didn't get any chances to go through those PRs.

PR #990, opened 2018-01-16
Fix tight loop in getFileInfo

ivankelly:
The argument to assertions is not evaluated if assertions is disabled,
which was messing up the refcounting for the fileinfo backing
cache. We ended up with dead fileinfo objects in the guava caches,
which triggered an infinite loop.

This patch fixes that by moving the tryRetain() out of the assert. It
also adds defensive code to getFileInfo to ensure that if a fileinfo
object is dead, that it doesn't exist any longer in the caches.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileInfoBackingCache.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileInfoBackingCache.java
@@ -52,7 +52,8 @@ class FileInfoBackingCache {
                 // have been able to get a reference to it here.
                 // The caller of loadFileInfo owns the refence, and is
                 // responsible for calling the corresponding #release().
-                assert(fi.tryRetain());
+                boolean retained = fi.tryRetain();
+                assert(retained);
                 return fi;
             }
         } finally {
@@ -67,7 +68,8 @@ class FileInfoBackingCache {
             fileInfos.put(ledgerId, fi);
 
             // see comment above for why we assert
-            assert(fi.tryRetain());
+            boolean retained = fi.tryRetain();
+            assert(retained);
             return fi;
         } finally {
             lock.writeLock().unlock();
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/IndexPersistenceMgr.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/IndexPersistenceMgr.java
@@ -225,10 +225,12 @@ public class IndexPersistenceMgr {
                     // cache. They shouldn't if refcounting is correct, but if someone
                     // does a double release, the fileinfo will be cleaned up, while
                     // remaining in the cache, which could cause a tight loop in this method.
-                    if (writeFileInfoCache.asMap().remove(ledger, fi)
-                        || readFileInfoCache.asMap().remove(ledger, fi)) {
-                        LOG.error("Dead fileinfo({}) forced out of cache. Must have been double-released somewhere.",
-                                  fi);
+                    boolean inWriteMap = writeFileInfoCache.asMap().remove(ledger, fi);
+                    boolean inReadMap = readFileInfoCache.asMap().remove(ledger, fi);
+                    if (inWriteMap || inReadMap) {
+                        LOG.error("Dead fileinfo({}) forced out of cache (write:{}, read:{}). "
+                                  + "It must have been double-released somewhere.",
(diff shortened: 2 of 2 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/IndexPersistenceMgr.java
```diff
@@ -220,7 +220,19 @@ CachedFileInfo getFileInfo(final Long ledger, final byte masterKey[]) throws IOE
...
                     fi = readFileInfoCache.get(ledger, loader);
                 }
-            } while (!fi.tryRetain());
+                if (!fi.tryRetain()) {
+                    // defensively ensure that dead fileinfo objects don't exist in the
+                    // cache. They shouldn't if refcounting is correct, but if someone
+                    // does a double release, the fileinfo will be cleaned up, while
+                    // remaining in the cache, which could cause a tight loop in this method.
+                    if (writeFileInfoCache.asMap().remove(ledger, fi)
+                        || readFileInfoCache.asMap().remove(ledger, fi)) {
```
eolivelli: if writeFileInfoCache#remove returns true we will miss to check readFileInfoCache#remove (the 'or' will not evaluate the second operand)

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/IndexPersistenceMgr.java
```diff
@@ -220,7 +220,19 @@ CachedFileInfo getFileInfo(final Long ledger, final byte masterKey[]) throws IOE
...
+                        LOG.error("Dead fileinfo({}) forced out of cache. Must have been double-released somewhere.",
```
eolivelli: Is there a way to automatically recover from this situation ? if not I would prefer to crash the bookie and not to run with a broken cache, what do you think @ivankelly  ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileInfoBackingCache.java
```diff
@@ -52,7 +52,8 @@ CachedFileInfo loadFileInfo(long ledgerId, byte[] masterKey) throws IOException
                 // have been able to get a reference to it here.
                 // The caller of loadFileInfo owns the refence, and is
                 // responsible for calling the corresponding #release().
-                assert(fi.tryRetain());
+                boolean retained = fi.tryRetain();
+                assert(retained);
```
eolivelli: if it is important that retained is true we should check it explicitly and crash, even if assertions are disabled

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/IndexPersistenceMgr.java
ivankelly: Oops, fixed.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileInfoBackingCache.java
ivankelly: In this case, retained can only be false if someone screws up some code in this file. Noone else has a reference to fi at this point.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/IndexPersistenceMgr.java
ivankelly: automatic recover is removal from the caches. The fileinfo object is dead, so nothing can be done with it in any case.

ivankelly:
@eolivelli regarding crashing, I don't have a very strong opinion, but lean against it. Any issue here is a programming error, but with the defensive code, it should be safe. I certainly don't want to put a System.exit in here, we'd need to add some interface stuff for crashing, and that's out of scope for this fix.

eolivelli:
@ivankelly 

>  certainly don't want to put a System.exit in here, we'd need to add some interface stuff for crashing, and that's out of scope for this fix.

I agree at 100%

sijie:
@ivankelly can you create a separate issue for cleaning up the usage of `assert`? we'd better not use `assert` in general. for tests, we need to use junit assertion, for other codes, we need to throw exceptions.  also if it is possible, add a checkstyle rule to enforce it.

ivankelly:
@sijie created #995

eolivelli:
merging now on master and branch-4.6, thank you @ivankelly

PR #991, opened 2018-01-16
PendingReadOp: add counter for speculative reads

athanatos:
(@bug W-3324107@)
Signed-off-by: Dustin Castor <dcastorsalesforce.com>
[Ported to current master, added test]


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
```diff
@@ -470,6 +473,7 @@ boolean complete(int bookieIndex, BookieSocketAddress host, ByteBuf buffer) {
         heardFromHostsBitSet = new BitSet(getLedgerMetadata().getEnsembleSize());
 
         readOpLogger = lh.bk.getReadOpLogger();
+        speculativeReadCounter = lh.bk.speculativeReadCounter;
```
eolivelli: Better use a getter, it simplifies mocking

athanatos: Alright.

athanatos:
@eolivelli Pushed fix.

PR #993, opened 2018-01-16
PendingReadOp: improve logging for speculative read targets

athanatos:
And add more log messages around speculative read
to understand what bookies were tried, before attempting
speculative read.

(@bug W-3315760)
[Fixed merge conflicts and checkstyle issues]

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
@@ -61,8 +61,9 @@ class PendingReadOp implements ReadEntryCallback, SafeRunnable {
     private ScheduledFuture<?> speculativeTask = null;
     protected final List<LedgerEntryRequest> seq;
     private final CompletableFuture<LedgerEntries> future;
-    Set<BookieSocketAddress> heardFromHosts;
-    BitSet heardFromHostsBitSet;
+    private final Set<BookieSocketAddress> heardFromHosts;
+    private final BitSet heardFromHostsBitSet;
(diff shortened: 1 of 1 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
```diff
@@ -254,9 +255,10 @@ public String toString() {
                 @Override
                 public Boolean call() throws Exception {
                     if (!isComplete() && null != maybeSendSpeculativeRead(heardFromHostsBitSet)) {
-                        if (LOG.isDebugEnabled()) {
-                            LOG.debug("Send speculative read for {}. Hosts heard are {}, ensemble is {}.",
-                                    this, heardFromHostsBitSet, ensemble);
+                        if (LOG.isInfoEnabled()) {
```
sijie: Are we sure changing this from debug to info? At a high load cluster, you might have 5-10% or even more traffic do speculative reads to ensure predictable latency. This change will result in producing a lot of unnecessary logging. 

May I ask you a question? How useful this logging helps you guys at production?

athanatos: It's been pretty useful for us.  Perhaps we should simply bump logging to debug in the few classes where we want it. @jvrao thoughts?

sijie: yeah. that would be ideal, you can configure your log4j properties to enable debug in PendingReadOp.

athanatos:
@sijie Switched to debug.

PR #996, opened 2018-01-17
DoubleByteBuf fix for Netty > 4.1.12

eolivelli:
This is a port from Pulsar, DoubleByteBuf has problems with Netty Native Transport.
Original author is  @sschepens 

see https://github.com/apache/incubator-pulsar/pull/1056

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/util/DoubleByteBuf.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/util/DoubleByteBuf.java
@@ -17,6 +17,7 @@
 */
 package org.apache.bookkeeper.util;
 
+import com.google.common.collect.ObjectArrays;
 import io.netty.buffer.AbstractReferenceCountedByteBuf;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufAllocator;
@@ -361,7 +362,7 @@ public final class DoubleByteBuf extends AbstractReferenceCountedByteBuf {
 
     @Override
     public ByteBuffer nioBuffer(int index, int length) {
-        ByteBuffer dst = ByteBuffer.allocate(length);
+        ByteBuffer dst = isDirect() ? ByteBuffer.allocateDirect(length) : ByteBuffer.allocate(length);
         ByteBuf b = Unpooled.wrappedBuffer(dst);
         b.writerIndex(0);
         getBytes(index, b, length);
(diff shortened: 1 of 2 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/util/DoubleByteBuf.java
```diff
@@ -370,7 +371,10 @@ public ByteBuffer nioBuffer(int index, int length) {
 
     @Override
     public ByteBuffer[] nioBuffers(int index, int length) {
-        return new ByteBuffer[] { nioBuffer(index, length) };
+        if (b1.nioBufferCount() == 1 && b2.nioBufferCount() == 1) {
+            return new ByteBuffer[] { b1.nioBuffer(), b2.nioBuffer() };
+        }
+        return ObjectArrays.concat(b1.nioBuffers(), b2.nioBuffers(), ByteBuffer.class);
```
eolivelli: with this change we are no more observing index and length parameters
@merlimat  @ivankelly  @sschepens

sschepens: yup, this is incorrect, seems I failed somewhere, this logic has to be in `public ByteBuffer[] nioBuffers()`, gonna change it

sschepens: Just checked and it's OK on the Pulsar, seems you modified the incorrect method :smile:

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/DoubleByteBuf.java
```diff
@@ -361,7 +362,7 @@ public ByteBuffer internalNioBuffer(int index, int length) {
 
     @Override
     public ByteBuffer nioBuffer(int index, int length) {
-        ByteBuffer dst = ByteBuffer.allocate(length);
+        ByteBuffer dst = isDirect() ? ByteBuffer.allocateDirect(length) : ByteBuffer.allocate(length);
```
sschepens: this is the only really necessary change, the other changes are just optimizations

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/DoubleByteBuf.java
eolivelli: Oh thank you @sschepens  I have messed up with the diff, thank you very much

sschepens:
:+1:

Issue #1003, opened 2018-01-18
Bookkeeper-server shaded artifact test fail on branch-4.6


eolivelli:
if I run on branch-4.6
`mvn clean install`

I get this error:

```
Running org.apache.bookkeeper.tests.shaded.BookKeeperServerShadedArtifactTest
Tests run: 6, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.104 sec <<< FAILURE! - in org.apache.bookkeeper.tests.shaded.BookKeeperServerShadedArtifactTest
testProtobufIsShaded(org.apache.bookkeeper.tests.shaded.BookKeeperServerShadedArtifactTest)  Time elapsed: 0.005 sec  <<< FAILURE!
java.lang.AssertionError: Expected exception: java.lang.ClassNotFoundException


Results :

Failed tests: 
  BookKeeperServerShadedArtifactTest.testProtobufIsShaded Expected exception: java.lang.ClassNotFoundException
```

eolivelli:
odd fact:
if I run the tests inside this directory, all runs fine
bookkeeper/tests/backward

I think it is due to the use of Maven Reactor

when running tests inside "tests" all modules will load dependencies from "target/classes" directory, which is empty for bookkeeper-server-compat-4.0.0 (and other directories)

when running a single project dependencies are taken from .m2

eolivelli:
@sijie do you think this is a blocker for release ?

Somehow tests pass, it is a problem of the global configuration of the poms and we are refactoring all of this in 4.7

eolivelli:
if I run 
`mvn clean install`
the problems does not occur on backward tests, so it is not a show stopper for release.

Anyway I get another problem and this a blocker, to be fixed
```
Running org.apache.bookkeeper.tests.shaded.BookKeeperServerShadedArtifactTest
Tests run: 6, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.104 sec <<< FAILURE! - in org.apache.bookkeeper.tests.shaded.BookKeeperServerShadedArtifactTest
testProtobufIsShaded(org.apache.bookkeeper.tests.shaded.BookKeeperServerShadedArtifactTest)  Time elapsed: 0.005 sec  <<< FAILURE!
java.lang.AssertionError: Expected exception: java.lang.ClassNotFoundException


Results :

Failed tests: 
  BookKeeperServerShadedArtifactTest.testProtobufIsShaded Expected exception: java.lang.ClassNotFoundException

[...]

PR #1006, opened 2018-01-18
Issue-1003 bookkeeper-server shaded artifact test fail on branch-4.6

eolivelli:
Fix the pom in order to clean up compatibility tests for branch-4.6. BookKeeper server shaded stuff was picked from 4.7 branch but the structure of the project is very different so the dependency tree of test modules is not exactly the same

### DIFF
--- a/tests/bookkeeper-server-shaded-artifact-test/pom.xml
+++ b/tests/bookkeeper-server-shaded-artifact-test/pom.xml
@@ -41,6 +41,13 @@
           <groupId>org.apache.bookkeeper</groupId>
           <artifactId>bookkeeper-proto</artifactId>
         </exclusion>
+        <exclusion>
+          <!-- actually on branch 4.6 the dependency tree is different from 4.7, in which we have bookkeeper-proto module
+               in order to make tests pass on 4.6 we have to exclude this dependency manually
+               it is only important that the main artifact bookkeeper-server:shaded:jar contains the relocated version of protobuf -->
+          <groupId>com.google.protobuf</groupId>
+          <artifactId>protobuf-java</artifactId>
+        </exclusion>
       </exclusions>
     </dependency>
   </dependencies>


eolivelli:
@sijie I think that the problem is the same for which we have introduced bookkeeper-server-shaded artifact (instead of bookkeeper-server with classifier 'shaded')

It is a bug on Maven Shade Plugin, even if updating to latest release of that plugin the result is the same.

In fact if I use the bookkeeper-server-shaded artifact (the one which is similar to bookkeeper-server in 4.5.x) there is no issue.

We are not able to drop from the bookkeeper-server:shaded artifact the dependency from protobuf.
On master branch we have bookkeper-proto indirection and for some reason it works as expected

sijie:
@eolivelli where are we using `bookkeeper-server:shaded` artifact? we never use this artifact internally.

eolivelli:
It is used only in tests for that artifact.
I find no value in that artifact.
We should test the bookkeeper-server-shaded artifact

eolivelli:
Sorry Sijie wrong comment

eolivelli:
@sijie I think we are using bookkeeper-server with classifier shaded here 
https://github.com/apache/bookkeeper/blob/c3ef133edc88aa1a9d457909eedcdf4ec0e62169/tests/bookkeeper-server-shaded-artifact-test/pom.xml#L33

I can't remember the value of such artifact, it is mostly like bookkeeper-server-shaded but the pom is broken because it retains dependency to protobuf even it is shaded inside the jar 

We have tests for bookkeeper-server-shaded and they are OK
https://github.com/apache/bookkeeper/blob/c3ef133edc88aa1a9d457909eedcdf4ec0e62169/tests/bookkeeper-server-shaded-test/pom.xml#L31

eolivelli:
@sijie @ivankelly @merlimat  anyway my interest is about the ability of cutting a release from branch 4.6.
If think that this test failure is not a blocker, because it tests something that is not really important, I think I can roll the RC soon

sijie:
@eolivelli : 

The fix is correct. But we need to update the description of issuse #1003 and the pull request here to reflect what is the real problem here. otherwise, it is unclear for people to revisit this in future.

I have two comments below.

- it is not BC tests failing, it is the shaded-artifact test is failing. we need to update the description of #1003, because the caption is misleading.

- the reason why it is failing, because the shaded-artfiact doesn't generate reduced pom. it is a classified artifact. we need to exclude dependencies for this test module. protobuf was excluded when we exclude `bookkeeper-common` in master. However in 4.6, we have the direct protobuf dependency in bookkeeper-server, so when this change is cherry-picked back to branch-4.6, the protobuf is not correctly excluded. it is the same problem as when we imported circe-checksum in #935. 

Can you update the description, and update the caption of #1003? Then this change is ready to go, you are unblocked for release 4.6.1.

eolivelli:
I will do tomorrow, thank you to everybody worked on this problem

ivankelly:
what is the purpose of the "shaded" qualified artifact, if we have another module for shaded anyhow?

sijie:
@ivankelly 

the bookkeeper-server was by default shaing protobuf and guava, which is not ideally for projects like distributedlog/pulsar, which has tight connection. so I changed bookkeeper-server to not shade by default, by using a classifier to shade it to a `shaded` artifact. 

but the `shaded` artifact doesn't generate a reduced pom, because it is a classified artifact of `bookkeeper-server`. but it is published in 4.6.0, and we realized it only after 4.6.0 was release.

so to better overall come it, I added `bookkeeper-server-shaded` module recently for 4.7.0 and add test cases to catch that. and those changes were backported to 4.6.1.

we can remove `shaded` artifact in 4.7.0 since `bookkeeper-server-shaded` is there. but we'd better to keep `shaded` artifact in 4.6.1, because 4.6.1 is supposed to a bug fix for 4.6.0, which we shouldn't change the release artifacts, it should be releasing same artifacts as 4.6.0. 

Hope this make things clear.

ivankelly:
Ah, makes sense. Thanks @sijie

eolivelli:
@sijie I have updated the description and cleaned up the history

eolivelli:
closed with https://github.com/apache/bookkeeper/commit/807678c78468db3d40c7a2ccfc9c0cdbb1cc5f66

eolivelli:
@ivankelly sorry, I have included you in the list of reviewers for this commit in the commit message, but you did not give explicit approval

ivankelly:
no worries

PR #1012, opened 2018-01-18
[Merge Yahoo repo] CMS-1437: use BookieSocketAddress in DNSToSwitchMapping

jiazhai:

use BookieSocketAddress in DNSToSwitchMapping.
Change-Id: Id74b6d31176f6e34f0cd148eae35adc7da2f3808
Here is original commit:
https://github.com/yahoo/bookkeeper/commit/569fff21

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/net/DNSToSwitchMapping.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/net/DNSToSwitchMapping.java
@@ -29,25 +29,24 @@ import java.util.List;
 @Beta
 public interface DNSToSwitchMapping {
     /**
-     * Resolves a list of DNS-names/IP-addresses and returns back a list of
+     * Resolves a list of {@link BookieSocketAddress} and returns back a list of
      * switch information (network paths). One-to-one correspondence must be
      * maintained between the elements in the lists.
      * Consider an element in the argument list - x.y.com. The switch information
      * that is returned must be a network path of the form /foo/rack,
      * where / is the root, and 'foo' is the switch where 'rack' is connected.
-     * Note the hostname/ip-address is not part of the returned path.
+     * Note the {@link BookieSocketAddress} is not part of the returned path.
      * The network topology of the cluster would determine the number of
      * components in the network path.
-     *
-     * <p>If a name cannot be resolved to a rack, the implementation
-     * should return {@link NetworkTopology#DEFAULT_REGION_AND_RACK}. This
+     * If a {@link BookieSocketAddress} cannot be resolved to a rack, the implementation
+     * should return {@link NetworkTopology#DEFAULT_RACK}. This
      * is what the bundled implementations do, though it is not a formal requirement
      *
-     * @param names the list of hosts to resolve (can be empty)
+     * @param bookieAddressList the list of hosts to resolve (can be empty)
      * @return list of resolved network paths.
-     * If <i>names</i> is empty, the returned list is also empty
+     * If <i>bookieAddressList</i> is empty, the returned list is also empty
      */
-    List<String> resolve(List<String> names);
+    List<String> resolve(List<BookieSocketAddress> bookieAddressList);
 
     /**
      * Reload all of the cached mappings.
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/net/NetUtils.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/net/NetUtils.java
@@ -21,18 +21,18 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Preconditions.checkState;
 
 import java.net.InetAddress;
-import java.net.InetSocketAddress;
 import java.net.UnknownHostException;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
-
+import lombok.extern.slf4j.Slf4j;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
  * Network Utilities.
  */
+@Slf4j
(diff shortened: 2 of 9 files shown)


sijie:
hmm, I think this change will break BC, because it changes the interface of DNSToSwitchMapping, which is allowed to be provided by users when constructing the bookkeeper.

@merlimat @sboobna do you guys have any more context about this change?

bookkeeper-server/src/main/java/org/apache/bookkeeper/net/NetUtils.java
```diff
@@ -68,18 +66,13 @@ public static String normalizeHostName(String name) {
         return hostNames;
     }
 
-    public static String resolveNetworkLocation(DNSToSwitchMapping dnsResolver, InetSocketAddress addr) {
-        List<String> names = new ArrayList<String>(1);
-
-        if (dnsResolver.useHostName()) {
```
sijie: I didn't see where this logic is moved to. I think this change will also break `useHostName()`. and the tests doesn't catch this, so it means that we are missing some tests to cover this. we need to add a test to cover `useHostName`.

bookkeeper-server/src/main/java/org/apache/bookkeeper/net/DNSToSwitchMapping.java
```diff
@@ -29,25 +29,24 @@
...
      * is what the bundled implementations do, though it is not a formal requirement
      *
-     * @param names the list of hosts to resolve (can be empty)
+     * @param bookieAddressList the list of hosts to resolve (can be empty)
      * @return list of resolved network paths.
-     * If <i>names</i> is empty, the returned list is also empty
+     * If <i>bookieAddressList</i> is empty, the returned list is also empty
      */
-    List<String> resolve(List<String> names);
+    List<String> resolve(List<BookieSocketAddress> bookieAddressList);
```
sijie: This breaks BC. It would be good to add a default method and mark the old method as deprecated.

```
default List<String> resolve(List<BookieSocketAddress> bookieAddressList) {

    return Lists.transform(bookieAddressList, bookieAddr -> {
        if (useHostName) {
            return bookieAddr.getSocketAddress().getHostName();
        } else {
            return bookieAddr.getSocketAddress().getAddress().getHostAddress();
        }
    });
    

}
```

jiazhai: Thanks. will do it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/net/NetUtils.java
jiazhai: Thanks, this is following the original change. 
Since BookieSocketAddress contains the hostname and InetSocketAddress, Do we still need this logic?

jiazhai: OK. Thanks. I see. it is still BC issue.

bookkeeper-server/src/main/java/org/apache/bookkeeper/net/DNSToSwitchMapping.java
jiazhai: @sijie have both `List<String> resolve(List<BookieSocketAddress> bookieAddressList)` and `List<String> resolve(List<String> names)` will report `both methods have same erasure`. 

@merlimat 
How about rename  to another name for new method:
`List<String> resolveByAddresses(List<BookieSocketAddress> bookieAddressList)`

jiazhai:
@merlimat @sboobna  For BC issue, have same question with @sijie , How is the new method in DNSToSwitchMapping used in Pulsar?  
Is it OK to keep the old method name `List<String> resolve(List<BookieSocketAddress> bookieAddressList)` and add new method with new name `List<String> resolveAddress(List<String> names)`?  because same name "resolve" will cause "both methods have same erasure".

sijie:
@merlimat @sboobna can any of you take a look at @jiazhai 's comment? It would be good to understand the motivation of this change and how is this used by pulsar.

bookkeeper-server/src/main/java/org/apache/bookkeeper/net/DNSToSwitchMapping.java
merlimat: +1 for changing method name and providing default implementation

merlimat:
@sijie From the original issue: 

> Currently, the rackaware policy passes host address to the CachedDNSToSwitchMapping and host name to any other implementation of DNSToSwitchMapping. To be uniform, It should always use BookieSocketAddress.

I think in one place the hostname was used vs the IP in the other and that was leading to no caching.

sijie:
@merlimat 

> I think in one place the hostname was used vs the IP in the other and that was leading to no caching.

if that is the case, I don't think we need this change before in master `useHostName` was introduced to DNSToSwitchMapping. It ensures the resolver either gets ip or hostname. 

https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/net/DNSToSwitchMapping.java#L67

jiazhai:
Yes, and it is has to keep BC `useHostName`, while introduce BookieSocketAddress

sijie:
@merlimat can you check if we still need this PR?

asfgit:

sijie:
@merlimat it seems that we don't need this change any more, shall we close it or you want to wait until pulsar completes bumping the bookkeeper version?

ivankelly:
@jiazhai @sijie @merlimat Can we close this in the end? That yellow line on the spreadsheet is bugging me.

jiazhai:
will close it.

PR #1015, opened 2018-01-18
ZkRegistrationManager: add log line with root path for prepareFormat

athanatos:
Useful for auditability.

(@bug W-4026563@)
Signed-off-by: Zhimeng Shi <zshi@salesforce.com>
[Updated for current patch]


jvrao:
+1

PR #1018, opened 2018-01-19
Issue 1007: move checksum to proto

ArvinDevel:

Refactor checksum to proto
Master Issue: #<1007>

> ---
> quickly and easily:
>
>
>
> Otherwise:
> 
> 
> ---


ArvinDevel:
@eolivelli I want to replace the DigestType in client.api with DigestType in proto.checksum, because keep two is more redundant, do you think is it ok?

sijie:
@ArvinDevel 

I saw your question in the other PR. but didn't get a chance to reply. I will reply here:

`client.api` is the new API package, which is going to the bookkeeper api in future. the old api will be deprecated in future releases soon. so now we have 3 types of DigestType in the project.

`client.api.DigestType` is the user facing one, `bookkeeper.DigestType' is from the old one, `proto.DigestType` is the one used for storing the metadata. you can't change the `bookkeeper.DigestType` in the old api, because it is there for backward compatability. so I would suggest:

- leave `client.api.DigestType` there, because it is the user facing api.
- leave `bookkeeper.DigestType` there for BC reason. (it will be gone soon).
- in the remaining places, use the `proto.DigestType`, which means DigestManager uses `proto.DigestType`. 

does this make sense to you?

ArvinDevel:
make sense, this change follows the contract. As long as there is a solid contract, it will be fine. As for the microbenchmark test, the generated class has lots of un-used fields, which causes the spotbuts-check fail, how to fix it? I'm curious that why the generated class before this change doesn't causes check fail, which also contain un-used fields.

eolivelli:
Thank you @sijie for your clear explanation. I totally agree with your words

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/DigestType.java
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * DigestType for checksum.
+ * There are 3 digest types that can be used for verification. The CRC32 is
+ * cheap to compute but does not protect against byzantine bookies (i.e., a
+ * bookie might report fake bytes and a matching CRC32). The MAC code is more
+ * expensive to compute, but is protected by a password, i.e., a bookie can't
+ * report fake bytes with a mathching MAC unless it knows the password.
+ * The CRC32C, which use SSE processor instruction, has better performance than CRC32.
+ */
+public enum DigestType {
```
sijie: I would suggest not adding another `DigestType`, just reuse the [protobuf one](https://github.com/apache/bookkeeper/blob/master/bookkeeper-proto/src/main/proto/DataFormats.proto#L45), since the digest type here is used by digest manager, which is kind of "internal" to users.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeper.java
```diff
@@ -653,7 +653,9 @@ public RegistrationClient getRegClient() {
      * cheap to compute but does not protect against byzantine bookies (i.e., a
      * bookie might report fake bytes and a matching CRC32). The MAC code is more
      * expensive to compute, but is protected by a password, i.e., a bookie can't
-     * report fake bytes with a mathching MAC unless it knows the password
+     * report fake bytes with a mathching MAC unless it knows the password.
+     * legacy DigestType for back compatible,
+     * new added DigestType should add it in both here and proto.checksum.DigestType.
```
eolivelli: Nit: I think this new lines are not well written.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeper.java
```diff
@@ -653,7 +653,9 @@ public RegistrationClient getRegClient() {
      * cheap to compute but does not protect against byzantine bookies (i.e., a
      * bookie might report fake bytes and a matching CRC32). The MAC code is more
      * expensive to compute, but is protected by a password, i.e., a bookie can't
-     * report fake bytes with a mathching MAC unless it knows the password
+     * report fake bytes with a mathching MAC unless it knows the password.
+     * legacy DigestType for back compatible,
```
eolivelli: Typo: back compatible -> backward compatibility

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/DigestManager.java
```diff
@@ -210,14 +210,17 @@ long verifyDigestAndReturnLac(ByteBuf dataReceived) throws BKDigestMatchExceptio
...
         verifyDigest(entryId, dataReceived);
         dataReceived.readerIndex(METADATA_LENGTH + macCodeLength);
         return dataReceived;
     }
 
-    static class RecoveryData {
+    /**
+     * A representation of RecoveryData.
+     */
+    public static class RecoveryData {
```
eolivelli: Maybe we can add 'final' here

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/DigestManager.java
```diff
@@ -226,9 +229,17 @@ public RecoveryData(long lastAddConfirmed, long length) {
             this.length = length;
         }
 
+        public long getLastAddConfirmed() {
+            return lastAddConfirmed;
```
eolivelli: Make all fields final?

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/MockBookKeeperTestCase.java
```diff
@@ -333,7 +336,11 @@ protected void setupBookieClientReadEntry() {
             long entryId = (Long) args[3];
 
             executor.submitOrdered(ledgerId, () -> {
-                DigestManager macManager = new CRC32DigestManager(ledgerId);
+                DigestManager macManager = null;
+                try {
+                    macManager = DigestManager.instantiate(ledgerId, new byte[2], DigestType.CRC32);
+                } catch (GeneralSecurityException gse){
```
eolivelli: Rethrow or at least log

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/MockBookKeeperTestCase.java
```diff
@@ -363,7 +370,11 @@ protected void setupBookieClientReadEntry() {
                 (BookkeeperInternalCallbacks.ReadEntryCallback) args[3];
 
             executor.submitOrdered(ledgerId, () -> {
-                DigestManager macManager = new CRC32DigestManager(ledgerId);
+                DigestManager macManager = null;
+                try {
+                    macManager = DigestManager.instantiate(ledgerId, new byte[2], DigestType.CRC32);
+                } catch (GeneralSecurityException gse){
```
eolivelli: Same add above

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/MockBookKeeperTestCase.java
```diff
@@ -387,7 +398,11 @@ protected void setupBookieClientReadEntry() {
             throws BKException.BKDigestMatchException {
         ByteBuf toSendCopy = Unpooled.copiedBuffer(toSend);
         toSendCopy.resetReaderIndex();
-        DigestManager macManager = new CRC32DigestManager(ledgerId);
+        DigestManager macManager = null;
+        try {
+            macManager = DigestManager.instantiate(ledgerId, new byte[2], DigestType.CRC32);
+        } catch (GeneralSecurityException gse){
```
eolivelli: Same as above

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/DoubleByteBuf.java
```diff
@@ -17,7 +17,6 @@
 */
 package org.apache.bookkeeper.util;
 
-import com.google.common.collect.ObjectArrays;
```
eolivelli: This will need rebase

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/ClientUtil.java
```diff
@@ -31,7 +34,13 @@ public static ByteBuf generatePacket(long ledgerId, long entryId, long lastAddCo
 
     public static ByteBuf generatePacket(long ledgerId, long entryId, long lastAddConfirmed,
                                                long length, byte[] data, int offset, int len) {
-        CRC32DigestManager dm = new CRC32DigestManager(ledgerId);
+        DigestManager dm = null;
+        try {
+            // CRC32 need not password, and is support definitely currently.
+            dm = DigestManager.instantiate(ledgerId, new byte[2], CRC32);
+        } catch (GeneralSecurityException gse) {
+
```
eolivelli: Rethrow or at least log

ArvinDevel:
@sijie @eolivelli I have fixed all the comments.

sijie:
@eolivelli please review this since you were in the PR thread before.

sijie:
@eolivelli please review this when you have time.

jvrao:
+1 for this; Do we have a separate issue to verify the checksum in bookie write path?

sijie:
@jvrao I created #1046 for adding verify logic in the bookie write path.

@ArvinDevel wondering if you are interested in adding this feature.

ArvinDevel:
okay, I'll try my best to add this feature.

PR #1020, opened 2018-01-19
Utilities for bookies running in test containers

ivankelly:
The changes include:
- Updating configuration in bk_server.conf for one or all bookies
- Run commands on all or one bookie
- Wait for all bookies in the test are up

There are also improvements to DockerUtils#runCommand:
- Better logging
- Catches RunTime exceptions, as Docker-java throws them and it could
  fail the test
- Throw an exception is the command exits with anything but 0

Master Issue: #903


ivankelly:
Jenkins is flakes in one test each.

Travis waiting on OSX. Merging now.

PR #1021, opened 2018-01-19
Fixups for broken versions in all versions images

ivankelly:
Various versions of bookkeeper have shipped broken by default. This
patch fixes them up in the all versions image, so that they can be
tested.

Master Issue: #903


ivankelly:
Jenkins java8 is green. Travis waiting on OSX, linux is green. Merging so I can push some more changes.

PR #1022, opened 2018-01-19
Only run integration tests with -DintegrationTests

ivankelly:
Integration tests have a higher system requirement than normal unit
tests, and take longer to run, so they should be explicitly enabled.

Master Issue: #903

### DIFF
--- a/tests/integration-tests-base/pom.xml
+++ b/tests/integration-tests-base/pom.xml
@@ -63,6 +63,8 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-surefire-plugin</artifactId>
         <configuration>
+          <!-- only run tests when -DintegrationTests is specified //-->
+          <skipTests>true</skipTests>
           <systemPropertyVariables>
             <currentVersion>${project.version}</currentVersion>
             <maven.buildDirectory>${project.build.directory}</maven.buildDirectory>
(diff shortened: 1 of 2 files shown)


tests/integration-tests-base/pom.xml
```diff
@@ -72,4 +74,25 @@
     </plugins>
   </build>
 
+  <profiles>
```
sijie: How does this work if you don't specify in the root pom.xml?

Can you show an example command?

tests/integration-tests-base/pom.xml
```diff
@@ -63,6 +63,8 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-surefire-plugin</artifactId>
         <configuration>
+          <!-- only run tests when -DintegrationTests is specified //-->
+          <skipTests>true</skipTests>
```
eolivelli: Wouldn't it be better to use failsafe and do not run on test goal but with verify?

sijie: good point. +1

tests/integration-tests-base/pom.xml
ivankelly: mvn test -DintegrationTests

tests/integration-tests-base/pom.xml
ivankelly: I looked a bit at failsafe. Does verify need to be called explicitly? As opposed to test which runs with package and install?

eolivelli: It has been designed expecially for integration tests I suppose.

Even using -DintegrationTests needs explicit request from user.

Maybe it is better to:
1) use directly a profile, -PintegrationTests (you can activate more than one profile on the command line)
2) use failsafe and so the verify command

To be honest I have never used failsafe but as far as I know it is the 'maven way'.

ivankelly: there is a profile "integrationTests", so it can be activated by -P, I created the activate by system property so that docker would also be activated.

I _want_ it to be explicit for the user to run these tests. I don't want users to have to have docker installed to be able to get through a "mvn install".

sijie:
@eolivelli can you review this PR again? so we can move forward with it.

ivankelly:
Merging.

Re: failsafe vs surefire, after reading a bit, it doesn't make a difference here. Arquillian takes care of setup and teardown before and after the suite, so the main usecase for failsafe doesn't exist. The only other difference is that it runs during a different stage, which doesn't have an effect here since the tests will be separate modules anyhow.

PR #1023, opened 2018-01-19
Arquillian topologies module

ivankelly:
Will allow topologies to be shared among modules, which is important
as arquillian builds and pulls down the containers on a per
module/suite basis.

Currently contains a single topology, which is 3 BK nodes and 1 ZK
nodes. The BK nodes contain all released versions, with no version
started by default.

Master Issue: #903

### DIFF
--- a/tests/integration-tests-topologies/src/main/resources/cube-definitions/3-node-all-version-unstarted.yaml
+++ b/tests/integration-tests-topologies/src/main/resources/cube-definitions/3-node-all-version-unstarted.yaml
@@ -1,0 +1,64 @@
+#/**
+# * Copyright 2007 The Apache Software Foundation
+# *
+# * Licensed to the Apache Software Foundation (ASF) under one
+# * or more contributor license agreements.  See the NOTICE file
+# * distributed with this work for additional information
+# * regarding copyright ownership.  The ASF licenses this file
+# * to you under the Apache License, Version 2.0 (the
+# * "License"); you may not use this file except in compliance
+# * with the License.  You may obtain a copy of the License at
+# *
+# *     http://www.apache.org/licenses/LICENSE-2.0
+# *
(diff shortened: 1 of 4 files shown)


ivankelly:
@sijie @eolivelli @jiazhai This and #1022 are the last two PRs for support code before I start pushing the tests themselves. See #930 for the tests. I'll push them one by one so they're easier to review.

tests/integration-tests-topologies/src/main/resources/cube-definitions/3-node-all-version-unstarted.yaml
```diff
@@ -0,0 +1,64 @@
...
+# * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# * See the License for the specific language governing permissions and
+# * limitations under the License.
+# */
+networks:
+  testnetwork*:
+    driver: bridge
+
+zookeeper*:
+  image: jplock/zookeeper
```
sijie: nit: any reason why no use an "official" image? shall we include the version here, otherwise, it will fetch the latest image, no?

ivankelly: ah, last time I looked there wasn't an official one, and jplock was the most popular, so it imprinted on my brain. Will use official.

eolivelli:
Last question...why are you using 4.x branch ? We have 3.5.3.
I don't know other companies but I am running 3.5 in production.
I know we are only using only features of 4.x on the wire, but maybe it is better to have 3.5

sijie:
@eolivelli I think 3.4.x is okay and mostly deployed on the production. the only requirement is to have a version to make the behavior deterministic. whether to use 3.4.x or 3.5.x is not important here.

Issue #1024, opened 2018-01-19
BP-26: Move the distributedlog library to bookkeeper


sijie:
**BP**

This is the master ticket for tracking BP-26 :

DistributedLog is an extension of Apache BookKeeper, which offers *reopenable* log streams as its storage primitives. It is tightly built over bookkeeper ledgers, and provides an easier-to-use abstraction and api to use. Applications can use *named* log streams rather than *numbered* ledgers to store their data. For example, users can use log streams as files to storge objects, checkpoints and other more general filesystem related use cases.

This BP proposes moving the distributedlog library as part of bookkeeper.

Proposal PR - #1025 

- tasks:
- [x] copy the distributedlog modules under `stream/distributedlog` module. #1068 
- [x] update version of distributedlog modules #1069 
- [x] add the modules to root pom.xml and provide a profile to activate it. #1069
- [x] update CI to have a CI to build the whole suites
- [ ] integrate tutorials and documentation for that.
- [x] integrate into the dist package 
- [x] integrate into docker image
- [x] build the shading jar in bookkeeper way

sijie:
Closed this as all the tasks have been done.

PR #1025, opened 2018-01-19
BP-26: Move the development of distributedlog library to bookkeeper

sijie:

This is a BP proposing moving the development of distributedlog library to bookkeeper. See details in the BP file.

Master Issue: #1024


hsaputra:
+1

site/bps/BP-26-move-distributedlog-core-library.md
```diff
@@ -0,0 +1,60 @@
...
+This is a new library moved in bookkeeper. It will *NOT* touch any existing bookkeeper modules and ledger api.
+
+### Proposed Changes
+
+This proposal will *ONLY* move following library-only modules from distributedlog repo:
+
+- distributedlog-core: the log stream library that build over bookkeeper ledgr api. It doesn't introduce any service components. Library only.
+- distributedlog-io/dlfs: A hdfs filesystem api wrapper over the log stream api, to provide filesystem-like usage over bookkeeper.
+
+This proposal will *NOT* move other service components like "distributedlog-proxy".
```
ivankelly: I'm unclear on what you mean when you say you'll *ONLY* move the above modules, and *NOT* move the following modules. I assume the whole code base is coming in, no?

sijie: @ivankelly no this proposal is only proposing moving the library not the proxy service for a few reasons. 1) what bookkeeper needs is a *distributedlog*-ish library to simplify the usage of bookkeeper. *ledger* api is not usable for most of the users. 2) moving the library is to mimic the impact of other bookkeeper users who don't need dlog. 3) the dlog proxy was written in twitter-finagle, which depends on a specialized libthrift which is only available in twitter maven. so whether to move dlog proxy or improve it with grpc is a broader discussion than just moving the dlog library. I would defer that in a later discussion.

Hope this makes things clear.

ivankelly: So there will still be a distributedlog repo, which will continue to have the proxy service?

AFAIK, we do have other things which are depending on finagle (stats provider i think), and thereby depend on the custom libthrift (which is in public repos now). The finagle team were talking about moving to latest apache libthrift just before christmas.

site/bps/BP-26-move-distributedlog-core-library.md
```diff
@@ -0,0 +1,60 @@
...
+
+- those modules will be moved under `stream/distributedlog` directory at apache bookkeeper repo.
```
ivankelly: "those modules" in this sentence is therefore referring to dlog-core and dlfs? I would spell this out explicitly here, because I took it it mean all modules except those two, given the previous sentence. Hence the confusion.

site/bps/BP-26-move-distributedlog-core-library.md
sijie: > there will still be a distributedlog repo, which will continue to have the proxy service?

yes. 

> we do have other things which are depending on finagle (stats provider i think), 

stats provider doesn't depend on finagle-thrift.

site/bps/BP-26-move-distributedlog-core-library.md
sijie: will update.

site/bps/BP-26-move-distributedlog-core-library.md
ivankelly: It's twitter-http server, via zipkin-core

site/bps/BP-26-move-distributedlog-core-library.md
sijie: @ivankelly updated. please take a look.

site/bps/BP-26-move-distributedlog-core-library.md
sijie: @ivankelly 

1) twitter server wraps all the utilities such as logging, metrics, tracing. they are dependencies but not necessarily enabled at runtime. and twitter-http-server is one implementation of http-server, it is only used when you load twitter-http server in the classpath.

2) it is a bit different story in distributedlog-proxy uses finagle-thrift as the wire protocol. Unfortunately finagle-thrift depends on a specialized protocol called thrift-mux for multiplexing requests, which only available at twitter's folk, not at the official thrift library. so it is a bit hard to extend to other language bindings. there were multiple issues reported in DL when people tried to create different language bindings, and it is hard to address at current finagle-thrift protocol. there was also discussion on supporting a gRPC proxy, which make multiple language bindings much easier. so it would be good to think about proxy part for different factors. that's why I didn't include proxy in this BP to keep things easier.

Hope this make you clear.

sijie: fyi - finagle thrift mux - https://twitter.github.io/finagle/guide/Protocols.html

ivankelly: Ah, I thought the problem was pulling the custom libthrift in just as a dep.

site/bps/BP-26-move-distributedlog-core-library.md
```diff
@@ -0,0 +1,62 @@
...
+
+- distributedlog-core: the log stream library that build over bookkeeper ledgr api. It doesn't introduce any service components. Library only.
+- distributedlog-io/dlfs: A hdfs filesystem api wrapper over the log stream api, to provide filesystem-like usage over bookkeeper.
+
+This proposal will *NOT* move other service components like "distributedlog-proxy".
+
+The steps to make this change are described as following:
+
+- the proposed modules (`distributedlog-core` and `distributedlog-io/dlfs`) will be moved under `stream/distributedlog` directory at apache bookkeeper repo.
+- a new "stream" profile will be added to the root `pom.xml` file. The distributedlog module will only be build when "-Pstream" is specified
```
ivankelly: rather than skip the build, why not enable skipTests by default, and set skipTests=false with -Pstream. This way breaking changes won't creep in, but they won't have to pay the penalty of running full tests.

sijie: either works for me. we can have ci jobs running with "-Pstream",  which breaking changes still won't creep in.

sijie:
Closed this BP as `Accepted`.

- +1: 7 (3 bindings)
- -1: 0

will create sub-tasks in master ticket #1024

PR #1028, opened 2018-01-22
Test upgrade from old server version to new version

ivankelly:
This patch contains two tests.
- Upgrade between all released version, up to latest version
- Upgrade direct from 4.1.0 to latest version

In all cases the old client should continue to work with the new version.
In one case(4.2->4.3), we broken client compatibility so the the new client
stops working with the old server.

These tests replace and expand on the testing done by
TestBackwardCompat#testCompat410 with the old BC tests.

Master Issue: #903


tests/backward-compat-upgrade-direct/src/test/groovy/org/apache/bookkeeper/tests/TestCompatUpgradeDirect.groovy
```diff
@@ -0,0 +1,123 @@
...
+@FixMethodOrder(MethodSorters.NAME_ASCENDING)
+class TestCompatUpgradeDirect {
+    private static final Logger LOG = LoggerFactory.getLogger(TestCompatUpgradeDirect.class)
+    private static byte[] PASSWD = "foobar".getBytes()
+
+    @ArquillianResource
+    DockerClient docker
+
+    @Test
+    public void test0_upgradeDirect410toCurrent() throws Exception {
```
sijie: any reason why you choose this naming pattern: 'test${number}_${test-name}'? we typically don't use "_" in java method, I would prefer removing it. the name can be just "testUpgradeDirect410toCurrent" or just "upgradeDirect410toCurrent".

tests/backward-compat-upgrade-direct/pom.xml
```diff
@@ -0,0 +1,36 @@
...
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="
+  http://maven.apache.org/POM/4.0.0
+  http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.apache.bookkeeper.tests</groupId>
+    <artifactId>integration-tests-base-groovy</artifactId>
```
sijie: any reason why this module's parent is "integration-tests-base-groovy"? can't we just include `integration-tests-base-groovy` as a dependency?

it is also an interesting layout to me. I think we typically organize the parent module in in hierarchical. it seems that you are using a sibling module as the parent. any reason? it might be good to follow how other modules are organized and make it consistent across the project.

it probably can be organized in this way:

- `tests/integration-tests`
- `tests/integration-tests/base`
- `tests/integration-tests/base-groovy`
- `tests/integration-tests/utils`
- `tests/integration-tests/topologies`
- `tests/integration-tests/backward-compat`

ivankelly: I used a parent instead of dependency, because i want plugin settings to be inherited too, such as the groovy stuff and surefire stuff. I can move them around to change the structure if you like. Could we do it after the whole series is in though?

tests/backward-compat-upgrade-direct/src/test/groovy/org/apache/bookkeeper/tests/TestCompatUpgradeDirect.groovy
ivankelly: Because the setup and teardown happen per suite, i need to order the tests, which is why the numbers are there. the ```_``` is just to break up the name.

tests/backward-compat-upgrade-direct/pom.xml
sijie: sure. it's better to organize them in a better hierarchy, so people is easy to understand and maintain. please create an issue for it.

ivankelly: #1030 created. We need to start triaging issues so this stuff gets picked up.

PR #1034, opened 2018-01-22
Servers formatted without instance id can't join a cluster with insta…

ivankelly:
Test that old client can continue to work with servers which are using their
hostname as their bookie id.

This replaces TestBackwardCompat#testCompatWrites and TestBackwardCompat#testCompatReads
from the old BC tests.

Master Issue: #903


ivankelly:
It's an arquillian limitation. Setup and teardown occur per suite, so if a test needs a fresh environment, it needs a new module.

sijie:
@ivankelly 

can you help understand this a bit more? if setup and teardown is per suite, can't you have multiple suites in one module? I don't think the docker environment is used cross suites, right?

for example, if I compare [backward-compat-hierarchical-ledger-manager](https://github.com/apache/bookkeeper/blob/master/tests/backward-compat-hierarchical-ledger-manager/src/test/resources/arquillian.xml) and [backward-compat-upgrade-direct](https://github.com/apache/bookkeeper/blob/master/tests/backward-compat-upgrade-direct/src/test/resources/arquillian.xml), they are all using same [3-node-all-version-unstarted.yaml](https://github.com/apache/bookkeeper/blob/master/tests/integration-tests-topologies/src/main/resources/cube-definitions/3-node-all-version-unstarted.yaml). 

Can you explain more why a new module is needed?

ivankelly:
With arquillian the docker containers are started before all tests in a module, and are torn down after all tests in a module. Looking at it again, I don't think it uses the concept of suites, but even if it did, and we made it work, it doesn't expose suites internally, so we wouldn't be able to dump container logs by suite, which would make debugging failures painful. Personally, I'd prefer it worked more like ```@BeforeClass``` and ```@AfterClass```. The fact that it doesn't is part of the reason I was lukewarm about using arquillian at the start. In the end though, I find the module easy enough to work with. If you want a new test, just copy a directory and fixup a few places.

Also, I think it's particularly bad for BC cases, because they involve upgrading bookies, so each test needs to start with a clean slate.

> for example, if I compare backward-compat-hierarchical-ledger-manager and backward-compat-upgrade-direct, they are all using same 3-node-all-version-unstarted.yaml.
> 
> Can you explain more why a new module is needed?

hierarchical-ledger-manager starts bookies on 4.2.0, then upgrades all to latest. If upgrade-direct were to run on the same set of containers, then the format in zookeeper, and the format in the bookie journal and ledger dirs would already be compatible with the latest, so the test would be pointless.

sijie:
@ivankelly okay.

sijie:
@ivankelly can you rebase this PR? It is ready to go.

ivankelly:
@sijie rebased

ivankelly:
j8 & j9 failures are flakes. Merging.

Issue #1030, opened 2018-01-22
BC test cleanup


ivankelly:
See https://github.com/apache/bookkeeper/pull/1028#discussion_r162941741

- [x] Move all backwards compat tests under a common directory
- [x] Also move tests into a backward-compat java package.
- [x] close all classloaders in groovy tests
- [x] Add pointers on running arquillian tests
- [x] Add integration tests to CI

ivankelly:
Also move tests into a backward-compat java package.

ivankelly:
Also, close all classloaders in groovy tests

eolivelli:
Currently we have a directory named 'tests', what about renaming it 'ittests' o 'itests' (for ' integration tests')

ivankelly:
It would be under a common directory under tests, likes tests/backwards-compat

sijie:
This issue has been addressed by all the subtasks linked above

PR #1035, opened 2018-01-22
Ledgers without masterkey in metadata can't be recovered

ivankelly:
Test that ledgers, created without a master key in the metadata, cannot be recovered until
the master key is known.

Replaces TestBookieRecovery from the old tests.

Master Issue: #903


tests/backward-compat-recovery-no-password/src/test/groovy/org/apache/bookkeeper/tests/TestCompatRecoveryNoPassword.groovy
```diff
@@ -0,0 +1,248 @@
...
+            bka.close();
+        }
+
+        adminConf.setBookieRecoveryDigestType(BookKeeper.DigestType.CRC32)
+        adminConf.setBookieRecoveryPasswd(passwdCorrect)
+
+        bka = new BookKeeperAdmin(adminConf)
+        try {
+            bka.recoverBookieData(failedBookieId)
+            Assert.fail("Shouldn't be able to recover with wrong digest")
```
eolivelli: I don't know if we can use 'import static' in Groovy.

I wonder if we can have some flavour of checkstyle for Groovy

/cc @sijie

tests/backward-compat-recovery-no-password/src/test/groovy/org/apache/bookkeeper/tests/TestCompatRecoveryNoPassword.groovy
```diff
@@ -0,0 +1,248 @@
...
+        byte[] passwdCorrect = "AAAAAA".getBytes()
+        byte[] passwdBad = "BBBBBB".getBytes()
+
+        String currentVersion = System.getProperty("currentVersion")
+        String zookeeper = BookKeeperClusterUtils.zookeeperConnectString(docker)
+
+        BookKeeperClusterUtils.legacyMetadataFormat(docker)
+
+        // Create a 4.1.0 client, will update /ledgers/LAYOUT
+        def v410CL = MavenClassLoader.forBookKeeperVersion("4.1.0")
```
eolivelli: I think we should try to close all classloaders

If I undersand correctly we are using forkCount = 0, so no new JVM will be spawned for test.
We should take care of not leaving memory (classloader) leaks

tests/backward-compat-recovery-no-password/src/test/groovy/org/apache/bookkeeper/tests/TestCompatRecoveryNoPassword.groovy
sijie: I don't think we enforce any checkstyle about import static. We only disallow "import *". We are okay to use "Assert.fail" or "import static ..fail;" and use "fail". 

Also I don't know how checkstyle is applied to groovy. I think we need an issue to enforce checkstyle on groovy as well.

tests/backward-compat-recovery-no-password/src/test/groovy/org/apache/bookkeeper/tests/TestCompatRecoveryNoPassword.groovy
ivankelly: Added a note to #1030

tests/backward-compat-recovery-no-password/src/test/groovy/org/apache/bookkeeper/tests/TestCompatRecoveryNoPassword.groovy
ivankelly: Create #1043

ivankelly:
@eolivelli do you have more comments, or can I merge?

PR #1037, opened 2018-01-22
BookKeeperClusterTestCase,UpdateLedgerOpTest: remove call to setAllow…

athanatos:
(@bug W-2962634@)
Signed-off-by: Andrey Yegorov <ayegorovsalesforce.com>
[Ported to master]

### DIFF
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/UpdateLedgerOpTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/UpdateLedgerOpTest.java
@@ -55,7 +55,6 @@ public class UpdateLedgerOpTest extends BookKeeperClusterTestCase {
 
     public UpdateLedgerOpTest() {
         super(3);
-        baseConf.setAllowLoopback(true);
         baseConf.setGcWaitTime(100000);
     }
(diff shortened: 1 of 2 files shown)


eolivelli:
Can you give a motivation for this change ?

athanatos:
@eolivelli  It seems superfluous and it confused me when I was debugging something else.  I can drop it if you like, but this seems simpler.

sijie:
@athanatos we are trying to use loopback as possible as we can, this would provide a more `stable` environment for running all the tests in CI environment. since all the tests are running in one machine, I am not sure how this would be different when using loopback vs not using loopback.

eolivelli:
I would prefer to keep the ability of using loopback as Sijie described there is more benefit than trouble.
If you do not have strong feeling about this I would like to close this patch as won't fix

athanatos:
@eolivelli @sijie Oh, sorry, I wasn't clear.  It's that TestBKConfiguration already sets it, so these calls  are superfluous.  It confused me briefly, so I figured I'd clean it up.  I can drop it if you think this is better.

sijie:
@athanatos oh, then that makes sense to me. +1

Issue #1039, opened 2018-01-22
BKClient tests with BookieErrors


reddycharan:
**FEATURE REQUEST**

1. Please describe the feature you are requesting.

This is testcode request. end-to-end tests for validating BKClient readEntry calls incase of combination of bookie timeouts and data corruption.

2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have). Are you currently using any workarounds to address this issue?

this is just new testcode. Must have for us, since these tests are missing in the current codebase.

PR #1040, opened 2018-01-22
ISSUE #1039: BKClient tests with BookieErrors

reddycharan:

- end-to-end tests for validating BKClient readEntry
calls incase of combination of bookie timeouts and
data corruption

Master Issue: #1039


sijie:
@reddycharan just one general comment (not related to this PR): it would be good if you could spend some time updating the description when you sent the PR (e.g. move the description down to below "Descriptions of the changes in this PR" and update the master issue). so the merge script can have a better commit message.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookKeeperClientTestsWithBookieErrors.java
```diff
@@ -0,0 +1,342 @@
...
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Test the bookkeeper client with errors from Bookies.
+ */
+public class BookKeeperClientTestsWithBookieErrors extends BaseTestCase {
```
sijie: I think it is good enough to test one time of digest type, because the test is focusing on testing bookie errors not different implementation of digest types. we unnecessarily increases the number of duplicated test cases to run, which increase the possibility of flakiness and also the testing time. 

so I would suggest just test with CRC32 (or even just CRC32C), which is the commonly used.

reddycharan: hmm..i thought about it. But in this testsuite, for few testcases I'm corrupting data. So I would like to make sure digest error is observed in both the digestmanagers. It is kind of end-to-end negative scenario for various digestmanagers.

reddycharan:
@sijie will be mindful about PR description from now on. The last couple of PRs are for just testcases, so I didn't have much to mention in all of the sections.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookKeeperClientTestsWithBookieErrors.java
sijie: well. I think the point if you are testing digest manager cases, you should just test the digest manager, you don't need to involve other components. if you are testing end-to-end, you should just use one digest manager. remember all the digest manager are implementing same interface, there is no need to test different digest manager end-to-end.

I would suggest splitting the test case into two.

1) you can test corrupting data on various digest managers, without involving other components.
2) you can test end-to-end using one digest type.

sijie:
@reddycharan left a comment, which I would prefer if you want parameterized tests, it should be done on digest manager only. for end-to-end testing, it would be great to just test one.

regarding the PR description, you can edit PR description any time after you sent the PR.

Issue #1041, opened 2018-01-23
New testcases for SortedLedgerStorage, EntryMemTable and LedgerHandleAdv


reddycharan:
**FEATURE REQUEST**

1. Please describe the feature you are requesting.

Adding new testcases for 

1) New Testclass for ZeroBuffer
2) New testcases for missing testcoverage in
SortedLedgerStorage, EntryMemTable and LedgerHandleAdv

2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have). Are you currently using any workarounds to address this issue?

this is testcode and it is nice to have

3. Provide any additional detail on your proposed use case for this feature.

PR #1042, opened 2018-01-23
ISSUE #1041: Adding new testcases

reddycharan:

- New Testclass for ZeroBuffer
- New testcases for missing testcoverage in
SortedLedgerStorage, EntryMemTable and LedgerHandleAdv

Master Issue: #1041

### DIFF
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieWriteLedgerTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieWriteLedgerTest.java
@@ -28,6 +28,7 @@ import io.netty.buffer.Unpooled;
 
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.List;
@@ -35,9 +36,11 @@ import java.util.Map;
 import java.util.Random;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutionException;
 
 import org.apache.bookkeeper.client.AsyncCallback.AddCallback;
+import org.apache.bookkeeper.client.BKException.BKLedgerClosedException;
 import org.apache.bookkeeper.client.BookKeeper.DigestType;
 import org.apache.bookkeeper.meta.LongHierarchicalLedgerManagerFactory;
 import org.apache.bookkeeper.net.BookieSocketAddress;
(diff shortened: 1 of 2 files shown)


bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieWriteLedgerTest.java
```diff
@@ -344,6 +347,88 @@ public void testLedgerCreateWithCustomMetadata() throws Exception {
...
     /*
+     * Verify the functionality of Advanced Ledger which accepts ledgerId as
+     * input and returns LedgerHandleAdv. LedgerHandleAdv takes entryId for
+     * addEntry, and let user manage entryId allocation.
+     * This testcase is mainly added for covering missing code coverage branches
+     * in LedgerHandleAdv
+     *
+     * @throws Exception
+     */
+    @Test(timeout = 60000)
```
sijie: nit: remove timeout

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieWriteLedgerTest.java
```diff
@@ -344,6 +347,88 @@ public void testLedgerCreateWithCustomMetadata() throws Exception {
...
+        lh.addEntry(0, entry.array());
+
+        // here asyncAddEntry(final long entryId, final byte[] data, final
+        // AddCallback cb, final Object ctx) method is
+        // called which is not covered in any other testcase
+        entry = ByteBuffer.allocate(4);
+        entry.putInt(rng.nextInt(maxInt));
+        entry.position(0);
+        entries1.add(entry.array());
+        CountDownLatch latch = new CountDownLatch(1);
```
sijie: line 378-389 can be simplified with the sync method, no?

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieWriteLedgerTest.java
```diff
@@ -344,6 +347,88 @@ public void testLedgerCreateWithCustomMetadata() throws Exception {
...
+        // here addEntry is called with incorrect offset and length
+        entry = ByteBuffer.allocate(4);
+        entry.putInt(rng.nextInt(maxInt));
+        entry.position(0);
+        try {
+            lh.addEntry(2, entry.array(), -3, 9);
+            fail("AddEntry is called with negative offset and incorrect length,"
+                    + "so it is expected to throw IndexOutOfBoundsException");
+        } catch (RuntimeException exception) {
+            // expected IndexOutOfBoundsException
```
sijie: if it is expected to "IndexOutOfBoundsException", can you just catch "IndexOutOfBoundsException"?

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieWriteLedgerTest.java
```diff
@@ -392,6 +477,79 @@ public void testLedgerCreateAdvWithLedgerIdInLoop() throws Exception {
...
+
+        assertTrue("Here we are expecting Bookies are configured to use SortedLedgerStorage",
+                baseConf.getSortedLedgerStorageEnabled());
+
+        long ledgerId;
+        int ledgerCount = 10;
+
+        List<List<byte[]>> entryList = new ArrayList<List<byte[]>>();
+        LedgerHandle[] lhArray = new LedgerHandle[ledgerCount];
+        long skipListSizeLimit = baseConf.getSkipListSizeLimit();
```
sijie: this is not a good unit test. can you set the skip list size limit to a smaller value? you shouldn't use the default value of testing purpose.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieWriteLedgerTest.java
```diff
@@ -801,6 +959,24 @@ private void readEntries(LedgerHandle lh, List<byte[]> entries) throws Interrupt
...
     }
 
+    private void readEntriesAndValidateDataArray(LedgerHandle lh, List<byte[]> entries)
+            throws InterruptedException, BKException {
+        ls = lh.readEntries(0, entries.size() - 1);
+        int index = 0;
+        while (ls.hasMoreElements()) {
+            byte[] originalData = entries.get(index++);
+            byte[] receivedData = ls.nextElement().getEntry();
+            LOG.debug("Length of originalData: " + originalData.length);
```
sijie: use '{}'

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieWriteLedgerTest.java
```diff
@@ -801,6 +959,24 @@ private void readEntries(LedgerHandle lh, List<byte[]> entries) throws Interrupt
...
+            LOG.debug("Length of receivedData: " + receivedData.length);
+            assertTrue(
+                    String.format("LedgerID: %d EntryID: %d OriginalDataLength: %d ReceivedDataLength: %d", lh.getId(),
+                            (index - 1), originalData.length, receivedData.length),
+                    originalData.length == receivedData.length);
+            assertTrue(String.format("Checking LedgerID: %d EntryID: %d  for equality", lh.getId(), (index - 1)),
```
sijie: user assertArrayEquals. it is better than `assertTrue`, because it would show the diff when the arrays don't equal.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieWriteLedgerTest.java
```diff
@@ -801,6 +959,24 @@ private void readEntries(LedgerHandle lh, List<byte[]> entries) throws Interrupt
...
+    private void readEntriesAndValidateDataArray(LedgerHandle lh, List<byte[]> entries)
+            throws InterruptedException, BKException {
+        ls = lh.readEntries(0, entries.size() - 1);
+        int index = 0;
+        while (ls.hasMoreElements()) {
+            byte[] originalData = entries.get(index++);
+            byte[] receivedData = ls.nextElement().getEntry();
+            LOG.debug("Length of originalData: " + originalData.length);
+            LOG.debug("Length of receivedData: " + receivedData.length);
+            assertTrue(
```
sijie: use assertEquals, it is better than "assertTrue"

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieWriteLedgerTest.java
reddycharan: actually with the old code it was throwing ArrayIndexOutOfBoundsException, but with the recent changes (ByteBuffer getting replaced with ByteBuf in the internal class implementations) it is throwing IndexOutOfBoundsException. So to keep this oblivious to the internal implementation, I'm catching RuntimeException.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieWriteLedgerTest.java
reddycharan: changed

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieWriteLedgerTest.java
reddycharan: done

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieWriteLedgerTest.java
reddycharan: done

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieWriteLedgerTest.java
reddycharan: done

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieWriteLedgerTest.java
reddycharan: hmm..I wrote this testsuite about an year ago. Don't quite remember why I chose async (probably for code coverage issue). Anyhow it should be ok to use async.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieWriteLedgerTest.java
reddycharan: im not quite sure if it is easy/worthy enough to change the conf value just for this testcase. The issue I see here is BookKeeper cluster is started as part of @Before method and the conf values should be set in this method. I dont see a way to have customized conf value just for this testcase.

sijie: you have a few options here:

1) what is the side effects of setting smaller skip list size limit? if there is not side effects, you can set the setting in the constructor, and the same setting will be applied to all the test cases. Ideally all the other tests should run okay with any skip list size limit.

2) you can stop the cluster, update the configuration and start the cluster in this test case.

3) you can move this test case to a separate test suite, which is configured with smaller skip list limit size.

thoughts?

reddycharan: done, set it in the constructor.

PR #1044, opened 2018-01-23
AuditorPeriodicCheckTest: add test for entries with failed bookie writes

athanatos:
Validate that the auditor's periodic check will detect entries
with failed bookie writes and repair them.

Signed-off-by: Rithin <rithin.shetty@salesforce.com>
(@bug W-3311952@)(@bug W-3311965@)

### DIFF
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/replication/AuditorPeriodicCheckTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/replication/AuditorPeriodicCheckTest.java
@@ -22,6 +22,7 @@ package org.apache.bookkeeper.replication;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertTrue;
 
 import io.netty.buffer.ByteBuf;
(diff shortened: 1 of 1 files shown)


bookkeeper-server/src/test/java/org/apache/bookkeeper/replication/AuditorPeriodicCheckTest.java
```diff
@@ -348,4 +354,114 @@ public void run() {
...
+        for (int i = 0; i < numBookies; i++) {
+            ReplicationWorker rw = new ReplicationWorker(
+                    zkc, bsConfs.get(i), NullStatsLogger.INSTANCE);
+            rw.start();
+            l.add(rw);
+        }
+        underReplicationManager.close();
+
+        // Wait for ensemble to change after replication
+        Thread.sleep(3000);
```
eolivelli: We should have some more deterministic way of waiting.
But I think thus change can go as it is now, then we will replace all Thread.sleep in the (hopefully near) future

athanatos: Yeah, I haven't had the time to improve this one.  If it turns out to be a problem I'll take another look.

PR #1045, opened 2018-01-24
More testcases for LedgerHandle

reddycharan:

- New testcases for missing testcoverage in
LedgerHandle and ReadOnlyLedgerHandle

### DIFF
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/test/BookieReadWriteTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/test/BookieReadWriteTest.java
@@ -38,16 +38,20 @@ import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import org.apache.bookkeeper.client.AsyncCallback.AddCallback;
+import org.apache.bookkeeper.client.AsyncCallback.CloseCallback;
 import org.apache.bookkeeper.client.AsyncCallback.ReadCallback;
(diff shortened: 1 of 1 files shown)


bookkeeper-server/src/test/java/org/apache/bookkeeper/test/BookieReadWriteTest.java
```diff
@@ -953,13 +957,139 @@ public void testReadFromOpenLedgerZeroAndOne() throws Exception {
...
+            Assert.fail("Test3 - asyncClose failed because of exception - " + KeeperException.Code.get(rcArray[0]));
+        }
+
+        // close of readonlyhandle should not affect the writehandle
+        writeNEntriesLastWriteSync(lh, 5);
+        lh.close();
+    }
+
+    @Test
+    public void testLedgerHandle() throws Exception {
```
eolivelli: This name is really generic.

another, more important piint, I think it would be better to use mockito based suite, you so not need a real cluster to test this methods

sijie: @eolivelli I believe these are ported back from salesforce. I would suggest not bringing mockito in the context, because mock based suite is a totally different direction than this PR. We can have follow up tasks to improve them using mocks if you like.

eolivelli: Ok let's complete the port and move forward

PR #1052, opened 2018-01-25
Remove old compat stuff

ivankelly:
As compat tests are now docker based, the shading stuff is no longer
needed.

Master Issue: #903


eolivelli:
I like this very much !

I am +1 but I did not check that all tests have been migrated to the new system

This will also reduce the amount of stuff we are deploying nightly (SNAPSHOTS)

great work @ivankelly

sijie:
@eolivelli 

> This will also reduce the amount of stuff we are deploying nightly (SNAPSHOTS)

just fyi - starting from 4.6.0, we already don't deploy those bc artifacts: https://github.com/apache/bookkeeper/blob/master/tests/pom.xml#L57

eolivelli:
@sijie I forgot about it. Thank you for explanation

ivankelly:
Merging this

Issue #1049, opened 2018-01-24
Add config to limit users allowed to start a bookie process


athanatos:
If the bookie is started by the wrong user (like root), the on disk state can wind up with the wrong permissions.  Add a config to prevent such an operational error.

PR #1065, opened 2018-01-26
ISSUE #1049: Add PERMITTED_STARTUP_USERS to limit bookie startup users

athanatos:
Starting the bookie as the wrong user (such as root) can cause problems
as the on disk files may end up with the wrong permissions.  Add
PERMITTED_STARTUP_USERS to protect against this kind of operational
error.

(@bug W-3599751@)
Signed-off-by: Dustin Castor <dcastorsalesforce.com>
[Adapted to current code]

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
@@ -22,7 +22,6 @@ import static org.apache.bookkeeper.conf.ClientConfiguration.CLIENT_AUTH_PROVIDE
 import java.net.URL;
 import java.util.Iterator;
 import java.util.List;
-
 import javax.net.ssl.SSLEngine;
 
 import org.apache.bookkeeper.feature.Feature;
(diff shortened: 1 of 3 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieServer.java
```diff
@@ -159,6 +162,28 @@ public synchronized void shutdown() {
         running = false;
     }
 
+    /**
+     * Ensure the current user can start-up the process if it's restricted.
+     */
+    private void validateUser(ServerConfiguration conf) throws AccessControlException {
+        if (conf.containsKey("permittedStartupUsers")) {
```
sijie: use the constant defined in AbstractConfiguration.

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
```diff
@@ -118,6 +120,20 @@ protected AbstractConfiguration() {
...
+     * Limit who can start the application to prevent future permission errors.
+     */
+    public void setPermittedStartupUsers(String s) {
+        setProperty(PERMITTED_STARTUP_USERS, s);
+    }
+
+    /**
+     * Get array of users specified in this property.
+     */
+    public String[] getPermittedStartupUsers() {
```
eolivelli: Can't we have consistent return type, in regards to the setter?

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieServer.java
athanatos: Will do, thanks.

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
athanatos: @eolivelli Well, the reason for the asymmetry is that org.apache.commons.configuration.AbstractConfiguration.setProperty has magic to split strings containing the delimiter (',' by default) into a list prior to storing internally.  getStringArray there simply either returns the stored array or, if it's a string, returns it as a single element array.  I could instead choose to make setPermittedStartupUsers take an array, but it would make the unit tests (which are the only user) less useful since it would bypass the string splitting behavior.  If there's ever another user, it would be perfectly ok to add a variant that takes a String[] as well, but I don't see a reason to do it now.

PR #1070, opened 2018-01-29
Cleanup classloaders in backward-compat tests

ivankelly:
The BC tests run in the maven process (for classloader reasons), so we
need to make sure we clean stuff up to avoid running out of memory, as
the classloaders will load a bunch of jars into memory.

Master Issue: #1030


ivankelly:
they're autoclosable. I could wrap everything in try () {} but then each method gets ugly as hell with nesting

eolivelli:
@ivankelly yes that's was my same thought.
I think we can take current approach, we won't close them only upon failure, hopefully never
#shipit

sijie:
If you don’t like the nesting of try-block, you can try the lombok cleanup annotation: https://projectlombok.org/features/Cleanup

ivankelly:
Changed it to use lombok

PR #1074, opened 2018-01-29
URLDecode key/cert paths in TestTLS

ivankelly:
In TestTLS we load keys/certs from the classpath, which means we get
the filename as a URL. Since it is a URL it is URL encoded for special
characters. Usually this is not an issue, but in the case that there
are special characters in the path, these will be encoded, and then
when we try to open the file we get a FileNotFoundException. This
occurs on jenkins, when there are two parallel builds on the same
job (it adds a @N to the workspace name), and was causing all TLS
tests to fail in that case.

### DIFF
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/tls/TestTLS.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/tls/TestTLS.java
@@ -17,6 +17,7 @@
  */
 package org.apache.bookkeeper.tls;
 
+import static java.nio.charset.StandardCharsets.UTF_8;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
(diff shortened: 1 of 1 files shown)


sijie:
hmm I thought I fixed this at #880. it seems that it was overwritten by other changes?

ivankelly:
Looks like https://github.com/apache/bookkeeper/issues/933 overwrote it.

bookkeeper-server/src/test/java/org/apache/bookkeeper/tls/TestTLS.java
```diff
@@ -99,6 +101,10 @@ public TestTLS(String keyStoreFormat, String trustStoreFormat) {
         this.serverTrustStoreFormat = TLSContextFactory.KeyStoreType.valueOf(trustStoreFormat);
     }
 
+    private String getResourcePath(String resource) throws Exception {
+        return URLDecoder.decode(this.getClass().getClassLoader().getResource(resource).getPath(), UTF_8.name());
```
sijie: I think it is probably better to use `toURI().getPath()`, rather than us doing the magic decoding. 

[URI.getPath](https://docs.oracle.com/javase/7/docs/api/java/net/URI.html#getPath())

Issue #1067, opened 2018-01-27
RecoveryReadOp: a single hanging bookie can prevent a ledger from fencing


athanatos:
PendingReadOp as written won't fail a read with NoSuchEntry until all reads sent have responded.  This is a problem for RecoveryReadOp because we need to see NoSuchEntry on an entry in order to fix LAC.  For recovery reads, as soon as wQ-aQ+1 bookies have responded that they do not have the entry, we should return NoSuchEntry.

athanatos:
Working on it, should have patch on Monday.

PR #1077, opened 2018-01-30
ISSUE #1067: PendingReadOp: recovery, return NoSuchEntry on wQ-aQ+1 errors

athanatos:
In the case of a recovery read, we rely on a NoSuchEntry response on the
first missing entry to determine the final LAC.  As such, we need to
consider an entry to be gone once we see wQ-aQ+1 NoSuchEntry/Ledger
responses from bookies.  Otherwise, a zombie bookie can prevent ledger
recovery from suceeding indefinitely (it'll hit the timeout each time).

There was some preexisting logic from
https://issues.apache.org/jira/browse/BOOKKEEPER-365
b6c1a8bbd7c2d44c2edb59d9938fa073f6f478de,
but it seems to have been too conservative in that it waited for all
responses and required that there were no other errors present.  It
seems now to be unnecessary, and so has been removed.

TestParallelRead.testFailParallelReadMissingEntryImmediately seemed to
rely on the previous logic working outside of recovery, but I believe it
was really meant to test the recovery case, so it has been adjusted.

(@bug W-4651456@)

### DIFF
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/TestFencing.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/TestFencing.java
@@ -335,6 +335,41 @@ public class TestFencing extends BookKeeperClusterTestCase {
     }
 
     /**
+     * create a ledger and write entries.
+     * sleep a bookie
+     * Ensure that fencing proceeds even with the bookie sleeping
+     */
+    @Test
+    public void testFencingWithHungBookie() throws Exception {
+        LedgerHandle writelh = bkc.createLedger(digestType, "testPasswd".getBytes());
+
+        String tmp = "Foobar";
+
+        final int numEntries = 10;
+        for (int i = 0; i < numEntries; i++) {
+            writelh.addEntry(tmp.getBytes());
+        }
+
+        CountDownLatch sleepLatch = new CountDownLatch(1);
+        sleepBookie(writelh.getLedgerMetadata().getEnsembles().get(0L).get(1), sleepLatch);
+
+        LedgerHandle readlh = bkc.openLedger(writelh.getId(),
+                                             digestType, "testPasswd".getBytes());
+
+        try {
+            writelh.addEntry(tmp.getBytes());
+            LOG.error("Should have thrown an exception");
+            fail("Should have thrown an exception when trying to write");
+        } catch (BKException.BKLedgerFencedException e) {
+            // correct behaviour
+        }
+
+        sleepLatch.countDown();
+        readlh.close();
+        writelh.close();
+    }
+
+    /**
      * Test that fencing doesn't work with a bad password.
      */
     @Test
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
@@ -74,7 +74,7 @@ class PendingReadOp implements ReadEntryCallback, SafeRunnable {
     OpStatsLogger readOpLogger;
     private final Counter speculativeReadCounter;
 
-    final int maxMissedReadsAllowed;
+    final int requiredBookiesMissingEntryForRecovery;
     final boolean isRecoveryRead;
     boolean parallelRead = false;
     final AtomicBoolean complete = new AtomicBoolean(false);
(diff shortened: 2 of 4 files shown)


athanatos:
The main issue here is that we'd like to be able to return NoSuchEntry on entry e for a recovery read as soon as we can rule out a successful write for entry e.  To that end, it seems sufficient to wait until we have more than wQ-aQ NoSuchEntry/Ledger responses.

There is existing code (for both seq and parallel) that returns NoSuchEntry if
1) we've tried to send to all bookies
2) all bookies have either responded or netty has reported that it cannot send the read (BKException.Code.BookieHandleNotAvailableException)
3) at least wQ-aQ+1 bookies have responded with NoSuchEntry/Ledger
4) no errors other than NoSuchEntry/Ledger or BookieHandleNotAvailableException have been seen

@ivankelly That logic seems to have been introduced in b6c1a8bbd7c2d44c2edb59d9938fa073f6f478de .  Is that logic only intended to apply to the recovery read case, or is it important more generally?  If the former, I think I should be able to simply replace it, right?

sijie:
@athanatos I added this logic for [BOOKKEEPER-365](https://issues.apache.org/jira/browse/BOOKKEEPER-365). The logic is mainly for recovery. so I think your change is reasonable. I will go through the context and refresh my mind and comment at this pr.

ivankelly:
@athanatos the requirement for (Qw - Qa) + 1, is only important for recovery. This whole class could do with a refactor, but what you've added looks fine. RecoveryRead should be it's own Op, and should always be parallel.

athanatos:
@sijie If that code was meant to apply to recovery, than I think this patch obsoletes it.  I've pushed another commit which removes that logic and clarifies the conditions a bit.  There was one test
(TestParallelRead.testFailParallelReadMissingEntryImmediately) which relied on this behavior, but it seemed to be intended for the recovery case, so I adjusted the test case.

If this is ok, I'll squash the commits down and clean up the commit message.

sijie:
@athanatos the changes looks good to me. 

(I don't think you need to squash the commits. the merge script will anyway do that job)

sijie:
@ivankelly can you review it since you were involved in the discussion?

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/TestFencing.java
```diff
@@ -334,6 +334,46 @@ public void testFencingInteractionWithBookieRecovery2() throws Exception {
...
     }
 
+    /**
+     * create a ledger and write entries.
+     * sleep a bookie
+     * Ensure that fencing proceeds even with the bookie sleeping
+     */
+    @Test
+    public void testFencingWithHungBookie() throws Exception {
+        System.setProperty("digestType", digestType.toString());
```
ivankelly: What is this for?

jvrao:
+1

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/TestFencing.java
sijie: @athanatos do you mind answering this comment? otherwise this change is ready to go.

athanatos: @ivankelly Apologies, it's left over from the test I adapted.  I'll clean it up.

athanatos:
@ivankelly Repushed.

horizonzy:
@athanatos Hi, I have a question about this change.
E:3 WQ:3 AQ:3.

`Ledger 1` exist at three bookie.
Bookie1: lac-12
Bookie2: lac-11
Bookie3: lac-11

When client try to recover ledger 1. It will read entry from three bookies. There are two situation which will cause different result.

1. Client read bookie2 `entry 12`, got NoSuchEntry failed. Then read bookie3 `entry 12`, got NoSuchEntry failed again. It reach `requiredBookiesMissingEntryForRecovery`, end recovery. `Ledger 1` lac is 11 finally.
2. Client read bookie1 `entry12` succeed.  RecoverAddEntry to bookie2 and bookie3. Then read `entry 13`, all bookie NoSuchEntry failed, end recovery. `Ledger 1` lac is 12 finally.


The order of client `RecoveryReadOp` send to bookie will cause different result.

athanatos:
I haven't really looked at this code for about 4 years, I'm afraid I can't really be of much use.  Sorry!

horizonzy:
> I haven't really looked at this code for about 4 years, I'm afraid I can't really be of much use. Sorry!

Fine, thanks reply.

PR #1084, opened 2018-01-30
Lombok doesn't work with groovy

ivankelly:
So the @Cleanup annotation does nothing. I've changed it in the BC
test to a traditional try {} finally {} since groovy doesn't do
try-with-resources either, and its equivalent doesn't do multiple
objects.

Master Issue: #903


ivankelly:
yup, answer is nope. fails silently too, which is annoying. I only it it because a test was exhausting zk connections, as it wasn't closing the BK client.

sijie:
the change looks good to me. but just out of curiosity, have you tried - https://github.com/groovy/groovy-eclipse/wiki/Groovy-Eclipse-Maven-plugin#project-lombok ?

ivankelly:
Ya, I tried that agent, didn't seem to do anything.

Issue #1075, opened 2018-01-29
Add a noop digest implementation


athanatos:
In situations where digest overhead is undesirable or where it's useful to test without digest overhead, it would be good to have a noop digest option.

athanatos:
Will have patch up shortly.

PR #1087, opened 2018-01-30
ISSUE #1075: Add a noop digest implentation

athanatos:
This digest will add a predefined digest key without actually computing
it. This can be used for testing or in situations/use-cases where digest
is considered overhead.

(@bug W-3245776@)
[Adapted to current patch]

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/DummyDigestManager.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/DummyDigestManager.java
@@ -18,34 +18,27 @@
  * under the License.
  *
  */
-package org.apache.bookkeeper.client.api;
+package org.apache.bookkeeper.proto.checksum;
 
-import org.apache.bookkeeper.common.annotation.InterfaceAudience.Public;
-import org.apache.bookkeeper.common.annotation.InterfaceStability.Unstable;
+import io.netty.buffer.ByteBuf;
+
 
 /**
- * Digest type.
- *
- * @since 4.6
+ * This class provides a noop digest implementation.
  */
-@Public
-@Unstable
-public enum DigestType {
+public class DummyDigestManager extends DigestManager {
+    public DummyDigestManager(long ledgerId) {
+        super(ledgerId);
+    }
 
-    /**
-     * Entries are verified by applied CRC32 algorithm.
-     */
-    CRC32,
-    /**
-     * Entries are verified by applied MAC algorithm.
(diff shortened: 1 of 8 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/DummyDigestManager.java
```diff
@@ -0,0 +1,50 @@
...
+public class DummyDigestManager extends DigestManager {
+    private static final byte[] dummyDigest = new byte[] { 0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x6, 0x07 };
+
+    public DummyDigestManager(long ledgerId) {
+        super(ledgerId);
+    }
+
+    @Override
+    int getMacCodeLength() {
+        return 8;
```
sijie: any reason why a dummy digest manager needs a 8 bytes here? can't we just make it 0? the digest here is anyway useless, no?

athanatos: Not sure.  Perhaps there is value in emulating the size overhead?  @jvrao Thoughts?

jvrao: I really don't really recall why we made up the digest and kept it at 8. Maybe tried to keep the payload size in sync with CRC32.

sijie: @jvrao any idea to waste extra 4 bytes, if it is not even used.

athanatos: @jvrao I'll switch it to an empty payload.

jvrao: I am fine with it.

athanatos:
@eolivelli Repushed with new api test.

sijie:
@eolivelli please review this since you were involved in this PR

athanatos:
@eolivelli Done.

sijie:
ping @eolivelli

eolivelli:
retest this please

sijie:
retest this please

asfgit:

sijie:
retrigger travis ci build

sijie:
retest this please

Issue #1086, opened 2018-01-30
 io.netty.util.internal.OutOfDirectMemoryError under continuous heavy load. 


dlg99:
Bug report

Ran long stress test (~1hr) of the write load that was bottlenecked by NIC bandwidth (10G).
Some bookies in the cluster could not handle this load (due to compaction, disk configuration that we tested etc.)
Got a write timeout, client's logs had "io.netty.util.internal.OutOfDirectMemoryError".
OODME appeared even in shorter runs that succeeded (ensemble changes/bookie reconnects succeeded before request timeout).
Client had 10G heap and 10G of direct memory allocated.

Expected the load to complete without write errors, possibly with temporary throughput decrease.

After the investigation it appears that PCBC does not respect netty's isWritable() state and keeps on writing to a channel that cannot write fast enough.
Specifically this is possible because we had Ack Quorum == 2 and Write Quorum == 3.
Write is considered successful when AQ replied, so 1 "slow channel" keeps on getting data written to it this can use up all available direct memory.

dlg99:
related: https://github.com/apache/bookkeeper/issues/628

sijie:
This is fixed by #1088

PR #1088, opened 2018-01-30
ISSUE #1086 (@bug W-4146427@) Client-side backpressure in netty (Fixes: io.netty.util.internal.OutOfDirectMemoryError under continuous heavy load)

dlg99:

Added client-side support for netty backpressure.
Now PCBC will actually respect netty's behavior configured with
clientWriteBufferLowWaterMark and clientWriteBufferHighWaterMark config parameters.
It is highly recommended setting them reasonably high for expected size of requests and load
Otherwise client may end up doing frequent waits / throughput may drop.
     
Master Issue: #1086


sijie:
@merlimat can you take a look at this change?

dlg99:
@sijie I start wit p.2:

backpressure is enabled all the time. I.e. CHANNEL_WAIT_TIMEOUT_ON_WRITE does not affect read request, LAC etc.
I.e. read has different ways to issue speculative retries to other bookie.

It works because write will get blocked in PendingAddOp when it is submitting requests
```java
                sendWriteRequest(writeSet.get(i));
            }
```

sendWriteRequest gets blocked if we block on netty.
In our case app limits number of requests in flight.
i.e. it can have 50 writes in flight and in the current ensemble 2 bookies able to handle this while the 3rd one is slow or goes through long GC.
Without this change we end up submitting data to netty to all 3 bookies and submit more as soon as two of them ack the write. netty in this case keeps on buffering data for the 3rd bookie and finally we were getting OODME.
With this change request ends up being blocked in sendWriteRequest to a slow bookie until it either succeeds or fails to submit (hence CHANNEL_WAIT_TIMEOUT_ON_WRITE to limit wait for writes specifically).

The change does not help if app can submit unlimited number fo requests, I totally agree. 
I think that should be addressed in a separate change building on top of this one. 

There is also server side of the backpressure story not addressed in this change, specifically:
- server has to stop accepting requests if it cannot process them fast enough
- server has to do something if it cannot send responses to client fast enough (slow client case) -> either stop accepting requests, or drop responses, or combo of two

p.1:  I have comparison of throughput with different sizes of HWM for netty (LWM = HWM-1M).
Without the change test failed with write error at anywhere between 35min and 58min.
In this case I managed to run load overnight, no OODME, no write errors.

hwatermark-tests

sijie:
@dlg99 great! I will take a closer look at your result. thank you for that.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
```diff
@@ -849,17 +857,84 @@ private ChannelFuture closeChannel(Channel c) {
...
+        if (channel == null || !channel.isActive() || channel != this.channel) {
             errorOut(key);
             return;
         }
 
+        // channel.isWritable() check will actually respect netty's behavior configured with
+        // clientWriteBufferLowWaterMark and clientWriteBufferHighWaterMark config parameters.
+        // It is highly recommended setting them reasonably high for expected size of requests and load
+        // Otherwise client may end up doing frequent waits in this loop.
+        if (!channel.isWritable()) {
```
eolivelli: what about skipping the whole block if the feature is disabled?

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
```diff
@@ -849,17 +857,84 @@ private ChannelFuture closeChannel(Channel c) {
...
+            final long startTime = MathUtils.nowInNano();
+            final long maxSleepUntil = startTime + (timeoutMillis > 0
+                    ? TimeUnit.MILLISECONDS.toNanos(timeoutMillis)
+                    : TimeUnit.DAYS.toNanos(1));
+            synchronized (channelMonitor) {
```
eolivelli: Isn't a day too much ad upper bound?

dlg99: I considered 1 day === infinity from request point of view. As in we are waiting until we get notified. just a way to skip "if" later

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
dlg99: channel.isWritable() is netty channel's attribute. It switches to false when clientWriteBufferHighWaterMark of unsent data is reached. it switches to true when some data is sent and buffer usage drops below clientWriteBufferLowWaterMark. 
This is netty's way to let us know that we are trying to push to much and cannot keep up.
the fact that we did not pay any attention to it was a problem addressed here.

It is of course possible to just fail request in this case but in case of busy system / maxed out NIC capacity it will result in too many failed requests.

sijie:
@dlg99 

I see the metrics do show great results. However I have one concern about this change, it is basically blocking the scheduler threads on waiting a slow netty channel, this can result in bad side-effects - e.g responses might not be able to be processed, other requests will also potentially be blocked as well.

so I guess the benchmark itself is only testing writes and most likely one one or a few ledgers. In this case, the benchmark can easily show improvements, because there is effectively only one client/ledger handle is writing. However a real-world use case can be different, this change here will cause a situation - a slow bookie can potentially impact the whole application that use the bookkeeper client.

Ideally I think the throttling/back-pressure should be pop back to the ledger handles who issue the write request, rather than asking PerChannelBookieClient to do the blocking-wait work. I am not sure what would be the best approach here. 

One thought I have here:

- when a ledger handle (effectively PendingAddOp) attempts to sendWriteRequest, PCBC checks if the channel is writable, if it is not writable, fail the write request with a special exception. PCBC itself doesn't block on waiting channel to become writable.

- when PendingAddOp receives this special exception, it knows the channel is not available for write. then you can apply the throttling logic here:

1) you can notify the ledger handle that a channel is not available, asking the ledger handle to slow down applications or block the application
2) at the same time, you can listen on the channel state change. once the channel is writable, issue the write to the channel, and unblock the application for submitting add entries.

In this way, you can block the application thread rather than the internal io scheduler which can be shared across multiple ledger handles.

sijie:
@dlg99 any ideas on my comment?

dlg99:
@sijie sorry for the delay, missed the notification. 

I re-implemented it with the following idea in mind:
Do not block worker pool. we can write to netty channel even if it is not writable (as it happens without this change). When channel is not writable still write it to netty and send similar flag on PCBC and so on, bubbling up this status.
LedgerHandle checks if it needs to slow down and waits until either time out expires or write allowed. 
If timeout expires, LedgerHandle sets flag "allowFastFail" and still sends op downstream. in case allowFastFail is set and channel is still not writable PCBC fails op so ensemble change etc can happen.

sijie:
@dlg99 sgtm

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieClient.java
```diff
@@ -150,6 +150,16 @@ private int getRc(int rc) {
...
     }
 
+    public boolean isWritable(BookieSocketAddress address) {
+        final PerChannelBookieClientPool pcbcPool = lookupClient(address);
+        if (pcbcPool == null) {
+            // let write fail with whatever error it produces
+            return true;
+        }
+
+        return pcbcPool.isWritable(0L /*not really used in obtain()*/);
```
jvrao: If we have multiple channels in the pool you are checking only one here isn't it?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
```diff
@@ -93,6 +95,7 @@ static PendingAddOp create(LedgerHandle lh, ByteBuf payload, AddCallback cb, Obj
         op.callbackTriggered = false;
         op.hasRun = false;
         op.requestTimeNanos = Long.MAX_VALUE;
+        op.allowFastFailOnBlockedWrite = false;
```
jvrao: Do we really need this mechanism to pass the fastFail? Why not make it default if throttle time is enabled? or come up with another configuration option? Here we are doing it by default in the read and write path, and for sure LaC we can do the same.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
```diff
@@ -863,17 +902,44 @@ private ChannelFuture closeChannel(Channel c) {
...
             return;
         }
 
+        if (isWritable != channel.isWritable()) {
+            // isWritable is volatile so simple "isWritable = channel.isWritable()" would be slower
+            isWritable = !isWritable;
+        }
+
+        if (allowFastFail && !isWritable) {
+            errorOut(key);
```
jvrao: This returns BookieHandleNotAvailableException. It is usually interpreted as not able to connect to bookie basically cannelPool returning NULL. I think we should return a different error here (busy??). Not sure if we can handle it
differently but we can at least bubble it back to the application where the application will know the reason.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
dlg99: This is unrelated to client configuration and enabled in runtime per op. 
This is a flag on the op which controls the behavior in case if netty channel is not writable:
- If set it is false the op is pushed into the netty's buffer and non-writable status is bubbled up eventually forcing LedgerHandle to block and wait.
- If set to true it means that LedgerHandle knew that underlying bookie/channel is not writable, waited, given up, and is ok with failing this quickly. I.e. write op in this case waits for writable bookies in WQ, times out, submits op with the flag. Requests to non-writable bookies fail immediately triggering ensemble change as usual.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
```diff
@@ -863,17 +902,44 @@ private ChannelFuture closeChannel(Channel c) {
...
+            //errorOut(key, BKException.Code.TooManyRequestsException);
```
jvrao: Ah looks like you have it planned. Good.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
jvrao: My request is to shift the focus to the channel. All we need to know how many requests/data-size channel can accept once it turned non-writable. (another tunable)? Reset this counter as soon as channel turns writable. If the caller waited or not should not influence the channel to submit the request of fail-fast. That avoids sending flag down.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -148,6 +157,7 @@
         this.enableParallelRecoveryRead = bk.getConf().getEnableParallelRecoveryRead();
         this.recoveryReadBatchSize = bk.getConf().getRecoveryReadBatchSize();
         this.writeFlags = writeFlags;
+        this.waitForWriteSetMs = bk.getConf().getWaitTimeoutOnBackpressureMillis();
```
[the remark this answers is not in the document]
dlg99: I don't think reads vs writes matter in this case.
Writes wait for full WQ (i.e. if when any 1 out of 3 not writable), reads wait one (i.e. when 3 out of 3 not writable) so they should not be affected by wait that much. Writes have a chance to recover with ensemble change but we don't want frequent ensemble changes; reads will simply fail. 
Combined with netty's L/H water mark configs this really should happen in extreme cases only.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -197,6 +207,7 @@ public long getBookiePendingRequests(BookieSocketAddress bookieSocketAddress) {
         ensembleChangeCounter = bk.getStatsLogger().getCounter(BookKeeperClientStats.ENSEMBLE_CHANGES);
         lacUpdateHitsCounter = bk.getStatsLogger().getCounter(BookKeeperClientStats.LAC_UPDATE_HITS);
         lacUpdateMissesCounter = bk.getStatsLogger().getCounter(BookKeeperClientStats.LAC_UPDATE_MISSES);
+        sendWaitTimer = bk.getStatsLogger().getOpStatsLogger(BookKeeperClientStats.CLIENT_SEND_WAIT_TIMER);
```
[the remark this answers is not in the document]
dlg99: same as BookieWatcher's newEnsembleTimer and replaceBookieTimer.
I'll see what is  the better name might be for it

asfgit:

jvrao:
@dlg99 looks like you overwrote your commit and force pushed. It is nice to make another commit to reflect review comments so the history will be available to look back. May be for future.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -990,6 +1011,78 @@ void asyncRecoveryAddEntry(final byte[] data, final int offset, final int length
...
     }
 
+    private boolean isWritesetWritable(DistributionSchedule.WriteSet writeSet,
+                                       long key, int allowedNonWritableCount) {
+        if (allowedNonWritableCount < 0) {
+            allowedNonWritableCount = 0;
+        }
+
+        final int sz = writeSet.size();
+        final int requiredWritable = sz - allowedNonWritableCount;
```
jvrao: Nice, you are flipping the logic here. Any reason why caller's themselves can't do it? A reader needs to send 1, and the writer needs to send AQ. isn't it? Maybe there is a good reason.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandleAdv.java
```diff
@@ -211,6 +211,11 @@ public String toString() {
             return;
         }
 
+        if (!waitForWritable(distributionSchedule.getWriteSet(op.getEntryId()),
+                op.getEntryId(), 0, waitForWriteSetMs)) {
```
jvrao: All we need is AQ channels, right?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -1034,6 +1127,11 @@ public String toString() {
             return;
         }
 
+        if (!waitForWritable(distributionSchedule.getWriteSet(op.getEntryId()), op.getEntryId(),
+                0, waitForWriteSetMs)) {
+            op.allowFastFailOnBlockedServer();
```
jvrao: I would have loved to contain this logic at the channel/PCBC level. But I can live with this. :)

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandleAdv.java
dlg99: WQ. 
if AQ < WQ, 1 channel is not writable and we decide to wait for AQ only we have the following options:
-- eventually get out of direct memory error.
-- fail all writes to the blocked channel ==> frequent ensemble changes caused by slightest fluke in data transfer
Sender is not slowed down in this case and keeps on sending more.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
dlg99: PCBC runs in context of different thread (worker pool)

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/DefaultPerChannelBookieClientPool.java
```diff
@@ -84,14 +84,22 @@ public void intialize() {
...
+            return clients[0];
         }
         int idx = MathUtils.signSafeMod(key, clients.length);
-        clients[idx].connectIfNeededAndDoOp(callback);
+        return clients[idx];
+    }
+
+    @Override
+    public void obtain(GenericCallback<PerChannelBookieClient> callback, long key) {
+        getClient(key).connectIfNeededAndDoOp(callback);
```
sijie: Can we do smarter here?

randomly pickup a channel, if the channel is not writable, do iteration start this randomly picked channel to pick next writable channel?

so the logic here would work for multiple connections case.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieClient.java
```diff
@@ -150,6 +150,12 @@ private int getRc(int rc) {
         return faultyBookies;
     }
 
+    public boolean isWritable(BookieSocketAddress address, long key) {
+        final PerChannelBookieClientPool pcbcPool = lookupClient(address);
+        // let write fail with whatever error it produces
```
sijie: is the comment aligned with the code here?

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ClientConfiguration.java
```diff
@@ -797,6 +798,33 @@ public ClientConfiguration setPCBCTimeoutTimerNumTicks(int numTicks) {
...
+    }
+
+    /**
+     * Timeout controlling wait on request send in case of unresponsive bookie(s)
+     * (i.e. bookie in long GC etc.)
+     *
+     * @param value
+     *        negative value disables the feature
+     *        0 to allow request to fail immediately
+     *        Default is 500
```
sijie: missed `@return` tag.

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ClientConfiguration.java
```diff
@@ -797,6 +798,33 @@ public ClientConfiguration setPCBCTimeoutTimerNumTicks(int numTicks) {
...
+     * Timeout controlling wait on request send in case of unresponsive bookie(s)
+     * (i.e. bookie in long GC etc.)
+     *
+     * @return timeout value
+     *        negative value disables the feature
+     *        0 to allow request to fail immediately
+     *        Default is 500
+     */
+    public long getWaitTimeoutOnBackpressureMillis() {
+        return getLong(WAIT_TIMEOUT_ON_BACKPRESSURE, 500);
```
sijie: I think we should make this fault value as -1, util we have benchmarks show 500 is in general good for most of applications.

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ClientConfiguration.java
```diff
@@ -416,7 +417,7 @@ public ClientConfiguration setClientReceiveBufferSize(int bufferSize) {
      * @return netty channel write buffer low water mark.
      */
     public int getClientWriteBufferLowWaterMark() {
-        return getInt(CLIENT_WRITEBUFFER_LOW_WATER_MARK, 32 * 1024);
+        return getInt(CLIENT_WRITEBUFFER_LOW_WATER_MARK, 384 * 1024);
```
sijie: why 384?

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ClientConfiguration.java
```diff
@@ -437,7 +438,7 @@ public ClientConfiguration setClientWriteBufferLowWaterMark(int waterMark) {
      * @return netty channel write buffer high water mark.
      */
     public int getClientWriteBufferHighWaterMark() {
-        return getInt(CLIENT_WRITEBUFFER_HIGH_WATER_MARK, 64 * 1024);
+        return getInt(CLIENT_WRITEBUFFER_HIGH_WATER_MARK, 512 * 1024);
```
sijie: why 512?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
```diff
@@ -501,6 +504,10 @@ PendingReadOp parallelRead(boolean enabled) {
         return this;
     }
 
+    void allowFastFailOnBlockedServer() {
```
sijie: "BlockedServer" is a confusing term here, since we don't really use "blocker" at any other places. I would suggest changing it to "allowFailFastOnUnwritableChannel".

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
```diff
@@ -78,6 +78,7 @@
     final boolean isRecoveryRead;
     boolean parallelRead = false;
     final AtomicBoolean complete = new AtomicBoolean(false);
+    boolean allowFastFail = false;
```
sijie: `fail-fast` is better than `fast-fail`, since `fail-fast` is a well known term : https://en.wikipedia.org/wiki/Fail-fast

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
```diff
@@ -389,6 +390,7 @@ synchronized BookieSocketAddress sendNextRead() {
                 return null;
             }
 
+            // ToDo: pick replica with writable PCBC
```
sijie: if the TODO is not implemented, please create an issue here and add the github issue at the TODO comment.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperClientStats.java
```diff
@@ -67,6 +67,7 @@
     String CHANNEL_START_TLS_OP = "START_TLS";
     String CHANNEL_TIMEOUT_START_TLS_OP = "TIMEOUT_START_TLS";
     String NETTY_EXCEPTION_CNT = "NETTY_EXCEPTION_CNT";
+    String CLIENT_SEND_WAIT_TIMER = "CLIENT_SEND_WAIT_TIMER";
```
sijie: we don't really use "SEND" as the term at the project. It has to be either add or read.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -791,6 +801,17 @@ public void onFailure(Throwable cause) {
                                                               long lastEntry) {
         PendingReadOp op = new PendingReadOp(this, bk.getScheduler(), firstEntry, lastEntry);
         if (!bk.isClosed()) {
+            // waiting on the first one.
+            // not very helpful if there are multiple ensembles or if bookie goes into unresponsive
+            // state later after N requests sent.
+            // unfortunately it seems that alternatives are:
+            // - send reads one-by-one (our case anyway)
+            // - block worker pool (not good)
+            DistributionSchedule.WriteSet ws = distributionSchedule.getWriteSet(firstEntry);
```
sijie: `WriteSet` is a recycle object, you need to recycle it after usage.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -990,6 +1011,78 @@ void asyncRecoveryAddEntry(final byte[] data, final int offset, final int length
...
+        }
+
+        final long startTime = MathUtils.nowInNano();
+        boolean success = true;
+
+        if (durationMs == 0) {
+            success = isWritesetWritable(writeSet, key, allowedNonWritableCount);
+        } else {
+            int backoff = 1;
+            final Instant deadline = Instant.now().plus(durationMs, ChronoUnit.MILLIS);
```
sijie: better to move the instant allocation to be after `isWritesetWritable` check, so we don't need do any object allocation on normal case, where channels are writable almost all the time. Otherwise line 1053 will become an problem when the traffic is high.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -990,6 +1011,78 @@ void asyncRecoveryAddEntry(final byte[] data, final int offset, final int length
...
+
+            while (!isWritesetWritable(writeSet, key, allowedNonWritableCount)) {
+                if (Instant.now().isBefore(deadline)) {
```
sijie: better to just use `Stopwatch`, which we have between used over all the places, rather than using `Instant` here.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -990,6 +1011,78 @@ void asyncRecoveryAddEntry(final byte[] data, final int offset, final int length
...
+                    long maxSleep = Duration.between(Instant.now(), deadline).toMillis();
+                    if (maxSleep < 0) {
+                        maxSleep = 1;
+                    }
+                    long sleepMs = Math.min(backoff, maxSleep);
+
+                    try {
+                        Thread.sleep(sleepMs);
```
sijie: for readability, better to use `TimeUnit.MILLISECONDS.sleep(sleepMs)`

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperClientStats.java
dlg99: @sijie this is not actually add or read. This is time for either add or read request when request send is blocked.
I am open for suggestions of better name, would CLIENT_NETTY_WAIT_TIMER or CLIENT_WRITE_WAIT_TIMER sound better? though the wait is in ledger handle. CLIENT_PCBC_WRITE_WAIT_TIMER might be better. though "write" can be confused with add op.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -791,6 +801,17 @@ public void onFailure(Throwable cause) {
                                                               long lastEntry) {
         PendingReadOp op = new PendingReadOp(this, bk.getScheduler(), firstEntry, lastEntry);
         if (!bk.isClosed()) {
+            // waiting on the first one.
+            // not very helpful if there are multiple ensembles or if bookie goes into unresponsive
+            // state later after N requests sent.
+            // unfortunately it seems that alternatives are:
+            // - send reads one-by-one (our case anyway)
```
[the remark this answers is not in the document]
dlg99: Will do

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
dlg99: good catch, thank you.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
dlg99: will do. https://github.com/apache/bookkeeper/issues/1239

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ClientConfiguration.java
dlg99: More interesting question is why 64K. 
Without backpressure that value is simply ignored. netty would switch isWritable on and off but won't prevent one from writing to the channel.
With backpressure I've tested effect of different values of HWM (chart attached to the CR) and throughput with backpressure and default HWM of 64K was way too low. Most of the entries in my test were 64KB.

Also from what I read, channel allocated whole chunk from PooledbyteBufAllocator to not deal with concurrent access.
"The PooledbyteBufAllocator allocates 16 MiB per chunk by default.  If all chunks are full or the pool was not able to find a room for the requested buffer, it will allocate another chunk.  If all buffers allocated from the chunk are released, the chunk will be deallocated automatically." https://groups.google.com/forum/#!searchin/netty/PooledByteBufAllocator$20chunk%7Csort:date/netty/3riW1A26QKc/5pjs-n-tuHYJ
As far as I can tell this still stands.

So HWM < chunk size makes sense to reduce number of submitted requests that netty won't be able to write out anyway if server is not reading. But larger HWM helps with throughput when bookies are healthy / accommodates spikes in load better.

So I decided to increase default to safe (less than chunk size) value that nevertheless improves throughput in case of backpressure.

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ClientConfiguration.java
dlg99: goes along with answer on why HWM changed.
Pretty much set to 75% of HWM after couple of quick experiments.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/DefaultPerChannelBookieClientPool.java
dlg99: I'd keep this out of the scope of this change. This will require separate set of tests to prove that it helps. 
I.e. this is only important for num of channels > 0.
All these channels are to the same bookie so even when there are more than one channel and one channel is already blocked most likely bookie is not reading fast enough (i.e. hit NIC limit on client or on bookie, or long GC on bookie etc).
All that switching to another channel achieves there is buffering request in the client just in another channel's buffer.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
```diff
@@ -450,6 +451,18 @@ void cleanDisconnectAndClose() {
         close();
     }
 
+    /**
+     *
+     * @return boolean, true is PCBC is writable
+     */
+    public boolean isWritable() {
```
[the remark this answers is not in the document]
dlg99: there is no isWritable method in ChannelInboundHandlerAdapter which PCBC extends.
Annotation is not needed.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperClientStats.java
jvrao: How about CHANNEL_WRITE?

dlg99:
More perf tests (off, on no wait, on wait 4ms):

backpressure-community

sijie:
@jvrao since you were reviewing this before, can you review it again?

@merlimat : can you review this when you have time? this would be related to the effort you were doing for performance.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
jvrao: I like to have 'requiredWritableCount' than a negative 'allowedNonWritableCount' as I mentioned in my about comment, which I think is easy to understand. @dlg99  any reason to pass the -ve (not writable) and flip to writable in this routine?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandleAdv.java
jvrao: The reason we choose AQ < WQ is to ride through some of these issues where WQ-AQ paths to the bookie are sluggish. I believe treatment must be same if a bookie is slow/sluggish or a channel to a bookie is slow.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
jvrao: @dlg99 I am not doing a -1, but I think caller itself specifying the requiredWriteCount will make the logic clean. But, that is up to you. I don't need to review this point again.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandleAdv.java
jvrao: This is limiting the ability to make progress and taking away the advantage of WQ<AQ for low latency tail. I think it is OK to take an ensemble change. Let me know what you think.

dlg99: @jvrao there is an option to set waitTimeoutOnBackpressureMs=0 and get this behavior (immediate fail/ensemble change). 
I'd argue that in case of spiky traffic it is ok for netty to switch channel in non-writable state and back so to avoid frequent ensemble changes I'd wait a little. Ensemble changes aren't exactly free (takes time + more data in zk)

dlg99:
@sijie I haven't heard anything from @merlimat; JV is ok with the change. 
I'll merge this Thursday afternoon unless I hear objections by then (community meeting + a little bit of time after)

sijie:
@dlg99 got for it

PR #1091, opened 2018-01-31
Issue 1063: Write keeps refCnt longer

jvrao:

Current code keeps the toSend buffers until client receives
resonses from all wq bookies. From the senders perspective,
It is not required to keep the refcount on it at the PendingAddOp
level, as the ref is taken at bookie client level.

Keeping these buffers longer will increase the memory pressure
on the client.

Master Issue: #1091 

> ---
> quickly and easily:
>
>
>
> Otherwise:
> 
> 
> ---

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
@@ -208,7 +208,6 @@ class PendingAddOp extends SafeRunnable implements WriteCallback {
      * Initiate the add operation.
      */
     public void safeRun() {
-        hasRun = true;
         if (callbackTriggered) {
             // this should only be true if the request was failed due
             // to another request ahead in the pending queue,
@@ -232,6 +231,7 @@ class PendingAddOp extends SafeRunnable implements WriteCallback {
                 sendWriteRequest(writeSet.get(i));
             }
         } finally {
+            hasRun = true;
             writeSet.recycle();
         }
     }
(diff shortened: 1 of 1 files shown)


jvrao:
@kishorekasi @dlg99

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
```diff
@@ -232,6 +231,7 @@ public void safeRun() {
                 sendWriteRequest(writeSet.get(i));
             }
         } finally {
+            hasRun = true;
```
sijie: any reason why this is moved down to here?

jvrao: I thought it is good to set it after it actually finished running. Do you see any issues with it?

sijie: there is a return at line 216? 

if the purpose is to setting this after it actually finished running, you should have pull all the logic in a try-finally block, no?

jvrao: yeah; I will move back the change.

jvrao:
> new unit test case to cover this changed branches. Maybe the fact that all tests are passing is a good signal of the goodness of the change

That's my take on this. Since this is not a new change in functionality, passing existing tests must be good enough. IMO..

eolivelli:
Okay. Looks good to me let's ship this change

sijie:
thanks @jvrao . merge it and also cherry-picked to branch-4.6, since it is a good fix to be picked at 4.6.

PR #1092, opened 2018-01-31
Dont log ledgermetadata which contains password.

reddycharan:

Dont log ledgermetadata contains password used to create
the ledger. Logs are not supposed to reveal such sensitive info.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
@@ -327,6 +327,10 @@ public class LedgerMetadata implements org.apache.bookkeeper.client.api.LedgerMe
     }
 
     LedgerMetadataFormat buildProtoFormat() {
+        return buildProtoFormat(true);
+    }
+
+    LedgerMetadataFormat buildProtoFormat(boolean withPassword) {
         LedgerMetadataFormat.Builder builder = LedgerMetadataFormat.newBuilder();
         builder.setQuorumSize(writeQuorumSize).setAckQuorumSize(ackQuorumSize)
             .setEnsembleSize(ensembleSize).setLength(length)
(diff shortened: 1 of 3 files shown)


sijie:
well I think there are two questions here:

- if we remove ledger metadata here, this logging statement is most likely useless. is that expected behavior we want? If the concern is about `password`, can't we just not dump password instead of removing the entire ledger metadata here.

- "password": is safesforce really using this for authentication? we already have tls and correct authentication framework to do authentication. "password" should be considered dead and removed at some point though. I am asking this is mostly to learn your use case. AFAIK, the password is not really used at either yahoo or twitter.

eolivelli:
I can confirm that in my projects I use passwords only in order to double check that the ledger is the one that is to be read, for instance while having multiple libraries that share the same bk cluster.

Offtopic:
We already said with Charan that it would be good to add password checking wven for deleteLedger, but only ad a way of preventing human errors.

reddycharan:
yeah, even I've mixed opinion on the fix for it. We are getting string representation of LedgerMetadata from the serialize method and its underlying methods. I don't feel it is correct to change the logic there. Probably a new method can be added to LedgerMetadata say "toSafeString" which would call toString method and filter out 'password' field if it is present. It is going to be hacky, but I don't see any better solution if we want to continue logging the changed metadata.

sijie:
+1 on "toSafeString"

eolivelli:
Good idea toSafeString

athanatos:
toSafeString works for me!

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
```diff
@@ -336,8 +340,9 @@ LedgerMetadataFormat buildProtoFormat() {
             builder.setCtime(ctime);
         }
 
-        if (hasPassword) {
-            builder.setDigestType(digestType).setPassword(ByteString.copyFrom(password));
```
reddycharan: @sijie here, I was not sure why builder.setDigestType is kept in if block. I moved it out of this if block. I guess, it is the right thing to do.

sijie: I think it was initially for BC, because we don't put any password and digesttype in metadata long time ago. so when it was introduced, password and digesttype are added at the same time. so that was the reason why digest is skipped. 

I am not sure what version introduced, it was probably 4.1.0 or 4.2.0. However I would suggest keeping the logic as same as before. 

```
if (hasPassword) {
   builder.setDigestType(digestType);
   if (withPassword) {
       ...
   }
}
```

we will have a separate issues to clean up these BC related legacy code when we enforce EOL of releases.

reddycharan:
@sijie reverted the change as you suggested for DigestType (for the sake of BC), I think this should be good to go.

sijie:
+1 will merge it soon

Issue #1000, opened 2018-01-17
BP-27: New BookKeeper CLI


sijie:
*BP*

This is the master ticket for tracking BP-27:

`BookieShell` is becoming a gaint class to contain all the sub command. It is a good time to think about how to break this class down:

- group the commands into categories, e.g. `bookie`, `cluster`, `metadata`.
- break this class into smaller classes for different commands, and improve the testability
- it is probably good to rethink using jcommander to implement this.

This BP proposes refactoring the bookkeeper commands in `BookieShell` in new CLI tool.

Proposal PR - #1093

jiazhai:
made a rough group based on the 4 groups,
```
## bookie
Done： lastmark

Todo:

- CMD_BOOKIEFORMAT
 // Formats the local data present in current bookie server.
class BookieFormatCmd extends MyCommand

- CMD_RECOVER
 // Recover command for ledger data recovery for failed bookie.
class RecoverCmd extends MyCommand

- CMD_READLOG
 // Command to read entry log files.
class ReadLogCmd extends MyCommand

- CMD_READLOGMETADATA
// Command to print metadata of entrylog.

class ReadLogMetadataCmd extends MyCommand

- CMD_READJOURNAL
// Command to read journal files.

class ReadJournalCmd extends MyCommand

- CMD_LISTFILESONDISC
List the files in JournalDirectory/LedgerDirectories/IndexDirectories
class ListDiskFilesCmd extends MyCommand

- CMD_DECOMMISSIONBOOKIE
// Command to trigger AuditTask by resetting lostBookieRecoveryDelay and then make sure the
// ledgers stored in the bookie are properly replicated.
class DecommissionBookieCmd extends MyCommand


## cluster
done： listBookies

Todo:
- CMD_BOOKIEINFO
// Command to retrieve bookie information like free disk space, etc from all the bookies in the cluster.
class BookieInfoCmd extends MyCommand

- CMD_INITNEWCLUSTER
// Intializes new cluster by creating required znodes for the cluster.
class InitNewCluster extends MyCommand

- CMD_NUKEEXISTINGCLUSTER
// Nuke bookkeeper metadata of existing cluster in zookeeper.
class NukeExistingCluster extends MyCommand {
}

- CMD_LEDGER
//  dump ledger index entries into readable format.
class LedgerCmd extends MyCommand {

- CMD_READ_LEDGER_ENTRIES
// Command for reading ledger entries.
[...]

PR #1093, opened 2018-01-31
BP-27: New BookKeeper CLI

sijie:

`BookieShell` is the current bookkeeper cli for interacting and operating a bookkeeper cluster. However, this class is getting bigger with more commands added to it. It is facing a few problems for maintenance and extensibility.

- All commands sit in one gaint shell class. It is hard to tell if a command is used for managing a bookie only or if a command is used for managing a cluster.
- Lack of unit tests. This class has very few test coverage. Most of the commands (introduced in early days) don't have a unit test.
- Lack of extensibility. If a new function component (for example, dlog) is introduced, it is a bit hard to extend this CLI to have commands for new function component.

All these problems lead to the proposal here. This proposal is to propose refactoring/redesigning the bookkeeper CLI to allow better managebility for maintenance, better test coverage and better extensibility for new function components.

Master Issue: #1000

### DIFF
--- a/site/bps/BP-27-new-bookkeeper-cli.md
+++ b/site/bps/BP-27-new-bookkeeper-cli.md
@@ -1,0 +1,109 @@
+---
+title: "BP-27: New BookKeeper CLI"
+issue: https://github.com/apache/bookkeeper/1000
+state: 'Under Discussion'
+release: "N/A"
+---
+
+### Motivation
+
+`BookieShell` is the current bookkeeper cli for interacting and operating a bookkeeper cluster. However, this class is getting bigger with more commands added to it. It is facing a few problems for maintenance and extensibility.
+
+- All commands sit in one gaint shell class. It is hard to tell if a command is used for managing a bookie only or if a command is used for managing a cluster.
+- Lack of unit tests. This class has very few test coverage. Most of the commands (introduced in early days) don't have a unit test.
+- Lack of extensibility. If a new function component (for example, dlog) is introduced, it is a bit hard to extend this CLI to have commands for new function component.
+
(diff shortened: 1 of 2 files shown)


sijie:
Put up a PR #1094 for demonstrating the idea.

sijie:
> Can I suggest that as you submit the changes, you submit each command individually, so that it's easier to review?

you mean this PR in #1094 or you mean the changes in future?

If you mean #1094 , the change there is mainly for setting up the framework and only 3 commands on each category is moved. so I think it is already a minimal change that I can get to for this review - you can get an overview on how the new CLI looks like.

If you mean the changes in futures, yes, there will be and should be one command per PR.

> Integration tests for these would also be good, since unit testing can't test the bash script part.

You don't need to test bash script every command. The testing of bash script should mainly just focus on if it is taking the environment settings correctly, invoking the CLI program correctly. That means the tests should be:

1) bash script testing: focus on testing the script itself, if it is taking the environment settings correctly, invoking the CLI program correctly. And it should stop after testing it can invoke the program. 

2) CLI program testing: focus on testing parsing and ensure the correct command is invoking and right parameters are passing into command.

3) Command logic testing: focus on testing the logic for individual commands.

This BP covers both 2) and 3). while 1) is related but it is a different topic in general, we need to test all the scripts we shipped along with the release. so it will have to be a different task than the one here.

Hope this make the clarification.

ivankelly:
> you mean this PR in #1094 or you mean the changes in future?

Future changes. #1094 is already 2.5K loc, which is going to take a long time to give a good review.

> This BP covers both 2) and 3). while 1) is related but it is a different topic in general, we need to test all the scripts we shipped along with the release. so it will have to be a different task than the one here.

Sure.

sijie:
@ivankelly create #1102 for bash script testing.

ivankelly:
👍

PR #1094, opened 2018-01-31
BP-27 (part 1): New BookKeeper CLI Skeleton

sijie:

This is first implementation of BP-27 (#1000). This change includes:

- introduce a bookkeeper-tools module for keep the commands and also for developing the new CLI
- introduce 4 command groups: `cluster`, `bookie`, `client`, `metadata`.
- move 3 command: `listbookies` under `cluster` group, `lastmark` under `bookie` group, `simpletest` under `client` group.
- unit tests for those 3 command.


sijie:
This change is ready to review now

sijie:
@jvrao @reddycharan: since you guys contributed a lot of shell commands, it would be great that you guys can spend some time on reviewing this one.

bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/commands/client/SimpleTestCommand.java
```diff
@@ -0,0 +1,80 @@
...
+    @Override
+    public String name() {
+        return "simpletest";
+    }
+
+    @Override
+    protected void run(BookKeeper bk) throws Exception {
+        byte[] data = new byte[100]; // test data
+
+        try (WriteHandle wh = result(bk.newCreateLedgerOp()
```
reddycharan: where is this result method defined?

bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/commands/client/SimpleTestCommand.java
```diff
@@ -0,0 +1,80 @@
...
+            .withEnsembleSize(ensembleSize)
+            .withWriteQuorumSize(writeQuorumSize)
+            .withAckQuorumSize(ackQuorumSize)
+            .withDigestType(DigestType.CRC32C)
```
reddycharan: in the old code it is MAC, I guess this change is intentional here

bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/commands/cluster/ListBookiesCommand.java
```diff
@@ -0,0 +1,80 @@
...
+    @Override
+    protected void run(RegistrationClient regClient) throws Exception {
+        List<BookieSocketAddress> bookies = Lists.newArrayList();
+        if (readwrite) {
+            bookies.addAll(
+                result(
+                    regClient.getWritableBookies()
+                ).getValue()
+            );
+            bookies.addAll(
```
reddycharan: in the previous code we were reporting readwrite bookies only in this case

bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/commands/client/SimpleTestCommand.java
sijie: yeah. I changed it to CRC32C intentionally, because I think CRC32C is far faster than MAC. since it is a test command, I don't think this change would cause any BC concerns. unless I misunderstood this command.

bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/commands/cluster/ListBookiesCommand.java
sijie: oops. my bad. I will change it back to report readwrite bookies only. nice catch.

sijie: done. reverted the semantic to only printing readwrite bookies for "-rw" flag.

sijie:
@reddycharan I think I addressed all your comments.

sijie:
ping @reddycharan

bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/commands/client/SimpleTestCommand.java
```diff
@@ -0,0 +1,80 @@
...
+public class SimpleTestCommand extends ClientCommand {
+
+    @Parameter(names = { "-e", "--ensemble-size" }, description = "Ensemble size (default 3)")
+    private int ensembleSize = 3;
+    @Parameter(names = { "-w", "--write-quorum-size" }, description = "Write quorum size (default 2)")
+    private int writeQuorumSize = 2;
+    @Parameter(names = { "-a", "--ack-quorum-size" }, description = "Ack quorum size (default 2)")
+    private int ackQuorumSize = 2;
+    @Parameter(names = { "-n", "--num-entries" }, description = "Entries to write (default 100)")
+    private int numEntries = 100;
```
reddycharan: in previous code, default value was 1000

bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/commands/cluster/ListBookiesCommand.java
```diff
@@ -0,0 +1,75 @@
...
+                    regClient.getWritableBookies()
+                ).getValue()
+            );
+        } else if (readonly) {
+            bookies.addAll(
+                result(
+                    regClient.getReadOnlyBookies()
+                ).getValue()
+            );
+        }
```
reddycharan: I see 2 issues here,

1) how can we make sure only one option is set
2) how can we make sure atleast one option is set

bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/helpers/DiscoveryCommand.java
```diff
@@ -0,0 +1,59 @@
...
+                executor,
+                NullStatsLogger.INSTANCE,
+                Optional.empty());
+            run(regClient);
+        }
+    }
+
+    @Override
+    protected void run(BookKeeper bk) throws Exception {
+        throw new UnsupportedOperationException("Discovery command only uses registration client");
```
reddycharan: does this has to be UnsupportedOperation?

shouldn't we use bk's regClient (bk.getRegClient()) ?

bookkeeper-tools/src/main/java/org/apache/bookkeeper/tools/cli/commands/CmdBase.java
```diff
@@ -0,0 +1,91 @@
...
+    protected final ServerConfiguration conf;
+
+    protected CmdBase(String cmdName, ServerConfiguration conf) {
+        this(cmdName, conf, new JCommander());
+    }
+
+    protected CmdBase(String cmdName, ServerConfiguration conf, JCommander commander) {
+        this.conf = conf;
+        this.commander = commander;
+        this.commander.setProgramName("bookie-shell " + cmdName);
```
reddycharan: is this hyphen required?

bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/commands/client/SimpleTestCommand.java
sijie: yes. I do this intentionally, because for a simple test command it is okay to just test with 100 entries, 1000 seems to too many for me.

but I didn't change the original behavior for old bookie shell, this behavior is only changed for new bookie shell. so it doesn't break any BC here.

bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/commands/cluster/ListBookiesCommand.java
sijie: nice catch. will add the validation

bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/helpers/DiscoveryCommand.java
sijie: this is done intentionally. for `discovery command`, we use `run(RegistrationClient)` not `run(BookKeeper)`, but we make DiscoveryCommand inherits ClientCommand. making it throw unsupported exception is to prevent anyone using this branch.

bookkeeper-tools/src/main/java/org/apache/bookkeeper/tools/cli/commands/CmdBase.java
sijie: it isn't.

bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/commands/cluster/ListBookiesCommand.java
sijie: done.

1. add a `validateArgs` method to let each command validate args before actual running. 
2. enforce "One and only one of --readwrite and --readonly must be specified"

bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/helpers/DiscoveryCommand.java
sijie: done. let `DiscoveryCommand` inherits `Command` directly, instead of `ClientCommand`.

sijie:
@reddycharan I have addressed your comments. please review it again.

bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/commands/cluster/ListBookiesCommand.java
yzang: Is there any command to list all bookies and their state? For example, by default, we list all bookies running in the cluster and their running mode if no args is given. If we specify -ro, we only list read only bookies?

sijie:
> Is there any command to list all bookies and their state? For example, by default, we list all bookies running in the cluster and their running mode if no args is given. If we specify -ro, we only list read only bookies?

Good point! I think the original change I made supports listing all bookies by default. I reverted it to limit to use either "-ro" or "-rw" only as the old CLI does per @reddycharan 's suggestion.

@reddycharan do you think we should remove this limitation on new bookkeeper CLI to support "list all bookies" if no arg is provided or both args are provided.

reddycharan:
yeah, there is value in printing all bookies if no option is set. But in your earlier change you were printing all bookies for -rw option.

sijie:
@reddycharan yes. I fixed that after your first comment. your second comment was asking for printing bookies when "one and only one flag is set". so does it make sense for me to revert that? then the behavior will be:

- "-rw" prints read/write bookies
- "-ro" prints readonly bookies
- no args prints both
- "-rw" and "-ro" print both

can we agree on this is the expected behavior?

reddycharan:
I have an issue with "-rw" and "-ro" print both". In your logic if both the options are provided then you are printing both "-rw" and "-ro" bookies with no distinction. If you are distinguishing them with enough log messages ("List of ReadWrite bookies: .. List of ReadOnly bookies: ..") then ok

sijie:
@reddycharan okay I will update headlines like "readwrite bookies" and "readonly bookies". does that make sense?

reddycharan:
yes, with clear logging it should be good.

bin/bookkeeper-cli
```diff
@@ -0,0 +1,153 @@
+#!/usr/bin/env bash
```
reddycharan: any reason for keeping it in bookkeeper/bin folder instead of bookkeeper/bookkeeper-server/bin ?

bin/bookkeeper-cli
```diff
@@ -0,0 +1,153 @@
...
+# * distributed under the License is distributed on an "AS IS" BASIS,
+# * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# * See the License for the specific language governing permissions and
+# * limitations under the License.
+# */
+
+BINDIR=`dirname "$0"`
+BK_HOME=`cd $BINDIR/..;pwd`
+
+DEFAULT_CONF=$BK_HOME/conf/bk_server.conf
```
reddycharan: here $BK_HOME would be 'bookkeeper' folder but not 'bookkeeper/bookkeeper-server' folder

bin/bookkeeper-cli
sijie: the thought here is eventually moving bin/ conf/ directories to the root module. since we will split bookkeeper-server into client and server module in future, and bookkeeper-dist is for packing the binary distribution. moving bin to top-level will make things easier and manageable. I am doing this for this new bookkeeper CLI first and will move other script after we split bookkeeper-server module in 4.8.

reddycharan: to un this shell script, do I have to config / set anything, I'm getting following error

/Users/cguttapalem/Workspace/Community/bookkeeper/bin
cguttapalem@cguttapale-ltm1:~/Workspace/Community/bookkeeper/bin$ ./bookkeeper-cli 
JAVA_HOME not set, using java from PATH. (/usr/bin/java)
Fail to load sub command 'cluster' : null
java.lang.reflect.InvocationTargetException
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:422)
	at org.apache.bookkeeper.tools.cli.BookieShell.newCommandInstance(BookieShell.java:69)
	at org.apache.bookkeeper.tools.cli.BookieShell.setupShell(BookieShell.java:122)
	at org.apache.bookkeeper.tools.cli.BookieShell.run(BookieShell.java:142)
	at org.apache.bookkeeper.tools.cli.BookieShell.main(BookieShell.java:205)
Caused by: java.lang.NoClassDefFoundError: org/apache/bookkeeper/stats/StatsLogger
	at org.apache.bookkeeper.tools.cli.commands.CmdCluster.<init>(CmdCluster.java:32)
	... 8 more
Caused by: java.lang.ClassNotFoundException: org.apache.bookkeeper.stats.StatsLogger
	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:331)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	... 9 more

sijie:
@reddycharan I updated this PR to make listbookies to list all bookies if no args is provided or both args are provided.

sijie:
retest this please

bin/bookkeeper-cli
reddycharan: in bookkeeper/conf folder there isn't bk_server.conf, it is in bookkeeper/bookkeeper-server/conf folder

so getting following error -

Failed to load configuration file '/Users/cguttapalem/Workspace/Community/bookkeeper/conf/bk_server.conf' : Unable to load the configuration from the URL file:/Users/cguttapalem/Workspace/Community/bookkeeper/conf/bk_server.conf

bin/bookkeeper-cli
reddycharan: @sijie after rebasing changes and rebuilding, this error is not reproing now, wondering in which commit you fixed it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/commands/bookie/LastMarkCommand.java
```diff
@@ -0,0 +1,61 @@
...
+import org.apache.bookkeeper.bookie.LogMark;
+import org.apache.bookkeeper.conf.ServerConfiguration;
+import org.apache.bookkeeper.tools.cli.helpers.BookieCommand;
+import org.apache.bookkeeper.util.DiskChecker;
+
+/**
+ * A bookie command to print the last log marker.
+ */
+@Parameters(commandDescription = "Print last log marker")
+public class LastMarkCommand extends BookieCommand {
```
reddycharan: cguttapalem@cguttapale-ltm1:~/Workspace/Community/bookkeeper/bin$ ./bookkeeper-cli bookie

Usage: bookie-shell bookie [options] [command] [command options]
  Commands:
    lastmark      Print last log marker
      Usage: lastmark [options]

------------------------------------

in the last line of the output, there shouldn't be "[options]" in "Usage" line, since there are no options for "lastmark" command

bin/bookkeeper-cli
sijie: do you run with a clean build after you switch the branch?

bin/bookkeeper-cli
sijie: yes. that is done intentionally at this PR. was planning to move the bin/ and conf/ to root directory after we complete the new bookkeeper CLI. otherwise we will have two bk_server.conf in place, which make things a bit hard to keep it in sync.

for testing purpose, you can manually copy bookkeeper-server/conf/bk_server.conf to conf.

If you feel strong that we need to copy bookkeeper-server/conf/bk_server.conf  right now, we don't have to wait until new bookkeeper cli is completed, I am fine to copy the file as well.

bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/commands/bookie/LastMarkCommand.java
sijie: @reddycharan this output is managed by jcommander. I don't think it is easy to get round at this moment. is this a real concern?

reddycharan: cguttapalem@cguttapale-ltm1:~/Workspace/Community/bookkeeper/bin$ ./bookkeeper-cli 

Usage: bookie-shell [options] [command] [command options]
  Options:
    -c, --conf
       Bookie Configuration File
    -h, --help
       Show this help message
       Default: false
  Commands:
    bookie      Commands on operating a single bookie
      Usage: bookie [options]

    client      Commands that interact with a cluster
      Usage: client [options]

    cluster      Commands that operate a cluster
      Usage: cluster [options]

    metadata      Commands that interact with metadata storage
      Usage: metadata [options]

--------------------------------------------------------------------
1) it is definitely not obvious for the first time users

2) In the following line what does "bookie-shell" refer to? because when I tried "./bookkeeper-cli bookie-shell bookie" it doesn't output just "bookie" command  usage.

Usage: bookie-shell [options] [command] [command options]

3) in the existing shell we have couple of options for commands - ledgeridformat and entryformat. It seems it is missing in the new CLI

4) I'm concerned about flexibility with jcommander framework for commands with complicated options like 'readlog' command - (readlog      [-msg] <entry_log_id | entry_log_file_name> [-ledgerid <ledgerid> [-entryid <entryid>]] [-startpos <startEntryLogBytePos> [-endpos <endEntryLogBytePos>]])

bin/bookkeeper-cli
reddycharan: probably not..and yeah that could be reason for why it didnt repro without any code change

bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/commands/bookie/LastMarkCommand.java
reddycharan: 5. What is the expected output for the command "./bookkeeper-cli --help bookie" ? isn't it supposed to print help message just for bookie command?

It is printing the full output of bookkeeper-cli

cguttapalem@cguttapale-ltm1:~/Workspace/Community/bookkeeper/bin$ ./bookkeeper-cli --help bookie
 
Usage: bookie-shell [options] [command] [command options]
  Options:
    -c, --conf
       Bookie Configuration File
    -h, --help
       Show this help message
       Default: false
  Commands:
    bookie      Commands on operating a single bookie
      Usage: bookie [options]

    client      Commands that interact with a cluster
      Usage: client [options]

    cluster      Commands that operate a cluster
      Usage: cluster [options]

    metadata      Commands that interact with metadata storage
      Usage: metadata [options]

reddycharan: 6) in the output for the following command - what does "[options]" referred to in the following line

"Usage: bookie-shell client [options] [command] [command options]"

cguttapalem@cguttapale-ltm1:~/Workspace/Community/bookkeeper/bin$ ./bookkeeper-cli client
JAVA_HOME not set, using java from PATH. (/usr/bin/java)
Usage: bookie-shell client [options] [command] [command options]
  Commands:
    simpletest      Simple test to create a ledger and write entries to it.
      Usage: simpletest [options]
        Options:
          -a, --ack-quorum-size
             Ack quorum size (default 2)
             Default: 2
          -e, --ensemble-size
             Ensemble size (default 3)
             Default: 3
          -n, --num-entries
             Entries to write (default 100)
             Default: 100
          -w, --write-quorum-size
             Write quorum size (default 2)
             Default: 2

reddycharan:
@sijie  sounds good..I'll wrap up this CR by the end of today. For sure by tomorrow I'll give my +1

sijie:
@reddycharan updated this PR to use "bookkeeper-cli" instead of "bookie-shell" in the command output, to avoid confusion.

bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/commands/bookie/LastMarkCommand.java
sijie: "bookie-shell client [options] [command] [command options]" this is formatted by jcommander.
 
for a given command, it has options (seen as "[options]") and then sub-commands ("[command]"). so here "[options]" is the options of "client" command. but there is no "options" defined for "client" command, so it is not printed in the output. it only prints the "commands" for "client" command: "simpletest"

bookkeeper-tools/src/main/java/org/apache/bookkeeper/tools/cli/BookKeeperCLI.java
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    static JCommander newJCommander() {
+        return new JCommander();
+    }
+
+    @SuppressWarnings("unchecked")
+    @VisibleForTesting
+    public static Object newCommandInstance(Class cls, ServerConfiguration config) throws Exception {
+        return cls.getConstructor(ServerConfiguration.class).newInstance(config);
```
reddycharan: any reason for using reflection? instead of instantiating directly in the static method (line 52-59)

sijie: no particular reason for using reflection. was trying to defer object initialization when it is needed. any concern on using reflection?

reddycharan: I would prefer not to use it unless there is strong need for reflection. In this case you are doing both adding commands class and instantiating them in the same class (BookKeeperCLI.java), So I dont see the need for it. Also it would be little annoying to findout from call hierarchy who are creating instances of that class and which constructor is used. Also tomorrow if we change the signature of the constructors without changing the reflection instantiation calls, reflection instantiation would fail with in runtime instead of compile time.

reddycharan:
"@reddycharan updated this PR to use "bookkeeper-cli" instead of "bookie-shell" in the command output, to avoid confusion."

im not in favor of having '-' in command name and script file name. People make all kind of mistakes with usage of '-' in the script file name/command name. Probably removing '-' and making it "bookkeepercli" might be better. But again we are having this script file (bookkeeper-cli) only for running shell commands right? its not for starting bookies, or localbookie, right? in that case having word shell in the script file name or command name might be better say - "bookieshellcli" or "bookkeepershellcli". If we are planning to have bookie start and localbookie start functionality as well in that script then we can just leave it as "bookkeepercli"

reddycharan:
overall LGTM. Other than the need of reflection and naming script/command with '-'. Regarding help usage message, it can be improved, like it is showing [options] unnecessarily in couple of places and from the help usage command it is not clear what is command vs sub-command. But for now I'm fine with it, hoping it can be improved/enhanced later.

bookkeeper-tools/src/main/java/org/apache/bookkeeper/tools/cli/BookKeeperCLI.java
sijie: @reddycharan my concern of instantiating all the commands is we basically construct all the commands, which it is completely waste resources since there is only command group is going to touched. if you check the code, we only use it for command group, for each command group we construct the sub commands. 

How about keep this for now? since it is a new CLI tool, we can decide and change what is the best solution after porting all the changes. At that point, if you still think we should not use reflection, we can change it.

sijie:
> People make all kind of mistakes with usage of '-' in the script file name/command name.

what are the mistakes of using "-" in script name?

I am using 'bookkeeper-cli' is to avoid using same name or similar as 'bookie shell'. because these two scripts need to co-exists for a while before we can deprecate "bookie shell". I would like to avoid confusion. 

If this concern is more about "naming", I would like to defer the official to after finishing porting all sub commands. It is still unclear to me why "-" is bad.

bookkeeper-tools/src/main/java/org/apache/bookkeeper/tools/cli/BookKeeperCLI.java
reddycharan: it is ok for now.

reddycharan:
it is matter of convenience and less error-prone to not have '-' in script filename/command name. When we spell it, people might take it incorrectly as an option, "bookkeeper -cli". Anyhow I'll leave the decision to public opinion.

 LGTM! for the first iteration.

sijie:
@reddycharan good point about "-". Created an issue #1153 for finalizing the new bookkeeper-cli name.

asfgit:

sijie:
retrigger the travis CI

sijie:
all CI are green except "integration tests". since this PR is about a new CLI tool, it is irrelative to integration tests. going to ignore CI to merge it.

sijie:
IGNORE CI

PR #1097, opened 2018-02-01
Use loopback network interface for testcases.

reddycharan:

- tests would be more reliable irrespective of the environment (and availability of network connection), if the loopback address is used in testsuites.
- unless loopback address is set explicitly, in my env (while on company's VPN),
UpdateCookieCmdTest is failing most of the times.
- currently allowLoopback is set to true in BookKeeperClusterTestCase,
but it doesn't make Bookies to use loopback interface address.
- loopback network interface should be set explicitly as
listening interface to use loopback address.

Note: 

1) will create Issue for this depending on the PR feedback.
2) in other places (where setAllowLoopback is set) also loopback interface needs to be set


athanatos:
lgtm

jiazhai:
retest this please

reddycharan:
I ran locally AuditorLedgerCheckerTest and BookieClientTest multiple times and they are running fine. Not sure why tests failed here. Also different tests in Java8 and Java9 build.

reddycharan:
@sijie I made changes to the initial commit and replaced the initial commit. Can you sign-off on this. Thanks.

sijie:
@reddycharan I don't think the failures are related.

PR #1098, opened 2018-02-01
Run checkstyle plugin at validate phase

sijie:

Most of the modules run checkstyle plugin at validate phase. However a few modules are still running at `test-compile` phase.
Update the checkstyle plugins in those modules to run at validate phase. So the behavior is consistent across the project.


sijie:
I noticed this when working on #1095.

PR #1099, opened 2018-02-01
Fix auditor shutdown logic and move decommision tests out of BookKeeperAdminTest

sijie:

- the auditor shutdown logic is problematic. most of the tests can finish quickly however it spend more 30 seconds on shutting down.
  because the shutdown logic will be blocked until `awaitTermination` timed out.
- most of the tests in BookKeeperAdminTest don't need 6 bookies. so move the decommission tests to a separate class.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
@@ -707,8 +707,7 @@ public class Auditor {
      */
     public void shutdown() {
         LOG.info("Shutting down auditor");
-        submitShutdownTask();
-
+        executor.shutdown();
         try {
             while (!executor.awaitTermination(30, TimeUnit.SECONDS)) {
                 LOG.warn("Executor not shutting down, interrupting");
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieDecommissionTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieDecommissionTest.java
@@ -1,0 +1,176 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.bookkeeper.client;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+
+import java.util.Iterator;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.bookkeeper.bookie.Bookie;
+import org.apache.bookkeeper.client.BKException.BKIllegalOpException;
+import org.apache.bookkeeper.client.BookKeeper.DigestType;
+import org.apache.bookkeeper.conf.ServerConfiguration;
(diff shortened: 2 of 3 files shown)


sijie:
An example jenkins job on how long each test takes. Most of the tests run more than 1 minutes.

https://builds.apache.org/job/bookkeeper_precommit_pullrequest_java8/239/org.apache.bookkeeper$bookkeeper-server/testReport/org.apache.bookkeeper.client/BookKeeperAdminTest/

sijie:
@jvrao @reddycharan can you guys take the change here? I think the decommission test was contributed by you guys. It would be good if you guys can check to ensure my change doesn't change the testing purpose.

reddycharan:
will look into this. Here you have reduced numofbookies to 2 and moved decomm tests to a new testsuite with 6 bookies. Thats it, right?

sijie:
@reddycharan yes

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
```diff
@@ -707,8 +707,7 @@ public void processResult(int rc, String s, Object obj) {
      */
     public void shutdown() {
         LOG.info("Shutting down auditor");
-        submitShutdownTask();
-
+        executor.shutdown();
```
reddycharan: wondering why did you make this change? why are you replacing submitShutdownTask call?

sijie: I explained this in the description:

"the auditor shutdown logic is problematic. most of the tests can finish quickly however it spend more 30 seconds on shutting down.
because the shutdown logic will be blocked until awaitTermination timed out." 

if Auditor is running, it might be blocking in the thread, submit shutdown task will never be executed, then it will have to wait for 30 seconds timeout (`awaitTermination`) and issue `shutdownNow`. this change issues shutdown, which will interrupt the auditor task and shutdown the executor thread. 

Hope this make things clear

reddycharan: yeah I get that, but I'm wondering why it wasn't like this before? why did the original author - https://github.com/apache/bookkeeper/commit/e5b0dd0d9e47394202f5a20cedf409f487d90beb went through this extra round of hassle for shutdown deliberately. Is it ok now to not to have this extra scrutiny with the shutdown process?

sijie: No idea why it was done in that way. @ivankelly can chime in since he made that change.

ivankelly: Long time ago, but I think it was to avoid killing a running audits before it was done, since it's a long running task.

sijie: Okay so the question is can we change it to shutdown. Because even we don’t shutdown, it will be shutdown in 30 seconds.

sijie: @ivankelly can you check if it is okay for us to change it to shutdown?

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieDecommissionTest.java
```diff
@@ -0,0 +1,176 @@
...
+    private static DigestType digestType = DigestType.CRC32;
+    private static final String PASSWORD = "testPasswd";
+
+    public BookieDecommissionTest() {
+        super(NUM_OF_BOOKIES, 480);
+        baseConf.setOpenLedgerRereplicationGracePeriod(String.valueOf(30000));
+        setAutoRecoveryEnabled(true);
+    }
+
+    @FlakyTest("https://github.com/apache/bookkeeper/issues/502")
```
reddycharan: is it still flaky?

reddycharan:
LGTM +1

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieDecommissionTest.java
sijie: I have a subsequent change to remove FlakyTest at #1100

sijie:
retest this please

asfgit:

sijie:
IGNORE CI (All CI passed except "Integration Tests". This change is purely on changing tests. So it doesn't need "Integration Tests" to verify it)

sijie:
IGNORE CI

PR #1100, opened 2018-02-01
(WIP) Improve decommission test

### LINKED ISSUE #502: BookKeeperAdminTest.testDecommissionBookie is flaky
**BUG REPORT**

1. Please describe the issue you observed:

- What did you do?

run ci job

- What did you expect to see?

all tests should pass

- What did you see instead?

BookKeeperAdminTest.testDecommissionBookie failed.

https://builds.apache.org/job/bookkeeper-precommit-pullrequest/org.apache.bookkeeper$bookkeeper-server/336/testReport/junit/org.apache.bookkeeper.client/BookKeeperAdminTest/testDecommissionBookie/

sijie:

- remove the hardcoded 30 seconds wait time
- reduce `openLedgerRereplicationGracePeriod` from 30 seconds to 100 ms
- reenable `testDecommissionBookie` test case (this fixes #502)

This change is based on #1099.


sijie:
@jvrao @reddycharan can you guys take the change here? I think the decommission test was contributed by you guys. It would be good if you guys can check to ensure my change doesn't change the testing purpose.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperAdmin.java
```diff
@@ -1430,19 +1432,44 @@ public void triggerAudit()
...
-         * force audittask and let the underreplicationmanager process
-         * to do its replication process
-         */
-        Thread.sleep(30 * 1000);
+        try {
+            waitAuditorStrategy.call();
+        } catch (InterruptedException ie) {
+            throw ie;
+        } catch (Exception e) {
+            throw new TimeoutException("Encountered exception on waiting auditor auditing the bookies");
```
reddycharan: not sure if the TimeoutException is the appropriate one in this case.

sijie: okay any better exceptions?

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieDecommissionTest.java
```diff
@@ -86,7 +91,18 @@ public void testDecommissionBookie() throws Exception {
          * this decommisionBookie should make sure that there are no
          * underreplicated ledgers because of this bookie
          */
-        bkAdmin.decommissionBookie(Bookie.getBookieAddress(killedBookieConf));
+        bkAdmin.decommissionBookie(
+            Bookie.getBookieAddress(killedBookieConf),
+            () -> {
+                Auditor auditor = recoveryProcess.getAuditor();
+                if (null != auditor) {
```
reddycharan: why do we have this if block? in which case it can be null and is it ok to be null for our testscenario?

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieDecommissionTest.java
```diff
@@ -44,14 +46,17 @@
 
     public BookieDecommissionTest() {
         super(NUM_OF_BOOKIES, 480);
-        baseConf.setOpenLedgerRereplicationGracePeriod(String.valueOf(30000));
-        setAutoRecoveryEnabled(true);
+        baseConf.setOpenLedgerRereplicationGracePeriod(100);
+        setAutoRecoveryEnabled(false);
```
reddycharan: I'm not sure how can this be ok? you disabled AutoRecoveryService for all the bookies by setting this to false, but in each testcase you are starting  AutoRecoveryMain (which starts daemon for Auditor and ReplicationWorker) just for one bookie. How come having just one bookie ReplicationWorker would be sufficient for these testcases where we are killing bookie/s? 

My concern here is if the bookie which is having AutoRecoveryMain is already having copy of fragments of that ledger, then it would ignore replicating that under replicated ledger to that Bookie, right?

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookKeeperAdminTest.java
```diff
@@ -61,7 +61,7 @@
     public BookKeeperAdminTest() {
         super(numOfBookies, 480);
         baseConf.setLostBookieRecoveryDelay(lostBookieRecoveryDelayInitValue);
-        baseConf.setOpenLedgerRereplicationGracePeriod(String.valueOf(30000));
+        baseConf.setOpenLedgerRereplicationGracePeriod(500);
```
reddycharan: is this change required? does it affect anyhow? for the sake of simplicity cannt we just leave it to default value

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieDecommissionTest.java
```diff
@@ -44,14 +46,17 @@
 
     public BookieDecommissionTest() {
         super(NUM_OF_BOOKIES, 480);
-        baseConf.setOpenLedgerRereplicationGracePeriod(String.valueOf(30000));
-        setAutoRecoveryEnabled(true);
+        baseConf.setOpenLedgerRereplicationGracePeriod(100);
```
reddycharan: same comment as above

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieDecommissionTest.java
sijie: this is just for safety guard

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieDecommissionTest.java
sijie: the way I see this - what decommission test is doing here is to testing the decommission behavior. we are not testing killing auto recovery. we don't need to start autorecovery on every bookie, it is resource consumed and don't have more meanings to the test itself. move auto recovery process out of the bookies, it can make the auto recovery trigger more deterministic, rather than replying on non-deterministic sleeping behavior.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieDecommissionTest.java
```diff
@@ -86,7 +91,18 @@ public void testDecommissionBookie() throws Exception {
...
          * underreplicated ledgers because of this bookie
          */
-        bkAdmin.decommissionBookie(Bookie.getBookieAddress(killedBookieConf));
+        bkAdmin.decommissionBookie(
+            Bookie.getBookieAddress(killedBookieConf),
+            () -> {
+                Auditor auditor = recoveryProcess.getAuditor();
+                if (null != auditor) {
+                    auditor.submitLostBookieRecoveryDelayChangedEvent().get();
+                    log.info("Wait until auditor completes the auditing task.");
```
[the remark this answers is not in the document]
sijie: will update to a more meaningful sentence.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookKeeperAdminTest.java
sijie: well the point of having this change is to reduce the test time on this test case. 30 seconds is too much for a test... why do we need want to test 30 seconds...

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieDecommissionTest.java
sijie: same as above. the point of this change is to reduce the test time. we unnecessarily have a lot of such 30 seconds sleep in quite a few test cases. I would like to kill them or reduce them to a small value.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieDecommissionTest.java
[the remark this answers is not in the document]
reddycharan: no I mean, isn't the log line supposed to be before auditor.submitLostBookieRecoveryDelayChangedEvent() call?

log.info("Wait until auditor completes the auditing task.");
auditor.submitLostBookieRecoveryDelayChangedEvent().get();

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieDecommissionTest.java
reddycharan: hmm..in hypothetical scenario if it is null, then aren't we just ignoring it? is that ok to just ignore the fact that auditor is null and proceed with the test logic without waiting for auditor to do its auditing bookies task? 

Anyhow, I'm still not convinced how can we be ok with Auditor to be null? If there is no Auditor, then there is no replication process, right?

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieDecommissionTest.java
reddycharan: I dont get your point  - "we don't need to start autorecovery on every bookie"

lets say there are 6 Bookies - BK0, BK1, BK2,..BK5 and Ledger with quorum 3,3 is created.
you started AutoRecovery process for only one Bookie - BK0. So for BK0 Auditor and ReplicationWorker daemons are running.

LID-P (ledger) : BK0, BK1, BK2 (ensemble)

Now if BK1 is killed, so LID-P will be underreplicated. My question is though AutoRecovery (Auditor, ReplicationWorker) are running for BK0, LID-P will remain under-replicated, right? Since, BK0 is already part of ensemble of LID-P, ReplicationWorker running with BK0 wouldn't be able to do anything.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookKeeperAdminTest.java
reddycharan: ok..i forgot that in this test testDecommissionForLedgersWithMultipleSegmentsAndNotWriteClosed, there are not writeclosed ledgers and they will remain underreplicated for openLedgerRereplicationGracePeriod.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieDecommissionTest.java
reddycharan: ok..i forgot that in this test testDecommissionForLedgersWithMultipleSegmentsAndNotWriteClosed, there are not writeclosed ledgers and they will remain underreplicated for openLedgerRereplicationGracePeriod. 

Anyhow isn't 100 millsec too less?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperAdmin.java
reddycharan: BKAuditException? even in the InterruptedException case also. 

This pattern is used in other places also

https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java#L692

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieDecommissionTest.java
sijie: well I think the test is to wait until the open fragments are fenced (unless I missed something). so if the test is to wait until open fragments are fenced and replicated, then 100ms isn't too less.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieDecommissionTest.java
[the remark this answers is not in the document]
sijie: I mean my intention to add logging after that, so I will update the sentence to a more meaningful sentence.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieDecommissionTest.java
sijie: @reddycharan I think you have a wrong understanding on how autorecovery works. autorecovery is a completely isolated thing from bookies, it doesn't really stick to any bookies. It can be co-run with bookies, or it can be run as separate service. that means you can run less or more autorecovery instances that bookies.

in your example, if BK1 is killed, autorecovery daemon will detect that and choose a bookie to replace BK1 and replicate entries from BK0 and BK2 to the newly chosen bookie.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieDecommissionTest.java
sijie: > If there is no Auditor, then there is no replication process, right?

If there is no auditor, there is no replication process then the test will fail.

The logic here is just to avoid throw NPE (it is a coding style problem). I can change it to checkNotNull(auditor) to explicitly throw NPE if you like.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieDecommissionTest.java
reddycharan: ok..had private chat with Sijie. https://github.com/apache/bookkeeper/commit/6fcabfc80c5bd6cdfa9252cdfc4ed209cc0620a6 change has decoupled ReplicationWorker and Bookie completely, and now there is no "targetBookie" factor in ReplicationWorker. So I'm good with this change. On a different note, ReplicationWorker.java javadoc is not updated with that change and it is still using "targetBookie" in its comments.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieDecommissionTest.java
reddycharan: since recoveryProcess.start() is async call here, as we discuss wait until auditor is not null here.

sijie:
@reddycharan since this branch depends on #1099 . I will address your comments here as we discussed offline after #1099 is in.

reddycharan:
LGTM +1

asfgit:

sijie:
rebase to latest master and address the comments to wait until auditor is started.

@reddycharan can you review it again?

jvrao:
Looks like reviews are finished on this pull request. Jenkins maven install X is blocking the merge?

sijie:
@jvrao BookieDecommissionTest.testDecommissionBookie is still flaky. since this change is to remove `@FlakyTest` annotation, I have to investigate more to see why it is still flaky at jenkins job before we are confident to merge it.

sijie:
Turned this to WIP since it is still a flaky test. will work on and turn it back to review after I found the problem.

PR #1108, opened 2018-02-01
Replace DoubleByteBuf with CompositeByteBuf because of perf regression with Netty > 4.1.12

merlimat:
Starting from Netty-4.1.13 some internal behavior has changed and that has introduced a performance penalty when using the `DoubleByteBuf`.

The problem resides in the fact that netty falls back to calling `nioBuffers()` and that is creating an array of unpooled direct `ByteBuffers`. That is pretty heavy and the JVM is taking 1 to 3 seconds GC pause to reclaim them.

Short term fix is to go back to `CompositeByteBuf` which doesn't have the problem (the `DirectByteBuffer` instances are pooled and reused). I have verified that GC is back to normal after this patch.

More mid-term, since `CompositeByteBuf` allocates a bunch of objects (counted ~10 per entry) I have a change to replace it with a `ByteBufList` holder class. I have the change almost done but it will take a while to test it to verify corner cases.

This should be considered for a 4.6.2 fix release.


ivankelly:
@merlimat have you considered pushing the modified compositebytebuf back to netty?

jvrao:
@dlg99 Please have a look.

+1 form my side; looks like a good short-term step.

merlimat:
> @merlimat have you considered pushing the modified compositebytebuf back to netty? 

Right now, in Pulsar I have changed it to `ByteBufPair` which is very simple in that it's just a holder of 2 buffers, without trying to mimic the buffer interface. 
https://github.com/apache/incubator-pulsar/blob/40d37b10731707724b28f11254c4a72697f57bee/pulsar-common/src/main/java/org/apache/pulsar/common/api/ByteBufPair.java

For BK, it's a bit more involved because we have 3 buffers in some places. I have a WIP at https://github.com/merlimat/bookkeeper/commit/c7d2b2ef6bafba32d8697edd673475e3376360e9 though it requires a bit more testing

If the `ByteBufPair` and/or `ByteBufList` are generally useful, I surely will submit to Netty.

dlg99:
I experimented with getting rid of DoubleByteBuffer completely (on older/internal codebase).

allocations are from its

```java
        ByteBuffer dst = isDirect() ? ByteBuffer.allocateDirect(length) : ByteBuffer.allocate(length);
```
Composite also ends up doing 
```java
        ByteBuffer merged = ByteBuffer.allocate(length).order(order());
```
if it is wrapped around > 1 buffers.
so you are trading a chance for direct buffer allocation for always on-heap allocation. 

We use DoubleByteBuf to wrap (header, data) tuple i.e. in computeDigestAndPackageForSending.
Header is of fixed size for given ledger. so I got rid of DoubleByteBuf and preallocated enough space in single buffer to have data after offset for header and populated it. 
I am not saying that this is the best way to do this but either getting rid of it, or having header/digest as a separate field in proto, or pooling buffers in nioBuffer().

hope I am making sense.

jvrao:
> preallocated enough space in single buffer to have data after offset for header and populated it
Depending on the digest our header size changes. Hence we need to take care of that situation too.

merlimat:
> allocations are from its
> 
>     public ByteBuffer nioBuffer(int index, int length) {
>         ByteBuffer dst = isDirect() ? ByteBuffer.allocateDirect(length) : ByteBuffer.allocate(length);
> 
> Composite also ends up doing
> 
>         ByteBuffer merged = ByteBuffer.allocate(length).order(order());
> 
> if it is wrapped around > 1 buffers.
> so you are trading a chance for direct buffer allocation for always on-heap allocation.

I don't see the heap being used when wrapping direct buffers. I got a memory dump when running after the patch and I don't see the heap being polluted.

> We use DoubleByteBuf to wrap (header, data) tuple i.e. in computeDigestAndPackageForSending.
> Header is of fixed size for given ledger. so I got rid of DoubleByteBuf and preallocated enough space > in single buffer to have data after offset for header and populated it.

Though this would mean to copy each payload into a buffer that contains header+payload, right?

dlg99:
@jvrao 
> Depending on the digest our header size changes. Hence we need to take care of that situation too.

it is set for ledger when ledger is created.

@merlimat 

.allocate() creates HeapByteBuffer which is backed by byte[]
May I suggest using java mission control to profile memory allocations instead of heap dumps?
Let me know if you want me to send steps on running it.

> Though this would mean to copy each payload into a buffer that contains header+payload, right?

As I mentioned, I did the changes on old code where client did not rely on netty buffers that much. 
But the idea was to allocate (header length + data length) space at least (i.e. from pool), write payload with offset, fill in header later.  Client would get (buffer, int dataOffset, int dataLen) as parameters. If 0 to dataOffset is enough for the header just write it there and return slice.

merlimat:
> .allocate() creates HeapByteBuffer which is backed by byte[]
> May I suggest using java mission control to profile memory allocations instead of heap dumps?
> Let me know if you want me to send steps on running it.

I got allocation tracing from YourKit. I think the difference is that I'm writing with v2 BK protocol (no protobuf). With v3 we're forcing a byte[] because there's no way to have protobuf to serialize on direct memory (other than patching protobuf....)

sijie:
marked this for both 4.6.2 and 4.7.0

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/DoubleByteBuf.java
```diff
@@ -17,447 +17,20 @@
...
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.OutputStream;
-import java.lang.reflect.Constructor;
-import java.nio.ByteBuffer;
-import java.nio.ByteOrder;
-import java.nio.channels.FileChannel;
-import java.nio.channels.GatheringByteChannel;
-import java.nio.channels.ScatteringByteChannel;
+public class DoubleByteBuf extends CompositeByteBuf {
```
sijie: we might need to add an updated javadoc comment back @merlimat

eolivelli:
Please fix checkstyle, which wants the java doc comment

sijie:
@merlimat do you need any help to fix the checkstyle problem, so we can move forward with merging this PR.

merlimat:
I'd close this one in favor of #1141

merlimat:
(maybe this PR could be used for 4.6.2, but for 4.7.0 I'd rather go with the #1441 approach)

sijie:
@merlimat yeah. this PR is good for 4.6.2. so if you can fix the checkstyle here, I can merge it to 4.7.0 and 4.6.2. and #1441 can come later in master for 4.7.0. How does that sound?

merlimat:
Fixed javadocs

PR #1113, opened 2018-02-02
BP-28: Etcd as metadata store

sijie:
Descriptions of the changes in this PR:

This BP is exploring using Etcd as the metadata store.

PR #1114, opened 2018-02-02
(WIP) BP-28 (prototype): use Etcd as metadata store

sijie:

This is a prototype change for demonstrating how to implement the metadata interfaces using Etcd api for BP-28 (#1113)

*BE AWARE*

This is only for prototype purpose. The initial change was done one year and half ago and re-picked to current bookkeeper repo. As @jvrao requested, pushed it out to raise the discussion.

### DIFF
--- a/metadata-stores/etcd/pom.xml
+++ b/metadata-stores/etcd/pom.xml
@@ -1,0 +1,84 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+   Licensed to the Apache Software Foundation (ASF) under one or more
+   contributor license agreements.  See the NOTICE file distributed with
+   this work for additional information regarding copyright ownership.
+   The ASF licenses this file to You under the Apache License, Version 2.0
+   (the "License"); you may not use this file except in compliance with
+   the License.  You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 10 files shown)


sijie:
@jvrao as we discussed today, I pushed my prototype changes here and raised a BP for discussion.

jvrao:
I think this is waiting on the URI change. Once that is done, we can push this in. I would like some README on how to use this. Also, metadata tests to use both etcd and zk.

asfgit:

sijie:
@jvrao I assumed you approved the proposal itself, but not this change here, right? This change needs to update to complete it with URI changes.

metadata-stores/etcd/pom.xml
```diff
@@ -0,0 +1,84 @@
...
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.bookkeeper</groupId>
+            <artifactId>bookkeeper-server</artifactId>
+            <version>${project.parent.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.coreos</groupId>
+            <artifactId>jetcd-core</artifactId>
+            <version>0.0.1</version>
```
eolivelli: Nit: This version is referred in the other pom, maybe it is better to use here the variable

sijie:
changed this to WIP, since this is not reviewable. I will update this once #1117 (BP-29) is completed.

PR #1115, opened 2018-02-02
Change default ledger manager factory from `Flat` to `Hierachical`

sijie:

the discussion can be found at : http://mail-archives.apache.org/mod_mbox/bookkeeper-dev/201802.mbox/%3CCAO2yDyYdZ8%3D4tViNX1uL0Z67KX78JPD2XBGd7ermjO%3DK%3DscvcQ%40mail.gmail.com%3E

### DIFF
--- a/deploy/kubernetes/gke/bookkeeper.statefulset.yml
+++ b/deploy/kubernetes/gke/bookkeeper.statefulset.yml
@@ -39,8 +39,6 @@ data:
     BK_ledgerDirectories: "/bookkeeper/data/ledgers"
     BK_indexDirectories: "/bookkeeper/data/ledgers"
     BK_zkServers: zookeeper
-    # the default manager is flat, which is not good for supporting large number of ledgers
-    BK_ledgerManagerType: "hierarchical"
     # TODO: Issue 458: https://github.com/apache/bookkeeper/issues/458
     #BK_statsProviderClass: org.apache.bookkeeper.stats.prometheus.PrometheusMetricsProvider
     # use hostname as bookie id for StatefulSets deployment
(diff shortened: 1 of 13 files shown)


sijie:
> Maybe we can add @deprecated annotations to FlatLedgerManagerFactory class?

good idea

sijie:
@yzang : make `FlatLedgerManagerFactory` as `@deprecated`

deploy/kubernetes/gke/bookkeeper.yaml
```diff
@@ -30,8 +30,6 @@ data:
     BK_ledgerDirectories: "/bookkeeper/data/ledgers"
     BK_indexDirectories: "/bookkeeper/data/ledgers"
     BK_zkServers: zookeeper
-    # the default manager is flat, which is not good for supporting large number of ledgers
-    BK_ledgerManagerType: "hierarchical"
```
eolivelli: We could leave this line so that it is explicit this important configuration parameter

deploy/kubernetes/gke/bookkeeper.statefulset.yml
```diff
@@ -39,8 +39,6 @@ data:
     BK_ledgerDirectories: "/bookkeeper/data/ledgers"
     BK_indexDirectories: "/bookkeeper/data/ledgers"
     BK_zkServers: zookeeper
-    # the default manager is flat, which is not good for supporting large number of ledgers
-    BK_ledgerManagerType: "hierarchical"
```
eolivelli: We could leave this line so that it is explicit this important configuration parameter

sijie: `ledgerMangerType` is already deprecated long time ago. it doesn't make any sense to leave it here. we used the wrong parameter here so I don't see a reason to have it though. for people who is new to bookkeeper, he should be knowing less settings as possible.

PR #1116, opened 2018-02-02
Cleanup registration client interface

sijie:

This change is mainly to remove `zookeeper` reference from metadata interface. The existence of `Optional<ZooKeeper>` is to allow passing an external zookeeper client to bookkeeper, so bookkeeper can reuse that client instance. That is useful for services like pulsar broker, which they can only instantiate a zookeeper client and pass the zookeeper client around to construct bookkeeper client. However, this pollutes the registration client interface, change the method to `Optional<Object>` as an optional context object and let the implementation interpret it.


sijie:
@jiazhai since you did the interface abstraction, can you take a look to see if this work for you.

@merlimat since pulsar is the main user of external zookeeper instance, can you review to make sure it is okay from your perspective.

to be clear on this change, it would be ideal to remove that field completely. however since bookkeeper still supports passing external zookeeper client and most likely it won't be gone for any time soon, so I stick to just change it to `Optional<Object>` to remove zookeeper reference from metadata interfaces entirely. Let the implementation itself interpret this context object.

PR #1120, opened 2018-02-05
Cherry picks/i 161

jiazhai:
This is cherry-pick from yahoo repo of branch yahoo-4.3.

original change is:
https://github.com/yahoo/bookkeeper/commit/1ddd14a1
Only reorder read sequence if any bookie is not available


sijie:
@eolivelli - I believe this fixes both rackware and region-aware. region-ware is a hierarchical placement over rackwares. don't think we need to change it to be only rackware.

sijie:
@jiazhai I think this change changes the behavior, so the rackware placement policy tests might need to be updated. the test cases seem to be failing both java8 and java9.

jiazhai:
Thanks @eolivelli @sijie , Yes, this will involves both rackware and region-aware. Will update the test cases.

sijie:
@jiazhai the change looks good to me now. +1

sijie:
retest this please

PR #1121, opened 2018-02-05
[Merge Yahoo repo]: AsyncReadLastEntry should trigger callback with error when ledger is empty

jiazhai:
This is cherry-pick from yahoo repo of branch yahoo-4.3.

original change is:
https://github.com/yahoo/bookkeeper/commit/9560ab0f
AsyncReadLastEntry should trigger callback with error when ledger is empty


sijie:
@jiazhai I think this cherry-pick is not correct.

You need to cherry-pick changes in 372a99db49

```
diff --git a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
index 33e7af09b..885dda7ae 100644
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
@@ -460,13 +460,30 @@ public class LedgerHandle {
             return;
         }

-        asyncReadEntriesInternal(firstEntry, lastEntry, cb, ctx);
+        asyncReadEntriesInternal(firstEntry, lastEntry, cb, ctx, false);
     }

-    void asyncReadEntriesInternal(long firstEntry, long lastEntry, ReadCallback cb, Object ctx) {
+    /**
+     * Read the last entry in the ledger
+     *
+     * @param cb
+     *            object implementing read callback interface
+     * @param ctx
+     *            control object
+     */
+    public void asyncReadLastEntry(ReadCallback cb, Object ctx) {
+        long lastEntryId = getLastAddConfirmed();
+        asyncReadEntriesInternal(lastEntryId, lastEntryId, cb, ctx, true);
+    }
```

when you cherry-picked 372a99db49, please add a sync call of `asyncReadLastEntry`, and create an issue for adding this method into new ledger api.

Then cherry-pick 9560ab0f92

jiazhai:
Thank @sijie @eolivelli . cherry-pick 372a99db49 was handling in https://github.com/apache/bookkeeper/pull/898.  will add a sync call of `asyncReadLastEntry` here.

jiazhai:
will wait for #898.

sijie:
@jiazhai as we chatted offline, I don't think you need to wait for #898. you can cherry-pick the `asyncReadLastEntry` changes in #898 and combine it with this change here. because I think `asyncReadLastEntry` was removed in #898 since it was an irrelevant change.

jiazhai:
@sijie Thanks for the comments. changed the code. CI failure seems caused by other un-related issue.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -787,6 +787,24 @@ public void onFailure(Throwable cause) {
...
+    /*
+     * Read the last entry in the ledger
+     *
+     * @param cb
+     *            object implementing read callback interface
+     * @param ctx
+     *            control object
+     */
+    public void asyncReadLastEntry(ReadCallback cb, Object ctx) {
+        long lastEntryId = getLastAddConfirmed();
```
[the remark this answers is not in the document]
jvrao: +1 on Sijie's comment

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -787,6 +788,35 @@ public void onFailure(Throwable cause) {
...
+            asyncReadEntriesInternal(lastEntryId, lastEntryId, cb, ctx);
+        }
+    }
+
+    public org.apache.bookkeeper.client.api.LedgerEntry readLastEntry()
+        throws InterruptedException, ExecutionException {
+        long lastEntryId = getLastAddConfirmed();
+        if (lastEntryId < 0) {
+            // Ledger was empty, so there is no last entry to read
+            return null;
```
sijie: throw NoSuchEntryException to keep it consistent with asyncReadLastEntry.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadOnlyLedgerHandle.java
```diff
@@ -178,4 +180,19 @@ public String toString() {
     protected void initializeExplicitLacFlushPolicy() {
         explicitLacFlushPolicy = ExplicitLacFlushPolicy.VOID_EXPLICITLAC_FLUSH_POLICY;
     }
+
+    @Override
+    public void asyncReadLastEntry(ReadCallback cb, Object ctx) {
+        asyncReadLastConfirmed(new ReadLastConfirmedCallback() {
+            @Override
+            public void readLastConfirmedComplete(int rc, long lastConfirmed, Object ctx) {
+                if (lastConfirmed < 0) {
```
sijie: you need check `rc` first. if `rc` is not `OK`, then fail `ReadCallback`. Otherwise do the logic to read actual entry.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -787,6 +788,35 @@ public void onFailure(Throwable cause) {
...
+        if (lastEntryId < 0) {
+            // Ledger was empty, so there is no last entry to read
+            cb.readComplete(BKException.Code.NoSuchEntryException, this, null, ctx);
+        } else {
+            asyncReadEntriesInternal(lastEntryId, lastEntryId, cb, ctx);
+        }
+    }
+
+    public org.apache.bookkeeper.client.api.LedgerEntry readLastEntry()
+        throws InterruptedException, ExecutionException, BKException {
```
sijie: we should not throw "ExcuctionException" for sync methods. keep the behaviour same as other sync methods.

you can check https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java#L594 as an example.

you can use SyncReadCallback as other methods are using.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadOnlyLedgerHandle.java
```diff
@@ -178,4 +180,25 @@ public String toString() {
...
+                    if (lastConfirmed < 0) {
+                        // Ledger was empty, so there is no last entry to read
+                        cb.readComplete(BKException.Code.NoSuchEntryException, ReadOnlyLedgerHandle.this, null, ctx);
+                    } else {
+                        asyncReadEntriesInternal(lastConfirmed, lastConfirmed, cb, ctx);
+                    }
+                } else {
+                    LOG.error("ReadException in asyncReadLastEntry, ledgerId: {}, lac: {}, rc:{}",
+                        lastConfirmed, ledgerId, rc);
+                    cb.readComplete(BKException.Code.ReadException, ReadOnlyLedgerHandle.this, null, ctx);
```
sijie: cb.readCpomplete(rc, ...)

you can pass the rc through to the application.

sijie:
@jvrao can you review this PR again since you were involved in this PR?

sijie:
retest this please

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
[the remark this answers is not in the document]
jvrao: @sijie is this comment addressed?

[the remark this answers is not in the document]
sijie: Yes it is addressed

asfgit:

sijie:
all CI jobs are green except "Integration Tests". but this change is adding a new method, and doesn't change any existing logic. so we don't really need this to pass "Integration Tests". will ignore CI and merge it.

sijie:
IGNORE CI

Issue #1123, opened 2018-02-06
BP-29: Metadata API module


sijie:
**BP**

This is the master ticket for tracking BP-29 :

We have already abstracted all the metadata operations into interfaces. And all the bookkeeper implementations only reply on metadata interfaces, rather than depending on zookeeper. This proposal is to organize the metadata interfaces and its implementations in a separate module and make bookkeeper implementation only depends on metadata interfaces, not depends on zookeeper. This would a few benefits:

- It allows supporting different metadata storages, without bringing in dependencies of metadata store implementation directly into bookkeeper-server module. The development of different metadata storage can be done without interleaving with each other.
- It would define a clean module dependency between bookkeeper implementation and metadata api, and how bookkeeper load a different metadata implementation.

Proposal PR - #1117

PR #1127, opened 2018-02-07
BP-29 (task 1) : Introduce `metadataServiceUrl`

sijie:

- introduce a new setting `metadataServiceUrl` for replacing implementation specific settings `zkServers` and `zkLedgersRootPath`.
- introduce a metadata driver interface for managing all the metadata interface
- introduce a metadata driver manager for resolving `metadataServiceUrl` and load specific metadata service url based on service scheme. This
  would replace configuring multiple different manager classes using one single `metadataServiceUrl`.

Please NOTE: this change only introduce the new classes, but it doesn't change the client and bookie to use the new setting. It will be done in
a subsequent change.

Master Issue: #1123 

> ---
> quickly and easily:
>
>
>
> Otherwise:
> 
> 
> ---


bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/MetadataDrivers.java
```diff
@@ -0,0 +1,290 @@
...
+    }
+
+    private static void loadInitialClientDrivers() {
+        Set<String> driverList = Sets.newHashSet();
+
+        // add default zookeeper based driver
+        driverList.add(ZK_CLIENT_DRIVER_CLASS);
+
+        // load drivers from system property
+        String driversStr = System.getProperty(BK_METADATA_CLIENT_DRIVERS_PROPERTY);
```
eolivelli: I think it should be better to use standard Java Service Provider API. It is 'standard' and JPMS aware

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/MetadataDrivers.java
```diff
@@ -0,0 +1,290 @@
...
+        MetadataBookieDriverInfo(Class<? extends MetadataBookieDriver> driverClass) {
+            this.driverClass = driverClass;
+            this.driverClassName = driverClass.getName();
+        }
+
+    }
+
+    @Getter(AccessLevel.PACKAGE)
+    private static final ConcurrentMap<String, MetadataClientDriverInfo> clientDrivers;
+    @Getter(AccessLevel.PACKAGE)
```
eolivelli: I can't find any code to release resources/unload the drivers.
Can you please give me a pointer?

sijie: I think I forgot to add the unload interfaces. However since we also use this pattern on dlog namespace driver, I don't see a real need to unload drivers. 

Also This only caches the class information when the drivers are included in the classpath, it doesn't contain any real resources that we are using.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/MetadataDrivers.java
sijie: This pattern was following JDBC in java 7. I can change to use ServiceLoader as we are in java 8 (the way that JDBC is using in java 8).

sijie: actually I think jdbc can still load drivers from [system properties](http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/tip/src/share/classes/java/sql/DriverManager.java#l572). 

So I would keep this setting here and add the ServiceLoader support later once we have a clean module definition, unless there are other major concerns on current approach.

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
```diff
@@ -175,6 +179,58 @@ public void loadConf(CompositeConfiguration baseConf) {
...
+        if (null == serviceUri) {
+            // no service uri is defined, fallback to old settings
+            String ledgerManagerType;
+            Class<? extends LedgerManagerFactory> factoryClass = getLedgerManagerFactoryClass();
+            if (factoryClass == null) {
+                ledgerManagerType = HierarchicalLedgerManagerFactory.NAME;
+            } else {
+                if (!AbstractZkLedgerManagerFactory.class.isAssignableFrom(factoryClass)) {
+                    // this is a non-zk implementation
+                    throw new UnsupportedOperationException("metadata service uri is not supported for "
```
jvrao: This is a no service URI case. But the log messages reads so..

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/MetadataClientDriver.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.concurrent.ScheduledExecutorService;
+import org.apache.bookkeeper.common.annotation.InterfaceAudience.LimitedPrivate;
+import org.apache.bookkeeper.common.annotation.InterfaceStability.Evolving;
+import org.apache.bookkeeper.conf.ClientConfiguration;
+import org.apache.bookkeeper.discover.RegistrationClient;
+import org.apache.bookkeeper.meta.exceptions.MetadataException;
+import org.apache.bookkeeper.stats.StatsLogger;
+
+/**
+ * Driver to manage all the metadata managers required by a bookkeeper client.
```
jvrao: Lets remove 'client' at the end. Bk Servers may need access too.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/MetadataDrivers.java
eolivelli: Okay for the system property for now

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/MetadataDrivers.java
eolivelli: I would prefer to have a way to cleanup things, think about webapps, in this case the webapplication need a way to cleanup things.
Or maybe we can have an *unregister* method to unload the driver explicitly

sijie: I can add an unregister method. But my point is it is never used, because in current approach, we only keep the mapping between scheme and the class. We don’t instantiate objects until resolving service uri. So there is really no place for you to unregister a “Class”. This is different from web apps. 

Franckly speaking, you only will have one implementation in your classpath and only use that implemention all the time. There is really no place or point for you to unregister.

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
sijie: the idea was: in the case no service uri is defined, if there is a customized factory class implemented, we should fail it and force the users to define a metadata service uri for this factory implementation. so from this perspective, it is a unsupported operation exception.

but if you have a better idea on what exception we should throw here, please help suggest one.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/MetadataDrivers.java
```diff
@@ -0,0 +1,290 @@
...
+ *
+ */
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+@Slf4j
+public final class MetadataDrivers {
+
+    static final String ZK_CLIENT_DRIVER_CLASS = "org.apache.bookkeeper.meta.zk.ZKMetadataClientDriver";
+    static final String ZK_BOOKIE_DRIVER_CLASS = "org.apache.bookkeeper.meta.zk.ZKMetadataBookieDriver";
+    static final String BK_METADATA_CLIENT_DRIVERS_PROPERTY = "bookkeeper.metadata.client.drivers";
+    static final String BK_METADATA_BOOKIE_DRIVERS_PROPERTY = "bookkeeper.metadata.bookie.drivers";
```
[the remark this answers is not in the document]
sijie: what do you mean "etcd" here?

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/exceptions/package-info.java
```diff
@@ -0,0 +1,23 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/**
+ * Exceptions for metadata service.
+ */
+package org.apache.bookkeeper.meta.exceptions;
```
athanatos: Missing newline

sijie: I don't think we have a check style enforcing 'newline' though. is this a concern?

sijie:
@eolivelli @jvrao  I already addressed your comments. please review this again.

sijie:
ping @jvrao @eolivelli , please review so that I can move forward with other changes.

sijie:
ping @jvrao

eolivelli:
retest this please

sijie:
retest this please

asfgit:

eolivelli:
@sijie do you thibj we can merge this one or better to wait for the second part?

sijie:
@eolivelli I will merge it after CI passed. I need this for subsequent changes.

sijie:
retest this please

sijie:
merged to latest master (so integration tests can actually run)

PR #1129, opened 2018-02-07
Upgrade JaCoCo and Upgrade Coveralls.io configuration

eolivelli:
- Upgrade JaCoCo to version 0.8.0 which supports Java 9
- Update Coveralls.io configuration in order to handle NAR-plugin files
- Update surefire configuration in order to not fail the build while using -Pcode-coverage

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -389,6 +389,9 @@
                 <groupId>org.eluder.coveralls</groupId>
                 <artifactId>coveralls-maven-plugin</artifactId>
                 <version>4.3.0</version>
+                <configuration>
+                    <sourceDirectories>circe-checksum/target/nar/nar-generated</sourceDirectories>
+                </configuration>
               </plugin>
               <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
@@ -400,12 +403,14 @@
 	          <redirectTestOutputToFile>true</redirectTestOutputToFile>
 	          <reuseForks>false</reuseForks>
 	          <forkedProcessTimeoutInSeconds>1800</forkedProcessTimeoutInSeconds>
+                  <!-- we want build to complete even in case of failures -->
+                  <testFailureIgnore>true</testFailureIgnore>
                 </configuration>
              </plugin>
              <plugin>
                <groupId>org.jacoco</groupId>
                <artifactId>jacoco-maven-plugin</artifactId>
-               <version>0.7.9</version>
+               <version>0.8.0</version>
                <executions>
                  <execution>
                    <goals>


pom.xml
```diff
@@ -389,6 +389,9 @@
                 <groupId>org.eluder.coveralls</groupId>
                 <artifactId>coveralls-maven-plugin</artifactId>
                 <version>4.3.0</version>
+                <configuration>
+                    <sourceDirectories>circe-checksum/target/nar/nar-generated</sourceDirectories>
```
sijie: any idea why we need to cover `nar-generated` here?

eolivelli: Because coveralls.io plugin can't find the sources of generated classes.
I am wondering if would it be better to exclude all generated sources from code coverage.
I forgot to mark this PR as WIP

sijie: Understand what sourcediectories mean. My question is why do we need to test coverage for generated code. I don’t see a reason to cover generated code, unless I missed anything.

eolivelli: Sorry, I have misunderstood your comment.
I am working on skipping all generated code, it makes no sense to check it

eolivelli:
@sijie updated the patch, now we are excluding NarSystem from the report

it seems that protobuf stuff is already excluded automatically (I cannot find those classes on report)

see this one (it does not report NarSystem any more)
https://coveralls.io/builds/15432041

eolivelli:
There real next step from this point it to understand the report and identity code paths not covered and/or understand way they are not reported as "covered" even there are tests.

I will go on, there is no need to merge this PR at the moment

asfgit:

eolivelli:
This is current error

> 2018-02-16T14:48:13.769 [INFO] --- jacoco-maven-plugin:0.8.0:report-aggregate (default) @ codecoverage ---
> 2018-02-16T14:48:13.997 [INFO] Loading execution data file /home/jenkins/jenkins-slave/workspace/bookkeeper-code-coverage-wip/bookkeeper-server/target/jacoco.exec
> 2018-02-16T14:48:14.892 [INFO] Analyzed bundle 'bookkeeper-server' with 935 classes
> 2018-02-16T14:48:14.893 [WARNING] Classes in bundle 'bookkeeper-server' do no match with execution data. For report generation the same class files must be used as at runtime.
> 2018-02-16T14:48:14.893 [WARNING] Execution data for class org/apache/bookkeeper/server/service/BookieService does not match.

eolivelli:
Jenkins JOB:
https://builds.apache.org/job/bookkeeper-code-coverage-wip/

eolivelli:
This is the latest result
https://coveralls.io/jobs/34016730

@sijie this patch is good to go for me

PR #1130, opened 2018-02-07
handle zookeeper session expire in zk ledger manager

sijie:

cherry-pick twitter/bookkeeper@dfcda5cc2efdc03db99fe126499f8e3347f50484

This includes a test case to test `AbstractZkLedgerManager`, including

    1) create/delete/read/write ledger metadata
    2) register/unregister listeners
    3) handling various watched events

This also fixes a couple of cases in the `AbstractZkLedgerManager` what is caught with test case.

> ---
> quickly and easily:
>
>
>
> Otherwise:
> 
> 
> ---


sijie:
I think this change was actually in master when we contributed back ensemble watcher. but it was commented out on waiting the retryable zookeeper client. somehow that change was removed by cleaning TODO items.

I am cherry-pciking this change back. I am marked as "WIP", as I would like to write a test case to test it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractZkLedgerManager.java
```diff
@@ -179,8 +179,16 @@ protected AbstractZkLedgerManager(AbstractConfiguration conf, ZooKeeper zk) {
 
     @Override
     public void process(WatchedEvent event) {
-        LOG.info("Received watched event {} from zookeeper based ledger manager.", event);
+        LOG.debug("Received watched event {} from zookeeper based ledger manager.", event);
         if (Event.EventType.None == event.getType()) {
+            if (Event.KeeperState.Expired == event.getState()) {
+                LOG.info("ZooKeeper client expired on ledger manager.");
```
eolivelli: Maybe it would be useful to log the session id, with the prev code we had the event, now       the event is logged at debug level. Don't know if event.toString contained the session id..
Anyway while analyzing zk session expiration events it is useful to have the id

sijie: Event.toString doesn't contain any session information.

It is also a bit tricky to log session id as well. because the underneath retry logic in the zookeeper client we are using. at the time the watcher callback was trigger, the underneath client might already reconnect and a new session id is available. if we try to log the session id, it might be actually misleading. 

If we want to achieve accurate logging with session id, we have to refactor the way how we submit callback, which I think it is an excessive change here.

eolivelli: Clear. Okay

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
```diff
@@ -52,6 +53,7 @@
  *
  * <p>It provides parsing and serialization methods of such metadata.
  */
+@EqualsAndHashCode
```
jvrao: What is the significance of adding this annotation as part of this change?

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractZkLedgerManager.java
```diff
@@ -409,7 +426,7 @@ public void processResult(int rc, String path, Object ctx, byte[] data, Stat sta
     public void writeLedgerMetadata(final long ledgerId, final LedgerMetadata metadata,
                                     final GenericCallback<Void> cb) {
         Version v = metadata.getVersion();
-        if (Version.NEW == v || !(v instanceof LongVersion)) {
+        if (!(v instanceof LongVersion)) {
```
jvrao: Why are we removing the first check? the (v instanceof LongVersion) just checks the instance, not the actual compare right?

sijie: this is a bug. if you take a look at this line (https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractZkLedgerManager.java#L416), it directly casts v to LongVersion. so the passed-in `metadata.getVersion()` has to be `LongVersion`.

sijie:
@jvrao I have addressed your comment, please review it again.

sijie:
ping @jvrao

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractZkLedgerManager.java
eolivelli: @sijie do you mean that it is now possible for *v* to be '==' to Version.NEW because it cannot be of the same class so that the test is not useful?

sijie: @eolivelli I mean v can not be `Version.New` because line 416 expects a LongVersion. also the ledger manager api have separate api for CreateLedgerMetadata and WriteLedgerMetadata. so we don't expect Version.New and Version.ANY on `WriteLedgerMetadata`.

asfgit:

sijie:
retest this please

sijie:
retest this please

sijie:
merged latest master

sijie:
integration tests passed BC tests but timed out on smoking test. it seems that the smoke tests can not connect to zookeeper docker image. don't think it is related to this PR, since all precommits job passed and BC tests passed. 

so going to ignore integration tests CI result to merge this change first.

sijie:
IGNORE CI

sijie:
can't easily cherry-pick this change to branch 4.6 because it requires the mock executor for testing. will cherry-pick this later.

PR #1131, opened 2018-02-09
Split all versions images into released and latest

ivankelly:
The released tarballs rarely change, while the latest changes every
build. This patch splits the all version image into a base images with
the released tarballs, and a image which includes all released and the
very latest tarball.

This allows the released image to be cached/pushed to dockerhub, which
will in turn allow us to speed up our CI builds that use the all
version image.


tests/docker-all-versions-image/all-released-image/Dockerfile
```diff
@@ -17,16 +17,16 @@
 # under the License.
 #
 
-FROM centos:7
+FROM openjdk:8-jdk
```
sijie: any reason what it was changed from centos7 to openjdk?

tests/docker-all-versions-image/latest-image/pom.xml
```diff
@@ -0,0 +1,111 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <groupId>org.apache.bookkeeper.tests</groupId>
+    <artifactId>all-versions-image</artifactId>
+    <version>4.7.0-SNAPSHOT</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>org.apache.bookkeeper.tests</groupId>
+  <artifactId>docker-latest-version-image</artifactId>
```
sijie: this seems to be a bit confusing to me. The name here sounds like an image include the latest version, which is actually not. This image is actually "all-versions-image". so I would suggest:

- name the parent module as "test-docker-images"
- name the module that builds the image of all released versions : "docker-all-released-versions-image"
- name the module that builds the image of all versions : "docker-all-versions-image"

tests/docker-all-versions-image/all-released-image/Dockerfile
ivankelly: to avoid having to pull epel (which we need for supervisor). openjdk is debian based, as has supervisor in the base repos.

tests/docker-all-versions-image/latest-image/pom.xml
ivankelly: sure, will change

ivankelly: the docker- is redundant, in the submodules, i'll leave it off

ivankelly:
retest this please

ivankelly:
retest this please

PR #1132, opened 2018-02-09
Fix shutdown race which left ZK session open

ivankelly:
There was a race when shutting down a bookie, where both the main
thread and the shutdown hook thread would try to close the bookie
service. This would result in them racing to set the lifecycle state
and neither would end up cleaning up properly. Specifically, the
starter latch would be counted down, so the main thread would start to
close, then the shutdown hook would run, try to close, and get an
exception when it tried to change the state. Once the all shutdown
hooks end, the process exits, even if main hasn't completed. This
leaves the zookeeper session open.

I've removed the close from main(). The shutdown hook always runs
on a graceful shutdown, so that's the place to do the cleanup.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/server/Main.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/server/Main.java
@@ -219,7 +219,6 @@ public class Main {
             // the server is interrupted
             log.info("Bookie server is interrupted. Exiting ...");
         }
-        server.close();
         return ExitCode.OK;
     }
 


ivankelly:
retest this please

sijie:
If I understand this change right, the fix can be just a simple 2-lines change, no?

in ComponentShutdownHook, move `aliveLatch.countDown()` to be after `component.close` and add it in the catch block.


Unless I missed anything here.

ivankelly:
@sijie yes, that change would also fix the immediate issue. It could also be fixed in a number of one-line ways.
- synchronization on #close() method
- Simply removing the #close() in #main()

I can pare down the patch to this if needed, but right now it's trying to address some fundamental issues with the lifecycle stuff, i.e. that it's taking action based on perceived state, and updating that state, without any synchronization between the threads.

sijie:
@ivankelly just fyi the lifecycle stuff wasn't designed for concurrent access. it assumes that only one thread is managing the lifecycle state. this PR changes the behavior of `beforeStart`, `beforeStop` and `beforeStop`, which was supposed to run before actual state changes. so I would suggest just adding synchronization to fix race condition, which doesn't change the original behavior and it is enough. because I don't think we need lifecycle management to have handle many concurrent accesses.

ivankelly:
If lifecycle changes are coming from multiple threads, it should be capable of concurrent access. Otherwise it's broken and should be removed. 

The synchronized #close isn't a good fix. The removal of close() from main is better. I can strip the patch down to this to get things moving, and push the other changes separately.

sijie:
please strip down the change to only fix the race condition first.

ivankelly:
Merging

PR #1133, opened 2018-02-09
Integration smoke test

ivankelly:
Most basic integration test we can have.
- Boots a cluster
- Writes a ledger
- Reads back the ledger
- Shutdown the cluster

If this test fails, something is very wrong.

### DIFF
--- a/tests/integration/smoke/pom.xml
+++ b/tests/integration/smoke/pom.xml
@@ -15,36 +15,40 @@
    See the License for the specific language governing permissions and
    limitations under the License.
 -->
-<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
-         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
-  <packaging>pom</packaging>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="
+  http://maven.apache.org/POM/4.0.0
+  http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
   <parent>
-    <groupId>org.apache.bookkeeper</groupId>
-    <artifactId>bookkeeper</artifactId>
+    <groupId>org.apache.bookkeeper.tests.integration</groupId>
+    <artifactId>tests-parent</artifactId>
     <version>4.7.0-SNAPSHOT</version>
   </parent>
-  <groupId>org.apache.bookkeeper.tests</groupId>
-  <artifactId>tests-parent</artifactId>
-  <name>Apache BookKeeper :: Tests</name>
-  <modules>
-    <module>shaded</module>
-    <module>docker-all-versions-image</module>
-    <module>integration-tests-base</module>
-    <module>integration-tests-base-groovy</module>
-    <module>integration-tests-utils</module>
-    <module>integration-tests-topologies</module>
-    <module>backward-compat</module>
-    <module>integration</module>
-  </modules>
+
+  <groupId>org.apache.bookkeeper.tests.integration</groupId>
+  <artifactId>smoke</artifactId>
+  <packaging>jar</packaging>
+  <name>Apache BookKeeper :: Tests :: Backward Compatibility :: Smoke test</name>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.apache.bookkeeper</groupId>
+      <artifactId>bookkeeper-server</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+  </dependencies>
+
   <build>
     <plugins>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-deploy-plugin</artifactId>
-        <version>${maven-deploy-plugin.version}</version>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <version>${maven-surefire-plugin.version}</version>
         <configuration>
-          <skip>true</skip>
+          <!-- smoke test should never flake //-->
+          <rerunFailingTestsCount>0</rerunFailingTestsCount>
         </configuration>
       </plugin>
     </plugins>
--- a/tests/integration/smoke/src/test/java/org/apache/bookkeeper/tests/integration/TestSmoke.java
+++ b/tests/integration/smoke/src/test/java/org/apache/bookkeeper/tests/integration/TestSmoke.java
@@ -1,0 +1,93 @@
+/*
+* Licensed to the Apache Software Foundation (ASF) under one
+* or more contributor license agreements.  See the NOTICE file
+* distributed with this work for additional information
+* regarding copyright ownership.  The ASF licenses this file
+* to you under the Apache License, Version 2.0 (the
+* "License"); you may not use this file except in compliance
+* with the License.  You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+package org.apache.bookkeeper.tests.integration;
+
+import com.github.dockerjava.api.DockerClient;
+
+import java.util.Enumeration;
+
+import org.apache.bookkeeper.client.BookKeeper;
+import org.apache.bookkeeper.client.LedgerEntry;
+import org.apache.bookkeeper.client.LedgerHandle;
+import org.apache.bookkeeper.tests.BookKeeperClusterUtils;
+
+import org.jboss.arquillian.junit.Arquillian;
+import org.jboss.arquillian.test.api.ArquillianResource;
+
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@RunWith(Arquillian.class)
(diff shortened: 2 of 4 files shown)


tests/integration/smoke/src/test/java/org/apache/bookkeeper/tests/integration/TestSmoke.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    @Test
+    public void testBootWriteReadShutdown() throws Exception {
+        Assert.assertTrue(BookKeeperClusterUtils.startAllBookiesWithVersion(docker, currentVersion));
+
+        String zookeeper = BookKeeperClusterUtils.zookeeperConnectString(docker);
+        BookKeeper bk = new BookKeeper(zookeeper);
+        long ledgerId;
+        try (LedgerHandle writelh = bk.createLedger(BookKeeper.DigestType.CRC32, PASSWD)) {
```
eolivelli: We can use new API here

tests/integration/smoke/src/test/java/org/apache/bookkeeper/tests/integration/TestSmoke.java
```diff
@@ -0,0 +1,93 @@
...
+            ledgerId = writelh.getId();
+            for (int i = 0; i < 100; i++) {
+                writelh.addEntry(("entry-" + i).getBytes());
+            }
+        }
+
+        Thread.sleep(10000);
```
eolivelli: Why this sleep?

tests/integration/smoke/src/test/java/org/apache/bookkeeper/tests/integration/TestSmoke.java
```diff
@@ -0,0 +1,93 @@
...
+                Assert.assertEquals(readBack, "entry-" + i++);
+            }
+            Assert.assertEquals(i, 100);
+        }
+
+        bk.close();
+
+        if (!BookKeeperClusterUtils.stopAllBookies(docker)) {
+            Thread.sleep(20000);
+            Assert.fail("A bookie failed to stop");
```
eolivelli: This sleep is needed?

tests/integration/smoke/pom.xml
```diff
@@ -0,0 +1,56 @@
...
+  http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.apache.bookkeeper.tests.integration</groupId>
+    <artifactId>tests-parent</artifactId>
+    <version>4.7.0-SNAPSHOT</version>
+  </parent>
+
+  <groupId>org.apache.bookkeeper.tests.integration</groupId>
+  <artifactId>smoke</artifactId>
```
eolivelli: Maybe it is better to have this id equal to the name of the directory?

ivankelly: ? the directory is smoke

tests/integration/smoke/src/test/java/org/apache/bookkeeper/tests/integration/TestSmoke.java
ivankelly: nope, will remove
was only there for debugging

tests/integration/smoke/src/test/java/org/apache/bookkeeper/tests/integration/TestSmoke.java
ivankelly: likewise

tests/integration/smoke/pom.xml
eolivelli: sorry, I was looking at the wrong pom

ivankelly:
regarding the api, I'd like to see some issues resolved with it before I start using it in tests. It's something I'm planning to come back to, but I wanted to get the BC test stuff out of the way first.

eolivelli:
fine for me, it makes sense to use the new API once is will be considered stable or 'beta'

sijie:
> it makes sense to use the new API once is will be considered stable or 'beta'

I think it is the reverse, we should make tests to use API before considering it as stable or beta.

sijie:
> regarding the api, I'd like to see some issues resolved with it before I start using it in tests.

what are the issues blocking using it in tests.

tests/integration/smoke/src/test/java/org/apache/bookkeeper/tests/integration/TestSmoke.java
```diff
@@ -0,0 +1,89 @@
...
+            BookKeeperClusterUtils.formatAllBookies(docker, currentVersion);
+        }
+    }
+
+    @Test
+    public void testBootWriteReadShutdown() throws Exception {
+        Assert.assertTrue(BookKeeperClusterUtils.startAllBookiesWithVersion(docker, currentVersion));
+
+        String zookeeper = BookKeeperClusterUtils.zookeeperConnectString(docker);
+        BookKeeper bk = new BookKeeper(zookeeper);
```
sijie: I think we should be using bookkeeper builder rather than calling the bookkeeper constructor with zookeeper string directly. (just to be clear, the old api also has builders for constructing bookkeeper instance).

tests/integration/smoke/src/test/java/org/apache/bookkeeper/tests/integration/TestSmoke.java
```diff
@@ -0,0 +1,89 @@
...
+            Enumeration<LedgerEntry> entries = readlh.readEntries(0, lac);
+            while (entries.hasMoreElements()) {
+                LedgerEntry e = entries.nextElement();
+                String readBack = new String(e.getEntry());
+                Assert.assertEquals(readBack, "entry-" + i++);
+            }
+            Assert.assertEquals(i, 100);
+        }
+
+        bk.close();
```
sijie: nit: put bk in a try-block. or use cleanup annotation.

tests/integration/smoke/pom.xml
```diff
@@ -0,0 +1,56 @@
...
+  <packaging>jar</packaging>
+  <name>Apache BookKeeper :: Tests :: Backward Compatibility :: Smoke test</name>
```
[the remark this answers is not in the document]
ivankelly: done

tests/integration/smoke/src/test/java/org/apache/bookkeeper/tests/integration/TestSmoke.java
ivankelly: That would make the line 
```
try (BookKeeper bk = BookKeeper.forConfig(new Configuration().setZkServers(zookeeper)).build) {
```
rather than
```
try (BookKeeper bk = new BookKeeper(zookeeper)) {
```

It's hard to argue that the builder is clearer. If it was something like
```
try (BookKeeper bk = BookKeeper.newBuilder().withZooKeeper(zookeeper).build())
```
then sure, but currently we have to pass in the connect string though a configuration, which kinda sucks.

tests/integration/smoke/src/test/java/org/apache/bookkeeper/tests/integration/TestSmoke.java
ivankelly: done

tests/integration/smoke/src/test/java/org/apache/bookkeeper/tests/integration/TestSmoke.java
ivankelly: and this is also true for the new api.

ivankelly:
> what are the issues blocking using it in tests.

@sijie Nothing blocking persay, but I want the asyncClose issue revolved before I start using them. I think what you suggest before about changing the methods to have sync and async methods like pulsar is the best solution to this, but this would require changing all tests that use them after the change.

Also, way we return CompletableFuture<> from everything is awkward to use in tests. You have to put .get() at the end of all calls, and can't directly catch specific exceptions. I'm aware I pushed for this at the time, having tried to use the apis a bit left me wiser.

Anyhow, as I said to @eolivelli, I'm planning to come back to the API stuff, but I want the BC test stuff out of my head first.

tests/integration/smoke/src/test/java/org/apache/bookkeeper/tests/integration/TestSmoke.java
sijie: the point here is not about whether the builder is clear or not, for sure it has cons on builder method. the point here is more about we should not use any "deprecated" method. more even, we are getting hard to get rid of direct zookeeper reference, so I would like to see we avoid using direct zookeeper constructor if possible.

sijie:
> You have to put .get() at the end of all calls, and can't directly catch specific exceptions.

you can use `FutureUtils.result(...)` which already unwrap the exception from ExecutionException, so you can catch specific exceptions.

ivankelly:
>  can use FutureUtils.result(...) which already unwrap the exception from ExecutionException, so you can catch specific exceptions.

Wrapping all calls in FutureUtils.result is ugly as hell

tests/integration/smoke/src/test/java/org/apache/bookkeeper/tests/integration/TestSmoke.java
ivankelly: Our api shouldn't be evolving in a way that simple stuff becomes less clear. I'll happily change to builder if and when the builder is as easy to use as the direct zookeeper constructor, but this isn't the case now.

sijie:
> Wrapping all calls in FutureUtils.result is ugly as hell

well. I think "ugly" is just a developer preference, the sync methods anyway need to use "get" or "result(...)". at twitter, we only expose async method and provide a util class `Await` to wait synchronously.

so the question here is more about whether we want to expose sync method for every method we are adding, or just providing a util class or let the application use CompletableFuture.get to wait synchronously.

ivankelly:
> so the question here is more about whether we want to expose sync method for every method we are adding, or just providing a util class or let the application use CompletableFuture.get to wait synchronously.

I think yes, but this PR isn't the forum for this. As I said, I'll bring all this up again once the compat stuff is finished.

sijie:
sure

sijie:
retest this please

ivankelly:
@sijie ok to merge this?

PR #1135, opened 2018-02-09
Specify repo in MavenClassLoader

ivankelly:
This is required for testing yahoo's custom bookkeeper client, as it
doesn't reside in maven central.

### DIFF
--- a/tests/integration-tests-utils/src/test/java/org/apache/bookkeeper/tests/MavenClassLoaderTest.java
+++ b/tests/integration-tests-utils/src/test/java/org/apache/bookkeeper/tests/MavenClassLoaderTest.java
@@ -39,9 +39,4 @@ public class MavenClassLoaderTest {
                                    org.apache.zookeeper.KeeperException.NoNodeException.class));
     }
 
-    @Test
-    public void testLoadSpecifyingMavenRepo() throws Exception {
-        MavenClassLoader.forArtifact("https://raw.githubusercontent.com/yahoo/bookkeeper/mvn-repo",
-                                     "org.apache.bookkeeper:bookkeeper-server:4.3.1.85-yahoo");
-    }
 }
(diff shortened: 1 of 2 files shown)


tests/integration-tests-utils/src/test/java/org/apache/bookkeeper/tests/MavenClassLoaderTest.java
```diff
@@ -39,4 +39,9 @@ public void testNoZooKeeperInterference() throws Exception {
                                    org.apache.zookeeper.KeeperException.NoNodeException.class));
     }
 
+    @Test
+    public void testLoadSpecifyingMavenRepo() throws Exception {
```
eolivelli: Honestly I don't like very much to depend on remote resources.
The best would be for me to start an http server, with wiremo and then use it

ivankelly: What's the issue with remote resources for a test that'll only normally run in CI? If github is down, the CI job will fail much earlier.

sijie: I am not sure what is the purpose of this test? It seems that we are testing the functionality that provides by maven resolver. didn't see much value to do this though.

eolivelli: What about dropping the testcase  ?

ivankelly: @eolivelli actually I misunderstood your original point. Yes, unit tests shouldn't have remote resource dependencies. I thought you meant we shouldn't use this in the integration tests.
@sijie I wrote the test case before the code to validate that what I assumed the library would do is actually what the library did.

I think the test has value, but given the hassle of having to boot a http server and mock up a maven repo, just to test a test utility, I'm just going to remove the test.

ivankelly:
retest this please

PR #1137, opened 2018-02-09
Fix test regression due to change in default ledger manager

ivankelly:
TestBenchmark#testReadThroughputLatency() regressed due to the change
in 3ddc5db, where the default ledger manager was changed from flat to
hierarchical.

The fix is to use flat for this test. I've also cleaned up the test a
little, and added some code to the benchmark to make the test run
faster.

### DIFF
--- a/bookkeeper-benchmark/src/test/java/org/apache/bookkeeper/benchmark/TestBenchmark.java
+++ b/bookkeeper-benchmark/src/test/java/org/apache/bookkeeper/benchmark/TestBenchmark.java
@@ -19,13 +19,6 @@
  */
 package org.apache.bookkeeper.benchmark;
 
-import com.google.common.util.concurrent.ThreadFactoryBuilder;
-
-import java.util.concurrent.Executors;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Future;
-import java.util.concurrent.TimeUnit;
-
 import org.apache.bookkeeper.client.BookKeeper;
 import org.apache.bookkeeper.client.LedgerHandle;
 import org.apache.bookkeeper.net.BookieSocketAddress;
@@ -81,20 +74,27 @@ public class TestBenchmark extends BookKeeperClusterTestCase {
 
     @Test
     public void testReadThroughputLatency() throws Exception {
-        ExecutorService executor = Executors.newSingleThreadExecutor(
-                new ThreadFactoryBuilder().setNameFormat("read-throughput-latency-test").build());
-        Future<Integer> f = executor.submit(() -> {
-                BenchReadThroughputLatency.main(new String[] {
-                        "--zookeeper", zkUtil.getZooKeeperConnectString(),
-                        "--listen", "10"});
-                return 0;
-            });
+        final AtomicBoolean threwException = new AtomicBoolean(false);
+        Thread t = new Thread() {
+                public void run() {
+                    try {
+                        BenchReadThroughputLatency.main(new String[] {
+                                "--zookeeper", zkUtil.getZooKeeperConnectString(),
+                                "--listen", "10"});
+                    } catch (Throwable t) {
+                        LOG.error("Error reading", t);
+                        threwException.set(true);
+                    }
+                }
+            };
+        t.start();
 
-        Thread.sleep(2000);
+        Thread.sleep(10000);
         byte data[] = new byte[1024];
         Arrays.fill(data, (byte)'x');
 
         long lastLedgerId = 0;
+        Assert.assertTrue("Thread should be running", t.isAlive());
         for (int i = 0; i < 10; i++) {
             BookKeeper bk = new BookKeeper(zkUtil.getZooKeeperConnectString());
             LedgerHandle lh = bk.createLedger(BookKeeper.DigestType.CRC32, "benchPasswd".getBytes());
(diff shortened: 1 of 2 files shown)


bookkeeper-benchmark/src/test/java/org/apache/bookkeeper/benchmark/TestBenchmark.java
```diff
@@ -66,27 +81,20 @@ public void testBookie() throws Exception {
...
+                new ThreadFactoryBuilder().setNameFormat("read-throughput-latency-test").build());
+        Future<Integer> f = executor.submit(() -> {
+                BenchReadThroughputLatency.main(new String[] {
+                        "--zookeeper", zkUtil.getZooKeeperConnectString(),
+                        "--listen", "10"});
+                return 0;
+            });
 
-        Thread.sleep(10000);
+        Thread.sleep(2000);
```
sijie: since you are changing here, can we improve it to get rid of the sleep logic here?

bookkeeper-benchmark/src/test/java/org/apache/bookkeeper/benchmark/TestBenchmark.java
```diff
@@ -100,14 +108,7 @@ public void run() {
...
         }
-        for (int i = 0; i < 60; i++) {
-            if (!t.isAlive()) {
-                break;
-            }
-            Thread.sleep(100);
-        }
-
-        Assert.assertFalse("Thread should be finished", t.isAlive());
+        Assert.assertEquals(Integer.valueOf(0), f.get(30, TimeUnit.SECONDS));
```
sijie: can we get rid of timed wait here? we should just rely on the unit test framework to timeout the test case.

bookkeeper-benchmark/src/test/java/org/apache/bookkeeper/benchmark/TestBenchmark.java
ivankelly: Would be very hard to without changing the ReadThroughputLatency benchmark completely. That's out of scope for the fix. I can remove all changes that are unrelated to the core problem, but this fix needs to go in ASAP. Postcommit[1] has been broken for 9 days now.

[1] https://builds.apache.org/view/B/view/BookKeeper/job/bookkeeper_postcommit_master_java8/

bookkeeper-benchmark/src/test/java/org/apache/bookkeeper/benchmark/TestBenchmark.java
ivankelly: changed.

bookkeeper-benchmark/src/test/java/org/apache/bookkeeper/benchmark/TestBenchmark.java
sijie: personally I would suggest fixing the problem only rather than coupling with additional refactor. because it seems that this refactor doesn't really address any problem, which it seems to be unnecessary for fixing the CI.

ivankelly: The refactor reduces the time we spend waiting. Will remove it though.

ivankelly:
@sijie can you look at this again?

bookkeeper-benchmark/src/test/java/org/apache/bookkeeper/benchmark/TestBenchmark.java
```diff
@@ -38,6 +39,13 @@ public TestBenchmark() {
         super(5);
     }
 
+    @Before
+    public void setUp() throws Exception {
+        baseConf.setLedgerManagerFactoryClassName("org.apache.bookkeeper.meta.FlatLedgerManagerFactory");
```
eolivelli: Why are you using a string and not an explicit reference to the class?
Having a strong compile time reference will ease refactoring and code analysis

ivankelly: Because, if you reference the class,
```
[WARNING] /home/ivan/src/bookkeeper/checkouts/bench-flake/bookkeeper-benchmark/src/test/java/org/apache/bookkeeper/benchmark/TestBenchmark.java:[44,73] org.apache.bookkeeper.meta.FlatLedgerManagerFactory in org.apache.bookkeeper.meta has been deprecated
```

FlatLedgerManagerFactory is never going to move. That in itself would be a BC break.

eolivelli: Okay

sijie: just fyi, you can use full qualified import and `suppresswarnings("deprecation")` on the setup method.

ivankelly: I'm well aware of suppressWarnings, but it was a small nit, that didn't add anything, the patch was already green with CI, was otherwise ready to go, just waiting on approvals and I wanted to get this change in asap, as master CI has been broken for 7 days now, and it's blocking other patches.

ivankelly:
merging this now.

bookkeeper-benchmark/src/test/java/org/apache/bookkeeper/benchmark/TestBenchmark.java
sijie: yes, I am not blocking this PR. the comment was just "FYI"

PR #1141, opened 2018-02-11
Replace DoubleByteBuf with ByteBufList

merlimat:
This PR aims at a better approach to fix the issue reported in #1108 

> Starting from Netty-4.1.13 some internal behavior has changed and that has introduced a performance penalty when using the DoubleByteBuf.
>
> The problem resides in the fact that netty falls back to calling nioBuffers() and that is creating an array of unpooled direct ByteBuffers. That is pretty heavy and the JVM is taking 1 to 3 seconds GC pause to reclaim them.

Instead of using `CompositeByteBuf` which has the downside of needing multiple object allocations per each usage (~7-8), we should have a way to aggregate multiple buffers into a single "logical" frame with no additional overhead.

This PR introduces a `ByteBufList` class which is just a holder of a list of buffers, without imitating the `ByteBuf` API. You can pass a `ByteBufList` to the channel and have the encoder take care of the multiple writes. 

Instances of ByteBufList are ref-counted and released to a pool after the write is completed. There are no object allocations or memory copies involved.


jvrao:
@dlg99  FYI.

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/ByteBufList.java
```diff
@@ -0,0 +1,304 @@
...
+
+    /**
+     * Append a {@link ByteBuf} at the end of this {@link ByteBufList}.
+     */
+    public void add(ByteBuf buf) {
+        buffers.add(buf);
+    }
+
+    /**
+     * Prepend a {@link ByteBuf} at the end of this {@link ByteBufList}.
```
jvrao: s/end of/head of

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
```diff
@@ -560,8 +562,7 @@ void addEntry(final long ledgerId, byte[] masterKey, final long entryId, ByteBuf
                     .setOperation(OperationType.ADD_ENTRY)
                     .setTxnId(txnId);
 
-            byte[] toSendArray = new byte[toSend.readableBytes()];
-            toSend.getBytes(toSend.readerIndex(), toSendArray);
+            byte[] toSendArray = toSend.toArray();
```
jvrao: We still have a buffer copy into a newly allocated array, and this is happening per bookie, WQ times.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieNettyServer.java
```diff
@@ -302,6 +303,9 @@ protected void initChannel(SocketChannel ch) throws Exception {
                         new BookieSideConnectionPeerContextHandler();
                     ChannelPipeline pipeline = ch.pipeline();
 
+                    // For ByteBufList, skip the usual LengthFieldPrepender and have the encoder itself to add it
+                    pipeline.addLast("bytebufList", ByteBufList.ENCODER_WITH_SIZE);
```
eolivelli: If I understand correctly, we are not changing the structure on the wire, it is 100% compatible with old clients, can you confirm @merlimat ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
merlimat: Yes, this is because protobuf cannot take a direct memory buffer with the payload.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieNettyServer.java
merlimat: Yes, there is absolutely no change in the wire format. The "encoder" here is to translate the `ByteBufList` into a series of `ByteBuf` to be written on the socket. 

The `ENCODER_WITH_SIZE` is to avoid the regular `LenghtFieldPrepender` to insert the size header for each item of the list. Since the `ByteBufList.Encoder` is the last encoder in the sequence, `LenghtFieldPrepender` will ignore the `ByteBufList` and we're manually adding the size (only once) in the encoder itself.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
jvrao: Why can't we have the change done at pendingAddOp level and avoid 3 copies? These are immutable buffers, no need to have WQ copies.

dlg99: we can replace ByteString.copyFrom(toSendArray) with UnsafeByteOperations.unsafeWrap(..). IIRC  copyFrom() may end up actually doing copy.
 
Also please look into using combo of UnsafeByteOperations.unsafeWrap-ing each ByteBuf separately with ByteString.concat on top. This should skip extra allocations (I haven't tried that yet)

merlimat: I agree with the above comments, though I'd rather leave this outside the scope of this PR which is a 1-1 replacement for `DoubleByteBuf`. The `byte[]` copies are not changing with this change.

eolivelli:
Tests cannot compile due to a warning, can you fix it please? So that we can move forward

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
dlg99: ok, LGTM then.

merlimat:
> Tests cannot compile due to a warning, can you fix it please? So that we can move forward 

Sure, will fix now

sijie:
@merlimat can you fix #1108 first? it would be good to fix #1108 and make it to master and 4.6.2, since you already have the pull request there. then this change can come after #1108 is in.

Otherwise if we take this change in, you have to redo #1108 for 4.6.x.

merlimat:
Merged #1108 and rebased on master

asfgit:

sijie:
@merlimat 

it seems that benchmark module can not compile

```
[ERROR] /home/travis/build/apache/bookkeeper/bookkeeper-benchmark/src/main/java/org/apache/bookkeeper/benchmark/BenchBookie.java:[189,32] incompatible types: io.netty.buffer.ByteBuf cannot be converted to org.apache.bookkeeper.util.ByteBufList
[ERROR] /home/travis/build/apache/bookkeeper/bookkeeper-benchmark/src/main/java/org/apache/bookkeeper/benchmark/BenchBookie.java:[206,32] incompatible types: io.netty.buffer.ByteBuf cannot be converted to org.apache.bookkeeper.util.ByteBufList
[ERROR] /home/travis/build/apache/bookkeeper/bookkeeper-benchmark/src/main/java/org/apache/bookkeeper/benchmark/BenchBookie.java:[224,32] incompatible types: io.netty.buffer.ByteBuf cannot be converted to org.apache.bookkeeper.util.ByteBufList
```

PR #1142, opened 2018-02-12
Gather disk information before CI build

ivankelly:
I suspect that we are getting quite a few flakes because of over-full
disks in the CI workers. I'd like to confirm this hypothesis and
gather some information before taking corrective action.

### DIFF
--- a/.test-infra/jenkins/job_bookkeeper_precommit_java8.groovy
+++ b/.test-infra/jenkins/job_bookkeeper_precommit_java8.groovy
@@ -22,6 +22,11 @@ import common_job_properties
 mavenJob('bookkeeper_precommit_pullrequest_java8') {
   description('precommit verification for pull requests of <a href="http://bookkeeper.apache.org">Apache BookKeeper</a> in Java 8.')
 
+  // Temporary information gathering to see if full disks are causing the builds to flake
+  preBuildSteps {
+    shell("df -h")
+  }
+
   // Execute concurrent builds if necessary.
   concurrentBuild()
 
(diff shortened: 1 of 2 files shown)


ivankelly:
Test here:
https://builds.apache.org/job/bookkeeper_precommit_pullrequest_java9_testdf/1/console

.test-infra/jenkins/job_bookkeeper_precommit_java8.groovy
```diff
@@ -22,6 +22,11 @@ import common_job_properties
 mavenJob('bookkeeper_precommit_pullrequest_java8') {
   description('precommit verification for pull requests of <a href="http://bookkeeper.apache.org">Apache BookKeeper</a> in Java 8.')
 
+  // Temporary information gathering to see if full disks are causing the builds to flake
+  preBuildSteps {
+    shell("df -h")
```
eolivelli: maybe we should add also a "pwd" ? in order to know which is the disk on which the build is running ?

ivankelly: Can do. From the test there only is /. But will add

ivankelly: Added a ps -ef also, in case something else is also running.

eolivelli:
What about logging info about ulimit ? We are not running as root I hope

ivankelly:
No, run as jenkins.

https://builds.apache.org/job/bookkeeper_precommit_pullrequest_java9_testdf/2/console <- testing

ivankelly:
I want to see what the state of the disks is before we start the test. I want to see what the disks are in the case of a failure. Is the /tmp separate? If so, couldn't we run in the workspace. Absolutely this is for debugging purposes, CI is flaking on environmental issues. It needs to be debugged.

ivankelly:
@sijie @eolivelli I checked with the infra folks on hipchat. they say it's fine

ivankelly:
Merging this.

PR #1145, opened 2018-02-12
Require green CI before merge

ivankelly:
CI has been broken since the 2018-02-06, due to a non-green patch
being merged. With the recent flake fixes, CI should only go green if
there's a good reason. More over, any remaining problems will get
fixed faster, if people non-green builds start being a pain for
people.

So, this patch blocks merging with the merge script, if all checks are
not green. It can be overriden by adding a comment with the text
"Ignore CI" (case insensitive), on the PR.


ivankelly:
Merging

PR #1148, opened 2018-02-15
Run integration tests on precommit

ivankelly:
Builds integration test docker images and runs all tests under tests/.


ivankelly:
Has been tested here:
https://builds.apache.org/view/B/view/BookKeeper/job/bookkeeper_precommit_integrationtests/

Only difference in this version is the commit trigger.

ivankelly:
don't merge yet, it will fail building image, because i forgot a script in #1146

ivankelly:
#1151 should be merged before this

PR #1151, opened 2018-02-15
Add script missing from custom yahoo docker image

ivankelly:
427eb81 references a script which I never added to the commit,
which means the image doesn't build.

This commit adds the script.

### DIFF
--- a/tests/docker-images/all-released-versions-image/scripts/install-pulsar-tarball.sh
+++ b/tests/docker-images/all-released-versions-image/scripts/install-pulsar-tarball.sh
@@ -1,0 +1,42 @@
+#!/usr/bin/env bash
+#
+#/**
+# * Licensed to the Apache Software Foundation (ASF) under one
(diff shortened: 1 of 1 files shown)


tests/docker-images/all-released-versions-image/scripts/install-pulsar-tarball.sh
```diff
@@ -0,0 +1,42 @@
...
+# * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# * See the License for the specific language governing permissions and
+# * limitations under the License.
+# */
+
+set -e
+
+TARBALL=$(ls apache-pulsar-*-incubating-bin.tar.gz | tail -1)
+
+if [ -f $T.asc ]; then
```
eolivelli: Why the signature is always present?

ivankelly: copy paste from other file, will change

PR #1156, opened 2018-02-16
Call static methods and build callbacks in MavenClassLoader

ivankelly:
To properly test some functionality we need to be able to call static
methods (e.g. for the creation of ByteBuf), and also to be able to
build callbacks (e.g. to call async methods).

This patch implements methods for doing this. For static methods, you
need to pass the className, the method name and a list of
arguments. It first tries to find the method using non-primitive
types, but falls back to trying primative types if possible.

Invokation looks like:
```
cl.callStaticMethod("org.apache.bookkeeper.client.BKException",
                    "create", [-101])
```

For callbacks you specify an interface and the a groovy closure.
It is assumed the closure will take as many arguments as the
callback. The invokation handler doesn't check the method name
called on the callback object, so it only works with callbacks
with a single method.

Invokation looks like:
```
def callback = classLoader.createCallback(
    "org.apache.bookkeeper.client.AsyncCallback\$AddCallback",
        { rc, _ledger, entryId, ctx ->
            // do something
        })
```

### DIFF
--- a/tests/integration-tests-utils/src/main/java/org/apache/bookkeeper/tests/MavenClassLoader.java
+++ b/tests/integration-tests-utils/src/main/java/org/apache/bookkeeper/tests/MavenClassLoader.java
@@ -127,7 +127,7 @@ public class MavenClassLoader implements AutoCloseable {
                         } else {
                             return k;
                         }
-                    } catch (IllegalAccessException|NoSuchFieldException nsfe) {
+                    } catch (IllegalAccessException | NoSuchFieldException nsfe) {
                         return k;
                     }
                 }).toArray(Class[]::new);
--- a/tests/integration-tests-utils/src/main/java/org/apache/bookkeeper/tests/MavenClassLoader.java
+++ b/tests/integration-tests-utils/src/main/java/org/apache/bookkeeper/tests/MavenClassLoader.java
@@ -23,10 +23,14 @@ package org.apache.bookkeeper.tests;
 import com.google.common.collect.Lists;
 
 import java.io.File;
+import java.lang.reflect.InvocationHandler;
+import java.lang.reflect.Method;
+import java.lang.reflect.Proxy;
 import java.net.URL;
 import java.net.URLClassLoader;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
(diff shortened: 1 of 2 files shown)


ivankelly:
retest this please // to pull in integrationtests

tests/integration-tests-utils/src/main/java/org/apache/bookkeeper/tests/MavenClassLoader.java
```diff
@@ -102,6 +106,47 @@ public static MavenClassLoader forBookKeeperVersion(String version) throws Excep
...
+
+    public Object callStaticMethod(String className, String methodName, ArrayList<?> args) throws Exception {
+        Class<?> klass = Class.forName(className, true, classloader);
+
+        try {
+            Class<?>[] paramTypes = args.stream().map((a)-> a.getClass()).toArray(Class[]::new);
+            return klass.getMethod(methodName, paramTypes).invoke(null, args.stream().toArray(Object[]::new));
+        } catch (NoSuchMethodException nsme) {
+            // maybe the params are primitives
+            Class<?>[] paramTypes = args.stream().map((a) -> {
```
eolivelli: I never used such kind of code for coercions. Do you this it is needed?
I will look into my code and send some pointer

ivankelly: It is. int.class is different to Integer.class, so without this, it won't find the method. I hit it when trying to find the static constructor for BKException. Groovy passed in an Integer as the argument. The method itself takes an int.

asfgit:

tests/integration-tests-utils/src/main/java/org/apache/bookkeeper/tests/MavenClassLoader.java
eolivelli: Sure, I know, my point was about this way of doing coercions which I never seen and I wrote lot of code about this problems.
Any way this works and I think we can keep your proposal. I have no time today and this weekend and no access to laptop.

tests/integration-tests-utils/src/main/java/org/apache/bookkeeper/tests/MavenClassLoader.java
```diff
@@ -102,6 +106,47 @@ public static MavenClassLoader forBookKeeperVersion(String version) throws Excep
...
+                    Class<?> k = a.getClass();
+                    try {
+                        Object type = k.getField("TYPE").get(null);
+                        if (type instanceof Class<?>) {
+                            return (Class<?>)type;
+                        } else {
+                            return k;
+                        }
+                    } catch (IllegalAccessException|NoSuchFieldException nsfe) {
```
eolivelli: nit: IllegalAccessException(space)|(space)NoSuchFieldException

tests/integration-tests-utils/src/main/java/org/apache/bookkeeper/tests/MavenClassLoader.java
ivankelly: There's probably a much better way to do this, I just went for the quickest that works.

tests/integration-tests-utils/src/main/java/org/apache/bookkeeper/tests/MavenClassLoader.java
ivankelly: There's nothing in the sun style guide about this, but in the tutorial there's no spaces. Personally I prefer to leave out the spaces, as it clusters all the types together visually.

https://docs.oracle.com/javase/tutorial/essential/exceptions/catch.html

ivankelly:
retest this please // docker wasn't running on integration node

ivankelly:
retest this please (have deleted individual retry job)

ivankelly:
@sijie added test for the static method stuff. For the callback is more awkward, as it implies having groovy. the compat test that uses it will implicitly test it though

sijie:
retest this please

sijie:
all CI passed. integration test is still pending (due to too many integration tests are in the queue running). since this change is about maven class loader, it has nothing to do with integration tests, so ignore integration test CI to merge this pull request.

sijie:
IGNORE CI

ivankelly:
MavenClassloader is _only_ used in integration tests. Should be fine though, as nothing existing is modified.

PR #1160, opened 2018-02-16
Dump config

reddycharan:

Dump config
- add utility method to AbstractConfiguration to return config as string and log it
- dump the config during the start of the BookieServer

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
@@ -618,4 +618,22 @@ public abstract class AbstractConfiguration<T extends AbstractConfiguration>
      * Trickery to allow inheritance with fluent style.
      */
     protected abstract T getThis();
+
+    /**
+     * converts the config into string format, by appending the config values
+     * and separated by 'separator'.
+     *
+     * @param separator
+     *            separator to separate the configs in string
+     */
+    public String configAsString(String separator) {
+        StringBuilder configString = new StringBuilder();
+        Iterator<String> keys = this.getKeys();
+        while (keys.hasNext()) {
+            String key = (String) keys.next();
+            Object value = getProperty(key);
+            configString.append(key + "=" + value + separator);
+        }
+        return configString.toString();
+    }
 }
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
@@ -20,13 +20,18 @@ package org.apache.bookkeeper.conf;
 import static org.apache.bookkeeper.conf.ClientConfiguration.CLIENT_AUTH_PROVIDER_FACTORY_CLASS;
 
 import java.net.URL;
+import java.util.HashMap;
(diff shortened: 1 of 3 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
```diff
@@ -618,4 +618,22 @@ public String getTLSEnabledProtocols() {
...
     protected abstract T getThis();
+
+    /**
+     * converts the config into string format, by appending the config values
+     * and separated by 'separator'.
+     *
+     * @param separator
+     *            separator to separate the configs in string
+     */
+    public String configAsString(String separator) {
```
sijie: http configuration service is also dumping config. can we reuse (or consolidate) the logic of these two places.

reddycharan: not sure what you are referring to? where is Http configuration used and how it can be reused?

sijie: this is what I am referring to : https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/server/http/service/ConfigurationService.java#L52

I was thinking of adding `String asJson()` in `AbstractConfiguration`. so it can be used in 1) configuration service and 2) logging purpose as here. that is what I was talking about consolidating the logic on how we dump the configuration. so we don't need to have multiple logic on dumping configuration.

eolivelli: I am ok for the json version, but it can be a second change.
I have this method which dumps the config in several projects.
I have an additional use case:
I start the bookie programmatically and  then I dump the actual configuration to a file so that it is possible to use bookkeeper shell commands.
So I like this function

sijie: well - if you are saving an configuration object to a file, you should consider using PropertiesConfiguration to save the composite configuration. commons configuration already provides extensive utils for this purpose. why do you want to reinvent this again?

I think the point here is - if we want this configuration to be dumped, it is good to have one implementation rather than multiple implementations.

eolivelli: PropertiesConfiguration -> I never noticed this utility, thank you

I think having the config in JSON format inside log files is not very intuitive, but I can live with it, as far as they are "pretty printed"

sijie: the config here is just plain key/value pairs. so there is no much difference between a properties-like format and a json format. and yes, they are pretty printed if you check this: https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/util/JsonUtil.java#L34

eolivelli: okay for JSON for me

reddycharan: ok..will add "String asJson()" in AbstractConfiguration class and will de-dupe the code.

PR #1168, opened 2018-02-18
Fix TestBKSyncLogReader

sijie:

*Problem*

The test case `TestBKSyncLogReader` is waiting an entry beyond a control record, where the number of control records can
vary depend on the timing of flush()/commit(). This introduces the flakiness of this test.

*Solution*

This change changes to wait for readahead go beyond the last user record, which is more deterministic.


sijie:
retest this please

sijie:
retest this please

jvrao:
@sijie  can we have a DL tag and add that flag to all DL related changes. Something we can use for all subprojects and major initiatives. I did something similar for the durability. We can for sure use it for the new metadata store.

sijie:
@jvrao yes. I have added tag `area/dlog`. I forgot to label this PR.

sijie:
retest this please

sijie:
the `ReadAheadReader` has caught up at line 172. so we can't use that again. The logic here is testing reads after readahead reader caught up.

Actually I think a better approach is to rewrite how we verify the read records. will update the PR with a different approach. will ping it after finish it.

philipsu522:
LGTM +1

sijie:
@jiazhai @yzang @philipsu522 I have changed the way how the test was written. can you guys review it again?

stream/distributedlog/core/src/test/java/org/apache/distributedlog/TestBKSyncLogReader.java
```diff
@@ -184,26 +184,19 @@ public void testReadRecordsAfterReadAheadCaughtUp() throws Exception {
...
-        // and return null to claim it as caughtup
-        LogRecord record = reader.readNext(false);
-        int numReads = 0;
+        // resume reading from sync reader util it consumes 20 records
         long expectedTxId = 1L;
-        while (null != record) {
-            ++numReads;
+        for (int i = 0; i < 20; i++) {
+            LogRecord record = reader.readNext(false);
+            while (null == record) {
```
yzang: Nit: we can probably use "do ... while" here?

sijie:
> Why do we remove the sleep between read? What if the polling speed is faster than the read ahead reading speed, and when we try to read the 21st record, we get null, but read ahead do get the record after the assert is done?

A good point.

the `sleep` was used for polling next readahead entry id. I have rewritten the test case to eliminate polling next readahead entry id.

The approach I am using is using blockingRead to read until we finish reading 20 records. `readNext(false)` is a blocking call, which will wait up to the wait time configured at the configuration.

https://github.com/apache/bookkeeper/blob/master/stream/distributedlog/core/src/main/java/org/apache/distributedlog/BKSyncLogReader.java#L145

PR #1171, opened 2018-02-19
Enforce zkAcls on initNewCluster

sijie:

initNewCluster was introduced #979. The new command doesn't honor to zookeeper acls settings. This change is to change those zookeeper calls to use `zkAcls`.


sijie:
@reddycharan can you review this change? #978 introduced `initNewCluster` in bookie registration manager. but the new methods don't use zk acls. This change is fixing that.

reddycharan:
LGTM

sijie:
IGNORE IT CI

sijie:
IT ci failure is not related to this change here.

PR #1172, opened 2018-02-19
Use BookieServer#getLocalAddress as the identifier in the test

sijie:

*Problem*

Issue #1097 introduced using `loopback` nic as the bookie server identifier. However a few test cases construct the bookie socket address using `InetAddress.getLocalHost()` and the bookie port. This bookie socket address can be different from the actual socket address that a bookie is advertising to zookeeper. It will cause test cases fail with certain network settings.

*Solution*

This change use `BookieServer#getLocalAddress()` to retrieve the actual bookie socket address. so it would respect to the server configuration and use the right socket address in the test.

### DIFF
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieRecoveryTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieRecoveryTest.java
@@ -28,7 +28,6 @@ import static org.junit.Assert.fail;
 import io.netty.buffer.ByteBuf;
 
 import java.io.IOException;
-import java.net.InetAddress;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Enumeration;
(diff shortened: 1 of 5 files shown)


sijie:
@reddycharan can you review this change? it is related to the loopback change at #1097

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieRecoveryTest.java
```diff
@@ -426,12 +421,7 @@ public void testSyncBookieRecoveryToSpecificBookie() throws Exception {
         writeEntriestoLedgers(numMsgs, 10, lhs);
 
         // Call the sync recover bookie method.
-        BookieSocketAddress bookieSrc = new BookieSocketAddress(InetAddress.getLocalHost().getHostAddress(),
-          initialPort);
-        BookieSocketAddress bookieDest = new BookieSocketAddress(InetAddress.getLocalHost().getHostAddress(),
-          newBookiePort);
-        LOG.info("Now recover the data on the killed bookie (" + bookieSrc + ") and replicate it to the new one ("
-          + bookieDest + ")");
+        LOG.info("Now recover the data on the killed bookie (" + bookieSrc + ") and replicate it to other bookies");
```
reddycharan: nit: in the previous log line it was saying that it copied to destination bookie. Here it is saying "replicate it to other bookies", is that correct?

reddycharan:
LGTM

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieRecoveryTest.java
sijie: yes. because `bookieDest` is actually not used.

sijie:
IGNORE IT CI

sijie:
IT ci failure is unrelated to this change. so ignore IT ci to allow merging this change.

PR #1174, opened 2018-02-19
Use travis_wait on running mvn test on travis

sijie:

*Problem*

The travis build can take more than 30 minutes, which exceeds the default limit. so the travis builds on master usually fail

*Solution*

Configure the travis build to use `travis_wait` and configure a higher timeout value.

This was set in dlog travis build : https://github.com/apache/distributedlog/blob/master/.travis.yml#L41


sijie:
it is travis change and travis build passed. so ignore other CI to merge it.

sijie:
IGNORE CI

PR #1175, opened 2018-02-19
Create a forked JVM for each bc test

ivankelly:
Until now, BC tests have been running in the maven processes because
they do funny stuff with the classloader which didn't seem to work
with later versions of the surefire plugin. However, using a version
of surefire less than or equal to 2.8.1, we can use
useSystemClassLoader to the same effect, and therefore fork a JVM for
each test. This is desirable to isolate the tests from each other.


ivankelly:
@sijie i've also doubled the amount of mem BC tests use to 4G. A lot of the failures I've seen are JVM running out of memory

sijie:
@ivankelly sgtm

PR #1178, opened 2018-02-19
Fix name on smoke test in pom

ivankelly:
It says backwards compat, should be integration


sijie:
this change only changes the module name. so ignore IT CI to allow merging it.

sijie:
IGNORE IT CI

PR #1179, opened 2018-02-19
Handle for 4.6.0 and 4.6.1 specially in bc test shutdown

ivankelly:
There is a race in 4.6.0 and 4.6.1 which can cause the zk session to
be leaked. The bc test should take this into account, and wait for the
session for expire before continuing.


sijie:
all CI passed. the IT one failed due to OOM. The OOM has been fixed by #1177 (doubling the memory usage to 4G). so ignore IT CI to merge it.

sijie:
IGNORE IT CI

PR #1180, opened 2018-02-19
V2 decoder marks the reads index of the payload on write

ivankelly:
This means that any storage implementation that receives that payload,
can reset the reader index, without jumping into bytes it knows
nothing about.

This was the source of an issue when using V2 protocol with
DbLedgerStorage where the data could not be successfully read back
from the ledger.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtoEncoding.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtoEncoding.java
@@ -167,6 +167,9 @@ public class BookieProtoEncoding {
                 // Read ledger and entry id without advancing the reader index
                 ledgerId = packet.getLong(packet.readerIndex());
                 entryId = packet.getLong(packet.readerIndex() + 8);
+                // mark the reader index so that any resets will return to the
+                // start of the payload
+                packet.markReaderIndex();
                 return BookieProtocol.ParsedAddRequest.create(
                         version, ledgerId, entryId, flags,
                         masterKey, packet.retain());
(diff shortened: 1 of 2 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtoEncoding.java
```diff
@@ -167,6 +167,9 @@ public Object decode(ByteBuf packet)
                 // Read ledger and entry id without advancing the reader index
                 ledgerId = packet.getLong(packet.readerIndex());
                 entryId = packet.getLong(packet.readerIndex() + 8);
+                // mark the reader index so that any resets will return to the
+                // start of the payload
+                packet.markReaderIndex();
```
merlimat: I think the fix should be in the DbLedgerStorage. it should not depend on the marker being set.

Eg, in `DbLedgerStorage` : 
```java
    public long addEntry(ByteBuf entry) throws IOException {
        long startTime = MathUtils.nowInNano();

        long ledgerId = entry.readLong();
        long entryId = entry.readLong();
        entry.resetReaderIndex();
```

could be changed into 
```java
long ledgerId = entry.getLong(entry.readerIndex());
long entryId = entry.getLong(entry.readerIndex() + 8);
```

without need for the reset index.

ivankelly: Sure. I was looking there, but I put in the decoder because that's how it was in yahoo-4.3 branch. i'll leave it here, and remove the dependency on the behaviour in DbStorage too.

sijie:
the failed IT ci seems to be unrelated. ignore IT CI to allow merging it.

sijie:
IGNORE IT CI

PR #1183, opened 2018-02-20
Fix flaky travis build on checking license header

sijie:

*Problem*

In some environments, `BK_VERSION` is computed as `Downloading`, which is not a valid bookkeeper version. It fails the `dev/check-binary-license` script.

```
[0K$ if [ "$TRAVIS_OS_NAME" == "linux" ]; then dev/check-binary-license ./bookkeeper-dist/server/target/bookkeeper-server-${BK_VERSION}-bin.tar.gz; fi
tar: ./bookkeeper-dist/server/target/bookkeeper-server-Downloading: Cannot open: No such file or directory
tar: Error is not recoverable: exiting now
tar: ./bookkeeper-dist/server/target/bookkeeper-server-Downloading: Cannot open: No such file or directory
tar: Error is not recoverable: exiting now
tar: ./bookkeeper-dist/server/target/bookkeeper-server-Downloading: Cannot open: No such file or directory
tar: Error is not recoverable: exiting now
```

*Solution*

Remove this `BK_VERSION` setting and use "*" in the file path.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,12 +54,11 @@ before_install:
             echo "Enable testing distributedlog modules if this pull request modifies files under directory `stream/distributedlog`."
         fi
     fi
-    export BK_VERSION=`mvn org.apache.maven.plugins:maven-help-plugin:2.1.1:evaluate     -Dexpression=project.version | grep -Ev '(^\[|Download\w+:)' 2> /dev/null`
 
 script:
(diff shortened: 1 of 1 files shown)


ivankelly:
The IT failure is a result of there being no docker networks available, probably due to failed runs of tests. will push fix.

sijie:
IGNORE IT CI

sijie:
This change is only related to travis ci.

PR #1184, opened 2018-02-20
Prune networks from previous runs before IT

ivankelly:
If a build fails, it can leave networks around. Run prune before the
integration tests to avoid this.


.test-infra/jenkins/job_bookkeeper_precommit_integrationtests.groovy
```diff
@@ -39,6 +39,7 @@ freeStyleJob('bookkeeper_precommit_integrationtests') {
         shell('pwd')
         shell('df -h')
         shell('ps -eo euser,pid,ppid,pgid,start,pcpu,pmem,cmd')
+        shell('docker network prune -f') // clean up any dangling networks from previous runs
         shell('docker system events > docker.log & echo $! > docker-log.pid')
```
eolivelli: Can this interfere with other jenkins jobs on asf shared ci?

ivankelly:
It will only prune networks that have no containers attached.

ivankelly:
Actually, we can filter by name, to make it extra safe. Will push new patch.

PR #1186, opened 2018-02-20
Propogate exception when bookie shutdown fails

ivankelly:
When a bookie fails to shutdown cleanly, we should propogate the
exception that caused the problem to the main thread.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/server/Main.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/server/Main.java
@@ -217,7 +217,7 @@ public class Main {
             // the server is interrupted
             log.info("Bookie server is interrupted. Exiting ...");
         } catch (ExecutionException ee) {
-            log.error("Error in bookie shutdown", ee);
+            log.error("Error in bookie shutdown", ee.getCause());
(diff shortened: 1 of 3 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/server/Main.java
```diff
@@ -211,13 +211,14 @@ static int doMain(String[] args) {
...
-        CountDownLatch aliveLatch = new CountDownLatch(1);
-        ComponentStarter.startComponent(server, aliveLatch);
         try {
-            aliveLatch.await();
+            ComponentStarter.startComponent(server).get();
         } catch (InterruptedException ie) {
             // the server is interrupted
             log.info("Bookie server is interrupted. Exiting ...");
+        } catch (ExecutionException ee) {
+            log.error("Error in bookie shutdown", ee);
```
eolivelli: Better to log ee.getCause() ?

ivankelly: sure, changed

ivankelly:
retest this please

really strange error in IT. a docker container is destroyed immediately after it's created :/

PR #1187, opened 2018-02-20
Allow default methods in MavenClassloader callbacks

ivankelly:
AsyncCallback.AddCallback has recently acquired a default method,
which breaks the assumption that there is only one method on these
callbacks. However, if does have a default method that calls the one
method. This patch takes advantage of that to allow us to use
AddCallback in tests.


ivankelly:
This is the last change needed before I push the yahoo BC test

ivankelly:
retest this please

failure was in BookieWriteLedgerTest/testLedgerCreateAdvWithLedgerIdInLoop in both cases, due to this being a very slow test and it running twice on the same machine at the same time :/ looking into it.

PR #1189, opened 2018-02-20
Always reset interrupted state after catching InterruptedException

ivankelly:
There's many places we did not do it. I've added it.


eolivelli:
Is this a general cleanup patch or did you fall into some real problem?

ivankelly:
General cleanup, though some of the CI stuff I've been fighting was threads seeming to ignore being interrupted, which could be related.

PR #1191, opened 2018-02-21
Fix GSSAPIBookKeeperTest

sijie:

*Problem*

in GSSAPIBookKeeperTest, `InetAddress.getLocalHost().getHostName()` was used for adding principle to kdc keytab. However in the tests, it uses advertisedAddress (introduced by #1097).

So it causes authentication failed because server can't be found.

*Solution*

Change to use `Bookie. getBookieAddress(..)` to get the real hostname used by bookie for adding keytab entries.


sijie:
@reddycharan can you review this again? this test case fails at my laptop as well, due to #1097

reddycharan:
LGTM.

PR #1192, opened 2018-02-21
BP-29 (task 3): use metadata service uri for constructing registration client

sijie:

- Deprecated `getRegistrationClientClass` and changed bookkeeper client to use metadata service uri to construct metadata driver.
- Add a test case for TestZKRegistrationClient and fixes issues caught by the test case.

Master Issue: #1123 

> ---
> quickly and easily:
>
>
>
> Otherwise:
> 
> 
> ---


sijie:
@jiazhai since you were the main author of registration interfaces, can you review this change?

ivankelly:
The integration test failure looks legit

sijie:
@jiazhai I have fixed the ci failures.

jiazhai:
+1, lgtm

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
```diff
@@ -215,6 +215,8 @@ public String getMetadataServiceUri() throws ConfigurationException {
                     ledgerManagerType = org.apache.bookkeeper.meta.FlatLedgerManagerFactory.NAME;
                 } else if (factoryClass == LongHierarchicalLedgerManagerFactory.class) {
                     ledgerManagerType = LongHierarchicalLedgerManagerFactory.NAME;
+                } else if (factoryClass == org.apache.bookkeeper.meta.MSLedgerManagerFactory.class) {
```
ivankelly: Is anyone still using this?

sijie: don't think any more is using them. would like to deprecate the whole interface after completing BP-29.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeper.java
```diff
@@ -525,9 +532,9 @@ public BookKeeper(ClientConfiguration conf, ZooKeeper zk, EventLoopGroup eventLo
         // initialize ledger manager
         try {
             this.ledgerManagerFactory =
-                AbstractZkLedgerManagerFactory.newLedgerManagerFactory(conf, regClient.getLayoutManager());
-        } catch (IOException | InterruptedException e) {
-            throw e;
+                this.metadataDriver.getLedgerManagerFactory();
```
ivankelly: As with regClient. We don't need a member field for ledgerManagerFactory, as it is owned by the metadataDriver, so metadataDriver should clean up. Either that, or metadataDriver should be a pure factory and not hold a reference to ledgerManagerFactory and regClient, but as they likely rely on a shared resource (zk client?) it's probably better to just let metadataDriver own all.

bookkeeper-server/src/main/java/org/apache/bookkeeper/discover/ZKRegistrationClient.java
```diff
@@ -133,13 +122,15 @@ public void safeRun() {
         @Override
         public void accept(Versioned<Set<BookieSocketAddress>> bookieSet, Throwable throwable) {
             if (throwable != null) {
-                scheduleWatchTask(ZK_CONNECT_BACKOFF_MS);
-                firstRunFuture.completeExceptionally(throwable);
+                if (firstRunFuture.isDone()) {
```
ivankelly:
```java
if (!firstRunFuture.completeExceptionally(throwable)) {
    scheduleWatchTask(ZK_CONNECT_BACKOFF_MS); 
}
```

But in any case, I don't get what you are doing with this code. Why not reschedule if firstRunFuture is complete?

bookkeeper-server/src/main/java/org/apache/bookkeeper/discover/ZKRegistrationClient.java
```diff
@@ -303,13 +238,21 @@ public ZooKeeper getZk() {
         CompletableFuture<Void> f = new CompletableFuture<>();
         if (null == watchWritableBookiesTask) {
             watchWritableBookiesTask = new WatchTask(bookieRegistrationPath, f);
+        } else {
+            FutureUtils.proxyTo(
```
ivankelly: huh?
```watchWritableBookiesTask.firstRunFuture``` is ```f```

File: bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/MetadataDrivers.java [hunk 5]
```diff
@@ -167,17 +167,28 @@ private static void loadInitialBookieDrivers() {
...
 
         String scheme = metadataBackendScheme.toLowerCase();
         MetadataClientDriverInfo oldDriverInfo = clientDrivers.get(scheme);
-        if (null != oldDriverInfo) {
+        if (null != oldDriverInfo && !allowOverride) {
             return;
         }
         MetadataClientDriverInfo newDriverInfo = new MetadataClientDriverInfo(driver);
-        oldDriverInfo = clientDrivers.putIfAbsent(scheme, newDriverInfo);
+        if (allowOverride) {
```
ivankelly: It'd be better to handle all the special case branching based on a boolean flag in one place.
```java
if (null != oldDriverInfo) {
    if (allowOverride) {
        log.debug("Metadata client driver for {} is already there.", scheme); 
    } else {
        return;
    }
}
clientDrivers.put(scheme, newDriverInfo);
```

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/MetadataDrivers.java
```diff
@@ -191,17 +202,28 @@ public static void registerClientDriver(String metadataBackendScheme,
...
+                                            Class<? extends MetadataBookieDriver> driver,
+                                            boolean allowOverride) {
         if (!initialized) {
             initialize();
         }
 
         String scheme = metadataBackendScheme.toLowerCase();
         MetadataBookieDriverInfo oldDriverInfo = bookieDrivers.get(scheme);
-        if (null != oldDriverInfo) {
+        if (null != oldDriverInfo && !allowOverride) {
```
ivankelly: As above

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/zk/ZKMetadataDriverBase.java
```diff
@@ -120,23 +125,27 @@ protected void initialize(AbstractConfiguration<?> conf,
                               Optional<Object> optionalCtx) throws MetadataException {
         this.conf = conf;
 
-        String metadataServiceUriStr;
+        final String metadataServiceUriStr;
```
ivankelly: why not construct metadataServiceUri directly?

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
```diff
@@ -141,7 +141,7 @@ private void initialize(ServerConfiguration conf, ZooKeeper zkc)
             LedgerManagerFactory ledgerManagerFactory = AbstractZkLedgerManagerFactory
                     .newLedgerManagerFactory(
                         conf,
-                        bkc.getRegClient().getLayoutManager());
+                        bkc.getMetadataClientDriver().getLayoutManager());
```
ivankelly: not for this patch, but ledgermanagerfactories are created from the metadata driver. Shouldn't we do the same here?

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
ivankelly: :+1:

sijie:
> There seems to be a lot of stuff in patch not directly related to changing to use a URI.

they are related to registration client: 1) client need to use uri for reflection registration client. 2) I am adding/improve test cases to cover this change.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeper.java
```diff
@@ -142,6 +145,9 @@
     // Features
     final Feature disableEnsembleChangeFeature;
 
+    final MetadataClientDriver metadataDriver;
+    // Registration client to discover bookies
+    final RegistrationClient regClient;
```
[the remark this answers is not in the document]
sijie: I keep it there to reduce the amount of code that I need to change.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeper.java
```diff
@@ -441,21 +447,22 @@ public BookKeeper(ClientConfiguration conf, ZooKeeper zk, EventLoopGroup eventLo
...
                 statsLogger,
                 java.util.Optional.ofNullable(zkc));
+            this.regClient = this.metadataDriver.getRegistrationClient();
         } catch (ConfigurationException ce) {
-            LOG.error("Failed to initialize registration client", ce);
-            throw new IOException("Failed to initialize registration client", ce);
-        } catch (BKException be) {
-            LOG.error("Failed to initialize registration client", be);
-            throw new IOException("Failed to initialize registration client", be);
+            LOG.error("Failed to initialize metadata client driver", ce);
```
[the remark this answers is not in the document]
sijie: Updated the logging statements to give more meaningful actions.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeper.java
sijie: I am trying to reduce the amount code to change for places that references `ledgerManagerFactory`. this should be done via a cleanup later.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeper.java
```diff
@@ -1447,6 +1454,7 @@ public void close() throws BKException, InterruptedException {
             eventLoopGroup.shutdownGracefully();
         }
         this.regClient.close();
```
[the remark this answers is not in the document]
sijie: done

bookkeeper-server/src/main/java/org/apache/bookkeeper/discover/ZKRegistrationClient.java
sijie: the register method return a completefuture. so I am aligning the behavior with the method signature. because if register failure, the task was removed, we don't need to retry this.

bookkeeper-server/src/main/java/org/apache/bookkeeper/discover/ZKRegistrationClient.java
sijie: done.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/MetadataDrivers.java
sijie: this changes the logging behavior though. the logging message is added after `put`, that is to handle concurrent puts. the change you are proposing here doesn't do that.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/MetadataDrivers.java
sijie: same as above

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/zk/ZKMetadataDriverBase.java
sijie: any issues you are seeing not construct `metadataServiceUri`? I am not sure what to fix here?

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
sijie: that is in subsequent change. all the changes I am dealing here is about bookkeeper client and reg client. I will have a subsequent change for bookie and reg manager, a change for ledger manager and a change (or few changes) for auditor & underreplicated managers.

sijie:
@ivankelly replied and addressed your comments.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/zk/ZKMetadataDriverBase.java
ivankelly: what I'm saying is that metadataServiceUriStr isn't needed. It's used on one other place, and metadataServiceUri could equally be used there directly.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeper.java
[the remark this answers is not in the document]
ivankelly: It's very little code change to remove it, and it makes it clearer that metadataDriver is replacing regClient.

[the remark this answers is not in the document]
ivankelly: In fact, in the latest change, it's only ever referenced in the constructor, and getZkHandle (which should be killed very soon in any case).

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeper.java
ivankelly: In the latest patch, it is only ever referenced a couple of lines down in the constructor.

bookkeeper-server/src/main/java/org/apache/bookkeeper/discover/ZKRegistrationClient.java
```diff
@@ -327,9 +270,17 @@ public synchronized void unwatchWritableBookies(RegistrationListener listener) {
 
     @Override
     public synchronized CompletableFuture<Void> watchReadOnlyBookies(RegistrationListener listener) {
-        CompletableFuture<Void> f = new CompletableFuture<>();
+        CompletableFuture<Void> f;
```
ivankelly: nit: add final to ensure all branches assign something. if returning a value is something miss from scala.

bookkeeper-server/src/main/java/org/apache/bookkeeper/discover/ZKRegistrationClient.java
ivankelly: ok, so the logic is basically, if the failure has been notified to the client, give up, otherwise retry.

Really we should be using something like observables here, but that's not for this change. In any case, make the check and completion atomic as above, since we're dealing with multiple threads here.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/MetadataDrivers.java
ivankelly: Ah, true. Still, it would be nice to check allow override only once, to so the read doesn't have to keep so much state in mind while reading. How about.
```
MetadataClientDriverInfo oldDriverInfo = clientDrivers.computeIfAbsent(
        scheme, (k) -> new MetadataClientDriverInfo(driver));
if (null != oldDriverInfo) {
    log.debug("Metadata client driver for {} is already there.", scheme);
    if (allowOverride) {
        log.debug("Overriding client driver for {}", scheme);
        clientDrivers.put(scheme, new MetadataClientDriverInfo(driver));
    }
}
```

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
ivankelly: :+1:

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeper.java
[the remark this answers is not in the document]
sijie: @ivankelly MetadataClientDriver is not replacing RegClient. it is just changing how regclient was built here. I will have subsequent changes to clean up after removing `getZkHandle`.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeper.java
sijie: there is a subsequent change on ledger manager factory. would defer changing that to ledger manager factory change, not sure it worth being stuck at here.

bookkeeper-server/src/main/java/org/apache/bookkeeper/discover/ZKRegistrationClient.java
sijie: yes that is aligned to the method signature, I don't change the original design here.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/MetadataDrivers.java
sijie: not sure this is better than current implementation. would prefer stick to current implementation unless there is a huge concern here.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeper.java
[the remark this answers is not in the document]
ivankelly: It's replacing it as far as BookKeeper is concerned. BookKeeper only has RegClient to manage the ownership of regClient. This is now handed off to MetadataClientDriver.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeper.java
ivankelly: for ledgerManagerFactory sure, I'd like regClient removed though.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/zk/ZKMetadataDriverBase.java
ivankelly: Sure.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/MetadataDrivers.java
ivankelly: My problem with this is that allowOverride is brought into the default flow of the method. It's even the first branch from the if on L183. It means multiple returns from the method. Logging only occurs if there is a race. The existence in the map is checked twice. And what should be the non-default path is intertwined with the default.
None of these are huge in themselves, but they combine to make me dislike the current flow a lot. Another alternative, which would be a least deal with some of the intertwining...
```
MetadataClientDriverInfo oldDriverInfo = clientDrivers.get(scheme);
if (null != oldDriverInfo && !allowOverride) {
    return;
}
MetadataClientDriverInfo newDriverInfo = new MetadataClientDriverInfo(driver);
oldDriverInfo = clientDrivers.putIfAbsent(scheme, newDriverInfo);
if (null != oldDriverInfo) {
    log.debug("Metadata client driver for {} is already there.", scheme);
    if (allowOverride) {
        log.debug("Overriding client driver for {}", scheme);
        clientDrivers.put(scheme, newDriverInfo);
    }
}
```

sijie:
> It's replacing it as far as BookKeeper is concerned. BookKeeper only has RegClient to manage the ownership of regClient. This is now handed off to MetadataClientDriver.

this doesn't make the call. bookkeeper is keeping this reference for simplifying the accesses to this class, just like bookkeeper/ledger handles keep references to counters/op stats loggers.

at your initial comment: `In the latest patch, it is only ever referenced a couple of lines down in the constructor`, which is not entirely true for `regClient`. you only see the places you need to change when you actually do the code change and I did the code change and I know the places I need to update in order to compromise removing `regClient` and that's why I didn't remove it.

but anyway, since you feel really strong on this. I don't want to be stuck here. I changed all the places to remove `regClient` as you requested. 

> My problem with this is that allowOverride is brought into the default flow of the method. It's even the first branch from the if on L183. It means multiple returns from the method. Logging only occurs if there is a race. The existence in the map is checked twice. And what should be the non-default path is intertwined with the default. None of these are huge in themselves, but they combine to make me dislike the current flow a lot. Another alternative, which would be a least deal with some of the intertwining...

no strong preference on how things should be implemented here. since you have a strong preference on branches, I would change to what you suggested.

sijie:
@ivankelly addressed your comments regarding `regClient` and `MetadataDrivers`

sijie:
ping @ivankelly

ivankelly:
@sijie lgtm, will merge

Issue #570, opened 2017-10-17
Multiple active entrylogs


ivankelly:
JIRA: https://issues.apache.org/jira/browse/BOOKKEEPER-1041

Reporter: Venkateswararao Jujjuri (JV) @jvrao

Current bookkeeper is tuned for rotational HDDs. It has one active entrylog, and all the ledger/entries go to the same entrylog until it is rotated out. This is perfect for HDDs as seeks and moving head allover the disk platter is very expensive. But this is very inefficient for SSDs, as each SSD can handle multiple parallel writers, also this method is extremely inefficient for compaction as it causes write amplification and inefficient disk space usage.

Our proposal is to have multiple active entrylogs and a configuration param on how many parallel entrylogs the system can have. This way one can have ability to configure to have less (may be  one) ledger per entrylog.

### Comments from JIRA

---
*Enrico Olivelli* 2017-04-20T07:28:28.619+0000

{quote}
 But this is very inefficient for HDDs
{quote}

[~jujjuri]
did you mean SSD ? 

---
*Venkateswararao Jujjuri (JV)* 2017-04-21T16:42:29.203+0000

Yes I mean SSD. Corrected. Thanks.

---
*Charan Reddy Guttapalem* 2017-05-18T15:01:22.566+0000

Issue: 

In Bookie's EntryLogger, we are having only one current active entryLog and all the ledger/entries go to the same entrylog. This is perfect for HDDs as file syncs, seeks and moving head allover the disk platter is very expensive. But having single active Entry Log is  inefficient for SSDs, as each SSD can handle multiple parallel writers. Also, having single active EntryLog (irrespective of LedgerStorage type - interleaved/sorted), is inefficient for compaction, since entries of multiple ledgers will end up in an entrylog.

Also in SortedLedgerStorage , in the addEntry request we flush EntryMemtable, if it reaches the sizelimit. Because of this we are observing unpredictable tail latency for addEntry request. When EntryMemTable snapshot of size (64 MB) is flushed all at once, this may affect the journal addentry latency. Also, if the rate of new add requests surpasses the rate at which the EntryMemTable's previous snapshot is flushed, then at a point the current EntryMemTable map will reach the limit and since the previous snapshot flush is in progress, EntryMemTable will throttle new addRequests, which would affect addEntry latency.

Goals:

[...]

Issue #995, opened 2018-01-17
Remove asserts from code


ivankelly:
We currently assert in production code in a few places. These are disabled at runtime, and so have little value. We should replace these with thrown exceptions, or crash the bookie if they are serious enough.

We should also add a checkstyle rule to forbid them.

PR #1201, opened 2018-02-24
ISSUE #570: Entrylog per ledger

reddycharan:

Current bookkeeper is tuned for rotational HDDs. It has one active entrylog,
and all the ledger/entries go to the same entrylog until it is rotated out.
This is perfect for HDDs as seeks and moving head allover the disk platter
is very expensive. But this is very inefficient for SSDs, as each SSD
can handle multiple parallel writers, also this method is extremely
inefficient for compaction as it causes write amplification
and inefficient disk space usage.

Our proposal is to have configurable feature to have entrylog per ledger
and also to have parallel EntryMemTable flusher for SortedLedgerStorage.

Implementation Details

- created Interface and corresponding implementations for EntryLogManager
- changed logic of EntryMemTable flush parallely to have just one Runnable per ledger
- moved the complete logic of flushIntervalInBytes from EntryLogger to BufferedChannel

Master Issue: #570

### DIFF
--- a/bookkeeper-server/conf/bk_server.conf
+++ b/bookkeeper-server/conf/bk_server.conf
@@ -620,3 +620,15 @@ zkEnableSecurity=false
 # dbStorage_rocksDB_numLevels=-1
 # dbStorage_rocksDB_numFilesInLevel0=4
 # dbStorage_rocksDB_maxSizeInLevel1MB=256
+
+# config enabling/disabling entrylog per ledger feature.
+# entryLogPerLedgerEnabled=false
+
+# in entryLogPerLedger feature, the time duration used for lastaccess eviction policy for cache
+# entrylogMapAccessExpiryTimeInSeconds=300
+
+# In the case of multipleentrylogs, multiple threads can be used to flush the memtable
+# numOfMemtableFlushThreads=4
+
+# memtableFlushTimeoutInSeconds specifies the amount of time main flushthread has to wait for the processor threads to complete the flush
+# memtableFlushTimeoutInSeconds=120
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedChannel.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedChannel.java
@@ -41,20 +41,46 @@ public class BufferedChannel extends BufferedReadChannel implements Closeable {
     // The buffer used to write operations.
     protected final ByteBuf writeBuffer;
     // The absolute position of the next write operation.
-    protected volatile long position;
+    protected final AtomicLong position;
+
+    /*
+     * if unpersistedBytesBound is non-zero value, then after writing to
+     * writeBuffer, it will check if the unpersistedBytes is greater than
+     * unpersistedBytesBound and then calls flush method if it is greater.
+     *
+     * It is a best-effort feature, since 'forceWrite' method is not
+     * synchronized and unpersistedBytes is reset in 'forceWrite' method before
+     * calling fileChannel.force
+     */
+    protected final long unpersistedBytesBound;
+
+    /*
+     * it tracks the number of bytes which are not persisted yet by force
+     * writing the FileChannel. The unpersisted bytes could be in writeBuffer or
+     * in fileChannel system cache.
+     */
+    protected final AtomicLong unpersistedBytes;
 
     // make constructor to be public for unit test
     public BufferedChannel(FileChannel fc, int capacity) throws IOException {
         // Use the same capacity for read and write buffers.
-        this(fc, capacity, capacity);
+        this(fc, capacity, 0L);
     }
 
-    public BufferedChannel(FileChannel fc, int writeCapacity, int readCapacity) throws IOException {
+    public BufferedChannel(FileChannel fc, int capacity, long unpersistedBytesBound) throws IOException {
+        // Use the same capacity for read and write buffers.
+        this(fc, capacity, capacity, unpersistedBytesBound);
+    }
+
+    public BufferedChannel(FileChannel fc, int writeCapacity, int readCapacity, long unpersistedBytesBound)
+            throws IOException {
         super(fc, readCapacity);
         this.writeCapacity = writeCapacity;
-        this.position = fc.position();
-        this.writeBufferStartPosition.set(position);
+        this.position = new AtomicLong(fc.position());
+        this.writeBufferStartPosition.set(position.get());
         this.writeBuffer = ByteBufAllocator.DEFAULT.directBuffer(writeCapacity);
+        this.unpersistedBytes = new AtomicLong(0);
+        this.unpersistedBytesBound = unpersistedBytesBound;
     }
 
     @Override
@@ -70,20 +96,34 @@ public class BufferedChannel extends BufferedReadChannel implements Closeable {
      * @param src The source ByteBuffer which contains the data to be written.
      * @throws IOException if a write operation fails.
      */
-    public synchronized void write(ByteBuf src) throws IOException {
+    public void write(ByteBuf src) throws IOException {
         int copied = 0;
-        int len = src.readableBytes();
-        while (copied < len) {
-            int bytesToCopy = Math.min(src.readableBytes() - copied, writeBuffer.writableBytes());
-            writeBuffer.writeBytes(src, src.readerIndex() + copied, bytesToCopy);
-            copied += bytesToCopy;
+        boolean shouldForceWrite = false;
+        synchronized (this) {
+            int len = src.readableBytes();
+            while (copied < len) {
+                int bytesToCopy = Math.min(src.readableBytes() - copied, writeBuffer.writableBytes());
+                writeBuffer.writeBytes(src, src.readerIndex() + copied, bytesToCopy);
+                copied += bytesToCopy;
 
-            // if we have run out of buffer space, we should flush to the file
-            if (!writeBuffer.isWritable()) {
-                flushInternal();
+                // if we have run out of buffer space, we should flush to the
+                // file
+                if (!writeBuffer.isWritable()) {
+                    flushInternal();
+                }
+            }
+            position.addAndGet(copied);
+            unpersistedBytes.addAndGet(copied);
+            if (unpersistedBytesBound > 0) {
+                if (unpersistedBytes.get() > unpersistedBytesBound) {
+                    flushInternal();
+                    shouldForceWrite = true;
+                }
             }
         }
-        position += copied;
+        if (shouldForceWrite) {
+            forceWrite(false);
+        }
     }
 
     /**
@@ -91,7 +131,7 @@ public class BufferedChannel extends BufferedReadChannel implements Closeable {
      * @return
      */
     public long position() {
-        return position;
+        return position.get();
     }
 
     /**
@@ -110,11 +150,15 @@ public class BufferedChannel extends BufferedReadChannel implements Closeable {
      * @throws IOException if the write or sync operation fails.
      */
     public void flush(boolean shouldForceWrite) throws IOException {
+        flush(shouldForceWrite, false);
+    }
+
+    public void flush(boolean shouldForceWrite, boolean forceMetadata) throws IOException {
         synchronized (this) {
             flushInternal();
         }
         if (shouldForceWrite) {
-            forceWrite(false);
+            forceWrite(forceMetadata);
         }
     }
 
@@ -138,6 +182,20 @@ public class BufferedChannel extends BufferedReadChannel implements Closeable {
         // the force write, any flush that happens after this may or may
         // not be flushed
         long positionForceWrite = writeBufferStartPosition.get();
+        /*
+         * since forceWrite method is not called in synchronized block, to make
+         * sure we are not undercounting unpersistedBytes, setting
+         * unpersistedBytes to the current number of bytes in writeBuffer.
+         *
+         * since we are calling fileChannel.force, bytes which are written to
+         * filechannel (system filecache) will be persisted to the disk. So we
+         * dont need to consider those bytes for setting value to
+         * unpersistedBytes.
+         *
+         */
+        synchronized (this) {
+            unpersistedBytes.set(writeBuffer.readableBytes());
+        }
         fileChannel.force(forceMetadata);
         return positionForceWrite;
     }
@@ -187,5 +245,13 @@ public class BufferedChannel extends BufferedReadChannel implements Closeable {
     public synchronized void clear() {
         super.clear();
         writeBuffer.clear();
+    }
+
+    public synchronized int getNumOfBytesInWriteBuffer() {
+        return writeBuffer.readableBytes();
+    }
+
+    long getUnpersistedBytes() {
+        return unpersistedBytes.get();
     }
 }
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
@@ -25,6 +25,10 @@ import static com.google.common.base.Charsets.UTF_8;
 import static org.apache.bookkeeper.bookie.TransactionalEntryLogCompactor.COMPACTING_SUFFIX;
 import static org.apache.bookkeeper.util.BookKeeperConstants.MAX_LOG_SIZE_LIMIT;
 
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.RemovalListener;
+import com.google.common.cache.RemovalNotification;
 import com.google.common.collect.MapMaker;
 import com.google.common.collect.Sets;
 
@@ -50,11 +54,16 @@ import java.nio.ByteBuffer;
 import java.nio.channels.FileChannel;
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.Iterator;
-import java.util.LinkedList;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Optional;
 import java.util.Set;
+import java.util.SortedMap;
+import java.util.TreeMap;
+import java.util.concurrent.Callable;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
@@ -63,13 +72,17 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
-import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 
-import org.apache.bookkeeper.bookie.LedgerDirsManager.LedgerDirsListener;
+import org.apache.bookkeeper.bookie.LedgerDirsManager.NoWritableLedgerDirException;
 import org.apache.bookkeeper.conf.ServerConfiguration;
 import org.apache.bookkeeper.util.IOUtils;
 import org.apache.bookkeeper.util.collections.ConcurrentLongLongHashMap;
 import org.apache.bookkeeper.util.collections.ConcurrentLongLongHashMap.BiConsumerLong;
+import org.apache.commons.lang.mutable.MutableInt;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -82,18 +95,19 @@ import org.slf4j.LoggerFactory;
  */
 public class EntryLogger {
     private static final Logger LOG = LoggerFactory.getLogger(EntryLogger.class);
+    private static final Long INVALID_LEDGERID = new Long(-1);
+    // log file suffix
+    private static final String LOG_FILE_SUFFIX = ".log";
 
-    private static class BufferedLogChannel extends BufferedChannel {
+    class BufferedLogChannel extends BufferedChannel {
         private final long logId;
         private final EntryLogMetadata entryLogMetadata;
         private final File logFile;
+        private Long ledgerId = INVALID_LEDGERID;
 
-        public BufferedLogChannel(FileChannel fc,
-                                  int writeCapacity,
-                                  int readCapacity,
-                                  long logId,
-                                  File logFile) throws IOException {
-            super(fc, writeCapacity, readCapacity);
+        public BufferedLogChannel(FileChannel fc, int writeCapacity, int readCapacity, long logId, File logFile,
+                long unpersistedBytesBound) throws IOException {
+            super(fc, writeCapacity, readCapacity, unpersistedBytesBound);
             this.logId = logId;
             this.entryLogMetadata = new EntryLogMetadata(logId);
             this.logFile = logFile;
@@ -113,13 +127,25 @@ public class EntryLogger {
         public ConcurrentLongLongHashMap getLedgersMap() {
             return entryLogMetadata.getLedgersMap();
         }
+
+        public boolean isLedgerDirFull() {
+            return ledgerDirsManager.isDirFull(logFile.getParentFile());
+        }
+
+        public Long getLedgerId() {
+            return ledgerId;
+        }
+
+        public void setLedgerId(Long ledgerId) {
+            this.ledgerId = ledgerId;
+        }
+
     }
 
-    volatile File currentDir;
     private final LedgerDirsManager ledgerDirsManager;
-    private final AtomicBoolean shouldCreateNewEntryLog = new AtomicBoolean(false);
+    private final boolean entryLogPerLedgerEnabled;
 
-    private volatile long leastUnflushedLogId;
+    RecentEntryLogsStatus recentlyCreatedEntryLogsStatus;
 
     /**
      * locks for compaction log.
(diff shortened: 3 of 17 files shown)


reddycharan:
**Note:** This is the first iteration for this work item. Will add testscases in the following commits. Also this work has been moved back and forth (from design perspective and change in priorities in our schedule), and rebased multiple times, so it might need more final touches. But it is ready for code review, in the mean time I'll rebase test work from my personal repo and add following commits to the PR.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -82,18 +95,19 @@
  */
 public class EntryLogger {
     private static final Logger LOG = LoggerFactory.getLogger(EntryLogger.class);
+    private static final Long INVALID_LEDGERID = new Long(-1);
+    // log file suffix
+    private static final String LOG_FILE_SUFFIX = ".log";
 
-    private static class BufferedLogChannel extends BufferedChannel {
+    class BufferedLogChannel extends BufferedChannel {
```
sijie: you change this to inner class, because of line 132. however I am not sure we should add `isLedgerDirFull` should be part of `BufferedLogChannel`. `isSomethingFull` should be part of ledger dirs manager, not part of an entry log channel.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -1435,4 +1859,40 @@ static long fileName2LogId(String fileName) {
...
+    /**
+     * Datastructure which maintains the status of logchannels. When a
+     * logChannel is created entry of < entryLogId, false > will be made to this
+     * sortedmap and when logChannel is rotated and flushed then the entry is
+     * updated to < entryLogId, true > and all the lowest entries with
+     * < entryLogId, true > status will be removed from the sortedmap. So that way
+     * we could get least unflushed LogId.
+     *
+     */
+    class RecentEntryLogsStatus {
```
sijie: this can be static, no?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -1435,4 +1859,40 @@ static long fileName2LogId(String fileName) {
...
+        private SortedMap<Long, Boolean> entryLogsStatusMap;
```
sijie: final

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -1435,4 +1859,40 @@ static long fileName2LogId(String fileName) {
...
+            this.leastUnflushedLogId = leastUnflushedLogId;
+        }
+
+        synchronized void createdEntryLog(Long entryLogId) {
+            entryLogsStatusMap.put(entryLogId, false);
+        }
+
+        synchronized void flushRotatedEntryLog(Long entryLogId) {
+            entryLogsStatusMap.replace(entryLogId, true);
+            while ((!entryLogsStatusMap.isEmpty()) && (entryLogsStatusMap.get(entryLogsStatusMap.firstKey()))) {
```
sijie:
```
map.entrySet().removeIf(entry -> {
     boolean flushed = entry.getValue();
     if (flushed) {
         leastUnflushedLogId = entry.getKey() + 1;
     }
     return flushed;
});
```

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -386,12 +412,37 @@ public BufferedReadChannel getFromChannels(long logId) {
...
+        return recentlyCreatedEntryLogsStatus.getLeastUnflushedLogId();
     }
 
-    synchronized long getCurrentLogId() {
-        return logChannel.getLogId();
+    long getPreviousAllocatedEntryLogId() {
+        return entryLoggerAllocator.getPreallocatedLogId();
+    }
+
+    boolean rollLogsIfEntryLogLimitReached() throws IOException {
```
sijie: I think this method should belong to EntryLogManager. so you don't need to expose `getCopyOfCurrentLogs()` to EntryLogger. for the simple entry log manager, you don't need to covert it back to a HashSet and do the while loop here.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -406,98 +457,39 @@ File getCurCompactionLogFile() {
...
-            // Append ledgers map at the end of entry log
-            appendLedgersMap(logChannel);
-
-            BufferedLogChannel newLogChannel = entryLoggerAllocator.createNewLog();
-            logChannelsToFlush.add(logChannel);
-            LOG.info("Flushing entry logger {} back to filesystem, pending for syncing entry loggers : {}.",
-                    logChannel.getLogId(), logChannelsToFlush);
-            for (EntryLogListener listener : listeners) {
-                listener.onRotateEntryLog();
+    void createNewLog(Long ledgerId) throws IOException {
```
sijie: I think it is better to have this method part of EntryLogManager as well, like rollLog should be part of EntryLogManager. so all the details on how to organize the mapping between ledgerId to entryLog can be hidden. and EntryLogger doesn't have to know details about locking.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -794,6 +847,372 @@ private long readLastLogId(File f) {
...
+         * Returns eligible writable ledger dir for the creation next entrylog
+         */
+        File getDirForNextEntryLog(List<File> writableLedgerDirs);
+    }
+
+    class EntryLogManagerForSingleEntryLog implements EntryLogManager {
+
+        private BufferedLogChannel activeLogChannel;
+        private Lock lockForActiveLogChannel;
+        private final Set<BufferedLogChannel> rotatedLogChannels;
```
sijie: you are changing the semantic when flushing rotated log channels. the original implementation is using List rather than Set, so the entry log files are actually flushing in sequence, which is much easier for debugging. and this change here is completely changing the expected behavior. just because it is going to adopt to per ledger change.

as what I suggested in the other places, `EntryLogManager` should do a much higher level abstraction on `createNewLog`, `rollLog`, which then hide how to manage entry logs, locking, flush, and checkpoint in the implementation. Rather than let one implementation adopt the pattern of the other implementation.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -406,98 +457,39 @@ File getCurCompactionLogFile() {
...
+                logChannel.flush(false);
+
+                // Append ledgers map at the end of entry log
+                appendLedgersMap(ledgerId);
+
+                BufferedLogChannel newLogChannel = entryLoggerAllocator.createNewLog();
+                entryLogManager.setCurrentLogForLedger(ledgerId, newLogChannel);
+                LOG.info("Flushing entry logger {} back to filesystem, pending for syncing entry loggers : {}.",
+                        logChannel.getLogId(), entryLogManager.getCopyOfRotatedLogChannels());
+                if (!entryLogPerLedgerEnabled) {
```
sijie: if you have an interface `createNewLog` at `EntyLogManager`, you don't need to do such branch logic here. because the simple entry log manager can do the roate entry log, while the per ledger entry log manager doesn't have to.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -506,78 +498,86 @@ void createNewLog() throws IOException {
...
     /**
      * Append the ledger map at the end of the entry log.
      * Updates the entry log file header with the offset and size of the map.
      */
-    private void appendLedgersMap(BufferedLogChannel entryLogChannel) throws IOException {
-        long ledgerMapOffset = entryLogChannel.position();
-
-        ConcurrentLongLongHashMap ledgersMap = entryLogChannel.getLedgersMap();
-        int numberOfLedgers = (int) ledgersMap.size();
+    private void appendLedgersMap(Long ledgerId) throws IOException {
```
sijie: I think a better change here is to move `appendLedgersMap` as part of `BufferedLogChannel`. so this method doesn't have to be aware of `ledgerId`.  the awareness of `ledgerId` should be in per ledger log manager.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -588,93 +588,146 @@ public void accept(long ledgerId, long size) {
...
+                }
+            }
+            if (entryLogPreAllocationEnabled) {
+                /*
+                 * We should submit new callable / create new instance of future only if the previous preallocation is
+                 * null or if it is done. This is needed because previous preallocation has not completed its execution
+                 * since it is waiting for lock on EntryLoggerAllocator.this.
+                 */
+                if ((preallocation == null) || preallocation.isDone()) {
+                    preallocation = allocatorExecutor.submit(new Callable<BufferedLogChannel>() {
```
sijie: please use lambda here as how the original code was written.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -588,93 +588,146 @@ public void accept(long ledgerId, long size) {
...
+                     * for this it means preallocation future must have either
+                     * completed creating new log or still waiting for lock on
+                     * this object to execute allocateNewLog method. So we
+                     * should try to get result of the future without waiting.
+                     * If it fails with TimeoutException then call
+                     * allocateNewLog explicitly since we are holding the lock
+                     * on this anyhow.
+                     *
+                     */
+                    bc = preallocation.get(0, TimeUnit.MILLISECONDS);
```
sijie: I think this change the behavior, which can cause an entry log is allocated but never used, no?

The problem why you need to change the behavior, is because you changed to use `synchronized` here. is the change here really needed?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -588,93 +588,146 @@ public void accept(long ledgerId, long size) {
...
                 }
-                String logFileName = Long.toHexString(preallocatedLogId) + suffix;
-                for (File dir : list) {
-                    newLogFile = new File(dir, logFileName);
+                /*
+                 * make sure there is no entrylog which already has the same
+                 * logID. Have to check all the ledegerdirs. If already there is
+                 * an entrylog with that logid then move to next ID.
+                 */
+                logFileName = Long.toHexString(preallocatedLogId) + ".log";
```
sijie: ".log" should be using the new CONSTANT you introduced?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -794,6 +847,372 @@ private long readLastLogId(File f) {
...
+         */
+        void removeFromRotatedLogChannels(BufferedLogChannel rotatedLogChannelToRemove);
+
+        /*
+         * Returns eligible writable ledger dir for the creation next entrylog
+         */
+        File getDirForNextEntryLog(List<File> writableLedgerDirs);
+    }
+
+    class EntryLogManagerForSingleEntryLog implements EntryLogManager {
```
sijie: this class can be static, no?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -794,6 +847,372 @@ private long readLastLogId(File f) {
...
+
+        private BufferedLogChannel activeLogChannel;
+        private Lock lockForActiveLogChannel;
```
sijie: final

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedChannel.java
```diff
@@ -110,11 +150,15 @@ public long getFileChannelPosition() {
      * @throws IOException if the write or sync operation fails.
      */
     public void flush(boolean shouldForceWrite) throws IOException {
+        flush(shouldForceWrite, false);
+    }
+
+    public void flush(boolean shouldForceWrite, boolean forceMetadata) throws IOException {
```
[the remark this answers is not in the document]
reddycharan: Call Hierarchy of this method : 

flush(boolean, boolean) : void - org.apache.bookkeeper.bookie.BufferedChannel
	flush(boolean) : void - org.apache.bookkeeper.bookie.BufferedChannel
	flushCurrentLog(BufferedLogChannel, boolean) : void - org.apache.bookkeeper.bookie.EntryLogger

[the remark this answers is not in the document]
sijie: @reddycharan sorry, my question was who is using `forceMetadata = true`? typically we don't use `forceMetadata=true` (if I remember correctly)

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
reddycharan: yes, it can be static

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
```diff
@@ -2586,4 +2603,79 @@ public void setRegistrationManagerClass(
...
         return this;
     }
+
+
+    /*
+     * specifies if entryLog per ledger is enabled. If it is enabled, then there
+     * would be a active entrylog for each ledger
+     */
+    public boolean isEntryLogPerLedgerEnabled() {
+        return this.getBoolean(ENTRY_LOG_PERLEDGER_ENABLED, false);
```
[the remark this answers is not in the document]
reddycharan: valid point.. and also I was planning to validate that "USE_TRANSACTIONAL_COMPACTION" and "ENTRY_LOG_PERLEDGER_ENABLED" are not enabled together. When ENTRY_LOG_PERLEDGER_ENABLED is enabled, compaction is not required in the first place.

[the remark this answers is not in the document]
sijie: sgtm

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SyncThread.java
```diff
@@ -119,6 +119,12 @@ public void startCheckpoint(Checkpoint checkpoint) {
         });
     }
 
+    void start() {
```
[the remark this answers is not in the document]
reddycharan: since 'flushInterval' variable is already there in this class, I used it. 

Anyhow, I'll make the changes as you suggested. Also I'm planning to use "scheduleWithFixedDelay" instead of scheduleAtFixedRate, so that way it is more certain that interval between two executions is 'flushInterval'.

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
```diff
@@ -2586,4 +2603,79 @@ public void setRegistrationManagerClass(
     protected ServerConfiguration getThis() {
         return this;
     }
+
+
+    /*
+     * specifies if entryLog per ledger is enabled. If it is enabled, then there
+     * would be a active entrylog for each ledger
```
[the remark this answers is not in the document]
reddycharan: will do

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryMemTable.java
```diff
@@ -263,6 +285,81 @@ private long flushSnapshot(final SkipListFlusher flusher, Checkpoint checkpoint)
...
+                    try {
+                        while (!latch.await(memtableFlushTimeoutInSeconds, TimeUnit.SECONDS)) {
+                            logger.error("Entrymemtable parallel flush has not completed in {0} secs, so waiting again",
+                                    memtableFlushTimeoutInSeconds);
+                        }
+                        assert (exceptionWhileFlushingParallelly.get() == null);
+                        flushBytesCounter.add(flushedSize.get());
+                        clearSnapshot(keyValues);
+                    } catch (InterruptedException ie) {
+                        logger.error("Got Interrupted exception while waiting for the flushexecutor "
```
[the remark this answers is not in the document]
reddycharan: will do

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryMemTable.java
```diff
@@ -263,6 +285,81 @@ private long flushSnapshot(final SkipListFlusher flusher, Checkpoint checkpoint)
...
+                            }
+                        });
+                    }
+
+                    try {
+                        while (!latch.await(memtableFlushTimeoutInSeconds, TimeUnit.SECONDS)) {
+                            logger.error("Entrymemtable parallel flush has not completed in {0} secs, so waiting again",
+                                    memtableFlushTimeoutInSeconds);
+                        }
+                        assert (exceptionWhileFlushingParallelly.get() == null);
```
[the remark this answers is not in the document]
reddycharan: assert statements are used in four other places in this class. For eg: - https://github.com/apache/bookkeeper/blob/cff50ee31feadf63bad1ccc1ccc3abcdf0fea79d/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryMemTable.java#L273

thought it should be ok to use assert statements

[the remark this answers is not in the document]
sijie: apache/bookkeeper#995 is for cleaning up the usage of assert

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryMemTable.java
```diff
@@ -263,6 +285,81 @@ private long flushSnapshot(final SkipListFlusher flusher, Checkpoint checkpoint)
...
+                                    }
+                                }
+                                latch.countDown();
+                            }
+                        });
+                    }
+
+                    try {
+                        while (!latch.await(memtableFlushTimeoutInSeconds, TimeUnit.SECONDS)) {
+                            logger.error("Entrymemtable parallel flush has not completed in {0} secs, so waiting again",
```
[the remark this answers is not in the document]
reddycharan: no it doesn't, that was mistake. will fix that

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryMemTable.java
```diff
@@ -263,6 +285,81 @@ private long flushSnapshot(final SkipListFlusher flusher, Checkpoint checkpoint)
...
+                                        return;
+                                    }
+                                }
+                                latch.countDown();
+                            }
+                        });
+                    }
+
+                    try {
+                        while (!latch.await(memtableFlushTimeoutInSeconds, TimeUnit.SECONDS)) {
```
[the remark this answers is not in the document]
reddycharan: actually, here I'm not doing much with await logic. Initially i wanted the timeout logic, but based on our internal feedback i toned down the functionality of it and just used it for logging purpose. It is just used for logging purpose, saying that flush didn't complete in "memtableFlushTimeoutInSeconds" period and hence waiting again..

Anyway, if you want me to remove the config all together, I can remove it and use some static value variable (120 secs) here. But I feel it would be helpful to log this line if for any reason it didn't complete the flush in memtableFlushTimeoutInSeconds period before going back to wait logic.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryMemTable.java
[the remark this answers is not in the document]
reddycharan: ok will remove the assert line.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryMemTable.java
```diff
@@ -263,6 +285,81 @@ private long flushSnapshot(final SkipListFlusher flusher, Checkpoint checkpoint)
...
+                            @Override
+                            public void safeRun() {
+                                for (EntryKeyValue entryKeyValue : entryKeyValuesOfALedger) {
+                                    try {
+                                        flusher.process(ledgerId, entryKeyValue.getEntryId(),
+                                                entryKeyValue.getValueAsByteBuffer());
+                                        flushedSize.addAndGet(entryKeyValue.getLength());
+                                    } catch (NoLedgerException exception) {
+                                        logger.info("Got NoLedgerException while flushing entry: {}. The ledger "
+                                                + "must be deleted " + "after this entry is added to the Memtable",
```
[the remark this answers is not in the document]
reddycharan: will do

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryMemTable.java
```diff
@@ -263,6 +285,81 @@ private long flushSnapshot(final SkipListFlusher flusher, Checkpoint checkpoint)
...
+                                                entryKeyValue);
+                                        break;
+                                    } catch (Exception exc) {
+                                        logger.error(
+                                                "Got Exception while trying to flush process entry: " + entryKeyValue,
```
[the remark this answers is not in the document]
reddycharan: will do

bookkeeper-server/conf/bk_server.conf
```diff
@@ -620,3 +620,15 @@ zkEnableSecurity=false
 # dbStorage_rocksDB_numLevels=-1
 # dbStorage_rocksDB_numFilesInLevel0=4
 # dbStorage_rocksDB_maxSizeInLevel1MB=256
+
+# config enabling/disabling entrylog per ledger feature.
+# entryLogPerLedgerEnabled=false
```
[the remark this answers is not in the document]
reddycharan: will do

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryMemTable.java
```diff
@@ -263,6 +285,81 @@ private long flushSnapshot(final SkipListFlusher flusher, Checkpoint checkpoint)
...
+                                                exc);
+                                        if (isFlushThreadInterrupted.compareAndSet(false, true)) {
+                                            exceptionWhileFlushingParallelly.set(exc);
```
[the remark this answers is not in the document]
reddycharan: OrderedSafeExecutor doesn't provides cancel API. Also, I don't think it as an issue if we don't cancel the pending writes. This flush operation is considered failure and it will be tried. Even if we try to cancel/remove the pending tasks there isn't any guarantee that it would be cancelled/removed.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
reddycharan: will do

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
reddycharan: yes it can be

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
reddycharan: will do

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
reddycharan: will do

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
reddycharan: will do

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
reddycharan: here we are not removing all the entries whose value is true. We are removing just the entries which has true value untill the first false value. But yeah, probably we can use streams in some form, for simplicity sake lets keep as it is. May be I should add a comment describing what I'm intending to do here.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
reddycharan: yes, it can be moved to BufferedLogChannel

reddycharan:
@eolivelli I just want to have option for reviewers to see the full picture. Im ok to change the PR title. any suggestions?

eolivelli:
Something like a prefix DRAFT or DO NOT MERGE

/cc @sijie

jvrao:
> Something like a prefix DRAFT or DO NOT MERGE
@reddycharan as I suggested RFC (Request For Comments) may convey the message.

sijie:
@reddycharan since most of the changes are already in in separated PRs, shall I close this one?

reddycharan:
first five sub-tasks of this Issue are already merged, the last one is in code review stage - https://github.com/apache/bookkeeper/pull/1391 . So this WIP PR is not needed anymore. Will close this PR.

PR #1204, opened 2018-02-26
Allow rackaware policy to be notified of any rack change

merlimat:
This PR is a port from Yahoo branch: https://github.com/yahoo/bookkeeper/commit/dbfd2073b6e687a5b4b9c56a0f0c3436f177db58

It adds a hook so that a placement policy implementation can add a way to notify the main rackaware policy that the rack mappings were updated.

This was used to have a specialized policy that keeps the mapping (bookie -> rack) in ZooKeeper. Whenever the new bookies are added (or updated), the administrator would update the mapping and we need to start using the new settings.


sijie:
```
junit.framework.AssertionFailedError: expected:<3> but was:<2>
	at org.apache.bookkeeper.client.TestRackawarePolicyNotificationUpdates.testNotifyRackChange(TestRackawarePolicyNotificationUpdates.java:116)
```

@merlimat it seems the new test case failed at java 9 jenkins build. can you take a look?

PR #1206, opened 2018-02-26
Allow derived implementations of RackawareEnsemblePlacementPolicy

merlimat:
Leave public constructor so that it's possible to create a placement policy that derives from RackawareEnsemblePlacementPolicy.

For example, in Pulsar we have an isolation policy that works in conjunction with the rack-aware: https://github.com/apache/incubator-pulsar/blob/ac94698df6c66c4b84faa9f96db71f0019bf955e/pulsar-zookeeper-utils/src/main/java/org/apache/pulsar/zookeeper/ZkIsolatedBookieEnsemblePlacementPolicy.java#L45


eolivelli:
Can we add some javadoc to state that this class is supposed to be extended?

sijie:
> Can we add some javadoc to state that this class is supposed to be extended?

we are not making this class public. It should be falling into the category of "PrivateLimited". so no javadoc should be public for this class.

merlimat:
> Can we add some javadoc to state that this class is supposed to be extended? 

This class is not necessarily supposed to be extended: it already provides rack-aware feature out of the box. In our use case we just wanted to add more functionality to support bookie isolation in different groups.

The class is already public, though the constructors are package-protected. I don't think that it changes a lot on what it's considered public API. In many other implementation classes we have public constructors.

merlimat:
Merging for now to start moving Pulsar to bk 4.7 snapshots. We can discuss the issue more separately.

PR #1209, opened 2018-02-27
Added MockBookKeeper for unit tests

merlimat:
This mock class is what we use in Pulsar to do unit tests for the managed-ledger library. 

This approach might be useful in general, in all cases when creating unit tests of an application that uses the BookKeeper API. 

The `MockBookKeeper` behaves like the regular BK client, except it needs no bookies or zookkeeper. All ledgers and data are just kept in an hashmap in memory.

It is possible to inject failure in the sync/async calls, even after N steps. This is very useful to trigger failure in complex operations and cover this paths in unit tests.

### DIFF
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/MockBookKeeper.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/MockBookKeeper.java
@@ -1,0 +1,280 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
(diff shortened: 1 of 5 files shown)


bookkeeper-server/src/test/java/org/apache/bookkeeper/client/MockBookKeeper.java
```diff
@@ -0,0 +1,280 @@
...
+    public ZooKeeper getZkHandle() {
+        return zkc;
+    }
+
+    @Override
+    public ClientConfiguration getConf() {
+        return super.getConf();
+    }
+
+    Map<Long, MockLedgerHandle> ledgers = new ConcurrentHashMap<Long, MockLedgerHandle>();
```
sijie: nit: line 62-65 can be final

sijie:
the failed "java 9" jenkins job is unrelated to this change.

sijie:
IGNORE CI

PR #1217, opened 2018-02-28
 Shade bookkeeper in distributedlog-core-shaded and allow using shaded bookkeeper client to resolve ledger manager factory class

sijie:

*Motivation*

1. when `distributedlog-core-shaded` is co-used with other bookkeeper jars, it would cause class conflicts because `distributedlog-core-shaded` shades
most of its dependencies except bookkeeper. so it is better to also shade bookkeeper to avoid such conflicts.

2. if we shade bookkeeper, it would cause a problem on resolving ledger manager factory class stored on layout metadata.

```
java.io.IOException: Failed to instantiate ledger manager factory org.apache.bookkeeper.meta.HierarchicalLedgerManagerFactory
	at dlshade.org.apache.bookkeeper.meta.AbstractZkLedgerManagerFactory.attemptToResolveShadedLedgerManagerFactory(AbstractZkLedgerManagerFactory.java:256) ~[distributedlog-core-shaded-4.7.0-SNAPSHOT.jar:4.7.0-SNAPSHOT]
	at dlshade.org.apache.bookkeeper.meta.AbstractZkLedgerManagerFactory.newLedgerManagerFactory(AbstractZkLedgerManagerFactory.java:203) ~[distributedlog-core-shaded-4.7.0-SNAPSHOT.jar:4.7.0-SNAPSHOT]
```

*Solution*

- shade bookkeeper dependencies in `distributedlog-core-shaded`
- add flags to allow using shaded bookkeeper jar to resolve ledger manager factory

> ---
> quickly and easily:
>
>
>
> Otherwise:
> 
> 
> ---

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
@@ -332,7 +332,7 @@ public abstract class AbstractConfiguration<T extends AbstractConfiguration>
      * @return ledger manager factory class name.
      */
     public boolean isShadedLedgerManagerFactoryClassAllowed() {
-        return getBoolean(ALLOW_SHADED_LEDGER_MANAGER_FACTORY_CLASS);
+        return getBoolean(ALLOW_SHADED_LEDGER_MANAGER_FACTORY_CLASS, false);
     }
 
     /**
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
@@ -73,6 +73,8 @@ public abstract class AbstractConfiguration<T extends AbstractConfiguration>
     // Ledger Manager
     protected static final String LEDGER_MANAGER_TYPE = "ledgerManagerType";
     protected static final String LEDGER_MANAGER_FACTORY_CLASS = "ledgerManagerFactoryClass";
+    protected static final String ALLOW_SHADED_LEDGER_MANAGER_FACTORY_CLASS = "allowShadedLedgerManagerFactoryClass";
+    protected static final String SHADED_LEDGER_MANAGER_FACTORY_CLASS_PREFIX = "shadedLedgerManagerFactoryClassPrefix";
     protected static final String METADATA_SERVICE_URI = "metadataServiceUri";
     protected static final String ZK_LEDGERS_ROOT_PATH = "zkLedgersRootPath";
     protected static final String ZK_REQUEST_RATE_LIMIT = "zkRequestRateLimit";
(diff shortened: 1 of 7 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
```diff
@@ -310,6 +312,60 @@ public String getLedgerManagerType() {
...
 
+    /**
+     * Set the flag to allow using shaded ledger manager factory class for
+     * instantiating a ledger manager factory.
+     *
+     * @param allowed
+     *          the flag to allow/disallow using shaded ledger manager factory class
+     * @return configuration instance.
+     */
+    public T setAllowShadedLedgerManagerFactoryClass(boolean allowed) {
```
ivankelly: I don't think this is needed. If you set "ShadedLedgerManagerFactoryClassPrefix" it implies you want to allow it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractZkLedgerManagerFactory.java
```diff
@@ -194,15 +200,78 @@ public static LedgerManagerFactory newLedgerManagerFactory(
...
+                return factoryClass.replace(shadedClassPrefix, "");
+            }
+        } else {
+            return factoryClass;
+        }
+    }
+
+    private static boolean isSameLedgerManagerFactory(
+            AbstractConfiguration<?> conf, String leftFactoryClass, String rightFactoryClass) {
+        leftFactoryClass = normalizedLedgerManagerFactoryClassName(
```
ivankelly:
```
return leftFactoryClass.endsWith(rightFactoryClass) || rightFactoryClass.endsWith(leftFactoryClass)
```

sijie: you can't do simply do that, because left or right can be null and it needs to take the flag into account.

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
sijie: would prefer having an explicit flag on controlling this

sijie:
> You should add a note in the PR comment that it's needed because you may load the class name from the zookeeper server metadata.

updated

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
ivankelly: Sure. My issue with it is that it exposes it in a public api, so it's hard to change later. Not super strongly opposed though.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractZkLedgerManagerFactory.java
ivankelly: sure.

ivankelly:
I think this breaks a unit test.

sijie:
Fixed broken unit test.

PR #1218, opened 2018-02-28
Support LacPiggyback, LongPoll and ExplicitLac in db ledger storage

sijie:

*Motivation*

DbLedgerStorage was developed in yahoo and just merged recently to master. This implementation only works for v2 protocol.
It doesn't work with v3 protocol. When using v3 protocol clients (e.g. dlog library) to talk to bookies use DbLedgerStorage,
the clients hang on waiting responses. because a few methods in DbLedgerStorage throw UnsupportedException

*Solution*

This change focuses on adding the implementations for those methods, in order to support LacPiggyback, LongPoll and ExplicitLac
for v3 protocol. The implementation follows what we had in FileInfoCache but did it in a much simpler way, since all the persistence
information has been managed by db ledger storage itself. All the information required by `LacPiggyback`, `LongPoll` and `ExplicitLac`
are transient.

This change doesn't introduce any new test cases. It reuses existing test cases to test those features using DbLedgerStorage. These
tests are:

- ExplicitLacTest: for testing explicit lac
- TestPiggybackLAC: for testing piggback lac
- TestReadLastConfirmedLongPoll: for testing long polling lac
- TestReadLastConfirmedAndEntry: for testing entry piggyback


sijie:
NOTE: in this pull requests reused a lot of existing test cases to test those features with DbLedgerStorage. 

Created #1219 for tracking how these test cases can be improved.

bookkeeper-server/src/test/java/org/apache/bookkeeper/util/TestUtils.java
```diff
@@ -52,4 +60,21 @@ public static boolean hasLogFiles(File ledgerDirectory, boolean partial, Integer
...
+
+    public static void waitUtilLacUpdated(ReadHandle rh, long newLac) throws Exception {
+        long lac = rh.getLastAddConfirmed();
+        while (lac < newLac) {
+            TimeUnit.MILLISECONDS.sleep(20);
+            lac = rh.readLastAddConfirmed().get();
+        }
+    }
+
+    public static long waitUtilExplicitLacUpdated(LedgerHandle rh, long newLac) throws Exception {
```
ivankelly: typo: waitUntil

bookkeeper-server/src/test/java/org/apache/bookkeeper/util/TestUtils.java
```diff
@@ -52,4 +60,21 @@ public static boolean hasLogFiles(File ledgerDirectory, boolean partial, Integer
         }
         return result;
     }
+
+    public static void waitUtilLacUpdated(ReadHandle rh, long newLac) throws Exception {
```
ivankelly: typo: waitUntil

bookkeeper-server/src/test/java/org/apache/bookkeeper/util/TestUtils.java
ivankelly: Why are you using ReadHandle for one and LedgerHandle for the other?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
```diff
@@ -74,10 +83,121 @@
...
+     * This class borrows the logic from FileInfo.
+     *
+     * <p>This class is used for holding all the transient states for a given ledger.
+     */
+    private static class TransientLedgerInfo
+            extends Watchable<LastAddConfirmedUpdateNotification>
+            implements AutoCloseable {
+
+        // lac
+        private Long lac = null;
```
ivankelly: use LedgerHandle.INVALID_ENTRY_ID and you'll avoid the autoboxing and all the null checks.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
```diff
@@ -74,10 +83,121 @@
...
+        @Override
+        public void close() {
+           synchronized (this) {
+               if (isClosed) {
+                   return;
+               }
+               isClosed = true;
+           }
+           // notify watchers
+            notifyWatchers(Long.MAX_VALUE);
```
ivankelly: watchers are being told that the lac for the ledger is MAX_VALUE? Seems like this will violate TOAB.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
```diff
@@ -167,12 +287,41 @@ public void initialize(ServerConfiguration conf, LedgerManager ledgerManager, Le
...
+        RemovalListener<Long, TransientLedgerInfo> ledgerInfoRemovalListener = this::handleLedgerEviction;
+        CacheBuilder<Long, TransientLedgerInfo> builder = CacheBuilder.newBuilder()
+            .initialCapacity(conf.getFileInfoCacheInitialCapacity())
+            .maximumSize(conf.getOpenFileLimit())
+            .concurrencyLevel(concurrencyLevel)
+            .removalListener(ledgerInfoRemovalListener);
+        if (conf.getFileInfoMaxIdleTime() > 0) {
+            builder.expireAfterAccess(conf.getFileInfoMaxIdleTime(), TimeUnit.SECONDS);
+        }
+        transientLedgerInfoCache = builder.build(new CacheLoader<Long, TransientLedgerInfo>() {
```
ivankelly: Can you a closure here to be more compact.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
```diff
@@ -167,12 +287,41 @@ public void initialize(ServerConfiguration conf, LedgerManager ledgerManager, Le
         ledgerIndex = new LedgerMetadataIndex(conf, KeyValueStorageRocksDB.factory, baseDir, stats);
         entryLocationIndex = new EntryLocationIndex(conf, KeyValueStorageRocksDB.factory, baseDir, stats);
 
+        // build the ledger info cache
+        int concurrencyLevel = Math.max(1, Math.max(conf.getNumAddWorkerThreads(), conf.getNumReadWorkerThreads()));
+        RemovalListener<Long, TransientLedgerInfo> ledgerInfoRemovalListener = this::handleLedgerEviction;
```
ivankelly: why not pass the method directly to .removalListener()?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
```diff
@@ -167,12 +287,41 @@ public void initialize(ServerConfiguration conf, LedgerManager ledgerManager, Le
...
         gcThread = new GarbageCollectorThread(conf, ledgerManager, this, statsLogger);
 
         registerStats();
     }
 
+    /**
+     * When a ledger is evicted from transient ledger info cache, we can just simply discard the object.
+     */
+    private void handleLedgerEviction(RemovalNotification<Long, TransientLedgerInfo> notification) {
+        TransientLedgerInfo fileInfo = notification.getValue();
```
ivankelly: rename fileInfo to ledgerInfo

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
```diff
@@ -712,23 +874,38 @@ public EntryLogger getEntryLogger() {
 
     @Override
     public long getLastAddConfirmed(long ledgerId) throws IOException {
-        throw new UnsupportedOperationException();
+        TransientLedgerInfo ledgerInfo = transientLedgerInfoCache.getIfPresent(ledgerId);
+        Long lac = null != ledgerInfo ? ledgerInfo.getLastAddConfirmed() : null;
+        if (null == lac) {
+            ByteBuf bb = getEntry(ledgerId, BookieProtocol.LAST_ADD_CONFIRMED);
```
ivankelly: It'd be better to do this in the loader, and not rely on a getLastAddConfirmed coming before getExplicitLAC or any of the other calls. If it's in the loader, it becomes easier to reason about the state. This method becomes 
```
return transientLedgerInfoCache.getUnchecked(ledgerId).getLastAddConfirmed();
```
No checking, because we can assume ledger info cache is up to date.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
ivankelly: A cache is by definition transient. I'd remove transient from the name.

sijie:
> I miss one piece of this story...when you reboot the bookie this info is lost, is this true?

> For long poll is not problem but for explicitlac it will be a problem, no?

we don't persist any of this because we don't actually need to.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
sijie: null has a different meaning than INVALID_ENTRY_ID, when it is null, we need to read the last entry to examine the lac.

even we can work this around to use long, would still prefer keeping this consistent with the implementation in SortedLedgerStorage, which is safer.

bookkeeper-server/src/test/java/org/apache/bookkeeper/util/TestUtils.java
sijie: I would prefer using new API whenever I can. readExplicitLastAddConfirmed is not available in the new API, that forces me going back to LedgerHandle.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
sijie: I can't. because it is not an interface.  I've used closures whenever I can and my idea will ask me to do that. If I don't do it, that means it can't be.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
sijie: because you can't. you need to do assignment for type inference

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
sijie: no. I have kept the cache as simple and the loading will never throw any exceptions. If I move the getEntry as part of loader, it complicates error handling in cache loading, which I don't want to.

ivankelly:
@sijie @eolivelli It looks like the explicit LAC info is 100% transient in this case. I would suggest leaving out explicit LAC until we have persistence for it in DB storage.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
sijie: this is not true.

1) in FileInfoCache, FileInfo is not trasient, it deals with persistence. However in here, TransientLedgerInfo doesn't deal with persistence. that makes the difference. 
2) I make it explicit "Transient", to make the difference with FileInfo and also `LedgerData` in LedgerMetadataIndex. Otherwise, "LedgerInfo" and "LedgerData" sound too close to each other.

sijie:
> But On other storage types we are persisting ExplicitLAC, why don't we need it in this case. My question is just to understand clearly

we don't persist any `ExplicitLac` at any type of ledger storage, because we don't need to. That is by the design when this feature was introduced.

@ivankelly 

we don't need to add any persistence for it.

sijie:
@ivankelly fixed typos and replied to your comments.

eolivelli:
@sijie okay, thank you. clear to me

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
sijie: we don't use the value notified as the source-of-truth. the value is used for trigger the logic when the lac is changed. we re-read lac and put that value in the response. we did this because LAC can potentially be updated between the watcher received notification and sending the response.  re-read lac would give the most recent value, so the client can have as many entries to read before hitting lac again.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
ivankelly: it does work, I've just tried it. Only a minor comment though, not too worried about it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
ivankelly: The whole problem with file info cache is that we are using something that should be transient as non-transient. That's why we need the backing cache, which isn't really a cache. Anyhow, was only a suggestion.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
ivankelly: fair enough.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
ivankelly: null only has a different meaning because you're loading uninitialized entries into the cache. I'd do it differently, but won't block on this.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
ivankelly: ok. seems a bit ropey, since MAX_VALUE is a valid value for lac in theory, and someone could try to add an optimization in future and end up sending it to the client, but that's not for this patch to deal with anyhow.

sijie: "MAX_VALUE" can't really be a valid value for lac. what would be the entry id to carry "MAX_VALUE" lac?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
sijie: that means you need to refactor the whole logic, not just ledger storage, which I will hesitate to do it.

sijie:
> No idea why it's persisted in other storage impls.

I have explained in the comments, it is not persisted in any storage impls.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
ivankelly: not valid read from storage, no. But a watcher could receive a positive long from the Watchable, and conceivably consider it valid.

ivankelly:
I see. I assumed since it was in *File*Info, it was persisted.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
sijie: > But a watcher could receive a positive long from the Watchable, and conceivably consider it valid.

yes the only case for Long.MAX_VALUE is from watchers. Then the meaning given for Long.MAX_VALUE is the watcher logic. At this case, Long.MAX_VALUE would then be effectively same as using some negative values.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
```diff
@@ -712,23 +874,38 @@ public EntryLogger getEntryLogger() {
...
+            bb.skipBytes(2 * Long.BYTES); // skip ledger id and entry id
+            lac = bb.readLong();
```
merlimat: `bb` should be released at this point

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
ivankelly: Which means the watcher will need to know that MAX_VALUE is a special value. It's a implicit contract.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
```diff
@@ -74,10 +83,121 @@
...
+            super(WATCHER_RECYCLER);
+            this.ledgerId = ledgerId;
+            this.ledgerIndex = ledgerIndex;
+        }
+
+        synchronized Long getLastAddConfirmed() {
+            return lac;
+        }
+
+        Long setLastAddConfirmed(long lac) {
```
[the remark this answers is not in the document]
sijie: @merlimat the whole pipeline about lac is expecting Long. even we return `long` here, we need to boxing back to Long. changing it not only involve changes in DbLedgerStorage but also involves all the request/response processing logic. so keep it as consistent as the other ledger storage impls.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
sijie: yes. I was actually fixing that for InterleavedLedgerStorage as well. not sure why it didn't get in this PR. pushing it.

sijie:
@merlimat fixed unit test failures, since `DBledgerStorageTest` doesn't put lac on the entries. addressed the bytebuf ref leaking issue on `getLastAddConfirmed` on all the storage impls.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
sijie: yes that's the contract.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/InterleavedLedgerStorage.java
```diff
@@ -242,10 +242,13 @@ public long getLastAddConfirmed(long ledgerId) throws IOException {
...
-                bb.readLong(); // ledger id
-                bb.readLong(); // entry id
-                lac = bb.readLong();
-                lac = ledgerCache.updateLastAddConfirmed(ledgerId, lac);
+                try {
+                    bb.skipBytes(2 * Long.BYTES); // skip ledger & entry id
+                    lac = bb.readLong();
+                    lac = ledgerCache.updateLastAddConfirmed(ledgerId, lac);
+                } finally {
+                    bb.release();
```
jvrao: Leak in existing code?

sijie: yeah seem so

PR #1221, opened 2018-02-28
Refactored ReadResponse ref count handling

ivankelly:
This contains a number of changes.

- V2 bookie protocol
  - Add retain and release methods to all responses. For read response
    it handles the data buffer.
  - ReadResponses always have a buffer now, even if empty.

- Server side
  - In the v2 read handler, releasing of the buffer in the case of
    error is left to the very end.

- Client side
  - Per channel bookie clients own the buffer for read responses.
    If a ReadCallback want it to live past the lifetime of the call
    it must call retain.

This change was originally e8643140 in the yahoo-4.3 branch.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
@@ -23,6 +23,7 @@ package org.apache.bookkeeper.client;
 import com.google.common.collect.Lists;
 import com.google.common.util.concurrent.ListenableFuture;
 import io.netty.buffer.ByteBuf;
+
 import java.util.ArrayList;
 import java.util.BitSet;
 import java.util.HashSet;
@@ -139,7 +140,6 @@ class PendingReadOp implements ReadEntryCallback, SafeRunnable {
             } catch (BKDigestMatchException e) {
                 readOpDmCounter.inc();
                 logErrorAndReattemptRead(bookieIndex, host, "Mac mismatch", BKException.Code.DigestMatchException);
-                buffer.release();
                 return false;
             }
 
@@ -154,7 +154,6 @@ class PendingReadOp implements ReadEntryCallback, SafeRunnable {
                 writeSet.recycle();
                 return true;
             } else {
-                buffer.release();
                 return false;
             }
         }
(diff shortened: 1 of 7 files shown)


ivankelly:
@merlimat can you check my understanding of the change matches your original intent?

sijie:
retest this please

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
```diff
@@ -23,6 +23,7 @@
 import com.google.common.collect.Lists;
 import com.google.common.util.concurrent.ListenableFuture;
 import io.netty.buffer.ByteBuf;
+
```
jvrao: Nit: any reason for a new line here?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
```diff
@@ -589,12 +588,15 @@ public void readEntryComplete(int rc, long ledgerId, final long entryId, final B
         heardFromHosts.add(rctx.to);
         heardFromHostsBitSet.set(rctx.bookieIndex, true);
 
+        buffer.retain();
```
jvrao: This is additional retain you added here. Who's responsibility is to release this buffer?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
ivankelly: probably came in in cherry-pick. will remove.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
ivankelly: We pass it into entry#complete, if entry#complete accepts the buffer, it gets passed to entryImpl.setEntryBuf, which takes ownership, and entry#complete returns true, otherwise it returns false, and it gets cleaned up in the else branch. entryImpls either get passed to the client (which is then responsible for cleanup), or cleaned up in submitCallback L637.

ivankelly:
merging this

PR #1223, opened 2018-02-28
Enable buiding website on travis job if a PR is modifying `site`

sijie:

PR #1224, opened 2018-03-02
Improve Travis Build

sijie:

*Problem*

There are a couple of problem with current travis build

- apache/bookkeeper#1223 introduce `set -e` on script section, which it cause some weird early-exiting problem.
- we run two installs. one is the default install by travis, the other one is install with apache-rat/spotbug checks in script section, which make the build slow.
- travis_retry and travis_wait spread across different commands in the script section

*Solution*

- move build logic into a bash script file and invoke the script
- skip install section in travis

*Result*

This reduces travis build time by 10+ mins and make it stable

NOTES:

The changes under `site` and `stream/distributedlog` are used `empty commits` to trigger corresponding builds in travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ matrix:
       dist: trusty
       env: CUSTOM_JDK="openjdk8"
 
-before_install:
+before_install: 
 - |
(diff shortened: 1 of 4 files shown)


sijie:
The travis CI passed. the failure is on jenkins integration tests job (timeout at building docker image)

sijie:
IGNORE IT CI

PR #1225, opened 2018-03-02
Issue #570: getting rid of unnecessary synchronization in InterleavedLedgerStorage

reddycharan:

This is < sub-task1 > of Issue #570.

In InterleavedLedgerStorage, since the initial version of it, addEntry and processEntry methods are synchronized, but they are not required to be synchronized. 

Removal of unnecessary synchronization in InterleavedLedgerStorage methods, as described in http://mail-archives.apache.org/mod_mbox/bookkeeper-dev/201707.mbox/%3CCAO2yDyZ946fp2S_qR2iL178hPiXgrnFGb%3DpvkyK4ReSYAtNLBw%40mail.gmail.com%3E

Master Issue: #570

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/InterleavedLedgerStorage.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/InterleavedLedgerStorage.java
@@ -76,7 +76,7 @@ public class InterleavedLedgerStorage implements CompactableLedgerStorage, Entry
     GarbageCollectorThread gcThread;
 
     // this indicates that a write has happened since the last flush
-    private AtomicBoolean somethingWritten = new AtomicBoolean(false);
+    private final AtomicBoolean somethingWritten = new AtomicBoolean(false);
 
     // Expose Stats
     private OpStatsLogger getOffsetStats;
@@ -362,10 +362,9 @@ public class InterleavedLedgerStorage implements CompactableLedgerStorage, Entry
 
     @Override
     public synchronized void flush() throws IOException {
-        if (!somethingWritten.get()) {
+        if (!somethingWritten.compareAndSet(true, false)) {
             return;
         }
-        somethingWritten.set(false);
         flushOrCheckpoint(false);
     }
 
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/InterleavedLedgerStorage.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/InterleavedLedgerStorage.java
@@ -34,6 +34,8 @@ import java.util.Map;
 import java.util.NavigableMap;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+
 import org.apache.bookkeeper.bookie.Bookie.NoLedgerException;
 import org.apache.bookkeeper.bookie.CheckpointSource.Checkpoint;
 import org.apache.bookkeeper.bookie.EntryLogger.EntryLogListener;
@@ -74,7 +76,7 @@ public class InterleavedLedgerStorage implements CompactableLedgerStorage, Entry
     GarbageCollectorThread gcThread;
 
     // this indicates that a write has happened since the last flush
-    private volatile boolean somethingWritten = false;
+    private AtomicBoolean somethingWritten = new AtomicBoolean(false);
 
     // Expose Stats
     private OpStatsLogger getOffsetStats;
@@ -264,7 +266,7 @@ public class InterleavedLedgerStorage implements CompactableLedgerStorage, Entry
 
 
     @Override
-    public synchronized long addEntry(ByteBuf entry) throws IOException {
+    public long addEntry(ByteBuf entry) throws IOException {
         long ledgerId = entry.getLong(entry.readerIndex() + 0);
         long entryId = entry.getLong(entry.readerIndex() + 8);
         long lac = entry.getLong(entry.readerIndex() + 16);
@@ -360,10 +362,10 @@ public class InterleavedLedgerStorage implements CompactableLedgerStorage, Entry
 
     @Override
     public synchronized void flush() throws IOException {
-        if (!somethingWritten) {
+        if (!somethingWritten.get()) {
             return;
         }
-        somethingWritten = false;
+        somethingWritten.set(false);
         flushOrCheckpoint(false);
     }
 
@@ -419,12 +421,12 @@ public class InterleavedLedgerStorage implements CompactableLedgerStorage, Entry
         processEntry(ledgerId, entryId, entry, true);
     }
 
-    protected synchronized void processEntry(long ledgerId, long entryId, ByteBuf entry, boolean rollLog)
+    protected void processEntry(long ledgerId, long entryId, ByteBuf entry, boolean rollLog)
             throws IOException {
         /*
          * Touch dirty flag
          */
-        somethingWritten = true;
+        somethingWritten.set(true);
 
         /*
          * Log the entry
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
@@ -111,7 +111,7 @@ public class EntryLogTest {
         assertTrue(meta.getLedgersMap().containsKey(3L));
     }
 
-    private ByteBuf generateEntry(long ledger, long entry) {
+    private static ByteBuf generateEntry(long ledger, long entry) {
         byte[] data = generateDataString(ledger, entry).getBytes();
         ByteBuf bb = Unpooled.buffer(8 + 8 + data.length);
         bb.writeLong(ledger);
@@ -120,8 +120,8 @@ public class EntryLogTest {
         return bb;
     }
 
-    private String generateDataString(long ledger, long entry) {
-        return new String("ledger-" + ledger + "-" + entry);
+    private static String generateDataString(long ledger, long entry) {
+        return ("ledger-" + ledger + "-" + entry);
     }
 
     @Test
@@ -401,7 +401,7 @@ public class EntryLogTest {
         assertEquals(Sets.newHashSet(0L, 1L, 2L, 3L), logger.getEntryLogsSet());
     }
 
-    class LedgerStorageWriteTask implements Callable<Boolean> {
+    static class LedgerStorageWriteTask implements Callable<Boolean> {
         long ledgerId;
         int entryId;
         LedgerStorage ledgerStorage;
@@ -424,7 +424,7 @@ public class EntryLogTest {
         }
     }
 
-    class LedgerStorageReadTask implements Callable<Boolean> {
+    static class LedgerStorageReadTask implements Callable<Boolean> {
         long ledgerId;
         int entryId;
         LedgerStorage ledgerStorage;
@@ -482,13 +482,13 @@ public class EntryLogTest {
         InterleavedLedgerStorage ledgerStorage = ((InterleavedLedgerStorage) bookie.ledgerStorage);
 
         int numOfLedgers = 70;
-        int numEntries = 3000;
+        int numEntries = 2000;
         // Create ledgers
         for (int i = 0; i < numOfLedgers; i++) {
             ledgerStorage.setMasterKey(i, "key".getBytes());
         }
 
-        ExecutorService executor = Executors.newFixedThreadPool(40);
+        ExecutorService executor = Executors.newFixedThreadPool(10);
         List<LedgerStorageWriteTask> writeTasks = new ArrayList<LedgerStorageWriteTask>();
         for (int j = 0; j < numEntries; j++) {
             for (int i = 0; i < numOfLedgers; i++) {
@@ -504,11 +504,10 @@ public class EntryLogTest {
             LedgerStorageWriteTask writeTask = writeTasks.get(i);
             long ledgerId = writeTask.ledgerId;
             int entryId = writeTask.entryId;
-            Assert.assertTrue(
-                    "WriteTask should have been completed successfully ledgerId: " + ledgerId + " entryId: " + entryId,
-                    future.isDone() && (!future.isCancelled()));
-            Assert.assertTrue("Position for ledgerId: " + ledgerId + " entryId: " + entryId + " should have been set",
-                    future.get());
+            Assert.assertTrue("WriteTask should have been completed successfully, but it is cancelled. ledgerId: "
+                    + ledgerId + " entryId: " + entryId, !future.isCancelled());
+            Assert.assertTrue("WriteTask should have been completed successfully. ledgerId: " + ledgerId + " entryId: "
+                    + entryId, future.get());
 
         }
 
@@ -526,9 +525,8 @@ public class EntryLogTest {
             Future<Boolean> future = readTasksFutures.get(i);
             long ledgerId = readTasks.get(i).ledgerId;
             int entryId = readTasks.get(i).entryId;
-            Assert.assertTrue(
-                    "ReadTask should have been completed successfully ledgerId: " + ledgerId + " entryId: " + entryId,
-                    future.isDone() && (!future.isCancelled()));
+            Assert.assertTrue("ReadTask should have been completed successfully, but it is cancelled. ledgerId: "
+                    + ledgerId + " entryId: " + entryId, (!future.isCancelled()));
             Assert.assertTrue("ReadEntry of ledgerId: " + ledgerId + " entryId: " + entryId
                     + " should have been completed successfully", future.get());
         }
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
@@ -38,6 +38,11 @@ import java.io.IOException;
 import java.io.RandomAccessFile;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
 
 import org.apache.bookkeeper.conf.ServerConfiguration;
 import org.apache.bookkeeper.conf.TestBKConfiguration;
@@ -107,12 +112,16 @@ public class EntryLogTest {
     }
 
     private ByteBuf generateEntry(long ledger, long entry) {
-        byte[] data = ("ledger-" + ledger + "-" + entry).getBytes();
+        byte[] data = generateDataString(ledger, entry).getBytes();
         ByteBuf bb = Unpooled.buffer(8 + 8 + data.length);
         bb.writeLong(ledger);
         bb.writeLong(entry);
         bb.writeBytes(data);
         return bb;
+    }
+
+    private String generateDataString(long ledger, long entry) {
+        return new String("ledger-" + ledger + "-" + entry);
     }
 
     @Test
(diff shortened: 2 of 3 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/InterleavedLedgerStorage.java
```diff
@@ -360,10 +362,10 @@ public void checkpoint(Checkpoint checkpoint) throws IOException {
 
     @Override
     public synchronized void flush() throws IOException {
-        if (!somethingWritten) {
+        if (!somethingWritten.get()) {
```
ivankelly: Use a compareAndSet here. 

Do we need synchronization on the flush method at all?

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
```diff
@@ -391,4 +400,137 @@ public void testGetEntryLogsSet() throws Exception {
...
+     * operations using InterleavedLedgerStorage.
+     */
+    @Test
+    public void testConcurrentWriteAndReadCallsOfInterleavedLedgerStorage() throws Exception {
+        File ledgerDir = createTempDir("bkTest", ".dir");
+        ServerConfiguration conf = TestBKConfiguration.newServerConfiguration();
+        conf.setJournalDirName(ledgerDir.toString());
+        conf.setLedgerDirNames(new String[] { ledgerDir.getAbsolutePath()});
+        conf.setLedgerStorageClass(InterleavedLedgerStorage.class.getName());
+        Bookie bookie = new Bookie(conf);
```
ivankelly: Do we need to construct a full bookie?

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
```diff
@@ -391,4 +400,137 @@ public void testGetEntryLogsSet() throws Exception {
...
+        // 6 seconds for them to complete
+        List<Future<Boolean>> writeTasksFutures = executor.invokeAll(writeTasks, 6, TimeUnit.SECONDS);
+        for (int i = 0; i < writeTasks.size(); i++) {
+            Future<Boolean> future = writeTasksFutures.get(i);
+            LedgerStorageWriteTask writeTask = writeTasks.get(i);
+            long ledgerId = writeTask.ledgerId;
+            int entryId = writeTask.entryId;
+            Assert.assertTrue(
+                    "WriteTask should have been completed successfully ledgerId: " + ledgerId + " entryId: " + entryId,
+                    future.isDone() && (!future.isCancelled()));
```
ivankelly: This assert is redundant with the next one, as .get() will throw an exception if cancelled, and will wait for done.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
```diff
@@ -391,4 +400,137 @@ public void testGetEntryLogsSet() throws Exception {
...
+        InterleavedLedgerStorage ledgerStorage = ((InterleavedLedgerStorage) bookie.ledgerStorage);
+
+        int numOfLedgers = 70;
+        int numEntries = 3000;
+        // Create ledgers
+        for (int i = 0; i < numOfLedgers; i++) {
+            ledgerStorage.setMasterKey(i, "key".getBytes());
+        }
+
+        ExecutorService executor = Executors.newFixedThreadPool(40);
```
ivankelly: What's the smallest machine you've run this on? If we have 40 threads spinning and they're aren't blocked on I/O, things are going to grind on laptop.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
```diff
@@ -391,4 +400,137 @@ public void testGetEntryLogsSet() throws Exception {
...
+        LedgerStorage ledgerStorage;
+
+        LedgerStorageReadTask(long ledgerId, int entryId, LedgerStorage ledgerStorage) {
+            this.ledgerId = ledgerId;
+            this.entryId = entryId;
+            this.ledgerStorage = ledgerStorage;
+        }
+
+        @Override
+        public Boolean call() {
```
ivankelly: Instead of returning false on errors, you could throw Exception, and have that carry the context of the error (ledger id, entry id etc). This would simplify the loops in the test itself, as you wouldn't have to do anything to get the ledger and entry id, as it'd already be in the exception message. In fact, the loop could be reduced to

```executor.invokeAll(writeTasks, 6, TimeUnit.SECONDS).forEach(f -> f.join())```

**Comment by ivankelly** [comment]
@reddycharan 
Could you summarize the email in PR description?

File: bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java (hunk 5 above)
Comment (reddycharan) [reply bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java]: invokeAll returns list of Futures and join is method of CompletableFuture but not Future.

File: bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java (hunk 3 above)
Comment (reddycharan) [reply bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java]: agreed, checking future.isDone is redundant here. but I would like to check "isCancelled" status before calling ".get". It would be helpful to diagnose the issue incase of failure (cancelled), if there is assert log statement instead of seeing CancellationException in the test results with no real context regarding which task has failed because of cancellation.

File: bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java (hunk 2 above)
Comment (reddycharan) [reply bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java]: probably not, but it is easier to get handle of fully constructed ledgerstorage instead of we creating instance of ledgerstorage and calling initialize explicitly with right set of arguments. Just followed how things are handled in other tests in this testsuite.

File: bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java (hunk 4 above)
Comment (reddycharan) [reply bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java]: i just ran it on my macboop pro. I was just trying some big number. But yeah 40 threads is bit much I'll bring it down to 10-15.

File: bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java (hunk 5 above)
Comment (reddycharan) [reply bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java]: other than removing redundant .isDone check, I would keep the rest of the test logic as it is now. For diagnosing in the case of failure, it is helpful to have assert failures and proper error log lines (by catching the error and log it) rather than failing with generic cancellation exception or executionexception when we call future.get().

File: bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/InterleavedLedgerStorage.java [hunk 6]
```diff
@@ -74,7 +76,7 @@
     GarbageCollectorThread gcThread;
 
     // this indicates that a write has happened since the last flush
-    private volatile boolean somethingWritten = false;
+    private AtomicBoolean somethingWritten = new AtomicBoolean(false);
```
sijie: nit : make it final

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
```diff
@@ -107,14 +112,18 @@ public void testCorruptEntryLog() throws Exception {
...
-        byte[] data = ("ledger-" + ledger + "-" + entry).getBytes();
+        byte[] data = generateDataString(ledger, entry).getBytes();
         ByteBuf bb = Unpooled.buffer(8 + 8 + data.length);
         bb.writeLong(ledger);
         bb.writeLong(entry);
         bb.writeBytes(data);
         return bb;
     }
 
+    private String generateDataString(long ledger, long entry) {
```
sijie: nit: make it static

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
```diff
@@ -107,14 +112,18 @@ public void testCorruptEntryLog() throws Exception {
...
+        return new String("ledger-" + ledger + "-" + entry);
```
sijie: nit: redundant constructor.

```
return "ledger-" + ledger + "-" + entry;
```

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
```diff
@@ -391,4 +400,137 @@ public void testGetEntryLogsSet() throws Exception {
 
         assertEquals(Sets.newHashSet(0L, 1L, 2L, 3L), logger.getEntryLogsSet());
     }
+
+    class LedgerStorageWriteTask implements Callable<Boolean> {
```
sijie: nit: can be static if `generateDataString` is static.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
```diff
@@ -391,4 +400,137 @@ public void testGetEntryLogsSet() throws Exception {
...
+                ledgerStorage.addEntry(generateEntry(ledgerId, entryId));
+            } catch (IOException e) {
+                LOG.error("Got Exception for AddEntry call. LedgerId: " + ledgerId + " entryId: " + entryId, e);
+                return false;
+            }
+            return true;
+        }
+    }
+
+    class LedgerStorageReadTask implements Callable<Boolean> {
```
sijie: nit: it can be static.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
sijie: +1 for following other tests in this testsuite for consistency.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/InterleavedLedgerStorage.java
reddycharan: changing it to compareAndSet.

but yeah I considered removing synchronized for flush method, but it is not clear regarding the transactional guarantees of the underlying methods (methods which get called by ILS.flush). So I didn't change this, moreover it is called by requestFlush, which is called by SyncThread.shutdown and SyncThread.start, which are one time events and not related to this work item. So left unchanged.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
reddycharan: done

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
reddycharan: done

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
reddycharan: changing

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
reddycharan: changing

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/InterleavedLedgerStorage.java
reddycharan: changing

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
```diff
@@ -391,4 +400,135 @@ public void testGetEntryLogsSet() throws Exception {
...
+        InterleavedLedgerStorage ledgerStorage = ((InterleavedLedgerStorage) bookie.ledgerStorage);
+
+        int numOfLedgers = 70;
+        int numEntries = 2000;
+        // Create ledgers
+        for (int i = 0; i < numOfLedgers; i++) {
+            ledgerStorage.setMasterKey(i, "key".getBytes());
+        }
+
+        ExecutorService executor = Executors.newFixedThreadPool(10);
```
eolivelli: It seems to me that we are noy shutting down this pool

sijie:
@reddycharan it is okay to go in 4.7. I also have a few features to close for 4.7, that's why I haven't really called a feature frozen yet.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
jiazhai: maybe could use `try(resource){}` blocks here.

reddycharan: Bookie doesn't implements AutoCloseable, Closeable to use with try-with-resources. But again, I'm just following Bookie instances is used in other testcases.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
reddycharan: fixing that

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
ivankelly: It was for diagnosing the errors that I suggested putting the ledgerid and entryid in the message of the thrown exception. You're right about join though. Using get would make things messy, so possibly not worth it.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
ivankelly: I've seen this pattern a few times where we construct a whole bookie, and everything that goes with it, just to get the ledger dirs manager or ledger storage. It makes our tests super heavy, and hard to inject mocked behaviour. This is part of the reason we have Thread.sleep all over the place in the test.

Anyhow, I won't block for this, but it's something we should be aware of and try to avoid in future. If you're concerned about long argument lists, we can create utility methods to hide the complexity.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/InterleavedLedgerStorage.java
ivankelly: with the compareAndSet, the only thing protected here is the flushOrCheckpoint(), which is called unprotected from checkpoint() above, so I think it would be safe. 

It only seems to be called from the syncthread, which means it's effectively synchronized anyhow since it's all on a single thread, and the synchronized here was only to protect the somethingWritten flag. 

It's no harm to leave it in, just seems weird to be mixing atomics and synchronization without good reason.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
```diff
@@ -391,4 +400,137 @@ public void testGetEntryLogsSet() throws Exception {
...
+            return true;
+        }
+    }
+
+    /**
+     * test concurrent write operations and then concurrent read
+     * operations using InterleavedLedgerStorage.
+     */
+    @Test
+    public void testConcurrentWriteAndReadCallsOfInterleavedLedgerStorage() throws Exception {
```
ivankelly: This change effects SortedLedgerStorage too. Add a variant of the test that tests against that.

sijie: why? I don't understand why you need a sorted ledger storage test here. the change is in entry log itself. the entry log is used by all ledger storage implementation, that doesn't mean you need to test all ledger storage implementation, which is redundant. the test should focus on testing what it is changing, not testing all the variants that is calling this class, otherwise we will end up have lots of redundant tests testing same logic, which make the whole build time very long.

so I don't think we need redundant tests here.

ivankelly: > I don't understand why you need a sorted ledger storage test here. the change is in entry log itself.

This change is in interleavedledgerstorage, which sortedledgerstorage derives from. It removes synchronization from processEntry, which sortedledgerstorage calls directly. It seems there's no other synchronization in sortedledgerstorage though, so it should be ok.

> the test should focus on testing what it is changing, 

which is _exactly_ why we shouldn't be constructing a full bookie to test it.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
reddycharan: I don't see why using .get would be messy. Just FYI about Junit, "The JUnit framework captures only assertion errors in the main thread running the test. It is not aware of exceptions from within new spawn threads. In order to do it right, you should communicate the thread's termination state to the main thread." So to get the status/result of task executed in non-main thread we should call for .get of future. Agreed, instead of returning boolean we may throw exception with proper message, but for diagnosing / logging purpose I feel it is better to fail with Assert statements  and enough logging in the tasks. I see it is just a matter of preference and there is no much difference in choosing one over the other.

reddycharan:
@sijie I think this fix should be good to go now.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
sijie: > This change is in interleavedledgerstorage, which sortedledgerstorage derives from. It removes synchronization from processEntry, which sortedledgerstorage calls directly. It seems there's no other synchronization in sortedledgerstorage though, so it should be ok.

this synchronization change only makes sense at InterleavedLedgerStorage, because there can be concurrent calls to `processEntry` when using InterleavedLedgerStorage. However this is not true in SortedLedgerStorage, because ledger storage flushes entries in a single-threaded executor. so I don't see why you need a test from SortedLedgerStorage. 

The test case is testing concurrent reads and write on EntryLogTest, InterleavedLedgerStorage is the tool to use for testing. It is not testing concurrent reads and writes on ledger storage. 

so 1) the testing object is EntryLog not ledger storage 2) concurrent reads and writes only make sense in interleaved ledger storage. so it is not right to add a duplicated test using sorted ledger storage.

> which is exactly why we shouldn't be constructing a full bookie to test it.

that's different. what I am talking here is you need to know what you are exactly _testing_. as I commented above, the testing subject is EntryLog not LedgerStorage.

whether constructing a full bookie or constructing a ledger storage is just a tool to setup a test case for testing EntryLog. 

the whole test suite is using Bookie already. for consistency, it is okay to use Bookie here. I don't see a strong reason to block this change just because of that.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
ivankelly: because .get() throws checked exceptions, so in the snippet I posted, you would have to catch and rethrow as an unchecked exception. 

I was suggest throwing exceptions in the callable thread precisely because asserts are only meaningful in the main thread (they're java.lang.Error). Throwing an exception in the throwable will actually be easer to debug, because you'll get the stacktrace directly on failure, so you can tell what line failed without having to even go to the logs.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
ivankelly: > whether constructing a full bookie or constructing a ledger storage is just a tool to setup a test case for testing EntryLog.

If it's only testing EntryLogger, then only entrylogger should be constructed.

But the change has changed flushing also, and removed locking that was previous around access to the index, so the how ledger storage should be tested. 

> the whole test suite is using Bookie already. for consistency, it is okay to use Bookie here. I don't see a strong reason to block this change just because of that.

The rest of the suite is testing badly. We shouldn't propagate bad practices. I don't consider consistency a good argument against this.

ivankelly: @reddycharan it would be good to have a thread running in a tight loop calling flush() on the ledgerstorage, while the adds and reads are happening.

sijie: > If it's only testing EntryLogger, then only entrylogger should be constructed.

in order to make test entry logger, you need to setup all the facility for do concurrent writes and reads. and InterleavedLedgerStorage is the facility. what is wrong with this? It is a granularity problem on the testing facility, if you just just using EntryLogger you need to setup too many additional stuffs for testing.

> But the change has changed flushing also, and removed locking that was previous around access to the index, so the how ledger storage should be tested.

yes. that's why it is using InterleaveLedgerStorage. because it is a matter for InterleaveLedgerStorage. it is not a matter of SortedLedgerStorage.

reddycharan: > @reddycharan it would be good to have a thread running in a tight loop calling flush() on the ledgerstorage, while the adds and reads are happening.

fair enough I can add flush tasks  to the write and read tasks list, so that flush would be executed along with write/read operations.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
reddycharan: first thing 

1) calling .get() on future is inevitable, since thats the only way you would get the result of the future.

> Instead of returning false on errors, you could throw Exception, and have that carry the context of the error (ledger id, entry id etc). This would simplify the loops in the test itself.

2) even if I change the logic to throw exception instead of boolean in the case of failure, we are not getting rid of loops. Because incase of CancellationException we would get generic cancellation exception instead of any of our exception with customized message. So atleast for this case we need to do what we are doing.

Again as I said this is matter of preference how we deal with return status of tasks/futures in junit test env. If you see it as a dealbreaker for you, I'll go ahead and change it to throw exception, as I don't see value in dragging this further.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
```diff
@@ -391,4 +402,174 @@ public void testGetEntryLogsSet() throws Exception {
...
+        /*
+         * add some flush tasks to the list of writetasks list.
+         */
+        for (int i = 0; i < (numOfLedgers * numEntries) / 500; i++) {
+            writeAndFlushTasks.add(rand.nextInt(writeAndFlushTasks.size()), new LedgerStorageFlushTask(ledgerStorage));
+        }
+
+        // invoke all those write tasks all at once concurrently and set timeout
+        // 6 seconds for them to complete
+        List<Future<Boolean>> writeTasksFutures = executor.invokeAll(writeAndFlushTasks, 6, TimeUnit.SECONDS);
```
ivankelly: I just notice that there's no interleaving of read and write tasks. There should be.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
```diff
@@ -391,4 +402,174 @@ public void testGetEntryLogsSet() throws Exception {
...
+            for (int i = 0; i < numOfLedgers; i++) {
+                writeAndFlushTasks.add(new LedgerStorageWriteTask(i, j, ledgerStorage));
+            }
+        }
+
+        /*
+         * add some flush tasks to the list of writetasks list.
+         */
+        for (int i = 0; i < (numOfLedgers * numEntries) / 500; i++) {
+            writeAndFlushTasks.add(rand.nextInt(writeAndFlushTasks.size()), new LedgerStorageFlushTask(ledgerStorage));
```
ivankelly: Collections.shuffle(writeAndFlushTasks, rand) would have same effect.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
```diff
@@ -391,4 +402,174 @@ public void testGetEntryLogsSet() throws Exception {
...
+    @Test
+    public void testConcurrentWriteAndReadCallsOfInterleavedLedgerStorage() throws Exception {
+        File ledgerDir = createTempDir("bkTest", ".dir");
+        ServerConfiguration conf = TestBKConfiguration.newServerConfiguration();
+        conf.setJournalDirName(ledgerDir.toString());
+        conf.setLedgerDirNames(new String[] { ledgerDir.getAbsolutePath()});
+        conf.setLedgerStorageClass(InterleavedLedgerStorage.class.getName());
+        Bookie bookie = new Bookie(conf);
+        InterleavedLedgerStorage ledgerStorage = ((InterleavedLedgerStorage) bookie.ledgerStorage);
+        Random rand = new Random();
```
ivankelly: Specify a seed, so that the test is more deterministic.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
ivankelly: You only need one flush task, with
```
while (someCondition) {
    ledgerStorage.flush();
}
```
We only flush from a single thread. If there are multiple flush tasks, this could trigger two of them running at the same time.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
reddycharan: for better coverage it is better to have indeterminism, especially here we are trying to make sure there are no race conditions and atomic/transaction issues.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
reddycharan: > Collections.shuffle(writeAndFlushTasks, rand) would have same effect.

anyhow I need to run for loop for creating and adding flush tasks. So it is not a biggie to add an element at random collection.

> We only flush from a single thread. If there are multiple flush tasks, this could trigger two of them running at the same time.

in production code thats true. But there is no harm in running multiple flush tasks at a time and it is better to validate concurrency aspect of that method, anyhow ledgerStorage.flush method is synchronized.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
reddycharan: I'm not sure if it is needed, but anyhow there is no harm in adding so. Adding write tasks to read tasks list as well.

ivankelly: actually, looking at the code again, the read path hasn't changed and was never synchronized anyhow, so this only serves to verify that we can read what the right task has written. Since this is the case, we should read back in a straight loop, and not in tasks.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
ivankelly: But what happens if test hits a failure because of randomness? I can guarantee that 99 times out of 100, noone looks at the failure, reruns for the pass, or just ignores CI. Given that this is how things have always been, I'd prefer more determinism in the order things run, and only hope for randomness in how the system scheduler is scheduling threads.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
ivankelly: sure.

ivankelly:
@reddycharan test is failing. Saying that flush task was cancelled. I don't see how that can be the case. There's no calls to cancel on the future anywhere.

ivankelly:
```
2018-03-06 01:43:23,095 - INFO  - [main:ScanAndCompareGarbageCollector@103] - Over Replicated Ledger Deletion : enabled=true, interval=86400000
2018-03-06 01:43:23,095 - INFO  - [main:GarbageCollectorThread@240] - Minor Compaction : enabled=true, threshold=0.20000000298023224, interval=3600000
2018-03-06 01:43:23,096 - INFO  - [main:GarbageCollectorThread@242] - Major Compaction : enabled=true, threshold=0.800000011920929, interval=86400000
```
Are you expecting to test GC as part of your test?

reddycharan:
this is little surprising. We are seeing cancellationexception because time given for the executor (6 secs) to execute the read and flush tasks has elapsed. https://github.com/apache/bookkeeper/pull/1225/files#diff-7b84d31d42b1b95ba2d985bf2746b448R554

When time has elapsed, the pending tasks are cancelled. It seems it infact waited for 6 seconds from the logs 

2018-03-06 **01:43:23,096** - INFO  - [main:GarbageCollectorThread@242] - Major Compaction : enabled=true, threshold=0.800000011920929, interval=86400000
2018-03-06 **01:43:29,188** - ERROR - [pool-29-thread-4:InterleavedLedgerStorage@331] - Exception flushing Ledger cache

Locally in my macbook pro I ran few times and it is working fine, just curious what is the test env. machines config?

trying to run this test in loop in my macbook

..

those logs is just noise, it is just info logs which we get because of Bookie initialization.

reddycharan:
hmm..I'm able to run successfully 64 times in loop on my machine (MacbookPro), wondering if this failure is consistently repro'able in build test machines or intermittent?

ivankelly:
> When time has elapsed, the pending tasks are cancelled. It seems it infact waited for 6 seconds from the logs

Remove the timeout. There's a timeout on each test anyhow. The jenkins machines are quite beefy in terms of ram and cpu. I think they're on HDD. And two jobs can execute at a time, so that could have been unhelpful.

reddycharan:
yes, testcase must has failed because of timeout issue. Locally on my MacbookPro (using SSD) (not loaded machine) those list of tasks are taking around 2.5 secs and 1.5 secs. I should reduce the number of tasks and also probably remove the timeouts.

ivankelly:
@reddycharan number of tasks should be fine. I'd just remove the timeout. Up to you though.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/EntryLogTest.java
reddycharan: It should be ok to do parallel read tasks, on one side we are testing concurrency aspect of read and also we are reducing the time of the testcase by doing parallel reads.

reddycharan:
This PR should be good to go.

PR #1227, opened 2018-03-04
 Fix AbstractLedgerManagerFactory on handling both shaded class and original class co-exists case

sijie:

*Problem*

#1217 addresses the issue when using a shaded class accessing a bookkeeper cluster.  However the assumption was made when the classpath only has shaded classes. However, it didn't address the case when shaded class and original class can co-exist. because IOException is thrown instead of ClassNotFoundException at following code:

```
                Class<?> theCls = Class.forName(layout.getManagerFactoryClass());
                if (!LedgerManagerFactory.class.isAssignableFrom(theCls)) {
                    throw new IOException("Wrong ledger manager factory " + layout.getManagerFactoryClass());
                }
```

*Solution*

This change fixes the issue.


sijie:
shading is a bit tricky on testing. so I didn't attempt to add a test case for this.

the issue was observed at pulsar and verified this fix address the problem.

sijie:
retest this please

Issue #1229, opened 2018-03-05
PendingAddOp can get recycled before it gets executed


dlg99:
**BUG REPORT**

Scenario:
1. many async writes submitted/pending for the same ledger.
2. Write fails due to not enough bookies.
3. Ledger handle cancels all pending adds. 
4. PendingAddOps of these adds get recycled.
5. Pending add finally gets its chance to execute.
6. LH at that point happens to be null (recycled already) hence Preconditions.checkNotNull throws exception.
7. Exception is logged and swallowed by SafeRunnable.

Problems:
- unnecessary exceptions affecting performance and generating garbage in the log.
- it is possible that recycled op gets reused and executed instead of original one. I think it is possible to get second write of the same entry to single bookie in this case in leu of write to another bookie though I did not try to repro that specific case.

Minimized version of repro, worked on current master / 3e51b8116c363604f93bdd0c6b6b71ef3387c8a8 :

throw in the following test into SlowBookieTest suite:

```java
    @Test
    public void testBookieFailure() throws Exception {
        ClientConfiguration conf = new ClientConfiguration();
        conf.setReadTimeout(5)
                .setAddEntryQuorumTimeout(5)
                .setAddEntryQuorumTimeout(10)
                .setZkServers(zkUtil.getZooKeeperConnectString());

        BookKeeper bkc = new BookKeeper(conf);

        byte[] pwd = new byte[] {};
        final LedgerHandle lh = bkc.createLedger(4, 4, 3, BookKeeper.DigestType.CRC32, pwd);
        final AtomicBoolean finished = new AtomicBoolean(false);
        final AtomicBoolean failTest = new AtomicBoolean(false);
        final byte[] entry = "Test Entry".getBytes();

        Thread t = new Thread(() -> {
            try {
                while (!finished.get()) {
                    lh.asyncAddEntry(entry, (rc, lh1, entryId, ctx) -> {
                        if (rc != BKException.Code.OK) {
                            failTest.set(true);
                            finished.set(true);
                        }
                    }, null);
                }
            } catch (Exception e) {
                LOG.error("Exception in add entry thread", e);
                failTest.set(true);
            }
        });

        killBookie(0);

        t.start();

        Thread.sleep(2_000);

        finished.set(true);
        t.join();

        assertFalse(failTest.get());

        lh.close();

        LedgerHandle lh2 = bkc.openLedger(lh.getId(), BookKeeper.DigestType.CRC32, pwd);
[...]

PR #1230, opened 2018-03-05
 Issue 1229: PendingAddOp can get recycled before it gets executed

sijie:
*Problem*

The PendingAddOp can be recycled when it is cancelled before it is executed. so it will hit NPE when it is actually executed. This is a bug introduced by #1091

*Solution*

Only recycle PendingAddOp after it has been run.

Master Issue: #1229 

> ---
> quickly and easily:
>
>
>
> Otherwise:
> 
> 
> ---

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
@@ -92,10 +92,10 @@ class PendingAddOp extends SafeRunnable implements WriteCallback {
         op.entryLength = payload.readableBytes();
 
         op.completed = false;
-        op.ackSet = lh.distributionSchedule.getAckSet();
-        op.addOpLogger = lh.bk.getAddOpLogger();
-        op.addOpUrCounter = lh.bk.getAddOpUrCounter();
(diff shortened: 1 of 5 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
```diff
@@ -422,7 +422,8 @@ private void maybeRecycle() {
             ReferenceCountUtil.release(toSend);
             toSend = null;
         }
-        if (toSend == null && pendingWriteRequests == 0) {
+        // only recycle a pending add op after it has been run.
+        if (hasRun && toSend == null && pendingWriteRequests == 0) {
```
dlg99: maybe we should simply do 
```java
if (!hasRun) return;
```
at the beginning of maybeRecycle()

dlg99:
thank you, @sijie 
awesome and simple fix.

sijie:
@dlg99 since the test case in SlowBookieTest is not really catching the bug here, I added another test case specific to PendingAddOp to catch the sequence of cancel-before-run.

dlg99:
@sijie test lgtm

PR #1231, opened 2018-03-06
Dont shade netty at distributedlog-core-shaded

sijie:

*Problem*

JVM crashed when `distributedlog-core-shaded` (where netty was shaded) is used in a project where netty isn't shaded
and epoll is enabled. The conflict happens on loading the native epoll transport because shaded netty doesn't actually
shade the native jni library. so signature mismatch when both version of netty try to resolve jni library.

*Solution*

Dont shade netty at all. Let the application resolve which netty version to use.


sijie:
travis and IT passed. since this change is only touch the dlog shaded jar, ignore CI results to merge this, allowing to unblock pulsar.

sijie:
IGNORE CI

PR #1233, opened 2018-03-06
Reduce running time for testLedgerCreateAdvWithLedgerIdInLoop

ivankelly:
This test has flaked in CI regularly mainly because it takes a long
time to run.

This changes reduces the running time by.
1. Creating the ledgers with a ensemble size of 1
2. Creating and writing to the ledgers in parallel
3. Using async methods to write entries to the ledgers

This brought the test from ~60 seconds to ~7 seconds locally.

### DIFF
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieWriteLedgerTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieWriteLedgerTest.java
@@ -631,6 +631,7 @@ public class BookieWriteLedgerTest extends
                         // Read and verify
                         LOG.info("Read entries for ledger: {}", handle.getId());
                         readEntries(handle, entries);
+                        entries.forEach(ByteBuf::release);
                         handle.close();
                         bkc.deleteLedger(handle.getId());
                     } catch (InterruptedException ie) {
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieWriteLedgerTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieWriteLedgerTest.java
@@ -42,7 +42,6 @@ import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutionException;
-import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 
 import org.apache.bookkeeper.bookie.Bookie;
@@ -586,14 +585,14 @@ public class BookieWriteLedgerTest extends
     public void testLedgerCreateAdvWithLedgerIdInLoop() throws Exception {
         int ledgerCount = 40;
 
-        long maxId = Long.MAX_VALUE;
-        if (!baseConf.getLedgerManagerFactoryClass().equals(LongHierarchicalLedgerManagerFactory.class)) {
+        long maxId = 9999999999L;
+        if (baseConf.getLedgerManagerFactoryClass().equals(LongHierarchicalLedgerManagerFactory.class)) {
             // since LongHierarchicalLedgerManager supports ledgerIds of decimal length upto 19 digits but other
             // LedgerManagers only upto 10 decimals
-            maxId = 9999999999L;
+            maxId = Long.MAX_VALUE;
         }
 
-        rng.longs(ledgerCount, 1, maxId) // generate a stream of ledger ids
+        rng.longs(ledgerCount, 0, maxId) // generate a stream of ledger ids
             .mapToObj(ledgerId -> { // create a ledger for each ledger id
                     LOG.info("Creating adv ledger with id {}", ledgerId);
                     return bkc.newCreateLedgerOp()
@@ -604,15 +603,21 @@ public class BookieWriteLedgerTest extends
                         .thenApply(writer -> { // Add entries to ledger when created
                                 LOG.info("Writing stream of {} entries to {}",
                                          numEntriesToWrite, ledgerId);
-                                List<Integer> entries = rng.ints(numEntriesToWrite, 0, maxInt)
-                                    .mapToObj(Integer::valueOf)
+                                List<ByteBuf> entries = rng.ints(numEntriesToWrite, 0, maxInt)
+                                    .mapToObj(i -> {
+                                            ByteBuf entry = Unpooled.buffer(4);
+                                            entry.retain();
+                                            entry.writeInt(i);
+                                            return entry;
+                                        })
                                     .collect(Collectors.toList());
                                 CompletableFuture<?> lastRequest = null;
-                                for (int i = 0; i < entries.size(); i++) {
-                                    ByteBuf entry = Unpooled.buffer(4);
-                                    entry.writeInt(entries.get(i));
-                                    LOG.info("Writing {}:{} as {}", ledgerId, i, entries.get(i));
-                                    lastRequest = writer.write(i, entry);
+                                int i = 0;
+                                for (ByteBuf entry : entries) {
+                                    long entryId = i++;
+                                    LOG.info("Writing {}:{} as {}",
+                                             ledgerId, entryId, entry.slice().readInt());
+                                    lastRequest = writer.write(entryId, entry);
                                 }
                                 lastRequest.join();
                                 return Pair.of(writer, entries);
@@ -622,9 +627,9 @@ public class BookieWriteLedgerTest extends
             .forEach(e -> { // check that each set of adds succeeded
                     try {
                         WriteAdvHandle handle = e.getLeft();
-                        List<Integer> entries = e.getRight();
+                        List<ByteBuf> entries = e.getRight();
                         // Read and verify
-                        LOG.info("Read entries for ledger: {}", handle);
+                        LOG.info("Read entries for ledger: {}", handle.getId());
                         readEntries(handle, entries);
                         handle.close();
                         bkc.deleteLedger(handle.getId());
@@ -1209,14 +1214,17 @@ public class BookieWriteLedgerTest extends
         }
     }
 
-    private void readEntries(ReadHandle reader, List<Integer> entries) throws Exception {
-        AtomicInteger index = new AtomicInteger(0);
-        try (LedgerEntries readEntries = reader.read(0, numEntriesToWrite - 1).join()) {
-            readEntries.forEach(e -> {
-                    int origEntry = entries.get(index.getAndIncrement());
-                    int readEntry = e.getEntryBuffer().readInt();
-                    assertEquals(origEntry, readEntry);
-                });
+    private void readEntries(ReadHandle reader, List<ByteBuf> entries) throws Exception {
+        assertEquals("Not enough entries in ledger " + reader.getId(),
+                     reader.getLastAddConfirmed(), entries.size() - 1);
+        try (LedgerEntries readEntries = reader.read(0, reader.getLastAddConfirmed()).join()) {
+            int i = 0;
+            for (org.apache.bookkeeper.client.api.LedgerEntry e : readEntries) {
+                int entryId = i++;
+                ByteBuf origEntry = entries.get(entryId);
+                ByteBuf readEntry = e.getEntryBuffer();
+                assertEquals("Unexpected contents in " + reader.getId() + ":" + entryId, origEntry, readEntry);
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


ivankelly:
@reddycharan I tried to add you as a reviewer to this, but it wouldn't find your name. Are you in the apache org in github?

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieWriteLedgerTest.java
```diff
@@ -579,45 +584,58 @@ public void addComplete(int rc, LedgerHandle lh, long entryId, Object ctx) {
...
-            long lid = lhArray[lc].getId();
-            LOG.info("readEntries for lc: {} ledgerId: {} ", lc, lhArray[lc].getId());
-            readEntries(lhArray[lc], entryList.get(lc));
-            lhArray[lc].close();
-            bkc.deleteLedger(lid);
+        long maxId = Long.MAX_VALUE;
+        if (!baseConf.getLedgerManagerFactoryClass().equals(LongHierarchicalLedgerManagerFactory.class)) {
+            // since LongHierarchicalLedgerManager supports ledgerIds of decimal length upto 19 digits but other
+            // LedgerManagers only upto 10 decimals
+            maxId = 9999999999L;
```
reddycharan: nit: probably it is easier to assign maxId to 9999999999L and assign Long.MAX_VALUE if it is LongHierarchicalLedgerManagerFactory

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieWriteLedgerTest.java
```diff
@@ -579,45 +584,58 @@ public void addComplete(int rc, LedgerHandle lh, long entryId, Object ctx) {
...
         }
+
+        rng.longs(ledgerCount, 1, maxId) // generate a stream of ledger ids
```
reddycharan: nit: why cannt randomNumberOrigin be 0 ?

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieWriteLedgerTest.java
```diff
@@ -579,45 +584,58 @@ public void addComplete(int rc, LedgerHandle lh, long entryId, Object ctx) {
...
+                        .withPassword(ledgerPassword).makeAdv().withLedgerId(ledgerId)
+                        .execute()
+                        .thenApply(writer -> { // Add entries to ledger when created
+                                LOG.info("Writing stream of {} entries to {}",
+                                         numEntriesToWrite, ledgerId);
+                                List<Integer> entries = rng.ints(numEntriesToWrite, 0, maxInt)
+                                    .mapToObj(Integer::valueOf)
+                                    .collect(Collectors.toList());
+                                CompletableFuture<?> lastRequest = null;
+                                for (int i = 0; i < entries.size(); i++) {
```
reddycharan: instead of entries.size() can't we just use numEntriesToWrite?

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieWriteLedgerTest.java
```diff
@@ -1191,6 +1209,17 @@ private void readEntries(LedgerHandle lh, List<byte[]> entries) throws Interrupt
         }
     }
 
+    private void readEntries(ReadHandle reader, List<Integer> entries) throws Exception {
+        AtomicInteger index = new AtomicInteger(0);
+        try (LedgerEntries readEntries = reader.read(0, numEntriesToWrite - 1).join()) {
+            readEntries.forEach(e -> {
+                    int origEntry = entries.get(index.getAndIncrement());
+                    int readEntry = e.getEntryBuffer().readInt();
+                    assertEquals(origEntry, readEntry);
```
reddycharan: nit: probably adding ledgerid and entryid to the assert message would be helpful in diagnosing in the case of failure

reddycharan: also here for thoroughness we should do ByteBuf comparisons not int comparisons

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieWriteLedgerTest.java
```diff
@@ -579,45 +584,58 @@ public void addComplete(int rc, LedgerHandle lh, long entryId, Object ctx) {
...
+                                    entry.writeInt(entries.get(i));
+                                    LOG.info("Writing {}:{} as {}", ledgerId, i, entries.get(i));
+                                    lastRequest = writer.write(i, entry);
+                                }
+                                lastRequest.join();
+                                return Pair.of(writer, entries);
+                            });
+                })
+            .parallel().map(CompletableFuture::join) // wait for all creations and adds in parallel
+            .forEach(e -> { // check that each set of adds succeeded
```
reddycharan: I know junit assert would work only in the main thread and thats why you are using foreach..but wondering if you run the read operations as well in parallel stream, store the status of each operation and finally in foreach loop you can validate the status by using junit asserts

Since you parallelized write operations why not read operations as well. Just a thought, its not a blocker for me.

reddycharan:
> @reddycharan I tried to add you as a reviewer to this, but it wouldn't find your name. Are you in the apache org in github?

don't think so. is there any todo list to get one?

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieWriteLedgerTest.java
ivankelly: No reason, changed.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieWriteLedgerTest.java
ivankelly: changed.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieWriteLedgerTest.java
ivankelly: I've changed it to use a for-each loop instead.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieWriteLedgerTest.java
ivankelly: Added

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieWriteLedgerTest.java
ivankelly: Every operation on the stream after the parallel() is run in parallel, so reading is parallel too. Regarding junit assert, at its core it's a Throwable. Parallel will run the operations in a ForkJoinPool. If an operation throws a throwable, the join will throw it in the main thread. So if you assert in a parallel op, it gets thrown up into main thread. I've experimented with this a bit to ensure it's the case.

ivankelly:
@reddycharan 
> don't think so. is there any todo list to get one?

You can link your apache and github account with gitbox (https://gitbox.apache.org/)

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieWriteLedgerTest.java
```diff
@@ -579,45 +583,64 @@ public void addComplete(int rc, LedgerHandle lh, long entryId, Object ctx) {
...
+                        .withDigestType(org.apache.bookkeeper.client.api.DigestType.CRC32)
+                        .withPassword(ledgerPassword).makeAdv().withLedgerId(ledgerId)
+                        .execute()
+                        .thenApply(writer -> { // Add entries to ledger when created
+                                LOG.info("Writing stream of {} entries to {}",
+                                         numEntriesToWrite, ledgerId);
+                                List<ByteBuf> entries = rng.ints(numEntriesToWrite, 0, maxInt)
+                                    .mapToObj(i -> {
+                                            ByteBuf entry = Unpooled.buffer(4);
+                                            entry.retain();
```
reddycharan: the way ByteBuf instance is used in this testcase makes me question the specifications and expectations of WriteAdvHandle.write(long entryId, ByteBuf data) API.

1) Regarding the refcount of ByteBuf, from what I understand the API method which takes ByteBuf as argument shouldn't increase or decrease the refcount of 'ReferenceCounted' object.

I mean
```
myApplicationMethod() {

    ByteBuf byteBuf = createNewInstanceOfByteBuf();
    ...
    APIClass.APIMethod(bytebuf);
    ...
    byteBuf.release(); 
}
```
here when I create byteBuf the refCount would typically be 1. Now by passing this refcounted object to the standard API method, I don’t expect the APIMethod to decrease the refCount (modify refCount) and finally it is the responsibility of the caller (application) method to release it by calling release method explicitly when they are done with the byteBuf.

But in our case  WriteAdvHandle.write(long entryId, ByteBuf data) is decreasing the refcount, this is because of https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java#L422 . Hence @ivankelly is calling https://github.com/apache/bookkeeper/pull/1233/commits/23a548d95474969729a59a1d3676ecf0de24d34a#diff-d168699d534345c67678296447f8375cR609 , explicit retain() immediately after creating ByteBuf. 

**Que:** Why are we doing this? What is the general practice with regards to usage of refcounted objects in API methods? Have we mentioned the expected behavior in detail in our API docs?

2) Regarding the the updates to readerIndex/writerIndex of ByteBuf by API method. AFAIU if the API method is doing read operation from the byteBuf then after calling API method with ByteBuf argument, I wouldn’t expect the readerIndex of the byteBuf to be at 0.

I mean

```
myApplicationMethod() {

    ByteBuf byteBuf = createNewInstanceOfByteBuf();
    // add data to byteBuf
    APIClass.APIReadMethod(bytebuf);
    // here I would expect the readerIndex and writerIndex of byteBuf to be the same,
   // since APIReadMethod has consumed/read all of the data in the byteBuf
}
```
But in our case, WriteAdvHandle.write(long entryId, ByteBuf data) doesn’t advances readerIndex of data ByteBuf and this might be because of the following clone https://github.com/apache/bookkeeper/blob/19846b23de3be13e28082d4ba73e21bb8fa39ad2/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtocol.java#L271 

I would expect it to behave in the same way as other core java APIs deal with ByteBuffer - for eg: https://docs.oracle.com/javase/7/docs/api/java/nio/channels/FileChannel.html#read(java.nio.ByteBuffer)  https://netty.io/4.0/api/io/netty/buffer/ByteBuf.html#readBytes-io.netty.buffer.ByteBuf- here position of ByteBuffer and readerIndex of ByteBuf is advanced.

**Que:** Are we following general practices of Buffers with our API?  Have we mentioned the expected behavior in detail in our API docs?

3) In this test case @ivankelly is explicitly calling retain immediately after creation of ByteBuf but the corresponding release call isn’t there. Wondering is it ok to be lenient? Are we doing corresponding release calls in all the places where ByteBuff is created. https://netty.io/4.0/api/io/netty/util/ReferenceCounted.html says that initial count would be 1 when an ReferenceCounted object is created.

My bad, I wasn’t paying attention to new API and netty buffer changes. But the way this test case handled with ByteBuf made me delve into this. IIRC @merlimat and @kishorekasi were making changes to the netty buffers and @eolivelli  introduced new client API.

@sijie @jvrao  FYI

ivankelly:
@reddycharan I guess I comes down to a question of whether you expect the API to take ownership of a passed in ByteBuf or not. i.e. if I create a bytebuf and pass it in, do I expect the API to clean it up. In our APIs we do make this assumption. I kinda prefer it this way, as then you should be able to forget about the bytebuf once passed in, but this isn't a strong preference. This is documented in the apidocs, though it's not very loud. We should probably have a full non-javadoc page of documentation dedicated to our ByteBuf usage.

Anyhow, the retain in question is there because I want to keep those ByteBufs alive beyond the lifetime of the write call so they can be compared later. I neglected to release them out of laziness. Added the release now.

reddycharan:
testcode change LGTM!

For client API specifications and expectations regarding ByteBuf refcount and reader/writerindexes we discussed in todays community meet. This needs broader conversation will create Question Issues for better tracking.

sijie:
Thank you @ivankelly and @reddycharan. will catch up with the bytebuf discussion.

reddycharan:
@sijie I created Question Issues, lets discuss about it there

https://github.com/apache/bookkeeper/issues/1237
https://github.com/apache/bookkeeper/issues/1238

sijie:
@reddycharan thank you

PR #1236, opened 2018-03-08
Issue #570: make changes to SyncThread/checkpoint logic.

reddycharan:

make changes to SyncThread/checkpoint logic, so that incase of
entrylogperledger, checkpoint happens for every flushInterval
but not when active entrylog is created/rolled over.

This is < sub-task3  > of Issue #570

Master Issue: #570

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
@@ -63,6 +63,7 @@ import org.apache.bookkeeper.bookie.BookieException.DiskPartitionDuplicationExce
 import org.apache.bookkeeper.bookie.BookieException.InvalidCookieException;
 import org.apache.bookkeeper.bookie.BookieException.MetadataStoreException;
 import org.apache.bookkeeper.bookie.BookieException.UnknownBookieIdException;
+import org.apache.bookkeeper.bookie.CheckpointSource.Checkpoint;
 import org.apache.bookkeeper.bookie.Journal.JournalScanner;
 import org.apache.bookkeeper.bookie.LedgerDirsManager.LedgerDirsListener;
 import org.apache.bookkeeper.bookie.LedgerDirsManager.NoWritableLedgerDirException;
@@ -696,9 +697,7 @@ public class Bookie extends BookieCriticalThread {
         String ledgerStorageClass = conf.getLedgerStorageClass();
         LOG.info("Using ledger storage: {}", ledgerStorageClass);
         ledgerStorage = LedgerStorageFactory.createLedgerStorage(ledgerStorageClass);
-        syncThread = new SyncThread(conf, getLedgerDirsListener(), ledgerStorage, checkpointSource);
 
-        Checkpointer checkpointer;
         /*
          * with this change https://github.com/apache/bookkeeper/pull/677,
          * LedgerStorage drives the checkpoint logic. But with multiple entry
@@ -707,10 +706,26 @@ public class Bookie extends BookieCriticalThread {
          * SyncThread.start does that and it is started in Bookie.start method.
          */
         if (entryLogPerLedgerEnabled) {
-            checkpointer = (checkpoint) -> {
+            syncThread = new SyncThread(conf, getLedgerDirsListener(), ledgerStorage, checkpointSource) {
+                @Override
+                public void startCheckpoint(Checkpoint checkpoint) {
+                    /*
+                     * in the case of entryLogPerLedgerEnabled, LedgerStorage
+                     * dont drive checkpoint logic, but instead it is done
+                     * periodically by SyncThread. So startCheckpoint which
+                     * will be called by LedgerStorage will be no-op.
+                     */
+                }
+
+                @Override
+                public void start() {
+                    executor.scheduleAtFixedRate(() -> {
+                        doCheckpoint(checkpointSource.newCheckpoint());
+                    }, conf.getFlushInterval(), conf.getFlushInterval(), TimeUnit.MILLISECONDS);
+                }
             };
         } else {
-            checkpointer = syncThread;
+            syncThread = new SyncThread(conf, getLedgerDirsListener(), ledgerStorage, checkpointSource);
         }
 
         ledgerStorage.initialize(
@@ -720,7 +735,7 @@ public class Bookie extends BookieCriticalThread {
             indexDirsManager,
             stateManager,
             checkpointSource,
-            checkpointer,
+            syncThread,
             statsLogger);
 
 
@@ -846,9 +861,7 @@ public class Bookie extends BookieCriticalThread {
          * checkpoint which reduce the chance that we need to replay journals
          * again if bookie restarted again before finished journal replays.
          */
-        if (entryLogPerLedgerEnabled) {
-            syncThread.scheduleCheckpointAtFixedRate(conf.getFlushInterval());
-        }
+        syncThread.start();
 
         // start bookie thread
         super.start();
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
@@ -847,7 +847,7 @@ public class Bookie extends BookieCriticalThread {
          * again if bookie restarted again before finished journal replays.
          */
         if (entryLogPerLedgerEnabled) {
-            syncThread.start();
+            syncThread.scheduleCheckpointAtFixedRate(conf.getFlushInterval());
         }
 
         // start bookie thread
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
@@ -112,6 +112,7 @@ public class Bookie extends BookieCriticalThread {
     final List<Journal> journals;
 
     final HandleFactory handles;
+    final boolean entryLogPerLedgerEnabled;
 
     static final long METAENTRY_ID_LEDGER_KEY = -0x1000;
     static final long METAENTRY_ID_FENCE_KEY  = -0x2000;
@@ -688,6 +689,7 @@ public class Bookie extends BookieCriticalThread {
                          conf, ledgerDirsManager, statsLogger.scope(JOURNAL_SCOPE + "_" + i)));
         }
 
+        this.entryLogPerLedgerEnabled = conf.isEntryLogPerLedgerEnabled();
         CheckpointSource checkpointSource = new CheckpointSourceList(journals);
 
         // Instantiate the ledger storage implementation
@@ -696,6 +698,21 @@ public class Bookie extends BookieCriticalThread {
         ledgerStorage = LedgerStorageFactory.createLedgerStorage(ledgerStorageClass);
         syncThread = new SyncThread(conf, getLedgerDirsListener(), ledgerStorage, checkpointSource);
 
+        Checkpointer checkpointer;
+        /*
+         * with this change https://github.com/apache/bookkeeper/pull/677,
+         * LedgerStorage drives the checkpoint logic. But with multiple entry
+         * logs, checkpoint logic based on a entry log is not possible, hence it
+         * needs to be timebased recurring thing and it is driven by SyncThread.
+         * SyncThread.start does that and it is started in Bookie.start method.
+         */
+        if (entryLogPerLedgerEnabled) {
+            checkpointer = (checkpoint) -> {
+            };
+        } else {
+            checkpointer = syncThread;
+        }
+
         ledgerStorage.initialize(
             conf,
             ledgerManager,
@@ -703,7 +720,7 @@ public class Bookie extends BookieCriticalThread {
             indexDirsManager,
             stateManager,
             checkpointSource,
-            syncThread,
+            checkpointer,
             statsLogger);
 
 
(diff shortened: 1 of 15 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SyncThread.java
```diff
@@ -119,6 +119,12 @@ public void startCheckpoint(Checkpoint checkpoint) {
         });
     }
 
+    void start() {
```
sijie: Can you rename this to "scheduleCheckpointAtFixedRate(long flushIntervalMs)" as I suggested in #1201?

- it is not actually a "start" action because the scheduler has been started, this is more a "schedule" action
- flushInterval doesn't belong to SyncThread any more.

Also can you update the comment about 'flushInterval' in ServerConfiguration and conf/bk_server.conf, `site/_data/config/bk_server.yaml`? `flushInterval` is now only works when `entryLogPerLedger` enabled.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/LedgerStorageCheckpointTest.java
```diff
@@ -0,0 +1,580 @@
...
+        handle.close();
+
+        Assert.assertNotEquals("bytesWrittenSinceLastFlush shouldn't be zero", 0,
+                entryLogger.bytesWrittenSinceLastFlush);
+        Assert.assertNotEquals("There should be logChannelsToFlush", 0, entryLogger.logChannelsToFlush.size());
+
+        /*
+         * wait for atleast flushInterval period, so that checkpoint can happen.
+         */
+        Thread.sleep(conf.getFlushInterval() + 500);
```
ivankelly: Don't put sleeps in tests. Use MockExecutorController.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/LedgerStorageCheckpointTest.java
```diff
@@ -0,0 +1,580 @@
...
+    /*
+     * In this testcase, SortedLedgerStorage is used and validate if the
+     * checkpoint is called for every flushinterval period.
+     */
+    @Test
+    public void testPeriodicCheckpointForSortedLedgerStorage() throws Exception {
+        testPeriodicCheckpointForLedgerStorage(SortedLedgerStorage.class.getName());
+    }
+
+    public void testPeriodicCheckpointForLedgerStorage(String ledgerStorageClassName) throws Exception {
```
ivankelly: What are you actually testing here? There are two changes in the main code. They should be tested separately. For the SyncThread change, you just need to validate that if perLedgerEntryLog is true, it runs periodically. You only need to construct a SyncThread for this. LedgerStorage and everything else can be mocked.

For the InterleavedLedgerStorage, you want to check that if perLedgerEntryLog is set, then flush always gets called instead of checkpoint. Similarly this can be done with mocks.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -696,14 +698,29 @@ public Bookie(ServerConfiguration conf, StatsLogger statsLogger)
...
 
+        Checkpointer checkpointer;
+        /*
+         * with this change https://github.com/apache/bookkeeper/pull/677,
+         * LedgerStorage drives the checkpoint logic. But with multiple entry
+         * logs, checkpoint logic based on a entry log is not possible, hence it
+         * needs to be timebased recurring thing and it is driven by SyncThread.
+         * SyncThread.start does that and it is started in Bookie.start method.
+         */
+        if (entryLogPerLedgerEnabled) {
```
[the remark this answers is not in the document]
reddycharan: hmm, I've no strong opinion either way.. @sijie what do you think about having no-op start() method in SyncThread class and Bookie calling .start method in all cases?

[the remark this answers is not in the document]
eolivelli: I had the same idea as @ivankelly but I never clocked on 'submit' review.
So +1 for me for having a common base class. But this is not blocker for me

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
```diff
@@ -179,6 +179,11 @@
     // Stats
     protected static final String ENABLE_TASK_EXECUTION_STATS = "enableTaskExecutionStats";
 
+    /*
+     * config specifying if the entrylog per ledger is enabled or not.
+     */
+    protected static final String ENTRY_LOG_PERLEDGER_ENABLED = "entryLogPerLedgerEnabled";
```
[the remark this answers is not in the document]
sijie: why "_"? we don't use "_" in the setting name.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -696,14 +698,29 @@ public Bookie(ServerConfiguration conf, StatsLogger statsLogger)
         ledgerStorage = LedgerStorageFactory.createLedgerStorage(ledgerStorageClass);
         syncThread = new SyncThread(conf, getLedgerDirsListener(), ledgerStorage, checkpointSource);
 
+        Checkpointer checkpointer;
+        /*
+         * with this change https://github.com/apache/bookkeeper/pull/677,
```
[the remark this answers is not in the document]
sijie: no, it should be the issue link. because  the issue link will contain all the review comments, but git sha doesn't contain any review comments.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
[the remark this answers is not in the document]
sijie: I am fine with current implementation with boolean flag. but not against to add a no-op start().

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/LedgerManagerFactory.java
```diff
@@ -94,7 +94,7 @@ LedgerUnderreplicationManager newLedgerUnderreplicationManager()
      * @param lm
      *            Layout manager
      */
```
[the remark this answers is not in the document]
sijie: it is probably just need rebase. not a big deal, not sure it is worth a comment here.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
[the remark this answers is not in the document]
eolivelli: Okay

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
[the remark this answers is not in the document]
ivankelly: the setting name is entryLogPerLedgerEnabled, not entryLogPerledgerEnabled. It's the variable I want updated to be consistent with the setting name, as the setting name treats Per and Ledger as separate words. As they are.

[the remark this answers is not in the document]
reddycharan: changed it

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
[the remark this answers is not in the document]
reddycharan: added start method to Checkpointer interface.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/InterleavedLedgerStorage.java
```diff
@@ -331,9 +333,16 @@ private void flushOrCheckpoint(boolean isCheckpointFlush)
...
+            /*
+             * if it is just a checkpoint flush and if entryLogPerLedger is not
+             * enabled, then we just flush rotated entry log files in entry
+             * logger.
+             *
+             * In the case of entryLogPerLedgerEnabled we need to flush both
+             * rotatedlogs and currentlogs. Hence we call entryLogger.flush in
+             * the case of entrylogperledgerenabled.
+             */
+            if (isCheckpointFlush && !entryLogPerLedgerEnabled) {
```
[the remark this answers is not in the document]
reddycharan: for now moved this if-else block to EntryLogger.checkpoint. In the subsequent tasks when I introduce EntrylogManager, I'll move the implementation of checkpoint to individual implementations of EntryLogManager.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/LedgerStorageCheckpointTest.java
reddycharan: using MockExecutorController.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/LedgerStorageCheckpointTest.java
reddycharan: since it is Checkpoint logic, I want to be cautious. Hence added testing at Bookie level. For each testcase I added comment describing what I'm intending to do in the testcase.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SyncThread.java
```diff
@@ -108,14 +103,13 @@ public void startCheckpoint(Checkpoint checkpoint) {
         });
     }
 
-    public Future<Void> requestFlush() {
+    public Future requestFlush() {
```
ivankelly: nit: why did you remove the generic here?

ivankelly:
retest this please // looks like failures are overfull disks

ivankelly:
retest this please // H23 full disk killed integration test. infra have taken it offline now

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SyncThread.java
reddycharan: 1) it is returning void/null.
2) its return value is not used anywhere.
3) MockExecutorController doesn't support controlSubmitCallable.

sijie: > it is returning void/null.

then you might consider returning Future<?>

> MockExecutorController doesn't support controlSubmitCallable.

oh, we can add support for submit callables

reddycharan:
@sijie @eolivelli not sure what is going on with the test env. Can you please check this. If things are fine with tests then this PR is good to go.

eolivelli:
retest this please

reddycharan:
@eolivelli can I /how to do that?

sijie:
@eolivelli was about to just manual trigger java8 build. "retest .. please" will take a long time to finish for all builds..

sijie:
@reddycharan you can comment the command as what @eolivelli commented. it will trigger jenkins to rebuild the jobs. but please don't comment the same command again, otherwise jekins will schedule builds again...

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/SortedLedgerStorageCheckpointTest.java
```diff
@@ -110,15 +110,27 @@ public void setUp() throws Exception {
...
-            log.info("Checkpoint the storage at {}", checkpoint);
-            try {
-                storage.checkpoint(checkpoint);
-                checkpoints.add(checkpoint);
-            } catch (IOException e) {
-                log.error("Failed to checkpoint at {}", checkpoint, e);
+        this.checkpointer = new Checkpointer() {
+            @Override
+            public void startCheckpoint(Checkpoint checkpoint) {
+                // TODO Auto-generated method stub
```
eolivelli: Nit: remove this comment

sijie:
@reddycharan : you can checkout this http://bookkeeper.apache.org/community/testing/

site/_data/config/bk_server.yaml
```diff
@@ -43,8 +43,8 @@ groups:
...
-    default: 100
+    description: When entryLogPerLedgerEnabled is enabled, checkpoint doesn't happens when a new active entrylog is created / previous one is rolled over. Instead SyncThread checkpoints periodically with 'flushInterval' delay (in milliseconds) in between executions. Checkpoint flushes both ledger entryLogs and ledger index pages to disk.  Flushing entrylog and index files will introduce much random disk I/O. If separating journal dir and ledger dirs each on different devices, flushing would not affect performance. But if putting journal dir and ledger dirs on same device, performance degrade significantly on too frequent flushing. You can consider increment flush interval to get better performance, but you need to pay more time on bookie server restart after failure. This config is used only when entryLogPerLedgerEnabled is enabled.
+    default: 10000
   - param: allowStorageExpansion
```
eolivelli: Arw we changing this?

sijie: @eolivelli it was already 10000 - https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java#L338 I don't think the change touches it.

reddycharan:
ok by commenting the command it will retrigger test/merge process. Since Enrico has commented that command a test run should have triggered and I don't need to issue that command again. Thanks @sijie for pointing me to the doc.

site/_data/config/bk_server.yaml
reddycharan: there is discrepancy in the default value of this config. In ServerConfiguration.java it is mentioned as 10000 (10 secs) but in this file it is mentioned as 100. So I'm correcting it.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/SortedLedgerStorageCheckpointTest.java
reddycharan: removed it.

reddycharan:
@sijie I fixed nit comment made by @eolivelli (removing auto-generated comment). This should be good to go.

sijie:
@reddycharan cool. merging it. now

sijie:
thanks @reddycharan for this great change!

PR #1249, opened 2018-03-12
Fixed Journal static empty array list recycling

merlimat:
When `journalSyncData=false`, we passing `EMPTY_ARRAY_LIST` to the `ForceWriteThread` when the journal is rolled over. 

I have seen this exception:

```
18:39:52.285 [ForceWriteThread] ERROR org.apache.bookkeeper.bookie.BookieCriticalThread - Uncaught exception in thread ForceWriteThread and is exiting!
java.lang.NullPointerException: null
	at io.netty.util.Recycler$DefaultHandle.recycle(Recycler.java:219) ~[io.netty-netty-all-4.1.21.Final.jar:4.1.21.Final]
	at org.apache.bookkeeper.common.collections.RecyclableArrayList.recycle(RecyclableArrayList.java:55) ~[org.apache.bookkeeper-bookkeeper-server-shaded-4.7.0-SNAPSHOT.jar:4.7.0-SNAPSHOT]
	at org.apache.bookkeeper.bookie.Journal$ForceWriteRequest.recycle(Journal.java:402) ~[org.apache.bookkeeper-bookkeeper-server-shaded-4.7.0-SNAPSHOT.jar:4.7.0-SNAPSHOT]
	at org.apache.bookkeeper.bookie.Journal$ForceWriteRequest.access$1(Journal.java:399) ~[org.apache.bookkeeper-bookkeeper-server-shaded-4.7.0-SNAPSHOT.jar:4.7.0-SNAPSHOT]
	at org.apache.bookkeeper.bookie.Journal$ForceWriteThread.run(Journal.java:506) ~[org.apache.bookkeeper-bookkeeper-server-shaded-4.7.0-SNAPSHOT.jar:4.7.0-SNAPSHOT]
```

This is due to calling `recycle()` twice on the same `EMPTY_ARRAY_LIST`  instance. The solution is that `EMPTY_ARRAY_LIST` should not be a pooled instance and we should check that before recycling.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
@@ -66,7 +66,7 @@ public class Journal extends BookieCriticalThread implements CheckpointSource {
 
     private static final RecyclableArrayList.Recycler<QueueEntry> entryListRecycler =
         new RecyclableArrayList.Recycler<QueueEntry>();
-    private static final RecyclableArrayList<QueueEntry> EMPTY_ARRAY_LIST = entryListRecycler.newInstance();
+    private static final RecyclableArrayList<QueueEntry> EMPTY_ARRAY_LIST = new RecyclableArrayList<>();
 
     /**
      * Filter to pickup journals.
--- a/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/collections/RecyclableArrayList.java
+++ b/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/collections/RecyclableArrayList.java
@@ -45,6 +45,14 @@ public final class RecyclableArrayList<T> extends ArrayList<T> {
 
     private final Handle<RecyclableArrayList<T>> handle;
 
+    /**
+     * Default non-pooled instance.
+     */
+    public RecyclableArrayList() {
+        super();
+        this.handle = null;
+    }
+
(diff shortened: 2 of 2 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -66,7 +66,7 @@
 
     private static final RecyclableArrayList.Recycler<QueueEntry> entryListRecycler =
         new RecyclableArrayList.Recycler<QueueEntry>();
-    private static final RecyclableArrayList<QueueEntry> EMPTY_ARRAY_LIST = entryListRecycler.newInstance();
+    private static final RecyclableArrayList<QueueEntry> EMPTY_ARRAY_LIST = new RecyclableArrayList<>();
```
eolivelli: I can't remember why we can't use a simple ArrayList if we cannot recycle it

sijie: https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java#L409

merlimat: Yes, the `ForceWriteThread` is taking a `RecyclableArrayList` so that it knows it can call `recycle()`. Otherwise we would have to check : 

```java
if (list instanceof RecyclableArrayList) {
  ((RecyclableArrayList) list).recycle();
}
```

PR #1250, opened 2018-03-13
Fixed Journal stats names

merlimat:
In `BOOKKEEPER-1009: Use multiple journals in bookie` 123eccd435a4a96a9147ed4a24efbe9025fe79ba there was a change in the metrics name that would be affecting also user not running with multiple journal.

It is a bit inconvenient to aggregate the stats in the metrics collector (how to aggregate 99pct latency for example). 

I think the best option is to have a single metric even when using multiple journal threads.


sijie:
@merlimat : you need to update [OpStatTest](https://github.com/apache/bookkeeper/blob/2d9c327fee60f5025ce9adad4d7c863958ab59ef/bookkeeper-server/src/test/java/org/apache/bookkeeper/test/OpStatTest.java#L116) since it is using the old journal format.

PR #1251, opened 2018-03-13
BP-29 (task 4): use metadata service uri for constructing registration manager and ledger manager factory

sijie:

The main changes are at two places:

- Deprecating using ServerConfiguration#getRegistrationManagerClass(). Changed Bookie.java to use MetadataBookieDriver to getRegistrationManager().
- Use MetadataBookieDriver/MetadataClientDriver to retrieve ledger manager factory in Bookie.java / BookKeeper.java

The other changes are in-place changes for replacing using registration manager to using metadata drivers.

- provides a few util functions in `MetadataDrivers` to run `functions` with registration manager / ledger manager factory

Master Issue: #1123 

> ---
> quickly and easily:
>
>
>
> Otherwise:
> 
> 
> ---


sijie:
NOTE: most of the changes here are just replacing registration manager with metadata driver. since I am changing the way how the registration manager was constructed, so I have touch the places where registration manager was referenced and used.

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
```diff
@@ -204,7 +204,8 @@ public String getMetadataServiceUri() throws ConfigurationException {
             String ledgerManagerType;
             Class<? extends LedgerManagerFactory> factoryClass = getLedgerManagerFactoryClass();
             if (factoryClass == null) {
-                ledgerManagerType = HierarchicalLedgerManagerFactory.NAME;
+                // set the ledger manager type to "null", so the driver implementation knows that the type is not set.
+                ledgerManagerType = "null";
```
eolivelli: Wouldn't it more user friendly something like 'auto' instead of 'null'.
Like zk+auto://.....
Instead of
zk+null://

Or maybe I did not understand correctly

sijie: "zk+null" comes from bc + testing and used internally when `metadataServiceUri` is not provided. by default user should "zk://..." for the purpose of "auto". I don't want to introduce a concept called "auto", which is confusing.

sijie: @eolivelli if you feel strong about "zk+null" for the internal usage,  I can change in a subsequent PR. merging this PR now to unblock subsequent changes.

eolivelli: @sijie  totally makes sense.

eolivelli:
+1 looks good to me

PR #1252, opened 2018-03-13
Set `ENABLE_DIGEST_TYPE_AUTODETECTION` to true as default value

sijie:

*Motivation*

We have recorded digest type in the ledger metadata since BK 4.5. it would be great to let client automatically detect the digest type
on opening ledger. So turn on the autodetection feature by default.

*Result*

digest type autodetection is on by default.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ClientConfiguration.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ClientConfiguration.java
@@ -239,10 +239,12 @@ public class ClientConfiguration extends AbstractConfiguration<ClientConfigurati
      * <p>Ignores provided digestType, if enabled and uses one from ledger metadata instead.
      * Incompatible with ledger created by bookie versions < 4.2
      *
+     * <p>It is turned on by default since 4.7.
+     *
      * @return flag to enable/disable autodetection of digest type.
      */
     public boolean getEnableDigestTypeAutodetection() {
-        return getBoolean(ENABLE_DIGEST_TYPE_AUTODETECTION, false);
+        return getBoolean(ENABLE_DIGEST_TYPE_AUTODETECTION, true);
     }
 
     /**


sijie:
>  may this change break existing tests, or at least change the meaning of what is tested ?

CI will tell. 

> we should add a test about this change, at least on the new API

There is already change testing the feature itself. so I don't think it need a new test for that, which sounds redundant and not useful.

ivankelly:
There are legit test failures. A BC test fails too. If you can configure it to be compatible, it's not a problem though.

merlimat:
> We have recorded digest type in the ledger metadata since BK 4.5. 

I think the metadata was there since 4.2

sijie:
retest this please

sijie:
@dlg99 can you take a look at this change? I think the logic of 'digest type autodetection' has some problems. since you were the author of this feature, it would be good for you to review my change.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeper.java
```diff
@@ -741,7 +741,8 @@ ZooKeeper getZkHandle() {
         return ((ZKMetadataClientDriver) metadataDriver).getZk();
     }
 
-    protected ClientConfiguration getConf() {
+    @VisibleForTesting
+    public ClientConfiguration getConf() {
```
eolivelli: This is not good.
Unfortunately this object is mutable.
You can you Whitebox in tests

sijie: > This is not good.

forget "VisibleForTesting" annotation, it is very typical to have `getConf` api in a client to retrieve the configuration used by a given client. which I don't think make `getConf` is considered "not good".

I am adding "VisibleForTesting" because this only exists in old bookkeeper api. Ideally the new api should provide `getConf` as well.

eolivelli: If I remember correctly sometimes we are picking current values from the configuration.
Maybe it is better to return a copy. We already have the 'features' feature which enable explicitly dynamic reconfiguration.
If you think this is really useful to add it now I am fine. In fact actually the configuration object is already mutable and we do not copy it at instantiation time of the Bookkeeper object.

We can drop the annotation and let it 'public'. We can also add it in the new API.
The only thing I would like is to make it clear in docs that the result must not be changed, and so it would be better to return a copy or an immutable object

eolivelli: If I remember correctly sometimes we are picking current values from the configuration.
Maybe it is better to return a copy. We already have the 'features' feature which enable explicitly dynamic reconfiguration.
If you think this is really useful to add it now I am fine. In fact actually the configuration object is already mutable and we do not copy it at instantiation time of the Bookkeeper object.

We can drop the annotation and let it 'public'. We can also add it in the new API.
The only thing I would like is to make it clear in docs that the result must not be changed, and so it would be better to return a copy or an immutable object

sijie: @eolivelli 

> We can drop the annotation and let it 'public'. We can also add it in the new API.

I see the value of putting `getConf` in the api, I can see it is useful in multiple places, especially when you are using a client in a service component. But I don't want to raise the discussion of exposing this in api, because that should be done in a separate discussion. so I put `@VisibleForTesting` for this PR, which it is the purpose of `@VisibleForTesting`. If you feel really strong about `getConf` or want to get into the whole discussion of `Immutability` of configuration (which doesn't belong this change), I am fine reverting my test code to use whitebox.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookKeeperTest.java
```diff
@@ -105,9 +105,19 @@ public void testConstructionNotConnectedExplicitZk() throws Exception {
...
+
+    @Test
+    public void testBookkeeperDigestPasswordWithoutAutoDetection() throws Exception {
+        testBookkeeperDigestPassword(false);
+    }
+
+    void testBookkeeperDigestPassword(boolean autodetection) throws Exception {
         ClientConfiguration conf = new ClientConfiguration();
         conf.setZkServers(zkUtil.getZooKeeperConnectString());
+        conf.setEnableDigestTypeAutodetection(autodetection);
```
eolivelli: I think it is better to enhance MockBookKeeperTestCase and do not export getConf() in the scope of this change.

sijie: done

sijie:
retest this please

sijie:
retest this please

PR #1254, opened 2018-03-13
Fixes for multiple journals recovery

merlimat:
Currently, the implementation of multiple journals in bookie is broken because all journal instances are using the same `logMark` file to store the journal marker, stepping into each others.

Adding here a suffix for `logMark` to have unique files: `logMark.0`, `logMark.1` etc.
If there's 1 single journal, we'll keep using `logMark` so there's no compatibility issue.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
@@ -182,7 +182,7 @@ public class Journal extends BookieCriticalThread implements CheckpointSource {
             List<File> writableLedgerDirs = ledgerDirsManager
                     .getWritableLedgerDirs();
             for (File dir : writableLedgerDirs) {
-                File file = new File(dir, "lastMark");
+                File file = new File(dir, lastMarkFileName);
                 FileOutputStream fos = null;
                 try {
                     fos = new FileOutputStream(file);
@@ -211,7 +211,7 @@ public class Journal extends BookieCriticalThread implements CheckpointSource {
             ByteBuffer bb = ByteBuffer.wrap(buff);
             LogMark mark = new LogMark();
             for (File dir: ledgerDirsManager.getAllLedgerDirs()) {
-                File file = new File(dir, "lastMark");
+                File file = new File(dir, lastMarkFileName);
                 try {
                     try (FileInputStream fis = new FileInputStream(file)) {
                         int bytesRead = fis.read(buff);
@@ -576,6 +576,10 @@ public class Journal extends BookieCriticalThread implements CheckpointSource {
 
     private final LastLogMark lastLogMark = new LastLogMark(0, 0);
 
+    private static final String LAST_MARK_DEFAULT_NAME = "lastMark";
+
+    private final String lastMarkFileName;
+
     /**
      * The thread pool used to handle callback.
      */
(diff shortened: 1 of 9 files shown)


reddycharan:
@merlimat small request. Can you please wait for my PR to get merged - https://github.com/apache/bookkeeper/pull/1236 (it is almost approved and tests ran fine). It doesn't touch exactly these components, but indirectly tries to read logmark file in the testcases.

sijie:
@reddycharan I have merged #1236

sijie:
@merlimat LastMarkCommandTest.testCommand failed

ivankelly:
@merlimat I don't understand the initial problem. Surely the lastMark file it store in the specify journal directory, so the file path is different anyhow?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -634,6 +639,11 @@ public Journal(File journalDirectory, ServerConfiguration conf, LedgerDirsManage
 
         this.removePagesFromCache = conf.getJournalRemovePagesFromCache();
         // read last log mark
+        if (conf.getJournalDirs().length == 1) {
+            lastMarkFileName = LAST_MARK_DEFAULT_NAME;
+        } else {
+            lastMarkFileName = LAST_MARK_DEFAULT_NAME + "." + journalIndex;
```
eolivelli: What happens if the user changes the order of directories on the config file?

ivankelly: I think the cookie stops it booting (but we should doublecheck)

eolivelli:
@ivankelly last mask is in ledger dirs and not in journal dirs

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
merlimat: Yes the cookie is based on the list that comes from the config file, so changing the order will invalidate the cookie

reddycharan:
Hey @merlimat in https://github.com/apache/bookkeeper/commit/88addff33a9a9474b069ac197d46d2d672ae988d "lastMark" filename is hardcoded. Probably it also needs to be fixed.

merlimat:
> Hey @merlimat in 88addff "lastMark" filename is hardcoded. Probably it also needs to be fixed.

@reddycharan I don't think that is a problem. I don't see usages of harcoded `"lastMark" outside that checkpoint test, where's it's anyway fine since it's running with 1 single journal.

```
$ git grep '"lastMark"'
bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileSystemUpgrade.java:                    || name.equals("lastId") || name.startsWith("lastMark")) {
bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java:    private static final String LAST_MARK_DEFAULT_NAME = "lastMark";
bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/LedgerStorageCheckpointTest.java:        File lastMarkFile = new File(ledgerDir, "lastMark");
bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/LedgerStorageCheckpointTest.java:        File lastMarkFile = new File(ledgerDir, "lastMark");
bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/LedgerStorageCheckpointTest.java:        File lastMarkFile = new File(ledgerDir, "lastMark");
bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/LedgerStorageCheckpointTest.java:        File lastMarkFile = new File(ledgerDir, "lastMark");
bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/LedgerStorageCheckpointTest.java:        lastMarkFile = new File(ledgerDir, "lastMark");
```

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -634,6 +639,11 @@ public Journal(File journalDirectory, ServerConfiguration conf, LedgerDirsManage
 
         this.removePagesFromCache = conf.getJournalRemovePagesFromCache();
         // read last log mark
+        if (conf.getJournalDirs().length == 1) {
```
ivankelly: What happens when we have a bookie with multiple journals, it's been using lastMark up until now, this change goes in, and it doesn't find any lastMark.* file? We'll end up replaying the entire journal, no? Perhaps we should check if no lastMark.* file exists, if a lastMark file exists.

merlimat: The current state of "multiple journals" is very broken, since the bookie most likely fails to restart (since the last mark is can be pointing to a txn log that doesn't exist for a particular journal instance). I don't think it has ever been used for anything else than PoC perf testing at this point (or at least I hope no one was depending on this :) ).

sijie:
retest this please

PR #1256, opened 2018-03-13
Issue 1255: Bookie should not advance the journal marker before creating the index file

sijie:

*Problem*

Currently Bookie journal 'new ledger' entry if a ledger doesn't exist at ledger storage. This 'new ledger' entry is journaled before adding the entry to ledger storage. so this would cause a problem on checkpointing.

- journal 'new ledger' entry

```
    /**
     * Retrieve the ledger descriptor for the ledger which entry should be added to.
     * The LedgerDescriptor returned from this method should be eventually freed with
     * #putHandle().
     *
     * @throws BookieException if masterKey does not match the master key of the ledger
     */
    private LedgerDescriptor getLedgerForEntry(ByteBuf entry, final byte[] masterKey)
            throws IOException, BookieException {
        final long ledgerId = entry.getLong(entry.readerIndex());

        LedgerDescriptor l = handles.getHandle(ledgerId, masterKey);
        if (masterKeyCache.get(ledgerId) == null) {
            // Force the load into masterKey cache
            byte[] oldValue = masterKeyCache.putIfAbsent(ledgerId, masterKey);
            if (oldValue == null) {
                // new handle, we should add the key to journal ensure we can rebuild
                ByteBuffer bb = ByteBuffer.allocate(8 + 8 + 4 + masterKey.length);
                bb.putLong(ledgerId);
                bb.putLong(METAENTRY_ID_LEDGER_KEY);
                bb.putInt(masterKey.length);
                bb.put(masterKey);
                bb.flip();

                getJournal(ledgerId).logAddEntry(bb, false /* ackBeforeSync */, new NopWriteCallback(), null);
            }
        }

        return l;
    }
```

- add entry to ledger storage
```
    /**
     * Add an entry to a ledger as specified by handle.
     */
    private void addEntryInternal(LedgerDescriptor handle, ByteBuf entry,
                                  boolean ackBeforeSync, WriteCallback cb, Object ctx)
            throws IOException, BookieException {
        long ledgerId = handle.getLedgerId();
        long entryId = handle.addEntry(entry);

        writeBytes.add(entry.readableBytes());

        if (LOG.isTraceEnabled()) {
            LOG.trace("Adding {}@{}", entryId, ledgerId);
        }
        getJournal(ledgerId).logAddEntry(entry, ackBeforeSync, cb, ctx);
    }

```

The problematic sequence can be described as below:

- thread t1 is adding the first entry of ledger L1
- thread t2 is adding entries of ledger L2
- t1 is adding a journal entry of 'new ledger L1'
- t2 is adding entries after t1 adds the journal entry and before t1 adding the entry to ledger storage
- after t2 added entries, checkpoint happens in the ledger storage. it would roll the journal mark, which will claim the 'new ledger L1' entry as persistent.
- if the bookie restarts, it would fail with no such ledger exception.

The problem can be produced using a unit test: https://github.com/sijie/bookkeeper/commit/5053a717cd578aeb88236d373553d7494501b801

*Solution*

The fix is simple - just make sure the 'new ledger' journal entry is added after an entry is added to ledger storage. so it make sure when checkpoint happen it will flush and create the ledger before moving the journal mark.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SyncThread.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SyncThread.java
@@ -31,6 +31,8 @@ import java.util.concurrent.Future;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 
+import lombok.AccessLevel;
+import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 
 import org.apache.bookkeeper.bookie.CheckpointSource.Checkpoint;
@@ -60,6 +62,7 @@ import org.apache.bookkeeper.util.MathUtils;
 @Slf4j
 class SyncThread implements Checkpointer {
 
+    @Getter(AccessLevel.PACKAGE)
     final ScheduledExecutorService executor;
     final int flushInterval;
     final LedgerStorage ledgerStorage;
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/CheckpointOnNewLedgersTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/CheckpointOnNewLedgersTest.java
@@ -65,6 +65,7 @@ public class CheckpointOnNewLedgersTest {
         conf = TestBKConfiguration.newServerConfiguration();
         conf.setGcWaitTime(gcWaitTime);
         conf.setLedgerStorageClass(InterleavedLedgerStorage.class.getName());
+        conf.setJournalDirsName(new String[] { bkDir.toString() });
         conf.setLedgerDirNames(new String[] { bkDir.toString() });
         conf.setEntryLogSizeLimit(10 * 1024);
 
@@ -176,6 +177,14 @@ public class CheckpointOnNewLedgersTest {
         // construct a new bookie to simulate "bookie restart from crash"
         Bookie newBookie = new Bookie(conf);
         newBookie.start();
+
+        for (int i = 0; i < numEntries; i++) {
+            ByteBuf entry = newBookie.readEntry(l2, i);
+            assertNotNull(entry);
+            assertEquals(l2, entry.readLong());
+            assertEquals((long) i, entry.readLong());
+            entry.release();
+        }
 
         ByteBuf entry = newBookie.readEntry(l1, 0L);
         assertNotNull(entry);
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/CheckpointOnNewLedgersTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/CheckpointOnNewLedgersTest.java
@@ -1,0 +1,188 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.bookkeeper.bookie;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.spy;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.Unpooled;
+import java.io.File;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ThreadLocalRandom;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.bookkeeper.bookie.Journal.LastLogMark;
+import org.apache.bookkeeper.bookie.storage.ldb.DbLedgerStorage;
+import org.apache.bookkeeper.conf.ServerConfiguration;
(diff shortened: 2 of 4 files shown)


sijie:
@yzang : this PR address the issue that we saw at bookkeeper

yzang:
This issue is keep happening at Twitter, the error we keep getting is:

> Exception while replaying journals, shutting down
> java.io.IOException: Missing ledger signature
> 	at org.apache.bookkeeper.bookie.FileInfo.readHeader(FileInfo.java:153)
> 	at org.apache.bookkeeper.bookie.FileInfo.checkOpen(FileInfo.java:205)
> 	at org.apache.bookkeeper.bookie.FileInfo.checkOpen(FileInfo.java:176)
> 	at org.apache.bookkeeper.bookie.FileInfo.getMasterKey(FileInfo.java:428)
> 	at org.apache.bookkeeper.bookie.IndexPersistenceMgr.readMasterKey(IndexPersistenceMgr.java:387)
> 	at org.apache.bookkeeper.bookie.LedgerCacheImpl.readMasterKey(LedgerCacheImpl.java:154)
> 	at org.apache.bookkeeper.bookie.InterleavedLedgerStorage.readMasterKey(InterleavedLedgerStorage.java:174)
> 	at org.apache.bookkeeper.bookie.SortedLedgerStorage.readMasterKey(SortedLedgerStorage.java:38)
> 	at org.apache.bookkeeper.bookie.Bookie$3.process(Bookie.java:663)
> 	at org.apache.bookkeeper.bookie.Journal.scanJournal(Journal.java:677)
> 	at org.apache.bookkeeper.bookie.Journal.replay(Journal.java:721)
> 	at org.apache.bookkeeper.bookie.Bookie.readJournal(Bookie.java:628)
> 	at org.apache.bookkeeper.bookie.Bookie.start(Bookie.java:747)
> 	at org.apache.bookkeeper.proto.BookieServer.start(BookieServer.java:114)
> 	at org.apache.bookkeeper.proto.BookieServer.main(BookieServer.java:369)

Really nice to have this bug to be fixed!

sijie:
@yzang it would be nice to apply this change (and canary it at twitter's environment) to see if it can address the problems there.

jvrao:
@reddycharan please review this change.

reddycharan:
LGTM.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SyncThread.java
```diff
@@ -60,6 +62,7 @@
 @Slf4j
 class SyncThread implements Checkpointer {
 
+    @Getter(AccessLevel.PACKAGE)
```
ivankelly: why even add a getter? it's final

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/CheckpointOnNewLedgersTest.java
```diff
@@ -0,0 +1,197 @@
...
+    private CountDownLatch getLedgerDescCalledLatch;
+    private CountDownLatch getLedgerDescWaitLatch;
+
+    @Before
+    public void setup() throws Exception {
+        File bkDir = testDir.newFolder("dbLedgerStorageCheckpointTest");
+        File curDir = Bookie.getCurrentDirectory(bkDir);
+        Bookie.checkDirectoryStructure(curDir);
+
+        int gcWaitTime = 1000;
```
ivankelly: why run gc at all? it can only interfere with what you're trying to test.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/CheckpointOnNewLedgersTest.java
```diff
@@ -0,0 +1,197 @@
...
+        buffer.writerIndex(0);
+        buffer.writeLong(ledgerId);
+        buffer.writeLong(entryId);
+        buffer.writeLong(entryId - 1); // lac
+        buffer.writerIndex(entrySize);
+        return buffer;
+    }
+
+    @Test
+    public void testCheckpoint() throws Exception {
```
ivankelly: This test looks like you're trying to hit a very specific sequence of events and thread interleaving. It would be clearer to not use threads at all and directly call the interleaving you as looking for.

yzang:
@sijie Yeah, we'll cherry pick the change and try a deploy to see if problem is fixed.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SyncThread.java
sijie: It is just personal practice. I generally dislike referencing field directly, which is not friend to testing at all.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/CheckpointOnNewLedgersTest.java
sijie: GC will not interfere the test. No matter I set the value of not, the test will always run as expected. I bring in gc to bring in chaos for testing. That said I don’t rely gc on testing. But if the test fails, that means something big to look into.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/CheckpointOnNewLedgersTest.java
sijie: You have to use threads, you can’t call directly. Because you need to bring in the race conditions to trigger this sequence, and you have to do that in a separate thread.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/CheckpointOnNewLedgersTest.java
ivankelly: In practice, if the test fails, we'll run it again, and not look into it. unit tests aren't really the place to adding fuzz.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/CheckpointOnNewLedgersTest.java
ivankelly: I'm not so sure, but anyhow, as I said, non-blocking. This can be merged now.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SyncThread.java
ivankelly: Sure. Not a fan of lombok, which is why I asked. We should eventually make construction of all these executors at the top level and pass them in via the constructor so we can mock them and get rid of non-determinism.

PR #1258, opened 2018-03-14
Allow to send callbacks from journal thread

merlimat:
In BK 4.3 it was possible to configure 0 threads for the dispatching of journal callback threads. 

There might be good reasons to not use a thread pool when dispatching the callback threads: 
 * Avoiding one more context switch 
 * Avoiding contention on the executor enqueuing
 * When using multiple journals, the journal thread is not a bottleneck anymore

We should allow the same convention (threads=0 means direct execution) that we have in other places.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
@@ -23,11 +23,13 @@ package org.apache.bookkeeper.bookie;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Stopwatch;
+import com.google.common.util.concurrent.MoreExecutors;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
 import io.netty.util.Recycler;
 import io.netty.util.Recycler.Handle;
+import io.netty.util.concurrent.DefaultThreadFactory;
 
 import java.io.File;
 import java.io.FileInputStream;
@@ -50,7 +52,6 @@ import org.apache.bookkeeper.stats.Counter;
 import org.apache.bookkeeper.stats.NullStatsLogger;
 import org.apache.bookkeeper.stats.OpStatsLogger;
 import org.apache.bookkeeper.stats.StatsLogger;
-import org.apache.bookkeeper.util.DaemonThreadFactory;
 import org.apache.bookkeeper.util.IOUtils;
 import org.apache.bookkeeper.util.MathUtils;
 import org.apache.bookkeeper.util.collections.GrowableArrayBlockingQueue;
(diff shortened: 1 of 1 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -625,8 +626,12 @@ public Journal(File journalDirectory, ServerConfiguration conf, LedgerDirsManage
         this.maxGroupWaitInNanos = TimeUnit.MILLISECONDS.toNanos(conf.getJournalMaxGroupWaitMSec());
         this.bufferedWritesThreshold = conf.getJournalBufferedWritesThreshold();
         this.bufferedEntriesThreshold = conf.getJournalBufferedEntriesThreshold();
-        this.cbThreadPool = Executors.newFixedThreadPool(conf.getNumJournalCallbackThreads(),
-                                                         new DaemonThreadFactory());
+        if (conf.getNumJournalCallbackThreads() > 0) {
+            this.cbThreadPool = Executors.newFixedThreadPool(conf.getNumJournalCallbackThreads(),
+                                                         new DefaultThreadFactory("bookie-journal-callback"));
+        } else {
+            this.cbThreadPool = MoreExecutors.newDirectExecutorService();
```
ivankelly: this thread pool gets used on two different threads(JournalThread & ForceWriteThread). Are you sure this is safe?

merlimat: I think it would be anyway either one case or the other. 

I think though that it should be safe anyway. If you have a "direct" executor, it is like a dummy executor. 

We could have threads `t1` and `t2` sharing the same instance. when `t1` submits one task, it gets executed immediately in `t1`. `t2` tasks will be executed in `t2` thread.

ivankelly: Previously I think it wasn't possible. if getNumJournalCallbackThreads was 0, then we would throw an exception creating the executor. If 1, then callbacks would be serialized. That said, if you had getNumJournalCallbackThreads > 1, you have no guarantee where you run, they would have to be thread safe anyhow.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -625,8 +626,12 @@ public Journal(File journalDirectory, ServerConfiguration conf, LedgerDirsManage
         this.maxGroupWaitInNanos = TimeUnit.MILLISECONDS.toNanos(conf.getJournalMaxGroupWaitMSec());
         this.bufferedWritesThreshold = conf.getJournalBufferedWritesThreshold();
         this.bufferedEntriesThreshold = conf.getJournalBufferedEntriesThreshold();
-        this.cbThreadPool = Executors.newFixedThreadPool(conf.getNumJournalCallbackThreads(),
-                                                         new DaemonThreadFactory());
+        if (conf.getNumJournalCallbackThreads() > 0) {
+            this.cbThreadPool = Executors.newFixedThreadPool(conf.getNumJournalCallbackThreads(),
```
sijie: interesting.. I thought we used OrderedScheduler here as well.

I actually checked twitter's branch, it does use ordered scheduler. I believe cbThreadPool was first introduced by twitter sometime back. 

https://github.com/twitter/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java#L577

sijie: the change looks good to me though. I just found this difference here.

sijie:
retest this please.

PR #1260, opened 2018-03-14
Recycle the AddResponse in WriteEntryProcessor

merlimat:
Instances of `AddResponse` are coming from a recycler object pool but we're not recycling them when we're done with them.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/WriteEntryProcessor.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/WriteEntryProcessor.java
@@ -28,6 +28,7 @@ import java.util.concurrent.TimeUnit;
 import org.apache.bookkeeper.bookie.BookieException;
 import org.apache.bookkeeper.net.BookieSocketAddress;
 import org.apache.bookkeeper.proto.BookieProtocol.ParsedAddRequest;
+import org.apache.bookkeeper.proto.BookieProtocol.Response;
 import org.apache.bookkeeper.proto.BookkeeperInternalCallbacks.WriteCallback;
 import org.apache.bookkeeper.util.MathUtils;
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


eolivelli:
@sijie  @merlimat  @ivankelly  are you OK with cherry pick to 4.6 branch, this is a importnat memory leak on bookie, no ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/WriteEntryProcessor.java
```diff
@@ -114,9 +115,10 @@ public void writeComplete(int rc, long ledgerId, long entryId,
             requestProcessor.addEntryStats.registerFailedEvent(MathUtils.elapsedNanos(startTimeNanos),
                     TimeUnit.NANOSECONDS);
         }
-        sendResponse(rc,
-                     ResponseBuilder.buildAddResponse(request),
-                     requestProcessor.addRequestStats);
+
+        Response response = ResponseBuilder.buildAddResponse(request);
+        sendResponse(rc, response, requestProcessor.addRequestStats);
```
ivankelly: sendResponse calls channel.writeAndFlush(response...). Are we sure that the writeAndFlush isn't keeping a reference to response somewhere?

ivankelly:
@eolivelli no memory leak here. There's no bytebuf in AddResponse, so it just means a recyclable object is being left for GC to clean up.

merlimat:
> are you OK with cherry pick to 4.6 branch, this is a importnat memory leak on bookie, no ?

@eolivelli "forgetting" to recycle doesn't leave a memory leak. It just doesn't take advantage of the pooling. 

You get an object from the recycler. If that is empty, it will allocated a new one (plust a `Handle` instance to track it). If it doesn't go back to pool, both the object and the "handle" are going to be be GCed. 

It's typically ok to not recycle on uncommon error paths for example.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/WriteEntryProcessor.java
merlimat: I believe that when the flush is called, the whole chain of handlers is invoked. Therefore, even if the channel is not writable, we should already have converted the add response into a buffer.

eolivelli:
@merlimat  @ivankelly  thank you for clarification, I took a look to Recycler implementation now (never did before)

sijie:
retest this please

merlimat:
Scratch this PR. This is actually wrong as tipped by the broken unit tests. 

The instances are actually already being recycled in `BookieProtoEncoding`. What I was seeing in the profiler is probably due that, with more than 1K instances in-flight, the recycler is actually throwing them away instead of putting them back to the pool. I'll try to understand a bit more but this change is wrong.

Issue #881, opened 2017-12-19
Separate initialize from format in docker image


sijie:
**FEATURE REQUEST**

1. Please describe the feature you are requesting.

Currently docker image uses `format` during startup to initialize the cluster. It would be good to separate `initialize` from `format`. The `initialize` command should only *create* znodes for brand new environment. It should not involve any logic on `reformatting`. 

Also make `initialize` an atomic operation.

2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have). Are you currently using any workarounds to address this issue?

*should-have*

3. Provide any additional detail on your proposed use case for this feature.

N/A

sijie:
The `metaformat` command has been implemented with two commands `initnewcluster ` and `nukeexistingcluster` in #978. we should update the docker file to use `initnewcluster` command.

PR #1263, opened 2018-03-15
Issue 881: use initialize insteadof format in docker image

jiazhai:

Currently docker image uses format during startup to initialize the cluster. It would be good to separate initialize from format. The initialize command should only create znodes for brand new environment. It should not involve any logic on reformatting.

- use initialize insteadof format in docker image.
- make initialize an atomic operation.
- add initialize and nuke command in makefile.
- fix error for non root user running.

This PR tested by local build, since 4.7 is not released, will bump bk version in Dockerfile once 4.7 released.

Master Issue: #881

### DIFF
--- a/docker/scripts/entrypoint.sh
+++ b/docker/scripts/entrypoint.sh
@@ -47,7 +47,9 @@ mkdir -p "${BK_journalDirectory}" "${BK_ledgerDirectories}" "${BK_indexDirectori
 # -------------- #
 # Allow the container to be started with `--user`
 if [ "$1" = '/opt/bookkeeper/bin/bookkeeper' -a "$(id -u)" = '0' ]; then
+    echo "This is root, will use user $BK_USER to run it"
     chown -R "$BK_USER:$BK_USER" "/opt/bookkeeper/" "${BK_journalDirectory}" "${BK_ledgerDirectories}" "${BK_indexDirectories}"
+    chmod -R +x "/opt/bookkeeper/"
     sudo -s -E -u "$BK_USER" /bin/bash "$0" "$@"
     exit
 fi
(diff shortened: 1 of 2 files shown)


sijie:
re-ran IT test - https://builds.apache.org/job/bookkeeper_precommit_integrationtests/359/

docker/scripts/entrypoint.sh
```diff
@@ -61,11 +63,50 @@ until /opt/bookkeeper/bin/bookkeeper org.apache.zookeeper.ZooKeeperMain -server
...
+        fi
+    else
+        echo "Other docker instance is doing initialize at the same time, will wait in this instance."
+        tenSeconds=1
+        while [ ${tenSeconds} -lt 10 ]
+        do
+            sleep 10
+            /opt/bookkeeper/bin/bookkeeper org.apache.zookeeper.ZooKeeperMain -server ${BK_zkServers} stat ${BK_zkLedgersRootPath}/available/readonly
+            if [ $? -eq 0 ]; then
+                echo "Waited $tenSeconds * 10 seconds, bookkeeper inited"
```
sijie: I don't think we need "Waited $tenSeconds * 10 seconds". Here the bookkeeper initialization is already done.

docker/scripts/entrypoint.sh
```diff
@@ -61,11 +63,50 @@ until /opt/bookkeeper/bin/bookkeeper org.apache.zookeeper.ZooKeeperMain -server
...
-/opt/bookkeeper/bin/bookkeeper shell metaformat -n || true
+
+# Init the cluster if required znodes not exist in Zookeeper.
+# Use ephemeral zk node as lock to keep initialize atomic.
+/opt/bookkeeper/bin/bookkeeper org.apache.zookeeper.ZooKeeperMain -server ${BK_zkServers} stat ${BK_zkLedgersRootPath}/available/readonly
+if [ $? -eq 0 ]; then
+    echo "Metadata of cluster already exists, no need format"
+else
+    # create ephemeral zk node bkInitLock, initiator who this node, then do init; other initiators will wait.
+    /opt/bookkeeper/bin/bookkeeper org.apache.zookeeper.ZooKeeperMain -server ${BK_zkServers} create -e ${BK_CLUSTER_ROOT_PATH}/bkInitLock
```
sijie: if there are two concurrent initialization, can the second one wait and retry?

jiazhai: Thanks, the second will wait the first.

docker/scripts/entrypoint.sh
jiazhai: Thanks. It is for concurrent initialization. This is not the successor for the init, it should wait the first complete. here it waited and know that initialization is already done, so it will break wait in next line.

sijie:
retest this please

PR #1264, opened 2018-03-15
Refactored LAC cache in DbLedgerStorage

merlimat:
The `TransientLedgerInfo` cache that holds the LAC (or explicit lac) for all ledgers has showed up in the profiler as a big contention point. 

### Modifications 

 * Only add an item in `TransientLedgerInfo` is someone has asked for LAC or has set explicity LAC. This will save lot of memory if many ledger are not requiring the LAC info.
 * Use `ConcurrentLongHashMap` to store the `TransientLedgerInfo`. There are few advantages: 
   - Key in the map is `long` and doesn't need boxing
   - The `get()` operation uses a stamped lock which is ideal in cases the number of `get()` is way bigger than `put()` operations, since it only needs a volatile variable read. In this case the map is only updated when new ledgers are added / removed, but we do a `get()` on each `addEntry()` operation.
  - Schedule periodical cleanup of the map for stale ledger info entries

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
@@ -25,28 +25,27 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import static org.apache.bookkeeper.bookie.LastAddConfirmedUpdateNotification.WATCHER_RECYCLER;
 
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.cache.CacheBuilder;
-import com.google.common.cache.CacheLoader;
-import com.google.common.cache.LoadingCache;
-import com.google.common.cache.RemovalListener;
-import com.google.common.cache.RemovalNotification;
 import com.google.common.collect.Lists;
 import com.google.protobuf.ByteString;
+
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
 import io.netty.util.concurrent.DefaultThreadFactory;
+
 import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.util.SortedMap;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.locks.Condition;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
+
 import org.apache.bookkeeper.bookie.Bookie;
 import org.apache.bookkeeper.bookie.Bookie.NoEntryException;
 import org.apache.bookkeeper.bookie.BookieException;
@@ -74,6 +73,7 @@ import org.apache.bookkeeper.stats.OpStatsLogger;
 import org.apache.bookkeeper.stats.StatsLogger;
 import org.apache.bookkeeper.util.DiskChecker;
 import org.apache.bookkeeper.util.MathUtils;
+import org.apache.bookkeeper.util.collections.ConcurrentLongHashMap;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -93,7 +93,7 @@ public class DbLedgerStorage implements CompactableLedgerStorage {
             implements AutoCloseable {
 
         // lac
-        private Long lac = null;
+        private volatile long lac = -1;
         // request from explicit lac requests
         private ByteBuffer explicitLac = null;
         // is the ledger info closed?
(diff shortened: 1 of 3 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
```diff
@@ -197,7 +208,9 @@ public void close() {
 
     private LedgerMetadataIndex ledgerIndex;
     private EntryLocationIndex entryLocationIndex;
-    private LoadingCache<Long, TransientLedgerInfo> transientLedgerInfoCache;
+
+    private static final long LEDGER_INFO_CACHING_TIME_MINUTES = 10;
```
sijie: Can we use `conf.getFileInfoMaxIdleTime` so it is configurable?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
```diff
@@ -287,23 +300,10 @@ public void initialize(ServerConfiguration conf, LedgerManager ledgerManager, Le
...
-            builder.expireAfterAccess(conf.getFileInfoMaxIdleTime(), TimeUnit.SECONDS);
-        }
-        transientLedgerInfoCache = builder.build(new CacheLoader<Long, TransientLedgerInfo>() {
-            @Override
-            public TransientLedgerInfo load(Long key) throws Exception {
-                return new TransientLedgerInfo(key, ledgerIndex);
-            }
-        });
+        transientLedgerInfoCache = new ConcurrentLongHashMap<>(16 * 1024,
+                Runtime.getRuntime().availableProcessors() * 2);
```
sijie: better to configure the concurrency level to be ` Math.max(1, Math.max(conf.getNumAddWorkerThreads(), conf.getNumReadWorkerThreads()))`?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
```diff
@@ -894,26 +901,43 @@ public long getLastAddConfirmed(long ledgerId) throws IOException {
...
     }
 
+    private TransientLedgerInfo getOrAddLedgerInfo(long ledgerId) {
+        TransientLedgerInfo tli = transientLedgerInfoCache.get(ledgerId);
+        if (tli != null) {
+            return tli;
+        } else {
+            TransientLedgerInfo newTli = new TransientLedgerInfo(ledgerId, ledgerIndex);
+            tli = transientLedgerInfoCache.putIfAbsent(ledgerId, newTli);
+            return tli != null ? tli : newTli;
```
sijie: better to close `newTli`?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
```diff
@@ -93,7 +93,7 @@
             implements AutoCloseable {
 
         // lac
-        private Long lac = null;
+        private volatile long lac = -1;
```
sijie: `-1L` has a meaning because `addEntry(0L)` carries -1L. I would suggest to use a different negative value, e.g. `-0xdead` or `Long.MIN_VALUE`.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
merlimat: The default for `getFileInfoMaxIdleTime` and I think we should have some time based eviction from this cache by default. The bookie might be storing many ledgers, though only a fraction of them might be actually be "active".

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
```diff
@@ -103,26 +103,30 @@
         // reference to LedgerMetadataIndex
         private final LedgerMetadataIndex ledgerIndex;
 
+        private long lastUpdated;
```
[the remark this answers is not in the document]
merlimat: fixed

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
merlimat: One thing to consider here is that if the `addWorkerThreads` is 0, it means the writes are coming directly from the io threads, and there the number depends on the number of cores.

for the `ConcurrentLongHashMap` the "parallelism" equates to the number of sections in the hash-map. Each section has its own read-write mutex. Since we only have 1 instance of this map per bookie, I don't think it's a problem to have multiple sections, since the memory increase is really tiny.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
merlimat: fixed

PR #1266, opened 2018-03-16
Fixed buffer release in v2 BookieProtocol

merlimat:
There was a buffers leak when using V2 bookie protocol. This was introduced with the refactoring and the merges from yahoo branch. 

Note: this was only happening when selecting  v2 protocol in client configuration.

The current code was missing a release in both the add and read entries path. I've been running it for some time and the ref-counting seems correct after the fix.


bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
```diff
@@ -1138,6 +1138,7 @@ static ReadV2ResponseCallback create(CompletionValue completionValue, long ledge
         @Override
         public void safeRun() {
             completionValue.handleV2Response(ledgerId, entryId, status, response);
+            response.release();
```
sijie: can we combine `release` and `recycle` like what we did for request?

PR #1274, opened 2018-03-17
Ensure we instantiate threads that extend Netty FastThreadLocalThread

merlimat:
Since we make heavy use of `Recycler` and other `FastThreadLocal` objects from Netty, we should make sure that all (critical) threads that we instantiate are inheriting from `FastThreadLocalThread`.

`FastThreadLocalThread` has an optimization that makes access to `FastThreadLocal` faster. It's a class that extends regular `Thread` and adds a field for the thread local access.

There are a couple of places where we're currently not using `FastThreadLocal`: 
 * `OrderedScheduler` builder
 * `BookieThread` (which is used in `Journal`, `ForceSyncThread`, etc..)


jvrao:
LGTM

PR #1268, opened 2018-03-16
(WIP) Enable bytebuf leak detection in tests

sijie:
Descriptions of the changes in this PR:

distributedlog already enables leak detection at some modules. it would be good to enable leak detection for all the modules across the project.

> ---
> Be sure to do all of the following to help us incorporate your contribution
> quickly and easily:
>
> If this PR is a BookKeeper Proposal (BP):
>
> - [ ] Make sure the PR title is formatted like:
>     `<BP-#>: Description of bookkeeper proposal`
>     `e.g. BP-1: 64 bits ledger is support`
> - [ ] Attach the master issue link in the description of this PR.
> - [ ] Attach the google doc link if the BP is written in Google Doc.
>
> Otherwise:
> 
> - [ ] Make sure the PR title is formatted like:
>     `<Issue #>: Description of pull request`
>     `e.g. Issue 123: Description ...`
> - [ ] Make sure tests pass via `mvn clean apache-rat:check install spotbugs:check`.
> - [ ] Replace `<Issue #>` in the title with the actual Issue number.
> 
> ---

PR #1275, opened 2018-03-19
Fix ByteBuf leaking at Journal

sijie:

BufferLogChannel uses a pooled bytebuf for buffering writes. However Journal only closes the underlying log file channel, but it doesn't close buffer log channel, which can cause bytebuf leaking.

This problem get exposed at #1268

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
@@ -904,6 +904,7 @@ public class Journal extends BookieCriticalThread implements CheckpointSource {
         paddingBuff.writeZero(paddingBuff.capacity());
         final int journalFormatVersionToWrite = conf.getJournalFormatVersionToWrite();
         final int journalAlignmentSize = conf.getJournalAlignmentSize();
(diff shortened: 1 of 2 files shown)


sijie:
No test is added here, because:

1) it is a bit to tricky to test "bytebuf is leak". 
2) if you enable leak detection as what #1268 does, and run BookieJournalRollingTest, you will see `LEAK` in the logging. After apply this change, `LEAK` loggings are gone. so effectively `BookieJournalRollingTest` is testing the case here.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -1089,8 +1093,6 @@ public void run() {
                 numEntriesToFlush++;
                 qe = null;
             }
-            logFile.close();
```
eolivelli: @sijie do you think this is reduntant ?

sijie: They are closed at the finally block.

jvrao:
@cguttapalem Plese review this.

PR #1276, opened 2018-03-19
Fix logging issue on PendingReadOp

sijie:

`EntryImpl` is recycled after callback is submitted for a given read op. However the object might be referenced after it is recycled, since read callback can come back after the read op is error-out. 

This doesn't affect any operation or correctness. However it might introduce some confused logging.

This also got exposed by #1268


sijie:
No new test is added because this is not a correctness. It has confused logging when running TestMaxSizeWorkersQueue with leak detection enabled. After applying this change, the confused loggings are gone.

PR #1277, opened 2018-03-19
WIP - Use ByteBufOutputStream to serialize ProtoBuf messages

eolivelli:
This patch fixes an issue found on a staging environment running on Java 9.

Responses and requests encoded using direct access to the underlying ByteBuf#array() result as corrupted packets on the receiving peer.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtoEncoding.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtoEncoding.java
@@ -20,7 +20,6 @@
  */
 package org.apache.bookkeeper.proto;
 
-import com.google.protobuf.CodedOutputStream;
 import com.google.protobuf.ExtensionRegistry;
 import com.google.protobuf.InvalidProtocolBufferException;
 import com.google.protobuf.MessageLite;
@@ -356,7 +355,7 @@ public class BookieProtoEncoding {
         ByteBuf buf = allocator.heapBuffer(size, size);
 
         try {
-            msg.writeTo(CodedOutputStream.newInstance(buf.array(), buf.arrayOffset() + buf.writerIndex(), size));
+            msg.writeTo(new ByteBufOutputStream(buf));
         } catch (IOException e) {
             // This is in-memory serialization, should not fail
             throw new RuntimeException(e);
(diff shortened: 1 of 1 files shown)


eolivelli:
I am marking this PR as Work-in-progress in order to gather reviews and comments.

This fix makes the staging environment work on Java 9.

This is the email thread on user list
http://mail-archives.apache.org/mod_mbox/bookkeeper-user/201803.mbox/%3CCACcefge48TTvEocqDbP%2BOCsBH6tuOOpJSS_tLnmHX3bKxXLysg%40mail.gmail.com%3E

Another fix is to use Unpooled ByteBufs instead of using the ByteBufAllocator returned by the Channel.

This fix may impact somehow performances because previously we were writing directly to the byte[] inside the ByteBuf.

This fix may open the way to using direct memory buffers on the v3 encoder as an heapBuffer is no more needed

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtoEncoding.java
```diff
@@ -356,7 +355,7 @@ private static ByteBuf serializeProtobuf(MessageLite msg, ByteBufAllocator alloc
         ByteBuf buf = allocator.heapBuffer(size, size);
 
         try {
-            msg.writeTo(CodedOutputStream.newInstance(buf.array(), buf.arrayOffset() + buf.writerIndex(), size));
+            msg.writeTo(new ByteBufOutputStream(buf));
```
jvrao: Is this a race or corruption happens all the time?  Change looks fine, but I did not understand the actual issue with the current code on why corruption happening.

sijie: @jvrao there is no conclusion yet. this is just marked as WIP.

eolivelli:
It seems that this code is working as well:


```
private static ByteBuf serializeProtobuf(MessageLite msg, ByteBufAllocator allocator) {
        int size = msg.getSerializedSize();
        ByteBuf buf = allocator.heapBuffer(size, size);                
        try {
            msg.writeTo(CodedOutputStream.newInstance(buf.array(), buf.arrayOffset() + buf.writerIndex(), size));
            buf.writerIndex(buf.capacity());
        } catch (IOException e) {
            // This is in-memory serialization, should not fail
            throw new RuntimeException(e);
        }
        for (int i = 0; i < size; i++) {
        }
        return buf;
    }
```

It seems that adding some code after the write makes the system work as expected.
Typical message size is about 120 bytes , ad we are talking about READ_ENTRY responses

as reported on Slack I have added some assertions, but just adding assertions makes the system work

eolivelli:
Closing this PR in favour of #1286

PR #1280, opened 2018-03-20
Improve write rejection in DbLedgerStorage 

merlimat:
This PR contain 2 changes that are much related to each other. 

1. Changed DbLedgerStorage write cache read-write lock into a `StampedLock`
   Main reason for using `StampedLock` is that this lock is always taken as a "read" lock, when putting entries in the the cache. It is protecting for the swapping of the write cache when the flush is triggered. Most of the time, we have many threads acquiring the read-lock. `StampedLock` is better in this scenario because there isn't any contention in the optimistic path (just a double volatile read).

2. Have a configurable throttle timing when the write cache is really full. 
    If we try to write faster than we can put data in the storage device, eventually both write caches will get full and we have to throttle writes. Added here a setting to configure the amount of time to block the thread (who's calling `Bookie.addEntry()`, it could be IO thread or a addEntryWorker depending on configuration).
   After the timeout, if we still weren't able to flush the write cache and insert the current entry, bookie will reject the write operation, without logging but rather using a counter to keep track of it.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieException.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieException.java
@@ -82,6 +82,7 @@ public abstract class BookieException extends Exception {
         int CookieNotFoundException = -105;
         int MetadataStoreException = -106;
         int UnknownBookieIdException = -107;
+        int OperationRejectedException = -108;
     }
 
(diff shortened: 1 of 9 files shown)


jiazhai:
seems CI failed.

merlimat:
> seems CI failed. 

Yes there was one test to adapt since the exception type changed. Fixed now.

Issue #659, opened 2017-10-24
Checkpoint logic is broken in SortedLedgerStorage


sijie:
*Problem*

Currently SortedLedgerStorage doesn't maintain any checkpoint related information. It shares the same `checkpoint holder` with InterleavedLedgerStorage. The checkpoint holder is updated when an entry log file is rotated. Unfortunately this can happen when compaction adds entries in the entry log files causing entry log file rotated and SortedLedgerStorage isn't aware of this. This can cause data not durably persisted in ledger storage before advancing journal checkpoint marker.

*Discussion*

The detail discussion can be found at:

http://mail-archives.apache.org/mod_mbox/bookkeeper-dev/201710.mbox/%3CCAAFz1KPbOZjSnmpHDt%2BdeU7FPQvv__XFPc_a69zhd31TuSFJ_g%40mail.gmail.com%3E

http://mail-archives.apache.org/mod_mbox/bookkeeper-dev/201710.mbox/%3CCAO2yDyZ-_3N6_YMGuG%3DG0k3Q-omDRxshBmegv%3DSMLb7dmsiBmQ%40mail.gmail.com%3E

*Solutions*

```
   1. Use the approach that I used before (as in the old commits at
   twitter’s branch). Ledger storage is responsible for instantiating the
   checkpoints.
   2. Sorted ledger storage maintains a separate checkpoint marker (can be
   using the checkpoint markers in memtables).
   3. Compaction avoids advancing checkpoint holder in entry log files.
   This can be covered by
   https://cwiki.apache.org/confluence/display/BOOKKEEPER/BP-6+-+Use+separate+log+for+compaction
```

PR #1281, opened 2018-03-21
Issue #570: Introducing EntryLogManager.

reddycharan:

Introducing EntryLogManager interface, which abstracts out current activeLogChannel,
rotatedLogChannels and corresponding lock for activeLogChannel. The current logic of
handling logs is moved to EntryLogManagerForSingleEntryLog class, in the
next sub-task EntryLogManagerForEntryLogPerLedger will be introduced. Also there
are minor changes to createNewLog logic and leastUnflushedLogId logic.

This is < sub-task5  > of Issue #570

Master Issue: #570

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
@@ -67,9 +67,9 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicBoolean;
 
+import org.apache.bookkeeper.bookie.LedgerDirsManager.LedgerDirsListener;
 import org.apache.bookkeeper.conf.ServerConfiguration;
 import org.apache.bookkeeper.util.IOUtils;
 import org.apache.bookkeeper.util.collections.ConcurrentLongLongHashMap;
@@ -386,7 +386,7 @@ public class EntryLogger {
         this.recentlyCreatedEntryLogsStatus = new RecentEntryLogsStatus(logId + 1);
         this.entryLoggerAllocator = new EntryLoggerAllocator(logId);
         if (entryLogPerLedgerEnabled) {
-            this.entryLogManager = new EntryLogManagerForSingleEntryLog() {
+            this.entryLogManager = new EntryLogManagerForSingleEntryLog(ledgerDirsManager) {
                 @Override
                 public void checkpoint() throws IOException {
                     /*
@@ -443,7 +443,7 @@ public class EntryLogger {
                 }
             };
         } else {
-            this.entryLogManager = new EntryLogManagerForSingleEntryLog();
+            this.entryLogManager = new EntryLogManagerForSingleEntryLog(ledgerDirsManager);
         }
     }
 
@@ -587,139 +587,110 @@ public class EntryLogger {
     class EntryLoggerAllocator {
 
         private long preallocatedLogId;
-        Future<BufferedLogChannel> preallocation = null;
+        private Future<BufferedLogChannel> preallocation = null;
         private ExecutorService allocatorExecutor;
+        private final Object createEntryLogLock = new Object();
+        private final Object createCompactionLogLock = new Object();
 
         EntryLoggerAllocator(long logId) {
             preallocatedLogId = logId;
             allocatorExecutor = Executors.newSingleThreadExecutor();
         }
 
-        synchronized long getPreallocatedLogId(){
+        synchronized long getPreallocatedLogId() {
             return preallocatedLogId;
         }
 
-        synchronized BufferedLogChannel createNewLog() throws IOException {
-            BufferedLogChannel bc;
-            if (!entryLogPreAllocationEnabled || null == preallocation) {
-                // initialization time to create a new log
-                bc = allocateNewLog();
-            } else {
-                // has a preallocated entry log
-                try {
-                    /*
-                     * both createNewLog and allocateNewLog are synchronized on
-                     * EntryLoggerAllocator.this object. So it is possible that
-                     * a thread calling createNewLog would attain the lock on
-                     * this object and get to this point but preallocation
-                     * Future is starving for lock on EntryLoggerAllocator.this
-                     * to execute allocateNewLog. Here since we attained lock
-                     * for this it means preallocation future must have either
-                     * completed creating new log or still waiting for lock on
-                     * this object to execute allocateNewLog method. So we
-                     * should try to get result of the future without waiting.
-                     * If it fails with TimeoutException then call
-                     * allocateNewLog explicitly since we are holding the lock
-                     * on this anyhow.
-                     *
-                     */
-                    bc = preallocation.get(0, TimeUnit.MILLISECONDS);
-                } catch (ExecutionException ee) {
-                    if (ee.getCause() instanceof IOException) {
-                        throw (IOException) (ee.getCause());
-                    } else {
-                        throw new IOException("Error to execute entry log allocation.", ee);
-                    }
-                } catch (CancellationException ce) {
-                    throw new IOException("Task to allocate a new entry log is cancelled.", ce);
-                } catch (InterruptedException ie) {
-                    Thread.currentThread().interrupt();
-                    throw new IOException("Intrrupted when waiting a new entry log to be allocated.", ie);
-                } catch (TimeoutException e) {
-                    LOG.warn("Received TimeoutException while trying to get preallocation future result,"
-                            + " which means that Future is waiting for acquiring lock on EntryLoggerAllocator.this");
+        BufferedLogChannel createNewLog() throws IOException {
+            synchronized (createEntryLogLock) {
+                BufferedLogChannel bc;
+                if (!entryLogPreAllocationEnabled){
+                    // create a new log directly
                     bc = allocateNewLog();
-                }
-            }
-            if (entryLogPreAllocationEnabled) {
-                /*
-                 * We should submit new callable / create new instance of future only if the previous preallocation is
-                 * null or if it is done. This is needed because previous preallocation has not completed its execution
-                 * since it is waiting for lock on EntryLoggerAllocator.this.
-                 */
-                if ((preallocation == null) || preallocation.isDone()) {
+                    return bc;
+                } else {
+                    // allocate directly to response request
+                    if (null == preallocation){
+                        bc = allocateNewLog();
+                    } else {
+                        // has a preallocated entry log
+                        try {
+                            bc = preallocation.get();
+                        } catch (ExecutionException ee) {
+                            if (ee.getCause() instanceof IOException) {
+                                throw (IOException) (ee.getCause());
+                            } else {
+                                throw new IOException("Error to execute entry log allocation.", ee);
+                            }
+                        } catch (CancellationException ce) {
+                            throw new IOException("Task to allocate a new entry log is cancelled.", ce);
+                        } catch (InterruptedException ie) {
+                            Thread.currentThread().interrupt();
+                            throw new IOException("Intrrupted when waiting a new entry log to be allocated.", ie);
+                        }
+                    }
+                    // preallocate a new log in background upon every call
                     preallocation = allocatorExecutor.submit(() -> allocateNewLog());
+                    return bc;
                 }
             }
-            LOG.info("Created new entry logger {}.", bc.getLogId());
-            return bc;
         }
 
-        synchronized BufferedLogChannel createNewLogForCompaction() throws IOException {
+        BufferedLogChannel createNewLogForCompaction() throws IOException {
+            synchronized (createCompactionLogLock) {
                 return allocateNewLog(COMPACTING_SUFFIX);
+            }
         }
 
-        synchronized BufferedLogChannel allocateNewLog() throws IOException {
-            return allocateNewLog(LOG_FILE_SUFFIX);
+        private synchronized BufferedLogChannel allocateNewLog() throws IOException {
+            return allocateNewLog(".log");
         }
 
         /**
          * Allocate a new log file.
          */
-        synchronized BufferedLogChannel allocateNewLog(String suffix) throws IOException {
-            File dirForNextEntryLog;
+        private synchronized BufferedLogChannel allocateNewLog(String suffix) throws IOException {
             List<File> list = ledgerDirsManager.getWritableLedgerDirsForNewLog();
-            dirForNextEntryLog = entryLogManager.getDirForNextEntryLog(list);
+            File dirForNextEntryLog = entryLogManager.getDirForNextEntryLog(list);
 
             List<File> ledgersDirs = ledgerDirsManager.getAllLedgerDirs();
             String logFileName;
-            while (true) {
+            // It would better not to overwrite existing entry log files
+            File testLogFile = null;
+            do {
                 if (preallocatedLogId >= Integer.MAX_VALUE) {
                     preallocatedLogId = 0;
                 } else {
                     ++preallocatedLogId;
                 }
-                /*
-                 * make sure there is no entrylog which already has the same
-                 * logID. Have to check all the ledegerdirs. If already there is
-                 * an entrylog with that logid then move to next ID.
-                 */
-                String preallocatedLogIdHexString = Long.toHexString(preallocatedLogId);
-                logFileName = preallocatedLogIdHexString + suffix;
-                boolean entryLogAlreadyExistsWithThisId = false;
+                logFileName = Long.toHexString(preallocatedLogId) + suffix;
                 for (File dir : ledgersDirs) {
-                    File[] entrylogFilesWithPreallocatedLogId = dir.listFiles((file) -> {
-                        if (file.getName().startsWith(preallocatedLogIdHexString + ".")) {
-                            LOG.warn("Found existed entry log " + file + " when trying to create it as a new log.");
-                            return true;
-                        }
-                        return false;
-                    });
-
-                    if ((entrylogFilesWithPreallocatedLogId != null)
-                            && (entrylogFilesWithPreallocatedLogId.length > 0)) {
-                        entryLogAlreadyExistsWithThisId = true;
+                    testLogFile = new File(dir, logFileName);
+                    if (testLogFile.exists()) {
+                        LOG.warn("Found existed entry log " + testLogFile
+                               + " when trying to create it as a new log.");
+                        testLogFile = null;
                         break;
                     }
                 }
-                if (!entryLogAlreadyExistsWithThisId) {
-                    break;
-                }
-            }
+            } while (testLogFile == null);
 
             File newLogFile = new File(dirForNextEntryLog, logFileName);
             FileChannel channel = new RandomAccessFile(newLogFile, "rw").getChannel();
+
             BufferedLogChannel logChannel = new BufferedLogChannel(channel, conf.getWriteBufferBytes(),
                     conf.getReadBufferBytes(), preallocatedLogId, newLogFile, conf.getFlushIntervalInBytes());
             logfileHeader.readerIndex(0);
             logChannel.write(logfileHeader);
 
-            for (File f : ledgersDirs) {
+            for (File f : list) {
                 setLastLogId(f, preallocatedLogId);
             }
+
             if (suffix.equals(LOG_FILE_SUFFIX)) {
                 recentlyCreatedEntryLogsStatus.createdEntryLog(preallocatedLogId);
             }
+
             LOG.info("Created new entry log file {} for logId {}.", newLogFile, preallocatedLogId);
             return logChannel;
         }
@@ -927,32 +898,8 @@ public class EntryLogger {
          */
         @Override
         public long addEntry(long ledger, ByteBuf entry, boolean rollLog) throws IOException {
-
             int entrySize = entry.readableBytes() + 4; // Adding 4 bytes to prepend the size
-            BufferedLogChannel logChannel = getCurrentLogForLedger(ledger);
-            boolean reachEntryLogLimit = rollLog ? reachEntryLogLimit(logChannel, entrySize)
-                    : readEntryLogHardLimit(logChannel, entrySize);
-            // Create new log if logSizeLimit reached or current disk is full
-            boolean diskFull = (logChannel == null) ? false
-                    : ledgerDirsManager.isDirFull(logChannel.getLogFile().getParentFile());
-            boolean allDisksFull = !ledgerDirsManager.hasWritableLedgerDirs();
-
-            /**
-             * if disk of the logChannel is full or if the entrylog limit is
-             * reached or if the logchannel is not initialized, then
-             * createNewLog. If allDisks are full then proceed with the current
-             * logChannel, since Bookie must have turned to readonly mode and
-             * the addEntry traffic would be from GC and it is ok to proceed in
-             * this case.
-             */
-            if ((diskFull && (!allDisksFull)) || reachEntryLogLimit || (logChannel == null)) {
-                if (logChannel != null) {
-                    logChannel.flushAndForceWriteIfRegularFlush(false);
-                }
-                createNewLog(ledger);
-            }
-
-            logChannel = getCurrentLogForLedger(ledger);
+            BufferedLogChannel logChannel = getCurrentLogForLedgerForAddEntry(ledger, entrySize, rollLog);
             ByteBuf sizeBuffer = sizeBufferForAdd.get();
             sizeBuffer.clear();
             sizeBuffer.writeInt(entry.readableBytes());
@@ -980,6 +927,9 @@ public class EntryLogger {
         }
 
         abstract BufferedLogChannel getCurrentLogForLedger(long ledgerId);
+
+        abstract BufferedLogChannel getCurrentLogForLedgerForAddEntry(long ledgerId, int entrySize, boolean rollLog)
+                throws IOException;
 
         abstract void setCurrentLogForLedgerAndAddToRotate(long ledgerId, BufferedLogChannel logChannel);
 
@@ -1052,10 +1002,67 @@ public class EntryLogger {
 
         private volatile BufferedLogChannel activeLogChannel;
         private long logIdBeforeFlush = INVALID_LID;
+        private final AtomicBoolean shouldCreateNewEntryLog = new AtomicBoolean(false);
+
+        EntryLogManagerForSingleEntryLog(LedgerDirsManager ledgerDirsManager) {
+            // Register listener for disk full notifications.
+            ledgerDirsManager.addLedgerDirsListener(getLedgerDirsListener());
+        }
+
+        private LedgerDirsListener getLedgerDirsListener() {
+            return new LedgerDirsListener() {
+                @Override
+                public void diskFull(File disk) {
+                    // If the current entry log disk is full, then create new
+                    // entry log.
+                    BufferedLogChannel currentActiveLogChannel = activeLogChannel;
+                    if (currentActiveLogChannel != null
+                            && currentActiveLogChannel.getLogFile().getParentFile().equals(disk)) {
+                        shouldCreateNewEntryLog.set(true);
+                    }
+                }
+
+                @Override
+                public void diskAlmostFull(File disk) {
+                    // If the current entry log disk is almost full, then create new entry
+                    // log.
+                    BufferedLogChannel currentActiveLogChannel = activeLogChannel;
+                    if (currentActiveLogChannel != null
+                            && currentActiveLogChannel.getLogFile().getParentFile().equals(disk)) {
+                        shouldCreateNewEntryLog.set(true);
+                    }
+                }
+            };
+        }
 
         @Override
         public synchronized long addEntry(long ledger, ByteBuf entry, boolean rollLog) throws IOException {
             return super.addEntry(ledger, entry, rollLog);
+        }
+
+        @Override
+        synchronized BufferedLogChannel getCurrentLogForLedgerForAddEntry(long ledgerId, int entrySize,
+                boolean rollLog) throws IOException {
+            if (null == activeLogChannel) {
+                // log channel can be null because the file is deferred to be created
+                createNewLog(UNASSIGNED_LEDGERID);
+            }
+
+            boolean reachEntryLogLimit = rollLog ? reachEntryLogLimit(activeLogChannel, entrySize)
+                    : readEntryLogHardLimit(activeLogChannel, entrySize);
+            // Create new log if logSizeLimit reached or current disk is full
+            boolean createNewLog = shouldCreateNewEntryLog.get();
+            if (createNewLog || reachEntryLogLimit) {
+                if (activeLogChannel != null) {
+                    activeLogChannel.flushAndForceWriteIfRegularFlush(false);
+                }
+                createNewLog(UNASSIGNED_LEDGERID);
+                // Reset the flag
+                if (createNewLog) {
+                    shouldCreateNewEntryLog.set(false);
+                }
+            }
+            return activeLogChannel;
         }
 
         @Override
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
@@ -59,7 +59,6 @@ import java.util.Map;
 import java.util.Set;
 import java.util.SortedMap;
 import java.util.TreeMap;
-import java.util.concurrent.Callable;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
@@ -71,7 +70,6 @@ import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 
-import org.apache.bookkeeper.bookie.LedgerDirsManager.NoWritableLedgerDirException;
 import org.apache.bookkeeper.conf.ServerConfiguration;
 import org.apache.bookkeeper.util.IOUtils;
 import org.apache.bookkeeper.util.collections.ConcurrentLongLongHashMap;
@@ -88,7 +86,7 @@ import org.slf4j.LoggerFactory;
  */
 public class EntryLogger {
     private static final Logger LOG = LoggerFactory.getLogger(EntryLogger.class);
-    static final Long UNASSIGNED_LEDGERID = Long.valueOf(-1);
+    static final long UNASSIGNED_LEDGERID = -1L;
     // log file suffix
     private static final String LOG_FILE_SUFFIX = ".log";
 
@@ -99,7 +97,7 @@ public class EntryLogger {
         private final long logId;
         private final EntryLogMetadata entryLogMetadata;
         private final File logFile;
-        private Long ledgerIdAssigned = UNASSIGNED_LEDGERID;
+        private long ledgerIdAssigned = UNASSIGNED_LEDGERID;
 
         public BufferedLogChannel(FileChannel fc, int writeCapacity, int readCapacity, long logId, File logFile,
                 long unpersistedBytesBound) throws IOException {
@@ -221,7 +219,7 @@ public class EntryLogger {
     private final LedgerDirsManager ledgerDirsManager;
     private final boolean entryLogPerLedgerEnabled;
 
-    RecentEntryLogsStatus recentlyCreatedEntryLogsStatus;
+    final RecentEntryLogsStatus recentlyCreatedEntryLogsStatus;
 
     /**
      * locks for compaction log.
@@ -235,7 +233,8 @@ public class EntryLogger {
     private volatile BufferedLogChannel compactionLogChannel;
 
     final EntryLoggerAllocator entryLoggerAllocator;
-    final EntryLogManager entryLogManager;
+    private final EntryLogManager entryLogManager;
+
     private final boolean entryLogPreAllocationEnabled;
     private final CopyOnWriteArrayList<EntryLogListener> listeners = new CopyOnWriteArrayList<EntryLogListener>();
 
@@ -397,19 +396,20 @@ public class EntryLogger {
                      * all the logs should be flushed.
                      *
                      */
-                    flushCurrentLogs();
-                    super.checkpoint();
+                    super.flush();
                 }
 
                 @Override
-                public void rollLogs() throws IOException {
+                public void prepareSortedLedgerStorageCheckpoint(long numBytesFlushed) throws IOException {
+                    // do nothing
                     /*
-                     * rollLogs is required for singleentrylog scenario, but it
-                     * is not needed for entrylogperledger scenario, since
-                     * entries of a ledger go to a entrylog (even during
-                     * compaction) and SyncThread drives periodic checkpoint
-                     * logic.
+                     * prepareSortedLedgerStorageCheckpoint is required for
+                     * singleentrylog scenario, but it is not needed for
+                     * entrylogperledger scenario, since entries of a ledger go
+                     * to a entrylog (even during compaction) and SyncThread
+                     * drives periodic checkpoint logic.
                      */
+
                 }
 
                 @Override
@@ -445,6 +445,10 @@ public class EntryLogger {
         } else {
             this.entryLogManager = new EntryLogManagerForSingleEntryLog();
         }
+    }
+
+    EntryLogManager getEntryLogManager() {
+        return entryLogManager;
     }
 
     void addListener(EntryLogListener listener) {
@@ -558,8 +562,8 @@ public class EntryLogger {
         }
     }
 
-    void rollLogs() throws IOException {
-        entryLogManager.rollLogs();
+    void prepareSortedLedgerStorageCheckpoint(long numBytesFlushed) throws IOException {
+        entryLogManager.prepareSortedLedgerStorageCheckpoint(numBytesFlushed);
     }
 
     void prepareEntryMemTableFlush() {
@@ -632,7 +636,7 @@ public class EntryLogger {
                     Thread.currentThread().interrupt();
                     throw new IOException("Intrrupted when waiting a new entry log to be allocated.", ie);
                 } catch (TimeoutException e) {
-                    LOG.debug("Received TimeoutException while trying to get preallocation future result,"
+                    LOG.warn("Received TimeoutException while trying to get preallocation future result,"
                             + " which means that Future is waiting for acquiring lock on EntryLoggerAllocator.this");
                     bc = allocateNewLog();
                 }
@@ -644,12 +648,7 @@ public class EntryLogger {
                  * since it is waiting for lock on EntryLoggerAllocator.this.
                  */
                 if ((preallocation == null) || preallocation.isDone()) {
-                    preallocation = allocatorExecutor.submit(new Callable<BufferedLogChannel>() {
-                        @Override
-                        public BufferedLogChannel call() throws IOException {
-                            return allocateNewLog();
-                        }
-                    });
+                    preallocation = allocatorExecutor.submit(() -> allocateNewLog());
                 }
             }
             LOG.info("Created new entry logger {}.", bc.getLogId());
@@ -669,19 +668,7 @@ public class EntryLogger {
          */
         synchronized BufferedLogChannel allocateNewLog(String suffix) throws IOException {
             File dirForNextEntryLog;
-            List<File> list;
-
-            try {
-                list = ledgerDirsManager.getWritableLedgerDirs();
-            } catch (NoWritableLedgerDirException nwe) {
-                if (!ledgerDirsManager.hasWritableLedgerDirs()) {
-                    list = ledgerDirsManager.getWritableLedgerDirsForNewLog();
-                } else {
-                    LOG.error("All Disks are not full, but getWritableLedgerDirs threw exception ", nwe);
-                    throw nwe;
-                }
-            }
-
+            List<File> list = ledgerDirsManager.getWritableLedgerDirsForNewLog();
             dirForNextEntryLog = entryLogManager.getDirForNextEntryLog(list);
 
             List<File> ledgersDirs = ledgerDirsManager.getAllLedgerDirs();
@@ -858,7 +845,7 @@ public class EntryLogger {
          * add entry to the corresponding entrylog and return the position of
          * the entry in the entrylog
          */
-        long addEntry(Long ledger, ByteBuf entry, boolean rollLog) throws IOException;
+        long addEntry(long ledger, ByteBuf entry, boolean rollLog) throws IOException;
 
         /*
          * gets the active logChannel with the given entryLogId. null if it is
@@ -877,29 +864,24 @@ public class EntryLogger {
         void checkpoint() throws IOException;
 
         /*
-         * roll entryLogs.
+         * flush both current and rotated logs.
          */
-        void rollLogs() throws IOException;
-
-        /*
-         * flush rotated logs.
-         */
-        void flushRotatedLogs() throws IOException;
-
-        /*
-         * flush current logs.
-         */
-        void flushCurrentLogs() throws IOException;
+        void flush() throws IOException;
 
         /*
          * close current logs.
          */
-        void closeCurrentLogs() throws IOException;
+        void close() throws IOException;
 
         /*
          * force close current logs.
          */
-        void forceCloseCurrentLogs();
+        void forceClose();
+
+        /*
+         *
+         */
+        void prepareSortedLedgerStorageCheckpoint(long numBytesFlushed) throws IOException;
 
         /*
          * this method should be called before doing entrymemtable flush, it
@@ -926,18 +908,25 @@ public class EntryLogger {
     }
 
     abstract class EntryLogManagerBase implements EntryLogManager {
-        final Set<BufferedLogChannel> rotatedLogChannels;
+        final List<BufferedLogChannel> rotatedLogChannels;
 
         EntryLogManagerBase() {
-            rotatedLogChannels = ConcurrentHashMap.newKeySet();
+            rotatedLogChannels = new CopyOnWriteArrayList<BufferedLogChannel>();
         }
+
+        private final FastThreadLocal<ByteBuf> sizeBufferForAdd = new FastThreadLocal<ByteBuf>() {
+            @Override
+            protected ByteBuf initialValue() throws Exception {
+                return Unpooled.buffer(4);
+            }
+        };
 
         /*
          * This method should be guarded by a lock, so callers of this method
          * should be in the right scope of the lock.
          */
         @Override
-        public long addEntry(Long ledger, ByteBuf entry, boolean rollLog) throws IOException {
+        public long addEntry(long ledger, ByteBuf entry, boolean rollLog) throws IOException {
 
             int entrySize = entry.readableBytes() + 4; // Adding 4 bytes to prepend the size
             BufferedLogChannel logChannel = getCurrentLogForLedger(ledger);
@@ -950,19 +939,21 @@ public class EntryLogger {
 
             /**
              * if disk of the logChannel is full or if the entrylog limit is
-             * reached of if the logchannel is not initialized, then
+             * reached or if the logchannel is not initialized, then
              * createNewLog. If allDisks are full then proceed with the current
              * logChannel, since Bookie must have turned to readonly mode and
              * the addEntry traffic would be from GC and it is ok to proceed in
              * this case.
              */
             if ((diskFull && (!allDisksFull)) || reachEntryLogLimit || (logChannel == null)) {
-                flushCurrentLog(logChannel, false);
+                if (logChannel != null) {
+                    logChannel.flushAndForceWriteIfRegularFlush(false);
+                }
                 createNewLog(ledger);
             }
 
             logChannel = getCurrentLogForLedger(ledger);
-            ByteBuf sizeBuffer = EntryLogger.this.sizeBuffer.get();
+            ByteBuf sizeBuffer = sizeBufferForAdd.get();
             sizeBuffer.clear();
             sizeBuffer.writeInt(entry.readableBytes());
             logChannel.write(sizeBuffer);
@@ -988,18 +979,27 @@ public class EntryLogger {
             return logChannel.position() + size > Integer.MAX_VALUE;
         }
 
-        abstract BufferedLogChannel getCurrentLogForLedger(Long ledgerId);
+        abstract BufferedLogChannel getCurrentLogForLedger(long ledgerId);
 
-        abstract void setCurrentLogForLedgerAndAddToRotate(Long ledgerId, BufferedLogChannel logChannel);
+        abstract void setCurrentLogForLedgerAndAddToRotate(long ledgerId, BufferedLogChannel logChannel);
 
-        public Set<BufferedLogChannel> getCopyOfRotatedLogChannels() {
-            return new HashSet<BufferedLogChannel>(rotatedLogChannels);
+        /*
+         * flush current logs.
+         */
+        abstract void flushCurrentLogs() throws IOException;
+
+        List<BufferedLogChannel> getRotatedLogChannels() {
+            return rotatedLogChannels;
         }
 
         @Override
+        public void flush() throws IOException {
+            flushCurrentLogs();
+            flushRotatedLogs();
+        }
+
         public void flushRotatedLogs() throws IOException {
-            Set<BufferedLogChannel> channels = getCopyOfRotatedLogChannels();
-            for (BufferedLogChannel channel : channels) {
+            for (BufferedLogChannel channel : rotatedLogChannels) {
                 channel.flushAndForceWrite(true);
                 // since this channel is only used for writing, after flushing the channel,
                 // we had to close the underlying file channel. Otherwise, we might end up
@@ -1011,7 +1011,7 @@ public class EntryLogger {
             }
         }
 
-        void flushCurrentLog(BufferedLogChannel logChannel, boolean forceMetadata) throws IOException {
+        void flushLogChannel(BufferedLogChannel logChannel, boolean forceMetadata) throws IOException {
             if (logChannel != null) {
                 logChannel.flushAndForceWrite(forceMetadata);
                 LOG.debug("Flush and sync current entry logger {}", logChannel.getLogId());
@@ -1022,7 +1022,7 @@ public class EntryLogger {
          * Creates a new log file. This method should be guarded by a lock,
          * so callers of this method should be in right scope of the lock.
          */
-        void createNewLog(Long ledgerId) throws IOException {
+        void createNewLog(long ledgerId) throws IOException {
             BufferedLogChannel logChannel = getCurrentLogForLedger(ledgerId);
             // first tried to create a new log channel. add current log channel to ToFlush list only when
             // there is a new log channel. it would prevent that a log channel is referenced by both
@@ -1038,7 +1038,7 @@ public class EntryLogger {
                 BufferedLogChannel newLogChannel = entryLoggerAllocator.createNewLog();
                 setCurrentLogForLedgerAndAddToRotate(ledgerId, newLogChannel);
                 LOG.info("Flushing entry logger {} back to filesystem, pending for syncing entry loggers : {}.",
-                        logChannel.getLogId(), getCopyOfRotatedLogChannels());
+                        logChannel.getLogId(), rotatedLogChannels);
                 for (EntryLogListener listener : listeners) {
                     listener.onRotateEntryLog();
                 }
@@ -1054,17 +1054,17 @@ public class EntryLogger {
         private long logIdBeforeFlush = INVALID_LID;
 
         @Override
-        public synchronized long addEntry(Long ledger, ByteBuf entry, boolean rollLog) throws IOException {
+        public synchronized long addEntry(long ledger, ByteBuf entry, boolean rollLog) throws IOException {
             return super.addEntry(ledger, entry, rollLog);
         }
 
         @Override
-        synchronized void createNewLog(Long ledgerId) throws IOException {
+        synchronized void createNewLog(long ledgerId) throws IOException {
             super.createNewLog(ledgerId);
         }
 
         @Override
-        public synchronized void setCurrentLogForLedgerAndAddToRotate(Long ledgerId, BufferedLogChannel logChannel) {
+        public synchronized void setCurrentLogForLedgerAndAddToRotate(long ledgerId, BufferedLogChannel logChannel) {
             BufferedLogChannel hasToRotateLogChannel = activeLogChannel;
             activeLogChannel = logChannel;
             if (hasToRotateLogChannel != null) {
@@ -1073,7 +1073,7 @@ public class EntryLogger {
         }
 
         @Override
-        public BufferedLogChannel getCurrentLogForLedger(Long ledgerId) {
+        public BufferedLogChannel getCurrentLogForLedger(long ledgerId) {
             return activeLogChannel;
         }
 
@@ -1097,11 +1097,6 @@ public class EntryLogger {
             flushRotatedLogs();
         }
 
-        @Override
-        public synchronized void rollLogs() throws IOException {
-            createNewLog(UNASSIGNED_LEDGERID);
-        }
-
         public long getCurrentLogId() {
             BufferedLogChannel currentActiveLogChannel = activeLogChannel;
             if (currentActiveLogChannel != null) {
@@ -1119,19 +1114,19 @@ public class EntryLogger {
                  * flushCurrentLogs method is called during checkpoint, so
                  * metadata of the file also should be force written.
                  */
-                flushCurrentLog(currentActiveLogChannel, true);
+                flushLogChannel(currentActiveLogChannel, true);
             }
         }
 
         @Override
-        public void closeCurrentLogs() throws IOException {
+        public void close() throws IOException {
             if (activeLogChannel != null) {
                 closeFileChannel(activeLogChannel);
             }
         }
 
         @Override
-        public void forceCloseCurrentLogs() {
+        public void forceClose() {
             if (activeLogChannel != null) {
                 forceCloseFileChannel(activeLogChannel);
             }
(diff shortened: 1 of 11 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -1441,4 +1658,40 @@ static long fileName2LogId(String fileName) {
...
+     * Datastructure which maintains the status of logchannels. When a
+     * logChannel is created entry of < entryLogId, false > will be made to this
+     * sortedmap and when logChannel is rotated and flushed then the entry is
+     * updated to < entryLogId, true > and all the lowest entries with
+     * < entryLogId, true > status will be removed from the sortedmap. So that way
+     * we could get least unflushed LogId.
+     *
+     */
+    static class RecentEntryLogsStatus {
+        private SortedMap<Long, Boolean> entryLogsStatusMap;
```
sijie: "private final"

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -802,88 +864,207 @@ private long readLastLogId(File f) {
         }
     }
 
+    interface EntryLogManager {
+        /*
+         * acquire lock for this ledger.
+         */
+        void acquireLock(Long ledgerId);
```
sijie: Can we try to hide these locking stuffs?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -802,88 +864,207 @@ private long readLastLogId(File f) {
...
+         * release lock for this ledger
+         */
+        void releaseLock(Long ledgerId);
+
+        /*
+         * sets the logChannel for the given ledgerId. The previous one will be
+         * removed from replicaOfCurrentLogChannels. Previous logChannel will be
+         * added to rotatedLogChannels.
+         */
+        void setCurrentLogForLedger(Long ledgerId, BufferedLogChannel logChannel);
```
sijie: All the `currentLogForLedger` `copy of rotated log channels` `copy of current logs` are kind of implementation detail of an implementation of EntryLogManager. Can we think of a better abstraction to hide these?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -116,6 +121,18 @@ public ConcurrentLongLongHashMap getLedgersMap() {
             return entryLogMetadata.getLedgersMap();
         }
 
+        public boolean isLedgerDirFull() {
```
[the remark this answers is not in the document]
reddycharan: ok will change that.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
reddycharan: currently in all the places 'synchronized' is used, but for entrylogperledger we should have a lock for each ledger/entrylog. Hence these lock methods in EntryLogManager. The contract with this interface is only in addEntry call "acquireLockByCreatingIfRequired" (because this call would create a lock if it is not created yet and we need this in addEntry method since addEntry will be the first call related to that ledger in write path) and for all other methods in EntryLogger when they are dealing/writing to that particular ledger/entrylog they would call "acquireLock". I don't see why it needs to be hidden. EntryLogger needs to synchronize, instead of synchronizing on its intrinsic lock it delegates it to EntryLogManager, and EntryLogManager deals with inner details but all it needs is call to acquireLock and releaseLock with ledgerId.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
reddycharan: will do

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
reddycharan: With my changes in this sub-task and subsequent sub-tasks I would like to abstract out the code/logic pertaining to singleentrylog/entrylogperledger and move it to the implementations of EntryLogManager, but leave the rest of the code as it is. With that reasoning, these methods in EntryLogManager and variables need to be exposed and EntryLogger class has to work with these methods.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -292,12 +308,28 @@ public EntryLogger(ServerConfiguration conf,
...
+                @Override
+                public void checkpoint() throws IOException {
+                    /*
+                     * In the case of entryLogPerLedgerEnabled we need to flush
+                     * both rotatedlogs and currentlogs. This is needed because
+                     * syncThread periodically does checkpoint and at this time
+                     * all the logs should be flushed.
+                     *
+                     */
+                    if (entryLogPerLedgerEnabled) {
```
reddycharan: nit: it should be simply, (will fix it in next iteration)

public void checkpoint() throws IOException {
    flushCurrentLogs();
    super.checkpoint();
}

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SortedLedgerStorage.java
```diff
@@ -167,14 +167,7 @@ public ByteBuf getEntry(long ledgerId, long entryId) throws IOException {
...
     public void checkpoint(final Checkpoint checkpoint) throws IOException {
-        long numBytesFlushed = memTable.flush(this, checkpoint);
-        if (numBytesFlushed > 0) {
-            // if bytes are added between previous flush and this checkpoint,
-            // it means bytes might live at current active entry log, we need
-            // roll current entry log and then issue checkpoint to underlying
-            // interleaved ledger storage.
-            entryLogger.rollLog();
-        }
+        memTable.flush(this, checkpoint);
```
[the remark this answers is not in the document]
reddycharan: First of all I’m not sure, why this is added with this commit https://github.com/apache/bookkeeper/commit/81cbba3cf620f2a6df48c0da6ee1ac019de24fbc. SortedLedgerStorage.checkpoint should be called from onSizeLimitReached’s scheduler. In onSizeLimitReached’s scheduler, memtable is flushed completely and if entrylog has reached its capacity then it will roll log. So here by the time startCheckpoint for the checkpoint cp is called all the entries must be flushed from memtable to entry log and the entry log must be rolled. So in this checkpoint method, the entries which come before checkpoint cp are already added to entry log and rolled over. So I don’t see the need of this extra rollLog call.

Mainly I removed rollLog method, instead I introduced rollLogsIfEntryLogLimitReached, which is not appropriate here.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SortedLedgerStorage.java
```diff
@@ -209,16 +202,12 @@ public void onSizeLimitReached(final Checkpoint cp) throws IOException {
...
                     memTable.flush(SortedLedgerStorage.this);
-                    long logIdAfterFlush = entryLogger.getCurrentLogId();
                     // in any case that an entry log reaches the limit, we roll the log and start checkpointing.
                     // if a memory table is flushed spanning over two entry log files, we also roll log. this is
                     // for performance consideration: since we don't wanna checkpoint a new log file that ledger
                     // storage is writing to.
-                    if (entryLogger.reachEntryLogLimit(0) || logIdAfterFlush != logIdBeforeFlush) {
-                        LOG.info("Rolling entry logger since it reached size limitation");
-                        entryLogger.rollLog();
+                    if (entryLogger.rollLogsIfEntryLogLimitReached()) {
```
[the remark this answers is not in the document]
reddycharan: i don't think it is significant to have any perf impact. First of all this will be called once EntryMemTable reaches size limit, and in this method I'm getting a collection of active entrylogs (in this case just one) and checks its size. For entrylogperledger this collection is required but for single entrylog it might not be required, but I want common implementation and it is called just once per EntryMemTable size limit, so it is not a thing to be concerned about.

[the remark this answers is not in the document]
sijie: @reddycharan well, the existence of the code had its reason. performance is one, I think the biggest problem is "checkpoint" logic can potentially never be triggered.

- during memtable flushes, rollLog is disabled for appending memtable entries.
- however background compaction can also advance entrylog, since it is appending entries into same entrylog.
- if a memtable flush spanning over two entry log files, `rollLogsIfEntryLogLimitReached` will most likely be false, so no checkpoint will happen.
- in the next memtable flush, if it is spanning over two entry log files again, then `rollLogsIfEntryLogLimitReached` still be false, no checkpoint will happen.

This was actually a behavior we observed at Twitter 3 years ago. I added this logic to prevent this happen. see twitter/bookkeeper@a7a8c88eaa2659b5d6c4e29f4eb9e2d184c7cfe0

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieInitializationTest.java
```diff
@@ -584,15 +584,6 @@ public void testWithDiskFullReadOnlyDisabledOrForceGCAllowDisabled() throws Exce
         } catch (NoWritableLedgerDirException e) {
             // expected
         }
-
```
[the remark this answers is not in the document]
reddycharan: in the existing implementation, when Bookie is initialized, it will eventually call EntryLogger.initialize() -> EntryLogger.createNewLog() and it would fail with exception because of https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDirsManager.java#L184 . But now with this change we are not creating newlog during initialization of Bookie ( and EntryLogger), EntryLog would be created only after receiving first EntryLogger.addEntry call, hence Bookie initialization would succeed in this case though IsForceGCAllowWhenNoSpace is set to false.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SortedLedgerStorage.java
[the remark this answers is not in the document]
reddycharan: brought back rollLogs

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -802,88 +876,217 @@ private long readLastLogId(File f) {
         }
     }
 
+    interface EntryLogManager {
```
ivankelly: Move into its own java file. EntryLogger is already huge.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -802,88 +876,217 @@ private long readLastLogId(File f) {
...
+         */
+        void checkpoint() throws IOException;
+
+        /*
+         * roll entryLogs.
+         */
+        void rollLogs() throws IOException;
+    }
+
+    class EntryLogManagerForSingleEntryLog implements EntryLogManager {
```
ivankelly: Move into it's own file.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -292,12 +304,37 @@ public EntryLogger(ServerConfiguration conf,
                 logId = lastLogId;
             }
         }
-        this.leastUnflushedLogId = logId + 1;
+        this.recentlyCreatedEntryLogsStatus = new RecentEntryLogsStatus(logId + 1);
```
ivankelly: Should be part of the entryLogManager.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -264,6 +274,8 @@ public EntryLogger(ServerConfiguration conf,
         // but the protocol varies so an exact value is difficult to determine
         this.maxSaneEntrySize = conf.getNettyMaxFrameSizeBytes() - 500;
         this.ledgerDirsManager = ledgerDirsManager;
+        this.conf = conf;
+        entryLogPerLedgerEnabled = conf.isEntryLogPerLedgerEnabled();
```
ivankelly: The entryLogManager should be constructed outside of the EntryLogger and passed in as an argument. The entryLogger shouldn't need to know which implementation of the entryLogManager it is using, which is the whole point of having an interface. Also injecting the implementation like this, will make for easier testing, and SortedLedgerStorage would not need to be touched at all (i.e. SortedLedgerStorage can be changed to always create the SingleLog variant).

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SortedLedgerStorage.java
```diff
@@ -78,6 +79,7 @@ public void initialize(ServerConfiguration conf,
                 .setNameFormat("SortedLedgerStorage-%d")
                 .setPriority((Thread.NORM_PRIORITY + Thread.MAX_PRIORITY) / 2).build());
         this.stateManager = stateManager;
+        this.isTransactionalCompactionEnabled = conf.getUseTransactionalCompaction();
```
ivankelly: Separate change, so should not be in this patch.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -513,78 +519,86 @@ void createNewLog() throws IOException {
...
     /**
      * Append the ledger map at the end of the entry log.
      * Updates the entry log file header with the offset and size of the map.
      */
-    private void appendLedgersMap(BufferedLogChannel entryLogChannel) throws IOException {
-        long ledgerMapOffset = entryLogChannel.position();
-
-        ConcurrentLongLongHashMap ledgersMap = entryLogChannel.getLedgersMap();
-        int numberOfLedgers = (int) ledgersMap.size();
+    private void appendLedgersMap(Long ledgerId) throws IOException {
```
reddycharan: will move this appendLedgersMap method to BufferedLogChannel. Because of that explicit need of converting from ledgerid to logchannel and synchronization wont be required.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
reddycharan: will do

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
reddycharan: will do

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
reddycharan: i kept the logic related to newlog creation and flush logs with in EntryLogger. So it should be here.

sijie:
@ivankelly : 

> Move into its own java file. EntryLogger is already huge.
> Move into it's own file.


I see you left a few comments around moving entry log manager out side of the entry logger class. It is a bit difficult in this patch, because entry log manager is sharing common stuffs (such as buffered log channels) between implementations. 

I would suggest not doing this large refactor (by moving stuffs out of entry logger). Let this PR focus on the interface abstraction and make EntryLogger responsible for managing EntryLogManager. Once we agree on the interface, we can do the cleanup by moving the interface out and have some common abstract implementation in subsequent changes.

Let's focus on abstracting interface at this PR, not to complicate stuffs.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
reddycharan: hmm..I considered moving out EntryLogManager, EntryLogManagerForSingleEntryLog and EntryLogManagerForEntryLogPerLedger out of EntryLogger, but  I realized that there is circular dependency. EntryLogger has a EntryLogManager reference, and for certain functionalities of EntryLogManager, it need to have reference of EntryLogger to call its methods in EntryLogManager. So I don’t see much benefits other than just creating new files and new classes (which I'm ok with).

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SortedLedgerStorage.java
reddycharan: this is part of this change itself. Provided more comments for why it is needed - https://github.com/apache/bookkeeper/pull/1281/files#diff-c5b1fb6226c382c12ed9f2dd9e6c8ed2R226

ivankelly: I don't see what the point of this flag even is. The external behaviour of SortedLedgerStorage should not change with this patch, so why the flag?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SortedLedgerStorage.java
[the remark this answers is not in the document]
ivankelly: @sijie if ```rollLogsIfEntryLogLimitReached``` is false, it will fall through to the else if statement, which covers the flush over two logs.

This is weird though, because it is really tied to a single log, but it acts like it's working with many.

sijie:
@ivankelly : 

appreciate you are working on this change. Here are my thought below:

> If there are circular dependencies, then the abstraction is broken from the very start. I've commented on how to fix it, and pushed a sketched patch to master...ivankelly:entrylogmanager which deals with a lot of the issues.

if `EntryLogManager` is an inner class (as what Charan) proposed, the abstraction is not *broken*. It is just need refactor later. I am saying "circular dependencies" is for what you said moving `EntryLogManager` out of the EntryLogger class. You need to move a lot of common code around to break the circular depenency, which I think that is too much for this PR or not worth at this moment. Because that is taking this PR into a whole "refactor" story, which I don't think we should take, because  1) we are losing the focus 2) it moves code around, make things a bit hard to review 3) the movement can be tricky, you don't really know if that works for multiple entry log.

The approach what I am suggesting is focusing on the interface - what is the right methods that EntryLogManager need for both single entry log and multiple entry logs. Get the interface done, Get the multiple entry log implementation, and do the refactor to move entry log manager out.  That is the most practical approach than the reverse. Otherwise we might end up moving code around and eventually found that this movement doesn't work for multiple entry log approach.

There is no real sense talking about inner class vs outer class before we get an interface abstraction that can work for both single entry log and multiple entry logs. That is my thought. but if you really feel strong about it, go for it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SortedLedgerStorage.java
sijie: @ivankelly 

> The external behaviour of SortedLedgerStorage should not change with this patch

it is not an external behavior of SortedLedgerStorage. It is the logic of SortedLedgerStorage. SortedLedgerStorage needs to realize the compaction mechanism to do checkpointing. That's related to the discussion below around checkpointing.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SortedLedgerStorage.java
```diff
@@ -209,16 +211,39 @@ public void onSizeLimitReached(final Checkpoint cp) throws IOException {
             public void run() {
                 try {
                     LOG.info("Started flushing mem table.");
-                    long logIdBeforeFlush = entryLogger.getCurrentLogId();
+                    long prevAllocLogIdBeforeFlush = entryLogger.getPreviousAllocatedEntryLogId();
```
sijie: This change is not correct. Because it should be "current" log id, not "previous allocated entry log id". That says the behavior is different between preallocation enabled and disabled.

That leads me to think that current entry log manager interface is not abstracted with right methods. because it abstracts a lot of methods around the underlying files, which underlying files are kind of tight with corresponding checkpointing and flushing logic. two different entry log manager can't really fit in this interface or checkpointing logic as well. 

So as what I said in my previous comment, we need to think about what is the right methods to put in EntryLogManager.

> I would suggest the EntryLogManager should hide the details on how to create log, allocate log,
> flush/checkpoint logs. the details such as preallocationLogId, unflushedLogId and how to lock them should be > hidden to the implementation, rather than defining a too finer interface here.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SortedLedgerStorage.java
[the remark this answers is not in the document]
sijie: @ivankelly that's what I commented before. I don't think the methods in EntryLogManager provide the right abstraction. That's what I also dislike your approach on asking moving the class out at this PR. At this PR, we need to figure out what are the right methods to put in an EntryLogManager before any kind of code movement. You comments about inner classes complicate things.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SortedLedgerStorage.java
reddycharan: "This change is not correct. Because it should be "current" log id, not "previous allocated entry log id". That says the behavior is different between preallocation enabled and disabled."

@sijie why would you say that behavior would be different between preallocation enabled and disabled? Is it because of async nature of preallocation?

Ok, I think I can do the following, which would solve 2 problems here.

I can add "getCurrentLogId" to the EntryLogManager Interface. For EntryLogManagerForSingleEntryLog implementation it would give the logId of the current activelog (which is current behavior). For EntryLogManagerForEntryLogPerLedger it would give some constant value all the times and it should be fine in entryLogPerLedger case, because checkpoint is not made when entrylog is rotated but instead SyncThread drives it periodically. 

Also it would remove the need of extra condition regarding "isTransactionalCompactionEnabled" which I added in this change.

So it would be clean and straightforward.

sijie: @reddycharan 

>  why would you say that behavior would be different between preallocation enabled and disabled? Is it because of async nature of preallocation?

yeah because preallocation is in background, preallocation can advance entry log id. if it does, the logic is not correct.

> I can add "getCurrentLogId" to the EntryLogManager Interface. For EntryLogManagerForSingleEntryLog implementation it would give the logId of the current activelog (which is current behavior).

That could work for now. Although I would still like to see if there is possibility to produce a cleaner interface without exposing those log ids. Let me think a bit more and come back.

reddycharan:
Thanks @ivankelly for taking the effort to consider ways to refactor the code. But this class is in deep down in the stack and I'm hesitant to make big code churn. With my changes in this sub-task and subsequent sub-tasks I would like to abstract out the code/logic pertaining to singleentrylog/entrylogperledger and move it to the implementations of EntryLogManager, but leave the rest of the code as it is to have minimal code churn and more focus. Once things are in place, I can consider the refactoring aspect of these classes.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SortedLedgerStorage.java
ivankelly: If you have to supply a flag, then there's something external to be observed.

Let me rephrase. As I understand it, we do not want the behaviour of SortedLedgerStorage flushing to change because of the change in this PR. So why are we adding a new configuration flag?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SortedLedgerStorage.java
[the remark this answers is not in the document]
ivankelly: > @ivankelly that's what I commented before. I don't think the methods in EntryLogManager provide the right abstraction. 

I agree about the methods providing the right abstraction. I was under the impression that #getPreviousAllocatedEntryLogId() would always return an incremented value if the current log has changed, but having read the code, this is not the case.

It seems to me that SortedLedgerStorage is reaching very far into the internals of the entrylogger to do what it wants to do. Since this is the case, we shouldn't try to shoehorn this into the abstraction. SortedLedgerStorage assumes that it uses a single log, so it should only ever be constructed with a SingleLogEntryLogmanager, and SingleLogEntryLogManager should expose the stuff that SortedLedgerStorage has required until now.

sijie:
@ivankelly 

> I'm of the opposite view. If the responsibility of each class is not clear, then you cannot possibly have a clean interface abstraction.

I hate to say this, but still I have to say it. because I don't want to get into these kind of arguments about "engineering-preferences" style arguments (e.g. circular dependency, two booleans) again and again. 

I still remembered a similar "circular dependency" argument that lasted for weeks when I introduced a better checkpoint mechanism 3-4 years ago. You disliked whatever you called "circular dependency" in the patch and let me change it to a way you preferred. but the approach turned out to have a correctness issue, and I had to revert it back (#659) recently to my original change that I did for Twitter (which has been running perfectly at twitter production for years). We could have been avoiding this kind of correctness issue if the review was more focused on logic and correctness rather than just thinking of breaking "circular dependency".

I am raising this old case here is not to point fingers. I want us to learn a lesson there, and avoid such case happening again. Because I see these review comments (around moving classes, breaking circular dependencies) generally are not really helpful. first, they make reviews losing focus; second, they have nothing to help with engineering progress; third, it exhausts energies for everyone involving in the PR - the author has the change his mind to one reviewer's mind, the other reviewers have to change their also, and etc. 

so I would be much happier to see if the reviews made are more focused on finding the correctness/logic issues with how the methods defined in this interface, rather than getting into yet-another-styleish-argument.

ivankelly:
@sijie 

> I hate to say this, but still I have to say it. because I don't want to get into these kind of arguments about "engineering-preferences" style arguments (e.g. circular dependency, two booleans) again and again.

These are not '"engineering-preferences" style arguments'. boolean parameters and circular dependencies are well established as bad practices.

> I still remembered a similar "circular dependency" argument that lasted for weeks when I introduced a better checkpoint mechanism 3-4 years ago. You disliked whatever you called "circular dependency" in the patch and let me change it to a way you preferred. but the approach turned out to have a correctness issue, and I had to revert it back (#659) recently to my original change that I did for Twitter (which has been running perfectly at twitter production for years). We could have been avoiding this kind of correctness issue if the review was more focused on logic and correctness rather than just thinking of breaking "circular dependency".

Since you brought it up, my objection to BOOKKEEPER-564 was that journal was calling to ledger storage and then ledger storage was then calling to journal. This is _exactly_ where the issue in #659 came from. My initial comments on [BOOKKEPER-564](https://issues.apache.org/jira/browse/BOOKKEEPER-564) were that [the mark should travel with the entries](https://issues.apache.org/jira/browse/BOOKKEEPER-564?focusedCommentId=13623992&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13623992). Anyhow, not to go into too much detail, but if you are trying to use the argument on that JIRA as an example of where arguing for certain engineering standard lead to a bug, you're pretty wide of the mark. SortedLedgerStorage wasn't even in the branch at that point. And what I was arguing against was exactly what caused the bug.

>  Because I see these review comments (around moving classes, breaking circular dependencies) generally are not really helpful. 

If I'm going to review code, I'm going to review it to the best of my ability to ensure that the code is maintainable in the future. If my reviews are unwanted, I can spend my time on other projects.

sijie:
> @sijie if you are ok with my comment regarding currentLogId, I can send next iteration change.

go for it now. I will think of how to clean it up later.

ivankelly:
@sijie I going to step back from this review and BK reviews in general because too many are devolving into arguments like this.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -83,18 +88,18 @@
  */
 public class EntryLogger {
     private static final Logger LOG = LoggerFactory.getLogger(EntryLogger.class);
+    private static final Long INVALID_LEDGERID = Long.valueOf(-1);
```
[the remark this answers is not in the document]
reddycharan: changed it to UNASSIGNED_LEDGERID

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -83,18 +88,18 @@
...
+    // log file suffix
+    private static final String LOG_FILE_SUFFIX = ".log";
 
     static class BufferedLogChannel extends BufferedChannel {
         private final long logId;
         private final EntryLogMetadata entryLogMetadata;
         private final File logFile;
+        private Long ledgerId = INVALID_LEDGERID;
```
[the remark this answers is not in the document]
reddycharan: changed it to 'ledgerIdAssigned'

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/CreateNewLogTest.java
```diff
@@ -100,9 +104,9 @@ public void testCreateNewLog() throws Exception {
         EntryLogger el = new EntryLogger(conf, ledgerDirsManager);
         // Calls createNewLog, and with the number of directories we
         // are using, if it picks one at random it will fail.
-        el.createNewLog();
-        LOG.info("This is the current log id: " + el.getCurrentLogId());
-        assertTrue("Wrong log id", el.getCurrentLogId() > 1);
+        el.createNewLog(0L);
+        LOG.info("This is the current log id: " + el.getPreviousAllocatedEntryLogId());
```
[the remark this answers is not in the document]
reddycharan: done

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -393,12 +513,37 @@ public BufferedReadChannel getFromChannels(long logId) {
...
+        return recentlyCreatedEntryLogsStatus.getLeastUnflushedLogId();
     }
 
-    synchronized long getCurrentLogId() {
-        return logChannel.getLogId();
+    long getPreviousAllocatedEntryLogId() {
+        return entryLoggerAllocator.getPreallocatedLogId();
+    }
+
+    boolean rollLogsIfEntryLogLimitReached() throws IOException {
```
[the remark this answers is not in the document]
reddycharan: changed EntryLogManager API as we discussed.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/ReadOnlyEntryLogger.java
```diff
@@ -54,7 +49,7 @@ protected boolean removeEntryLog(long entryLogId) {
     }
 
     @Override
-    public synchronized long addEntry(long ledger, ByteBuffer entry) throws IOException {
+    public synchronized long addEntry(Long ledgerId, ByteBuffer entry) throws IOException {
```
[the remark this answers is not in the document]
reddycharan: in EntryLogManagerForEntryLogPerLedger I maintain couple of datastructures, where key for HashMap is ledgerId and hence Long wrapper class is needed.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/CreateNewLogTest.java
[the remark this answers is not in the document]
reddycharan: in all these places i considered whether preallocation is enabled or not appropriately and  getPrevisousAllocatedEntryLogId is the appropriate one to validate the cases here.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -802,88 +881,230 @@ private long readLastLogId(File f) {
...
+            recentlyCreatedEntryLogsStatus.flushRotatedEntryLog(channel.getLogId());
+            entryLogManager.removeFromRotatedLogChannels(channel);
             LOG.info("Synced entry logger {} to disk.", channel.getLogId());
         }
-        // move the leastUnflushedLogId ptr
-        leastUnflushedLogId = flushedLogId + 1;
     }
 
     public void flush() throws IOException {
+        flushCurrentLogs();
```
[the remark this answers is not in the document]
reddycharan: i think it is incorrect to call flushrotatedlogs first and then flushcurrentlogs. Consider this scenario - if in between flushRotatedLogs, flushCurrentLogs calls current activelog is rotated and added to rotatedlogs list then we would miss flushing that log in this current flush call. So it has to be first flushCurrentLogs and then flushRotatedLogs

reddycharan:
Hey @sijie I think I addressed all of your comments and tests are looking good.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -788,89 +852,340 @@ private long readLastLogId(File f) {
...
+    void checkpoint() throws IOException {
+        entryLogManager.checkpoint();
+    }
+
+
+
     public void flush() throws IOException {
-        flushRotatedLogs();
-        flushCurrentLog();
+        entryLogManager.flushCurrentLogs();
```
sijie: for single-log manager, it is better to flush `rotated logs` before flushing `current logs`, because the way how it rotates files and do the checkpoint. I would prefer keeping existing logic unchanged to reduce the risks.

besides that, why not provide `flush` method in EntryLogManager? so different entry log manager implementation can implement its own flushing logic rather than exposing this knowledge to EntryLogger. If you provide `flush` method, it is possible that you don't need to expose `flushCurrentLogs` and `flushRotatedLogs`.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -788,89 +852,340 @@ private long readLastLogId(File f) {
...
+
+        /*
+         * Do the operations required for checkpoint.
+         */
+        void checkpoint() throws IOException;
+
+        /*
+         * roll entryLogs.
+         */
+        void rollLogs() throws IOException;
```
sijie: after the change of `prepareMemtableFlush` and `commitMemtableFlush`, rollLogs seems to be only the behavior of single-log manager. I don't think you need `rollLogs` here, unless I missed something.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -788,89 +852,340 @@ private long readLastLogId(File f) {
...
+
+        /*
+         * flush rotated logs.
+         */
+        void flushRotatedLogs() throws IOException;
```
sijie: `flushRotatedLogs` and `flushCurrentLogs` are only used in `flush()` method, why not just provide `flush()` method in EntryLogManager, like `checkpoint`.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -788,89 +852,340 @@ private long readLastLogId(File f) {
...
+
+        /*
+         * flush current logs.
+         */
+        void flushCurrentLogs() throws IOException;
+
+        /*
+         * close current logs.
+         */
+        void closeCurrentLogs() throws IOException;
```
sijie: the `closeCurrentLogs` and `forceCloseCurrentLogs` are the abstraction of the logic during `closing`. I think it is better to call it `close` and `closeNow` (or `forceClose`) - meaning closing the manager, force closing the manager. This would provide a better meaning and it is similar as `shutdown` and `shutdownNow` in scheduler.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/ReadOnlyEntryLogger.java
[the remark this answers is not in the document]
sijie: @reddycharan you can still use long for retreiving from a Long hashmap. why do you need to change here? in any cases long to Long is a boxing operation, if single-log manager doesn't need Long, it should be deferred to where it is really needed.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SortedLedgerStorage.java
sijie: I don't think we need this flag any more, no?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SortedLedgerStorage.java
```diff
@@ -173,7 +175,7 @@ public void checkpoint(final Checkpoint checkpoint) throws IOException {
             // it means bytes might live at current active entry log, we need
             // roll current entry log and then issue checkpoint to underlying
             // interleaved ledger storage.
-            entryLogger.rollLog();
+            entryLogger.rollLogs();
```
sijie: I think `numBytesFlushed > 0` logic is a behavior that specific to single-log manager. `rollLogs` is a not proper term for the meaning here. I would suggest adding a method in the EntryLogManager:

`void prepareCheckpoint(Checkpoint checkpoint, long numBytesFlushedBetweenCheckpoints)` -> it means prepare checkpoint in the entry logger. 

Then logic here can be changed to:

```
long numBytesFlushed = memTable.flush(this, checkpoint);
entryLogger.prepareCheckpoint(checkpoint, numBytesFlushed);
super.checkpoint(checkpoint);
```

in single-log manager, you can implement 

```
void prepareCheckpoint(Checkpoint checkpoint, long numBytesFlushedBetweenCheckpoint) {
    if (numBytesFlushedBetweenCheckpoint > 0) {
        rollLogs();
    }
}
```

in per-ledger-log manager, it is a no-op.

so in this way, you can get rid of `rollLogs` in EntryLogManager.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/CreateNewLogTest.java
```diff
@@ -93,16 +98,16 @@ public void testCreateNewLog() throws Exception {
...
         File newLogFile = new File(dir, logFileName);
         newLogFile.createNewFile();
 
         EntryLogger el = new EntryLogger(conf, ledgerDirsManager);
         // Calls createNewLog, and with the number of directories we
         // are using, if it picks one at random it will fail.
-        el.createNewLog();
-        LOG.info("This is the current log id: " + el.getCurrentLogId());
-        assertTrue("Wrong log id", el.getCurrentLogId() > 1);
+        ((EntryLogManagerBase) el.entryLogManager).createNewLog(0L);
```
sijie: - nit: provide a `getEntryLogManager` in EntryLogger. so we can mock it in future if needed.
- better to cast to `SingleEntryLogManager`. because you can expose `getCurrentLogId` in `SingleEntryLogManager`.

```
SingleEntryLogManager selm = (SingleEntryLogManager) (el.getEntryLogManager());
selm.createNewLog(0L);
assertTrue(selm.getCurrentLogId() > 1);
```

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SortedLedgerStorage.java
reddycharan: yes, it is not needed. Will remove.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/CreateNewLogTest.java
reddycharan: 1) I can add simple getter for EntryLogManager in EntryLogger, but wondering how would it help us in mocking if inside EntryLogger class direct reference of EntryLogger instance is used and using getter of EntryLogManager in all the places it is used in EntryLogger class is a stretch. 
2) if mocking is the requirement here then i can create initializeEntryLogManager method, which would initialize the EntryLogManager depending on the config, and for mocking purpose that method can be mocked to return what we want.
3) ok. in some places I wanted to test  getPreviousAllocatedEntryLogId, but it seems in this test it is not appropriate to change it to getPreviousAllocatedEntryLogId. will revert it back.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/ReadOnlyEntryLogger.java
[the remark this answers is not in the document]
reddycharan: ok..will revert it back to long primitive datatype in interface.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -84,6 +88,9 @@
  */
 public class EntryLogger {
     private static final Logger LOG = LoggerFactory.getLogger(EntryLogger.class);
+    static final Long UNASSIGNED_LEDGERID = Long.valueOf(-1);
```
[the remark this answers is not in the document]
reddycharan: changed it to long

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -120,21 +124,104 @@ public ConcurrentLongLongHashMap getLedgersMap() {
...
+        }
     }
 
-    volatile File currentDir;
     private final LedgerDirsManager ledgerDirsManager;
     private final boolean entryLogPerLedgerEnabled;
-    private final AtomicBoolean shouldCreateNewEntryLog = new AtomicBoolean(false);
 
-    private volatile long leastUnflushedLogId;
+    RecentEntryLogsStatus recentlyCreatedEntryLogsStatus;
```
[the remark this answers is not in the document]
reddycharan: done

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -788,89 +852,340 @@ private long readLastLogId(File f) {
...
+             * the addEntry traffic would be from GC and it is ok to proceed in
+             * this case.
+             */
+            if ((diskFull && (!allDisksFull)) || reachEntryLogLimit || (logChannel == null)) {
+                flushCurrentLog(logChannel, false);
+                createNewLog(ledger);
+            }
+
+            logChannel = getCurrentLogForLedger(ledger);
+            ByteBuf sizeBuffer = EntryLogger.this.sizeBuffer.get();
```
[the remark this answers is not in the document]
reddycharan: created sizeBufferForAdd threadlocal in EntryLogManagerBase.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -788,89 +852,340 @@ private long readLastLogId(File f) {
...
-                synchronized (this) {
-                    if (null == logChannelsToFlush) {
-                        logChannelsToFlush = channels;
-                    } else {
-                        logChannelsToFlush.addAll(0, channels);
-                    }
+
+        boolean reachEntryLogLimit(BufferedLogChannel logChannel, long size) {
+            if (logChannel == null) {
+                return false;
```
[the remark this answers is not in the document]
reddycharan: no. I don't want reachEntryLogLimit to do more than what it is supposed to do. let the caller deal with the extra logic.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -788,89 +852,340 @@ private long readLastLogId(File f) {
...
+                // we had to close the underlying file channel. Otherwise, we might end up
+                // leaking fds which cause the disk spaces could not be reclaimed.
+                closeFileChannel(channel);
+                recentlyCreatedEntryLogsStatus.flushRotatedEntryLog(channel.getLogId());
+                rotatedLogChannels.remove(channel);
+                LOG.info("Synced entry logger {} to disk.", channel.getLogId());
+            }
+        }
+
+        void flushCurrentLog(BufferedLogChannel logChannel, boolean forceMetadata) throws IOException {
```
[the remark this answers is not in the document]
reddycharan: will do

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -92,13 +99,10 @@
         private final long logId;
         private final EntryLogMetadata entryLogMetadata;
         private final File logFile;
+        private Long ledgerIdAssigned = UNASSIGNED_LEDGERID;
```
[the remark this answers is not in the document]
reddycharan: changed it

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
reddycharan: moved flushCurrentLogs and flushRotatedLogs from EntryLogManager interface to EntryLogManagerBase and added flush method in interface.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -499,176 +576,163 @@ void createNewLog() throws IOException {
...
-            List<File> list = ledgerDirsManager.getWritableLedgerDirsForNewLog();
-            Collections.shuffle(list);
-            // It would better not to overwrite existing entry log files
-            File newLogFile = null;
-            do {
+        synchronized BufferedLogChannel allocateNewLog(String suffix) throws IOException {
+            File dirForNextEntryLog;
+            List<File> list;
+
+            try {
```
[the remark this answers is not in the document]
reddycharan: removed this try catch block and reverted it to getWritableLedgerDirsForNewLog

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -880,39 +1195,6 @@ protected ByteBuf initialValue() throws Exception {
...
-            // log channel can be null because the file is deferred to be created when no writable ledger directory
-            // is available.
-            createNewLog();
-        }
-
-        int entrySize = entry.readableBytes() + 4; // Adding 4 bytes to prepend the size
-        boolean reachEntryLogLimit =
-            rollLog ? reachEntryLogLimit(entrySize) : readEntryLogHardLimit(entrySize);
-        // Create new log if logSizeLimit reached or current disk is full
-        boolean createNewLog = shouldCreateNewEntryLog.get();
```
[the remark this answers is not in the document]
reddycharan: i removed the need of LedgerDirsListener in EntryLogger. Instead in EntryLogManagerBase.addEntry, I check if dir is full

            boolean diskFull = (logChannel == null) ? false
                    : ledgerDirsManager.isDirFull(logChannel.getLogFile().getParentFile());

and diskFull will be used for determining if newlog needs to be created or not

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
reddycharan: rollLogs is not used in memtable flush related logic, but it is used in sortedledgerstorage.checkpoint logic. You made another comment related to rollLogs, I'm copying the same response..

i'm flexible here to make the change but

1) to begin with, from what i understand, this check (numBytesFlushed > 0) and rollLogs call is not needed. Since entries which come before checkpoint are already flushed from memtable (because of the way sortedledgerstorage.checkpoint) is called.
2) im not super convinced with moving from rollLogs method in interface to prepareCheckpoint, because this has meaning/use only for sortedledgerstorage and for single entrylog manager and having such a generic method name "prepareCheckpoint" seems to be a stretch.
3)in the method signature - "void prepareCheckpoint(Checkpoint checkpoint, long numBytesFlushedBetweenCheckpoint)" 'checkpoint' is not needed and we can remove it. parameter name 'numBytesFlushedBetweenCheckpoint' is not appropriate since it is just bytes added between previous flush and this checkpoint.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -788,89 +852,340 @@ private long readLastLogId(File f) {
...
     }
 
-    void flushRotatedLogs() throws IOException {
-        List<BufferedLogChannel> channels = null;
-        long flushedLogId = INVALID_LID;
-        synchronized (this) {
-            channels = logChannelsToFlush;
-            logChannelsToFlush = null;
+    abstract class EntryLogManagerBase implements EntryLogManager {
+        final Set<BufferedLogChannel> rotatedLogChannels;
```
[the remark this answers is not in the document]
reddycharan: 1) in both the cases (singleentrylog and entrylogperledger) rotatedLogChannels are dealt similarly. So it makes sense to have this variable and related methods (flushRotatedLogs) in Base class.
2) I dont see it to be much of an issue. Anyhow, I'll change it to List interface, but it is incorrect to have non-synchronized version of List for rotatedLogChannels. I'll use CopyOnWriteArrayList, so that the list can be modified while its iterator is being used.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
reddycharan: moved flushCurrentLogs and flushRotatedLogs from EntryLogManager interface to EntryLogManagerBase and added flush method in interface.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -499,176 +576,163 @@ void createNewLog() throws IOException {
...
+                    LOG.error("All Disks are not full, but getWritableLedgerDirs threw exception ", nwe);
+                    throw nwe;
+                }
+            }
+
+            dirForNextEntryLog = entryLogManager.getDirForNextEntryLog(list);
+
+            List<File> ledgersDirs = ledgerDirsManager.getAllLedgerDirs();
+            String logFileName;
+            while (true) {
```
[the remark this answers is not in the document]
reddycharan: the current logic with allocateNewLog is not completely correct for choosing the logId and the dir for newlog.
lets assume there are multiple ledgerdirs - dir1 and dir2, current value of preallocatedLogId is 5 and for what so ever reasons lets assume that there is a file with name "6.log" in dir2. Now, when we get call for allocateNewLog, with the current logic it would end up creating new entrylog with name "6.log" in dir1 though there is a file with same name in dir2. 

With my current code, I fixed it by making sure that there is no file under any ledgerdir with that logFileName before finalizing logFileName.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -499,176 +576,163 @@ void createNewLog() throws IOException {
...
+                     * for this it means preallocation future must have either
+                     * completed creating new log or still waiting for lock on
+                     * this object to execute allocateNewLog method. So we
+                     * should try to get result of the future without waiting.
+                     * If it fails with TimeoutException then call
+                     * allocateNewLog explicitly since we are holding the lock
+                     * on this anyhow.
+                     *
+                     */
+                    bc = preallocation.get(0, TimeUnit.MILLISECONDS);
```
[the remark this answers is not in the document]
reddycharan: i don't think it is correct for createNewLog to be guarded by createEntryLogLock and createNewLogForCompaction to be guarded by createCompactionLogLock when both call non-synchronized allocateNewLog method (which does operations on shared state, for eg:  preallocatedLogId) and also for clarity sake it would be easier to have all these methods synchronized on same lock (this). 

(same comment as I mentioned in code)
now since both createNewLog and allocateNewLog are synchronized on EntryLoggerAllocator.this object. So it is possible that a thread calling createNewLog would attain the lock on this object and get to this point but preallocation Future is starving for lock on EntryLoggerAllocator.this to execute allocateNewLog. Here since we attained lock for this it means preallocation future must have either completed creating new log or still waiting for lock on this object to execute allocateNewLog method. So we should try to get result of the future without waiting. If it fails with TimeoutException then call allocateNewLog explicitly since we are holding the lock on this anyhow.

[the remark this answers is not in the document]
sijie: @reddycharan okay, I remembered this context. however does this change need to be here for introducing EntryLogManager? If you can defer this change to a subsequent change, that would be better, also it is good for people understand the change here in future when looking back at the git history.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
reddycharan: btw, regarding order of flushRotatedLogs and flushCurrentLogs, I responded earlier as well - 

i think it is incorrect to call flushrotatedlogs first and then flushcurrentlogs. Consider this scenario - if in between flushRotatedLogs, flushCurrentLogs calls current activelog is rotated and added to rotatedlogs list then we would miss flushing that log in this current flush call. So it has to be first flushCurrentLogs and then flushRotatedLogs.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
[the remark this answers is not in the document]
sijie: - can you have a test case reproduce the behavior? I am not seeing how the case you described here can happen with original logic.

- if there is a problem with this logic, I think it is better to fix in a subsequent change. It is not really related to abstract EntryLogManager.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -788,89 +852,340 @@ private long readLastLogId(File f) {
...
+         */
+        @Override
+        public long addEntry(Long ledger, ByteBuf entry, boolean rollLog) throws IOException {
+
+            int entrySize = entry.readableBytes() + 4; // Adding 4 bytes to prepend the size
+            BufferedLogChannel logChannel = getCurrentLogForLedger(ledger);
+            boolean reachEntryLogLimit = rollLog ? reachEntryLogLimit(logChannel, entrySize)
+                    : readEntryLogHardLimit(logChannel, entrySize);
+            // Create new log if logSizeLimit reached or current disk is full
+            boolean diskFull = (logChannel == null) ? false
```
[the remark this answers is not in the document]
sijie: > it is not much of a change here. There are minor changes here and these changes make code more organized and appropriate.

understand you want to reorganize these. but if possible, can we do this in a subsequent change? it is hard to reason about why it was changed in this way. especially when other people  looking at the git commits in future.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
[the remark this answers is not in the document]
sijie: `reachEntryLogLimit` doesn't do more than it is supposed to do. a `null` log channel is effectively same a long channel reaches entry log limit, no?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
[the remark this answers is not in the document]
sijie: >  it is incorrect to have non-synchronized version of List for rotatedLogChannels. I'll use CopyOnWriteArrayList, so that the list can be modified while its iterator is being used.

why? the original code is synchronizing the rotatedLogChannels list properly. If you just move the logic into the abstraction, this logic should work.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
sijie: okay that's find, it you want to leave `rollLogs` here, I am fine.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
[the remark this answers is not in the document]
sijie: > i don't think it is correct for createNewLog to be guarded by createEntryLogLock and createNewLogForCompaction to be guarded by createCompactionLogLock when both call non-synchronized allocateNewLog method

I think we discussed that. Technically it is not a correctness issue. because the way how transactional compaction works.

The main question I have here, `preallocation.get(0, TimeUnit.MILLISECONS)`. if the allocation is still happening, this call will timeout immediately, and you go directly allocate a new entry log file again, which leaves an entry log with a smaller log id used later. this is a potential bug to me, because single-entry-log implementation assumes log id increasing at some extension (for both checkpoint, gc, compaction). 

regarding your point "but preallocation Future is starving for lock on EntryLoggerAllocator.this to execute allocateNewLog", can you show a test case around that?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
sijie: > Consider this scenario - if in between flushRotatedLogs, flushCurrentLogs calls current activelog is rotated and added to rotatedlogs list then we would miss flushing that log in this current flush call. So it has to be first flushCurrentLogs and then flushRotatedLogs.

Can you add a test case for that? I don't see how this can happen in single-entry-log implementation.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -950,19 +939,21 @@ public long addEntry(Long ledger, ByteBuf entry, boolean rollLog) throws IOExcep
...
+             * reached or if the logchannel is not initialized, then
              * createNewLog. If allDisks are full then proceed with the current
              * logChannel, since Bookie must have turned to readonly mode and
              * the addEntry traffic would be from GC and it is ok to proceed in
              * this case.
              */
             if ((diskFull && (!allDisksFull)) || reachEntryLogLimit || (logChannel == null)) {
-                flushCurrentLog(logChannel, false);
+                if (logChannel != null) {
+                    logChannel.flushAndForceWriteIfRegularFlush(false);
```
reddycharan: this is added because with this commit https://github.com/apache/bookkeeper/commit/30261eae3fd8ab25239d57cfb86a200d5f7b6b7d#diff-66d58c001864b142c348f9191407c410 it was deleted and it was a mistake.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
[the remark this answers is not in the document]
sijie: > I removed the need of LedgerDirsListener and another flag variable - shouldCreateNewEntryLog. Instead I check if the dir is full by calling "ledgerDirsManager.isDirFull(logChannel.getLogFile().getParentFile())" here. 

I don't think it is a good idea to check disk usage on every add request. This is much expensive than just check `shouldCreateNewEntryLog`. `shouldCreateNewEntryLog` is set when disk checker detect a disk is full, so it is an `asychronous` behavior to avoid any potential performance penalty.

> Also, I don't think it is appropriate to createNewLog in the case of "diskAlmostFull", because though disk is almost full,

createNewLog is *only* set when "current" disk is full at the case of `diskAlmotFull`



The main concern I have with getting rid of shouldCreateNewEntryLog is you change the `async-disk-checking` behavior to a `sync-disk-checking` on *every* add entry. this is unacceptable to me.

reddycharan:
Discussed with @sijie and decided to revert back some of the changes which are not completely related to introduction of EntryLogManager. Will try to address those issues in separate Issues/PRs.

@sijie added new commit for this PR, made changes as we discussed yesterday.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -788,89 +810,392 @@ private long readLastLogId(File f) {
...
-        List<BufferedLogChannel> channels = null;
-        long flushedLogId = INVALID_LID;
-        synchronized (this) {
-            channels = logChannelsToFlush;
-            logChannelsToFlush = null;
+    abstract class EntryLogManagerBase implements EntryLogManager {
+        final List<BufferedLogChannel> rotatedLogChannels;
+
+        EntryLogManagerBase() {
+            rotatedLogChannels = new CopyOnWriteArrayList<BufferedLogChannel>();
```
sijie: based on my understanding, you will use Set for per-ledger manager. why not move `rotatedLogChannels` to single-log manager, and keep the original logic unchanged? because I am not convinced that we need to change this to `CopyOnWriteArrayList`.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -788,89 +810,392 @@ private long readLastLogId(File f) {
...
+         * flush current logs.
+         */
+        abstract void flushCurrentLogs() throws IOException;
+
+        List<BufferedLogChannel> getRotatedLogChannels() {
+            return rotatedLogChannels;
+        }
+
+        @Override
+        public void flush() throws IOException {
```
sijie: I am not convinced this change is correct. it seems to me this might only be needed when you changed from `volatile List<BufferedLogChannel>` to `final CopyOnWriteArrayList<BufferedLogChannel>`. 

for sake of no changing the behavior of single-log manager, I would prefer keeping the behavior as same as before.

reddycharan: we missed discussing about this (order of flush in flush method)

i think it is incorrect to call flushrotatedlogs first and then flushcurrentlogs. Consider this scenario - if in between flushRotatedLogs, flushCurrentLogs calls current activelog is rotated and added to rotatedlogs list then we would miss flushing that log in this current flush call. So it has to be first flushCurrentLogs and then flushRotatedLogs.

even in the case of 'volatile List<BufferedLogChannel>' also i dont think it is correct. For the following reason - 

1) lets say flush method is called 
2) now as per the existing implementation if flushRotatedLogs is called first then lets after the execution of the following synchronized block (where 'logChannelsToFlush' is assigned to 'channels' and null is assigned to 'logChannelsToFlush'), lets say current active entrylog is rotated in another thread (because of reaching the size limit) then the current active entrylog will be added to new 'logChannelsToFlush' list. And when flushCurrentLogs is called it would flush the current active entrylog but not the one which is just rotated. So we missed flushing the just rotated log in this flush() call

https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java#L816

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
reddycharan: regarding rotatedlogs i don't see the need of having separate implementation for EntryLogManagerForSingleEntryLog and EntryLogManagerForEntryLogPerLedger thats why I kept it in EntryLogManagerBase. I don't have strong opinion on making this collection set or list, thats why I changed it back to List as we discussed (to minimize the change in behavior).

But for the sake of correctness and simplicity it has to be concurrentlist. 
for eg:
In this line 'logChannelsToFlush' is called in the log line, which would in turn call toString method of List and toString method of List returns "the string representation consists of a list of the collection's elements in the order they are returned by its iterator". So while iterator is iterating if an entrylog is added to rotatedLogChannels, LinkedList iterator will fail because "if the list is structurally modified at any time after the iterator is created, in any way except through the Iterator's own remove or add methods, the iterator will throw a ConcurrentModificationException."

https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java#L485

And because of CopyOnWriteArrayList, flushRotatedLogs method becomes fairly simple and straightforward.

Also in our case "rotatedLogChannels" list will be modified only when entrylog is rotated or when flushRotatedLogs is called, which is very less frequent. So having CopyOnWriteArrayList isn't going to change the behavior in any considerable way.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
sijie: @reddycharan 

first, I don't think flush current log should happen before rotated logs. because the way how checkpoint works for single-log manager. if current synchronization logic has problem, we should fix the synchronization, rather than reversing the order.  I am not sure about per-ledger log manager. 

second, I don't think this can happen with current "volatile" list approach. because the list creation and swap is `synchronized`. if you can show me a sequence and reproduce the problem in a test, I am convinced this is a problem. otherwise I will prefer us keeping the logic unchanged.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -788,89 +812,418 @@ private long readLastLogId(File f) {
...
+                logChannel.flushAndForceWrite(forceMetadata);
+                LOG.debug("Flush and sync current entry logger {}", logChannel.getLogId());
+            }
+        }
+
+        /*
+         * Creates a new log file. This method should be guarded by a lock,
+         * so callers of this method should be in right scope of the lock.
+         */
+        void createNewLog(long ledgerId) throws IOException {
```
sijie: please keep a method like `rollOldLogAndCreateNewLog`. because this method is replacing a lot of places that used to call `rollLog`. it would be good to have `roll log` in the method name, that would provide a more meaningful name.

```
public void rollOldLogAndCreateNewLog(long ledgerId) {
    createNewLog(long ledgerId);
}
```

sijie:
retest this please

Issue #1282, opened 2018-03-21
call to appendLedgersMap for compactionlog is missing


reddycharan:
**BUG REPORT**

1. Please describe the issue you observed:

with this change https://github.com/apache/bookkeeper/commit/3392beee5c70abe36d36604723e14d97b7764be9 compactionlog was introduced. But for compactionlog call to appendLedgersMap (// Append ledgers map at the end of entry log) isn't made. 

So for these logs, getEntryLogMetadata call has to call extractEntryLogMetadataByScanning for getting the metadata of entrylog. Which is a perf hit.

PR #1283, opened 2018-03-21
Issue #1282: call to appendLedgersMap in flushCompactionLog

reddycharan:

with this change https://github.com/apache/bookkeeper/commit/3392beee5c70abe36d36604723e14d97b7764be9 compactionlog was introduced. But for compactionlog call to appendLedgersMap (// Append ledgers map at the end of entry log) isn't made.

So for these logs, getEntryLogMetadata call has to call extractEntryLogMetadataByScanning for getting the metadata of entrylog. Which is a perf hit.

Master Issue: #1282

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
@@ -943,6 +943,7 @@ public class EntryLogger {
     void flushCompactionLog() throws IOException {
         synchronized (compactionLogLock) {
             if (compactionLogChannel != null) {
+                appendLedgersMap(compactionLogChannel);
                 compactionLogChannel.flushAndForceWrite(false);
                 LOG.info("Flushed compaction log file {} with logId.",
                     compactionLogChannel.getLogFile(),


sijie:
@yzang can you take a look at this change?

ivankelly:
@reddycharan will you update the PR description with the description from the issue, so we'll be able to see the reason for the change directly from the git log?

ivankelly:
retest this please

reddycharan:
@ivankelly updated PR description.

sijie:
@reddycharan did you look at @yzang 's comment?

reddycharan:
@yzang I'm not sure what you meant by "What if we flush the log channel successfully but not the ledger map?". In which case can it happen? Are you saying that this scenario is not handled completely?

Personally i"m not in favor of moving ledgersmap (appendLedgersMap logic) to flushRotatedLogs. let flushRotatedLogs method be about flushing and closing channels and setting state of the variables.

sijie:
@reddycharan 

>  I'm not sure what you meant by "What if we flush the log channel successfully but not the ledger map?". In which case can it happen? Are you saying that this scenario is not handled completely?

 so @yzang was asking if the flush ordering at line 494
```
             // flush the internal buffer back to filesystem but not sync disk
             // so the readers could access the data from filesystem.	
             logChannel.flush();	             logChannel.flush();
 	 
             // Append ledgers map at the end of entry log
             appendLedgersMap(logChannel);
```

@yzang : the `flush` here is writing all the buffer data back to filesystem. it is *NOT* flushing to disk. so the order at line 494 is correct: we write the buffer back to filesystem and append the ledger map after that. then checkpoint logic will fsync the entry log file (via flushAndForceWrite) (which happens at flushCurrentLog and flushRotatedLogs).


> Also shall we move line 494 to before line 841 instead? It makes more sense to append the ledger map in flush function than when creating a new log?

@yzang 

`flushRotatedLogs` only happens when checkpointing happen. The reason we put `appendLedgerMap()` at line 494 instead of before line 841 is a performance consideration. you can append ledger map into filesystem without fsync, so between you append ledger map and the real checkpoint happen, filesystem might already 'pdflush' the dirty pages back to disk. when the real fsync happen, there is no blocking I/O happening.

@reddycharan @yzang hope this clarify the questions.

PR #1286, opened 2018-03-22
Implement directly ChannelOutboundHandlerAdapter in BookieProtoEncoding#ResponseEncoder

eolivelli:
This change is mostly a clean up/refactor which drops intermediate MessageToMessageEncoder and MessageToMessageDecoder from BookieProtoEncoding

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtoEncoding.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtoEncoding.java
@@ -32,14 +32,13 @@ import io.netty.buffer.ByteBufOutputStream;
 import io.netty.buffer.Unpooled;
 import io.netty.channel.ChannelHandler.Sharable;
 import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelInboundHandlerAdapter;
 import io.netty.channel.ChannelOutboundHandlerAdapter;
 import io.netty.channel.ChannelPromise;
-import io.netty.handler.codec.MessageToMessageDecoder;
-import io.netty.handler.codec.MessageToMessageEncoder;
+import io.netty.util.ReferenceCountUtil;
 
 import java.io.IOException;
 import java.security.NoSuchAlgorithmException;
-import java.util.List;
 
 import org.apache.bookkeeper.proto.BookieProtocol.PacketHeader;
 import org.apache.bookkeeper.proto.checksum.MacDigestManager;
@@ -373,7 +372,7 @@ public class BookieProtoEncoding {
      * A request message encoder.
      */
     @Sharable
-    public static class RequestEncoder extends MessageToMessageEncoder<Object> {
+    public static class RequestEncoder extends ChannelOutboundHandlerAdapter {
 
         final EnDecoder reqPreV3;
         final EnDecoder reqV3;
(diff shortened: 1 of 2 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtoEncoding.java
```diff
@@ -452,20 +454,24 @@ protected void decode(ChannelHandlerContext ctx, Object msg, List<Object> out) t
...
                 LOG.error("Invalid response to encode to {}: {}", ctx.channel(), msg.getClass().getName());
-                out.add(msg);
+                ctx.write(msg, promise);
             }
         }
+
+        @Override
+        public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws Exception {
+            LOG.error("Generic error while encoding response", cause);
+        }
```
merlimat: I think this exception handler will be invoked in addition to the one we have already on the main handler.
Also, the exception here might not necessarely related to the encoder, but the `exceptionCaught()` is invoked for any exception that happens in the IO thread when serving this connection.

eolivelli: Will drop

eolivelli:
I will also change the decoder side once we agree on this approach

eolivelli:
retest this please

eolivelli:
@sijie  @merlimat  @ivankelly  I have updated the patch.
I will run this code on sites which reported the problem in the weekend and if no error occurs anymore I will remove the "WIP" flag.

eolivelli:
The fix is running without problem, my collegues are very happy. I am removing the flag and make the patch available for final review.
All lights are green.

I would like to cherry-pick this change to 4.6 branch

eolivelli:
@merlimat @sijie since you already took a look to this patch I will wait your ack before merging to master and to branch-4.6

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtoEncoding.java
```diff
@@ -412,28 +413,34 @@ protected void encode(ChannelHandlerContext ctx, Object msg, List<Object> out) t
             usingV3Protocol = true;
         }
 
-        @Override
-        protected void decode(ChannelHandlerContext ctx, Object msg, List<Object> out) throws Exception {
+         @Override
```
sijie: the space is not needed.

eolivelli: Will drop

eolivelli:
@sijie MessageToMessageEncoder uses reflection and does checks on objects types and it does other things not needed so my idea is that we are saving CPU cycles.

I will update the description, the change is more a cleanup/refactor and it does not address directly the problem, but actually with this refactor the problem does not occur anymore.

I agree that describing only part of the evidences is only confusing

sijie:
> MessageToMessageEncoder uses reflection and does checks on objects types and it does other things not needed so my idea is that we are saving CPU cycles.

"reflection" is only used during initialization on constructing the type matcher, no?

"checks on objects" -> the type we are using is Object and it is calling isInstance, which it is a very cheap op.

" my idea is that we are saving CPU cycles." -> I am not sure we are actually saving anything here.

--------------

To be honest, from code perspective, personally I would prefer MessageToMessage{Encoder,Decoder} because we don't need to handle "refcount release" stuff, which can become very tricky with netty version upgrades (e.g. if netty changes some of the behaviors)

However I understand you want to address your production issue and it seems this change can address your issue. I am fine with change as if the description is updated to what is actually changed here.

sijie:
> This change prevents errors seen when Bookie runs on java 9.
> There must be some race condition during serialization on messages when the bookie runs on Java 9, but there is way to create a simpler reproducer.
> Errors are reported only when the bookie is running with a real application

@eolivelli this description sounds claiming this change fixes race condition, which I don't think we should say that, because you don't really know what you are fixing.

I would suggest just saying this is a cleanup refactor. That is the most accurate description at this moment.

eolivelli:
@sijie thank you. Description updated. I will create an issue with logs and the full story and link to this patch next week

eolivelli:
@sijie I have addressed your comment and simplified the description

eolivelli:
retest this please

eolivelli:
I am going to merge this PR and cherry pick to 4.6 today, as soon as CI ends with success

PR #1288, opened 2018-03-23
Add sync variants of all methods in handles

ivankelly:
As discussed on the mailing list [1], this patch removes the
inconsistency around the naming of the close call on the new handle
APIs, by creating sync versions of each async calls, and renaming the
async versions to have the suffix "Async".

Most of the changes are very mechanical - just a copy of the old
method and some small fixups the javadoc. One thing to note is that
I've made a copy of the close and closeAsync methods in the
WriteHandle interface, so that the ReadHandle and Handle javadoc for
these methods do not have to talk about what it means to close/seal a
ledger.

Another change is that I've removed the SneakyThrows from close, that
would have also been needed on the other sync methods. Instead, I pass a
exception handler to FutureUtils which generates a BKException.

[1] https://lists.apache.org/thread.html/c3784cffb949438510d21e5eac8c0351865c6748c42c380e673a60db@%3Cdev.bookkeeper.apache.org%3E


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/Handle.java
```diff
@@ -21,7 +21,7 @@
 package org.apache.bookkeeper.client.api;
 
 import java.util.concurrent.CompletableFuture;
-import lombok.SneakyThrows;
+import org.apache.bookkeeper.client.impl.BKExceptionHandler;
```
sijie: Make BKExceptionHandler.HANDLER as part of BKException. rather than having a separate class. this also prevent api package referencing impl package, which allows producing a clean api package.

ivankelly:
@sijie updated

sijie:
@ivankelly you need to fix some tests.

eolivelli:
@ivankelly this is an important change to the public API.
I am totally fine with this change.

But shouldn't it need some kind of discussion/vote on the mailing list ?

Cc @sijie

eolivelli:
retest this please

eolivelli:
Sorry, I was not clear. The discussion happened but not in the usual form during last period (like BPs...) and no formal vote thread has been sent.

sijie:
@eolivelli - I think this is not proposing a brand new API. The original new API was introduced in a BP, this change here is more a incremental change to fix the inconsistency in the new API. so I don't think it worth a BP. but if you really feel strong about it, we can do a BP.

eolivelli:
Make

I feel fine with merging this PR and move forward.

eolivelli:
I did not want to close this one

ivankelly:
retest this please

ivankelly:
@eolivelli there was a integration test failure, so kicking again. I think it's unrelated, but want to be sure. Also, too late in the day to dig deeper :)

PR #1289, opened 2018-03-23
Allow multiple directories in DbLedgerStorage

merlimat:
In normal conditions, the max throughput that a bookie can sustain is mostly determined by how fast we can write entries on the journal. 

If we assume a *very* fast journal, for example using multiple journal threads and setting `journalSyncData=false` or with the journal on a RAM disk (there are some cases in which it might make sense), then the LedgerStorage becomes the "bottleneck". We need to be able to flush data on disk faster than the incoming rate, otherwise the write cache will get full and then we have to apply backpressure. 

In different testing scenario, the bottleneck has become the LedgerStoarge flush/checkpoint because it's done by a single thread. 

For smaller entries (1KB) the limit is ~250 K write/s and the limitation is due to preparing a batch with the offsets to write into rocksdb. Each insertion in the batch needs to cross JNI boundaries and that is expensive. 

For bigger entries (10KB), the limit is ~50K write/s (or 500MB /s). This mostly because the single flush thread cannot drive 100% IO utilization on the disks (since it's doing the rest of work). Raw disk writes in that test environment (12 HDDs) could reach 1.4GB/s.

This change is only relative to DbLedgerStorage and it's a simple refactoring that, when configured with multiple directories, it will hash a ledger into a certain directory. Each directory is flushed independently on its own thread.


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
```diff
@@ -20,323 +20,114 @@
...
-                LEDGER_INFO_CACHING_TIME_MINUTES, TimeUnit.MINUTES);
+        gcExecutor = Executors.newSingleThreadScheduledExecutor(new DefaultThreadFactory("GarbageCollector"));
 
-        entryLogger = new EntryLogger(conf, ledgerDirsManager);
-        gcThread = new GarbageCollectorThread(conf, ledgerManager, this, statsLogger);
+        ledgerStorageList = Lists.newArrayList();
+        for (File ledgerDir : ledgerDirsManager.getAllLedgerDirs()) {
+            // Create a ledger dirs manager for the single directory
+            File[] dirs = new File[1];
+            // Remove the `/current` suffix which will be appended again by LedgersDirManager
```
sijie: it might be worth checking if it is "/current" for safe

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
```diff
@@ -20,323 +20,114 @@
...
+            dirs[0] = ledgerDir.getParentFile();
+            LedgerDirsManager ldm = new LedgerDirsManager(conf, dirs, ledgerDirsManager.getDiskChecker(), statsLogger);
+            ledgerStorageList.add(newSingleDirectoryDbLedgerStorage(conf, ledgerManager, ldm, indexDirsManager,
```
sijie: there is a problem here - if you run with pulsar 1.22.0 (with yahoo branch), you upgrade to pulsar 2.0 (with bk 4.7). if you rollback, there is going to be a problem, it will cause data loss.

so you might consider having a defensing solution to prevent this happen or you have to document this well to make sure users are aware of this.

sijie: @merlimat - my question is if you already used multiple directories for storage. there is no change about number of directories. before this change, first directory is used for db index, after this change, all directories are used for db index.

It is probably not a problem if pulsar enforces people to use only one directory for ledger storage. but it is good to clarify that though.

merlimat: Oh I see. With db storage it was asserting that only 1 directory was present.

merlimat: (I mean, the assertion was in `DbLedgerStorage` itself)

sijie: oh i missed the assertion

merlimat:
In any case, I'm still testing this code.

sijie:
@merlimat db ledger storage test is failing

sijie:
@merlimat do you mind fixing the test failures?

PR #1298, opened 2018-03-27
newOpenLedgerOp and ReadHandle implementation for MockBookKeeper

ivankelly:
This will allow users of MockBookKeeper to use the new ReadHandle API.

There are two implementations of the ReadHandle interface. One pure,
which isn't backed by any production code, and another on
MockLedgerHandle which passes through to the pure implementation.


bookkeeper-server/src/test/java/org/apache/bookkeeper/client/MockBookKeeper.java
```diff
@@ -210,6 +213,66 @@ public void close() throws InterruptedException, BKException {
...
+
+            @Override
+            public OpenBuilder withLedgerId(long ledgerId) {
+                this.ledgerId = ledgerId;
+                return this;
+            }
+
+            @Override
+            public OpenBuilder withRecovery(boolean recovery) {
+                return this;
```
eolivelli: I guess this will be enhanced when we will implement a new mock WriteHandle

ivankelly: if needed. currently the mock calls asyncOpenLedger for asyncOpenNoRecovery. 

Originally this code comes from pulsar, where they don't use tailing anyhow.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/MockBookKeeper.java
```diff
@@ -210,6 +213,66 @@ public void close() throws InterruptedException, BKException {
         shutdown();
     }
 
+    @Override
+    public OpenBuilder newOpenLedgerOp() {
+        return new OpenBuilder() {
+            long ledgerId;
+            byte[] password;
+            org.apache.bookkeeper.client.api.DigestType digestType;
```
sijie: why not import this?

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/MockBookKeeper.java
```diff
@@ -210,6 +213,66 @@ public void close() throws InterruptedException, BKException {
         shutdown();
     }
 
+    @Override
+    public OpenBuilder newOpenLedgerOp() {
+        return new OpenBuilder() {
```
sijie: why not extending `OpenBuilderImpl` because most of the methods here are builder methods. what you really need is overriding the `execute` method.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/MockLedgerHandle.java
```diff
@@ -100,7 +108,7 @@ public void run() {
                 final Queue<LedgerEntry> seq = new ArrayDeque<LedgerEntry>();
                 long entryId = firstEntry;
                 while (entryId <= lastEntry && entryId < entries.size()) {
-                    seq.add(entries.get((int) entryId++));
+                    seq.add(new LedgerEntry(entries.get((int) entryId++).duplicate()));
```
sijie: there is a memory leak here, you don't need duplicate.

old `LedgerEntry` constructor will retain the buffer.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/MockReadHandle.java
```diff
@@ -0,0 +1,141 @@
...
+import org.apache.bookkeeper.client.api.ReadHandle;
+import org.apache.bookkeeper.client.impl.LedgerEntriesImpl;
+import org.apache.bookkeeper.client.impl.LedgerEntryImpl;
+
+
+/**
+ * Mock implementation of ReadHandle.
+ */
+@Slf4j
+public class MockReadHandle implements ReadHandle {
```
sijie: remove "public" to make it package protected.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/MockReadHandle.java
```diff
@@ -0,0 +1,141 @@
...
+        this.ledgerId = ledgerId;
+        this.metadata = metadata;
+        this.entries = entries;
+    }
+
+    @Override
+    public CompletableFuture<LedgerEntries> readAsync(long firstEntry, long lastEntry) {
+        CompletableFuture<LedgerEntries> promise = new CompletableFuture<>();
+        if (bk.isStopped()) {
+            promise.completeExceptionally(BKException.create(-1));
```
sijie: use `Code. ReadException` for -1.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/MockReadHandle.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    @Override
+    public CompletableFuture<LedgerEntries> readUnconfirmedAsync(long firstEntry, long lastEntry) {
+        return readAsync(firstEntry, lastEntry);
+    }
+
+    @Override
+    public CompletableFuture<Long> readLastAddConfirmedAsync() {
+        return CompletableFuture.completedFuture((long) (entries.size() - 1));
```
sijie: `return CompletableFuture.completedFuture(getLastAddConfirmed())`

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/MockReadHandle.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    @Override
+    public CompletableFuture<Long> tryReadLastAddConfirmedAsync() {
+        return readLastAddConfirmedAsync();
+    }
+
+    @Override
+    public long getLastAddConfirmed() {
+        return entries.size() - 1;
```
sijie: return the entry id of last entry

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/MockReadHandle.java
```diff
@@ -0,0 +1,141 @@
...
+    public boolean isClosed() {
+        return metadata.isClosed();
+    }
+
+    @Override
+    public CompletableFuture<LastConfirmedAndEntry> readLastAddConfirmedAndEntryAsync(long entryId,
+                                                                                      long timeOutInMillis,
+                                                                                      boolean parallel) {
+        CompletableFuture<LastConfirmedAndEntry> promise = new CompletableFuture<>();
+        // it doesn't make sense to implement long poll until WriteHandle is also implemented
```
sijie: the comment is misleading. long poll doesn't require write handle to implement. it is irrelative whether you use old write api or new write handle. please improve the comment or remove it.

jvrao:
What is mockBookKeeper? why does it need a new open interface? Any discussion on that?

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/MockBookKeeper.java
```diff
@@ -210,6 +214,43 @@ public void close() throws InterruptedException, BKException {
...
 
+    @Override
+    public OpenBuilder newOpenLedgerOp() {
+        return new OpenBuilderBase() {
+            @Override
+            public CompletableFuture<ReadHandle> execute() {
+                CompletableFuture<ReadHandle> promise = new CompletableFuture<ReadHandle>();
+
+                if (!validate()) {
+                    promise.completeExceptionally(new BKException.BKNoSuchLedgerExistsException());
```
sijie: return here

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/impl/OpenBuilderBase.java
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * Base class for open builders which does the mundane builder stuff.
+ */
+public abstract class OpenBuilderBase implements OpenBuilder {
+    static final Logger LOG = LoggerFactory.getLogger(OpenBuilderBase.class);
+
+    protected boolean recovery = false;
+    protected long ledgerId = -1;
+    protected byte[] password;
+    protected DigestType digestType = org.apache.bookkeeper.client.api.DigestType.CRC32;
```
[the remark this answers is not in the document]
ivankelly: yup. An oversight when moving.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/impl/OpenBuilderBase.java
```diff
@@ -0,0 +1,69 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * Base class for open builders which does the mundane builder stuff.
+ */
+public abstract class OpenBuilderBase implements OpenBuilder {
+    static final Logger LOG = LoggerFactory.getLogger(OpenBuilderBase.class);
+
+    protected boolean recovery = false;
+    protected long ledgerId = -1;
```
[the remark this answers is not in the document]
ivankelly: I was looking for it when I first wrote this. Seems to not exist. Added.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/MockBookKeeper.java
ivankelly: done.

ivankelly:
@jvrao 

> What is mockBookKeeper?

A mock implementation of the bookkeeper client, that can be used for testing without bringing up a whole BK client.

> why does it need a new open interface?

Because we added a new apis to the bookkeeper client. If a user is using mock bookkeeper, this change allows them to use the new apis for reading. In this specific case, in Pulsar we want to use ReadHandle.

sijie:
there are findbugs errors need to be fixed

PR #1301, opened 2018-03-28
Drop MD5 signatures from Release Guide and add a reference to git sha in VOTE email template

eolivelli:


sijie:
@eolivelli you also need another PR to update the `Download` page.

eolivelli:
I will do it as soon as possible (tomorrow).
Thank you I forgot about it
@sijie

PR #1302, opened 2018-03-28
Remove references to md5 checksums in download page

eolivelli:
According to current ASF suggestions, it is better to stop signing release with MD5, so we need to drop the links in the download page.

See paragraph which states "MD5 hashes are deprecated; please use SHA for new release" at https://www.apache.org/dev/release-signing.html


sijie:
@eolivelli can you put a description explaining why?

eolivelli:
@sijie  PR description updated. don't know if is is good to add a link to ASF website (which in turn will change in the future)

PR #1303, opened 2018-03-28
Make CRC32C_HASH a final static variable and log when we cannot use native crc32c library

merlimat:
For CRC32c there is a marked performance difference when the JNI library is available versus when it's not, since the fallback involves creating many `ByteBuffer`s and copy payloads. 

Additionally we should set the static variable to `final` so the optimizer can get rid of the `(CRC32C_HASH instanceof Sse42Crc32C)` checks.


merlimat:
retest this please

sijie:
retest this please

PR #1306, opened 2018-03-29
Minimize number of thread local instances in CRC32CDigestManager

merlimat:
CRC32CDigestManager uses thread local to store the current crc value (since the crc is computed on the headers and then on the payload). 

One issue is that the thread local instances are per-ledger and are never cleaned up from the thread local storage. As new ledgers get used, we'd have more memory used there. 

For digest we really need 1 single variable 

Changes: 
 * Use Netty FastThreadLocal instead of JDK ThreadLocal
 * Made the thread local variable static, since one thread is computing the checksum for an entry of a given ledge, before moving to next entry.
 * Minimized thread local `.get()` access since it's more expensive than regular variable access.
 * Removed the usage of `isNewCrc` since initial value for crc32c (as well as crc32) is 0


merlimat:
retest this please

sijie:
retest this please

PR #1307, opened 2018-03-29
Use reflection based CRC32 to pass direct memory pointer

merlimat:
The crc update with the bytebuffer is very expensive, especially in netty > 4.1.12.

```java
crc.get().update(data.nioBuffer());
```

Converting a direct`ByteBuf` into a `ByteBuffer` with `data.nioBuffer()` is actually resorting to allocating an unpooled `DirectByteBuffer` which is actually killing the GC (since direct memory is freed when the `DirectByteBuffer` instances are GCed, but the pauses are > 1 second).

This is an adaptation of a change from Yahoo branch at https://github.com/yahoo/bookkeeper/commit/6c01ca2921a998f3bf2e85cacb27867773e7ea28

Initially I though this change was not needed anymore in 4.7 codebase, and that is true for the CRC32c variant. For the regular CRC32, though, we need to avoid the bad behavior of unpooled direct buffers.

Additionally, as in #1306, the thread local variable was made static to avoid the leakage of instances per each ledger.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/DirectMemoryCRC32Digest.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/DirectMemoryCRC32Digest.java
@@ -29,7 +29,7 @@ import org.apache.bookkeeper.proto.checksum.CRC32DigestManager.CRC32Digest;
  * Specialized implementation of CRC32 digest that uses reflection on {@link CRC32} class to get access to
  * "updateByteBuffer" method and pass a direct memory pointer.
  */
-public class DirectMemoryCRC32Digest implements CRC32Digest {
+class DirectMemoryCRC32Digest implements CRC32Digest {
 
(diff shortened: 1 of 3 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/DirectMemoryCRC32Digest.java
```diff
@@ -0,0 +1,94 @@
...
+import java.lang.reflect.Method;
+import java.util.zip.CRC32;
+
+import org.apache.bookkeeper.proto.checksum.CRC32DigestManager.CRC32Digest;
+
+/**
+ * Specialized implementation of CRC32 digest that uses reflection on {@link CRC32} class to get access to
+ * "updateByteBuffer" method and pass a direct memory pointer.
+ */
+public class DirectMemoryCRC32Digest implements CRC32Digest {
```
sijie: nit: package private

Issue #1310, opened 2018-03-30
filter out confidential data (entry body and masterkey) from ProcessorV3 toString representation


reddycharan:
#**FEATURE REQUEST**

1. Please describe the feature you are requesting.

In certain logs instances of ProcessorV3 (for eg WriteEntryProcessorV3) are logged. For eg - https://github.com/apache/bookkeeper/blob/master/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/OrderedScheduler.java#L172

This would print output of toString method of WriteEntryProcessorV3, which gets its output from request.toString (https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PacketProcessorBaseV3.java#L94). request.toString would include all the fields of request like masterKey and body. For us 'body' is the customer data and masterKey contains ledger password details. So logging entry data and credentials details is not appropriate.

for eg. WriteEntryProcessorV3 toString representation
```
header {
  version: VERSION_THREE
  operation: ADD_ENTRY
  txnId: 2
}
addRequest {
  ledgerId: 0
  entryId: 0
  masterKey: "F\200\250<G\a3\033b\240\\\372\203 {\233;7^\373"
  body: "\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\377\377\377\377\377\377\377\377\000\000\000\000\000\000\000\t\000\000\000\000\317~\3405entrydata"
}
```

PR #1311, opened 2018-03-30
Issue #1310: Filter out body and masterKey from ProcessorV3s

reddycharan:

In certain logs instances of ProcessorV3 (for eg WriteEntryProcessorV3) are logged.

Logging entry data and credentials details is not appropriate. Filter out
those details

(have to consider other ProcessorV3s as well and testcases to validate this change)

Master Issue: #1310

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/WriteEntryProcessorV3.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/WriteEntryProcessorV3.java
@@ -170,4 +170,14 @@ class WriteEntryProcessorV3 extends PacketProcessorBaseV3 {
                          requestProcessor.addRequestStats);
         }
     }
+
+    /**
+     * this toString method filters out body and masterKey from the output.
+     * masterKey contains the password of the ledger and body is customer data,
+     * so it is not appropriate to have these in logs or system output.
+     */
+    @Override
+    public String toString() {
+        return RequestUtils.toSafeString(request);
+    }
 }
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadEntryProcessorV3.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadEntryProcessorV3.java
@@ -22,12 +22,15 @@ import com.google.common.util.concurrent.FutureCallback;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.SettableFuture;
 import com.google.protobuf.ByteString;
+
 import io.netty.buffer.ByteBuf;
 import io.netty.channel.Channel;
 import io.netty.util.ReferenceCountUtil;
+
 import java.io.IOException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.TimeUnit;
+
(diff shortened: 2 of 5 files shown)


reddycharan:
@sijie @merlimat before proceeding with other ProcessorV3s and testcases, I just wanted you to review this WIP PR.

eolivelli:
Sorry, just re read the issue.
It is fine to me, the only issue I am thinking is that it is quite expensive

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/WriteEntryProcessorV3.java
```diff
@@ -170,4 +170,18 @@ public void safeRun() {
...
     }
+
+    /**
+     * this toString method filters out body and masterKey from the output.
+     * masterKey contains the password of the ledger and body is customer data,
+     * so it is not appropriate to have these in logs or system output.
+     */
+    @Override
+    public String toString() {
+        Request.Builder reqBuilder = Request.newBuilder().setHeader(request.getHeader());
```
sijie: rather than constructing another request without masterkey. why not formatting the string itself (which it would void building another object)?

e.g.

```
@Override
public String toString() {

return MoreObjects.toStringHelper()
      .add("ledgerId", request.getLedgerId())
      ...
      .toString();

}
```

reddycharan: I've mixed opinion. 

AFAIU These kind of log lines (involving toString of ProcessorV3) are called rarely and may be it is ok to create a new request object. Also, it is easier to backlist one are two unwanted fields rather than whitelisting all other fields (including header/fields of header).

dlg99: 1. Response goes through the same OrderedSafeExecutor and can get logged similarly. Luckily, in the cases I see when response is wrapped into Runnable we also redefine toString for runnable.

2. BookieProtocol.Request defines toString as:

```
public String toString() {
            return String.format("Op(%d)[Ledger:%d,Entry:%d]", opCode, ledgerId, entryId);
        }
```

This should be enough, maybe add tx id to that

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/WriteEntryProcessorV3.java
```diff
@@ -170,4 +170,18 @@ public void safeRun() {
...
+        // clear body and masterKey
+        reqBuilder.setAddRequest(
+                AddRequest.newBuilder().mergeFrom(request.getAddRequest()).clearBody().clearMasterKey().buildPartial());
```
dlg99: looking at proto definition of request:

ReadRequest has masterKey as well, WriteLacRequest has body and masterKey etc.

reddycharan:
@sijie @dlg99 @eolivelli removed recreation of Request for toString. Now using MoreObjects.toStringHelper to build ProcessorV3's / request toString.

sijie:
@reddycharan can you rebase your branch?

reddycharan:
rebased it.

jvrao:
+1 LGTM

PR #1312, opened 2018-04-03
Upgrade Maven Assembly Plugin to 3.1.0

eolivelli:
Upgrade Maven Assembly Plugin to 3.1.0
    - upgrade Maven Assembly Plugin because 2.2.1 version has issues about file permissions
    - explicitly set tarLongFileMode to 'posix' in order to handle user ids with large values

PR #1313, opened 2018-04-03
Upgrade Maven Assembly Plugin to 3.1.0

eolivelli:
- upgrade Maven Assembly Plugin because 2.2.1 version has issues about file permissions
- explicitly set tarLongFileMode to 'posix' in order to handle user ids with large values


eolivelli:
I am marking this a blocker because the problem was found during 4.6.2 (see #1312) and it blocks the release even for 4.7.

Issue #1314, opened 2018-04-04
Allow ledger recovery writes on readonly bookies


sijie:
**FEATURE REQUEST**

1. Please describe the feature you are requesting.

Currently we allow fence requests going through readonly bookies, since fence requests are special read requests. However a ledger can only be sealed after it is successfully recovered. If bookies are in readonly, those recovery writes won't go through.

If there is a bookie outage happened (e.g. all bookies are readonly), all ledgers are not able to be sealed. It might be good to have a similar setting like [minUsableSizeForIndexFileCreation](https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java#L164) for recovery writes. 

This can improve operability during outage. 

2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have). Are you currently using any workarounds to address this issue?

*should-have*

3. Provide any additional detail on your proposed use case for this feature.

It is similar to #208 (for sorted ledger storage)

It can be helpful on pulsar handling bookkeeper outage : apache/pulsar#1458 apache/pulsar#1459

PR #1315, opened 2018-04-04
Issue 1314: Provide a mechanism to allow high priority writes to readonly bookies

sijie:

*Problem*

Currently we allow fence requests going through readonly bookies, since fence requests are special read requests. However a ledger can only be sealed after it is successfully recovered. If bookies are in readonly, those recovery writes won't go through.

If there is a bookie outage happened (e.g. all bookies are readonly), all ledgers are not able to be sealed. It might be good to have a similar setting like minUsableSizeForIndexFileCreation for recovery writes.

This can improve operability during outage.

*Solution*

- add a setting `minUsableSizeForHighPriorityWrites` to allow accepting high priority writes on readonly bookies.

Master Issue: #1314 

> ---
> quickly and easily:
>
>
>
> Otherwise:
> 
> 
> ---

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
@@ -886,24 +886,15 @@ public class Bookie extends BookieCriticalThread {
         return new LedgerDirsListener() {
 
             @Override
-            public void diskFull(File disk) {
-                // Nothing needs to be handled here.
-            }
-
-            @Override
-            public void diskAlmostFull(File disk) {
-                // Nothing needs to be handled here.
-            }
-
-            @Override
             public void diskFailed(File disk) {
                 // Shutdown the bookie on disk failure.
                 triggerBookieShutdown(ExitCode.BOOKIE_EXCEPTION);
             }
 
             @Override
-            public void allDisksFull() {
+            public void allDisksFull(boolean disksUnavailable) {
                 // Transition to readOnly mode on all disks full
+                stateManager.setHighPriorityWritesAvailability(!disksUnavailable);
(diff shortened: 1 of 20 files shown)


sijie:
need to write tests. but just putting this up to get review comments. @merlimat

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDirsManager.java
```diff
@@ -373,43 +373,49 @@ public NoWritableLedgerDirException(String errMsg) {
...
          * This will be notified whenever all disks are detected as full.
+         *
+         * <p>Normal writes will be rejected when disks are detected as "full". High priority writes
+         * such as ledger recovery writes can go through if disks are still available.
+         *
+         * @param disksUnavailable the parameter indicates whether disks are still available for
+         *                         taking high priority writes.
          */
-        void allDisksFull();
+        default void allDisksFull(boolean disksUnavailable) {}
```
merlimat: The negated logic of the parameter is a bit confusing (if `true` then it's not available).
What about naming it something like `boolean highPriorityWritesAllowed`?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -962,6 +955,15 @@ public boolean isReadOnly() {
         return stateManager.isReadOnly();
     }
 
+    /**
+     * Check whether Bookie is available for high priority writes.
+     *
+     * @return true if the bookie is able to take high priority writes.
+     */
+    public boolean isAvailableForHighPriorrityWrites() {
```
merlimat: double "r" in `Priorrity`

PR #1344, opened 2018-04-16
Allow to configure read-only mode in ZooKeeperClient

merlimat:
Underlying `Zookeeper` instance has an option to specify the client is fine to remain connected when the ZK quorum is lost, just in read-only mode. 

We should expose the "allow read-only mode" option in `ZooKeeperClient`.

One example of use case for this flag is when connecting to a ZK ensemble that is just used for configuration/metadata store, in which no ephemeral nodes are used. It's therefore better to keep the connection with ZK and be able to keep reading (possibly stale) data from ZK rather than no read at all. Concrete use case is for ZK session for Pulsar global ZK ensemble, where we read configuration data and we don't really need to have a valid session when brokers are trying to read.


ivankelly:
@merlimat Do you have a usecase in mind? Would be good to mention it in the commit message.

merlimat:
> @merlimat Do you have a usecase in mind? Would be good to mention it in the commit message. 

Good point. I've edited the description (which will be added in the commit log message)

sijie:
@merlimat it seems you need to fix tests

eolivelli:
retest this please

sijie:
@eolivelli retest doesn't help here, because there are some checkstyle errors in this PR.

eolivelli:
@sijie I knew but I kicked tests again because on jenkins the logs are lost (this pr is quite old), so I wanted to see the logs again.
I know that this PR need a fix from the author

sijie:
fyi, I rebased this PR to latest master and fixed the compilation issue. 

waiting for CI to pass and will then merge it for both 4.8 and 4.7.1.

sijie:
rerun integration tests

sijie:
run integration tests

sijie:
run integration tests

PR #1358, opened 2018-04-19
Issue #570: Introduce EntryMemTableWithParallelFlusher

reddycharan:

This is < sub-task4  > of Issue #570

When there is going to be entrylog per ledger, then for
better utilizing the availability of multiple entrylogs,
parallel flush of entrymemtable can be done. In this
sub-task EntryMemTableWithParallelFlusher is introduced,
which uses OrderedExecutor with "numOfMemtableFlushThreads"
number of threads.

Master Issue: #570

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryMemTableWithParallelFlusher.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryMemTableWithParallelFlusher.java
@@ -20,14 +20,9 @@
 package org.apache.bookkeeper.bookie;
 
 import java.io.IOException;
-import java.util.HashMap;
-import java.util.LinkedList;
-import java.util.List;
 import java.util.Map;
-import java.util.NavigableSet;
-import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.ConcurrentNavigableMap;
+import java.util.concurrent.Phaser;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
 import lombok.extern.slf4j.Slf4j;
@@ -42,15 +37,15 @@ import org.apache.bookkeeper.util.SafeRunnable;
  * EntryMemTableWithParallelFlusher.
  */
 @Slf4j
-public class EntryMemTableWithParallelFlusher extends EntryMemTable {
+class EntryMemTableWithParallelFlusher extends EntryMemTable {
 
     final OrderedExecutor flushExecutor;
-    private static final int MEMTABLE_FLUSH_TIMEOUT_IN_SECONDS = 60;
 
     public EntryMemTableWithParallelFlusher(final ServerConfiguration conf, final CheckpointSource source,
-            final StatsLogger statsLogger, OrderedExecutor flushExecutor) {
+            final StatsLogger statsLogger) {
         super(conf, source, statsLogger);
-        this.flushExecutor = flushExecutor;
+        this.flushExecutor = OrderedExecutor.newBuilder().numThreads(conf.getNumOfMemtableFlushThreads())
+                .name("MemtableFlushThreads").build();
     }
 
     /**
@@ -58,9 +53,9 @@ public class EntryMemTableWithParallelFlusher extends EntryMemTable {
      * EntryMemTable's flushSnapshot, but it uses flushExecutor
      * (OrderedExecutor) to process an entry through flusher.
      *
-     * <p>entryKeyValues of a ledger are grouped in a list and runnable is created
-     * to flush process all those entries and that runnable is submitted to the
-     * flushExecutor with ledgerId as the orderingKey.
+     * <p>SubMaps of the snapshot corresponding to the entries of the ledgers are
+     * created and submitted to the flushExecutor with ledgerId as the
+     * orderingKey to flush process the entries of a ledger.
      */
     @Override
     long flushSnapshot(final SkipListFlusher flusher, Checkpoint checkpoint) throws IOException {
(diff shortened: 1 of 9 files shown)


reddycharan:
i think tests failed because of space constraint in the test machine

2018-04-19 01:49:33,642 - ERROR - [main:DiskChecker@156] - Space left on device /tmp/IndexPersistenceMgr3196320391597035831Ledger/current : 18986319872, Used space fraction: 0.95140547 < threshold 0.95.

reddycharan:
retest this please

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SortedLedgerStorage.java
```diff
@@ -72,7 +74,13 @@ public void initialize(ServerConfiguration conf,
             checkpointSource,
             checkpointer,
             statsLogger);
-        this.memTable = new EntryMemTable(conf, checkpointSource, statsLogger);
+        if (conf.isEntryLogPerLedgerEnabled()) {
+            this.flushExecutor = OrderedExecutor.newBuilder().numThreads(conf.getNumOfMemtableFlushThreads())
+                    .name("MemtableFlushThreads").build();
+            this.memTable = new EntryMemTableWithParallelFlusher(conf, checkpointSource, statsLogger, flushExecutor);
```
sijie: it might be better to move flushExecutor to `EntryMemtableWithParallelFlusher` if possible, you can let `MemTable` implements `AutoClose`. #close on `EntryMemtable` is a no-op, while #close on `EntryMemtableWithParallelFlusher` is shutting down the executor. so we don't need this if-else checking in sorted ledger storage.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryMemTableWithParallelFlusher.java
```diff
@@ -0,0 +1,142 @@
...
+import org.apache.bookkeeper.common.util.OrderedExecutor;
+import org.apache.bookkeeper.conf.ServerConfiguration;
+import org.apache.bookkeeper.stats.StatsLogger;
+import org.apache.bookkeeper.util.SafeRunnable;
+
+/**
+ * EntryMemTableWithParallelFlusher.
+ */
+@Slf4j
+public class EntryMemTableWithParallelFlusher extends EntryMemTable {
```
sijie: nit: this class doesn't have to be 'public'. better to keep it as `package-protected`

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryMemTableWithParallelFlusher.java
```diff
@@ -0,0 +1,142 @@
...
+                        flushExecutor.executeOrdered(ledgerId.longValue(), new SafeRunnable() {
+                            @Override
+                            public void safeRun() {
+                                for (EntryKeyValue entryKeyValue : entryKeyValuesOfALedger) {
+                                    try {
+                                        flusher.process(ledgerId, entryKeyValue.getEntryId(),
+                                                entryKeyValue.getValueAsByteBuffer());
+                                        flushedSize.addAndGet(entryKeyValue.getLength());
+                                    } catch (NoLedgerException exception) {
+                                        log.debug("Got NoLedgerException while flushing entry: {}. "
```
sijie: add `if (log.isDebugEnabled())`

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryMemTableWithParallelFlusher.java
```diff
@@ -0,0 +1,142 @@
...
+                                                + "The ledger must be deleted "
+                                                + "after this entry is added to the Memtable",
+                                                entryKeyValue);
+                                        break;
+                                    } catch (Exception exc) {
+                                        log.error(
+                                                "Got Exception while trying to flush process entry: " + entryKeyValue,
```
sijie: use `{}` if possible

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryMemTableWithParallelFlusher.java
```diff
@@ -0,0 +1,142 @@
...
+                                                exc);
+                                        if (isFlushThreadInterrupted.compareAndSet(false, true)) {
+                                            exceptionWhileFlushingParallelly.set(exc);
+                                            mainFlushThread.interrupt();
+                                        }
+                                        // return without countdowning the latch since we got unexpected Exception
```
sijie: if you don't countdown, doesn't that cause the mainthread blocking forever?

Also I don't see a test case testing this part...

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryMemTableWithParallelFlusher.java
```diff
@@ -0,0 +1,142 @@
...
+     */
+    @Override
+    long flushSnapshot(final SkipListFlusher flusher, Checkpoint checkpoint) throws IOException {
+        AtomicLong flushedSize = new AtomicLong();
+        if (this.snapshot.compareTo(checkpoint) < 0) {
+            synchronized (this) {
+                EntrySkipList keyValues = this.snapshot;
+                if (keyValues.compareTo(checkpoint) < 0) {
+                    NavigableSet<EntryKey> keyValuesSet = keyValues.keySet();
+                    Map<Long, List<EntryKeyValue>> entryKeyValuesMap = new HashMap<Long, List<EntryKeyValue>>();
```
sijie: nit: have you benchmarked this before? my feeling this is going to be a gc churning ...

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/LedgerCacheTest.java
```diff
@@ -613,4 +655,107 @@ private ByteBuf generateEntry(long ledger, long entry) {
...
+        flushTestSortedLedgerStorage.addEntry(generateEntry(2, 2));
+        flushTestSortedLedgerStorage.addEntry(generateEntry(3, 2));
+
+        assertTrue("EntryMemTable SnapShot is expected to be empty", memTable.snapshot.isEmpty());
+        assertFalse("EntryMemTable is not expected to be empty", memTable.isEmpty());
+
+        // inject MemTableSizeLimitReached, so entrymemtable will be flushed
+        flushTestSortedLedgerStorage.setInjectMemTableSizeLimitReached(true);
+        flushTestSortedLedgerStorage.addEntry(generateEntry(1, 3));
+        Thread.sleep(1000);
```
sijie: hmm can we make the behavior more deterministic rather than depending on timing.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/LedgerCacheTest.java
```diff
@@ -613,4 +655,107 @@ private ByteBuf generateEntry(long ledger, long entry) {
         bb.writeBytes(data);
         return bb;
     }
+
+    @Test(timeout = 60000)
```
sijie: no timeout please

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/LedgerCacheTest.java
```diff
@@ -613,4 +655,107 @@ private ByteBuf generateEntry(long ledger, long entry) {
...
+        flushTestSortedLedgerStorage.setInjectMemTableSizeLimitReached(true);
+        flushTestSortedLedgerStorage.addEntry(generateEntry(1, 3));
+        Thread.sleep(1000);
+
+        // since we simulated sizeLimitReached, snapshot should have been created and flushed
+        assertTrue("EntryMemTable SnapShot is expected to be empty", memTable.snapshot.isEmpty());
+        assertEquals("Flusher called", 1, flushTestSortedLedgerStorage.getNumOfTimesFlushSnapshotCalled());
+    }
+
+    @Test(timeout = 60000)
```
sijie: no timeout please

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/LedgerCacheTest.java
```diff
@@ -613,4 +655,107 @@ private ByteBuf generateEntry(long ledger, long entry) {
...
+        flushTestSortedLedgerStorage.addEntry(generateEntry(1, 4));
+        flushTestSortedLedgerStorage.addEntry(generateEntry(2, 4));
+        flushTestSortedLedgerStorage.addEntry(generateEntry(3, 4));
+
+        // inject MemTableSizeLimitReached and FlushException, so entrymemtable flush will fail
+        flushTestSortedLedgerStorage.setInjectMemTableSizeLimitReached(true);
+        flushTestSortedLedgerStorage.setInjectFlushException(true, 1L);
+
+        flushTestSortedLedgerStorage.addEntry(generateEntry(1, 5));
+        Thread.sleep(1000);
```
sijie: please make this more deterministic rather than just sleeping

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/LedgerCacheTest.java
```diff
@@ -613,4 +655,107 @@ private ByteBuf generateEntry(long ledger, long entry) {
...
+        // since we simulate FlushException, memtable snapshot should not be empty
+        assertFalse("EntryMemTable SnapShot is not expected to be empty", memTable.snapshot.isEmpty());
+        assertEquals("Flusher called", 1, flushTestSortedLedgerStorage.getNumOfTimesFlushSnapshotCalled());
+
+        flushTestSortedLedgerStorage.setInjectFlushException(false, FlushTestSortedLedgerStorage.FORALLLEDGERS);
+        flushTestSortedLedgerStorage.addEntry(generateEntry(1, 5));
+        Thread.sleep(1000);
```
sijie: please make this more deterministic rather than just sleeping

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/LedgerCacheTest.java
```diff
@@ -613,4 +655,107 @@ private ByteBuf generateEntry(long ledger, long entry) {
...
+        /*
+         * since MemTableSizeLimitReached is already set to true, and flush
+         * exception is disabled, this time memtable snapshot should be flushed
+         */
+        assertTrue("EntryMemTable SnapShot is expected to be empty", memTable.snapshot.isEmpty());
+        assertEquals("Flusher called", 2, flushTestSortedLedgerStorage.getNumOfTimesFlushSnapshotCalled());
+    }
+
+    String[] createAndGetLedgerDirs(int numOfLedgerDirs) throws IOException {
```
sijie: static

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryMemTableWithParallelFlusher.java
reddycharan: will do

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryMemTableWithParallelFlusher.java
reddycharan: will do

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryMemTableWithParallelFlusher.java
reddycharan: hmm..if EntryMemTable class is public, then why should this class has to be package-protected?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryMemTableWithParallelFlusher.java
reddycharan: here we are logging exception also, I don't think there is any overloaded log.error method to do {}

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/LedgerCacheTest.java
reddycharan: this is old code, missed removing it. will do

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/LedgerCacheTest.java
reddycharan: this method uses createAndGetLedgerDirs, which is non-static method.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryMemTableWithParallelFlusher.java
sijie: well. EntryMemTable should be package-protected at the beginning. but we didn't do well on these stuffs before. it would be good to keep internal stuffs package-protected if possible.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryMemTableWithParallelFlusher.java
reddycharan: changed the logic here. with the new implementation there wont be any need of auxiliary datastructures in this method implementation.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryMemTableWithParallelFlusher.java
reddycharan: changed synchronization logic to use Phaser

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SortedLedgerStorage.java
reddycharan: moved

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/LedgerCacheTest.java
reddycharan: it was old code..i missed removing it.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/LedgerCacheTest.java
reddycharan: actually these sleep statements are not needed, in FlushTestSortedLedgerStorage.onSizeLimitReached memTable.flush is done synchronously, so removed these sleep statements.

reddycharan:
retest this please

reddycharan:
@sijie fixed all of your comments and tests have passed.

PR #1359, opened 2018-04-19
In EntryLogger.addEntry method, rollLog flag should not be ignored.

reddycharan:
In EntryLogger.addEntry method, rollLog flag should not be ignored.


sijie:
retest this please

Issue #791, opened 2017-11-29
Bookkeeper client allocates a lot of short-lived objects


dlg99:
We (Salesforce) upgraded NIC on our perf cluster to 20gbps  (as 10gbps had become a bottleneck) and could not get throughput to go above 15gbps. Proxy (our bk client/java app) used about 18 cores out of 36 (1800% CPU utilization in 'top').

Upon the investigation I've noticed that GC runs roughly 2 times per second (111 GCs in 60seconds) with GC pause of about 20ms.

From profiler (skipping proxy internal stacks) one can see that roughly 7G/sec are coming from protobuf + digest manager (see below), plus some extra (excluded here) from internal code.

I have some changes in work that reduced allocations to ~1.1GB/sec, need to finish some extra perf benchmarks. Changes include upgrade to protobuf 3.4 w/proto2 with use of unsafe wrappers. 
With these changes throughput went up close to 20gbps. CPU utilization is lower but still high (~1500%) with a lot of it coming from crc32, subject to another investigation/tuning/.. 
Majority (90%) of 1.1GB/sec are still coming from protobuf and it looks like the only other thing I can try there is to move to proto3 code gen. Quick prototype shown that simple move to proto3 (and its reduced immutability guarantees) results in flapping tests so the change is not that safe and requires more work, thus excluded form scope at this moment.

Top TLAB allocations, expanded to minimal relevant stacks (in 60 sec interval):
--------
Stack Trace	TLABs	Total TLAB Size(bytes)	Pressure(%)
bk-shade.com.google.proto_2.6.1.CodedInputStream.readRawBytesSlowPath(int)	347,980	192,147,664,488	35.738
   bk-shade.com.google.proto_2.6.1.CodedInputStream.readBytes()	347,980	192,147,664,488	35.738
      org.apache.bookkeeper.proto.BookkeeperProtocol$ReadResponse.<init>(CodedInputStream, ExtensionRegistryLite)	347,980	192,147,664,488	35.738
java.util.zip.CRC32.update(ByteBuffer)	36,317	97,655,274,184	18.163
   org.apache.bookkeeper.client.CRC32DigestManager.update(ByteBuf)	36,317	97,655,274,184	18.163
      org.apache.bookkeeper.client.DigestManager.verifyDigest(long, ByteBuf, boolean)	36,317	97,655,274,184	18.163
[...]

PR #1361, opened 2018-04-20
Issue #791: Eliminate byte[] copies in AddEntry code path

nicmichael:
The AddEntry code path in Bookkeeper Client performs several copies of the send buffer before sending it out to bookies, which not only costs CPU cycles for array copying, but also drives up object allocation rate and thus increases young GC frequency.

Currently, the send buffer is prepared by DigestManager.computeDigestAndPackageForSending(), which wraps the netty ByteBuf containing the data together with a (direct) header buffer into a ByteBufList. So far, we haven't yet allocated or copied any new buffers. But in order to send the buffer to the bookies, PendingAddOp.safeRun() will invoke sendWriteRequest() for each bookie in the write set, which will eventually invoke PerChannelBookieClient.addEntry() for each bookie. In here, we are now allocating two new byte[] and copying the buffer twice per bookie for V3 protocol with protobuf:
    byte[] toSendArray = toSend.toArray();      // 1st byte[] allocation and copy
    .setBody(ByteString.copyFrom(toSendArray)); // 2nd byte[] allocation and copy
If for example we're using a write set size of 3, then we allocate 6 byte arrays and copy the same buffer 6 times.

All 6 copies can be eliminated if DigestManager.computeDigestAndPackageForSending() would return a ByteBuf (or ByteBufList) that is backed by an array, which we can access and use without further copying it. PerChannelBookieClient.addEntry() could then wrap this array into a protobuf ByteString without performing any additional copies.

This change eliminates all byte[] allocations and copies in PCBC by:
1. allocating a new (pooled) array-backed heap ByteBuf for header, digest, and data in DigestManager.computeDigestAndPackageForSending()
2. copying the data payload into this newly allocated ByteBuf and returning it wrapped into a ByteBufList of size 1 (instead of header + body separately)
3. enhancing ByteBufList to give access to the underlying array if it consists of only a single ByteBuf that's backed by an array
4. extracting the underlying array in PerChannelBookieClient.addEntry() from ByteBufList and wrapping it into protobuf ByteString without copying it.

With this change, no more object allocations are required in this code path. DigestManager uses a pooled array (1). We're adding one arraycopy in (2) compared to the original code path, trading this against the allocations and copies we save later on.. Later steps, in particular step (4), is free of object allocation and array copies.
As a result, the object allocation rate in the AddEntry code path is reduced significantly. In case of a write set size of 3, it is reduced from 6 to 0. This translates into much less frequent young generation garbage collections (in my experiments with write-heavy workload by ~4x overall).


bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/DigestManager.java
```diff
@@ -89,17 +107,27 @@ public static DigestManager instantiate(long ledgerId, byte[] passwd, DigestType
...
+            sendBuffer.writeBytes(data.array(), data.arrayOffset(), data.readableBytes());
+        } else {
+            // fall-back to slow path with extra object allocation and copy
+            byte[] sendArray = new byte[sendBuffer.readableBytes()];
+            sendBuffer.getBytes(sendBuffer.readerIndex(), sendArray);
+            sendBuffer.writeBytes(sendArray);
+        }
 
-        return ByteBufList.get(headersBuffer, data);
+        return ByteBufList.get(sendBuffer);
```
eolivelli: Maybe in a follow up change we can get rid of ByteBufList and use directly the ByteBuf.

Cc @merlimat

eolivelli:
@sijie we can think about cherry picking to 4.7 branch

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/DigestManager.java
```diff
@@ -40,10 +41,19 @@
 public abstract class DigestManager {
     private static final Logger logger = LoggerFactory.getLogger(DigestManager.class);
 
+    /*
+     * This ByteBufAllocator gives us (pooled) heap ByteBuf's that are backed by an accessible array.
+     * The backing array can be used to instantiate a protobuf ByteString without copying the array.
+     * This allocator should be used when using protobuf to eliminate array copies when converting
+     * ByteBuf to ByteString.
+     */
+    private static final ByteBufAllocator HEAP_BYTEBUF_ALLOCATOR = new PooledByteBufAllocator(false);
```
sijie: I think you want need to provide here is an unpooled bytebuf allocator:

https://netty.io/4.0/api/io/netty/buffer/UnpooledByteBufAllocator.html#UnpooledByteBufAllocator-boolean-boolean-

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/DigestManager.java
```diff
@@ -89,17 +107,27 @@ public static DigestManager instantiate(long ledgerId, byte[] passwd, DigestType
      */
     public ByteBufList computeDigestAndPackageForSending(long entryId, long lastAddConfirmed, long length,
             ByteBuf data) {
-        ByteBuf headersBuffer = PooledByteBufAllocator.DEFAULT.buffer(METADATA_LENGTH + macCodeLength);
-        headersBuffer.writeLong(ledgerId);
-        headersBuffer.writeLong(entryId);
-        headersBuffer.writeLong(lastAddConfirmed);
-        headersBuffer.writeLong(length);
+        ByteBuf sendBuffer = byteBufAllocator.buffer(METADATA_LENGTH + macCodeLength + data.readableBytes());
```
sijie: I think it would be good to keep the original method:

if (useV2Protocol) {
    // use the original way
} else {
    // use the new way
}

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/DigestManager.java
sijie: @eolivelli no. see my comment.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
```diff
@@ -578,12 +581,13 @@ void addEntry(final long ledgerId, byte[] masterKey, final long entryId, ByteBuf
...
             }
 
-            byte[] toSendArray = toSend.toArray();
             AddRequest.Builder addBuilder = AddRequest.newBuilder()
                     .setLedgerId(ledgerId)
                     .setEntryId(entryId)
-                    .setMasterKey(ByteString.copyFrom(masterKey))
-                    .setBody(ByteString.copyFrom(toSendArray));
+                    .setMasterKey(UnsafeByteOperations.unsafeWrap(masterKey))
+                    .setBody(toSend.hasArray()
```
sijie: this can be:

```
if (toSend.hasArray) {
    UnsafeByteOperations.unsafeWrap(toSend.array(), toSend.arrayOffset(), toSend.readableBytes());
} else if (toSend.size() == 1) { // only one bytebuffer
   UnsafeByteOperations.unsafeWrap(toSend.getBuffer(0).nioBuffer());
} else {
   UnsafeByteOperations.unsafeWrap(toSend.toArray());
}
```

sijie:
@eolivelli yes we should cherry-pick this for 4.7.0

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/DigestManager.java
merlimat: @eolivelli this would make an additional copy of the payload buffer

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/DigestManager.java
```diff
@@ -57,22 +67,30 @@ void update(byte[] data) {
...
         macCodeLength = getMacCodeLength();
     }
 
     public static DigestManager instantiate(long ledgerId, byte[] passwd, DigestType digestType)
             throws GeneralSecurityException {
+        return instantiate(ledgerId, passwd, digestType, false);
+    }
+
+    public static DigestManager instantiate(long ledgerId, byte[] passwd, DigestType digestType,
+            boolean useDirectBufferAllocator) throws GeneralSecurityException {
```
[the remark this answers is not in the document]
nicmichael: Oh yes, I suppose so. I shouldn't have done that last change and should have stayed with Unpooled. Will revert again.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/DigestManager.java
nicmichael: Yes. We can just call Unpooled.buffer() (same as if we instantiate UnpooledByteBufAllocator(false)).

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/DigestManager.java
nicmichael: Ok

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
nicmichael: Ok.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/DigestManager.java
```diff
@@ -89,17 +96,48 @@ public static DigestManager instantiate(long ledgerId, byte[] passwd, DigestType
...
+
+            update(sendBuffer);
+            update(data);
+            populateValueAndReset(sendBuffer);
+
+            if (data.hasArray()) {
+                // fast path: single copy into existing array
+                sendBuffer.writeBytes(data.array(), data.arrayOffset(), data.readableBytes());
+            } else {
+                // fall-back to slow path with extra object allocation and copy
```
sijie: the logic here seems to be wrong. I think you need to write data into the buffer, no?

any reason why you can't just call `sendBuffer.writeBytes(data, data.readerIndex(), data.readableBytes())` to replace line 129-137?

https://netty.io/4.0/api/io/netty/buffer/ByteBuf.html#writeBytes-io.netty.buffer.ByteBuf-int-int-

nicmichael: This "transfers the specified source buffer's data to this buffer". The source is "data", our payload, and the destination is "sendBuffer", the buffer containing header, checksum, and payload. Am I missing something?

But I think we can replace the entire block by
sendBuffer.writeBytes(data);

It will figure out the offset/length itself. The distinction isn't necessary indeed.

nicmichael: I haven't yet figured out the Netty API and its use in bookkeeper.

It seems to me that writeBytes(ByteBuf src) will consider the source buffers reader index and readableBytes, and according to API doc "This method is basically same with {@link #writeBytes(ByteBuf, int, int)}, except that this method increases the {@code readerIndex} of the source buffer by the number of the transferred bytes while {@link #writeBytes(ByteBuf, int, int)} does not."

I'm not sure whether this difference matters or not... Are we reading from "data" (they payload) elsewhere again? If yes, I suppose we have to use writeBytes(ByteBuf, int, int) instead. Else #writeBytes(ByteBuf) would be just fine, right?

sijie: `sendBuffer.writeBytes(data)` is basically `sendBuffer.writeBytes(data, int, int)`. `writeBytes(data)` will modify the readerIndex data. typically the practice is if we don't need to modify the reader/writer index for a bytebuf, try not to modify it. that's much clearer

PR #1366, opened 2018-04-24
[Table Service] Dlog based checkpoint store

sijie:
*Motivation*

Currently the table range stores are using local filesystem as a checkpoint store. It is okay for running as standalone mode.
But it doesn't work to run in a distributed mode. This change is introducing a dlog based checkpoint store to make sure all
the sst files of table ranges are durably checkpointed into bookkeeper itself.

*Solution*

Introduced a dlog based checkpoint store.

### DIFF
--- a/stream/statelib/src/main/java/org/apache/bookkeeper/statelib/impl/rocksdb/checkpoint/dlog/DLCheckpointStore.java
+++ b/stream/statelib/src/main/java/org/apache/bookkeeper/statelib/impl/rocksdb/checkpoint/dlog/DLCheckpointStore.java
@@ -1,0 +1,154 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.bookkeeper.statelib.impl.rocksdb.checkpoint.dlog;
+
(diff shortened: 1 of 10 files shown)


stream/statelib/src/main/java/org/apache/bookkeeper/statelib/impl/rocksdb/checkpoint/dlog/DLCheckpointStore.java
```diff
@@ -0,0 +1,154 @@
...
+
+    @Override
+    public boolean fileExists(String filePath) throws IOException {
+        return namespace.logExists(filePath);
+    }
+
+    @Override
+    public long getFileLength(String filePath) throws IOException {
+        try (DistributedLogManager dlm = namespace.openLog(filePath)) {
+            return dlm.getLastTxId();
```
eolivelli: Is this right ?

stream/statelib/src/main/java/org/apache/bookkeeper/statelib/impl/rocksdb/checkpoint/dlog/DLInputStream.java
```diff
@@ -0,0 +1,232 @@
...
+
+/**
+ * The input stream for a distributedlog stream.
+ */
+@Slf4j
+class DLInputStream extends InputStream {
+
+    private static final long REOPEN_READER_SKIP_BYTES = 4 * 1024 * 1024; // 4MB
+
+    private static class RecordStream {
```
jiazhai: Is there a need to leverage LogReader.readBulk to read a few records each time? so the read may be faster.

sijie: we don't need to call `readBulk`. `readBulk` is a just a wrapper over `readNext`. DL has readahead mechanism, so there are already entries read into a readahead buffer. so `readNext` is just polling records out of the readahead buffer.

stream/statelib/src/main/java/org/apache/bookkeeper/statelib/impl/rocksdb/checkpoint/dlog/DLCheckpointStore.java
sijie: yes. it is right. we [use `txid` as the offset](https://github.com/apache/bookkeeper/pull/1366/files/4e27d748940e457f1800222be20ff18d22420675#diff-f90a2ff7fd99bf21f6f8bdb27b75b7c0R97).

sijie:
retest this please

PR #1372, opened 2018-04-27
OrderedScheduler#chooseThread(key) handle null key

sijie:

*Motivation*

It is more convenient if OrderedScheduler#chooseThread(key) can handle null key. so the applications who is using #chooseThread don't need to worry which method to choose - `chooseThread()` vs `chooseThread(key)`.

*Solution*

in `chooseThread(key)`, fallback to randomly pickup a thread if key is null.

### DIFF
--- a/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/OrderedExecutor.java
+++ b/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/OrderedExecutor.java
@@ -348,7 +348,11 @@ public class OrderedExecutor implements ExecutorService {
             return threads[0];
         }
 
-        return threads[MathUtils.signSafeMod(orderingKey.hashCode(), threads.length)];
+        if (null == orderingKey) {
+            return threads[rand.nextInt(threads.length)];
+        } else {
+            return threads[MathUtils.signSafeMod(orderingKey.hashCode(), threads.length)];
+        }
     }
 
     /**


sijie:
retest this please

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/OrderedExecutor.java
```diff
@@ -348,7 +348,11 @@ public ExecutorService chooseThread(Object orderingKey) {
             return threads[0];
         }
 
-        return threads[MathUtils.signSafeMod(orderingKey.hashCode(), threads.length)];
+        if (null == orderingKey) {
+            return threads[rand.nextInt(threads.length)];
+        } else {
+            return threads[MathUtils.signSafeMod(orderingKey.hashCode(), threads.length)];
+        }
```
jvrao: Looks fine, but wondering what is the usecase of this? I worry it might mask bugs as ordered safe executor is an absolute must to protect the correctness (wherever it is needed). So callers need to know what they are doing.

sijie: it is for convenience, so in the caller, I don't have to do `if (orderKey != null) chooseThread(orderKey); else chooseThread(..)`. I can have a separate util function to do that, but I feel it is better to do it at the executor level.

sijie:
retest this please

PR #1373, opened 2018-04-27
Make bookie state manager & registration manager reusable outside of bookie

sijie:

*Motivation*

Since 4.6.0, we have already abstracted a pretty good interfaces around registration service and state management. Most of the classes can actually be reused for use cases that have similar service discovery and writable/readonly state transition management.

*Solution*

The change is try to cleanup `BookieStateManager` and `ZKRegistrationManager` to make them more generic and can be reused for other usage.

*Result*

Registration Interfaces and StateManager can be used outside of bookie.


sijie:
retest this please

sijie:
retest this please

Issue #1343, opened 2018-04-14
Advisory (optimistic) write close 


reddycharan:
**FEATURE REQUEST**

1. Please describe the feature you are requesting.

With https://github.com/apache/bookkeeper/issues/570 entrylog per ledger feature will be introduced. With this feature there will be dedicated entrylog for each ledger. With the <sub-task6> of issue-570 EntryLogManagerForEntryLogPerLedger will be introduced. 

Since there is going to be entrylog per ledger, with the current Bookie implementation there is no way to know when the entrylog is writeclosed and the entrylog for the active ledger can be rotated. So in <sub-task6> of issue-57, as a first step of handling this, expireAfterAccess (each entrylog should be automatically rotated from the current active entrylog list once a fixed duration has elapsed after the  last access of entrylog for addEntry) will be implemented and also the maximum number of entrylogs that can be active at a given time (this is for limiting number of entrylogs/filedescriptors open at a given time).

The above mentioned approaches are preliminary ways of handling but it would be ideal to have explicit call to EntryLogger when the write for this ledger is done so that it can rotate the entrylog as soon as it is done writing. This will minimize the number of entrylogs/file descriptors that are open/active and also it will make progress in leastUnflushedLogId, so that GarbageCollectorThread can consider these entrylogs for garbage collection. So as part of next step, following can be done

1. have explicit write close request. This write close request should be sent to the current ensemble when the write handle is closed. This should be just optimistic write close operation and callback of this operation should be just logger, saying if it is succeeded / partially succeeded / failed. This should be done asynchronously and the write handle close operation should not be blocked for this response.
2. In the case of ledger recover open, readrequest (ReadEntryProcessorV3) with fence flag, can take care of calling appropriate methods in ledgerdescriptor / entry logger to rotate the entrylog for the ledger
3.  in the case of auto-replication case, LedgerFragmentReplicator already uses bookieclient for addEntry (bkc.getBookieClient().addEntry), the same bookieClient instance can be used to call explicit writeClose method in  bookieclient.
4. in the case of any write failure in bookie for an entry, then before sending error response to the client do entrylog rotation

2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have). Are you currently using any workarounds to address this issue?

This is should-have feature for entrylogperledger 

3. If there are some sub-tasks using -[] for each subtask and create a corresponding issue to map to the sub task:
[...]

PR #1374, opened 2018-04-28
BP-32: Advisory (optimistic) write close

reddycharan:

initial proposal for Advisory (optimisitic) write close

Master Issue: #1343

### DIFF
--- a/site/bps/BP-32-advisory-write-close.md
+++ b/site/bps/BP-32-advisory-write-close.md
@@ -1,0 +1,54 @@
+---
+title: "BP-32: Advisory (optimistic) write close"
+issue: https://github.com/apache/bookkeeper/issues/1343
+state: Under Discussion
+release: N/A
+---
+
+### Motivation
+
+With entrylog per ledger feature (https://github.com/apache/bookkeeper/issues/570) there will be dedicated entrylog for each ledger and it provides EntryLogManagerForEntryLogPerLedger (EntryLogManager). Since there is going to be entrylog per ledger, with the current Bookie implementation there is no way for EntryLogManagerForEntryLogPerLedger (EntryLogManager) to know when the entrylog is writeclosed, so that entrylog for the active ledger can be rotated. So it would be ideal to have explicit call to EntryLogger when the write for this ledger is done and it is writeclosed, so that it can rotate the entrylog as soon as it is write closed. This will minimize the number of entrylogs/file descriptors that are open/active and also it will make progress in leastUnflushedLogId, so that GarbageCollectorThread can consider these entrylogs for garbage collection.
+
+### Proposed Changes
+
+So in entrylog per ledger feature implementation, as a first step of handling this, expireAfterAccess (each entrylog should be automatically rotated from the current active entrylog list once a fixed duration has elapsed after the last access of entrylog for addEntry) will be implemented and also the maximum number of entrylogs that can be active at a given time (this is for limiting number of entrylogs/filedescriptors open at a given time).
+
+The above mentioned approaches are preliminary ways of handling but it would be ideal to have explicit call to EntryLogger when the write for this ledger is done so that it can rotate the entrylog as soon as it is write closed. So to implement this following can be done
+
+- have explicit write close request. This write close request should be sent to the current ensemble when the write handle is closed. This should be just optimistic write close operation and callback of this operation should be just logger, saying if it is succeeded / partially succeeded / failed. This should be done asynchronously and the write handle close operation should not be blocked for this response.
+
+- In the case of ledger recover open, readrequest (ReadEntryProcessorV3) with fence flag, can take care of calling appropriate methods in ledgerdescriptor / entry logger to rotate the entrylog for the ledger
+
+- in the case of auto-replication case, LedgerFragmentReplicator already uses bookieclient for addEntry (bkc.getBookieClient().addEntry), the same bookieClient instance can be used to call explicit writeClose method in bookieclient.
+
+- in the case of any write failure in bookie for an entry, then before sending error response to the client do entrylog rotation
+
+EntryLogManagerForEntryLogPerLedger is the class which is last layer in the stack to receive this writeClose call for a ledger, it would store this info in in-memory data structure and when next next checkpoint is called it would rotate entrylogs of these writeclosed ledgers.
+
+### Public Interfaces
+
+As explained above, this advisory write close should be transparent to Bookkeeper API user, this should be piggybagged in writehandle close call and other internal Bookie / BookieClient internal class methods (ReadEntryProcessorV3.readrequest and LedgerFragmentReplicator). But this feature introduces new protobuf message between Client and Bookie.
+
+message WriteCloseRequest {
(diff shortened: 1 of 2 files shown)


reddycharan:
@sijie I think I followed process needed for BP, let me know if I'm missing something here or if BP needs more details.

@jvrao fyi.

sijie:
@reddycharan thank you for sending this. will take a look.

can you update the bookkeeper_proposals.md also

- bump the bp number
- and add the BP link to "inprogress" section.

site/bps/BP-32-advisory-write-close.md
```diff
@@ -0,0 +1,54 @@
...
+message WriteCloseRequest {
+    required int64 ledgerId = 1;
+    required bytes masterKey = 2;    
+}
+
+message WriteCloseResponse {
+    required StatusCode status = 1;
+    required int64 ledgerId = 2;  
+}
+
```
jvrao: Do we need a response? It may be tiny, but unnecessary to send/queue/process.

jvrao: @sijie what do you think?

eolivelli: @jvrao good point

 it depends on how much we care to inform the bookie even in case of temporary errors. This will need retries, but I dont know if it is worth.
I think that for this BP the request will be fire and forget.
We can skip it now and then add a response in the future if needed

eolivelli: @jvrao good point

 it depends on how much we care to inform the bookie even in case of temporary errors. This will need retries, but I dont know if it is worth.
I think that for this BP the request will be fire and forget.
We can skip it now and then add a response in the future if needed

site/bps/BP-32-advisory-write-close.md
```diff
@@ -0,0 +1,54 @@
...
+### Proposed Changes
+
+So in entrylog per ledger feature implementation, as a first step of handling this, expireAfterAccess (each entrylog should be automatically rotated from the current active entrylog list once a fixed duration has elapsed after the last access of entrylog for addEntry) will be implemented and also the maximum number of entrylogs that can be active at a given time (this is for limiting number of entrylogs/filedescriptors open at a given time).
+
+The above mentioned approaches are preliminary ways of handling but it would be ideal to have explicit call to EntryLogger when the write for this ledger is done so that it can rotate the entrylog as soon as it is write closed. So to implement this following can be done
+
+- have explicit write close request. This write close request should be sent to the current ensemble when the write handle is closed. This should be just optimistic write close operation and callback of this operation should be just logger, saying if it is succeeded / partially succeeded / failed. This should be done asynchronously and the write handle close operation should not be blocked for this response.
```
sijie: I would prefer having a flag to control whether the client should wait on response or not. 

That says it would be good to allow application to configure whether it is an *advisory* close or an *enforced* close.

site/bps/BP-32-advisory-write-close.md
```diff
@@ -0,0 +1,54 @@
...
+
+### Proposed Changes
+
+So in entrylog per ledger feature implementation, as a first step of handling this, expireAfterAccess (each entrylog should be automatically rotated from the current active entrylog list once a fixed duration has elapsed after the last access of entrylog for addEntry) will be implemented and also the maximum number of entrylogs that can be active at a given time (this is for limiting number of entrylogs/filedescriptors open at a given time).
```
sijie: nit: it would be good to update the fileinfo to set the ledger state from `fenced` to `closed`, so it can trigger all long poll requests to be notified with last entry id, so the long polling readers know the ledger is closed without interacting with zookeeper.

this is a request from dlog perspective. but you dont need to implement in this BP. but it would be good to keep this in mind when design the request/response protocol, so I can extend in future.

site/bps/BP-32-advisory-write-close.md
```diff
@@ -0,0 +1,54 @@
...
+
+The above mentioned approaches are preliminary ways of handling but it would be ideal to have explicit call to EntryLogger when the write for this ledger is done so that it can rotate the entrylog as soon as it is write closed. So to implement this following can be done
+
+- have explicit write close request. This write close request should be sent to the current ensemble when the write handle is closed. This should be just optimistic write close operation and callback of this operation should be just logger, saying if it is succeeded / partially succeeded / failed. This should be done asynchronously and the write handle close operation should not be blocked for this response.
+
+- In the case of ledger recover open, readrequest (ReadEntryProcessorV3) with fence flag, can take care of calling appropriate methods in ledgerdescriptor / entry logger to rotate the entrylog for the ledger
```
sijie: It is unclear to me when you will send the write-close request.

site/bps/BP-32-advisory-write-close.md
```diff
@@ -0,0 +1,54 @@
...
+
+message WriteCloseResponse {
+    required StatusCode status = 1;
+    required int64 ledgerId = 2;  
+}
+
+### Compatibility, Deprecation, and Migration Plan
+
+- With this feature we are introducing new protocol message. Will do the required Compatibility testing. But since it is going to be advisory (optimistic) in nature, failure in this request in anyway should not affect functioning in any way.
+- Also, possibly this explicit write close can be used in future for other purposes.
```
sijie: +1 for keeping explicit write-close in mind

site/bps/BP-32-advisory-write-close.md
reddycharan: having client config would help this. 

But in the case of *enforced* close, success/failure behavior should be defined 

1) will we fail the writehandle close call?
2) what should be the behavior if it is partial success (part of the ensemble succeed but not the rest)?
3) In case of no response should we retry? 
4) In case of writehandle close call failure, should we expect API user to retry the writeHandle close call untill it is succeeded?

site/bps/BP-32-advisory-write-close.md
reddycharan: sure..this is lower level implementation detail. This should be doable/extendable.

site/bps/BP-32-advisory-write-close.md
sijie: "enforced-close" behavior should be similar as "we writing CLOSED state to ledger metadata in zookeeper". so "enforced-close" should carry the failure characteristics as "writing CLOSED state to ledger metadata“.


in details:

1. we should fail the writehandle close call. it is similar as if you can't write metadata to zookeeper, we fail the close call.

2. "enforced close" should be a quorum behavior as "fencing"

3. in case of no response: "no response" means timeout, we should retry.

4. 

> In case of writehandle close call failure, should we expect API user to retry the writeHandle close call untill it is succeeded?

the behavior should be same as "if writehandle close call failed due to failed to write to zookeeper, API user is responsible to retry".

site/bps/BP-32-advisory-write-close.md
sijie: I see it. make sense to me.

sijie:
This PR is approved with 2 +1. Marked it as `accepted`.

PR #1375, opened 2018-04-28
WIP BP14 - forceLedger bookie side implementation

eolivelli:
This is the bookie side implementation of the force() API as defined in BP-14, without the wire protocol handler.
We will introduce a new RPC which allows the client to ensure that all data sent for a  given ledger is persisted durably to the bookie, this is useful for DEFERRED_SYNC writers. In this case we are simulating a regular sync'd write to the journal choosen for the ledger.


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -1141,6 +1145,40 @@ private void addEntryInternal(LedgerDescriptor handle, ByteBuf entry,
...
+            if (oldValue == null) {
+                // new handle, we should add the key to journal ensure we can rebuild
+                ByteBuffer bb = ByteBuffer.allocate(8 + 8 + 4 + masterKey.length);
+                bb.putLong(ledgerId);
+                bb.putLong(METAENTRY_ID_LEDGER_KEY);
+                bb.putInt(masterKey.length);
+                bb.put(masterKey);
+                bb.flip();
+
+                getJournal(ledgerId).logAddEntry(bb, false /* ackBeforeSync */, new NopWriteCallback(), null);
```
jvrao: Not sure how I missed, but 'sync' flag would have made standard here instead of ackBeforeSync.

If sync == true -> sync to disk before ack
if sync == false -> no need to sync before ack; 
May be we can change it when we get a chance.
@sijie ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -1141,6 +1145,40 @@ private void addEntryInternal(LedgerDescriptor handle, ByteBuf entry,
         getJournal(ledgerId).logAddEntry(entry, ackBeforeSync, cb, ctx);
     }
 
+    /**
+     * Force write on the journal assigned to the given ledger.
+     * It works like a regular addEntry with ackBeforeSync=false but without really
+     * writing to disk.
```
jvrao: NiT; Remove "but without really writing to disk" ; I think that is implicit with the flag set to false.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -1141,6 +1145,40 @@ private void addEntryInternal(LedgerDescriptor handle, ByteBuf entry,
...
+     */
+    private void forceLedgerInternal(LedgerDescriptor handle,
+                                     WriteCallback cb, Object ctx, byte[] masterKey)
+            throws IOException, BookieException {
+        long ledgerId = handle.getLedgerId();
+
+
+        if (masterKeyCache.get(ledgerId) == null) {
```
jvrao: Can we avoid code duplication here? by introducing another method with a sync flag?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -1199,6 +1237,40 @@ public ByteBuf getExplicitLac(long ledgerId) throws IOException, Bookie.NoLedger
         return lac;
     }
 
+    /**
+     * Forces (sync) data on a ledger on the Journal.
```
jvrao: Nit: Consider changing this to -> " force sync given 'ledgerId' entries on the journal to the disk.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -1072,22 +1083,23 @@ public void run() {
                 if (qe == null) { // no more queue entry
                     continue;
                 }
+                if (qe.entryId != Bookie.METAENTRY_ID_FORCE_LEDGER) {
```
jvrao: Who will be consuming METAENTRY_ID_FORCE_LEDGER??

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
eolivelli: Okay

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
eolivelli: I did not introduce the nee method in order not to change addEntry and focus on this change.
I will do if you prefer. No problem.
I thought it would be a follow refactor/enhancement

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
eolivelli: Okay

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
eolivelli: It is just a placeholder. No data needs to be written to journal.
forceLedger is using the same code path as addEntry but with this special id.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
eolivelli: If you fell strong I will change.

I think that currently the exceptional case is sync=false so having ackBeforeSync=true lets us identify more clearly this case.
From the client it is an opt in feature, and it makes sense that here we see the client is requiring a special treatment

eolivelli:
@jvrao I will address your comments by changing the patch as soon as I am back from vacation.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
sijie: you don't need to change addEntry, but you can abstract line 1156 to line 1173 as a common method? so this common method can be called in addEntry and forceLedgerInternal, rather than duplicating code.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -313,7 +313,9 @@ public void run() {
                 LOG.debug("Acknowledge Ledger: {}, Entry: {}", ledgerId, entryId);
             }
             journalCbQueueSize.dec();
-            journalAddEntryStats.registerSuccessfulEvent(MathUtils.elapsedNanos(enqueueTime), TimeUnit.NANOSECONDS);
+            if (journalAddEntryStats != null) {
```
sijie: when this can be null?

if this can be null, that is a bug. the whole purpose of having NullStatsLogger is to avoid this null checking.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerStorage.java
```diff
@@ -103,12 +104,19 @@ void initialize(ServerConfiguration conf,
...
+     * Read an entry from storage. In case of {@link BookieProtocol#LAST_ADD_CONFIRMED} we are going to return the
+     * 'last' written entry.
+     *
+     * @throws Bookie.NoEntryException in case of missing entry
+     * @throws IOException in case of internal failure
+     *
+     * @return the entry, never null
      */
-    ByteBuf getEntry(long ledgerId, long entryId) throws IOException;
+    ByteBuf getEntry(long ledgerId, long entryId) throws IOException, Bookie.NoEntryException;
```
sijie: what is the purpose of the changes in this file?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
eolivelli: Wil do

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
eolivelli: I am setting it to null because forceLedger is not addEntry can add another logger if you prefer

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerStorage.java
eolivelli: Oh, I left the comment on the previous PR.
It is only documentation.
While digging into the internals I found it useful to add this docs in order not to need to run thru the implementations again

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
eolivelli: introduced new logger (JOURNAL_FORCE_LEDGER)

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerStorage.java
eolivelli: reverted

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
eolivelli: I prefer leaving ackBeforeSync in the scope of this patch, in order to keep it simple. We can do a follow up refactor

eolivelli:
@jvrao @sijie I have addressed all of your comments and answered to your questions.

plase take another look, hopefully the patch is ready to go

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -1201,6 +1227,40 @@ public ByteBuf getExplicitLac(long ledgerId) throws IOException, Bookie.NoLedger
...
+    public void forceLedger(long ledgerId, WriteCallback cb,
+                            Object ctx, byte[] masterKey)
+                throws IOException, BookieException {
+        long requestNanos = MathUtils.nowInNano();
+        boolean success = false;
+        try {
+            LedgerDescriptor handle = handles.getHandle(ledgerId, masterKey);
+            synchronized (handle) {
+                if (handle.isFenced()) {
+                    throw BookieException
```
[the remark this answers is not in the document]
eolivelli: Rethinking about fence/force I think you are right. While issuing a force() data is already on the bookie and so it is useless to break the writer. 
I thought it was a good idea because LAC on client advances only on force.
I am dropping this check

eolivelli:
@sijie ready again.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -1201,6 +1226,36 @@ public ByteBuf getExplicitLac(long ledgerId) throws IOException, Bookie.NoLedger
...
+     * for which the client receives acknoledgement for writes
+     * without the guarantee that data has been sync'd to disk (ackBeforeSync=true).
+     */
+    public void forceLedger(long ledgerId, WriteCallback cb,
+                            Object ctx, byte[] masterKey)
+                throws IOException, BookieException {
+        long requestNanos = MathUtils.nowInNano();
+        boolean success = false;
+        try {
+            LedgerDescriptor handle = handles.getHandle(ledgerId, masterKey);
```
sijie: why do you need get handle here?

eolivelli: Two (weak) reasons:
- validate masterkey
- execute inside the synchronized block like addEntry code path.

I have dropped the test about isFenced so the synchronized block maybe is useless.
All we are doing is to write on journal.

Code inside ensureLedgerOnMasterKeyCache should be treated as a per-ledger critical section ad wr are writing the meta entry the first time are seeing the ledger on the bookie.

Overall I think it makes sense and make forceLedger very similar to addEntry, but if you think it is an error or it could be a bottleneck I can drop it

eolivelli:
@jvrao please check again

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
eolivelli: As discussed in slack I will drop it.
There is no need to touch ledger storage for force Ledger.
I will update the patch soon

eolivelli:
@sijie  I have dropped access to LedgerStorage.
I am changing the stats from StatsLogger to Counter, because actually there is no value in registering elapsed time for only a write on journal internal queue.

please validate this change on stats again. thanks

eolivelli:
@jvrao  @sijie  I wlil take this branch as base for the next change.

I would like to merge as soon as possible in order not to have too much in-flight changes as I did the first time I started working on BP14

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -1201,6 +1209,22 @@ public ByteBuf getExplicitLac(long ledgerId) throws IOException, Bookie.NoLedger
...
+     * Force sync given 'ledgerId' entries on the journal to the disk.
+     * It works like a regular addEntry with ackBeforeSync=false.
+     * This is useful for ledgers with DEFERRED_SYNC write flag.
+     */
+    public void forceLedger(long ledgerId, WriteCallback cb,
+                            Object ctx, byte[] masterKey) {
+        if (LOG.isTraceEnabled()) {
+            LOG.trace("Forcing ledger {}", ledgerId);
+        }
+        ensureLedgerOnMasterKeyCache(ledgerId, masterKey);
```
sijie: I thought we agreed that we are not touching ledger storage at this moment at the slack channel. The method would be just as simple as `getJournal(ledgerId).forceLedger(ledgerId, cb, ctx)`, no?

Image this case, if there is no entries added in the ledger, the user calls #force(), do you think we need to create a ledger in the ledger storage?

eolivelli:
@jvrao  @sijie do I have green light for merging this one ?

I have dropped all the (direct/indirect) accesses to LedgerStorage and dropped masterKey from the wire protocol.

Next step is to review/merge https://github.com/apache/bookkeeper/pull/1393 

and the first client side patch is ready for review https://github.com/apache/bookkeeper/pull/853

PR #1382, opened 2018-05-03
ZKLedgerManager.writeLedgerMetadata should return appropriate error.

reddycharan:

ZKLedgerManager.writeLedgerMetadata should return
NoSuchLedgerExistsException error in the case it
gets KeeperException.Code.NONODE rc value from ZK.
ZKLedgerManager.removeLedgerMetadata does that as well.


sijie:
retest this please

PR #1352, opened 2018-04-17
[release-4.7.0] update BC tests

sijie:
Descriptions of the changes in this PR:

add 4.6.2 and 4.7.0 to BC tests list

PR #1386, opened 2018-05-03
Fix TestCompatUpgrade

sijie:

*Motivation*

 #1352 update the BC tests to include newer versions, however it removes the flags that used for 4.6.0 to handle badly shutdown.
 This change makes the BC tests become very flaky.

*Solution*

- apply `badlyshutdown` flag to both from 4.6.0 to 4.6.1 and from 4.6.1 to 4.6.2 upgrade


sijie:
This change only touches IT CI and IT CI passed.

sijie:
IGNORE CI

Issue #1387, opened 2018-05-03
LedgerHandleAdv incorrectly handles attempts to use LedgerHandle's specific API calls (add without specifying entry id)


dlg99:
**BUG REPORT**

1. Please describe the issue you observed:

- What did you do?

```java
lh = bkc.createLedgerAdv(ledgerId, 3, 3, 2, digestType, ledgerPassword, null);

lh.addEntry(entry.array());
```

- What did you expect to see?

Exception

- What did you see instead?

addEntry() hung

I'll submit PR soon.

dlg99:
My bad: mix up between community and internal (older) code. This is already fixed in community code, will add unit test anyway.

PR #1388, opened 2018-05-03
ISSUE #1387: added unit test for LedgerHandleAdv handling of attempts to use LedgerHandle's specific API calls (add without specifying entry id)

dlg99:

- added unit test.

Master Issue: #1387


dlg99:
@jvrao actually, community code does not need that. it is handled differently. 
will keep unit test.

PR #1391, opened 2018-05-04
Issue #570: EntryLogManagerForEntryLogPerLedger implementation

reddycharan:

This is < sub-task6  > of Issue #570

introducing EntryLogManagerForEntryLogPerLedger, which is the
implementation of EntryLogManager for entrylog per ledger
feature.

Master Issue: #570

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogManagerForEntryLogPerLedger.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogManagerForEntryLogPerLedger.java
@@ -1,0 +1,463 @@
+/**
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+
+package org.apache.bookkeeper.bookie;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
+import com.google.common.cache.RemovalListener;
+import com.google.common.cache.RemovalNotification;
+import io.netty.buffer.ByteBuf;
+import java.io.File;
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Optional;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.bookkeeper.bookie.EntryLogger.BufferedLogChannel;
+import org.apache.bookkeeper.bookie.LedgerDirsManager.LedgerDirsListener;
+import org.apache.bookkeeper.conf.ServerConfiguration;
+import org.apache.commons.lang.mutable.MutableInt;
+
+@Slf4j
+class EntryLogManagerForEntryLogPerLedger extends EntryLogManagerBase {
+
+    static class EntryLogAndLockTuple {
+        private final Lock ledgerLock;
+        private BufferedLogChannel entryLog;
+
+        public EntryLogAndLockTuple() {
+            ledgerLock = new ReentrantLock();
+        }
+
+        public Lock getLedgerLock() {
+            return ledgerLock;
+        }
+
+        public BufferedLogChannel getEntryLog() {
+            return entryLog;
+        }
+
+        public void setEntryLog(BufferedLogChannel entryLog) {
+            this.entryLog = entryLog;
+        }
+    }
+
+    private LoadingCache<Long, EntryLogAndLockTuple> ledgerIdEntryLogMap;
+    /*
+     * every time active logChannel is accessed from ledgerIdEntryLogMap
+     * cache, the accesstime of that entry is updated. But for certain
+     * operations we dont want to impact accessTime of the entries (like
+     * periodic flush of current active logChannels), and those operations
+     * can use this copy of references.
+     */
+    private final ConcurrentHashMap<Long, BufferedLogChannel> replicaOfCurrentLogChannels;
+    private final CacheLoader<Long, EntryLogAndLockTuple> entryLogAndLockTupleCacheLoader;
+    private EntryLogger.RecentEntryLogsStatus recentlyCreatedEntryLogsStatus;
+    private final int entrylogMapAccessExpiryTimeInSeconds;
+    private final int maximumNumberOfActiveEntryLogs;
+
+    EntryLogManagerForEntryLogPerLedger(ServerConfiguration conf, LedgerDirsManager ledgerDirsManager,
+            EntryLoggerAllocator entryLoggerAllocator, List<EntryLogger.EntryLogListener> listeners,
+            EntryLogger.RecentEntryLogsStatus recentlyCreatedEntryLogsStatus) throws IOException {
+        super(conf, ledgerDirsManager, entryLoggerAllocator, listeners);
+        this.recentlyCreatedEntryLogsStatus = recentlyCreatedEntryLogsStatus;
+        this.rotatedLogChannels = new CopyOnWriteArrayList<BufferedLogChannel>();
+        this.replicaOfCurrentLogChannels = new ConcurrentHashMap<Long, BufferedLogChannel>();
+        this.entrylogMapAccessExpiryTimeInSeconds = conf.getEntrylogMapAccessExpiryTimeInSeconds();
+        this.maximumNumberOfActiveEntryLogs = conf.getMaximumNumberOfActiveEntryLogs();
+        ledgerDirsManager.addLedgerDirsListener(getLedgerDirsListener());
+        this.entryLogAndLockTupleCacheLoader = new CacheLoader<Long, EntryLogAndLockTuple>() {
+            @Override
+            public EntryLogAndLockTuple load(Long key) throws Exception {
+                return new EntryLogAndLockTuple();
+            }
+        };
+        /*
+         * Currently we are relying on access time based eviction policy for
+         * removal of EntryLogAndLockTuple, so if the EntryLogAndLockTuple of
+         * the ledger is not accessed in
+         * entrylogMapAccessExpiryTimeInSeconds period, it will be removed
+         * from the cache.
+         *
+         * We are going to introduce explicit advisory writeClose call, with
+         * that explicit call EntryLogAndLockTuple of the ledger will be
+         * removed from the cache. But still timebased eviciton policy is
+         * needed because it is not guaranteed that Bookie/EntryLogger would
+         * receive successfully write close call in all the cases.
+         */
+        ledgerIdEntryLogMap = CacheBuilder.newBuilder()
+                .expireAfterAccess(entrylogMapAccessExpiryTimeInSeconds, TimeUnit.SECONDS)
+                .maximumSize(maximumNumberOfActiveEntryLogs)
+                .removalListener(new RemovalListener<Long, EntryLogAndLockTuple>() {
+                    @Override
+                    public void onRemoval(
+                            RemovalNotification<Long, EntryLogAndLockTuple> expiredLedgerEntryLogMapEntry) {
+                        removalOnExpiry(expiredLedgerEntryLogMapEntry);
+                    }
+                }).build(entryLogAndLockTupleCacheLoader);
+    }
+
+    /*
+     * This method is called when access time of that ledger has elapsed
+     * entrylogMapAccessExpiryTimeInSeconds period and the entry for that
+     * ledger is removed from cache. Since the entrylog of this ledger is
+     * not active anymore it has to be removed from
+     * replicaOfCurrentLogChannels and added to rotatedLogChannels.
+     *
+     * Because of performance/optimizations concerns the cleanup maintenance
+     * operations wont happen automatically, for more info on eviction
+     * cleanup maintenance tasks -
+     * https://google.github.io/guava/releases/19.0/api/docs/com/google/
+     * common/cache/CacheBuilder.html
+     *
+     */
+    private void removalOnExpiry(RemovalNotification<Long, EntryLogAndLockTuple> expiredLedgerEntryLogMapEntry) {
+        Long ledgerId = expiredLedgerEntryLogMapEntry.getKey();
+        log.debug("LedgerId {} is not accessed for entrylogMapAccessExpiryTimeInSeconds"
+                + " period so it is being evicted from the cache map", ledgerId);
+        EntryLogAndLockTuple entryLogAndLockTuple = expiredLedgerEntryLogMapEntry.getValue();
+        Lock lock = entryLogAndLockTuple.ledgerLock;
+        BufferedLogChannel logChannel = entryLogAndLockTuple.entryLog;
+        lock.lock();
+        try {
+            replicaOfCurrentLogChannels.remove(logChannel.getLogId());
+            rotatedLogChannels.add(logChannel);
+        } finally {
+            lock.unlock();
+        }
+    }
+
+    private LedgerDirsListener getLedgerDirsListener() {
+        return new LedgerDirsListener() {
+            @Override
+            public void diskFull(File disk) {
+                Set<BufferedLogChannel> copyOfCurrentLogs = getCopyOfCurrentLogs();
+                for (BufferedLogChannel currentLog : copyOfCurrentLogs) {
+                    if (disk.equals(currentLog.getLogFile().getParentFile())) {
+                        currentLog.setLedgerDirFull(true);
+                    }
+                }
+            }
+
+            @Override
+            public void diskWritable(File disk) {
+                Set<BufferedLogChannel> copyOfCurrentLogs = getCopyOfCurrentLogs();
+                for (BufferedLogChannel currentLog : copyOfCurrentLogs) {
+                    if (disk.equals(currentLog.getLogFile().getParentFile())) {
+                        currentLog.setLedgerDirFull(false);
+                    }
+                }
+            }
+        };
+    }
+
+    public void acquireLock(Long ledgerId) throws IOException {
+        try {
+            ledgerIdEntryLogMap.get(ledgerId).getLedgerLock().lock();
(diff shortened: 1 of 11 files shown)


bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/CreateNewLogTest.java
```diff
@@ -142,6 +151,237 @@ public void testCreateNewLogWithNoWritableLedgerDirs() throws Exception {
...
+                expectedPreAllocatedLogID, entryLoggerAllocator.getPreallocatedLogId());
+
+        int numOfLedgers = 6;
+
+        for (long i = 0; i < numOfLedgers; i++) {
+            /* since we are starting creation of new ledgers, entrylogid will be ledgerid */
+            entryLogManager.createNewLog(i);
+        }
+
+        Thread.sleep(100);
```
dlg99: why do you need sleep here?
if createNewLog() is async please consider something more predictable/less error prone. i.e. return future you can wait on.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/CreateNewLogTest.java
```diff
@@ -142,6 +151,237 @@ public void testCreateNewLogWithNoWritableLedgerDirs() throws Exception {
...
+
+        // new entrylog creation should succeed, though there is no writable ledgerDir
+        entryLogManager.createNewLog(ledgerId);
+    }
+
+    /*
+     * In this testcase it is validated if the entryLog is created in the
+     * ledgerDir with least number of current active entrylogs
+     */
+    @Test(timeout = 60000)
```
dlg99: IIRC we don't need "(timeout = 60000)" anymore

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/CreateNewLogTest.java
```diff
@@ -268,4 +508,64 @@ public void testCreateNewLogAndCompactionLog() throws Exception {
...
+        AtomicInteger createdEntryLogs = new AtomicInteger(0);
+        CountDownLatch startLatch = new CountDownLatch(1);
+        CountDownLatch createdLatch = new CountDownLatch(numOfLedgers * numOfThreadsForSameLedger);
+
+        for (long i = 0; i < numOfLedgers; i++) {
+            for (int j = 0; j < numOfThreadsForSameLedger; j++) {
+                long ledgerId = i;
+                new Thread() {
+                    @Override
+                    public void run() {
```
dlg99: you can safely use less verbose lambdas here, () -> {...}

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogManagerForEntryLogPerLedger.java
```diff
@@ -0,0 +1,463 @@
...
+             * of the file also should be force written.
+             */
+            flushLogChannel(logChannel, true);
+        }
+    }
+
+    @Override
+    public BufferedLogChannel createNewLogForCompaction() throws IOException {
+        throw new UnsupportedOperationException(
+                "When entryLogPerLedger is enabled, transactional compaction should have been disabled");
```
dlg99: is this checked in configuration as well? 
i'd prefer crash on start with message about wrong configuration to crash after few hrs of running.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogManagerForEntryLogPerLedger.java
```diff
@@ -0,0 +1,463 @@
...
+    void doEntryLogMapCleanup() {
+        ledgerIdEntryLogMap.cleanUp();
+    }
+
+    @VisibleForTesting
+    ConcurrentMap<Long, EntryLogAndLockTuple> getCacheAsMap() {
+        return ledgerIdEntryLogMap.asMap();
+    }
+    /*
+     * Returns writable ledger dir with least number of current active
```
eolivelli: This way if you lose a disk you will partially lose a ledger.
I wonder if it would be better to put all data of a ledger in the same directory.

In the future maybe we can think about selecting directory with an algorithm which takes into account some writeflag/metadata. Thinking about multi tenancy with a limited number of bookies

Certainly if you have few ledgers and many directory with this implementation you can use all of the disk space.



@jvrao

reddycharan:
retest this please

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogManagerForEntryLogPerLedger.java
```diff
@@ -0,0 +1,463 @@
...
+    /*
+     * every time active logChannel is accessed from ledgerIdEntryLogMap
+     * cache, the accesstime of that entry is updated. But for certain
+     * operations we dont want to impact accessTime of the entries (like
+     * periodic flush of current active logChannels), and those operations
+     * can use this copy of references.
+     */
+    private final ConcurrentHashMap<Long, BufferedLogChannel> replicaOfCurrentLogChannels;
+    private final CacheLoader<Long, EntryLogAndLockTuple> entryLogAndLockTupleCacheLoader;
+    private EntryLogger.RecentEntryLogsStatus recentlyCreatedEntryLogsStatus;
```
sijie: nit: it can be final, no?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogManagerForEntryLogPerLedger.java
```diff
@@ -0,0 +1,463 @@
...
+        public BufferedLogChannel getEntryLog() {
+            return entryLog;
+        }
+
+        public void setEntryLog(BufferedLogChannel entryLog) {
+            this.entryLog = entryLog;
+        }
+    }
+
+    private LoadingCache<Long, EntryLogAndLockTuple> ledgerIdEntryLogMap;
```
sijie: nit: final?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogManagerForEntryLogPerLedger.java
```diff
@@ -0,0 +1,463 @@
...
+     * cleanup maintenance tasks -
+     * https://google.github.io/guava/releases/19.0/api/docs/com/google/
+     * common/cache/CacheBuilder.html
+     *
+     */
+    private void removalOnExpiry(RemovalNotification<Long, EntryLogAndLockTuple> expiredLedgerEntryLogMapEntry) {
+        Long ledgerId = expiredLedgerEntryLogMapEntry.getKey();
+        log.debug("LedgerId {} is not accessed for entrylogMapAccessExpiryTimeInSeconds"
+                + " period so it is being evicted from the cache map", ledgerId);
+        EntryLogAndLockTuple entryLogAndLockTuple = expiredLedgerEntryLogMapEntry.getValue();
```
sijie:
```
A notification of the removal of a single entry. The key and/or value may be null if they were already garbage collected.
```

http://google.github.io/guava/releases/snapshot/api/docs/com/google/common/cache/RemovalNotification.html


It is better to check null for the value.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogManagerForEntryLogPerLedger.java
```diff
@@ -0,0 +1,463 @@
...
+        Lock lock = entryLogAndLockTuple.ledgerLock;
+        BufferedLogChannel logChannel = entryLogAndLockTuple.entryLog;
+        lock.lock();
+        try {
+            replicaOfCurrentLogChannels.remove(logChannel.getLogId());
```
sijie: just one question:

is the eviction behavior same, no matter of the `RemovalCause`?

http://google.github.io/guava/releases/snapshot/api/docs/com/google/common/cache/RemovalCause.html

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -85,6 +83,7 @@
         private final EntryLogMetadata entryLogMetadata;
         private final File logFile;
         private long ledgerIdAssigned = UNASSIGNED_LEDGERID;
+        volatile boolean ledgerDirFull = false;
```
sijie: I am not sure `ledgerDirFull` really belongs to `BufferedLogChannel`. `BufferedLogChannel` is a channel for a log file, which it shouldn't deal with `ledgerDirFull` or not. Can't you put the ledger dir full logic in entry log manager itself?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogManagerForEntryLogPerLedger.java
```diff
@@ -0,0 +1,463 @@
...
+            BufferedLogChannel hasToRotateLogChannel = getCurrentLogForLedger(ledgerId);
+            logChannel.setLedgerIdAssigned(ledgerId);
+            ledgerIdEntryLogMap.get(ledgerId).setEntryLog(logChannel);
+            replicaOfCurrentLogChannels.put(logChannel.getLogId(), logChannel);
+            if (hasToRotateLogChannel != null) {
+                replicaOfCurrentLogChannels.remove(hasToRotateLogChannel.getLogId());
+                rotatedLogChannels.add(hasToRotateLogChannel);
+            }
+        } catch (Exception e) {
+            log.error("Received unexpected exception while fetching entry from map", e);
```
sijie: nit: better to add `ledgerId` into the log statement, so it is easier for troubleshooting.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogManagerForEntryLogPerLedger.java
```diff
@@ -0,0 +1,463 @@
...
+            throw new IOException("Received unexpected exception while fetching entry from map", e);
+        } finally {
+            releaseLock(ledgerId);
```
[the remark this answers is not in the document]
sijie: +1 on what @dlg99 pointed out

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogManagerForEntryLogPerLedger.java
```diff
@@ -0,0 +1,463 @@
...
+    }
+
+    @Override
+    public BufferedLogChannel getCurrentLogForLedger(long ledgerId) throws IOException {
+        acquireLock(ledgerId);
+        try {
+            EntryLogAndLockTuple entryLogAndLockTuple = ledgerIdEntryLogMap.get(ledgerId);
+            return entryLogAndLockTuple.getEntryLog();
+        } catch (Exception e) {
+            log.error("Received unexpected exception while fetching entry from map", e);
```
sijie: nit: better to add ledgerId into the log statement, so it is easier for troubleshooting.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogManagerForEntryLogPerLedger.java
```diff
@@ -0,0 +1,463 @@
...
+            throw new IOException("Received unexpected exception while fetching entry from map", e);
+        } finally {
+            releaseLock(ledgerId);
+        }
+    }
+
+    public Set<BufferedLogChannel> getCopyOfCurrentLogs() {
+        return new HashSet<BufferedLogChannel>(replicaOfCurrentLogChannels.values());
```
sijie: nit: use ImmutableSet

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogManagerForEntryLogPerLedger.java
```diff
@@ -0,0 +1,463 @@
...
+
+    /*
+     * sets the logChannel for the given ledgerId. It will add the new
+     * logchannel to replicaOfCurrentLogChannels, and the previous one will
+     * be removed from replicaOfCurrentLogChannels. Previous logChannel will
+     * be added to rotatedLogChannels in both the cases.
+     */
+    @Override
+    public void setCurrentLogForLedgerAndAddToRotate(long ledgerId, BufferedLogChannel logChannel) throws IOException {
+        acquireLock(ledgerId);
```
sijie: what does the "lock" actually lock? does it prevent modifying the shared structure or the log channel itself? it is unclear to me. that says do we need to put the lock block for all the lines below?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogManagerForEntryLogPerLedger.java
sijie: +1

reddycharan: yes, it is validated in ServerConfiguration.validate method - https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java#L2268

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogManagerForEntryLogPerLedger.java
reddycharan: yes, it can be

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogManagerForEntryLogPerLedger.java
reddycharan: yes

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/CreateNewLogTest.java
reddycharan: yes, will remove

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogManagerForEntryLogPerLedger.java
reddycharan: valid point

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogManagerForEntryLogPerLedger.java
reddycharan: will do

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogManagerForEntryLogPerLedger.java
reddycharan: i followed simple rule, wherever in EntryLogManagerForSingleEntryLog, we are having code in 'synchronized' scope, i'm keeping the equivalent code of that in EntryLogManagerForEntryLogPerLedger, in guarded scope of ledgerId's lock.

so effectively ledgerId's lock is guarding everything related to this ledger - log channel and state information related to this ledger.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogManagerForEntryLogPerLedger.java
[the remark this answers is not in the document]
reddycharan: changed the logic of releaselock in the last minute, because of some issue in getIfPresent. So now it doesn't makes complete sense to do what I'm doing.

Will make changes as Andrey pointed out.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogManagerForEntryLogPerLedger.java
reddycharan: I'm not sure if it is correct to expect to pick same ledgerDir for a ledger. If a ledgerDir is full then it doesn't makes sense to pick the same ledgerdir when there is other available ledgerdirs.

This is not any new behavior I'm introducing here, it is the existing behavior. Here I'm making sure I'm picking ledgerDir with least number of current active logs, so no ledgerdir is overloaded.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
reddycharan: Hey Sijie, I considered about your concern before reintroducing this flag variable here. It would unnecessarily complicate the logic of LedgerDirsListener for maintaining the ledgerdir full status in the entrylogmanager implementation. Thats why I chose to keep it here. 

I don't think it is inappropriate to have it here, since there is final 'logfile' (File) variable for this BufferedLogChannel here, so I would consider 'ledgerDirFull' variable as extension of 'logFile' variable.

sijie: well if this logic is only to EntryLogPerLedgerManager, it should be maintained in EntryLogPerLedgerManager rather than polluting here. You can wrap BufferedLogChannel with a flag in EntryLogPerLedgerManager if you think you need that flag.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogManagerForEntryLogPerLedger.java
reddycharan: yes, in the existing implementation entry would be removed only in the following cases EXPIRED and SIZE. In these scenarios we would want the eviction behavior. COLLECTED is not relevant, since we are not using WeakKeys/WeakValues/SoftValues. When explicit writeclose is implemented we would do EXPLICIT removal by calling invalidate method explicitly, and even in that case we would need same behavior. In our implementation we would never get into REPLACED scenario, because we are not calling 'put' or 'putAll' method anywhere, value of the entry (EntryLogAndLockTuple) is CacheLoaded by the CacheLoader.

Being said that, I should rename this method and remove word 'Expiry' from name and update the method description. Thanks for pointing that out.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
reddycharan: Considered that as well, having new class BufferedLogChannelWithDirInfo extending BufferedLogChannelWithDirInfo in EntryLogManagerForEntryLogPerLedger class. But BufferedLogChannel instance is created by EntryLoggerAllocator https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLoggerAllocator.java#L162 , so if I create new BufferedLogChannelWithDirInfo class, then here I should have if/else block based on conf value.

is that ok?

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/CreateNewLogTest.java
reddycharan: guava to the rescue! removed these sleeps, for entryloggerallocator's allocatorexecutor using newDirectExecutorService()

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/CreateNewLogTest.java
reddycharan: changed it to lambda

sijie:
retest this please

reddycharan:
Tests passed!

sijie:
merged. @reddycharan thank you for your great contribution!

PR #1393, opened 2018-05-04
BP-14 forceLedger wire protocol server side implementation

eolivelli:
Implementation of the wire protocol for the BP-14 force ledger API on bookie side


eolivelli:
Note for reviewer:
please take into account only commit https://github.com/apache/bookkeeper/pull/1393/commits/9dcefd6284dd6b2304b6abfdc6598e90d1bb71a9

this branch is based on #1375

eolivelli:
Checkstyle is not passing. I will fix soon.

eolivelli:
@sijie @jvrao checkstyle fixed, this patch is ready for review.
I am working to client side stuff now

sijie:
@eolivelli can you rebase this PR?

eolivelli:
@sijie I saw your commit, I am just rebasing right now :-) 
 thank you

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
```diff
@@ -369,6 +378,40 @@ private void processAddRequestV3(final BookkeeperProtocol.Request r, final Chann
         }
     }
 
+    private void processForceLedgerRequestV3(final BookkeeperProtocol.Request r, final Channel c) {
+        ForceLedgerProcessorV3 forceLedger = new ForceLedgerProcessorV3(r, c, this);
+
+        final OrderedExecutor threadPool;
+        if (RequestUtils.isHighPriority(r)) {
```
eolivelli: @sijie I don't know if it is worth using this threadpool for forceLedger, I copied this from addEntry, for consistency

eolivelli:
@sijie  @jvrao reabase to current master and ready for final review

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
sijie: I think it is okay to be consistent

bookkeeper-proto/src/main/proto/BookkeeperProtocol.proto
```diff
@@ -63,6 +63,7 @@ enum OperationType {
     READ_LAC = 7;
     GET_BOOKIE_INFO = 8;
     START_TLS = 9;
+    FORCE_LEDGER = 10;
```
jvrao: I thought we are using SYNC_LEDGER instead of FORCE. Force doesn't tell what it is forcing. I was feeling weird in the previous patch itself but did not click on what it was I was missing in the name. :) s/force/sync makes lot of sense and code readability IMO

eolivelli:
@jvrao
the name force superseded "sync" after a quite long discussion between JV,Sijie,Ivan and me and we agreed on that final name.

The motivation is that Java APIs use the 'force' verb
https://docs.oracle.com/javase/7/docs/api/java/nio/channels/FileChannel.html#force(boolean)
on linux we have *fsync*/*fdatasync*, I don't know the windows world but with a quick search I found *fflush*, so using the Java name is more neutral

I am open to discussion and decide a new name, from my point of view (coming from Java world and Linux world) every of the two names make sense.

So @sijie  and @ivankelly were mostly involved in that discussion I I think we need to hear their voices I you feel strong about changing this name.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ForceLedgerProcessorV3.java
```diff
@@ -0,0 +1,137 @@
...
+
+        if (!isVersionCompatible()) {
+            forceLedgerResponse.setStatus(StatusCode.EBADVERSION);
+            return forceLedgerResponse.build();
+        }
+
+        BookkeeperInternalCallbacks.WriteCallback wcb = new BookkeeperInternalCallbacks.WriteCallback() {
+            @Override
+            public void writeComplete(int rc, long ledgerId, long entryId,
+                                      BookieSocketAddress addr, Object ctx) {
```
jvrao: maybe assert that the entryId is the special entryId? Just a thought.

jvrao:
Changes look fine to me, but somehow I think syncLedger makes more sense than forceLedger. @sijie ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ForceLedgerProcessorV3.java
eolivelli: in this case we are using WriteCallback as a trick , there is no real meaning in that entryId.

I could add this kind check, but in that case which error should we report to the client ? Any suggestion ?

jvrao:
> force superseded "sync" after a quite long discussion

I am fine if it is decided with a nod from @ivankelly  and @sijie. But somehow it doesn't go well with me as my background is UNIX/Linux :)

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ForceLedgerProcessorV3.java
jvrao: I said ASSERT. i.e bring down the bookie as it is an indication something seriously went wrong. At least add a comment about entryId.

eolivelli: okay, will do

eolivelli: @jvrao done

eolivelli:
@jvrao I have updated the patch with the assertions, let's wait for Sijie's (and eventually Ivan) final comments before merging.

eolivelli:
retest this please

eolivelli:
retest this please

eolivelli:
retest this please

eolivelli:
I kicked tests again, the error seemed not related to the patch ("no enough native threads"...)

eolivelli:
@sijie  @jvrao since these "names" (forceLedger vs sync) are only on bare code, not on public client API and not on wire protocol, I would like to merge this patch now so that bookie side changes are totally in
this way I can move forwards with steps on client: we cannot merge stuff on client if server side stuff is not complete

I can send a follow up change to rename variables and methods if we feel strong about changing *force* to *sync*.

Do this sound reasonable to you ?

sijie:
@jvrao I think we are using "force" to follow the convention in file channel java api - https://docs.oracle.com/javase/7/docs/api/java/nio/channels/FileChannel.html#force(boolean)

@eolivelli I understand you want to move this forward. but we need to wait until other people's replies. If there are not replies at a certain time, feel free to move forward.

eolivelli:
@sijie I will stay tuned.

I apologize to be annoying on this topic.

eolivelli:
jenkins had problems on disk, re-kicking tests

eolivelli:
retest this please

eolivelli:
Recap:
@sijie reviewed and approved
@jvrao reviewed and approved, suggesting to change the name from sync to force, but without being -1

I will wait untill Monday and than move forward.
We can change the name with a follow up change, we are not touching raw wire protocol, it is only a matter of variables and method names not exposed to the public

Thank you all guys for helping with this important change

PR #1394, opened 2018-05-04
Enable Prometheus stats provider to use custom http container

eolivelli:
This change enable Prometheus stats provider to:
- use an user supplied registry
- do not start internal Jetty server

The context is to run the Bookie inside a process together with a servlet container (like Apache Tomcat) and use the existing web endpoint instead of opening another http (plain) port


eolivelli:
please cherry pick to 4.7 branch

sijie:
in general, it would be good to seek a solution to combine the http server that reports metrics and the http servers that serve admin endpoints. we might have to create a task for that.

sijie:
retest this please

Issue #1390, opened 2018-05-04
Ensemble change on Delayed Write Failure


jvrao:
1. Please describe the feature you are requesting.
BK ignores any write failures *after* meeting ack quorum and successfully responding back to the application.
We should make the ensemble change on delayed write failures so one bad bookie should not cause extended under replication of many entries.

PR #1395, opened 2018-05-05
ISSUE #1390 Ensemble change on delayed write error

jvrao:
Error on delayed writes are dropped if the addEntry
is in complete state (ack quorum satisfied).
This change records the delayed write failure and forces
ensemble change onthe next write. This saves from having
extended under replicated status on the ledger and also
avoids unnecessary build up at PCBC channel.

> ---
> quickly and easily:
>
>
>
> Otherwise:
> 
> 
> ---

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
@@ -118,7 +118,7 @@ public class LedgerHandle implements WriteHandle {
     ScheduledFuture<?> timeoutFuture = null;
 
     final long waitForWriteSetMs;
-    Map<Integer, BookieSocketAddress> delayedWriteFailedBookies = new HashMap<Integer, BookieSocketAddress>();
+    private Map<Integer, BookieSocketAddress> delayedWriteFailedBookies = new HashMap<Integer, BookieSocketAddress>();
 
     /**
      * Invalid entry id. This value is returned from methods which
@@ -1755,7 +1755,7 @@ public class LedgerHandle implements WriteHandle {
         return new EnsembleInfo(newEnsemble, failedBookies, replacedBookies);
     }
 
-    void handleDelayeWriteBookieFailure() {
+    void handleDelayedWriteBookieFailure() {
         int curBlockAddCompletions = blockAddCompletions.get();
         if (bk.getDisableEnsembleChangeFeature().isAvailable()) {
             if (LOG.isDebugEnabled()) {
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
@@ -39,6 +39,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.EnumSet;
 import java.util.Enumeration;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -1749,6 +1755,42 @@ EnsembleInfo replaceBookieInMetadata(final Map<Integer, BookieSocketAddress> fai
         return new EnsembleInfo(newEnsemble, failedBookies, replacedBookies);
     }
 
+    void handleDelayeWriteBookieFailure() {
```
eolivelli: Typo?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -151,6 +153,10 @@
         }
     }
 
+    public Map<Integer, BookieSocketAddress> getDelayedWriteFailedBookies() {
```
eolivelli: Why do we have to expose as 'public' ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -117,6 +118,7 @@
     ScheduledFuture<?> timeoutFuture = null;
 
     final long waitForWriteSetMs;
+    Map<Integer, BookieSocketAddress> delayedWriteFailedBookies = new HashMap<Integer, BookieSocketAddress>();
```
eolivelli: Can you please explain why this is not  a ConcurrentMap ?

jvrao: Because of or ordered safe executor, I thought a regular map is good enough.

jvrao:
> I wonder if we should add a config option in order to enable explicitly this change
@eolivelli  I thought about it - Felt that it can just follow the same config path we use to make ensemble change on the failed add entry path. This patch still respects delayedEnsemble change and disableEnsembleChange configurations.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
jvrao: taken care

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
eolivelli: I was thinking the same. Just wanted to double check.

jvrao:
> I also wonder if it is possible to use mockito based client side framework and save us from starting a real cluster in tests. (Not blocker for me)

Are you making a general comment about tests in this class? if so, maybe we can open an issue and deal with that separately?

eolivelli:
For me it is okay.
I would like to know @sijie 's opinion

eolivelli:
Regarding the test: I did not think that this is not a new class, so okay for keeping your test as it is.

Regarding the map: okay for making it a concurrent map, in this case the map will be very small and rarely used so using a ConcurrentHashMap will not have significant impact and it makes things simpler. It can be a simple hashmap btw

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -117,6 +118,7 @@
     ScheduledFuture<?> timeoutFuture = null;
 
     final long waitForWriteSetMs;
+    private Map<Integer, BookieSocketAddress> delayedWriteFailedBookies = new HashMap<Integer, BookieSocketAddress>();
```
sijie: nit: final?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
jvrao: took care of it.

sijie:
retest this please

jvrao:
retest this please

sijie:
retest this please

PR #1397, opened 2018-05-08
Bookies should be from different racks in a Writequorum.

reddycharan:

Ideally RackAwareEnsemblePlacementPolicy should try to select
bookies from different racks for a write quorum. So in a
WriteQuorum, bookies should be from WriteQuorum number of racks,
provided there are sufficient number of available racks and
bookies.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
@@ -79,6 +79,7 @@ public class RackawareEnsemblePlacementPolicyImpl extends TopologyAwareEnsembleP
     int maxWeightMultiple;
     private Map<BookieNode, WeightedObject> bookieInfoMap = new HashMap<BookieNode, WeightedObject>();
     private WeightedRandomSelection<BookieNode> weightedSelection;
+    private int minNumRacksPerWriteQuorum;
 
     public static final String REPP_DNS_RESOLVER_CLASS = "reppDnsResolverClass";
     public static final String REPP_RANDOM_READ_REORDERING = "ensembleRandomReadReordering";
@@ -233,6 +234,7 @@ public class RackawareEnsemblePlacementPolicyImpl extends TopologyAwareEnsembleP
                                                               int stabilizePeriodSeconds,
                                                               boolean isWeighted,
                                                               int maxWeightMultiple,
+                                                              int minNumRacksPerWriteQuorum,
                                                               StatsLogger statsLogger) {
         checkNotNull(statsLogger, "statsLogger should not be null, use NullStatsLogger instead.");
         this.statsLogger = statsLogger;
@@ -242,6 +244,7 @@ public class RackawareEnsemblePlacementPolicyImpl extends TopologyAwareEnsembleP
         this.stabilizePeriodSeconds = stabilizePeriodSeconds;
         this.dnsResolver = new DNSResolverDecorator(dnsResolver, () -> this.getDefaultRack());
         this.timer = timer;
+        this.minNumRacksPerWriteQuorum = minNumRacksPerWriteQuorum;
 
         // create the network topology
         if (stabilizePeriodSeconds > 0) {
@@ -329,6 +332,7 @@ public class RackawareEnsemblePlacementPolicyImpl extends TopologyAwareEnsembleP
                 conf.getNetworkTopologyStabilizePeriodSeconds(),
                 conf.getDiskWeightBasedPlacementEnabled(),
                 conf.getBookieMaxWeightMultipleForWeightBasedPlacement(),
+                conf.getMinNumRacksPerWriteQuorum(),
                 statsLogger);
     }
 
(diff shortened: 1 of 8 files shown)


bookkeeper-server/src/test/java/org/apache/bookkeeper/client/TestRackawareEnsemblePlacementPolicy.java
```diff
@@ -886,11 +973,29 @@ static int getNumCoveredWriteQuorums(ArrayList<BookieSocketAddress> ensemble, in
...
                 racks.add(StaticDNSResolver.getRack(addr.getHostName()));
             }
-            numCoveredWriteQuorums += (racks.size() > 1 ? 1 : 0);
+            numCoveredWriteQuorums += (racks.size() == Math.min(writeQuorumSize, Math.max(availableNumOfRacks, 2)) ? 1
+                    : 0);
         }
         return numCoveredWriteQuorums;
     }
 
+    static int getMinNumOfRacksForAWriteQuorum(ArrayList<BookieSocketAddress> ensemble, int writeQuorumSize)
```
reddycharan: have to remove this method

reddycharan:
retest this please

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
```diff
@@ -542,9 +544,58 @@ public void handleBookiesThatJoined(Set<BookieSocketAddress> joinedBookies) {
...
+                    if (writeQuorumSize > 1) {
+                        /*
+                         * Ideally RackAwareEnsemblePlacementPolicy should try
+                         * to select bookies from different racks for a write
+                         * quorum. So in a WriteQuorum, bookies should be from
+                         * WriteQuorum number of racks. So we would add racks of
+                         * (WriteQuorumSize-1) neighbours (both sides) to the
+                         * exclusion list (~curRack).
+                         */
+                        for (int j = 1; j < writeQuorumSize; j++) {
```
sijie: can we have a separate setting like `minNumRacksPerWriteQuorum` and by default set it to 2? so that it doesn't change the default behavior here.

because not everyone has enough rack diversity. especially when running bookkeeper in cloud, we only can use `availability zone` as `rack`, so the number of `availability zone`s per region is typically limited to 2~3.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/TestRackawareEnsemblePlacementPolicy.java
```diff
@@ -525,17 +526,89 @@ public void testNewEnsembleWithMultipleRacks() throws Exception {
...
+                    null, new HashSet<>());
+            numCovered = getNumCoveredWriteQuorums(ensemble2, 2, numOfAvailableRacks);
             assertTrue(numCovered >= 1 && numCovered < 3);
         } catch (BKNotEnoughBookiesException bnebe) {
             fail("Should not get not enough bookies exception even there is only one rack.");
         }
     }
 
+    @Test
+    public void testWriteQuorumNumberOfRacks() throws Exception {
```
sijie: can you add a unit test case to cover `write-quorum-size` is larger than `number-of-racks`?

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/TestRackawareEnsemblePlacementPolicy.java
```diff
@@ -525,17 +526,89 @@ public void testNewEnsembleWithMultipleRacks() throws Exception {
...
+        BookieSocketAddress addr1 = new BookieSocketAddress("127.0.0.2", 3181);
```
sijie: nit: can we put line 548-588 into a for-loop?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
reddycharan: @jvrao

reddycharan:
retest this please

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/TestRackawareEnsemblePlacementPolicy.java
reddycharan: added testscenario

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/TestRackawareEnsemblePlacementPolicy.java
reddycharan: moved to for loop

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/TestRackawareEnsemblePlacementPolicy.java
reddycharan: removed this method

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
reddycharan: changed logic to honor minNumRacksPerWriteQuorum config value.

reddycharan:
retest this please

reddycharan:
@sijie just FYI, introduced new config for min number of racks in a writeQuorum as you pointed out.

PR #1399, opened 2018-05-10
Issue #1396: Export metrics in same http server

### LINKED ISSUE #1396: Export metrics in same http server
**FEATURE REQUEST**

Currently metrics provider and http admin endpoint are isolated due to the way how things are modularized. This requires two ports to be used, one is for metrics provider to expose metrics, while the other one is used by http admin endpoint for expose admin endpoints.

It would be good to allow export metrics in same http admin endpoint if it is enabled.

2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have). Are you currently using any workarounds to address this issue?

*should-have*

3. Provide any additional detail on your proposed use case for this feature.

sijie:

*Motivation*

Currently metrics provider and http admin endpoint are isolated due to the way how things are modularized.
This requires two ports to be used, one is for metrics provider to expose metrics, while the other one is used by
http admin endpoint for expose admin endpoints.

When implementing a bookkeeper operator on k8s, it becomes confusing on managing those ports and do health checks.
It would be good to allow export metrics in same http admin endpoint if it is enabled.

*Solution*

- Introduce a method `writeAllMetrics` in stats provider to allow stats provider exporting metrics.
- Add a `MetricsService` in bookie http service and export stats provider's metrics under `/metrics` endpoint.

This fixes #1396

Master Issue: #1396


bookkeeper-server/src/main/java/org/apache/bookkeeper/server/http/service/MetricsService.java
```diff
@@ -0,0 +1,78 @@
...
+
+        if (null == statsProvider) {
+            response.setCode(StatusCode.INTERNAL_ERROR);
+            response.setBody("Stats provider is not enabled. Please enable it by set statsProviderClass"
+                + " on bookie configuration");
+            return response;
+        }
+
+        // GET
+        try (StringWriter writer = new StringWriter(1024)) {
```
eolivelli: 1024 ?

sijie: the initial size of the buffer

eolivelli: Sorry, my comment was not clear 
Why did you choose 1024? just curious

sijie: metrics is usually larger than 1KB. so choose 1KB as the initial buffer size.

bookkeeper-stats-providers/prometheus-metrics-provider/src/main/java/org/apache/bookkeeper/stats/prometheus/PrometheusMetricsProvider.java
```diff
@@ -174,7 +176,7 @@ public StatsLogger getStatsLogger(String scope) {
     }
 
     @VisibleForTesting
-    void writeAllMetrics(Writer writer) throws IOException {
+    public void writeAllMetrics(Writer writer) throws IOException {
```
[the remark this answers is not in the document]
sijie: fixed.

PR #1402, opened 2018-05-12
Fix TestSmoke#testTailingReads

sijie:

*Motivation*

TestSmoke#testTailingReads is testing the tailing behavior at bookkeeper. However the current testing logic is a bit
problematic on checking whether the reader has caught up with the expected last add confirmed. so the read logic break
earlier before reading the expectedLastConfirmedEntry.

*Solution*

Fix the checking logic on breaking the read loop.


sijie:
retest this please

sijie:
IT tests passed. 

Jenkins Java 8/9 failed due to filled disks. so ignore CI to merge.

sijie:
IGNORE CI

Issue #1403, opened 2018-05-12
ArrayIndexOutOfBoundsException is thrown on readLastAddConfirmedAndEntry


sijie:
**BUG REPORT**

1. Please describe the issue you observed:

- What did you do?

write an integration test on testing long poll reads

- What did you expect to see?

integration test can pass

- What did you see instead?

the integration test hangs on.

```
2018-05-12 03:41:11,910 - ERROR - [BookKeeperClientWorker-OrderedExecutor-0-0:SafeRunnable@38] - Unexpected throwable caught
java.lang.ArrayIndexOutOfBoundsException: 2
	at org.apache.bookkeeper.client.RoundRobinDistributionSchedule$WriteSetImpl.get(RoundRobinDistributionSchedule.java:123)
	at org.apache.bookkeeper.client.ReadLastConfirmedAndEntryOp$SequenceReadRequest.sendNextRead(ReadLastConfirmedAndEntryOp.java:373)
	at org.apache.bookkeeper.client.ReadLastConfirmedAndEntryOp$SequenceReadRequest.logErrorAndReattemptRead(ReadLastConfirmedAndEntryOp.java:406)
	at org.apache.bookkeeper.client.ReadLastConfirmedAndEntryOp.readEntryComplete(ReadLastConfirmedAndEntryOp.java:602)
	at org.apache.bookkeeper.proto.PerChannelBookieClient$ReadCompletion$1.readEntryComplete(PerChannelBookieClient.java:1633)
	at org.apache.bookkeeper.proto.PerChannelBookieClient$ReadCompletion.handleReadResponse(PerChannelBookieClient.java:1714)
	at org.apache.bookkeeper.proto.PerChannelBookieClient$ReadCompletion.handleV3Response(PerChannelBookieClient.java:1689)
	at org.apache.bookkeeper.proto.PerChannelBookieClient$3.safeRun(PerChannelBookieClient.java:1304)
	at org.apache.bookkeeper.common.util.SafeRunnable.run(SafeRunnable.java:36)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at io.netty.util.concurrent.DefaultThreadFactory$DefaultRunnableDecorator.run(DefaultThreadFactory.java:138)
	at java.lang.Thread.run(Thread.java:748)
```

[...]

PR #1401, opened 2018-05-12
 Provide a vagrantfile to launch a linux vm for running integration tests

sijie:
Descriptions of the changes in this PR:

*Motivation*

Currently all bookkeeper's integration tests and bc tests are docker based and written
using arquillian framework. Due to the network issues in docker and the way how bookies
advertise themselves in the cluster, we can *ONLY* run integration tests in a linux
environment. It is inconvinient to people who is using macOS as the development environment.

*Solution*

This PR is providing a Vagrantfile to lanuch a dev vm that contains all the tools required
for running integration tests and bc tests. So people can actually debug and test integration
tests locally using a linux vm.

PR #1404, opened 2018-05-12
Issue 1403: ArrayIndexOutOfBoundsException is thrown on readLastAddConfirmedAndEntry

sijie:

*Motivation*

There are two bugs in `ReadLastAddConfirmedAndEntry`:

1) a regression was introduced by #657. the long poll read op is attempting to long-poll reading lac. since lac is stored across ensemble, so the retry logic assumes it will attempt over all the bookies in the ensemble. however #657 use a `write-quorum-size` write set for tracking those attempts. this leads to ArrayIndexOutOfBoundsException reported at #1403. The integrate tests added in this PR can easily reproduce this issue.

2) there was a bug on retry logic when a ledger whose ensemble size is larger than write quorum size. when this happens, it will claim lac is not advanced prior to attempt the bookie in the ensemble. so the client will never know lac is advanced if using long poll reads on such ledgers. The integrate tests added in this PR can also catch this issue.

disclaim: twitter uses long poll reads but never uses `ensembleSize > writeQuorumSize`. so this is not a problem for dlog users.

*Solution*

- introduce a `getWriteSetForLongPoll` call that uses `ensembleSize` for building the write set. this would address problem 1)
- fix the assignment of `numEmptyResponsesAllowed`, so the long poll reads can work with `ensembleSize > writeQuorumSize`
- add integration tests for long polling reads

at the same time, also add an integration test for normal tailing reads with


sijie:
This PR is based on #1402 and #1401 so it includes changes from those two PRs as well.

eolivelli:
retest this please

Issue #1409, opened 2018-05-17
Server-side backpressure


dlg99:
**FEATURE REQUEST**

1. Please describe the feature you are requesting.

It is possible for client(s) to overload bookie with the requests.
Bookie should stop accepting requests if it cannot process them anymore.

Without backpressure, bookie:

Scenario 1:
Bookie receives add entry requests and pushes them to downstream processing. 
These requests end up accumulated in journal queue. 
Bookie runs out of memory, goes into super long GC or just drags on processing requests with extremely high latency trying to handle all queued requests.

Scenario 2:
Bookie receives more requests than it can handle, memtable flush is super slow, bookie tries to slow things down (sleep 1ms as a throttling attempt in memtable)
Netty keeps on autoreading requests from the network into internal buffers
bookie runs out of memory, goes into super long GC or just drags on processing requests with >> 1sec latency trying to handle all queued requests.

Scenario 3:
Bookie receives a lot of read requests.
Bookie manages to actually read data on time
Bookie cannot send all read data back to the client because of NIC being maxed out, client is not responding etc.
Bookie runs out of memory, goes into super long GC or just drags on processing requests with >> 1sec latency trying to handle all queued requests.
(yes, I noticed an option to shed read requests)

2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have). Are you currently using any workarounds to address this issue?

must have

sijie:
This is addressed by #1410

Issue #1367, opened 2018-04-27
Size-based throttling in Bookie side


ArvinDevel:
feature request

**FEATURE REQUEST**

1. Please describe the feature you are requesting.
Size based throttling in Bookie side

2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have). Are you currently using any workarounds to address this issue?
nice-to-have

3. Provide any additional detail on your proposed use case for this feature.
Current bookie use `maxPendingAddRequestsPerThread` to limit server request processor‘s writing capacity, which can't deal different request size gracefully, and we can't setting it according to server resource accurately. In my test case, I have to change this setting to avoid the `TooManyRequestsException`.

dlg99:
see https://github.com/apache/bookkeeper/issues/1409 / https://github.com/apache/bookkeeper/pull/1410

PR #1410, opened 2018-05-17
Issue #1409: Added server side backpressure (@bug W-3651831@)

dlg99:
Added server side backpressure handling and related unit tests.

**Background:**

BK’s writes happen in this order on the server side:
First, ledger storage (Interleaved|Sorted), presumably non-blocking to some level.
Second, journal. 
Request is finished when journal’s write is fsynced.

Three major moving parts on the server side need to be taken into account:
- Journal; its performance (I/O delays) or journal’s batching delay (mis)configuration affect client latency. Journal has internal batching and separate thread for data fsync/request ack.
- InterleavedLedgerStorage/entry log. It will naturally block request before it reaches journal if blocked on I/O.
- SortedLedgerStorage. Sorted storage puts request into in-memory data structure/SkipList (aka memtable) until it reaches certain limit and flushes it to disk asynchronously.   

**Implementation:**

1. Limit number of requests in progress (separately for reads and writes since they are handled in different thread pools).

Requests in progress (RIPs) are requests being processed by threads in threadpool + requests waiting for the next thread. RIPs lifetime if from the moment it is received/read by netty to the moment response for the request is sent. 

Target limit of RIPs (heuristics) is ((number of processing threads)*2+(expected max batch size on journal)), so each thread can have one request to process and the next one waiting. 
It is assumed that we have enough memory to keep data for that many requests. It is impossible to estimate size of read request at the moment when it is received anyway.

Limit is configured by setting number of RIPs explicitly in config for the following reasons:
Easier to experiment with different numbers. I.e. we may want to experiment with different number of requests in progress, i.e. ((number of processing threads)+2*(expected max batch size on journal)) or simply (2*(expected max batch size on journal)).
There is an option to run request directly on netty thread so no config parameter to base initial value on and netty’s defaults can change between versions. 
Removes need for explicit enable/disable backpressure flag, instead we can set RIPs to zero. 

2. Pause netty’s autoread when limit is exceeded to prevent it from pulling more data before we track it as RIP. 

3. Limit number of requests in asynchronous write path (LedgerStorage)

InterleavedLedgerStorage will naturally block if write is slowed down due to i.e. fsync. 
SortedLedgerStorage has naive implementation of throttling that blocks request for 1ms if checkpoint (memtable flush) is in progress. This is replaced with block until space in memtable is available. The limit is set to 2*(skipListSize) where skipListSize is limit that triggers memtable flush. 
 
 4. Sending response must respect netty’s isWritable() flag and wait up to certain timeout, if needed. 
Drop response after timeout (client will not hear about that request) or close the channel (disconnect will notify the client that responses to requests from that connection will never happen).

Master Issue: #1409


sijie:
@dlg99 overall looks awesome. will do a detailed review tomorrow.

@merlimat please review this as well.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -935,11 +940,27 @@ public void run() {
             while (true) {
                 // new journal file to write
                 if (null == logFile) {
+
                     logId = logId + 1;
+                    final JournalChannel.BufferedChannelBuilder bcBuilder;
+                    if (getDelay > 0 || addDelay > 0 || flushDelay > 0) {
```
eolivelli: it would be better to use PowerMock and to not use this kind of "explict" instrumentation.
Recently we added a lot of test cases which tweak the journal

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SortedLedgerStorage.java
```diff
@@ -41,16 +42,23 @@
  * entries will be first added into a {@code MemTable}, and then be flushed back to the
  * {@code InterleavedLedgerStorage} when the {@code MemTable} becomes full.
  */
-public class SortedLedgerStorage extends InterleavedLedgerStorage
-        implements LedgerStorage, CacheCallback, SkipListFlusher {
+public class SortedLedgerStorage
```
eolivelli: this is a important change, SortedLedgerStorage will not be a InterleavedLedgerStorage anymore

cc @sijie @ivankelly  @merlimat

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ClientConfiguration.java
```diff
@@ -128,7 +128,6 @@
     protected static final String PCBC_TIMEOUT_TIMER_NUM_TICKS = "pcbcTimeoutTimerNumTicks";
     protected static final String TIMEOUT_TIMER_TICK_DURATION_MS = "timeoutTimerTickDurationMs";
     protected static final String TIMEOUT_TIMER_NUM_TICKS = "timeoutTimerNumTicks";
-    protected static final String WAIT_TIMEOUT_ON_BACKPRESSURE = "waitTimeoutOnBackpressureMs";
```
eolivelli: I this it would be better to have two separate properties, one for server and another for client.
Having two properties with the same name will be confiusing and/or we have to write docs to explain this fact

dlg99:
retest this please

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SortedLedgerStorage.java
jvrao: What are the implications of this change? @eolivelli ?

jvrao: @dlg99 can you please describe the motto of this change?

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
```diff
@@ -665,6 +668,34 @@ public T setNettyMaxFrameSizeBytes(int maxSize) {
...
+    /**
+     * Timeout controlling wait on request send in case of unresponsive bookie(s)
+     * (i.e. bookie in long GC etc.)
+     *
+     * @return timeout value
+     *        negative value disables the feature
+     *        0 to allow request to fail immediately
+     *        Default is -1 (disabled)
+     */
+    public long getWaitTimeoutOnBackpressureMillis() {
```
jvrao: If it is from the client, how does the client know if the bookie is slow/dead/hung or something backpressure related? Also, how does this manifest with rest of the timeouts, say, read/write timeout

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
```diff
@@ -614,6 +620,73 @@ public ServerConfiguration setJournalFormatVersionToWrite(int version) {
         return this;
     }
 
+    /**
+     * Get max number of adds in progress. 0 == unlimited.
+     *
+     * @return Max number of adds in progress.
+     */
+    public int getMaxAddsInProgress() {
```
jvrao: This sounds like a stat rather than config. Maybe change getMaxAddsInProgressLimit

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
```diff
@@ -214,6 +266,122 @@ public BookieRequestProcessor(ServerConfiguration serverCfg, Bookie bookie,
         this.getBookieInfoStats = statsLogger.getOpStatsLogger(GET_BOOKIE_INFO);
         this.getBookieInfoRequestStats = statsLogger.getOpStatsLogger(GET_BOOKIE_INFO_REQUEST);
         this.channelWriteStats = statsLogger.getOpStatsLogger(CHANNEL_WRITE);
+
+        this.addEntryBlockedStats = statsLogger.getOpStatsLogger(ADD_ENTRY_BLOCKED_WAIT);
+        this.readEntryBlockedStats = statsLogger.getOpStatsLogger(READ_ENTRY_BLOCKED_WAIT);
+
+        int maxAdds = serverCfg.getMaxAddsInProgress();
```
jvrao: int maxAddsInProgressLimit?

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
```diff
@@ -214,6 +266,122 @@ public BookieRequestProcessor(ServerConfiguration serverCfg, Bookie bookie,
...
+        addsSemaphore = maxAdds > 0 ? new Semaphore(maxAdds, true) : null;
+
+        int maxReads = serverCfg.getMaxReadsInProgress();
```
jvrao: Same comment

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
```diff
@@ -214,6 +266,122 @@ public BookieRequestProcessor(ServerConfiguration serverCfg, Bookie bookie,
...
+
+    }
+
+    protected void onAddRequestStart(Channel channel) {
+        if (addsSemaphore != null) {
+            if (!addsSemaphore.tryAcquire()) {
+                final long throttlingStartTimeNanos = MathUtils.nowInNano();
+                channel.config().setAutoRead(false);
+                addsBlocked.incrementAndGet();
+                addsSemaphore.acquireUninterruptibly();
```
jvrao: Info/Debug message will be helpful

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
dlg99: will do.
PowerMock in tests that extends BookKeeperClusterTestCase a bit problematic, likely due to  
https://github.com/powermock/powermock/issues/822

Figured a way how to use just mockito there, will update PR soon.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SortedLedgerStorage.java
dlg99: This simplifies testability i.e. SlowSortedLedgerStorage implementation. Reduces code duplication etc. I tried dealing with this without this change first and turn up ugly.

I am not worried about this as there is no reason to code directly against InterleavedLedgerStorage instead of CompactableLedgerStorage interface which SortedLedgerStorage still implements. Tests still pass though I had to change couple of tests that casted SortedLedgerStorage to InterleavedLedgerStorage. 
If there is some specific reason that I have missed that explains why we should not change that or a reason why we should ignore interface and code against concrete implementation please let me know.

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
dlg99: will do

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
```diff
@@ -614,6 +620,73 @@ public ServerConfiguration setJournalFormatVersionToWrite(int version) {
...
+        this.setProperty(MAX_ADDS_IN_PROGRESS, value);
+        return this;
+    }
+
+    /**
+     * Get max number of reads in progress. 0 == unlimited.
+     *
+     * @return Max number of reads in progress.
+     */
+    public int getMaxReadsInProgress() {
```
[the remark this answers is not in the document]
dlg99: will do

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
```diff
@@ -1507,6 +1581,10 @@ public long getSkipListSizeLimit() {
      * @return server configuration object.
      */
     public ServerConfiguration setSkipListSizeLimit(int size) {
+        if (size > (Integer.MAX_VALUE - 1) / 2) {
+            // gives max of 2*1023MB for mem table (one being checkpointed and still writable).
+            throw new IllegalArgumentException("skiplist size over " + ((Integer.MAX_VALUE - 1) / 2));
+        }
```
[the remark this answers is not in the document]
dlg99: I'll leave it at explicit error (fail on start, fast detection) over the quiet value change.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
dlg99: will do

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
dlg99: will do

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
```diff
@@ -214,6 +266,122 @@ public BookieRequestProcessor(ServerConfiguration serverCfg, Bookie bookie,
...
+        }
+    }
+
+    protected void onReadRequestStart(Channel channel) {
+        if (readsSemaphore != null) {
+            if (!readsSemaphore.tryAcquire()) {
+                final long throttlingStartTimeNanos = MathUtils.nowInNano();
+                channel.config().setAutoRead(false);
+                readsBlocked.incrementAndGet();
+                readsSemaphore.acquireUninterruptibly();
```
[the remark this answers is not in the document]
dlg99: will add

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
dlg99: will add

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
dlg99: all the client is see is that connection to bookie still exists but data cannot be send via it at the moment (or cannot be send fast enough to clean up existing backlog) and netty raises channel.isWritable() == false.
the rest is handled by client backpressure.

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ClientConfiguration.java
dlg99: will do

dlg99:
retest this please

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
```diff
@@ -214,6 +266,126 @@ public BookieRequestProcessor(ServerConfiguration serverCfg, Bookie bookie,
...
+            public Number getSample() {
+                return readsBlocked;
+            }
+        });
+
+    }
+
+    protected void onAddRequestStart(Channel channel) {
+        if (addsSemaphore != null) {
+            if (!addsSemaphore.tryAcquire()) {
```
sijie: nit:

currently all the throttling is based requests. it might be good to do throttling based on bytes, because that is much easier to translate to bandwidth usage rather requests. apache/bookkeeper#1367 is one ask for that.

that can be a future improvement.

dlg99: @sijie This is partially addressed this in memtable flush scenario but I wanted to keep configuration consistent across read and write requests (as in: requests in progress limit) and in case of reads one does not know size of the entry until index is read. 
We can think about to byte-based configuration later but I feel like it is not going to be a high priority if backpressure is dynamic (like with this change) and reacts on actual throughput and not a rate-base throttling like client used to have.

dlg99:
@jvrao @eolivelli Guys, I did the changes to address your comments. Can you please take another look?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SortedLedgerStorage.java
eolivelli: @dlg99 thank you for trying.
Go head with your change.
We already told at community meeting that this is not a problem

eolivelli:
@dlg99 it seems that we are missing @jvrao's approval as he is in the review loop

PR #1414, opened 2018-05-18
Ensure BufferedChannel instance is properly closed

merlimat:
Whenever the the `EntryLogger` is closing the `BufferedChannel` instance (the channel used for writing) it is closing the file descriptor but it's not calling `close()` on the object itself. 

This causes a small mem leak for each log file, because the `writeBuffer` from `BufferedChannel` is never released.

I think this should be considered for a 4.7.1 release.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedChannel.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedChannel.java
@@ -86,8 +86,9 @@ public class BufferedChannel extends BufferedReadChannel implements Closeable {
     }
 
     @Override
-    public void close() throws IOException {
+    public synchronized void close() throws IOException {
         writeBuffer.release();
+        fileChannel.close();
     }
 
     /**
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedChannel.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedChannel.java
@@ -23,6 +23,7 @@ package org.apache.bookkeeper.bookie;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufAllocator;
+import io.netty.util.ReferenceCountUtil;
 
 import java.io.Closeable;
 import java.io.IOException;
@@ -39,7 +40,7 @@ public class BufferedChannel extends BufferedReadChannel implements Closeable {
     // The position of the file channel's write pointer.
     protected AtomicLong writeBufferStartPosition = new AtomicLong(0);
     // The buffer used to write operations.
-    protected final ByteBuf writeBuffer;
+    protected ByteBuf writeBuffer;
     // The absolute position of the next write operation.
     protected final AtomicLong position;
 
@@ -87,7 +88,11 @@ public class BufferedChannel extends BufferedReadChannel implements Closeable {
 
     @Override
     public synchronized void close() throws IOException {
-        writeBuffer.release();
+        if (writeBuffer != null) {
+            ReferenceCountUtil.safeRelease(writeBuffer);
+            writeBuffer = null;
+        }
+
         fileChannel.close();
     }
 
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedChannel.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedChannel.java
@@ -40,7 +40,7 @@ public class BufferedChannel extends BufferedReadChannel implements Closeable {
     // The position of the file channel's write pointer.
     protected AtomicLong writeBufferStartPosition = new AtomicLong(0);
     // The buffer used to write operations.
-    protected ByteBuf writeBuffer;
+    protected final ByteBuf writeBuffer;
     // The absolute position of the next write operation.
     protected final AtomicLong position;
 
@@ -88,11 +88,7 @@ public class BufferedChannel extends BufferedReadChannel implements Closeable {
 
     @Override
     public synchronized void close() throws IOException {
-        if (writeBuffer != null) {
-            ReferenceCountUtil.safeRelease(writeBuffer);
-            writeBuffer = null;
-        }
-
+        ReferenceCountUtil.safeRelease(writeBuffer);
         fileChannel.close();
     }
 
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogManagerForEntryLogPerLedger.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogManagerForEntryLogPerLedger.java
@@ -61,6 +61,7 @@ import org.apache.bookkeeper.conf.ServerConfiguration;
 import org.apache.bookkeeper.stats.Counter;
 import org.apache.bookkeeper.stats.OpStatsLogger;
 import org.apache.bookkeeper.stats.StatsLogger;
+import org.apache.bookkeeper.util.IOUtils;
 import org.apache.bookkeeper.util.collections.ConcurrentLongHashMap;
 import org.apache.commons.lang3.mutable.MutableInt;
 
@@ -554,7 +555,7 @@ class EntryLogManagerForEntryLogPerLedger extends EntryLogManagerBase {
     public void close() throws IOException {
         Set<BufferedLogChannelWithDirInfo> copyOfCurrentLogsWithDirInfo = getCopyOfCurrentLogs();
         for (BufferedLogChannelWithDirInfo currentLogWithDirInfo : copyOfCurrentLogsWithDirInfo) {
-            EntryLogger.closeFileChannel(currentLogWithDirInfo.getLogChannel());
+            IOUtils.close(log, currentLogWithDirInfo.getLogChannel());
         }
     }
 
@@ -562,7 +563,7 @@ class EntryLogManagerForEntryLogPerLedger extends EntryLogManagerBase {
     public void forceClose() {
         Set<BufferedLogChannelWithDirInfo> copyOfCurrentLogsWithDirInfo = getCopyOfCurrentLogs();
         for (BufferedLogChannelWithDirInfo currentLogWithDirInfo : copyOfCurrentLogsWithDirInfo) {
-            EntryLogger.forceCloseFileChannel(currentLogWithDirInfo.getLogChannel());
+            IOUtils.close(log, currentLogWithDirInfo.getLogChannel());
         }
     }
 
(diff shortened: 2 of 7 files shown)


merlimat:
> lgtm. Will you mention in the description that it's the bytebuf that would have leaked.

@ivankelly updated description

sijie:
journal file close: https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java#L375

buffered channel close: https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java#L1069

merlimat:
Good point. I don't see why we should not close the file descriptor and release the buffer at once in the journal (after the last fsync is done). 
I don't think it's important to in this context to free the buffer earlier than that.

merlimat:
@sijie PTAL at 2nd commit. I think it should be a safe change

merlimat:
> in Journal line #1116-1117, there is double-close. you need to change that. I think there might be other places that have similar usage, we need to change that to be only one close. 

Ouch, I'll definitely take a closer look

merlimat:
@sijie Updated to fix the double close. I've checked around and couldn't find any other place to fix in Journal

eolivelli:
Good catch @sijie

sijie:
@merlimat 

this code change looks good for master, but I am concerned about the test failures, esp around "IllegalReferenceCountException". 

I think a safer change is to change Pooled buffer to Unpooled, *at least for 4.7.1*.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -1140,24 +1140,15 @@ public void shutdown() {
...
 
-    static void closeFileChannel(BufferedChannelBase channel) throws IOException {
-        if (null == channel) {
-            return;
-        }
-
-        FileChannel fileChannel = channel.getFileChannel();
-        if (null != fileChannel) {
-            fileChannel.close();
+    static void closeFileChannel(BufferedChannel channel) throws IOException {
```
reddycharan: @merlimat I don't see the need of having this static method. I think we can completely get rid of this method and the caller instead of calling "EntryLogger.closeFileChannel(bc)", they can just call "bc.close()"

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -1140,24 +1140,15 @@ public void shutdown() {
...
     }
 
-    static void forceCloseFileChannel(BufferedChannelBase channel) {
-        if (null == channel) {
-            return;
-        }
-        FileChannel fileChannel = channel.getFileChannel();
-        if (null != fileChannel) {
-            IOUtils.close(LOG, fileChannel);
+    static void forceCloseFileChannel(BufferedChannel channel) {
```
reddycharan: @merlimat I don't see the need for this method. The caller can directly call "IOUtils.close(LOG, channel)", name of this method 'forceCloseFileChannel' is misleading and also it makes no sense to have this static method in EntryLogger class where it is just passing call to other static method.

reddycharan:
@merlimat I made couple of comments for this change, can you please check them. Since it is release blocker / kind of critical fix, I would request you to resolve this sooner. Also for EntryLogManagerForEntryLogPerLedger case, this is going to be serious issue since there are going to be multiple entrylogs.

@sijie @eolivelli @jvrao fyi.

reddycharan:
Also having a unit test case would be helpful, like after flushRotatedLogs is called BufferedChannel's 'writeBuffer''s refcount is brought down to 0.

eolivelli:
retest this please

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
merlimat: Make sense. These 2 methods were doing more work before but they have no need to be there now. Changed.

merlimat:
> Also having a unit test case would be helpful, like after flushRotatedLogs is called BufferedChannel's 'writeBuffer''s refcount is brought down to 0. 

Yes, the `BufferedChannel` in `EntryLogger` is closed twice on shutdown. Added check to handle double close.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogManagerForEntryLogPerLedger.java
```diff
@@ -554,15 +555,15 @@ public File getDirForNextEntryLog(List<File> writableLedgerDirs) {
     public void close() throws IOException {
         Set<BufferedLogChannelWithDirInfo> copyOfCurrentLogsWithDirInfo = getCopyOfCurrentLogs();
         for (BufferedLogChannelWithDirInfo currentLogWithDirInfo : copyOfCurrentLogsWithDirInfo) {
-            EntryLogger.closeFileChannel(currentLogWithDirInfo.getLogChannel());
+            IOUtils.close(log, currentLogWithDirInfo.getLogChannel());
```
reddycharan: any reason why "IOUtils.close" is called instead of "currentLogWithDirInfo.getLogChannel().close()"

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -675,10 +675,12 @@ void removeCurCompactionLog() {
...
                     LOG.warn("Could not delete compaction log file {}", compactionLogChannel.getLogFile());
                 }
+
                 try {
-                    closeFileChannel(compactionLogChannel);
+                    compactionLogChannel.close();
                 } catch (IOException e) {
-                    LOG.error("Failed to close file channel for compaction log {}", compactionLogChannel.getLogId());
+                    LOG.error("Failed to close file channel for compaction log {}: {}", compactionLogChannel.getLogId(),
+                            e.getMessage());
```
reddycharan: instead of logging just exception message, it would be better to log complete exception, right? we would get complete call stack trace aswell

LOG.error("Failed to close file channel for compaction log {}", compactionLogChannel.getLogId(), e);

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -1084,9 +1084,7 @@ public void run() {
                             // check whether journal file is over file limit
                             if (shouldRolloverJournal) {
                                 // if the journal file is rolled over, the journal file will be closed after last
-                                // entry is force written to disk. the `bc` is not used anymore, so close it to release
-                                // the buffers in `bc`.
-                                IOUtils.close(LOG, bc);
```
reddycharan: why is this call removed?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedChannel.java
```diff
@@ -39,7 +40,7 @@
     // The position of the file channel's write pointer.
     protected AtomicLong writeBufferStartPosition = new AtomicLong(0);
     // The buffer used to write operations.
-    protected final ByteBuf writeBuffer;
+    protected ByteBuf writeBuffer;
```
reddycharan: any reason why 'final' is removed?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedChannel.java
```diff
@@ -86,8 +87,13 @@ public BufferedChannel(FileChannel fc, int writeCapacity, int readCapacity, long
     }
 
     @Override
-    public void close() throws IOException {
-        writeBuffer.release();
+    public synchronized void close() throws IOException {
+        if (writeBuffer != null) {
+            ReferenceCountUtil.safeRelease(writeBuffer);
+            writeBuffer = null;
```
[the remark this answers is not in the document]
merlimat: Yes, that's how it was earlier in this PR. The problem is that close() gets called twice on the active write channel when entrylogger gets shutdown. Let me try to rethink that

reddycharan:
@merlimat i see you addressed one of my comment, but still there are few unresponded comments. Can you please check them as well. I think it would be much helpful if we have some level of testing for bytebuf count in these fixes.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
merlimat: As the comment says, the `logFile` object is closed in the syncThread after the last entry is persisted. 

Since the `BufferedChannel` (`bc`) is a component of the `JournalChannel` (`logFile`), in this PR I have associated the `logFile.close()` to also close the internal buffered channel, since `logFile` is the "owner" of the buffered channel.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogManagerForEntryLogPerLedger.java
reddycharan: if null check is what you are looking for then probably doing null check is better instead of calling IOUtils, since it suppresses log message.

Also for the sake of consistency do the same thing as what you are doing in EntryLogManagerForSingleEntryLog 
https://github.com/apache/bookkeeper/pull/1414/files#diff-64c79da08c688a4e03da592ab424d871R215

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
reddycharan: thanks for the explanation.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogManagerForEntryLogPerLedger.java
merlimat: Sure

merlimat:
@reddycharan Addressed all your comments

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedChannel.java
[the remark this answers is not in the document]
jvrao: Thanks @merlimat calling twice is an issue and we should address that. :)

reddycharan:
@merlimat LGTM with the changes. My only remaining concern is we aren't having test coverage validating refcounts. Anyhow this is broader issue, I'm fine if you feel there is no necessary for adding test coverage for this fix now.

@sijie @jvrao @eolivelli just FYI about test coverage.

eolivelli:
rerun bookkeeper bookie test

eolivelli:

reddycharan:
@sijie can you do final review and also this PR was open for quite sometime.

Issue #1415, opened 2018-05-19
org.apache.distributedlog.fs.TestDLFileSystem fails on CI


eolivelli:
The tests seems flaky

org.apache.distributedlog.fs.TestDLFileSystem.testBasicIO

Failing for the past 1 build (Since Instabile#40 )
Took 0.88 s.
Error Message
/tmp/testBasicIO (Permission denied)
Stacktrace
java.io.FileNotFoundException: /tmp/testBasicIO (Permission denied)
	at org.apache.distributedlog.fs.TestDLFileSystem.testBasicIO(TestDLFileSystem.java:68

sijie:
it seems CI doesn't allow creating file under "/tmp".

so it might be worth changing the base dir to use `tmpDir`: https://github.com/apache/bookkeeper/blob/master/stream/distributedlog/io/dlfs/src/test/java/org/apache/distributedlog/fs/TestDLFileSystem.java#L64

Anyone interested in fixing this?

PR #1417, opened 2018-05-21
Issue-1415: [DLOG] org.apache.distributedlog.fs.TestDLFileSystem fails on CI

eolivelli:
Use a generated tempDir for DL FS tests instead of using an hard coded /tmp path

Fix for #1415

### DIFF
--- a/stream/distributedlog/core/src/test/java/org/apache/distributedlog/TestDistributedLogBase.java
+++ b/stream/distributedlog/core/src/test/java/org/apache/distributedlog/TestDistributedLogBase.java
@@ -97,12 +97,6 @@ public class TestDistributedLogBase {
     protected static int numBookies = 3;
     private static final List<File> tmpDirs = new ArrayList<File>();
 
-    protected static File newTempDir(String name) throws IOException {
-        File zkTmpDir = IOUtils.createTempDir(name, "distrlog");
-        tmpDirs.add(zkTmpDir);
-        return zkTmpDir;
-    }
-
     @BeforeClass
     public static void setupCluster() throws Exception {
         File zkTmpDir = IOUtils.createTempDir("zookeeper", "distrlog");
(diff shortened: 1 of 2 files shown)


stream/distributedlog/io/dlfs/src/test/java/org/apache/distributedlog/fs/TestDLFileSystem.java
```diff
@@ -61,7 +61,7 @@ public void testBasicIO() throws Exception {
         }
         assertTrue(fs.exists(path));
 
-        File tempFile = new File("/tmp/" + runtime.getMethodName());
+        File tempFile = new File(newTempDir(runtime.getMethodName()), "test.bin");
```
sijie: this might actually cause the files never deleted. because new temp dir will generate a new directory and it is never deleted. 

I would suggest using JUnit temp folder.

https://junit.org/junit4/javadoc/4.12/org/junit/rules/TemporaryFolder.html

https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieJournalForceTest.java#L72

eolivelli: I tried to use the same style as DL tests, I am not still used to this codebase.

Temporary Folder is better for me. Will change

stream/distributedlog/core/src/test/java/org/apache/distributedlog/TestDistributedLogBase.java
```diff
@@ -97,6 +97,12 @@
     protected static int numBookies = 3;
     private static final List<File> tmpDirs = new ArrayList<File>();
 
+    protected static File newTempDir(String name) throws IOException {
+        File zkTmpDir = IOUtils.createTempDir(name, "distrlog");
```
jiazhai: in the name, `zk` looks not needed, how about rename it?

eolivelli:
@jiazhai @sijie updated the patch with TemporaryFolder.
Please review, so that we can with CI soon

PR #1420, opened 2018-05-21
Allow bookie to start if failed to load extra server components

sijie:

*Motivation*

`extraServerComponents` provides the flexibility to allow bookie to start with extra server components.
It acts as a plugin mechanism to extend the functionality of bookies. However it is okay to allow bookie
start up when failed to load extra server components.

This is a change for allowing starting table service as extra components in bookies.

*Solution*

Catch the exception and log it when failed to load extra server components.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/server/Main.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/server/Main.java
@@ -48,6 +48,7 @@ import org.apache.commons.cli.Option;
 import org.apache.commons.cli.Options;
 import org.apache.commons.cli.ParseException;
 import org.apache.commons.configuration.ConfigurationException;
+import org.apache.commons.lang3.StringUtils;
 
 /**
  * A bookie server is a server that run bookie and serving rpc requests.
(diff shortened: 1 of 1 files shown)


sijie:
@eolivelli pls read the description, this is done in purpose.

eolivelli:
@sijie I read the description.
If a component fails to start the user can change the config and choose explicitly not to start it. 
If you let the process live maybe it will be hard to early detect the failure

sijie:
@eolivelli - `extraServerComponents` was introduced for non-stateful components like table services. 1) If one bookie fails to start the server components, it doesn't impact the cluster, it is okay for one bookie to not start those components. the most important thing is to not impact existing bookie functionalities. 2) the table service is in developer-preview, so it doesn't make sense to crash bookie if it fails to start due to any potential bugs.

jvrao:
@sijie  I agree with @eolivelli 's comment. There should be a configuration parameter if the user
wanted or don't want to load something. I don't see the value of load if possible. I am afraid this can lead to problems in future. Why not introduce a configuration value?

sijie:
@jvrao okay. will introduce the configuration value.

sijie:
@jvrao @eolivelli - put up a PR #1426 to address the comments.

PR #1421, opened 2018-05-21
Provide a util method to run functions with metadata client driver

sijie:

*Motivation*

Currently `MetadataDrivers` provides util methods to run functions with metadata bookie driver.
It is convinient to provide a util method to run functions with metadata client driver as well.

*Solution*

Provide a util method `runFunctionWithMetadataClientDriver` to run functions with metadata client driver.


sijie:
This is a util method used for a subsequent change, so the tests are covered by subsequent change.

jiazhai:
Oh, found the use case in PR 1422 . 👍

Issue #1205, opened 2018-02-26
BP-30: BookKeeper Table Service


sijie:
**BP**

This is the master ticket for tracking BP-30 :

This BP is proposing adding a table (aka key/value) service component as a contrib module to the bookkeeper, which can later be used for metadata storage. This BP together with other BPs forms the ideas of improving metadata management in bookkeeper.

BP Details: https://docs.google.com/document/d/155xAwWv5IdOitHh1NVMEwCMGgB28M3FyMiQSxEpjE-Y/edit#heading=h.56rbh52koe3f

Proposal PR - #1185 

Tasks:

- [x] copy the table service code as the contrib module #1207 
- [x] add the module to profile and only activate it for certain profiles #1216 
- [x] enable ci when code changes touch the module #1216 
- [x] update dist package, docker image
- [ ] documentation

eolivelli:
@sijie  seems we are only lacking the 'documentation' subtask and this will be closed

sijie:
@eolivelli yes the basic functions are there. so we can mark this BP as closed after we have documentation. there will be more improvements/changes in future, but they are separated from this BP.

xiemeilong:
When will this be documented. It seems a useful feature.

hsaputra:
@sijie , @eolivelli - seems like the issue is still missing documentation for the Table service API

eolivelli:
@hsaputra you are totally right !

we really need doc for this feature

cc @Jennifer88huang

sijie:
@hsaputra Apologized that I missed this message! Yes, we still need documentation. The feature is still a beta version. I know your team at Splunk is actively working on it. I am not sure if there are any major API changes or not. We might want to wait for your team to contribute the changes back to the bookkeeper project before @Jennifer88huang starts writing the documentation for it. What do you think?

hsaputra:
HI @sijie , sorry for late response. 
I think it would be ok to wait until the contribution is coming back upstream. It gives motivation to accelerate the initiative.

hsaputra:
HI @sijia-w , @eolivelli - After talking to team at Splunk working on the State Store upgrades, looks like there is no, or very little changes to the Table Service API themselves.

I have filed bug for adding documentation for this new feature. It should be ok for @Jennifer88huang  to start adding doc for this feature.

FYI @merlimat

sijie:
@hsaputra thank you for the heads up!  I will plan that with @Jennifer88huang.

[...]

PR #1422, opened 2018-05-21
[table service] start table service as an extra component of bookie

sijie:

*Motivation*

table service was built over bookkeeper ledgers. it is an extension to bookies' functionalities, so it is much convenient to start the service as one additional component with bookie, just like how we start `AutoRecovery` along with bookie.

*Solution*

- include `stream/server` module as part of bookkeeper server/all binary distributions
- abstract `StorageServer` to allow controlling whether to start bookie or not
- provide a ServerLifecycleComponent of `StorageServer`, so it can be used as an extra component of bookie

*Tests*

- improve the integration tests to start table service as part of bookie containers
- move `LocationClientTest` to container based integration tests

Master Issue: #1205


sijie:
This PR includes changes from #1421 and #1420

bookkeeper-server/src/main/java/org/apache/bookkeeper/server/Main.java
```diff
@@ -327,13 +328,18 @@ static LifecycleComponentStack buildBookieServer(BookieConfiguration conf) throw
...
+                    extraComponents,
+                    conf,
+                    rootStatsLogger);
+                for (ServerLifecycleComponent component : components) {
+                    serverBuilder.addComponent(component);
+                    log.info("Load lifecycle component : {}", component.getClass().getName());
+                }
+            } catch (Exception e) {
+                log.info("Failed to load extra components '{}' - {}. Continuing without those components.",
+                    StringUtils.join(extraComponents), e.getMessage());
```
eolivelli: This would not be safe.
Maybe the bookie is up but it is, not doing what it is supposed to do

jiazhai: Seems this is for handling issue of other components, which is not bookie. It is make sure the bookie still get running while other components meet exception.

sijie:
retest this please

sijie:
retest this please

PR #1427, opened 2018-05-23
Publish bookkeeper dev image to docker hub

sijie:

*Motivation*

We built a `bookkeeper-current` image that reflects latest master for integration tests. However currently we don't publish this docker image to any docker registry. So it is inconvenient to test latest master in dockerized environments.

*Solution*

- Add a dev script to publish `bookkeeper-current` image to configured docker registry specified at `DOCKER_REGISTRY`
- Update the nightly snapshot job to publish generated `bookkeeper-current` image to docker hub `apachebookkeeper`
- Add `bin/standalone` script to use `docker-compose` to launch a 3-nodes bookkeeper cluster locally. The script generates docker-compose.yml and handles mounting local volumes and port forwarding. so developers can use that for quick development.


bin/standalone.docker-compose
```diff
@@ -0,0 +1,177 @@
...
+
+
+${DOCKER_COMPOSE} $@ ${DOCKER_COMPOSE_OPTS}
+
+if [ $? == 0 -a "${ACTION}" == "up" ]; then
+    echo ""
+    echo "Standalone cluster '${CLUSTER_NAME}' is up running."
+    echo "Use following uris to connect to standalone cluster:"
+    echo ""
+    echo "  - metadata service uri = 'zk://localhost/ledgers'"
```
jiazhai: Could this be like: zk://metadata-service/ledgers ?

PR #1429, opened 2018-05-23
Fix bookkeeper post commit CI

sijie:

*Motivation*

apache/bookkeeper#1422 includes stream storage integration tests in tests/integration.
so we need to include `-Dstream` on building the tests.

*Solution*

Update the bookkeeper post commit CI jobs to include `-Dstream`


sijie:
IGNORE CI

sijie:
This PR only touches CI groovy files.

PR #1436, opened 2018-05-24
BP-14 force() API - client side implementation

eolivelli:
- Introduce the client side force() API
- Implementation on the client side wire protocol for FORCE_LEDGER RPC
- Disable ensemble changes for DEFERRED_SYNC writers
- Prevent v2 client from using force() API.

The force() API enables the client (usually with DEFERRED_SYNC write flags) to require a point of synchronization with all the bookies in the ensemble, to have guarantees about durability of previously written entries (and ackknowledgerd), this way LastAddConfirmed is able to advance.

For DEFERRED_SYNC writers LastAddConfirmed will advance only using this API

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperClientStats.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperClientStats.java
@@ -41,6 +41,7 @@ public interface BookKeeperClientStats {
 
     String ADD_OP = "ADD_ENTRY";
     String ADD_OP_UR = "ADD_ENTRY_UR"; // Under Replicated during AddEntry.
+    String FORCE_OP = "FORCE";
     String READ_OP = "READ_ENTRY";
     // Corrupted entry (Digest Mismatch/ Under Replication) detected during ReadEntry
     String READ_OP_DM = "READ_ENTRY_DM";
@@ -64,7 +65,9 @@ public interface BookKeeperClientStats {
     String CHANNEL_ADD_OP = "ADD_ENTRY";
     String CHANNEL_TIMEOUT_ADD = "TIMEOUT_ADD_ENTRY";
     String CHANNEL_WRITE_LAC_OP = "WRITE_LAC";
+    String CHANNEL_FORCE_OP = "FORCE";
     String CHANNEL_TIMEOUT_WRITE_LAC = "TIMEOUT_WRITE_LAC";
+    String CHANNEL_TIMEOUT_FORCE = "TIMEOUT_FORCE";
     String CHANNEL_READ_LAC_OP = "READ_LAC";
     String CHANNEL_TIMEOUT_READ_LAC = "TIMEOUT_READ_LAC";
     String TIMEOUT_GET_BOOKIE_INFO = "TIMEOUT_GET_BOOKIE_INFO";
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperClientStats.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperClientStats.java
@@ -41,7 +41,7 @@ public interface BookKeeperClientStats {
 
     String ADD_OP = "ADD_ENTRY";
     String ADD_OP_UR = "ADD_ENTRY_UR"; // Under Replicated during AddEntry.
-    String FORCE_OP = "FORCE";
+    String FORCE_OP = "FORCE"; // Number of force ledger operations
     String READ_OP = "READ_ENTRY";
     // Corrupted entry (Digest Mismatch/ Under Replication) detected during ReadEntry
     String READ_OP_DM = "READ_ENTRY_DM";
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/DistributionSchedule.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/DistributionSchedule.java
@@ -142,19 +142,12 @@ public interface DistributionSchedule {
     WriteSet getWriteSet(long entryId);
 
     /**
-     * Return the set of bookies indices to send the messages to for longpoll reads.
+     * Return the set of bookies indices to send the messages to the whole ensemble.
      *
-     * @param entryId expected next entry id to read.
-     * @return the set of bookies indices to read from.
+     * @param entryId entry id use to calculate the ensemble.
+     * @return the set of bookies indices to send the request.
      */
-    WriteSet getWriteSetForLongPoll(long entryId);
-
-    /**
-     * Return the set of bookies indices to send the messages to for force ledger.
-     *
-     * @return the set of bookies indices to send force request to.
-     */
-    WriteSet getWriteSetForForceLedger();
+    WriteSet getEnsembleSet(long entryId);
 
     /**
      * An ack set represents the set of bookies from which
@@ -208,7 +201,7 @@ public interface DistributionSchedule {
      * Returns an ackset object useful to wait for all bookies in the ensemble,
      * responses should be checked against this.
      */
-    AckSet getAckSetForForceLedger();
+    AckSet getEnsembleAckSet();
 
     /**
      * Interface to keep track of which bookies in an ensemble, an action
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/DistributionSchedule.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/DistributionSchedule.java
@@ -144,7 +144,7 @@ public interface DistributionSchedule {
     /**
      * Return the set of bookies indices to send the messages to the whole ensemble.
      *
-     * @param entryId entry id use to calculate the ensemble.
+     * @param entryId entry id used to calculate the ensemble.
      * @return the set of bookies indices to send the request.
      */
     WriteSet getEnsembleSet(long entryId);
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/DistributionSchedule.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/DistributionSchedule.java
@@ -150,6 +150,13 @@ public interface DistributionSchedule {
     WriteSet getWriteSetForLongPoll(long entryId);
 
     /**
+     * Return the set of bookies indices to send the messages to for force ledger.
+     *
+     * @return the set of bookies indices to send force request to.
+     */
+    WriteSet getWriteSetForForceLedger();
+
+    /**
      * An ack set represents the set of bookies from which
      * a response must be received so that an entry can be
      * considered to be replicated on a quorum.
@@ -197,6 +204,11 @@ public interface DistributionSchedule {
      */
     AckSet getAckSet();
 
+    /**
+     * Returns an ackset object useful to wait for all bookies in the ensemble,
+     * responses should be checked against this.
+     */
+    AckSet getAckSetForForceLedger();
 
     /**
      * Interface to keep track of which bookies in an ensemble, an action
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ForceLedgerOp.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ForceLedgerOp.java
@@ -90,7 +90,7 @@ class ForceLedgerOp extends SafeRunnable implements ForceLedgerCallback {
         checkState(!completed, "We are waiting for all the bookies, it is not expected an early exit");
 
         if (errored) {
-            // alredy failed, we alread called the callback reporting a failure
+            // already failed, do not fire error callbacks twice
             return;
         }
 
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ForceLedgerOp.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ForceLedgerOp.java
@@ -17,8 +17,8 @@
  */
 package org.apache.bookkeeper.client;
 
+import static com.google.common.base.Preconditions.checkState;
 import java.util.ArrayList;
-import java.util.BitSet;
 import java.util.concurrent.CompletableFuture;
 import org.apache.bookkeeper.common.concurrent.FutureUtils;
 import org.apache.bookkeeper.net.BookieSocketAddress;
@@ -34,7 +34,6 @@ class ForceLedgerOp extends SafeRunnable implements ForceLedgerCallback {
 
     private static final Logger LOG = LoggerFactory.getLogger(ForceLedgerOp.class);
     final CompletableFuture<Void> cb;
-    BitSet receivedResponseSet;
 
     DistributionSchedule.AckSet ackSet;
     boolean completed = false;
@@ -65,17 +64,15 @@ class ForceLedgerOp extends SafeRunnable implements ForceLedgerCallback {
         // remember that we are inside OrderedExecutor, this induces a strict ordering
         // on the sequence of events
         this.currentNonDurableLastAddConfirmed = lh.pendingAddsSequenceHead;
-        LOG.debug("force {} clientNonDurableLac {}", lh.ledgerId, currentNonDurableLastAddConfirmed);
-
+        if (LOG.isDebugEnabled()) {
+            LOG.debug("force {} clientNonDurableLac {}", lh.ledgerId, currentNonDurableLastAddConfirmed);
+        }
         // we need to send the request to every bookie in the ensamble
         this.currentEnsemble = lh.metadata.currentEnsemble;
-        this.ackSet = lh.distributionSchedule.getAckSetForForceLedger();
-        this.receivedResponseSet = new BitSet(
-                lh.getLedgerMetadata().getEnsembleSize());
-        this.receivedResponseSet.set(0,
-                lh.getLedgerMetadata().getEnsembleSize());
+        this.ackSet = lh.distributionSchedule.getEnsembleAckSet();
 
-        DistributionSchedule.WriteSet writeSet = lh.distributionSchedule.getWriteSetForForceLedger();
+        DistributionSchedule.WriteSet writeSet = lh.getDistributionSchedule()
+                                                   .getEnsembleSet(currentNonDurableLastAddConfirmed);
         try {
             for (int i = 0; i < writeSet.size(); i++) {
                 sendForceLedgerRequest(writeSet.get(i));
@@ -89,35 +86,30 @@ class ForceLedgerOp extends SafeRunnable implements ForceLedgerCallback {
     public void forceLedgerComplete(int rc, long ledgerId, BookieSocketAddress addr, Object ctx) {
         int bookieIndex = (Integer) ctx;
 
-        if (completed) {
-            return;
-        }
+        checkState(!completed, "We are waiting for all the bookies, it is not expected an early exit");
 
         if (BKException.Code.OK != rc) {
             lastSeenError = rc;
         }
 
-        // We got response.
-        receivedResponseSet.clear(bookieIndex);
-
         if (rc == BKException.Code.OK) {
-            if (ackSet.completeBookieAndCheck(bookieIndex) && !completed) {
+            if (ackSet.completeBookieAndCheck(bookieIndex)) {
                 completed = true;
                 // we are able to say that every bookie sync'd its own journal
                 // for every ackknowledged entry before issuing the force() call
-                LOG.debug("After force on ledger {} updating LastAddConfirmed to {} ",
-                        ledgerId, currentNonDurableLastAddConfirmed);
-                lh.lastAddConfirmed = currentNonDurableLastAddConfirmed;
+                if (LOG.isDebugEnabled()) {
+                    LOG.debug("After force on ledger {} updating LastAddConfirmed to {} ",
+                              ledgerId, currentNonDurableLastAddConfirmed);
+                }
+                lh.updateLastConfirmed(currentNonDurableLastAddConfirmed, lh.getLength());
                 FutureUtils.complete(cb, null);
-                return;
             }
         } else {
-            LOG.warn("ForceLedger did not succeed: Ledger {} on {}", new Object[]{ledgerId, addr});
-        }
-
-        if (receivedResponseSet.isEmpty()) {
-            completed = true;
+            // at least one bookie failed, as we are waiting for all the bookies
+            // we can fail immediately
+            LOG.info("ForceLedger did not succeed: Ledger {} on {}", ledgerId, addr);
             FutureUtils.completeExceptionally(cb, BKException.create(lastSeenError));
         }
+
     }
 }
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ForceLedgerOp.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ForceLedgerOp.java
@@ -37,6 +37,7 @@ class ForceLedgerOp extends SafeRunnable implements ForceLedgerCallback {
 
     DistributionSchedule.AckSet ackSet;
     boolean completed = false;
+    boolean errored = false;
     int lastSeenError = BKException.Code.WriteException;
     ArrayList<BookieSocketAddress> currentEnsemble;
 
@@ -88,6 +89,11 @@ class ForceLedgerOp extends SafeRunnable implements ForceLedgerCallback {
 
         checkState(!completed, "We are waiting for all the bookies, it is not expected an early exit");
 
+        if (errored) {
+            // alredy failed, we alread called the callback reporting a failure
+            return;
+        }
+
         if (BKException.Code.OK != rc) {
             lastSeenError = rc;
         }
@@ -108,6 +114,12 @@ class ForceLedgerOp extends SafeRunnable implements ForceLedgerCallback {
             // at least one bookie failed, as we are waiting for all the bookies
             // we can fail immediately
             LOG.info("ForceLedger did not succeed: Ledger {} on {}", ledgerId, addr);
+            errored = true;
+
+            // fail all pending writes
+            lh.errorOutPendingAdds(rc);
+
+            // notify the failure
             FutureUtils.completeExceptionally(cb, BKException.create(lastSeenError));
         }
 
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ForceLedgerOp.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ForceLedgerOp.java
@@ -65,7 +65,7 @@ class ForceLedgerOp extends SafeRunnable implements ForceLedgerCallback {
         // remember that we are inside OrderedExecutor, this induces a strict ordering
         // on the sequence of events
         this.currentNonDurableLastAddConfirmed = lh.pendingAddsSequenceHead;
-        LOG.debug("start force() on ledger {} capturing {} ", lh.ledgerId, currentNonDurableLastAddConfirmed);
+        LOG.debug("force {} clientNonDurableLac {}", lh.ledgerId, currentNonDurableLastAddConfirmed);
 
         // we need to send the request to every bookie in the ensamble
         this.currentEnsemble = lh.metadata.currentEnsemble;
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ForceLedgerOp.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ForceLedgerOp.java
@@ -33,7 +33,7 @@ import org.slf4j.LoggerFactory;
 class ForceLedgerOp extends SafeRunnable implements ForceLedgerCallback {
 
     private static final Logger LOG = LoggerFactory.getLogger(ForceLedgerOp.class);
-    CompletableFuture<Void> cb;
+    final CompletableFuture<Void> cb;
     BitSet receivedResponseSet;
 
     DistributionSchedule.AckSet ackSet;
@@ -43,7 +43,7 @@ class ForceLedgerOp extends SafeRunnable implements ForceLedgerCallback {
 
     long currentNonDurableLastAddConfirmed = LedgerHandle.INVALID_ENTRY_ID;
 
-    LedgerHandle lh;
+    final LedgerHandle lh;
 
     ForceLedgerOp(LedgerHandle lh, CompletableFuture<Void> cb) {
         this.lh = lh;
@@ -61,7 +61,9 @@ class ForceLedgerOp extends SafeRunnable implements ForceLedgerCallback {
 
     void initiate() {
 
-        // capture currentNonDurableLastAddConfirmed, we are inside OrderedExecutor
+        // capture currentNonDurableLastAddConfirmed
+        // remember that we are inside OrderedExecutor, this induces a strict ordering
+        // on the sequence of events
         this.currentNonDurableLastAddConfirmed = lh.pendingAddsSequenceHead;
         LOG.debug("start force() on ledger {} capturing {} ", lh.ledgerId, currentNonDurableLastAddConfirmed);
 
@@ -103,7 +105,7 @@ class ForceLedgerOp extends SafeRunnable implements ForceLedgerCallback {
                 completed = true;
                 // we are able to say that every bookie sync'd its own journal
                 // for every ackknowledged entry before issuing the force() call
-                LOG.info("After force on ledger {} updating LastAddConfirmed to {} ",
+                LOG.debug("After force on ledger {} updating LastAddConfirmed to {} ",
                         ledgerId, currentNonDurableLastAddConfirmed);
                 lh.lastAddConfirmed = currentNonDurableLastAddConfirmed;
                 FutureUtils.complete(cb, null);
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ForceLedgerOp.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ForceLedgerOp.java
@@ -116,9 +116,6 @@ class ForceLedgerOp extends SafeRunnable implements ForceLedgerCallback {
             LOG.info("ForceLedger did not succeed: Ledger {} on {}", ledgerId, addr);
             errored = true;
 
-            // fail all pending writes
-            lh.errorOutPendingAdds(rc);
-
             // notify the failure
             FutureUtils.completeExceptionally(cb, BKException.create(lastSeenError));
         }
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ForceLedgerOp.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ForceLedgerOp.java
@@ -1,0 +1,121 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.bookkeeper.client;
+
+import java.util.ArrayList;
+import java.util.BitSet;
+import java.util.concurrent.CompletableFuture;
+import org.apache.bookkeeper.common.concurrent.FutureUtils;
+import org.apache.bookkeeper.net.BookieSocketAddress;
+import org.apache.bookkeeper.proto.BookkeeperInternalCallbacks.ForceLedgerCallback;
+import org.apache.bookkeeper.util.SafeRunnable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This represents a request to sync the ledger on every bookie.
+ */
+class ForceLedgerOp extends SafeRunnable implements ForceLedgerCallback {
+
+    private static final Logger LOG = LoggerFactory.getLogger(ForceLedgerOp.class);
+    CompletableFuture<Void> cb;
+    BitSet receivedResponseSet;
+
+    DistributionSchedule.AckSet ackSet;
+    boolean completed = false;
+    int lastSeenError = BKException.Code.WriteException;
+    ArrayList<BookieSocketAddress> currentEnsemble;
+
+    long currentNonDurableLastAddConfirmed = LedgerHandle.INVALID_ENTRY_ID;
+
+    LedgerHandle lh;
+
+    ForceLedgerOp(LedgerHandle lh, CompletableFuture<Void> cb) {
+        this.lh = lh;
+        this.cb = cb;
+    }
+
+    void sendForceLedgerRequest(int bookieIndex) {
+        lh.bk.getBookieClient().forceLedger(currentEnsemble.get(bookieIndex), lh.ledgerId, this, bookieIndex);
+    }
+
+    @Override
+    public void safeRun() {
+        initiate();
+    }
+
+    void initiate() {
+
+        // capture currentNonDurableLastAddConfirmed, we are inside OrderedExecutor
+        this.currentNonDurableLastAddConfirmed = lh.pendingAddsSequenceHead;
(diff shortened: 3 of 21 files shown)


eolivelli:
This is still WIP but almost complete, missing items:
- prevent v2 clients to use force() API
- add test cases for corner cases
- disable ensemble changes with DEFERRED_SYNC (this can be another patch)

@jvrao @sijie if you have cycles you can start your review

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/DeferredSyncTest.java
```diff
@@ -49,10 +51,111 @@ public void testAddEntryLastAddConfirmedDoesNotAdvance() throws Exception {
...
+            for (int i = 0; i < NUM_ENTRIES - 1; i++) {
+                result(wh.appendAsync(DATA));
+            }
+            long lastEntryIdBeforeSuspend = result(wh.appendAsync(DATA));
+            assertEquals(NUM_ENTRIES - 1, lastEntryIdBeforeSuspend);
+            assertEquals(-1, wh.getLastAddConfirmed());
+
+            // one bookie will stop sending acks
+            BookieSocketAddress bookieAddress = wh.getLedgerMetadata().getEnsembleAt(wh.getLastAddPushed()).get(0);
+            suspendBookieWriteAcks(bookieAddress);
```
eolivelli: this test does not make really much sense. I will drop it.
This is testing:
- client send *addEntry*
- client send *force*
- client received ack for *force*
- client received ack for *addEntry*
this is not actually possible due to ordered executor

I will add a test with:
- client send *addEntry* for entry 0
- client send *force* at entry 0
- client send *addEntry* for entry 1
- client received ack for *addEntry*
- client received ack for *force*  (LAC MUST still be 0, force should 'capture' the last completed addEntry, not the last pushed one)
- client received ack for *addEntry*

I will add similar tests for WriteAdvHandle

eolivelli:
@jvrao  @sijie the patch is ready for review now

eolivelli:
retest this please

eolivelli:
retest this please

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -1790,6 +1851,14 @@ void handleDelayedWriteBookieFailure() {
             }
             return;
         }
+        if (writeFlags.contains(WriteFlag.DEFERRED_SYNC)) {
+            if (LOG.isDebugEnabled()) {
+                LOG.debug("Cannot perform ensemble changee with writeflags {}."
```
jiazhai: a typo `changee`

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
```diff
@@ -1563,6 +1601,55 @@ public void handleV3Response(BookkeeperProtocol.Response response) {
...
+
+        @Override
+        public void handleV3Response(BookkeeperProtocol.Response response) {
+            ForceLedgerResponse forceLedgerResponse = response.getForceLedgerResponse();
+            StatusCode status = response.getStatus() == StatusCode.EOK
+                ? forceLedgerResponse.getStatus() : response.getStatus();
+            long ledgerId = forceLedgerResponse.getLedgerId();
+
+            if (LOG.isDebugEnabled()) {
+                logResponse(status, "ledger", ledgerId);
```
jiazhai: seems `logResponse` already has an `isDebugEnabled()` internal.

sijie:
@jvrao @eolivelli : I am wondering how would be the relationship between `force()` and `explicitLac`. my feeling is that these two operations are kind of similar/same operation, basically we are requesting advancing lac in either case.

If you guys agree with that, it is probably making more sense to reuse `explicitLac` as the request for `force()` api. we can add a flag in `explicitLac` indicating whether we want to force the journal. 

if no, we need to figure out what does following behavior mean:

- if a client enables explicit lac, what `explicitLac` is exactly doing for deferred sync ledgers.

eolivelli:
@sijie explicitlac for deferred sync writers only sends the value of LastAddConfirmed and it is still a way to notify LAC changes without a real write. The difference is that LastAddConfirmed variable does not advance at every write.

I think they are different operations, and need different semantics. Force() must be acknowledged by the full ensemble.
You can use force() without sending explicitlac, in fact new LAC will be piggybacked at next write for instance.

I think it is better to keep force and explicitlac separated.

I will add test cases to cover the usage of explicit LAC with deferred sync flag.

I will check the implementation today, in case I am missing something.

Cc @jvrao

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ForceLedgerOp.java
```diff
@@ -0,0 +1,123 @@
...
+        initiate();
+    }
+
+    void initiate() {
+
+        // capture currentNonDurableLastAddConfirmed
+        // remember that we are inside OrderedExecutor, this induces a strict ordering
+        // on the sequence of events
+        this.currentNonDurableLastAddConfirmed = lh.pendingAddsSequenceHead;
+        LOG.debug("start force() on ledger {} capturing {} ", lh.ledgerId, currentNonDurableLastAddConfirmed);
```
jvrao: instead of capturing {} ; just say clientNonDurableLac {}

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RoundRobinDistributionSchedule.java
```diff
@@ -60,6 +60,11 @@ public WriteSet getWriteSetForLongPoll(long entryId) {
         return WriteSetImpl.create(ensembleSize, ensembleSize /* writeQuorumSize */, entryId);
     }
 
+    @Override
+    public WriteSet getWriteSetForForceLedger() {
+        return WriteSetImpl.create(ensembleSize, ensembleSize /* writeQuorumSize */, 0);
```
jvrao: Have a comment similar to longPoll method above.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -1099,6 +1099,67 @@ public void asyncAddEntry(final long entryId, final byte[] data, final int offse
...
+            }
+        }
+
+        if (wasClosed) {
+            // make sure the callback is triggered in main worker pool
+            try {
+                bk.getMainWorkerPool().submit(new SafeRunnable() {
+                    @Override
+                    public void safeRun() {
+                        LOG.warn("Attempt to use a closed ledger: {}", ledgerId);
```
jvrao: Nit: Force()  attempted on a closed ledger {}

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -1099,6 +1099,67 @@ public void asyncAddEntry(final long entryId, final byte[] data, final int offse
...
+                        return String.format("force(lid=%d)", ledgerId);
+                    }
+                });
+            } catch (RejectedExecutionException e) {
+                result.completeExceptionally(new BKException.BKInterruptedException());
+            }
+            return result;
+        }
+
+        if (pendingAddsSequenceHead == INVALID_ENTRY_ID) {
```
jvrao: Please add comments on what it really means.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -1790,6 +1851,14 @@ void handleDelayedWriteBookieFailure() {
             }
             return;
         }
+        if (writeFlags.contains(WriteFlag.DEFERRED_SYNC)) {
+            if (LOG.isDebugEnabled()) {
+                LOG.debug("Cannot perform ensemble changee with writeflags {}."
+                        + "Failed bookies {} for ledger {}.",
+                        writeFlags, delayedWriteFailedBookies, ledgerId);
+            }
```
jvrao: Maybe we should treat this also unrecoverable error, or even fail at the caller itself.
Because if AQ< WQ we can continue to write but force is going to fail anyway. 
To be consistent, any write error (even if it is delayed write error it needs to fail under this flag)

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
```diff
@@ -284,7 +284,9 @@ public void writeComplete(int rc, long ledgerId, long entryId, BookieSocketAddre
                 // Got an error after satisfying AQ. This means we are under replicated at the create itself.
                 // Update the stat to reflect it.
                 addOpUrCounter.inc();
-                if (!lh.bk.getDisableEnsembleChangeFeature().isAvailable() && !lh.bk.delayEnsembleChange) {
+                if (!lh.bk.getDisableEnsembleChangeFeature().isAvailable()
+                        && !lh.writeFlags.contains(WriteFlag.DEFERRED_SYNC)
```
jvrao: What is the use of adding this check here?

sijie:
> I think they are different operations, and need different semantics. Force() must be acknowledged by the full ensemble.
You can use force() without sending explicitlac, in fact new LAC will be piggybacked at next write for instance.

I am not saying they are different operations. what I am questioning here is  - is it better to extend the explicit lac request with a flag that force flushing the data? In that way the behavior would be better defined when people enable explicit lac on a deferred-sync ledger.

Or we have to clarify following cases:

1. normal ledger, without explicit lac
2. normal ledger, with explicit lac
3. deferred ledger, without explicit lac
4. deferred ledger, with explicit lac

The case 4 is the most confused case.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
```diff
@@ -565,6 +572,37 @@ void writeLac(final long ledgerId, final byte[] masterKey, final long lac, ByteB
...
+                                                                OperationType.FORCE_LEDGER);
+        // force is mostly like addEntry hence uses addEntryTimeout
+        completionObjects.put(completionKey,
+                              new ForceLedgerCompletion(completionKey, cb,
+                                                     ctx, ledgerId));
+
+        // Build the request
+        BKPacketHeader.Builder headerBuilder = BKPacketHeader.newBuilder()
+                .setVersion(ProtocolVersion.VERSION_THREE)
+                .setOperation(OperationType.FORCE_LEDGER)
```
jvrao: Did we make protobuf changes for this?

jvrao:
>Or we have to clarify following cases

I don't understand where the explicitLAC is clashing with the deferred_sync. ExplicitLAC is completely independent operatin, where if the client thinks it's LAC is ahead of piggybacked LAC, it can send an update. and differedSync dictates how the client LAC gets updated. Maybe I am missing something @sijie

sijie:
>  don't understand where the explicitLAC is clashing with the deferred_sync.

@jvrao : explicit lac is advancing lac, while force is also advancing lac. my point is why can we consolidate these two behaviors as one. to me, 'explicit lac' can be a background 'force'; 'force' is an explicit application-defined 'explict lac' operation.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ForceLedgerOp.java
eolivelli: done

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RoundRobinDistributionSchedule.java
eolivelli: done

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
eolivelli: done

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
eolivelli: done

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
eolivelli: actually with writeflag = DEFERRED_SYNC this block of code will never run.
I have copied this block for consistency with other cases of "enabled" disableEnsembleChangeFeature.

I can drop it if you prefer, this code will never run

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
eolivelli: We are acting as "enabled" disableEnsembleChangeFeature, so I kept the same exact beheaviour

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
eolivelli: There is a new RPC for FORCE_LEDGER, nothing else

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperClientStats.java
```diff
@@ -41,6 +41,7 @@
 
     String ADD_OP = "ADD_ENTRY";
     String ADD_OP_UR = "ADD_ENTRY_UR"; // Under Replicated during AddEntry.
+    String FORCE_OP = "FORCE";
```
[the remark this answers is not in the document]
eolivelli: done

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
eolivelli: done

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
eolivelli: this is consistent with all the rest of the file, I can change every where, but I think that it can be a slight save of resources.
I would prefer to keep the style consistent.
I will change if you feel strong about it

eolivelli:
@jiazhai @jvrao I have addressed all of your comments and provided answers.

eolivelli:
@sijie I have added a test case about using ExplicitLAC and DEFERRED_SYNC.

I think that current behaviour is what I am expecting:
1) LastAddConfirmed advances only in case of force()
2) ExplicitLAC timer sends notifications only when LastAddConfirmed, in case of DEFERRED_SYNC this is possible only after force()
3) without force() ExplicitLAC does nothing, as expected.

It seems to me that this is consistent with the expectations of DEFERRED_SYNC writeflag.

> is it better to extend the explicit lac request with a flag that force flushing the data? In that way the behavior would be better defined when people enable explicit lac on a deferred-sync ledger.

I think that the client controls LAC using force() and ExplicitLAC is a background task which seems to me like an optimization.
I don't see much benefit in having an "implicit" force with the ExplicitLAC notification.

Maybe it is like the initial proposal of having a "piggybacked LastAddPersisted" on AddResponse which would have made LAC to advance even without an explicit "force()".

We could add a flag which enables a timer to send explicit force() periodically, or that performs a force() during the execution of ExplicitLAC timer.
I prefer to keep things separated at wire protocol, we can compose the two RPCs, they are lightweight and we are in a background task (the ExplicitLAC timer)

cc @jvrao

sijie:
@eolivelli @jvrao : 

if we step back to think about the behavior here, both of these two features are kind of "flush/sync" behavior. forget about the implementation here, think from a storage system, image we have a "flush" primitive - it basically flushes unwritten data and make these data available. 

so,

"explicit lac" - it is a background "flush", flushing the data between lac and lastAddPushed. it is running in background.
"force" - it is an explicit "flush", flushing the data between current lac and lastAddPushed, it is explicitly.

those are exactly similar/same behavior from this sense. 

if you look at filesystem, filesystem has pdflush doing background flush and provides fsync systemcall for applications to do explicitly flush. but under the hood, they are doing the same thing - flushing the unflushed data.

if you map the behavior of bookkeeper to filesystem, "explicit lac" is exactly the background flush "pdflush"; while "force" is the system call. 

my main point here is if these two behaviors are similar/same, we should consolidate the underlying mechanism rather than having two separate routes to do same operations.

eolivelli:
@sijie 
I agree at 100% with your higher level description. I would like to have such "background flush" .

I am looking deeply into ExplicitLAC, and my feeling is that it will need to be reworked a bit in order to make it usable for this purpose.
It *seems* to me that ExplicitLAC is not integrated with the usual workflow: if you only configure SetExplicitLACInterval LAC is not advancing automagically.

It would be lovely to have a new "background flush" mechanism which performs a background "force()" and then makes LAC advance without writes, like ExplicitLAC.

Maybe the best approach would be to open a new followup task to enhance and rework ExplicitLAC.

cc @jvrao

sijie:
> Maybe the best approach would be to open a new followup task to enhance and rework ExplicitLAC.

It is fine to do stuffs in a separate task. I think my main point in this PR, we need to be aligned with all the concepts around these two mechanisms. otherwise it is hard to maintain going forward.

eolivelli:
@sijie 
In the limited scope of this PR I think that it is good that we are not breaking ExplicitLAC and it is working as expected (I have added a test case as proof).

I think that in general, despite BP14, the fact that LAC is only piggybacked on writes (and written finally on close/recovery...) is very annoying for tailing readers.
The best would be to enable ExplicitLAC by default but as far as I can see it is not fully integrated in the normal flow of a tailing reader (you have to explicitly read ExplicitLAC from the reader side).

In my opinion it is better for us to keep ExplicitLAC RPC as it is now.
We have choosen to keep force RPC away from ledger storage (no masterKey on wireprotocol), but in the future we can add an optional payload with masterKey and current client-side LAC.
We could make force() to really write a meta entry which contains the LAC and integrate this in the story and so ExplicitLAC won't be necessary anymore if we perform force() in background, even for non-DEFERRED_SYNC writers.

So a recap for this PR and my view of the story after BP14:
- the story we are writing is healthy and it does not break existing assumptions
- it opens the door to future enhancements, and maybe to the deprecation of ExplicitLAC, in favour of standard LAC (we are introducing a mechanism to let LAC advance on writer side without "writes")

I will be happy to continue personally this work and make BK more like a storage system

cc @jvrao

eolivelli:
retest this please

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/DistributionSchedule.java
```diff
@@ -149,6 +149,13 @@ public WriteSet copy() {
      */
     WriteSet getWriteSetForLongPoll(long entryId);
 
+    /**
+     * Return the set of bookies indices to send the messages to for force ledger.
+     *
+     * @return the set of bookies indices to send force request to.
+     */
+    WriteSet getWriteSetForForceLedger();
```
sijie: This will become unmanageable. We ended up adding getWriteSetFor for different type of requests. I would suggest: renaming `getWriteSetForLongPoll` to `getEnsembleWriteSet`, because those requests are required to send to the whole ensemble.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/DistributionSchedule.java
```diff
@@ -197,6 +204,11 @@ public WriteSet copy() {
      */
     AckSet getAckSet();
 
+    /**
+     * Returns an ackset object useful to wait for all bookies in the ensemble,
+     * responses should be checked against this.
+     */
+    AckSet getAckSetForForceLedger();
```
sijie: `getEnsembleAckSet(long entryId)`. same reason as `getEnsembleWriteSet`.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ForceLedgerOp.java
```diff
@@ -0,0 +1,123 @@
...
+
+        // We got response.
+        receivedResponseSet.clear(bookieIndex);
+
+        if (rc == BKException.Code.OK) {
+            if (ackSet.completeBookieAndCheck(bookieIndex) && !completed) {
+                completed = true;
+                // we are able to say that every bookie sync'd its own journal
+                // for every ackknowledged entry before issuing the force() call
+                LOG.debug("After force on ledger {} updating LastAddConfirmed to {} ",
```
sijie: `if (LOG.isDebugEnabled())`

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ForceLedgerOp.java
```diff
@@ -0,0 +1,123 @@
...
+                        ledgerId, currentNonDurableLastAddConfirmed);
+                lh.lastAddConfirmed = currentNonDurableLastAddConfirmed;
+                FutureUtils.complete(cb, null);
+                return;
+            }
+        } else {
+            LOG.warn("ForceLedger did not succeed: Ledger {} on {}", new Object[]{ledgerId, addr});
```
sijie: - I dont think we need `new Object[]`
- do we really need this `warn` message? It can be very noisy.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ForceLedgerOp.java
```diff
@@ -0,0 +1,123 @@
...
+        receivedResponseSet.clear(bookieIndex);
+
+        if (rc == BKException.Code.OK) {
+            if (ackSet.completeBookieAndCheck(bookieIndex) && !completed) {
+                completed = true;
+                // we are able to say that every bookie sync'd its own journal
+                // for every ackknowledged entry before issuing the force() call
+                LOG.debug("After force on ledger {} updating LastAddConfirmed to {} ",
+                        ledgerId, currentNonDurableLastAddConfirmed);
+                lh.lastAddConfirmed = currentNonDurableLastAddConfirmed;
```
sijie: use `LedgerHandle#updateLastConfirmed` - it is a safer method to use rather than just directly mutating here.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -1100,6 +1100,68 @@ public void asyncAddEntry(final long entryId, final byte[] data, final int offse
...
+            // updating lastAddPushed
+            if (metadata.isClosed()) {
+                wasClosed = true;
+            }
+        }
+
+        if (wasClosed) {
+            // make sure the callback is triggered in main worker pool
+            try {
+                bk.getMainWorkerPool().submit(new SafeRunnable() {
```
sijie: better use `executeOrdered`

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -1100,6 +1100,68 @@ public void asyncAddEntry(final long entryId, final byte[] data, final int offse
...
+                });
+            } catch (RejectedExecutionException e) {
+                result.completeExceptionally(new BKException.BKInterruptedException());
+            }
+            return result;
+        }
+
+        // early exit: no write has been issued yet
+        if (pendingAddsSequenceHead == INVALID_ENTRY_ID) {
+            bk.getMainWorkerPool().submit(new SafeRunnable() {
```
sijie: better to use `executeOrdered`

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -1822,6 +1892,17 @@ void handleBookieFailure(final Map<Integer, BookieSocketAddress> failedBookies)
             return;
         }
 
+        if (writeFlags.contains(WriteFlag.DEFERRED_SYNC)) {
+            blockAddCompletions.decrementAndGet();
+            if (LOG.isDebugEnabled()) {
```
sijie: The comment isn't aligned with the function call. Are you retrying or not?

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieDeferredSyncTest.java
```diff
@@ -0,0 +1,197 @@
...
+ */
+public class BookieDeferredSyncTest extends BookKeeperClusterTestCase {
+
+    public BookieDeferredSyncTest() {
+        super(1);
+    }
+
+    @Test
+    public void testWriteAndRecovery() throws Exception {
+        WriteHandle lh = result(bkc.newCreateLedgerOp()
```
sijie: `@Cleanup` or in a try block

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieDeferredSyncTest.java
```diff
@@ -0,0 +1,197 @@
...
+                .withEnsembleSize(1)
+                .withWriteQuorumSize(1)
+                .withAckQuorumSize(1)
+                .withWriteFlags(WriteFlag.DEFERRED_SYNC)
+                .withDigestType(org.apache.bookkeeper.client.api.DigestType.CRC32C)
```
sijie: nit: why are you using fully qualified name here?

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieDeferredSyncTest.java
```diff
@@ -0,0 +1,197 @@
...
+
+        try (ReadHandle readLh = result(bkc.newOpenLedgerOp()
+                .withLedgerId(ledgerId)
+                .withRecovery(true)
+                .withPassword(new byte[0])
+                .execute());) {
+
+            try (LedgerEntries entries = readLh.read(0, n - 1)) {
+                for (int i = 0; i < n; i++) {
+                    org.apache.bookkeeper.client.api.LedgerEntry entry = entries.getEntry(i);
```
sijie: nit: why are you using fully qualified name here?

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieDeferredSyncTest.java
```diff
@@ -0,0 +1,197 @@
...
+                .withDigestType(org.apache.bookkeeper.client.api.DigestType.CRC32C)
+                .withPassword(new byte[0])
+                .execute());
+
+        int n = 10;
+
+        long ledgerId = lh.getId();
+
+        for (int i = 0; i < n; i++) {
+            lh.append(("entry-" + i).getBytes());
```
sijie: nit: `getBytes(UTF_8)`

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieDeferredSyncTest.java
```diff
@@ -0,0 +1,197 @@
...
+        }
+
+        restartBookies();
```
sijie: what exactly this test is testing?

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieDeferredSyncTest.java
```diff
@@ -0,0 +1,197 @@
...
+
+        for (int i = 0; i < n; i++) {
+            lh.append(("entry-" + i).getBytes());
+        }
+
+        // this will close metadata, writing LastAddConfirmed = -1
+        assertEquals(-1, lh.getLastAddConfirmed());
+        lh.close();
+
+        restartBookies();
```
sijie: why we restart here?

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieDeferredSyncTest.java
```diff
@@ -0,0 +1,197 @@
...
+                for (int i = 0; i < n; i++) {
+                    org.apache.bookkeeper.client.api.LedgerEntry entry = entries.getEntry(i);
+                    assertEquals("entry-" + i, new String(entry.getEntryBytes()));
+                }
+            }
+        }
+    }
+
+    @Test
+    public void testCloseNoForce() throws Exception {
```
sijie: most of the code is duplicated with `testCloseWithForce`. please consolidate these into one common function `testClose(boolean withForce)`.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieDeferredSyncTest.java
```diff
@@ -0,0 +1,197 @@
...
+                for (int i = 0; i < n; i++) {
+                    org.apache.bookkeeper.client.api.LedgerEntry entry = entries.getEntry(i);
+                    assertEquals("entry-" + i, new String(entry.getEntryBytes()));
+                }
+            }
+        }
+    }
+
+    @Test
+    public void testForceRoundTripWithDeferredSync() throws Exception {
```
sijie: `RoundTrip`, what does that mean?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/DistributionSchedule.java
jvrao: I like that idea; Everything that needs to get whole ensemble set can have one API. maybe call it `getEnsembleSet`, no need to have **write** in it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ForceLedgerOp.java
```diff
@@ -0,0 +1,123 @@
...
+        }
+    }
+
+    @Override
+    public void forceLedgerComplete(int rc, long ledgerId, BookieSocketAddress addr, Object ctx) {
+        int bookieIndex = (Integer) ctx;
+
+        if (completed) {
+            return;
+        }
```
jvrao: Given that we are sending force() to all bookies in the ensemble and we won't be completed until we receive a response from all bookies, will we ever enter this condition? Maybe we need to assert here instead of just a return??

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ForceLedgerOp.java
```diff
@@ -0,0 +1,123 @@
...
+                        ledgerId, currentNonDurableLastAddConfirmed);
+                lh.lastAddConfirmed = currentNonDurableLastAddConfirmed;
+                FutureUtils.complete(cb, null);
+                return;
+            }
+        } else {
+            LOG.warn("ForceLedger did not succeed: Ledger {} on {}", new Object[]{ledgerId, addr});
+        }
+
+        if (receivedResponseSet.isEmpty()) {
```
jvrao: Ok; I am little confused with the logic here. We need to have the following logic.
1. Success case: we set `completed` flag when all bookies send  `Code.OK` and `receivedResponseSet.isEmpty`.
2. Failure case:  Set `completed` flag on the first error and ignore rest of the responses. 
Why do you need to wait until all responses in the failure case?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
jvrao: we are not retrying; so comment needs to be changed.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
jvrao: I don't think this check makes sense. This may not cause harm as `failedBookies` set is best effort scenario.
@sijie do you think we need this check here?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -1100,6 +1100,68 @@ public void asyncAddEntry(final long entryId, final byte[] data, final int offse
...
+        }
+
+        if (wasClosed) {
+            // make sure the callback is triggered in main worker pool
+            try {
+                bk.getMainWorkerPool().submit(new SafeRunnable() {
+                    @Override
+                    public void safeRun() {
+                        LOG.warn("Force() attempted on a closed ledger: {}", ledgerId);
+                        result.completeExceptionally(new BKException.BKLedgerClosedException());
```
jvrao: Should we do an implicit force() on close()? We are failing the force() if the ledger is closed, but the close() API doesn't appear to guarantee the force(). So what happens to applications that failed to force() before close()?

eolivelli: We have discussed about this point. Final answer was to leave this up to the application.
Filesystems usually do to fsync on close.

I have a test case for this scenario. If you do not perform force() LAC does not advance and metadata are sealed with the current LAC, which could be -1 if no force() has been issued and completed with success. This is consistent with what readers had seen, that is a LAC which never has been updated.
I see this is not obvious and should be clearly stated in high level docs about BP14, not only in javadocs

eolivelli: We have discussed about this point. Final answer was to leave this up to the application.
Filesystems usually do to fsync on close.

I have a test case for this scenario. If you do not perform force() LAC does not advance and metadata are sealed with the current LAC, which could be -1 if no force() has been issued and completed with success. This is consistent with what readers had seen, that is a LAC which never has been updated.
I see this is not obvious and should be clearly stated in high level docs about BP14, not only in javadocs

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
eolivelli: removed

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieClient.java
```diff
@@ -193,6 +194,30 @@ public PerChannelBookieClientPool lookupClient(BookieSocketAddress addr) {
         return clientPool;
     }
 
+    public void forceLedger(final BookieSocketAddress addr, final long ledgerId,
+            final ForceLedgerCallback cb, final Object ctx) {
+        final PerChannelBookieClientPool client = lookupClient(addr);
+        if (client == null) {
+            cb.forceLedgerComplete(getRc(BKException.Code.BookieHandleNotAvailableException),
+                              ledgerId, addr, null);
```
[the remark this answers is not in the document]
eolivelli: good catch. done

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/DistributionSchedule.java
eolivelli: done.

I will use currentNonDurableLastAddConfirmed as reference entryId (prev getWriteSetForLongPoll required an entry id)
for DEFERRED_SYNC writers ensemble change is disabled, so it is not very important the entryId, but I think this is consistent with the fact that we want the ensemble at the current point in the history of the ledger

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/DistributionSchedule.java
eolivelli: seems to me that entryId is not neeeded in this case, I will rename to getEnsembleActSet

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ForceLedgerOp.java
eolivelli: done

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ForceLedgerOp.java
eolivelli: good catch. Moving to info and removing Object[]

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieDeferredSyncTest.java
eolivelli: done

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieDeferredSyncTest.java
eolivelli: done

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieDeferredSyncTest.java
eolivelli: renamed and consolidated in only one method.
I meant the full trip from client to bookie and back again to the client, that is testing the full network stack for force RPC.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ForceLedgerOp.java
eolivelli: good catch

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ForceLedgerOp.java
eolivelli: good idea. reworked

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
eolivelli: done

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieDeferredSyncTest.java
eolivelli: I will add a comment, I am not going to "close" this handle because I want recovery to recover LAC, I don't want it to be written into metadata

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieDeferredSyncTest.java
eolivelli: IDE "fix import" problem. fixed

eolivelli:
@jvrao @sijie 
All of your comments addressed. Now it looks really better ! Thank you

Additional implementation to note:
**Error out** all pending adds in case of failure of **force()**

eolivelli:
@sijie  @jvrao 
waiting for your comments and hopefully approval.

I have found an edge case, but I don't know if it is worth handling it:
- we are erroring out pending adds in case of failure of a force
- we are NOT erroring out pending "force" operations in case of write failures

I this that anyway if any add fails, as we do not allow ensemble changes, the "force" will eventually fail

sijie:
> I have found an edge case, but I don't know if it is worth handling it.

I would suggest not to error out pending adds. a force failure just means not advancing lac. a force failure doesn't have any side effect to current state of a ledger handle.

eolivelli:
@sijie in the fsync / force metaphor I think it is better not to fail pending adds. 
It was a comment/request from @jvrao.

I am going to drop that op, but I need to ask to JV

sijie:
>  in the fsync / force metaphor I think it is better not to fail pending adds.

sorry your comment isn't clear to you. are you saying JV asked for failing pending adds when encountered issues on #force? or `not failing`?

eolivelli:
I meant @jvrao suggested to error our pending adds.
I think we do not need to do so.

I will remove that from the patch, but before spending time I would like to have acknowledgement feom JV

eolivelli:
@sijie  @jvrao FYI I have added a commit to drop 'error out pending adds" in case of failure in force operation.

So we are *not* erroring out pending adds, they will fail naturally in case of bookie failure.
a force() which fails simply does not advance LastAddConfirmed.

Hopefully @sijie  and @jiazhai will be still okay with the patch

jvrao:
@eolivelli  thanks a lot for pushing this change forward. Great work.!!

eolivelli:
@jvrao @sijie @jiazhai 
Thank you !

I have rebased to latest master, I will wait for CI to pass and then merge.
This was the last important part of BP-14 !
I will follow up with docs

I think existing BC tests are enough to guarantee that we did not break compatibility.

Now that BP-14 has got its final form I will start tests and benchmarks on downstream applications (once this change reaches Apache Snapshots Repository)

eolivelli:

eolivelli:

eolivelli:

eolivelli:

PR #1437, opened 2018-05-24
Use maven-exec-plugin to get the maven project version for CI jobs

sijie:

*Motivation*

The current approach to get project version isn't reliable on CI environment. It might have garbage output (e.g. exceptions).

*Solution*

Changed to use `exec-maven-plugin` to echo `${project.vesion}`. It provides a more reliable approach to get version on CI environments.

### DIFF
--- a/dev/common.sh
+++ b/dev/common.sh
@@ -21,7 +21,11 @@
 #
 
 function get_bk_version() {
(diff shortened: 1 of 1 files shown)


sijie:
IGNORE IT CI

PR #1441, opened 2018-05-25
Update bookkeeper dependencies : netty/protobuf/grpc

sijie:

*Motivation*

The netty version in bookkeeper is a bit dated than the version in pulsar.
Hence the grpc version is limited to the version that use same netty version.
It causes issues at pulsar using bookkeeper, because pulsar has been using
a newer version for longer time. This causes conflicts when pulsar client and
grpc are used together without proper shading.

*Solution*

Upgrade netty/protobuf/grpc dependencies.

- update netty from `4.1.12` to `4.1.22`, which is the one grpc is using and closer to the one that pulsar is using.
- update grpc from `1.5.0` to `1.12.0`, which is using netty `4.1.22` and protobuf `3.5.1`.
- update protobuf from `3.4.0` to `3.5.1`

Related Issue: apache/pulsar#1844


sijie:
@merlimat please review

sijie:
IGNORE IT CI

PR #1446, opened 2018-05-26
Add printreplicationworkerid option to ListUnderreplicatedCmd.

reddycharan:

Add printreplicationworkerid option to ListUnderreplicatedCmd.
This helps in knowing who is holding lock on UnderReplicated ledger,
and diagnosing that ReplicationWorker if we are seeing any issues in
rereplication.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
@@ -104,6 +104,7 @@ import org.apache.bookkeeper.proto.BookkeeperInternalCallbacks.GenericCallback;
 import org.apache.bookkeeper.replication.AuditorElector;
 import org.apache.bookkeeper.replication.ReplicationException;
 import org.apache.bookkeeper.replication.ReplicationException.CompatibilityException;
+import org.apache.bookkeeper.replication.ReplicationException.UnavailableException;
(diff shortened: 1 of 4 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
```diff
@@ -927,12 +930,26 @@ int runCmd(CommandLine cmdLine) throws Exception {
...
+                        urLedgerMapEntry.getValue().forEach((missingReplica) -> {
+                            System.out.println("\tMissingReplica : " + missingReplica);
                         });
                     }
+                    if (printReplicationWorkerId) {
+                        try {
+                            String replicationWorkerId = underreplicationManager
+                                    .getReplicationWorkerIdRereplicatingLedger(urLedgerId);
+                            if (replicationWorkerId != null) {
+                                System.out.println("\tReplicationWorkerId : " + replicationWorkerId);
```
eolivelli: Would it be useful to print something even in case of missing worker

reddycharan: i think it would be noisy. At a given point, the number of locks acquired would be roughly equal to number of ReplicationWorkers in the cluster. So for the rest of the under-replicated ledgers this value would be null and if we print something in case lock is not yet acquired for the under replicated ledger then it would become verbose and noisy.

reddycharan:
retest this please

PR #1430, opened 2018-05-24
[table service] client interceptor and storage container grpc proxy

sijie:
Descriptions of the changes in this PR:

This is a subsequent change after apache/bookkeeper#1428. 

*Motivation*

Current almost every grpc requests are wrapped into `StorageContainerRequest` and their responses
are wrapped into `StorageContainerResponse`. It makes things a bit complicated on adding new grpc
services.

*Changes*

To simplify things, this PR introduces two functionalities for simplifying dispatching container requests/responses.

1) *StorageContainerClientInterceptor*: A grpc `ClientInterceptor` that stamps container information (currently is `scId`) into the requests' metadata before sending the requests to the wire.

2) A simple grpc reverse proxy to dispatch grpc requests to the channels provided by a `ChannelFinder`.

*Tests*

1. Existing unit tests covered client interceptor changes.
2. Introduced a `stream-storage-tests-common` module to include common classes that would be used for testing.
3. Introduced a `PingPongService` for testing reverse proxy : unary/client-streaming/server-streaming/bidi-streaming.

Master Issue: #1205

PR #1448, opened 2018-05-28
[table service] use grpc reverse proxy to serve storage container requests

sijie:

*Motivation*

#1430 introduced client interceptor and grpc reverse proxy for storage container requests. This PR is the subsequent change to leverage client interceptor and reverse proxy for storage container requests.

*Changes*

**Client Changes**

Changed `StorageContainerChannel` to add client interceptor to stamp storage container id into the the requests' metadata.

**Server Changes**

- moved the stream storage logic out of storage containers to a `service` package. the main logic will be kept in `RangeStoreService` and `RangeStoreServiceFactory`.
- make the storage container logic for hosting `StorageContainerService`.
- Each storage container will run an inprogress grpc server for serving the grpc services registered to the container and provide an `channel` for accessing those grpc service.
- Changed the server to use reverse proxy for serving/proxying storage container requests.

*NOTE*

This change doesn't directly remove `StorageContainerRequest` and `StorageContainerResponse`.  It would be done in a subsequent change.

Master Issue: #1205


stream/storage/api/src/main/java/org/apache/bookkeeper/stream/storage/api/metadata/RangeStoreService.java
```diff
@@ -22,4 +23,9 @@
     extends MetaRangeStore,
     RootRangeStore,
     TableStore {
```
jiazhai: how about also add this in the interface comments in line 20?

sijie: I think data includes "table".

sijie:
retest this please

PR #1451, opened 2018-05-29
Update release procedure to use candidate tags for voting

sijie:

*Motivation*

We used final release tag during release voting procedure. However there might be changes
happen between different candidate votes. So we ended up retagging the release tag for
different votes.

*Solution*

Use a candidate tag for voting. After a release is approved, create the final release tag.

### DIFF
--- a/site/community/release_guide.md
+++ b/site/community/release_guide.md
@@ -263,6 +263,7 @@ Set up a few environment variables to simplify Maven commands that follow. This 
     RC_NUM="0"
     TAG="release-${VERSION}"
     RC_DIR="bookkeeper-${VERSION}-rc${RC_NUM}"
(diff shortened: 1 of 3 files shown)


site/community/release_guide.md
```diff
@@ -445,6 +446,24 @@ Copy the source release from the `dev` repository to the `release` repository at
...
+### Git tag
+
+Create and push a new signed for the released version by copying the tag for the final release tag, as follows
+
+```shell
+git tag -s "${TAG}" "${RC_TAG}"
+git push apache "${TAG}"
+```
+
+Remove rc tags:
```
ivankelly: I would leave them. It can sometimes be useful to see the diffs between candidates.

PR #1457, opened 2018-05-30
[table service] cleanup : cleanup protobuf definitions 

sijie:

*Motivation*

There are some unused fields and some fields that were not renamed correctly from col to namespace.

*Solution*

This is a cleanup PR to remove unused fields and rename "col" to "ns" or "namespace".

### DIFF
--- a/stream/proto/src/main/proto/stream.proto
+++ b/stream/proto/src/main/proto/stream.proto
@@ -89,7 +89,7 @@ message ParentRangesList {
 // from the beginning, we shared the metadata management
 // between streams and tables and distinguish them using
 // a flag that recorded in metadata.
-enum StreamType {
+enum StorageType {
     STREAM      = 0;
     TABLE       = 1;
 }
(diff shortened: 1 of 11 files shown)


stream/proto/src/main/proto/stream.proto
```diff
@@ -85,6 +85,15 @@ message ParentRangesList {
 // Stream
 //
 
+// since stream and table are similar and exchangable,
+// from the beginning, we shared the metadata management
+// between streams and tables and distinguish them using
+// a flag that recorded in metadata.
+enum StreamType {
```
eolivelli: Streamtype = table seems a bit weird.

What about something like StorageMode, DataModel....

sijie: How about StorageType?

eolivelli: Okay. Sounds good.

sijie: done

sijie:
retest this please

sijie:
retest this please

sijie:
retest this please

PR #1459, opened 2018-05-31
[table service] cleanup : provide unified service uri for resolving service endpoints

sijie:

*Motivation*

Currently there are multiple settings in `StorageClientSettings` on configuring how to resolve
service endpoints to find the services. It is a bit inconvinient and confusing

*Solution*

Provide a class `ServiceURI` for parsing service uri to retrieve common informations including:

- serviceName
- serviceInfos
- serviceUser
- serviceHosts
- servicePath

It extends the java.net.URI and does parsing and validation in the same place. Then it can be used
by the grpc name resolver to resolve service endpoints.

Also update dlog to use it for resolving its namespace uri.


bookkeeper-common/src/main/java/org/apache/bookkeeper/common/net/ServiceURI.java
```diff
@@ -0,0 +1,235 @@
...
+ * {@code zk+hierarchical://localhost:2181/ledgers} =&gt; ledger service uri using hierarchical ledger manager<br>
+ * {@code etcd://localhost/ledgers} =&gt; ledger service uri using etcd as metadata store<br>
+ * {@code distributedlog://localhost:2181/distributedlog} =&gt; distributedlog namespace<br>
+ * {@code distributedlog-bk://localhost:2181/distributedlog} =&gt; distributedlog namespace with bk backend<br>
+ * {@code bk://bookkeeper-cluster/} =&gt; stream storage service uri <br>
+ * {@code host1:port,host2:port} =&gt; a list of hosts as bootstrap hosts to a stream storage cluster}
+ * </blockquote>
+ *
+ * @since 4.8.0
+ */
```
jiazhai: 👍 great comments format.

stream/clients/java/base/src/main/java/org/apache/bookkeeper/clients/utils/GrpcChannels.java
```diff
@@ -0,0 +1,64 @@
...
+    private GrpcChannels() {}
+
+    /**
+     * Create a channel builder from <tt>serviceUri</tt> with client <tt>settings</tt>.
+     *
+     * @param serviceUri service uri
+     * @param settings client settings
+     * @return managed channel builder
+     */
+    public static ManagedChannelBuilder createChannelBuilder(String serviceUri,
```
jiazhai: nit: is there a need to passed in parameter with type of `ServiceURI` instead of `String`?  so the IllegalArgumentException will be catch earlier.

sijie: good point. I think `StorageClientSettings` validates the string on `#build`. so the string used by the internal implementation, should already kind of validate. so it is okay to have a string here, and having a string is make the caller simpler

PR #1470, opened 2018-06-03
Support build on JDK10

eolivelli:
- switch to "javac -h" in circe-checksum instead of 'javah' because in JDK10 javah has been dropped (see http://openjdk.java.net/jeps/313)
- add openjdk10 to Travis-CI


sijie:
retest this please

PR #1486, opened 2018-06-06
[table service] dlog configuration settings are not propagated

sijie:

*Motivation*

When stream storage server starts the bookie watcher service, it waits for at least `ensembleSize` bookies
available before starting other components. However the dlog settings are not propagated correctly. so when
using this component in an environment that only has 1 bookie (e.g. pulsar standalone). Startup will be hanging
on waiting 3 bookies.

*Solution*

load the dlog settings from the base configuration.


sijie:
run integration tests

sijie:
rebuild java9

Issue #1489, opened 2018-06-06
Better Prevent Read Outliers during short-term Bookie Slow-Down


nicmichael:
**BUG REPORT**

Issue #709 introduced a feature to put bookies which didn't respond to a read request for speculativeReadTimeout ms (default: 2000 ms) on a "slowBookie" list, on which they remain for bookieFailureHistoryExpirationMSec ms (default: 60000). During this period, bookies on the slowBookie list will be put behind any non-slow bookie in "writeSet" that determines the read order. As a consequence, if a bookie has been slow just once, it will not be considered for further reads for 60 seconds. This pushes more load (potentially unnecessarily) onto other bookies in the ensemble. In some cases, this may overload the other bookies, which might then as a result hit the speculativeReadTimeout as well. In some cases, I've seen all bookies end up on the slowBookie list though none of them had any longer lasting performance issues. At that point, the read order is decided based on the number of outstanding requests of those bookies. 

This might work well if bookies are slow for an extended period of time, but not if they only face a brief hickup. For example, we're setting speculativeReadTimeout = 30 to improve read latency and prevent outliers in case of slow bookies, for example due to Java GC in bookies (on the order of tens of ms) or any other short hickups bookies they may face. The current implementation does not address these needs adequately since it only detects slow bookies after speculativeReadTimeout ms (but not while a queue is building up on a bookie), and also puts bookies that have once been slow on the slowBookie list needlessly long without detecting whether or not the problem (e.g. Java GC) has been resolved.

We should have a feature that quicker reacts to short latency spikes or blocking of individual bookies, and quickly resumes directing load to them once the situation is resolved.

nicmichael:
The plot below shows the behaviour of the current implementation with RackawareEnsemblePlacementPolicy for a scenario with 3 bookies and a speculativeReadTimeout of 30 ms. Ticks on the x axis are minutes. 

Plot #1:
![analysis_slowbookies_zoom2b](https://user-images.githubusercontent.com/27270046/41051957-66178628-696c-11e8-88ce-952ac49d04a0.png)

[...]

Issue #709, opened 2017-11-08
Add Slow Bookkeeper Servers to Placement Policy for read ordering


philipsu522:
Feature Request

1. Please describe the feature you are requesting.
The bookkeeper client currently only differentiates between bookkeeper hosts that are available, readonly or unavailable when reading. However, clients will continue encountering lag if a single bookkeeper host in the quorum is slow in serving requests.

The reason is that the bookkeeper client picks bookkeeper hosts to read from in a round robin fashion. Therefore, if the read quorum is 3 and one of the 3 hosts are slow, then every third read will be served by a slow host, which results in lag building up.

The proposed solution for this is to maintain a list of slow bookkeeper hosts, which are tried only after readonly bookkeeper hosts are tried. Bookkeeper hosts can be categorized as "slow" when a speculative timeout occurs on that bookkeeper host.

2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have). Are you currently using any workarounds to address this issue?
should-have

PR #1504, opened 2018-06-08
Issue #1489: Better Prevent Read Outliers during short-term Bookie Slow-Down

nicmichael:
### Motivation

Bookies can temporarily be slow for a large number of reasons, often for just a brief time of few milliseconds to seconds such as during Java Garbage Collection or EntryLog compaction. For writes, latencies of individual bookies are masked by acknowledging the client after a quorum of bookies have replied. However for reads, we don't have any equivalent feature to mask short-term latencies of individual bookies yet (in case of SequenceReadRequests). This PR implements such a feature by reordering reads to prefer bookies with a high probability of being fast over bookies that are potentially slow.

### Changes
This change implements a configurable reordering of read requests in Bokkeeper client based on the number of pending requests to each bookie that could service the request. The intention is to mask the latency of one bookie by directing a read request to another bookie that could potentially service the request faster. This should help prevent read time outliers due to bookies that temporarily are responsing slow, for example due to Java garbage collection, compaction, or any other kind of hickup. Unlike the implementation for Issue #709, this algorithm quickly reacts to both an or decrease increase in queue length of a bookie relative to others, and allows to redirect requests long before they would hit the speculativeReadTimeout. Once the problem is resolved (e.g. Java GC finished), it will quickly direct requests to the previously "slow" bookie as its queue length decreases.

Reordering of reads is based on a threshold of relative queue length to other bookies. Setting the threshold very low will more frequently reorder the read set and potentially result in better latency, but will also reduce data affinity of reads. Reads send to other than the preferred bookie have a low chance to be served from file system cache on that bookie, and will likely result in a physical read. Small thresholds therefore shuffle read requests more among bookies and may lead to reduced file system cache reach and increased physical reads on disks. A larger timeout will maintain data affinity and avoid above problems, but only kick in once a bookie has built-up a considerable queue of requests. It therefore masks only slightly larger outliers, and leads to overall better efficiency.

Master Issue: #1489

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ClientConfiguration.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ClientConfiguration.java
@@ -149,6 +149,7 @@ public class ClientConfiguration extends AbstractConfiguration<ClientConfigurati
     // Ensemble Placement Policy
     protected static final String ENSEMBLE_PLACEMENT_POLICY = "ensemblePlacementPolicy";
     protected static final String NETWORK_TOPOLOGY_STABILIZE_PERIOD_SECONDS = "networkTopologyStabilizePeriodSeconds";
+    protected static final String READ_REORDER_THRESHOLD_PENDING_REQUESTS = "readReorderThresholdPendingRequests";
     protected static final String ENSEMBLE_PLACEMENT_POLICY_ORDER_SLOW_BOOKIES =
         "ensemblePlacementPolicyOrderSlowBookies";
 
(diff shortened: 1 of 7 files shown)


sijie:
@philipsu522 @yzang can you guys help reviewing this? this is related to handling slow bookies. you guys also have better experiences/context than me :)

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ClientConfiguration.java
```diff
@@ -1126,6 +1127,30 @@ public ClientConfiguration setEnsemblePlacementPolicy(Class<? extends EnsemblePl
         return this;
     }
 
+    /**
+     * Get the threshold for the number of pending requests beyond which to reorder
+     * reads. If <= zero, this feature is turned off.
```
eolivelli: Nit: use &lt; instead of '<'. This is HTML

philipsu522:
awesome - LGTM!

sijie:

sijie:
run bookkeeper-server bookie tests // BookieJournalTest is flaky due to #1508

sijie:
run integration tests

sijie:
run integration tests

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ClientConfiguration.java
nicmichael: Thanks for pointing out. Fixed in rev 2.

Issue #1490, opened 2018-06-06
BookieJournalForceTest is flaky


sijie:
*Problem*

```
2018-06-06\T\08:03:29.167 [ERROR] testForceLedger(org.apache.bookkeeper.bookie.BookieJournalForceTest)  Time elapsed: 0.263 s  <<< FAILURE!
java.lang.AssertionError: expected:<0> but was:<1>
	at org.apache.bookkeeper.bookie.BookieJournalForceTest.testForceLedger(BookieJournalForceTest.java:367)

```

sijie:
@eolivelli I found that `BookieJournalForceTest` is flaky. Can you take a look at this? I would be surprised about this, since this unit test is mock based.

eolivelli:
Sure. Will do soon. It is very strange.

sijie:
yeah. there might be some non-deterministic behavior in the test. but it is seen multiple times in the "bookie tests" in recent PRs.

eolivelli:
There are multiple threads which are playing in the game. I will debug as soon as I have time this week. We must not have flakes especially on new tests

PR #1508, opened 2018-06-10
ISSUE-1490 BookieJournalForceTest is flaky

eolivelli:

### Motivation
The test if flaky because the Journal is putting some ForceWriteRequest in the internal queue. This test controls the journal and the queue of ForceWriteRequests, but in the end this control is lost and the final assertion can fail if the journal 'works'.
My guess is about the flag setJournalFlushWhenQueueEmpty which is turned by default to true in tests in order t ospped up the execution.

### Changes

- add debug
- add toString to ForceWriteRequest
- add setJournalFlushWhenQueueEmpty(false) to the flaky test

Master Issue: #1490

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
@@ -22,6 +22,7 @@
 package org.apache.bookkeeper.bookie;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.MoreObjects;
 import com.google.common.base.Stopwatch;
 import com.google.common.util.concurrent.MoreExecutors;
 
@@ -392,12 +393,13 @@ public class Journal extends BookieCriticalThread implements CheckpointSource {
 
         @Override
         public String toString() {
-            return "ForceWriteRequest{"
-                    + "forceWriteWaiters=" + forceWriteWaiters
-                    + ", shouldClose=" + shouldClose
-                    + ", isMarker=" + isMarker
-                    + ", lastFlushedPosition=" + lastFlushedPosition
-                    + ", logId=" + logId + '}';
+            return MoreObjects.toStringHelper(ForceWriteRequest.class)
+                .add("forceWriteWaiters", forceWriteWaiters)
+                .add("shouldClose", shouldClose)
+                .add("isMarker", isMarker)
+                .add("lastFlushedPosition", lastFlushedPosition)
+                .add("logId", logId)
+                .toString();
         }
 
         public void closeFileIfNecessary() {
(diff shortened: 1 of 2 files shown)


bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieJournalForceTest.java
```diff
@@ -302,6 +302,8 @@ public void writeComplete(int rc, long ledgerId, long entryId, BookieSocketAddre
         LinkedBlockingQueue<ForceWriteRequest> supportQueue = new LinkedBlockingQueue<>();
         BlockingQueue<ForceWriteRequest> forceWriteRequests = mock(BlockingQueue.class);
         doAnswer((Answer) (InvocationOnMock iom) -> {
+            log.error("something put " + iom.getArgument(0) + " on journal queue",
```
eolivelli: this is the real debug, to see why we have an additional item in the queue

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -390,6 +390,16 @@ public int process(boolean shouldForceWrite) throws IOException {
             }
         }
 
+        @Override
+        public String toString() {
```
sijie: use `StringBuilder`.

or use lomok `ToString` annotation

or use guava `ToStringHelper`

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieJournalForceTest.java
sijie: use `{}`

eolivelli:
Apart from @sijie comments. The test is passed on CI. Let's kick it a couple of times to see if flakiness went away

eolivelli:
rerun bookie tests

eolivelli:

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
eolivelli: done with Guava

eolivelli:
@sijie I have fixed the code. I wrote those lines with the intention to drop them in the future.
Maybe they can stay.

CI started automatically with the push, let's see if the test now is stable

sijie:

sijie:

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieJournalForceTest.java
```diff
@@ -320,6 +322,7 @@ public void testForceLedger() throws Exception {
         Bookie.checkDirectoryStructure(Bookie.getCurrentDirectory(journalDir));
 
         ServerConfiguration conf = TestBKConfiguration.newServerConfiguration();
+        conf.setJournalFlushWhenQueueEmpty(false);
```
sijie: it is still unclear to me how this would cause the flakiness? can you explain more?

sijie:
@eolivelli - I sent a separate PR for resolving this issue - #1512

I think the problem is `adaptiveGroupWrite`, which will insert a marker back to forceWriteQueue, so any validation on forceWriteQueue length can fail since it is not empty. please take a look.

PR #1509, opened 2018-06-10
Issue 1476: LedgerEntry is recycled twice at ReadLastConfirmedAndEntryOp

### LINKED ISSUE #1476: recycle already.
bookkeeper 4.7, when using distributed log api, got the following error:

 [BookKeeperClientWorker-OrderedExecutor-0-0] ERROR org.apache.bookkeeper.common.util.SafeRunnable - Unexpected throwable caught 
java.lang.IllegalStateException: recycled already
	at io.netty.util.Recycler$Stack.pushNow(Recycler.java:560)
	at io.netty.util.Recycler$Stack.push(Recycler.java:550)
	at io.netty.util.Recycler$DefaultHandle.recycle(Recycler.java:209)
	at org.apache.bookkeeper.client.impl.LedgerEntryImpl.recycle(LedgerEntryImpl.java:163)
	at org.apache.bookkeeper.client.impl.LedgerEntryImpl.close(LedgerEntryImpl.java:154)
	at org.apache.bookkeeper.client.ReadLastConfirmedAndEntryOp$ReadLACAndEntryRequest.close(ReadLastConfirmedAndEntryOp.java:94)
	at org.apache.bookkeeper.client.ReadLastConfirmedAndEntryOp.submitCallback(ReadLastConfirmedAndEntryOp.java:530)
	at org.apache.bookkeeper.client.ReadLastConfirmedAndEntryOp.readEntryComplete(ReadLastConfirmedAndEntryOp.java:568)
	at org.apache.bookkeeper.proto.PerChannelBookieClient$ReadCompletion$1.readEntryComplete(PerChannelBookieClient.java:1559)
	at org.apache.bookkeeper.proto.PerChannelBookieClient$ReadCompletion.handleReadResponse(PerChannelBookieClient.java:1640)
	at org.apache.bookkeeper.proto.PerChannelBookieClient$ReadCompletion.handleV3Response(PerChannelBookieClient.java:1615)
	at org.apache.bookkeeper.proto.PerChannelBookieClient$3.safeRun(PerChannelBookieClient.java:1232)
	at org.apache.bookkeeper.common.util.SafeRunnable.run(SafeRunnable.java:36)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at io.netty.util.concurrent.DefaultThreadFactory$DefaultRunnableDecorator.run(DefaultThreadFactory.java:138)
	at java.lang.Thread.run(Thread.java:745)

infodog:

The issue #1476 is caused by peculative reads with object recycling, same request object will be added to the CompletionObjects multiple times with different txnid.  In fact the logic of process the request already take this into account, only on place inside `ReadLastConfirmedAndEntryOp.requestComplete` forget to check requestComplete before calling `submitCallback` which in turn call request.close.

### Motivation

to fix #1476 

### Changes

check `requestComplete` before `submitCallback` in `ReadLastConfirmedAndEntryOp.requestComplete`

Master Issue: #1476 

> ---
>
> ---

> ---
> quickly and easily:
>
>
>
> Otherwise:
> 
> 
> ---

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadLastConfirmedAndEntryOp.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadLastConfirmedAndEntryOp.java
@@ -557,22 +557,28 @@ class ReadLastConfirmedAndEntryOp implements BookkeeperInternalCallbacks.ReadEnt
             hasValidResponse = true;
 
             if (entryId != BookieProtocol.LAST_ADD_CONFIRMED) {
-                buffer.retain();
-                if (request.complete(rCtx.getBookieIndex(), bookie, buffer, entryId)) {
-                    // callback immediately
-                    if (rCtx.getLacUpdateTimestamp().isPresent()) {
-                        long elapsedMicros = TimeUnit.MILLISECONDS.toMicros(System.currentTimeMillis()
-                                - rCtx.getLacUpdateTimestamp().get());
-                        elapsedMicros = Math.max(elapsedMicros, 0);
-                        lh.bk.getReadLacAndEntryRespLogger()
-                            .registerSuccessfulEvent(elapsedMicros, TimeUnit.MICROSECONDS);
-                    }
+                if(requestComplete.get()){
+                    LOG.warn("Request is already completed, then right data come. (lac={}) from bookie({}) for (lid={}).",
+                            rCtx.getLastAddConfirmed(), bookie, ledgerId);
+                }
+                else {
+                    buffer.retain();
+                    if (request.complete(rCtx.getBookieIndex(), bookie, buffer, entryId)) {
+                        // callback immediately
+                        if (rCtx.getLacUpdateTimestamp().isPresent()) {
+                            long elapsedMicros = TimeUnit.MILLISECONDS.toMicros(System.currentTimeMillis()
+                                    - rCtx.getLacUpdateTimestamp().get());
+                            elapsedMicros = Math.max(elapsedMicros, 0);
+                            lh.bk.getReadLacAndEntryRespLogger()
+                                    .registerSuccessfulEvent(elapsedMicros, TimeUnit.MICROSECONDS);
+                        }
 
-                    submitCallback(BKException.Code.OK);
-                    requestComplete.set(true);
-                    heardFromHostsBitSet.set(rCtx.getBookieIndex(), true);
-                } else {
-                    buffer.release();
+                        submitCallback(BKException.Code.OK);
+                        requestComplete.set(true);
+                        heardFromHostsBitSet.set(rCtx.getBookieIndex(), true);
+                    } else {
+                        buffer.release();
+                    }
                 }
             } else {
                 emptyResponsesFromHostsBitSet.set(rCtx.getBookieIndex(), true);
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/util/DiskChecker.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/util/DiskChecker.java
@@ -153,17 +153,17 @@ public class DiskChecker {
             float free = (float) usableSpace / (float) totalSpace;
             float used = 1f - free;
             if (used > diskUsageThreshold) {
-                LOG.error("Space left on device {} : {}, Used space fraction: {} < threshold {}.",
+                LOG.error("Space left on device {} : {}, Used space fraction: {} > threshold {}.",
                         dir, usableSpace, used, diskUsageThreshold);
                 throw new DiskOutOfSpaceException("Space left on device "
-                        + usableSpace + " Used space fraction:" + used + " < threshold " + diskUsageThreshold, used);
+                        + usableSpace + " Used space fraction:" + used + " > threshold " + diskUsageThreshold, used);
             }
             // Warn should be triggered only if disk usage threshold doesn't trigger first.
             if (used > diskUsageWarnThreshold) {
-                LOG.warn("Space left on device {} : {}, Used space fraction: {} < WarnThreshold {}.",
+                LOG.warn("Space left on device {} : {}, Used space fraction: {} > WarnThreshold {}.",
                         dir, usableSpace, used, diskUsageThreshold);
                 throw new DiskWarnThresholdException("Space left on device:"
-                        + usableSpace + " Used space fraction:" + used + " < WarnThreshold:" + diskUsageWarnThreshold,
+                        + usableSpace + " Used space fraction:" + used + " > WarnThreshold:" + diskUsageWarnThreshold,
                         used);
             }
             return used;
--- a/dev/bk-merge-pr.py
+++ b/dev/bk-merge-pr.py
@@ -500,7 +500,7 @@ def is_jenkins_check(check):
     return check["context"].startswith("Jenkins:")
 
 def is_jenkins_passed(url):
-    jenkins_status = get_json("%sapi/json" % (url))
+    jenkins_status = get_json("%sapi/json?tree=result" % (url))
     return "SUCCESS" == jenkins_status['result'] 
 
 def is_integration_test_check(check):
--- a/site/_includes/footer.html
+++ b/site/_includes/footer.html
@@ -4,6 +4,9 @@
       <p>
         Copyright &copy; 2016 - {{ 'now' | date: '%Y' }} <a href="https://www.apache.org/">The Apache Software Foundation</a>,<br /> licensed under the <a href="http://www.apache.org/licenses/LICENSE-2.0">Apache License, version 2.0</a>.
       </p>
+      <p>
+        Apache BookKeeper, BookKeeper®, Apache®, the Apache feature logo, and the Apache BookKeeper logo are either registered trademarks or trademarks of The Apache Software Foundation.
+      </p>
     </div>
   </div>
 </footer>


eolivelli:
Thank you @infodog for looking into this.

How can you prove the fix addresses the problem?
Is there any test case we can add in order ensure wr won't fall into this problem in the future?

infodog:
I setup a bookkeeper cluster with 3 bookie server, and then setup a client read the logs , the client code is as following:
```
void downloadLog() throws IOException {
        long nextTxId = getLocalLastTxId();
        long lastShardTxId = 0;
        try {
            lastShardTxId = dlm.getLastTxId();

        } catch (Throwable t) {
            //当日志为空的时候，dlm.getLastTxId会出异常，所以需要try住
            //do nothing
        }
        LogReader reader = dlm.getInputStream(nextTxId);

        while (true) {
            try {

                while (true) {
                    try {
                        LogRecord record = reader.readNext(false);
                        if (record == null) {
                            break;
                        }
                        System.out.println("read record, txid=" + record.getTransactionId() + ", lastShardId=" + lastShardTxId);
                        if (record.getTransactionId() > getLocalLastTxId()) {
                            updateLocalLog(record);
                        }
                        // read next record
                    } catch (LogEmptyException t) {
                        //没事
                        System.out.println("no record in log.");
                    } catch (Throwable ioe) {
                        // handle the exception
                        nextTxId = getLocalLastTxId();
                        reader = dlm.getInputStream(nextTxId);
                    }

                }


                if (isSwitchingToMaster) {
                    return;
                }
                synchronized (downloadThreadMonitor) {
                    downloadThreadMonitor.wait(100);
                }
                if (isSwitchingToMaster) {
                    return;
                }
            } catch (Throwable e) {
                e.printStackTrace();
            }
        }
    }
```

when I start a writer write to the bookkeeper server, with serveral thoundsand writes, the client will throw the exception, saying that the object recycled problem, then the client will not receive any new data. In fact I start 2 clients in the same time to read from the server. And always one client fails, randomly. 

After apply the fix, my client will run and always get the updated data. So I think the fix addressed the data.

Maybe we can create a test case like this? But I dont know how to do that. 

By the way, When I debug the problem, I add some log statements to the code, and the logs shows there is still some problem hiding. Still shows something I can't understand.  Maybe I am not look carefully enough.

eolivelli:
@infodog I see your description.
It is better to fix the first known issue with your current work, then you will create a follow up issue if this is no enough.

@sijie I don't know DL codebase very much and I am not using long-poll so I can't be very helpful here.
Don't we have some "tailing reader" test case ?
Or maybe existing tests are too narrow that does not make the problem happen.

sijie:
@eolivelli I think the problem only happened when speculations happen. also it is not just on long poll reads, I think it will impact normal reads. those tests around speculations might end before real problems show up.

eolivelli:
So @sijie you are saying that we have some code path not covered by test cases.

It would be good to have minimal coverage of this change, just by using mockito.
How does it sounds to you?

sijie:
@eolivelli - sure, we need to add test cases for this. however I don't think the fix address the root cause. still looking ...

eolivelli:
@sijie  sure, go ahead.

sijie:
@infodog : 

your fix is corrected. however, it can be simplified with changing using `submitCallback` to use `completeRequest`. this would guarantee a request only being completed once.

I created a PR to your branch to improve it and also added a unit test to reproduce the problem and ensure your changes fix the problem: infodog/bookkeeper#1 if you merged it, the changes will be applied to your branch and showed up here.

then we should be ready to merge your fix.

sijie:
@eolivelli - changes in infodog/bookkeeper#1 include the unit test.

@jiazhai @yzang please review this PR as well, since it is dlog related.

infodog:
@sijie I think the fix of  infodog/bookkeeper#1 may have problem.

because 
`request.complete(rCtx.getBookieIndex(), bookie, buffer, entryId) `will be called even the request is already closed.

In request.complete

 ```
entryImpl.setLength(buffer.getLong(DigestManager.METADATA_LENGTH - 8));
 entryImpl.setEntryBuf(content);
```

the entryImpl may already recycled and owned by another request, so this will ruin another request's data?

**Comment by sijie** [comment]
@infodog - https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadLastConfirmedAndEntryOp.java#L127 use an atomic boolean to guarantee it only executed once. so the problem you described won't happen.

**Comment by infodog** [comment]
@sijie  but in https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadLastConfirmedAndEntryOp.java#L617  use `requestComplete` to guarantee request is only close once, while https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadLastConfirmedAndEntryOp.java#L127 use `complete` to guarantee , it's two diff variables. So it's possible after the request is closed at the same time entryImpl is recycled, the https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadLastConfirmedAndEntryOp.java#L127 code will be executed, and the data maybe ruined.

**Comment by sijie** [comment]
@infodog I see what your point. let me try to explain:

1) yes. there are two flags, the `completed` flag in request is to checking if the request itself is completed or not, the `requestComplete` flag is in the ReadLacOp is to checking if the op itself is completed or not. in some sense, we need these two different flags, since one op can potentially contain multiple requests. it might be a bit strange in this class, since in this case, one operation only have one request.

2) are there any race conditions between these two flags? the answer is no. because the request callback can only happen on one thread. so if a request successfully completed with a valid entry, `request.complete` returns true and then it will submit the callback, which will set `requestComplete` to true. so any future callback will not try to complete the request or submit callback again, this prevents the race condition you described.

3) just for your information, in bookkeeper client, the operations/callbacks for same ledger will be executed in same thread.

Hope this explains why the change in infodog/bookkeeper#1 work.

**Comment by infodog** [comment]
@sijie but some times requestComplete come before request.complte, such as the path on https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadLastConfirmedAndEntryOp.java#L584 , this will happen when the response without piggyback data comes before the response with data. 

In this case, the request will first close in requestComplete, the entryImpl will be recycled. Then the request.complete will be called. although this time request.close will not be called, but since entryImpl is already recycled , if entryImpl is resued by other request, then data will be ruined.

**Comment by sijie** [comment]
@infodog you are correct! I will revert infodog/bookkeeper#1 to your original fix, I will update the unit test to ensure verify the case your described.

**Comment by sijie** [comment]
@infodog I've updated infodog/bookkeeper#1 with your original fix and updated the unit test to add validations to make sure recycled entry will not be mutated by any subsequent callbacks. please take a look and let me know if it makes sense to you.

**Comment by infodog** [comment]
@sijie I still have  questions,

1. Although the fix will solve the object recycle problem, but the response with valid data is ignored. Will this case cause  problem to the application  using dl as a replication service? How the application using the dl client shoud do to handle this situation?
Is there any retry mechanism embeded by the bookkeeper client? I think as a quick fix, if this happed the client should throw a exception to the application, so that the application can implement its own retry mechanism. Maybe we should not treat this situation ( the response with valid data come after response withoud data situation) as an error one, when we receive an empty reponse, we should not close the request? Or we should impletement a retry mechanism after this situation happend?

2.  back to the fix detail, if request.complete is not called, then on https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadLastConfirmedAndEntryOp.java#L127

```
writeSet.recycle();
orderedEnsemble.recycle(); 
```
is not executed.
will this cause resource leak? 


3. if requestComplete already set the `buffer.retain();` [here](https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadLastConfirmedAndEntryOp.java#L127) is redundant

**Comment by sijie** [comment]
@infodog good questions, comments inline.

> Will this case cause problem to the application using dl as a replication service?

This will not cause any problems to the application using DL. Here we completed the request only when lac (last add confirmed) is advanced. it is better if there is an entry piggybacked, if there is no entry piggybacked, it is still good, because we received a valid advanced lac.

you might think we dropped a valid response. however the thought behind is to tell the client caller (which is DL library at this case) as soon as LAC is advanced, it can read entries till the new LAC. DL will react to the callback and know LAC is advanced and issue corresponding read requests. 

> How the application using the dl client shoud do to handle this situation?

You don't need to handle this situation. Since DL handles that.

> Is there any retry mechanism embeded by the bookkeeper client?

The conversation between normal reads and long poll reads and knowing LAC advance are all happending in the dl library. applications don't have to worry about that.

> will this cause resource leak?

This will not cause resource leak. The recycle there means putting back the objects to be reused. if recycle is not called, it doesn't mean resource leak, it only means the objects will not be put back to the object pool to be reused. 

the recycle objects are different from bytebuf where releasing bytebuf should be done when the bytebuf is not reused anymore, otherwise it will causing memory can't be released.

> if requestComplete already set the buffer.retain(), here is redundant

I am not sure what exactly you are asking. but I am guessing you are asking the boolean flag in the request. that boolean flag is needed, it is used for flagging whether the request is completed or not. the `requestComplete` is used for checking if the operation is completed (or callback is submitted or not). Those booleans have different meanings. It is not easy to tell the difference from first glance.



Hope this clarifies your questions.

**Comment by sijie** [comment]
@infodog if the comments make sense to you, it would be great if you can merge infodog/bookkeeper#1 to your branch, so it can be showed up here. then we can merge this change once it passes other committers's review and CI checks.

**Comment by infodog** [comment]
@sijie  Ok thanks very much for the explanation, I already merged to my branch.

**Comment by sijie** [comment]
@infodog thank you.

@jiazhai @yzang @eolivelli can you guys review this change?

**Comment by sijie** [comment]

**Comment by sijie** [comment]
rebuild java8

File: bookkeeper-server/src/test/java/org/apache/bookkeeper/client/ReadLastConfirmedAndEntryOpTest.java [hunk 1]
```diff
@@ -0,0 +1,252 @@
...
+        Entry<BookieSocketAddress, ReadLastConfirmedAndEntryHolder> firstBookieEntry = iter.next();
+        ReadLastConfirmedAndEntryHolder firstBookieHolder = firstBookieEntry.getValue();
+        ReadLastConfirmedAndEntryContext firstContext = firstBookieHolder.context;
+        firstContext.setLastAddConfirmed(entryId);
+        firstBookieHolder.getCallback()
+            .readEntryComplete(Code.OK, LEDGERID, BookieProtocol.INVALID_ENTRY_ID, null, firstContext);
+
+        // readEntryComplete above will release the entry impl back to the object pools.
+        // we want to make sure after the entry is recycled, it will not be mutated by any future callbacks.
+        LedgerEntryImpl entry = LedgerEntryImpl.create(LEDGERID, Long.MAX_VALUE);
```
eolivelli: @sijie  @infodog 
are you assuming that here we are getting the same entry as above ?

infodog: @eolivelli 
Yes I assuming that.

eolivelli: Is there any way to perform an assertion about this assumption ?
if that assumption is broken the test is not really useful.

I don't have a suggestion on how to capture the LedgerEntryImpl.

sijie: @eolivelli It is hard to capture the original entry. I mean it is possible, however that's going to make things a bit complicated, which I don't think it is worth doing that.

sijie:

sijie:
rubuild java8

sijie:

sijie:

sijie:
retest this please

sijie:
(somehow the jenkins precommit checks are not running after rebased. trying to close and reopen)

sijie:
run pr validation

sijie:
I think "bookie tests" precommit check is flaky due to #1516 .

there was one successful run - https://builds.apache.org/job/bookkeeper_precommit_bookie_tests/61/

so going to ignore CI to merge this bug fix.

sijie:
IGNORE CI

sijie:
merged the changes. thank you @infodog !

PR #1526, opened 2018-06-19
LedgerManagerIteratorTest: avoid ledger collision

athanatos:
For the non-LHLM managers, the ledger space seems to be small enough to
allow collisions to occasionally cause a test failure.  Remember
created ledgers to avoid collisions.

(@bug W-5104859@)


reddycharan:
LGTM

sijie:
rerun bookkeeper-server tls tests
rerun integration tests

Issue #1527, opened 2018-06-20
Make ExplicitLAC persistent


jvrao:
**FEATURE REQUEST**

ExplicitLAC is kept in the memory and it can be lost in bookie reboot.
Though it is an extreme corner case scenario, it can break one of the BK guarantees. 
" If you read once you can always read it".
If all the bookies of the Write stripe were rebooted, it can loose its explicitLAC value and the client
which was able to read the entry before the reboot, can't read it anymore.

eolivelli:
This will be great.
I hope you already have some idea to share

reddycharan:
@jvrao @sijie @eolivelli 

For persisting explicitLAC, we can follow the same approach as followed in persisting fencing information / stateBits (https://github.com/apache/bookkeeper/commit/d69986c4df6b51ed6ed27c5f45a7b6850e69dacd) in FileInfo file and special marker entry in Journal.

We can have a new marker METAENTRY_ID_LEDGER_EXPLICITLAC = -0x8000 and we can log a special entry in journal (just like fence entry) when we get setExplicitLAC call to the bookie. This special entry could be following 

```
    static ByteBuf createExplicitLACEntry(long ledgerId, long entryId) {
        ByteBuf bb = Unpooled.buffer(8 + 8 + 8);
        bb.writeLong(ledgerId);
        bb.writeLong(METAENTRY_ID_LEDGER_EXPLICITLAC);
        bb.writeLong(entryId);
        return bb;
    }
```

The long explicitLAC value could be stored in the header of the FileInfo after 'stateBits' value, as there is sufficient space remaining in FileInfo allocated header (1024). For explicitLAC, we can follow the same semantics as fence/statebits logic for flushHeader/checkpoint and replay journal .

But there is some small intricacy here regarding 'HEADER_VERSION' of the FileInfo. Since zero is valid ExplicitLac value for a Ledger, for backward compatibility purpose we cannot just consider the read zero value from HEADER after 'statebits' in old FileInfo file, as persisted 'ExplicitLAC'. So we may need to bump up 'HEADER_VERSION' of FileInfo for implementing this feature and handle compatibility scenarios.

jvrao:
Yeah; we need to bump up the version number, and need to handle the case for backward compatibility.

eolivelli:
Some thoughts (thinking out loud):
- I think we should unify ExplicitLAC and PiggyBackLAC, for instnace we should update this new field on FileInfo both on setExplicitLAC and on addEntry
[...]

PR #1532, opened 2018-06-29
ISSUE #1527: Make ExplicitLAC persistent

reddycharan:

For persisting explicitLAC, we can follow the
same approach as followed in persisting
fencing information / stateBits (d69986c)
in FileInfo file and special marker entry
in Journal.

### Motivation

ExplicitLAC is kept in the memory and it can be lost in bookie reboot.
Though it is an extreme corner case scenario, it can break one of the BK guarantees.
" If you read once you can always read it".
If all the bookies of the Write stripe were rebooted, it can loose its explicitLAC value and the client
which was able to read the entry before the reboot, can't read it anymore.

Master Issue: #1527

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
@@ -38,6 +38,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Lists;
 import com.google.common.util.concurrent.SettableFuture;
 import io.netty.buffer.ByteBuf;
+import io.netty.buffer.PooledByteBufAllocator;
 import io.netty.buffer.Unpooled;
 import java.io.File;
 import java.io.FileNotFoundException;
@@ -1202,7 +1203,7 @@ public class Bookie extends BookieCriticalThread {
     }
 
     static ByteBuf createExplicitLACEntry(long ledgerId, ByteBuf explicitLac) {
-        ByteBuf bb = Unpooled.buffer(8 + 8 + 4 + explicitLac.capacity());
+        ByteBuf bb = PooledByteBufAllocator.DEFAULT.directBuffer(8 + 8 + 4 + explicitLac.capacity());
         bb.writeLong(ledgerId);
         bb.writeLong(METAENTRY_ID_LEDGER_EXPLICITLAC);
         bb.writeInt(explicitLac.capacity());
@@ -1216,7 +1217,9 @@ public class Bookie extends BookieCriticalThread {
             long ledgerId = entry.getLong(entry.readerIndex());
             LedgerDescriptor handle = handles.getHandle(ledgerId, masterKey);
             synchronized (handle) {
+                entry.markReaderIndex();
                 handle.setExplicitLac(entry);
+                entry.resetReaderIndex();
                 ByteBuf explicitLACEntry = createExplicitLACEntry(ledgerId, entry);
                 getJournal(ledgerId).logAddEntry(explicitLACEntry, false /* ackBeforeSync */, writeCallback, ctx);
             }
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
@@ -32,6 +32,7 @@ import static org.apache.bookkeeper.bookie.BookKeeperServerStats.LD_INDEX_SCOPE;
 import static org.apache.bookkeeper.bookie.BookKeeperServerStats.LD_LEDGER_SCOPE;
 import static org.apache.bookkeeper.bookie.BookKeeperServerStats.READ_BYTES;
 import static org.apache.bookkeeper.bookie.BookKeeperServerStats.WRITE_BYTES;
+import static org.apache.bookkeeper.bookie.Bookie.METAENTRY_ID_FENCE_KEY;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Lists;
@@ -121,6 +122,7 @@ public class Bookie extends BookieCriticalThread {
     static final long METAENTRY_ID_LEDGER_KEY = -0x1000;
     static final long METAENTRY_ID_FENCE_KEY  = -0x2000;
     public static final long METAENTRY_ID_FORCE_LEDGER  = -0x4000;
+    static final long METAENTRY_ID_LEDGER_EXPLICITLAC  = -0x8000;
 
     private final LedgerDirsManager ledgerDirsManager;
     private LedgerDirsManager indexDirsManager;
@@ -784,6 +786,23 @@ public class Bookie extends BookieCriticalThread {
                                     + " but layout version (" + journalVersion
                                     + ") is too old to hold this");
                         }
+                    } else if (entryId == METAENTRY_ID_LEDGER_EXPLICITLAC) {
+                        if (journalVersion >= JournalChannel.V5) {
+                            int explicitLacBufLength = recBuff.getInt();
+                            ByteBuf explicitLacBuf = Unpooled.buffer(explicitLacBufLength);
+                            byte[] explicitLacBufArray = new byte[explicitLacBufLength];
+                            recBuff.get(explicitLacBufArray);
+                            explicitLacBuf.writeBytes(explicitLacBufArray);
+                            byte[] key = masterKeyCache.get(ledgerId);
+                            if (key == null) {
+                                key = ledgerStorage.readMasterKey(ledgerId);
+                            }
+                            LedgerDescriptor handle = handles.getHandle(ledgerId, key);
+                            handle.setExplicitLac(explicitLacBuf);
+                        } else {
+                            throw new IOException("Invalid journal. Contains explicitLAC " + " but layout version ("
+                                    + journalVersion + ") is too old to hold this");
+                        }
                     } else {
                         byte[] key = masterKeyCache.get(ledgerId);
                         if (key == null) {
@@ -1182,13 +1201,24 @@ public class Bookie extends BookieCriticalThread {
         }
     }
 
-    public void setExplicitLac(ByteBuf entry, Object ctx, byte[] masterKey)
+    static ByteBuf createExplicitLACEntry(long ledgerId, ByteBuf explicitLac) {
+        ByteBuf bb = Unpooled.buffer(8 + 8 + 4 + explicitLac.capacity());
+        bb.writeLong(ledgerId);
+        bb.writeLong(METAENTRY_ID_LEDGER_EXPLICITLAC);
+        bb.writeInt(explicitLac.capacity());
+        bb.writeBytes(explicitLac);
+        return bb;
+    }
+
+    public void setExplicitLac(ByteBuf entry, WriteCallback writeCallback, Object ctx, byte[] masterKey)
             throws IOException, BookieException {
         try {
             long ledgerId = entry.getLong(entry.readerIndex());
             LedgerDescriptor handle = handles.getHandle(ledgerId, masterKey);
             synchronized (handle) {
                 handle.setExplicitLac(entry);
+                ByteBuf explicitLACEntry = createExplicitLACEntry(ledgerId, entry);
+                getJournal(ledgerId).logAddEntry(explicitLACEntry, false /* ackBeforeSync */, writeCallback, ctx);
             }
         } catch (NoWritableLedgerDirException e) {
             stateManager.transitionToReadOnlyMode();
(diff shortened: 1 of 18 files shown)


eolivelli:
@reddycharan I agree there is no need to check fence flag.

Let's move forward with current approch

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -1182,13 +1201,24 @@ public void recoveryAddEntry(ByteBuf entry, WriteCallback cb, Object ctx, byte[]
         }
     }
 
-    public void setExplicitLac(ByteBuf entry, Object ctx, byte[] masterKey)
+    static ByteBuf createExplicitLACEntry(long ledgerId, ByteBuf explicitLac) {
+        ByteBuf bb = Unpooled.buffer(8 + 8 + 4 + explicitLac.capacity());
```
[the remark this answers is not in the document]
reddycharan: hmm..i just followed createLedgerFenceEntry approach- https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDescriptor.java#L56

eolivelli:
What about adding some test in this file, to narrow the scope of the test? This suite exercises and tests  directly the interface between Bookie and Journal

https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieWriteToJournalTest.java

eolivelli:
Could you also consider adding some test like these?
https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/test/java/org/apache/bookkeeper/proto/WriteEntryProcessorTest.java

I mean only unit tests around the 'Processor' for setExplicitLAC on bookie side.

reddycharan:
Hey @eolivelli , considering the importance/sensitivity of this fix, I added tests at multiple levels -

1. ExplicitLacTest - end-to-end test from BookKeeper client to Bookie level, verified at Bookkeeper client level
2. LedgerStorageTest - end-to-end scenario from BookKeeper client to Bookie level, but verified at Bookie API level (reading explicit lac using Bookie) to validate explicitlac entry added to journal and also verified by reading at FileInfo level to validate explicitlac persisted to FileInfo header.
3.  IndexPersistenceMgrTest - test scenario at IndexPersistenceMgr level, by doing read and write at IndexPersistenceMgr level, also old and new header versions of FileInfo are pre-created for compatibility (backward) validation.

So I guess, I added sufficient testing for this fix.

sijie:
rerun pr validation

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -784,6 +787,23 @@ public void process(int journalVersion, long offset, ByteBuffer recBuff) throws
...
+                            explicitLacBuf.writeBytes(explicitLacBufArray);
+                            byte[] key = masterKeyCache.get(ledgerId);
+                            if (key == null) {
+                                key = ledgerStorage.readMasterKey(ledgerId);
+                            }
+                            LedgerDescriptor handle = handles.getHandle(ledgerId, key);
+                            handle.setExplicitLac(explicitLacBuf);
+                        } else {
+                            throw new IOException("Invalid journal. Contains explicitLAC " + " but layout version ("
+                                    + journalVersion + ") is too old to hold this");
```
jvrao: Just curious, if the software is an older version, how can it even recognize that the entryId is METAENTRY_ID_LEDGER_EXPLICITLAC??

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileInfo.java
```diff
@@ -206,9 +218,11 @@ public synchronized void readHeader() throws IOException {
                 throw new IOException("Missing ledger signature while reading header for " + lf);
             }
             int version = bb.getInt();
-            if (version != HEADER_VERSION) {
+            if (version > CURRENT_HEADER_VERSION) {
                 throw new IOException("Incompatible ledger version " + version + " while reading header for " + lf);
```
jvrao: So this prevents software rollback period. isn't it? Should we consider coming up ? what are the downsides of that? yeah we may not read the explicitLAC other than that, is there anything else? I am worried if there are any situations where we make the cluster not usable.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -784,6 +787,23 @@ public void process(int journalVersion, long offset, ByteBuffer recBuff) throws
...
+                        }
                     } else {
```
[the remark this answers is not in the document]
reddycharan: this journalVersion is the version of journalfile but not "JournalChannel.CURRENT_JOURNAL_FORMAT_VERSION" (current code version), we would get to this 'else' case if we found a journal older than V5 but it is having entry where entryId METAENTRY_ID_LEDGER_EXPLICITLAC, which is not possible and expected in any case. Hence we are throwing IOException here just like https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java#L783

reddycharan:
@jvrao @sijie actually i missed considering Journal version bump for this change. It seems for other special entries in journal, JournalVersion is bumped 

https://github.com/apache/bookkeeper/commit/c65d6ada34215ee3a5466c495d565d745857c435
https://github.com/apache/bookkeeper/commit/d69986c4df6b51ed6ed27c5f45a7b6850e69dacd

can you confirm if Journal version bump is required here.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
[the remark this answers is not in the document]
jvrao: I mean any special entryIds are going to be treated as regular entryIds in this case. So my request is to create a mask to identify SPECIAL entries and filter them out. If there is anything special that the code doesn't know, flag an error and continue. I know this doesn't solve the previous release but future code changes should be good.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileInfo.java
jvrao: This code change is fine. I spoke to @sijie  and @eolivelli  this morning. We needed a bigger gate to identify ondisk changes. I will open a issue for that.

reddycharan:
@sijie and @eolivelli can you please complete the review. Its been waiting for quite sometime.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -121,6 +123,7 @@
     static final long METAENTRY_ID_LEDGER_KEY = -0x1000;
     static final long METAENTRY_ID_FENCE_KEY  = -0x2000;
     public static final long METAENTRY_ID_FORCE_LEDGER  = -0x4000;
+    static final long METAENTRY_ID_LEDGER_EXPLICITLAC  = -0x8000;
```
sijie: I think we need to bump the journal version

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -784,6 +787,23 @@ public void process(int journalVersion, long offset, ByteBuffer recBuff) throws
                                     + " but layout version (" + journalVersion
                                     + ") is too old to hold this");
                         }
+                    } else if (entryId == METAENTRY_ID_LEDGER_EXPLICITLAC) {
+                        if (journalVersion >= JournalChannel.V5) {
```
sijie: this should be JournalChannel.V6

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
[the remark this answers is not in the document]
sijie: +1 on having a mask as what JV suggested.

eolivelli:
For me is quite okay but I want to merge it only after cutting 4.8 branch.
This feature is not enough mature to be delivered with 4.8

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileInfo.java
```diff
@@ -271,10 +304,20 @@ private synchronized void checkOpen(boolean create, boolean openBeforeClose)
     private void writeHeader() throws IOException {
         ByteBuffer bb = ByteBuffer.allocate((int) START_OF_DATA);
         bb.putInt(SIGNATURE);
-        bb.putInt(HEADER_VERSION);
+        bb.putInt(this.headerVersion);
         bb.putInt(masterKey.length);
         bb.put(masterKey);
         bb.putInt(stateBits);
+        if (this.headerVersion >= V1) {
```
sijie: it would be better to make `fileInfo` version configurable in ServerConfiguration, as how we handled in Journal.

so people can upgrade binary but still sets the write version to be v0. so the bookie can run with new binary but writing old version fileinfo. this allows rollback binary.

after people verify the binary work, then can change the write version to v1. at this point, it can still rollback to configuration to v0 and using the new binary. in this approach, it gives some options for rollback.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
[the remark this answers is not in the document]
reddycharan: @sijie @jvrao valid point. for masking purpose, I'm going to assume all negative entryids are SPECIAL entries and if the current code / binary doesn't recognize a special entry then it will log a warn line and proceed.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
reddycharan: yes, bumped journal version

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileInfo.java
reddycharan: yep, adding config for this

reddycharan:
@sijie @jvrao addressed review comments. Can you please check the new iteration I sent.

regarding updating http://bookkeeper.apache.org/docs/latest/admin/upgrade/#46x-to-470-upgrade, I think it would be better and convenient to do it in different PR. Anyhow nothing is written so far for "4.7.x to 4.8.0 upgrade".

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/ExplicitLacTest.java
```diff
@@ -50,6 +53,12 @@ public ExplicitLacTest(Class<? extends LedgerStorage> storageClass) {
         super(1);
         this.digestType = DigestType.CRC32;
         baseConf.setLedgerStorageClass(storageClass.getName());
+        /*
+         * to persist explicitLac, journalFormatVersionToWrite should be atleast
+         * V6 and fileInfoFormatVersionToWrite should be atleast V1
+         */
+        baseConf.setJournalFormatVersionToWrite(6);
```
eolivelli: @reddycharan maybe it is better to use defined constants for these values (I mean '6' here and '1' below)

eolivelli:
retest this please

eolivelli:
@sijie  @reddycharan 

It seems to me that the only remaining concern from Sijie is to upgrade the guide.
If you Charan create a new task and have time to finish the work I would be happy to merge this change, so that we can keep it on the branch for a few days (and run downstream projects tests) before cutting the release branch

reddycharan:
@eolivelli created issue for upgrade doc - https://github.com/apache/bookkeeper/issues/1565. Will follow up with that once this is closed.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/ExplicitLacTest.java
reddycharan: yes, I thought about this. But "JournalChannel" is not public class (which contains defined constants) and similarly "FileInfo" is not public either. 

I guess for the same reason actual values are used in other places aswell - https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/test/java/org/apache/bookkeeper/conf/TestBKConfiguration.java#L50

https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java#L637

eolivelli: Oh, ok!
Thank you for your explanation

eolivelli: Oh, ok!
Thank you for your explanation

eolivelli:
retest this please

eolivelli:
rerun bookie tests
rerun integration tests

eolivelli:

sijie:
@jvrao since you left a few comments here before, do you want to review this again?

reddycharan:
Thanks all for your feedback on this work item. Gladly, tests have passed finally.

Issue #1540, opened 2018-07-12
Bookie/BookieServer components shutdown will fail to end/exit the BookieProcess


reddycharan:
**BUG REPORT**

1. Please describe the issue you observed:

If Bookie/BookieServer components are shutdown internally because of any fatal errors (ExitCode - INVALID_CONF, SERVER_EXCEPTION, ZK_EXPIRED, ZK_REG_FAIL, BOOKIE_EXCEPTION) then it will go through shutdown method logic and shutdowns components internal to Bookie/BookieServer but it will not succeed in bringing down the bookie process. 

This is because in BookieServer.main / server.Main.doMain it would wait for the startComponent future to complete http://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/server/Main.java#L227 .  The startComponent future will be market complete only in runtime shutdownhook - https://github.com/apache/bookkeeper/blob/master/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/component/ComponentStarter.java#L66. 

But the problem is nowhere in Bookie/BookieProcess shutdown we are calling System.exit() and hence the runtime shutdownhook is not executed to mark the startComponent future to complete. Hence Main.doMain will wait forever on this future though Bookie/BookieServer components are shutdown because of known fatal errors.

Following is the thread callstacktrace of main method which is waiting forever
(line numbers might not match, since we are using little older code)

main - priority:5 - threadId:0x00007fd60000d000 - nativeId:0x2d37b - state:WAITING
stackTrace:
java.lang.Thread.State: WAITING (parking)
at sun.misc.Unsafe.park(Native Method)
- parking to wait for <0x00007fcddf015af0> (a java.util.concurrent.CompletableFuture$Signaller)
at java.util.concurrent.locks.LockSupport.park(LockSupport.java:175)
at java.util.concurrent.CompletableFuture$Signaller.block(CompletableFuture.java:1693)
at java.util.concurrent.ForkJoinPool.managedBlock(ForkJoinPool.java:3323)
at java.util.concurrent.CompletableFuture.waitingGet(CompletableFuture.java:1729)
at java.util.concurrent.CompletableFuture.get(CompletableFuture.java:1895)
at org.apache.bookkeeper.server.Main.doMain(Main.java:215)
at org.apache.bookkeeper.server.Main.main(Main.java:189)
at org.apache.bookkeeper.proto.BookieServer.main(BookieServer.java:256)

[...]

PR #1543, opened 2018-07-14
Issue 1540: Bookie/BookieServer components shutdown will fail to end exit the BookieProcess

sijie:

 ### Motivation

Fixes the issue at #1540.

If Bookie/BookieServer components are shutdown internally because of any fatal errors
(ExitCode - INVALID_CONF, SERVER_EXCEPTION, ZK_EXPIRED, ZK_REG_FAIL, BOOKIE_EXCEPTION) then
it will go through shutdown method logic and shutdowns components internal to Bookie/BookieServer
but it will not succeed in bringing down the bookie process.

This is because in BookieServer.main / server.Main.doMain it would wait for the startComponent
future to complete
http://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/server/Main.java#L227 .
The startComponent future will be market complete only in runtime shutdownhook -
https://github.com/apache/bookkeeper/blob/master/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/component/ComponentStarter.java#L66.

But the problem is nowhere in Bookie/BookieProcess shutdown we are calling System.exit() and hence
the runtime shutdownhook is not executed to mark the startComponent future to complete. Hence
Main.doMain will wait forever on this future though Bookie/BookieServer components are shutdown
because of known fatal errors.

 ### Regression

Issue #508 introduced this regression. Before this change, the main thread is blocking using `BookieServer#join()`.
When bookie is dead for any reason, the DeathWatchThread will kill the bookie and bookie server. so the main thread will quite.
However after #508 is introduced, the lifecycle management is disconnected from the bookie and bookie server. so when they are dead,
lifecycle management is unaware of the situation and the main thread doesn't quite.

 ### Changes

- Add `UncaughtExceptionHandler` to lifecycle components
- When a lifecycle component hits an error, it can use `UncaughtExceptionHandler` to notify lifecycle component stack to shutdown the whole stack

Master Issue: #1540 

> ---
>
> ---

> ---
> quickly and easily:
>
>
>
> Otherwise:
> 
> 
> ---

### DIFF
--- a/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/component/ComponentStarter.java
+++ b/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/component/ComponentStarter.java
@@ -62,8 +62,19 @@ public class ComponentStarter {
      */
     public static CompletableFuture<Void> startComponent(LifecycleComponent component) {
         CompletableFuture<Void> future = new CompletableFuture<>();
-        Runtime.getRuntime().addShutdownHook(new Thread(
-            new ComponentShutdownHook(component, future), "component-shutdown-thread"));
+        final Thread shutdownHookThread = new Thread(
+            new ComponentShutdownHook(component, future),
+            "component-shutdown-thread"
+        );
+
+        // register a shutdown hook
+        Runtime.getRuntime().addShutdownHook(shutdownHookThread);
+
+        // register a component exception handler
+        component.setExceptionHandler((t, e) -> {
+            // start the shutdown hook when an uncaught exception happen in the lifecycle component.
+            shutdownHookThread.start();
+        });
 
         log.info("Starting component {}.", component.getName());
         component.start();
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieServer.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieServer.java
@@ -255,6 +255,12 @@ public class BookieServer {
         DeathWatcher(ServerConfiguration conf) {
             super("BookieDeathWatcher-" + conf.getBookiePort());
             watchInterval = conf.getDeathWatchInterval();
+            // set a default uncaught exception handler to shutdown the bookie server
+            // when it notices the bookie is not running any more.
+            setUncaughtExceptionHandler((thread, cause) -> {
+                shutdown();
+                LOG.info("BookieDeathWatcher exited loop!");
+            });
         }
 
         @Override
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieServer.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieServer.java
@@ -26,6 +26,7 @@ import static org.apache.bookkeeper.conf.AbstractConfiguration.PERMITTED_STARTUP
 
 import com.google.common.annotations.VisibleForTesting;
 import java.io.IOException;
+import java.lang.Thread.UncaughtExceptionHandler;
 import java.net.UnknownHostException;
 import java.security.AccessControlException;
 import java.util.Arrays;
@@ -34,6 +35,7 @@ import java.util.concurrent.TimeUnit;
 import org.apache.bookkeeper.bookie.Bookie;
 import org.apache.bookkeeper.bookie.BookieCriticalThread;
 import org.apache.bookkeeper.bookie.BookieException;
+import org.apache.bookkeeper.bookie.BookieException.BookieDeadException;
 import org.apache.bookkeeper.bookie.ExitCode;
 import org.apache.bookkeeper.bookie.ReadOnlyBookie;
 import org.apache.bookkeeper.client.BKException;
@@ -73,6 +75,9 @@ public class BookieServer {
     // Expose Stats
     private final StatsLogger statsLogger;
 
+    // Exception handler
+    private volatile UncaughtExceptionHandler uncaughtExceptionHandler = null;
+
     public BookieServer(ServerConfiguration conf) throws IOException,
             KeeperException, InterruptedException, BookieException,
             UnavailableException, CompatibilityException, SecurityException {
@@ -108,6 +113,18 @@ public class BookieServer {
         this.requestProcessor = new BookieRequestProcessor(conf, bookie,
                 statsLogger.scope(SERVER_SCOPE), shFactory);
         this.nettyServer.setRequestProcessor(this.requestProcessor);
+    }
+
+    /**
+     * Currently the uncaught exception handler is used for DeathWatcher to notify
+     * lifecycle management that a bookie is dead for some reasons.
+     *
+     * <p>in future, we can register this <tt>exceptionHandler</tt> to critical threads
+     * so when those threads are dead, it will automatically trigger lifecycle management
+     * to shutdown the process.
+     */
+    public void setExceptionHandler(UncaughtExceptionHandler exceptionHandler) {
+        this.uncaughtExceptionHandler = exceptionHandler;
     }
 
     protected Bookie newBookie(ServerConfiguration conf)
@@ -249,6 +266,12 @@ public class BookieServer {
                 }
                 if (!isBookieRunning()) {
                     shutdown();
+                    if (null != uncaughtExceptionHandler) {
+                        uncaughtExceptionHandler.uncaughtException(
+                            this,
+                            new BookieDeadException("Bookie is not running any more")
+                                .fillInStackTrace());
+                    }
                     break;
                 }
             }
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieServer.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieServer.java
@@ -258,8 +258,9 @@ public class BookieServer {
             // set a default uncaught exception handler to shutdown the bookie server
             // when it notices the bookie is not running any more.
             setUncaughtExceptionHandler((thread, cause) -> {
+                LOG.info("BookieDeathWatcher exited loop due to uncaught exception from thread {}",
+                    thread.getName(), cause);
                 shutdown();
-                LOG.info("BookieDeathWatcher exited loop!");
             });
         }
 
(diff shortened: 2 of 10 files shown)


sijie:
@reddycharan @jvrao  this is the fix.

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/component/ComponentStarter.java
```diff
@@ -62,8 +62,19 @@ public void run() {
      */
     public static CompletableFuture<Void> startComponent(LifecycleComponent component) {
         CompletableFuture<Void> future = new CompletableFuture<>();
-        Runtime.getRuntime().addShutdownHook(new Thread(
-            new ComponentShutdownHook(component, future), "component-shutdown-thread"));
+        final Thread shutdownHookThread = new Thread(
+            new ComponentShutdownHook(component, future),
+            "component-shutdown-thread"
```
eolivelli: I wonder if we can use a more meaning name. Not so important

sijie: what is your suggestion?

sijie:
rerun bookkeeper-server remaining tests

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/component/ComponentStarter.java
eolivelli: Something with the name/class  of the component

sijie:

sijie:

reddycharan:
@sijie 

hmm.. I kind of understood what you are doing, but I'm not sure why it is called "ExceptionHandler" (all across the code) and in the fist place why we need 'UncaughtExceptionHandler'.

In this change, BookieServer is signaling that it is not running and it is done, so https://github.com/apache/bookkeeper/pull/1543/files#diff-e7149cec76f474d8ae8a515cbe0464e8R74 can start ‘shutdownHookThread’ (ComponentShutdownHook) in ComponentStarter. So all it is doing is signaling  ‘ComponentStarter’ to start ‘shutdownHookThread’, which completes future, so that Main.doMain need not wait for future anymore. Also exception passed to handler is dropped/ignored in the uncaught exception handler you created https://github.com/apache/bookkeeper/pull/1543/files#diff-e7149cec76f474d8ae8a515cbe0464e8R74. So it doesn’t makes sense completely to call ExceptionHandler.

And the other (major) concern I have is, in the previous versions we make sure “System.Exit” is called in the shutdown path of Bookie/BookieServer - https://github.com/apache/bookkeeper/blob/branch-4.4/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieServer.java#L401 . This is completely missing in the current LifeCycleComponent infrastructure and also in your current change. This is needed because for whatsoever reason non-daemon thread could be running in this Bookie process, though we have come to this (or completed) shutdown path and it could prevent Bookie process from shutdown, no matter how thorough we are with our lifecyclecomponent and core components handling/management.

sijie:
> I'm not sure why it is called "ExceptionHandler" (all across the code) and in the fist place why do we need 'UncaughtExceptionHandler'.

 why I use "UncaughtExceptionHandler" is explained here: https://github.com/apache/bookkeeper/pull/1543/files#diff-87fe9faad6f1b8ff70975c11e676b4fbR122

I reused `UncaughtExceptionHandler` so that it can be easily used to attach to any critical thread, when those threads dead, it will trigger this shutdown logic.

I treated a bookie dead as an "uncaught" exception from a lifecycle component perspective, so registering an uncaught exception handler so the lifecycle management can react to those "uncaught" exception. once this mechanism is in placed, it can also be reused in other uncaught situation.

Hope this make sense.

> in the previous versions we make sure “System.Exit” is called in the shutdown path of Bookie/BookieServer

This is a good point. I will add `System.Exit` when the lifecycle shuts things down.

reddycharan:
>why I use "UncaughtExceptionHandler" is explained here: https://github.com/apache/bookkeeper/pull/1543/files#diff-87fe9faad6f1b8ff70975c11e676b4fbR122

> I reused UncaughtExceptionHandler so that it can be easily used to attach to any critical thread, when those threads dead, it will trigger this shutdown logic.

Yeah, I read your comment. But my concern here is it is not just simple exception handler, it starts ComponentShutdownHook (shutdownHookThread) thread, which completes component future and it has got nothing to do with the exception (because it drops/ignores the exception). So probably in LifeCycleComponent instead of “setExceptionHandler” method name “setComponentShutdownHandler” and in BookieServer instead of “setExceptionHandler” method name “bookieServiceShutdownHandler” would be more appropriate. I feel “ExceptionHandler” is being used more generously when we are not really handling exception.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieServer.java
```diff
@@ -249,6 +266,12 @@ public void run() {
                 }
                 if (!isBookieRunning()) {
                     shutdown();
+                    if (null != uncaughtExceptionHandler) {
+                        uncaughtExceptionHandler.uncaughtException(
+                            this,
+                            new BookieDeadException("Bookie is not running any more")
+                                .fillInStackTrace());
```
dlg99: Why do we need fillInStackTrace() here? 
Stack trace should be filled in by the Throwable's constructor already, assuming BookieException called super's constructor.

Now this got me looking at BookieException.
"public BookieException(int code)" sets private field code but does not call super().
All other constructors call super(..) variants but don't set code so we cannot rely on getCode(). Luckily, getCode() is not used so why do we even bother passing the code and having the field?

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieServer.java
```diff
@@ -249,6 +266,12 @@ public void run() {
                 }
                 if (!isBookieRunning()) {
                     shutdown();
+                    if (null != uncaughtExceptionHandler) {
+                        uncaughtExceptionHandler.uncaughtException(
```
dlg99: overall this looks weird. There is no uncaught exception but we are triggering uncaughtExceptionHandler with new exception, but only if it exists. 
I'd either:
- just throw new exception and let the handler really deal with it as with unhandled exception unless caught by higher level component. let app crash if there is no unhandled exception handler.
- make BookieServer implement lifecyclecomponent, rename shutdown into doStop and call stop() instead of shutdown to ensure proper notification of all subscribers.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieServer.java
sijie: > Why do we need fillInStackTrace() here?
Stack trace should be filled in by the Throwable's constructor already, assuming BookieException called super's constructor.

good question. I think some of the constructors are not calling super. I can fix the bookie exception constructors and remove `fillInStackTrace`

> variants but don't set code so we cannot rely on getCode()

we should set the code. will fix it.

> Luckily, getCode() is not used so why do we even bother passing the code and having the field?

the existing of the code is following BKException. will fix the constructors.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieServer.java
sijie: > make BookieServer implement lifecyclecomponent, rename shutdown into doStop and call stop() instead of shutdown to ensure proper notification of all subscribers.

this won't work as shutting down one component will not notify the lifecycle stack to shutdown the whole. so that's why I am using a component being dead as an "uncaught" situation to notify the lifecycle stack to handle this "uncaught" situation and shutdown the whole stack.

> just throw new exception and let the handler really deal with it as with unhandled exception unless caught by higher level component. let app crash if there is no unhandled exception handler.

what might be better here is following you suggested:

- pass the uncaught exception handler to DeathWatchThread. in DeathWatchThread, when it detects a bookie is not running, thrown a BookieBeadException. then the uncaught exception handler will kick in to handle this uncaught exception.

sijie:
> in the previous versions we make sure “System.Exit” is called in the shutdown path of Bookie/BookieServer - https://github.com/apache/bookkeeper/blob/branch-4.4/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieServer.java#L401 . This is completely missing in the current LifeCycleComponent infrastructure and also in your current change. 

@reddycharan sorry I took my comment back. `System.Exit` is called as before. It is called here: https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/server/Main.java#L202

sijie:
@reddycharan @dlg99 I addressed your comments.

- fix the constructor issue for BookieException
- change the way how DeathWatcherThread propagates exceptions. so DeatchWatcher will register the uncaught exception handler, when it notices that the bookie exception, it will throw a RuntimeException and it will trigger the Uncaught exception handler to handle this bookie dying situation. 

let me know if this approach looks better to you guys.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieServer.java
```diff
@@ -248,11 +267,13 @@ public void run() {
                     Thread.currentThread().interrupt();
                 }
                 if (!isBookieRunning()) {
-                    shutdown();
-                    break;
+                    LOG.info("BookieDeathWatcher noticed the bookie is not running any more, exiting the watch loop!");
```
reddycharan: just want to make sure that removal of "shutdown()" call here is intentional.

sijie: yes it is intentional. because the uncaught exception handle registered in death watcher thread will handle all the shutdown logic.

reddycharan:
LGTM +1

sijie:
retest this please

sijie:
@dlg99 @reddycharan can you review my latest commit d7867d9 ? without this change, some of tests will be crashing because it will trigger `Runtime.getRuntime().exit(-1)`. This latest commit will ensure it works for both tests and Main program.

sijie:
@dlg99 I addressed your comment around the logging part. let me know if it looks okay.

sijie:
rerun pr validation
rebuild java8
rebuild java9

sijie:
retest this please

PR #1547, opened 2018-07-16
Allow concurrent write requests for same entry in V2 protocol

ivankelly:
With the V2 protocol we have previously allowed a single client to
issue concurrent read requests for the same entry. We didn't allow the
same for write requests, as we always assume single writer. This
assumption isn't absolutely correct though. When opening a ledger, the
recovery process will rewrite entries after the LAC. To
PerChannelBookieClient these look no different to normal writes.

If a client tries to open a ledger multiple times concurrently,
currently some of the attempts can hang forever since there is no
support for concurrent writes.

This patch adds support for concurrent writes, so multiple concurrent
ledger open attempts will always complete (though some may fail for
unrelated reasons).

### DIFF
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/ConcurrentV2RecoveryTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/ConcurrentV2RecoveryTest.java
@@ -95,7 +95,7 @@ public class ConcurrentV2RecoveryTest extends BookKeeperClusterTestCase  {
                         });
                 futures.add(future);
             }
-            Thread.sleep(1000);
+
             latch.countDown();
             for (Future<?> f : futures) {
                 try {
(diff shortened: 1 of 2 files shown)


bookkeeper-server/src/test/java/org/apache/bookkeeper/client/ConcurrentV2RecoveryTest.java
```diff
@@ -0,0 +1,113 @@
...
+                                                        future.completeExceptionally(BKException.create(rc));
+                                                    } else {
+                                                        future.complete(handle);
+                                                    }
+                                                }, null);
+                            return future;
+                        });
+                futures.add(future);
+            }
+            Thread.sleep(1000);
```
eolivelli: This is not needed

sijie:

sijie:
rebuild java8
rebuild java9

eolivelli:
retest this please

sijie:
retest this please

eolivelli:

Issue #1563, opened 2018-07-20
Spammy logs on Bookie in ReadEntryProcessorV3 in case of tailing reader on empty ledger


eolivelli:
In case of a tailing reader which is looping over a readLastAddConfirmed() on an ledger without entries (so unknown to any bookie) we have a lot of logs written at INFO level

`No ledger found while reading entry: -1 from ledger: xxx`

We should decrease log level so that this log  does not spam the server and fills up the disk.

Please note that code is like this:

```
           if (RequestUtils.isFenceRequest(readRequest)) {
                LOG.info("No ledger found reading entry {} when fencing ledger {}", entryId, ledgerId);
            } else {
                LOG.info("No ledger found while reading entry: {} from ledger: {}", entryId, ledgerId);
            }
            return buildResponse(readResponse, StatusCode.ENOLEDGER, startTimeSw);
```

The first case about a fenced ledger may stay at INFO level, it is more "important" and "useful".
I am talking about the second case.

PR #1564, opened 2018-07-20
Issue-1563 Spammy logs on Bookie in ReadEntryProcessorV3 in case of tailing reader on empty ledger

eolivelli:
### Motivation

In case of an empty ledger the bookie floods the disk with an useless message.

### Changes

Make that line "No ledger found while reading entry" to be logged at 'debug' level.

Master issue #1563 

> ---
>
> - [x] [skip bookkeeper-server bookie tests]: skip testing `org.apache.bookkeeper.bookie` in bookkeeper-server module.
> - [x] [skip bookkeeper-server client tests]: skip testing `org.apache.bookkeeper.client` in bookkeeper-server module.
> - [x] [skip bookkeeper-server replication tests]: skip testing `org.apache.bookkeeper.replication` in bookkeeper-server module.
> - [x] [skip bookkeeper-server tls tests]: skip testing `org.apache.bookkeeper.tls` in bookkeeper-server module.
> - [x] [skip bookkeeper-server remaining tests]: skip testing all other tests in bookkeeper-server module.
> - [x] [skip integration tests]: skip docker based integration tests. if you make java code changes, you shouldn't skip integration tests.
> ---

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadEntryProcessorV3.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadEntryProcessorV3.java
@@ -221,8 +221,8 @@ class ReadEntryProcessorV3 extends PacketProcessorBaseV3 {
         } catch (Bookie.NoLedgerException e) {
             if (RequestUtils.isFenceRequest(readRequest)) {
                 LOG.info("No ledger found reading entry {} when fencing ledger {}", entryId, ledgerId);
-            } else {
-                LOG.info("No ledger found while reading entry: {} from ledger: {}", entryId, ledgerId);
+            } else if (LOG.isDebugEnabled()) {
+                LOG.debug("No ledger found while reading entry: {} from ledger: {}", entryId, ledgerId);
             }
             return buildResponse(readResponse, StatusCode.ENOLEDGER, startTimeSw);
         } catch (Bookie.NoEntryException e) {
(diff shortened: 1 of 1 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadEntryProcessorV3.java
```diff
@@ -221,8 +221,8 @@ protected ReadResponse getReadResponse() {
         } catch (Bookie.NoLedgerException e) {
             if (RequestUtils.isFenceRequest(readRequest)) {
                 LOG.info("No ledger found reading entry {} when fencing ledger {}", entryId, ledgerId);
-            } else {
-                LOG.info("No ledger found while reading entry: {} from ledger: {}", entryId, ledgerId);
+            } else if (LOG.isDebugEnabled()) {
```
sijie: a couple of comments:

- "No ledger found" should be over multiple places. v2 and v3 protocol and also I think LongPoll would have similar logging. we need to make sure the logging behaviors are consistent between places.

- I think it is useful if it is reading a specific entry but the ledger is not found. it is probably annoying when reading -1, since no entry is really added yet.

eolivelli: Totally agree! Will update patch soon

eolivelli:
@sijie 
I have changed the code in a way that we are skipping only this message for my case (entryId = -1)

I have checked various cases on the read path:

V2 read processor:
It does not ever write this log, it only writes at "debug" level, I don't know if it is worth to change it now

https://github.com/apache/bookkeeper/blob/89f8ee91ff768664b5da2c52fcac1d1b7f0614f0/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadEntryProcessor.java#L105

Long Pool Readers:
I don't know very well the overall workflow but LongPollReadEntryProcessorV3 is a subclass of ReadEntryProcessorV3 and this change is covering that case

PR #1567, opened 2018-07-24
Restore JaCoCo configuration on bookkeeper-server module

eolivelli:
With commit  c5a3621 - Sijie Guo <sijie@apache.org>
Move TimedOutTestsListener from dlog to bookkeeper-common

We had overwritten surefire configuration on mostly every profile in bookkeeper-server module and this disabled JaCoCo agent which is activated using 'code-coverage' Maven profile.

This change forces surefire configuration in case of 'code-coverage' profile

### DIFF
--- a/bookkeeper-server/pom.xml
+++ b/bookkeeper-server/pom.xml
@@ -185,24 +185,6 @@
       </plugin>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-surefire-plugin</artifactId>
-        <version>${maven-surefire-plugin.version}</version>
-        <configuration>
-          <argLine>-Xmx2G -Djava.net.preferIPv4Stack=true</argLine>
-          <redirectTestOutputToFile>${redirectTestOutputToFile}</redirectTestOutputToFile>
-          <reuseForks>false</reuseForks>
-          <forkedProcessTimeoutInSeconds>1800</forkedProcessTimeoutInSeconds>
-          <rerunFailingTestsCount>2</rerunFailingTestsCount>
-          <properties>
-            <property>
-              <name>listener</name>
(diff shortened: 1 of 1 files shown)


sijie:
@eolivelli I have one general question here - why can we run code-coverage report on all modules? there is no point to limit code coverage into one module.

eolivelli:
@sijie 
this is current configuration

`clean verify jacoco:report coveralls:report -Pcode-coverage -DrepoToken=$COVERALLS_REPO_TOKEN -Dmaven.test.failure.ignore=true -Dstream`

having -Dstream should run the tests of the full codebase

eolivelli:
rerun bookkeeper-server clients tests
rerun integration tests

eolivelli:

sijie:
@eolivelli my question is if jacoco is configured at root pom, why we need to add this to bookkeeper-server? it is unclear to me.

eolivelli:
This is because in your commit you have overridden surefire configuration for each profile,  in bookkeeper-server module only.
Maybe an alternative approach would be to add the listener in the root pom and drop your customization in bookkeeper-server.

We need that @argLine parameter in order to make JaCoCo work

sijie:
> Maybe an alternative approach would be to add the listener in the root pom and drop your customization in bookkeeper-server.

That's probably better. It is great to have this listener to dump stack traces when a test is timed out.

eolivelli:
I have tried to move the listener to the root pom but it makes all modules depend on bookkeeper-commons.
I will mess up all the dependency tree.
And actually I can't make it work for circe-checksum (I guess nar plugin does some magic)

Another option is to disable the listener at all, we are seeing that it is not helping so much.

I would like to go with this last option: drop the listener from bookkeeper-server module

eolivelli:
Okay I will revert last change

eolivelli:
This issue will be fixed with #1570

PR #1569, opened 2018-07-26
Replace guava multimap in PCBC with custom impl

ivankelly:
For a long time PerChannelBookieClient has used guava
LinkedListMultiMap to store conflicting V2 completion keys and
values. This is problematic though. Completion keys are pooled
objects. When a key-value pair is stored in a LinkedListMultiMap, if
it is the first value for that key, a collection is created for the
values, and added to a top-level map using the key, and then the key
and the value are added to the collection. When a second value is
added for the same key, the key and value are simply added to the
collection. The problem occurs when the first key is removed. PBCB
will recycle the key object, but this object is still being used in
the multimap in the top-level map. This causes all sorts of fun like
NullPointerException and IllegalStateException.

Because of this, this patch introduces a very simple multimap
implementation that only stores the key one time (in the collection)
and uses the hashCode of the key to separate the collections into
buckets. It's pretty inefficient, but this code it only hit in the
rare case where a client is trying to read or write the same entry
from the same ledger more than once at the same time.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
@@ -21,9 +21,6 @@ package org.apache.bookkeeper.proto;
 import static org.apache.bookkeeper.client.LedgerHandle.INVALID_ENTRY_ID;
 
 import com.google.common.base.Joiner;
-import com.google.common.collect.ArrayListMultimap;
-import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.ListMultimap;
 import com.google.common.collect.Sets;
 import com.google.protobuf.ByteString;
 import com.google.protobuf.ExtensionRegistry;
@@ -72,7 +69,6 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.EnumSet;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Optional;
 import java.util.Queue;
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
@@ -21,7 +21,8 @@ package org.apache.bookkeeper.proto;
 import static org.apache.bookkeeper.client.LedgerHandle.INVALID_ENTRY_ID;
 
 import com.google.common.base.Joiner;
-import com.google.common.collect.LinkedListMultimap;
+import com.google.common.collect.ArrayListMultimap;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ListMultimap;
 import com.google.common.collect.Sets;
 import com.google.protobuf.ByteString;
@@ -73,6 +74,7 @@ import java.util.Collections;
 import java.util.EnumSet;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Optional;
 import java.util.Queue;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
@@ -131,6 +133,7 @@ import org.apache.bookkeeper.util.MathUtils;
 import org.apache.bookkeeper.util.SafeRunnable;
 import org.apache.bookkeeper.util.StringUtils;
 import org.apache.bookkeeper.util.collections.ConcurrentOpenHashMap;
+import org.apache.bookkeeper.util.collections.SynchronizedHashMultiMap;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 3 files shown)


eolivelli:
There are a few checkstyle issues.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
```diff
@@ -1068,16 +1054,10 @@ void errorOutPendingOps(int rc) {
...
-        // on keys and see if we are successfully able to remove the key from
-        // the map. Because the add and the read methods also do the same thing
-        // in case they get a write failure on the socket. The one who
-        // successfully removes the key from the map is the one responsible for
-        // calling the application callback.
-        for (CompletionKey key : completionObjectsV2Conflicts.keySet()) {
-            while (completionObjectsV2Conflicts.get(key).size() > 0) {
-                errorOut(key, rc);
-            }
+        Optional<CompletionKey> multikey = completionObjectsV2Conflicts.getAnyKey();
```
eolivelli: this 'getAnyKey()' sounds a bit weird to me, therefore inside the implementation we allocating a bunch of objects (using stream API).
I have an alternative proposal:

`completionObjectsV2Conflicts.forEachValue( k -> errorOut(k, rc))`

the iteration will be internal to the implementation, we will save method calls and allocations and overall the code will look better

ivankelly: errorOut modifies the collection, so forEachValue will give you a ConcurrentModificationException. 

Regarding allocations, this isn't in a critical path, so I didn't take any care to avoid them, but escape analysis should get rid of most of them in any case.

eolivelli: Oh, I missed the internals of errorOut.

Okay for me

sijie:
This change is needed to be cherry-picked to `release/4.7.2`.

The problem has been observed using `4.7.1`.

```
09:58:15.041 [BookKeeperClientScheduler-OrderedScheduler-0-0] ERROR org.apache.bookkeeper.common.util.SafeRunnable - Unexpected throwable caught
java.util.NoSuchElementException: null
        at com.google.common.collect.LinkedListMultimap.checkElement(LinkedListMultimap.java:313) ~[com.google.guava-guava-21.0.jar:?]
        at com.google.common.collect.LinkedListMultimap.access$300(LinkedListMultimap.java:104) ~[com.google.guava-guava-21.0.jar:?]
        at com.google.common.collect.LinkedListMultimap$NodeIterator.previous(LinkedListMultimap.java:391) ~[com.google.guava-guava-21.0.jar:?]
        at com.google.common.collect.LinkedListMultimap$NodeIterator.<init>(LinkedListMultimap.java:332) ~[com.google.guava-guava-21.0.jar:?]
        at com.google.common.collect.LinkedListMultimap$1ValuesImpl.listIterator(LinkedListMultimap.java:760) ~[com.google.guava-guava-21.0.jar:?]
        at java.util.AbstractList.listIterator(AbstractList.java:299) ~[?:1.8.0_181]
        at java.util.AbstractSequentialList.iterator(AbstractSequentialList.java:239) ~[?:1.8.0_181]
        at org.apache.bookkeeper.proto.PerChannelBookieClient.checkTimeoutOnPendingOperations(PerChannelBookieClient.java:803) ~[org.apache.bookkeeper-bookkeeper-server-4.7.1.jar:4.7.1]
        at org.apache.bookkeeper.proto.DefaultPerChannelBookieClientPool.checkTimeoutOnPendingOperations(DefaultPerChannelBookieClientPool.java:100) ~[org.apache.bookkeeper-bookkeeper-server-4.7.1.jar:4.7.1]
        at org.apache.bookkeeper.proto.BookieClient.monitorPendingOperations(BookieClient.java:453) ~[org.apache.bookkeeper-bookkeeper-server-4.7.1.jar:4.7.1]
        at org.apache.bookkeeper.proto.BookieClient.lambda$new$0(BookieClient.java:118) ~[org.apache.bookkeeper-bookkeeper-server-4.7.1.jar:4.7.1]
        at org.apache.bookkeeper.util.SafeRunnable$1.safeRun(SafeRunnable.java:43) ~[org.apache.bookkeeper-bookkeeper-server-4.7.1.jar:4.7.1]
        at org.apache.bookkeeper.common.util.SafeRunnable.run(SafeRunnable.java:36) [org.apache.bookkeeper-bookkeeper-common-4.7.1.jar:4.7.1]
        at com.google.common.util.concurrent.MoreExecutors$ScheduledListeningDecorator$NeverSuccessfulListenableFutureTask.run(MoreExecutors.java:587) [com.google.guava-guava-21.0.jar:?]
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511) [?:1.8.0_181]
        at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:308) [?:1.8.0_181]
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:180) [?:1.8.0_181]
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:294) [?:1.8.0_181]
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149) [?:1.8.0_181]
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624) [?:1.8.0_181]
        at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30) [io.netty-netty-all-4.1.22.Final.jar:4.1.22.Final]
        at java.lang.Thread.run(Thread.java:748) [?:1.8.0_181]
```

/cc @ivankelly @merlimat

merlimat:
Yes, this should go into 4.7.2

PR #1572, opened 2018-07-30
Use readExplicitLAC instead of readEntry in order to get current LastAddConfirmed

eolivelli:
### Motivation

Readers have to explicitly call readExplicitLastConfirmed in order to leverage ExplicitLAC.
I would like to have a unified way to read LastAddConfirmed.

### Changes
This change makes the client use internally readExplicitLastConfirmed instead of using readLastAddConfirmed.
We need to introduce something similar to tryReadLastAddConfirmed in order to make it consistent  with readLastAddConfirmed.

It lacks support for Long Poll Reads, notably we would have to listen on ExplicitLAC modifications on the bookie side in order to wake up the readers even in case of setExplicitLAC from the writer and we would need to change protocol for readLastAddConfirmedAndEntry in order to return the ExplicitLAC.

We cannot support v2 clients.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
@@ -1338,11 +1338,11 @@ public class LedgerHandle implements WriteHandle {
         if (clientCtx.getConf().useExplicitLacForReads) {
             asyncReadExplicitLastConfirmed(cb, ctx);
         } else {
-            asyncReadLastConfirmedInternal(cb, ctx);
+            asyncReadPiggybackLastConfirmed(cb, ctx);
         }
     }
 
-    private void asyncReadLastConfirmedInternal(final ReadLastConfirmedCallback cb, final Object ctx) {
+    private void asyncReadPiggybackLastConfirmed(final ReadLastConfirmedCallback cb, final Object ctx) {
         boolean isClosed;
         long lastEntryId;
         synchronized (this) {
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
@@ -1388,11 +1388,11 @@ public class LedgerHandle implements WriteHandle {
         if (clientCtx.getConf().useExplicitLacForReads) {
             asyncTryReadExplicitLastConfirmed(cb, ctx);
         } else {
-            asyncTryReadLastConfirmedInternal(cb, ctx);
+            asyncTryReadPiggybackLastConfirmed(cb, ctx);
         }
     }
 
-    private void asyncTryReadLastConfirmedInternal(final ReadLastConfirmedCallback cb, final Object ctx) {
+    private void asyncTryReadPiggybackLastConfirmed(final ReadLastConfirmedCallback cb, final Object ctx) {
         boolean isClosed;
         long lastEntryId;
         synchronized (this) {
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
@@ -1335,7 +1335,14 @@ public class LedgerHandle implements WriteHandle {
      */
 
     public void asyncReadLastConfirmed(final ReadLastConfirmedCallback cb, final Object ctx) {
-        if (useLegacyMode) {
+        if (clientCtx.getConf().useExplicitLacForReads) {
+            asyncReadExplicitLastConfirmed(cb, ctx);
+        } else {
+            asyncReadLastConfirmedInternal(cb, ctx);
+        }
+    }
+
+    private void asyncReadLastConfirmedInternal(final ReadLastConfirmedCallback cb, final Object ctx) {
         boolean isClosed;
         long lastEntryId;
         synchronized (this) {
@@ -1361,9 +1368,6 @@ public class LedgerHandle implements WriteHandle {
             };
 
         new ReadLastConfirmedOp(this, clientCtx.getBookieClient(), getCurrentEnsemble(), innercb).initiate();
-        } else {
-        asyncReadExplicitLastConfirmed(cb, ctx);
-        }
     }
 
     /**
@@ -1381,7 +1385,14 @@ public class LedgerHandle implements WriteHandle {
      *          callback context
      */
     public void asyncTryReadLastConfirmed(final ReadLastConfirmedCallback cb, final Object ctx) {
-        if (legacyMode) {
+        if (clientCtx.getConf().useExplicitLacForReads) {
+            asyncTryReadExplicitLastConfirmed(cb, ctx);
+        } else {
+            asyncTryReadLastConfirmedInternal(cb, ctx);
+        }
+    }
+
+    private void asyncTryReadLastConfirmedInternal(final ReadLastConfirmedCallback cb, final Object ctx) {
         boolean isClosed;
         long lastEntryId;
         synchronized (this) {
@@ -1412,9 +1423,6 @@ public class LedgerHandle implements WriteHandle {
         };
         new TryReadLastConfirmedOp(this, clientCtx.getBookieClient(), getCurrentEnsemble(),
                                    innercb, getLastAddConfirmed()).initiate();
-} else {
-        asyncTryReadExplicitLastConfirmed(cb, ctx);
-}
     }
 
     /**
@@ -1672,6 +1680,7 @@ public class LedgerHandle implements WriteHandle {
     void asyncTryReadExplicitLastConfirmed(final ReadLastConfirmedCallback cb, final Object ctx) {
         boolean isClosed;
         synchronized (this) {
+            LedgerMetadata metadata = getLedgerMetadata();
             isClosed = metadata.isClosed();
             if (isClosed) {
                 lastAddConfirmed = metadata.getLastEntryId();
@@ -1696,7 +1705,7 @@ public class LedgerHandle implements WriteHandle {
                 }
             }
         };
-        new TryPendingReadLacOp(this, innercb).initiate();
+        new TryPendingReadLacOp(this, clientCtx.getBookieClient(), getCurrentEnsemble(), innercb).initiate();
     }
 
     /*
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
@@ -1335,6 +1335,7 @@ public class LedgerHandle implements WriteHandle {
      */
 
     public void asyncReadLastConfirmed(final ReadLastConfirmedCallback cb, final Object ctx) {
+        if (useLegacyMode) {
         boolean isClosed;
         long lastEntryId;
         synchronized (this) {
@@ -1360,6 +1361,9 @@ public class LedgerHandle implements WriteHandle {
             };
 
         new ReadLastConfirmedOp(this, clientCtx.getBookieClient(), getCurrentEnsemble(), innercb).initiate();
+        } else {
+        asyncReadExplicitLastConfirmed(cb, ctx);
+        }
     }
 
     /**
@@ -1377,6 +1381,7 @@ public class LedgerHandle implements WriteHandle {
      *          callback context
      */
     public void asyncTryReadLastConfirmed(final ReadLastConfirmedCallback cb, final Object ctx) {
+        if (legacyMode) {
         boolean isClosed;
         long lastEntryId;
         synchronized (this) {
@@ -1407,6 +1412,9 @@ public class LedgerHandle implements WriteHandle {
         };
         new TryReadLastConfirmedOp(this, clientCtx.getBookieClient(), getCurrentEnsemble(),
                                    innercb, getLastAddConfirmed()).initiate();
+} else {
+        asyncTryReadExplicitLastConfirmed(cb, ctx);
+}
     }
 
     /**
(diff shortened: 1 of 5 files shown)


eolivelli:
@sijie @jvrao can you take a look to this idea?

eolivelli:
@sijie @jvrao can you take a look to this idea?

sijie:
@eolivelli sorry I missed your comment. I typically ignore any WIP PRs...

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -1355,30 +1355,7 @@ synchronized void updateLastConfirmed(long lac, long len) {
...
-                        updateLastConfirmed(data.getLastAddConfirmed(), data.getLength());
-                        cb.readLastConfirmedComplete(rc, data.getLastAddConfirmed(), ctx);
-                    } else {
-                        cb.readLastConfirmedComplete(rc, INVALID_ENTRY_ID, ctx);
-                    }
-                }
-            };
-
-        new ReadLastConfirmedOp(this, innercb).initiate();
+        asyncReadExplicitLastConfirmed(cb, ctx);
```
eolivelli: @sijie here I am changing also *asyncReadLastConfirmed*

This is actually the core of the change.
I had to re implement *asyncTryReadLastConfirmed* because there is no existing support for explicitLAC

eolivelli:
@sijie  @jvrao I have rebased this patch onto current master.
I would like to insert it in 4.9.

I only have to add unit tests. All current tests about readLastAddConfirmed() and tryReadLastAddConfirmed() are passing even by enabling the feature by default.
So I only have to find interesting tests and transform them to a parametrized version which run the test with/without this new feature.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -1335,6 +1335,14 @@ synchronized void updateLastConfirmed(long lac, long len) {
...
 
     public void asyncReadLastConfirmed(final ReadLastConfirmedCallback cb, final Object ctx) {
+        if (clientCtx.getConf().useExplicitLacForReads) {
+            asyncReadExplicitLastConfirmed(cb, ctx);
+        } else {
+            asyncReadLastConfirmedInternal(cb, ctx);
+        }
+    }
+
+    private void asyncReadLastConfirmedInternal(final ReadLastConfirmedCallback cb, final Object ctx) {
```
sijie: asyncReadLastConfirmedInternal is a bad name. I would prefer calling it `asyncReadPiggybackLastConfirmed` (in contrast to `asyncReadExplicitLastConfirmed`.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -1335,6 +1335,14 @@ synchronized void updateLastConfirmed(long lac, long len) {
      */
 
     public void asyncReadLastConfirmed(final ReadLastConfirmedCallback cb, final Object ctx) {
+        if (clientCtx.getConf().useExplicitLacForReads) {
```
sijie: I am not sure using a flag is the best choice here. the bookkeeper client should be *smart* to figure out whether it should be reading explicit lac or piggybacked lac.

eolivelli: Maybe the name is misleading but it is already here.
This function will compose explicit and piggybacked LAC.
So actually it would be better to use always readExplicitLAC RPC, but we cannot enable it by default because it works only with recent bookies and with v3 protocol

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
eolivelli: Okay

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
jvrao: Client knows it is using V3 protocol or not; so why not use the explicitLAC for >= v3 automatically without a conf option?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/TryPendingReadLacOp.java
```diff
@@ -0,0 +1,152 @@
...
+                if (lacBuffer != null && lacBuffer.readableBytes() > 0) {
+                    long lac = lh.macManager.verifyDigestAndReturnLac(lacBuffer);
+                    if (lac > maxLac) {
+                        newLac = lac;
+                    }
+                }
+                // Extract lac from last entry on the disk
+                if (lastEntryBuffer != null && lastEntryBuffer.readableBytes() > 0) {
+                    RecoveryData recoveryData = lh.macManager.verifyDigestAndReturnLastConfirmed(lastEntryBuffer);
+                    long recoveredLac = recoveryData.getLastAddConfirmed();
```
jvrao: isn't it called piggyBackedLAC?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/TryPendingReadLacOp.java
```diff
@@ -0,0 +1,152 @@
+/**
```
jvrao: How is this different from PendingReadLacOp?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/TryPendingReadLacOp.java
```diff
@@ -0,0 +1,152 @@
...
+import org.apache.bookkeeper.net.BookieSocketAddress;
+import org.apache.bookkeeper.proto.BookieClient;
+import org.apache.bookkeeper.proto.BookkeeperInternalCallbacks.ReadLacCallback;
+import org.apache.bookkeeper.proto.checksum.DigestManager.RecoveryData;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This op is try to read last confirmed without involving quorum coverage checking.
+ * Use {@link PendingLacOp} if you need quorum coverage checking.
```
jvrao: You mean PendingReadLacOp?

jvrao: I am not sure if we ever need quorum for LAC. LAC means it has quorum, so we can take greatest LAC.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/TryPendingReadLacOp.java
```diff
@@ -0,0 +1,152 @@
...
+                    if (lac > maxLac) {
+                        newLac = lac;
+                    }
+                }
+                // Extract lac from last entry on the disk
+                if (lastEntryBuffer != null && lastEntryBuffer.readableBytes() > 0) {
+                    RecoveryData recoveryData = lh.macManager.verifyDigestAndReturnLastConfirmed(lastEntryBuffer);
+                    long recoveredLac = recoveryData.getLastAddConfirmed();
+                    if (recoveredLac > maxLac) {
+                        newLac = recoveredLac;
```
jvrao: You are overwriting newLac.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/TryPendingReadLacOp.java
jvrao: If we take out the coverageSet.checkCovered() check from PendingReadLacOp, then we don't need this right? I don't know why we have the quorum check in there, What scenario needs LAC  to be quorum checked? @sijie ?

ivankelly: Haven't looked at this patch, but this coverage check is there because fencing is piggybacked on the LAC read.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
eolivelli: ExplicitLAC is since 4.5. V3 protocol was already present. So we have bookies which talk v3 and do not know ExplicitLAC protocol: we will break compatibility from 4.9 clients and 4.4 bookies.
Honestly I don't think this is a real production problem because 4.4 is very old, but we don't have a clear EOL policy.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/TryPendingReadLacOp.java
eolivelli: @jvrao @ivankelly  @sijie 
This feature has the same semantics as TryReadLastConfirmedOp:

TryPendingReadLacOp -> PendingReadLacOp
TryReadLastConfirmedOp -> ReadLastConfirmedOp

We are trying to advance LAC and return as soon as we have an useful value, without full coverage checks.

AFAIK This feature is a used for Long-Poll reads, which I am not going to implement soon, at least not within 4.9 timeline.
In Long Poll reads tests I have found this API *asyncTryReadLastConfirmed*  and _for consistency_  I have created a similar one for ExplicitLAC based reads.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/TryPendingReadLacOp.java
eolivelli: renamed to newLac, it is clearer, thanks

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/TryPendingReadLacOp.java
eolivelli: good catch!

eolivelli:
For what is worth...
I am testing this patch with some projects and it works like a charm, on the writers you can enable ExplicitLAC and the readers are able to see the LAC advance without changing code (which is using readLastAddConfirmed and not readExplicitLastConfirmed).
No need to extra NOOP writes to make LAC advance, all is transparent.

**I think this is a great step forward for BK usability**

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/TryPendingReadLacOp.java
jvrao: > without full coverage checks.

This is what I am trying to understand. Once we have LAC updated (either piggy-back or explicit) it means that it has met Write Quorum. So I don't understand what is the additional coverage check needed. This is not a fencing read or reading beyond LAC, this operation reads LAC which must have been updated by the writer after write quorum, hence it is persisted.

eolivelli:
Ping @sijie

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
sijie: @eolivelli I don't understand your comment here. if a bookie doesn't know ExplicitLAC, the request will be rejected and the client will know about that, right? when client know that, it can fallback to disable using explicit lac, no?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -1377,6 +1385,14 @@ public void readLastConfirmedDataComplete(int rc, DigestManager.RecoveryData dat
      *          callback context
      */
     public void asyncTryReadLastConfirmed(final ReadLastConfirmedCallback cb, final Object ctx) {
+        if (clientCtx.getConf().useExplicitLacForReads) {
+            asyncTryReadExplicitLastConfirmed(cb, ctx);
+        } else {
+            asyncTryReadLastConfirmedInternal(cb, ctx);
```
sijie: Why the name here is not consistent with `asyncReadPiggybackLastConfirmed`?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
eolivelli: @sijie so you are suggesting to fallback to piggybackLac in case of "UnupportedOperation" error from bookie.
We should keep track of this per-bookie. I think this is very tricky, we need to add a lot of code and runtime overhead for the sake of compatibility with very old bookies (4.4).

We can have this configuration flag, enabled by default in 4.9 in case of v3+ protocol and disable for v2 protocol, and then drop it as soon as we decide to drop support for compatibility with such old versions.

eolivelli: @sijie  I see your concern. There is no hurry, we can work on this topic for 4.10.
I think that current asyncReadExplicitLastConfirmed() is better because it is actually merging the two values (piggy backed + explicit)

https://github.com/apache/bookkeeper/blob/e4265470881431fb7a78c8ab714c237239169e12/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadLacOp.java#L115

I think this is the future, I am not dropping the legacy method only for compatibility with 4.4.

The best would be to not have the flag, not perform any compatibility effort, and say that 4.9 clients are able to read from 4.4 bookies only using v2 protocol (in v2 protocol we don't have ExplicitLAC so we cannot use PendingReadLacOp)

PendingReadLacOp is already here, I am not adding it, I am only suggesting to use PendingReadLacOp for regular readLastAddConfirmed() instead of ReadLastConfirmedOp.

Ideally we should merge PendingReadLacOp with ReadLastConfirmedOp and drop PendingReadLacOp, but it is greater work, because ReadLastConfirmedOp handles fencing stuff

eolivelli:
@sijie  @jvrao  @ivankelly 
I am closing this one in favour of #1901 which is simpler and does not introduce new variants of read LAC.
We will deal with tryReadLastAddConfirmed in the future.
To me it is important to enable new API users to use ExplicitLAC

Issue #281, opened 2017-07-24
make ledger metadata immutable


sijie:
**FEATURE REQUEST**

1. Please describe the feature you are requesting.

Address the TODO item in *LedgerHandle*

                if (isClosed()) {
                    // TODO: make ledger metadata immutable
                    // Although the metadata is already closed, we don't need to proceed zookeeper metadata update, but
                    // we still need to error out the pending add ops.
                    //
                    // There is a race condition a pending add op is enqueued, after a close op reset ledger metadata state
                    // to unclosed to resolve metadata conflicts. If we don't error out these pending add ops, they would be
                    // leak and never callback.
                    //
                    // The race condition happen in following sequence:
                    // a) ledger L is fenced
                    // b) write entry E encountered LedgerFencedException, trigger ledger close procedure
                    // c) ledger close encountered metadata version exception and set ledger metadata back to open
                    // d) writer tries to write entry E+1, since ledger metadata is still open (reset by c))
                    // e) the close procedure in c) resolved the metadata conflicts and set ledger metadata to closed
                    // f) writing entry E+1 encountered LedgerFencedException which will enter ledger close procedure
                    // g) it would find that ledger metadata is closed, then it callbacks immediately without erroring out any pendings
                    synchronized (LedgerHandle.this) {
                        pendingAdds = drainPendingAddsToErrorOut();
                    }
                    errorOutPendingAdds(rc, pendingAdds);
                    cb.closeComplete(BKException.Code.OK, LedgerHandle.this, ctx);
                    return;
                }

2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have). Are you currently using any workarounds to address this issue?

*nice-to-have*

3. Provide any additional detail on your proposed use case for this feature.

sijie:
@ivankelly @jiazhai I marked this for 4.7.0 now. there have been discussions and concerns around the mutability of ledger metadata. let's pick this up in 4.7.0.

sijie:
I don't think we are able to do this in any time soon. dropping the release version at this moment.

ivankelly:
ya, not a hope for 4.7.0. This needs yahoo and saleforce merge to be 100% complete. We're close, but better to postpone til 4.8.0.

I remember when I did it last time, the change only took about a day.

ivankelly:
@reddycharan @jvrao Are salesforce planning to make any changes to the client metadata handling?

jvrao:
[...]

PR #1573, opened 2018-07-31
LedgerManager should return the metadata just written

ivankelly:
First part of changes to make ledger metadata immutable. The client
should only act on metadata which has been written to zookeeper. To
this end, we need to be able to get back from the ledger manager what
has been written to zookeeper (with the version number updated).

Master issue: #281

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractZkLedgerManager.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractZkLedgerManager.java
@@ -245,7 +245,7 @@ public abstract class AbstractZkLedgerManager implements LedgerManager, Watcher 
 
     @Override
     public void createLedgerMetadata(final long ledgerId, final LedgerMetadata metadata,
-            final GenericCallback<Void> ledgerCb) {
+            final GenericCallback<LedgerMetadata> ledgerCb) {
         String ledgerPath = getLedgerPath(ledgerId);
         StringCallback scb = new StringCallback() {
             @Override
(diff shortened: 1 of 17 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractZkLedgerManager.java
```diff
@@ -245,15 +245,15 @@ public void process(WatchedEvent event) {
...
+            final GenericCallback<LedgerMetadata> ledgerCb) {
         String ledgerPath = getLedgerPath(ledgerId);
         StringCallback scb = new StringCallback() {
             @Override
             public void processResult(int rc, String path, Object ctx, String name) {
                 if (rc == Code.OK.intValue()) {
                     // update version
                     metadata.setVersion(new LongVersion(0));
-                    ledgerCb.operationComplete(BKException.Code.OK, null);
+                    ledgerCb.operationComplete(BKException.Code.OK, metadata);
```
eolivelli: So this is the core of the change. I was missing it in the diff.

ivankelly: It doesn't really do anything right now, as it's just returning what was passed in. Once immutable metadata is there, this will be something like.

```
LedgerMetadata writtenMetadata = LedgerMetadataBuilder.from(metadata).withVersion(new LongVersion(0)).build();
ledgerCb.operationComplete(BKException.Code.OK, writtenMetadata);
````

ivankelly: There's a load of boring changes for this. I'm trying to split out into smaller patches to make the eventually meaty changes easier to review.

eolivelli: Sure, I see. I just marked this line as it is the only one which does something really useful in this change.

ivankelly: There's also one in the MSLedgerManager, but ya, that's the only slightly interesting part

eolivelli: Oh, there are a couple of similar lines.
Please go ahead. These were mostly comments for myself.

eolivelli:
To me as soon as CI is green we can merge this one. It won't create any harm

PR #1574, opened 2018-07-31
Disallow direct access to LedgerHandle#metadata

ivankelly:
This object has been accessed and mutated all over the client, which
makes it hard to do anything with the object. This patch removes the
direct accesses, so the object can only be accessed through an
accessor.

Master issue: #281

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
@@ -102,7 +102,7 @@ public class LedgerHandle implements WriteHandle {
     static final long PENDINGREQ_NOTWRITABLE_MASK = 0x01L << 62;
 
     final byte[] ledgerKey;
-    LedgerMetadata metadata;
+    private LedgerMetadata metadata;
     final BookKeeper bk;
     final long ledgerId;
     long lastAddPushed;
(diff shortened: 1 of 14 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -333,13 +335,24 @@ public LedgerMetadata getLedgerMetadata() {
         return metadata;
     }
 
+    boolean setLedgerMetadata(LedgerMetadata expected, LedgerMetadata newMetadata) {
+        synchronized (this) {
+            if (metadata == expected) {
```
eolivelli: add a comment ?
usually '==' seems an error in Java, but in this point we are doing it because it is expected

ivankelly: I've never heard of '==' being code smell. This code is standard check and set. Can add a comment though.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadLastConfirmedOp.java
```diff
@@ -52,22 +52,22 @@ public ReadLastConfirmedOp(LedgerHandle lh, LastConfirmedDataCallback cb) {
...
-        this.numResponsesPending = lh.metadata.getEnsembleSize();
+        this.numResponsesPending = lh.getLedgerMetadata().getEnsembleSize();
         this.coverageSet = lh.distributionSchedule.getCoverageSet();
     }
 
     public void initiate() {
-        for (int i = 0; i < lh.metadata.currentEnsemble.size(); i++) {
-            lh.bk.getBookieClient().readEntry(lh.metadata.currentEnsemble.get(i),
+        for (int i = 0; i < lh.getLedgerMetadata().currentEnsemble.size(); i++) {
+            lh.bk.getBookieClient().readEntry(lh.getLedgerMetadata().currentEnsemble.get(i),
```
eolivelli: isn't it better to "campture" lh.getLedgerMetadata() once ?
maybe it is guaranteed somewhere else that calling getLedgerMetadata twice will return exactly the same object because we are using safeOrderedExecutor, but it is not very clear just by reading this code

ivankelly:
rerun integration test

ivankelly:
rerun integration tests

ivankelly:
merging so I can push the next

eolivelli:
@ivankelly seems we have compilation errors, check CI

apart from that this patch can go for me, as you will soon re-refactor the code in which I left comments

PR #1575, opened 2018-07-31
Make each ensemble in ensemble list immutable

ivankelly:
Previously, the ensemble list was a Map<Long, ArrayList<BookieSocketAddress>>.
ArrayList is by definition mutable, so ensemble passed to metadata users are
always mutable.

This patch changes in ensembles in the list to be immutable. We were also leaking
the implementation of ledger metadata to the placement policy, so this has been
modified to use List<BookieSocketAddress> also.

Master issue: #281

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerFragmentReplicator.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerFragmentReplicator.java
@@ -26,7 +26,6 @@ import io.netty.buffer.Unpooled;
 import java.util.ArrayList;
 import java.util.Enumeration;
 import java.util.HashSet;
-import com.google.common.collect.ImmutableList;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
@@ -387,8 +386,7 @@ public class LedgerFragmentReplicator {
                 LOG.info("Bookie {} doesn't exist in ensemble {} anymore.", entry.getKey(), ensemble);
             }
         }
-        List<BookieSocketAddress> immutable = ImmutableList.copyOf(newEnsemble);
-        lh.getLedgerMetadata().updateEnsemble(fragmentStartId, immutable);
+        lh.getLedgerMetadata().updateEnsemble(fragmentStartId, newEnsemble);
         lh.writeLedgerConfig(new UpdateEnsembleCb(ensembleUpdatedCb,
                 fragmentStartId, lh, oldBookie2NewBookie));
     }
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerFragmentReplicator.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerFragmentReplicator.java
@@ -26,6 +26,7 @@ import io.netty.buffer.Unpooled;
 import java.util.ArrayList;
 import java.util.Enumeration;
 import java.util.HashSet;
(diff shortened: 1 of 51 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerFragmentReplicator.java
```diff
@@ -374,18 +375,20 @@ private static void updateEnsembleInfo(
...
             // change. if ensemble is already changed, skip replacing the bookie doesn't exist.
             if (deadBookieIndex >= 0) {
-                ensemble.set(deadBookieIndex, entry.getValue());
+                newEnsemble.set(deadBookieIndex, entry.getValue());
             } else {
                 LOG.info("Bookie {} doesn't exist in ensemble {} anymore.", entry.getKey(), ensemble);
             }
         }
+        List<BookieSocketAddress> immutable = ImmutableList.copyOf(newEnsemble);
+        lh.getLedgerMetadata().updateEnsemble(fragmentStartId, immutable);
```
eolivelli: So you are updating the internals of LedgerMetadata here?
I expected this something like:
newMetadata = lh.getLedgerMetadata().updateEnsemble(....)

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
```diff
@@ -707,29 +714,29 @@ private String toStringRepresentation(boolean withPassword) {
...
+        for (Entry<Long, ? extends List<BookieSocketAddress>> entry : newEnsembles.entrySet()) {
             ++i;
             if (ensembles.size() != i) {
                 // we should use last ensemble from current metadata
                 // not the new metadata read from zookeeper
                 long key = entry.getKey();
-                ArrayList<BookieSocketAddress> ensemble = entry.getValue();
-                ensembles.put(key, ensemble);
+                List<BookieSocketAddress> ensemble = entry.getValue();
+                ensembles.put(key, ImmutableList.copyOf(ensemble));
```
eolivelli: Why not using Collections.unmodifiableList ? Which will create a wrapper. In any case you are not creating here a 'deep copy' (clone each element:

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerFragmentReplicator.java
ivankelly: Yes, eventually, though it'll be more a builder pattern.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
ivankelly: oh, I'll start using that. This line of code is going to disappear fairly soon anyhow.

Regarding deep copy, BookieSocketAddress is immutable.

ivankelly:
retest this please // making sure fails aren't environmental (ran cleanly locally)

eolivelli:
@ivankelly seems we have compilation errors, check CI

apart from that this patch can go for me, as you will soon re-refactor the code in which I left comments

eolivelli:
@ivankelly it seemed to me that there were also compilation errors, not only javac, let's see CI

ivankelly:
@eolivelli checking now

ivankelly:
ah, dlog was messing with internals

PR #1582, opened 2018-08-03
Enhancing DecommissionBookieCmd

reddycharan:

- adding bookieid option to DecommissionBookieCmd, to enable decommissioning remote
bookie. This might be needed because, in an extreme scenario Bookie node might not
be completely accessible and hence we might need to decommission a Bookie remotely.
- also DecommissionBookieCmd should delete the cookie of the decommissioned Bookie.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
@@ -2511,7 +2511,7 @@ public class BookieShell implements Tool {
 
         @Override
         String getUsage() {
-            return "decommissionbookie [-bookieid <bookieaddress>]";
+            return CMD_DECOMMISSIONBOOKIE + " [-bookieid <bookieaddress>]";
         }
 
         @Override
@@ -2529,14 +2529,16 @@ public class BookieShell implements Tool {
                         .isBlank(remoteBookieidToDecommission) ? Bookie.getBookieAddress(bkConf)
                                 : new BookieSocketAddress(remoteBookieidToDecommission));
                 admin.decommissionBookie(bookieAddressToDecommission);
-                LOG.info("The ledgers stored in the given decommissioning bookie are properly replicated");
+                LOG.info("The ledgers stored in the given decommissioning bookie: {} are properly replicated",
+                        bookieAddressToDecommission);
                 runFunctionWithRegistrationManager(bkConf, rm -> {
                     try {
                         Versioned<Cookie> cookie = Cookie.readFromRegistrationManager(rm, bookieAddressToDecommission);
                         cookie.getValue().deleteFromRegistrationManager(rm, bookieAddressToDecommission,
                                 cookie.getVersion());
                     } catch (CookieNotFoundException nne) {
-                        LOG.warn("No cookie to remove, it could be deleted already : ", nne);
+                        LOG.warn("No cookie to remove for the decommissioning bookie: " + bookieAddressToDecommission
+                                + ", it could be deleted already : ", nne);
                     } catch (BookieException be) {
                         throw new UncheckedExecutionException(be.getMessage(), be);
                     }
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
@@ -2490,25 +2490,28 @@ public class BookieShell implements Tool {
     }
 
     /**
-     * Command to trigger AuditTask by resetting lostBookieRecoveryDelay and then make sure the
-     * ledgers stored in the bookie are properly replicated.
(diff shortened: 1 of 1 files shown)


reddycharan:
@jvrao @dlg99 can you take a look

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
```diff
@@ -2490,25 +2490,28 @@ public int runCmd(CommandLine cmdLine) throws Exception {
...
         String getDescription() {
             return "Force trigger the Audittask and make sure all the ledgers stored in the decommissioning bookie"
-                    + " are replicated.";
+                    + " are replicated and cookie of the decommissioned bookie is deleted from metadata server.";
         }
 
         @Override
         String getUsage() {
-            return CMD_DECOMMISSIONBOOKIE;
+            return "decommissionbookie [-bookieid <bookieaddress>]";
```
sijie: `CMD_DECOMMISSIONBOOKIE + " [-bookieid <bookieaddress>]"`

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
```diff
@@ -2521,8 +2524,26 @@ public int runCmd(CommandLine cmdLine) throws Exception {
...
             BookKeeperAdmin admin = new BookKeeperAdmin(adminConf);
             try {
-                BookieSocketAddress thisBookieAddress = Bookie.getBookieAddress(bkConf);
-                admin.decommissionBookie(thisBookieAddress);
+                final String remoteBookieidToDecommission = cmdLine.getOptionValue("bookieid");
+                final BookieSocketAddress bookieAddressToDecommission = (StringUtils
+                        .isBlank(remoteBookieidToDecommission) ? Bookie.getBookieAddress(bkConf)
+                                : new BookieSocketAddress(remoteBookieidToDecommission));
+                admin.decommissionBookie(bookieAddressToDecommission);
+                LOG.info("The ledgers stored in the given decommissioning bookie are properly replicated");
```
sijie: add the `bookieAddressToDecommission` to the logging statement.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
```diff
@@ -2521,8 +2524,26 @@ public int runCmd(CommandLine cmdLine) throws Exception {
...
+                runFunctionWithRegistrationManager(bkConf, rm -> {
+                    try {
+                        Versioned<Cookie> cookie = Cookie.readFromRegistrationManager(rm, bookieAddressToDecommission);
+                        cookie.getValue().deleteFromRegistrationManager(rm, bookieAddressToDecommission,
+                                cookie.getVersion());
+                    } catch (CookieNotFoundException nne) {
+                        LOG.warn("No cookie to remove, it could be deleted already : ", nne);
```
sijie: add the bookieAddressToDecommission to the logging statement.

sijie:
overall looks good. some minor comments on logging.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
```diff
@@ -2529,14 +2529,16 @@ public int runCmd(CommandLine cmdLine) throws Exception {
...
+                LOG.info("The ledgers stored in the given decommissioning bookie: {} are properly replicated",
+                        bookieAddressToDecommission);
                 runFunctionWithRegistrationManager(bkConf, rm -> {
                     try {
                         Versioned<Cookie> cookie = Cookie.readFromRegistrationManager(rm, bookieAddressToDecommission);
                         cookie.getValue().deleteFromRegistrationManager(rm, bookieAddressToDecommission,
                                 cookie.getVersion());
                     } catch (CookieNotFoundException nne) {
-                        LOG.warn("No cookie to remove, it could be deleted already : ", nne);
+                        LOG.warn("No cookie to remove for the decommissioning bookie: " + bookieAddressToDecommission
```
sijie: can you change it to use `{}`

reddycharan: @sijie considered about it, but in the Logger class there is no warn method which takes Object arg and Throwable t as arguments (I mean void warn(String format, Object arg1, Throwable t)). The only warn method which takes Throwable as argument is this - https://www.slf4j.org/api/org/slf4j/Logger.html#warn(java.lang.String,%20java.lang.Throwable)

sijie: https://www.slf4j.org/api/src-html/org/slf4j/Logger.html#line.509 doesn't this work for you?

reddycharan: changed as we discussed.

PR #1583, opened 2018-08-05
WIP - Add JaCoco agent in stream module when using code-coverage profile

eolivelli:
This is a WIP patch to force executing JaCoCo agent on 'stream' module.
Stream module pom.xml overrides surefire configuration from parent pom and so it is hiding special configuration parameters activated usually by the code-coverage profile.

### DIFF
--- a/stream/pom.xml
+++ b/stream/pom.xml
@@ -85,7 +85,25 @@
         </plugins>
       </build>
     </profile>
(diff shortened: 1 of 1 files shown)


eolivelli:
Closing this patch. Not needed as we will drop such profiles in 4.9

Issue #1584, opened 2018-08-06
LedgerHandleAdv should expose asyncAddEntry variant that takes ByteBuf (LedgerHandle exposes it as public)


dlg99:
**FEATURE REQUEST**

1. Please describe the feature you are requesting.

LedgerHandle.java:
```java
    public void asyncAddEntry(ByteBuf data, final AddCallback cb, final Object ctx) {
...
```

LedgerHandleAdv.java:

```java
    private void asyncAddEntry(final long entryId, ByteBuf data,
            final AddCallbackWithLatency cb, final Object ctx) {
...
```

It's useful to have this exposed as public for clients to make use of netty's allocator

2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have). Are you currently using any workarounds to address this issue?

must have

3. Provide any additional detail on your proposed use case for this feature.

I'll send a PR soon.

PR #1585, opened 2018-08-06
Issue #1584: LedgerHandleAdv should expose asyncAddEntry variant that takes ByteBuf (LedgerHandle exposes it as public)

dlg99:

- exposed asyncAddEntry as public, similarly to other variants.
- fixed ByteBuf retention

### Motivation

It's useful to have this exposed as public for clients to make use of netty's allocator and pass ByteBuf directly.

### Changes

exposed asyncAddEntry as public, similarly to other variants.

Master Issue: #1584 

> ---
>
> ---

> ---
> quickly and easily:
>
>
>
> Otherwise:
> 
> 
> ---

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
@@ -1126,6 +1126,26 @@ public class LedgerHandle implements WriteHandle {
     }
 
     /**
+     * Add entry asynchronously to an open ledger, using an offset and range.
+     * This can be used only with {@link LedgerHandleAdv} returned through
+     * ledgers created with {@link createLedgerAdv(int, int, int, DigestType, byte[])}.
+     *
+     * @param entryId
+     *            entryId of the entry to add.
+     * @param data
+     *            io.netty.buffer.ByteBuf of bytes to be written
+     * @param cb
+     *            object implementing callbackinterface
+     * @param ctx
+     *            some control object
+     */
+    public void asyncAddEntry(final long entryId, ByteBuf data,
+                              final AddCallbackWithLatency cb, final Object ctx) {
+        LOG.error("To use this feature Ledger must be created with createLedgerAdv() interface.");
+        cb.addCompleteWithLatency(BKException.Code.IllegalOpException, LedgerHandle.this, entryId, 0, ctx);
+    }
+
+    /**
      * {@inheritDoc}
      */
     @Override
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
@@ -1067,7 +1067,6 @@ public class LedgerHandle implements WriteHandle {
     }
 
     public void asyncAddEntry(ByteBuf data, final AddCallback cb, final Object ctx) {
-        data.retain();
         PendingAddOp op = PendingAddOp.create(this, data, writeFlags, cb, ctx);
         doAsyncAddEntry(op);
     }
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
@@ -92,7 +92,6 @@ class PendingAddOp extends SafeRunnable implements WriteCallback {
         op.entryId = LedgerHandle.INVALID_ENTRY_ID;
         op.currentLedgerLength = -1;
         op.payload = payload;
-        op.payload.retain();
         op.entryLength = payload.readableBytes();
 
         op.completed = false;
@@ -244,6 +243,8 @@ class PendingAddOp extends SafeRunnable implements WriteCallback {
         this.toSend = lh.macManager.computeDigestAndPackageForSending(
                 entryId, lh.lastAddConfirmed, currentLedgerLength,
                 payload);
+        // ownership of RefCounted ByteBuf was passed to computeDigestAndPackageForSending
+        payload = null;
 
         // We are about to send. Check if we need to make an ensemble change
         // becasue of delayed write errors
@@ -457,8 +458,10 @@ class PendingAddOp extends SafeRunnable implements WriteCallback {
     private void recyclePendAddOpObject() {
         entryId = LedgerHandle.INVALID_ENTRY_ID;
         currentLedgerLength = -1;
-        ReferenceCountUtil.release(payload);
-        payload = null;
+        if (payload != null) {
+            ReferenceCountUtil.release(payload);
+            payload = null;
+        }
         cb = null;
         ctx = null;
         ackSet.recycle();
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
@@ -92,6 +92,7 @@ class PendingAddOp extends SafeRunnable implements WriteCallback {
         op.entryId = LedgerHandle.INVALID_ENTRY_ID;
         op.currentLedgerLength = -1;
         op.payload = payload;
+        op.payload.retain();
         op.entryLength = payload.readableBytes();
 
         op.completed = false;
@@ -456,6 +457,7 @@ class PendingAddOp extends SafeRunnable implements WriteCallback {
     private void recyclePendAddOpObject() {
         entryId = LedgerHandle.INVALID_ENTRY_ID;
         currentLedgerLength = -1;
+        ReferenceCountUtil.release(payload);
         payload = null;
         cb = null;
         ctx = null;
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieWriteLedgerTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieWriteLedgerTest.java
@@ -1335,7 +1335,12 @@ public class BookieWriteLedgerTest extends
             int rc = cf.get();
             assertEquals("rc code is OK", BKException.Code.OK, rc);
 
-            TimeUnit.MILLISECONDS.sleep(250); // recycler runs asynchronously
+            for (int i =0; i < 10; i++) {
+                if (data.refCnt() == 0) {
+                    break;
+                }
+                TimeUnit.MILLISECONDS.sleep(250); // recycler runs asynchronously
+            }
             assertEquals("writing entry with id " + entryId + ", ref count on ByteBuf should be 0 ",
                     0, data.refCnt());
 
@@ -1373,7 +1378,12 @@ public class BookieWriteLedgerTest extends
             int rc = cf.get();
             assertEquals("rc code is OK", BKException.Code.OK, rc);
 
-            TimeUnit.MILLISECONDS.sleep(250); // recycler runs asynchronously
+            for (int i =0; i < 10; i++) {
+                if (data.refCnt() == 0) {
+                    break;
+                }
+                TimeUnit.MILLISECONDS.sleep(250); // recycler runs asynchronously
+            }
             assertEquals("writing entry with id " + entryId + ", ref count on ByteBuf should be 0 ",
                     0, data.refCnt());
 
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieWriteLedgerTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieWriteLedgerTest.java
@@ -34,6 +34,7 @@ import io.netty.buffer.AbstractByteBufAllocator;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.PooledByteBufAllocator;
 import io.netty.buffer.Unpooled;
+import io.netty.buffer.UnpooledByteBufAllocator;
 
 import java.io.IOException;
 import java.nio.ByteBuffer;
@@ -50,7 +51,6 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
-import io.netty.buffer.UnpooledByteBufAllocator;
 import org.apache.bookkeeper.bookie.Bookie;
 import org.apache.bookkeeper.bookie.BookieException;
 import org.apache.bookkeeper.client.AsyncCallback.AddCallback;
@@ -1350,7 +1350,7 @@ public class BookieWriteLedgerTest extends
     @Test
     @SuppressWarnings("unchecked")
     public void testLedgerCreateByteBufRefCnt() throws Exception {
-        final LedgerHandle lh = bkc.createLedger(5, 3, 2, digestType, ledgerPassword, null);;
+        final LedgerHandle lh = bkc.createLedger(5, 3, 2, digestType, ledgerPassword, null);
 
         final List<AbstractByteBufAllocator> allocs = Lists.newArrayList(
                 new PooledByteBufAllocator(true),
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieWriteLedgerTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieWriteLedgerTest.java
@@ -28,7 +28,11 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
+
+import com.google.common.collect.Lists;
+import io.netty.buffer.AbstractByteBufAllocator;
 import io.netty.buffer.ByteBuf;
+import io.netty.buffer.PooledByteBufAllocator;
 import io.netty.buffer.Unpooled;
 
 import java.io.IOException;
@@ -43,8 +47,10 @@ import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
+import io.netty.buffer.UnpooledByteBufAllocator;
 import org.apache.bookkeeper.bookie.Bookie;
 import org.apache.bookkeeper.bookie.BookieException;
 import org.apache.bookkeeper.client.AsyncCallback.AddCallback;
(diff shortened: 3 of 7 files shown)


dlg99:
@jvrao @sijie @eolivelli FYI

dlg99:
@eolivelli somewhat related topic, I noticed that LedgerHandle.java does data.retain there:

 ```java
    public void asyncAddEntry(ByteBuf data, final AddCallback cb, final Object ctx) {
        data.retain();
        PendingAddOp op = PendingAddOp.create(this, data, writeFlags, cb, ctx);
        doAsyncAddEntry(op);
    }
```

As I understand, netty assumes that ownership is passed to the caller and it is up to callee to retain object if it needs it after passing down. I.e. LedgerHandleAdv's asyncAddEntry variant does not call retain(). One of those needs to be fixed, or both if it is actually PendingAddOp.create who needs this retained.  
See http://netty.io/wiki/reference-counted-objects.html#who-destroys-it

**Comment by eolivelli** [comment]
@dlg99 regarding you comment... I will check today

**Comment by sijie** [comment]
1) the method is already there. So this PR only changes private to public.

2) it takes time for people to switching to use new APIs. And most of the production systems is using ledgerhandle, so since it is not adding brand new features, I don’t think it is problem for changing a private method to a public method.

3) the new api is still unstable. So the existing api still has values. We have to first switch internal usage to new API before getting rid of the old api. Until then, we have to support it since it is used widely in production.

**Comment by eolivelli** [comment]
@sijie I am not against this change.
Just wanted to understand.
The 'write' method is available on the old API as well.

**Comment by eolivelli** [comment]
@dlg99 you are right. Thank you very much

asyncAddEntry MUST call data.retain(), Not calling data.retain() will lead to a double release.

Currently the asyncAddEntry method is called by 2 asyncAddEntry methods which are wrapping a byte[], so ref counting is not really an issue.

It is also used in writeAsync, and this is not calling ByteBuf.retain, and so this is a bug.

I am currently using writeAsync  in an application, but with a wrapped byte[] so I didn't notice the bug.

Please add data.retain() in  the method you are changing, so it will handle correctly ref counts and which will also fix writeAsync.

We can also cherry pick this change to 4.7 branch

**Comment by sijie** [comment]
can we have a test case on reproducing the problem @eolivelli is describing?

**Comment by eolivelli** [comment]
Yep a test case would be very useful.

**Comment by dlg99** [comment]
@eolivelli @sijie I added unit tests and fixed data retention, please take a look.

@eolivelli problem with asyncAddEntry(ByteBuf..) is that LH has it as public (ok).
But for each LH's public method that adds an entry we have an override that takes entryId as a parameter and its implementation in LHAdv. That part was missing, specifically LHAdv's implementation was private and LH did not have that override. I would not consider this a breaking change of the API but rather a fix of inconsistency.

**Comment by dlg99** [comment]
@sijie some dlog's tests failed.
As far as I can tell now, I fixed them. At least the ones that failed and that ran on-by-one from IDE.
simple "cd bookkeeper/stream/distributedlog/core; mvn test" does not seem to run any tests (including doing it from higher level folders). I tried "mvn -Dstream test" - does not seem to run the tests as well.
Running them from IDE (intellij) results in netty's exception OutOfDirectMemory in majority of tests, that includes main branch without any changes so I did not troubleshoot this + I am not familiar with dlog.
I'll check on the build results.

**Comment by sijie** [comment]
@dlg99 thank you for looking into into. it seems tests are passing. so it should be fine.

File: bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java [hunk 1]
```diff
@@ -1067,7 +1067,6 @@ public void asyncAddEntry(final byte[] data, final int offset, final int length,
     }
 
     public void asyncAddEntry(ByteBuf data, final AddCallback cb, final Object ctx) {
-        data.retain();
```
sijie: one thing I am not very clear at this change is : 

if the change is move `data.retain` to PendingAddOp.create, why do we need to changes below?

- why do we need to do `ReferenceCountUtil.release(payload);` in PendingAddOp?
- why do we need to release in DigestManager?

I am a bit confused now. My biggest worry is if we need to change dlog code, that might mean it might potentially break other bookkeeper client.

Can you clarify the change here? I feel I am missing something here.

eolivelli: @sijie  @dlg99 
I think it is safer to not change PendingAddOp and DigestManager.
We only need to add a data.retain() in the asyncAddEntry private method that Andrey is changing to public, nothing more.
The change that I am proposing will not touch any other code path.

Current code is running very well with v2 protocol on the LedgerHandle (non-Adv I mean) and so we are sure that reference counting is done very well.
in the case of LedgerHandleAdv we never exposed the ByteBuf API and we always used ByteBuf which were wrappers for byte[] so they are not reference counted.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieWriteLedgerTest.java
```diff
@@ -1294,6 +1300,91 @@ public void testLedgerCreateAdvSyncAsyncAddDuplicateEntryIds() throws Exception
...
+            CompletableFuture<Integer> cf = new CompletableFuture<>();
+            lh.asyncAddEntry(entryId, data, (rc, handle, eId, qwcLatency, ctx) -> {
+                CompletableFuture<Integer> future = (CompletableFuture<Integer>) ctx;
+                future.complete(rc);
+            }, cf);
+
+            int rc = cf.get();
+            assertEquals("rc code is OK", BKException.Code.OK, rc);
+
+            TimeUnit.MILLISECONDS.sleep(250); // recycler runs asynchronously
```
eolivelli: @dlg99 
this will be our next flaky test in the list.

during 4.5 development I had added some tests about ref counts but they turned to be quite flaky

Maybe it is better not to add assertions but only see that the asyncAddEntry completes with success.
I understand that this way we are not testing the whole change but I guess we will save us from debugging a new flaky test.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/DeferredSyncTest.java
```diff
@@ -50,9 +50,9 @@ public void testAddEntryLastAddConfirmedDoesNotAdvance() throws Exception {
                 .withWriteFlags(WriteFlag.DEFERRED_SYNC)
                 .execute())) {
             for (int i = 0; i < NUM_ENTRIES - 1; i++) {
-                result(wh.appendAsync(DATA));
+                result(wh.appendAsync(DATA.retainedDuplicate()));
```
eolivelli: In theory this code is safe and it is an "application" that is calling this API as expected.

the fact the you have changed this test is an hint about an error which you are introducing.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
dlg99: This thing grew a bit out of "let's make method public" change ;)

I'll start with DigestManager:
it handles V2 and V3 protocol.
For V2 protocol it returns ByteBufList which wraps received ByteBuf data and needs it until ByteBufList is released.
ByteBufList (aka toSend object in PendingAddOp) is released when callback is triggered. On ByteBufList release it callse .release() on all its parts, including ByteBuf data.

For V3 protocol currently it copies content into new unpooled ByteBuf, original data is no longer needed and no longer used so it should be released. This keeps behavior consistent for v2 and v3 in regards of ownership expectations of data (as in: computeDigestAndPackageForSending gets ownership of 1 refCnt of the passed ByteBuf and releases it when data is no longer needed or passes to a component that will do it).

This matches netty's "general rule of thumb is that the party who accesses a reference-counted object lastly is responsible for the destruction of the reference-counted object."

PendingAddOp keeps reference to payload.
Looking closer at it, there is no need to retain data unless we decide to i.e. log it after callback triggered and toSend was released. I'll actually remove it and re-run bookie tests locally, probably will push the change.
in this case payload's release in PendingAddOp should happen in recycle() only if payload was never used (hasRun == false). 

If I understood correctly, retain() call was added without accompanying release to fix failing test(s), i.e. DeferredSyncTest. This works with i.e. buffers produced by Unpooled.wrappedBuffer(..) but not when one tries to use Pooled allocator where unreleased reference will lead to a leak. I approached it by changing the test.

I understand your concerns about dlog. I propose the following:

1. Let's agree on the ownership rules for the ByteBufs and fix the client to use these rules consistently.

2. Let's review dlog's code. Any API calls in write path that pass ByteBuf directly are suspects, as well as parts where .slice() or .duplicate() is called. Tests caught one place where ref count had to be incremented. The rule of thumb is if ByteBuf is used after being passed somewhere, it has to be retained.
I am not familiar with dlog's codebase so I'll need help. We can add asserts on recCnt to existing tests, similar to ones in the new tests I added.
Existing tests caught few cases of reusing released ByteBuf (EnvelopedEntryWriter.getBuffer() ) and I did not find any other suspects after brief look at the dlog's code.
We can try enabling paranoid leak detection in unit tests, but at least initially it will catch false positives related to i.e. keeping references like
static final ByteBuf DATA = Unpooled.wrappedBuffer("foobar".getBytes()); 
and never releasing them at the end of the test. These won't be hard to fix but tedious.

3. We can add a warning to release notes.

eolivelli: @dlg99 thank you for taking care of this.

I took a deeper look ad PendingAddOp#payload.
It should release *payload* after computeDigestAndPackageForSending and as you are saying it should release *payload* in case of hasRun == false.

So in order to make it clearer I suggest adding release in PendingAddOp and not in DigestManager. This way it is clear how  PendingAddOp is changing reference counts.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/DeferredSyncTest.java
eolivelli: @dlg99  regarding this test I think that the application in this case is already retaining *once* the DATA ByteBuf, so the test is not to be changed.

same for DL

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
sijie: @dlg99 

I haven't fully read your reply yet. but just to clarify what I was saying.

I uses dlog as an example. I am kind of less worried about dlog, since it is in the same repo and have all the tests running to make sure the change is correct at some extend. 

I am more worried about the applications that use bk that isn't in this repo, for example pulsar. it is a bit hard to tell what would the impact to those applications. it sounds like those applications have to change their behavior around reference counting. That's why I was worried this change become a behavior breaking change.

sijie:
@merlimat can you please spend sometime on reviewing this? this change might require some changes in pulsar managed ledger. but I think the change here is correct and make the ownership of ByteBuf much clearer. so take a look and lets try to see if we can agree on this rule.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
```diff
@@ -92,6 +92,7 @@ static PendingAddOp create(LedgerHandle lh, ByteBuf payload, EnumSet<WriteFlag>
         op.entryId = LedgerHandle.INVALID_ENTRY_ID;
         op.currentLedgerLength = -1;
         op.payload = payload;
+        op.payload.retain();
```
sijie: @dlg99 the change looks clear to me. but I would suggest lets add more comments around this `payload` field (line #60). basically payload should not be used anymore after it is computed to `toSend`, because the digest manager will either pass the reference count to `toSend` (for v2) and be released (by v3).

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
```diff
@@ -456,6 +457,7 @@ private void maybeRecycle() {
     private void recyclePendAddOpObject() {
         entryId = LedgerHandle.INVALID_ENTRY_ID;
         currentLedgerLength = -1;
+        ReferenceCountUtil.release(payload);
```
sijie: I think payload should only be released if `toSend` is not computed. if `toSend` is computed, the ref cnt of payload is either passed on to `toSend` (v2) or released (v3). unless I missed something from your last comment.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
dlg99: @sijie one way or another we break something. 
We had inconsistency between LH and LHAdv.
This is new API overall and maintaining bug-compatibility feels like an overhead. Fixing it later won't be easier.
I'd love to make it consistent between LH and LHAdv, and consistent with netty's recommendations.
as alternative, I can slap .retain() call back into LH.asyncAddEntry to maintain that compatibility.
I'd very much like to not have to use it with LHAdv's variant because it would mean that I need to remember to release buffer (as in: try/finally around everything) after each call unlike the rest of the netty calls. But that means keeping inconsistency which will bite us eventually.

I suggest we sleep on it and decide tomorrow.
I'll look at Enrico's and your comments tomorrow and will address everything that does not touch subject of ref-cnt ownership.
Maybe we need to chat, like in slack tomorrow or I'll do my best to call in to community meeting this Thursday.

eolivelli:
I am marking this PR as blocker for 4.8, we have to fix it now

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/DeferredSyncTest.java
dlg99: @eolivelli assuming we agree that "the party who accesses a reference-counted object lastly is responsible for the destruction of the reference-counted object":

static final ByteBuf DATA = Unpooled.wrappedBuffer("foobar".getBytes());
created DATA with refCnt == 1

1st call to
result(wh.appendAsync(DATA));
will work ok, take ownership of 1 refCnt of DATA and will reduce DATA's refCnt to 0 when no longer needs it.
2nd call to 
result(wh.appendAsync(DATA));
will fail.

It is up to the calling method to decide whether object has to be retained and reused in caller's context hence 
result(wh.appendAsync(DATA.retainedDuplicate()));
Could as well do
DATA.retain();
result(wh.appendAsync(DATA));

Normally at the end of the test we should release DATA. As far as I see, netty recommends for unit tests the following:
static final ByteBuf DATA = ReferenceCountUtil. releaseLater( Unpooled.wrappedBuffer("foobar".getBytes()) );
to release it automatically.

Now let's imagine that appendAsync does not take ownership but calls retain() internally. 
Now in simple prod case when data is not used after write need to do something like that:

ByteBuf data = PooledAllocator.DEFAULT.buffer(size);
...
try {
  wh.appendAsync(data);
} finally {
  data.release();
}

eolivelli: I think this is an unexpected behaviour (Maybe I need a coffee and I am missing something)

```static final ByteBuf DATA = Unpooled.wrappedBuffer("foobar".getBytes());
created DATA with refCnt == 1
1st call to
result(wh.appendAsync(DATA));
will work ok, take ownership of 1 refCnt of DATA and will reduce DATA's refCnt to 0 when no longer needs it.
```


This is not good, I (application) have a reference to DATA, I am using it, refcount = 1.
I pass it to BK, and I am continuing to hold a reference to it.
If BK uses it, it must call 'retain' any time BK wants (refcount > 1), but the net effect MUST be that when BK finished its own work the refcount is back to 1.
When my application does not need that ByteBuf anymore, it will call release() and refcount will be 0 and then Netty will free or recycle the buffer.

Is is not this the expected behaviour, that is 'net effect of BK = 0" ?

@reddycharan already brought up this discussion about the 'net effect'

dlg99: That's different from what app had before when it used i.e. byte arrays for data.
But this is expected with netty ref-counted ByteBuffs.
I'll refer you to http://netty.io/wiki/reference-counted-objects.html#who-destroys-it "Who destroys it?" chapter again.

eolivelli: @dlg99 I understand your point.

Following that guide the client which is performing addEntry somehow says 'now this bytebuf is one of yours'

eolivelli:
rerun integration tests

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
merlimat: @dlg99 The main reason for doing `retain()` from within `asyncAddEntry()` was to have a "friendly API". 

In general, the contract with Netty ByteBufs is that if some API is going to take temporarely ownership of a buffer, it should retain (and later release() accordingly). 

In this case, the retain is to make sure the buffer is valid until the end of the write operation, irrespective of what the BK app is doing with the buffer. 

For example, without this retain, an application should make sure the buffer is valid till the end, thus doing something like: 

```java
data.retain();
ledger.asyncAddEntry(data, (rc, entryId) -> {
    // do something 
    data.release();
}, ctx);
```

This would just force the application to do the same. 

The problem with the ByteBufList that is currently computed is that this happens from the ordered exectutor and not from the same thread that is calling the `asyncAddEntry()`, thus the buffer might be already invalid if application wouldn't call retain() itself.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
```diff
@@ -456,7 +458,10 @@ private void maybeRecycle() {
     private void recyclePendAddOpObject() {
         entryId = LedgerHandle.INVALID_ENTRY_ID;
         currentLedgerLength = -1;
-        payload = null;
+        if (payload != null) {
+            ReferenceCountUtil.release(payload);
```
merlimat: `ReferenceCountUtil.safeRelease(payload);` would automatically do the if/null check

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/DeferredSyncTest.java
merlimat: @dlg99 

> "the party who accesses a reference-counted object lastly is responsible for the destruction of the reference-counted object"

I think the statement is a bit misleading. The destruction happens when the ref-count reaches 0. Each party will not know exactly when that is the case, they will just call `release()` and move on. 

I think a better way to describe the pattern is in term of "temporary ownership", or "ensuring the buffer is valid until I'm done". 

This way: 
 * If I'm passed a buffer, I assume it's a valid one
 * If I use the buffer immediately (in same thread), there's no need to retain/release
 * If I want to store a ref to the buffer and use it later (eg: pass to other thread, or keep it around in case I need to retransmit), then I need to retain/release. As a rule of thumb, I would say that this should be hidden in the API, because the caller shouldn't be forced to know what I'm going to do internally with that given buffer.

stream/distributedlog/core/src/main/java/org/apache/distributedlog/EnvelopedEntryWriter.java
```diff
@@ -165,7 +165,7 @@ public synchronized ByteBuf getBuffer() throws InvalidEnvelopedEntryException, I
         if (null == finalizedBuffer) {
             finalizedBuffer = finalizeBuffer();
         }
-        return finalizedBuffer.slice();
+        return finalizedBuffer.retainedSlice();
```
merlimat: This is changing the net counts for this buffer. I don't really know much on this code, but would this still be correct when `envelopeBeforeTransmit` is false?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
dlg99: @merlimat there are a few discussions around this in the CR.

my counterarguments:

1. netty's http://netty.io/wiki/reference-counted-objects.html#who-destroys-it 
"The general rule of thumb is that the party who accesses a reference-counted object lastly is responsible for the destruction of the reference-counted object."

2. add's callback does not pass data as a parameter, does not need to access it, so no need to release within it as long as there is no retain().

This would do:
```java
ledger.asyncAddEntry(data, (rc, entryId) -> {
    // do something 
}, ctx);
```

now, if app needs data after addition:
```java
data.retain();
try {
  ledger.asyncAddEntry(data, (rc, entryId) -> {
      // do something 
  }, ctx);
  ...
  log.info("data was {}", asHex(data));
} finally {
   data.release()
}
```

alternative, with data.retain happening within asyncAddEntry one would have to do something like:

```java
try {
  ledger.asyncAddEntry(data, (rc, entryId) -> {
      // do something 
  }, ctx);
} finally {
  data.release();
}
```
or do that in the callback, as in your example.

Optionally we can follow what netty is doing with some of ByteBuf's methods:
have asyncAddEntry(ByteBuf...)
and asyncAddEntryAndRetain(ByteBuf...)
with well documented differences.

Note that all these data retention things only apply to methods that use netty's ByteBuf directly, nothing changes when byte[] on java's ByteBuffer variant used.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/DeferredSyncTest.java
dlg99: @merlimat 

> If I'm passed a buffer, I assume it's a valid one

ok

> If I use the buffer immediately (in same thread), there's no need to retain/release

There is no need to retain, but a need to release() when you don't need it anymore. 

> If I want to store a ref to the buffer and use it later (eg: pass to other thread, or keep it around in case I need to retransmit), then I need to retain/release. 

then you release it later. when you are done with it.
when you pass it to another thread that will release it independently, do call retain() or pass retainedDuplicate() to that thread/runnable.

> As a rule of thumb, I would say that this should be hidden in the API, because the caller shouldn't be forced to know what I'm going to do internally with that given buffer

It won't be hidden as now one who passed that buffer has to wrap all calls to addAsync into try/finally and release that buffer, unlike with the rest of netty's API that uses ByteBufs.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
dlg99: @merlimat please look at the unit tests that I added
specifically assert for refcount https://github.com/apache/bookkeeper/pull/1585/files#diff-d168699d534345c67678296447f8375cR1344
has to pass (refcount after add == 0).
Otherwise it means that:
- whoever created ByteBuf has to deal with its release, despite netty's guidelines.
- if not released we risk memleak.

stream/distributedlog/core/src/main/java/org/apache/distributedlog/EnvelopedEntryWriter.java
dlg99: @merlimat I am not familiar with dlog's code. 
This mimicked what the tests expected/extra retain from addAsync so I figured it cannot be worse that it was initially.

I hope @sijie can help with the question on expectations from envelopeBeforeTransmit

sijie:
regarding the dlog question, the dlog logic should be irrelative to this change. the change essentially removes `retain` from asyncAddEntry, and assume there is 1 reference count from the passed-in bytebuf that bookkeeper client needs to release. from the dlog code change perspective, it moves `retain` outside of `asyncAddEntry` and combines with `slice`. so the change is correct.

more in details:

> I don't really know much on this code, but would this still be correct when envelopeBeforeTransmit is false?

no matter `envelopeBeforeTransmit` is true or false, `finalizeBuffer` will keep one reference count for the buffer during the lifecycle of `EnvelopedEntryWriter`, when `EnvelopedEntryWriter` ends its lifecycle (via completeTransmit or abortTransmit), it will release the reference count of the buffer. so when caller calls `getBuffer`, it retained the slice. so the caller will be responsible for releasing the buffer. in the context of this PR, the caller is calling `bookkeeper#asyncAddEntry(entryWriter.getBuffer())`, so the retained buffer is passed in bookkeeper and bookkeeper will release it after it finishes using it.

sijie:
@merlimat : I think @dlg99 's change is kind of following the behavior that Netty is using and recommends. Netty releases `ByteBuf` sent to the channels after using them. It is kind of making the behavior consistent across the board, from API to network wire. what do you think?

sijie:
@eolivelli : 

> I think for an application it is better that BK is the last owner of the ByteBuf, otherwise it should handle the release in the callback and it makes things very tricky.

I don't think either solution requires application to handle release in the callback, unless I misunderstood your comment.

I think contract is whether bk should retain a reference count for its usage during the lifecycle of an entry inside bk (current approach - Matteo's temporary ownership theory) or bk doesn't have to retain a reference count for its usage during lifecycle (the proposed approach - sender-defer-decision-to-receiver).

looking into the contract between bk client and netty, it is following sender-defer-decision-to-receiver. bk (sender) passes a reference-counted bytebuf to netty (receiver). netty (receiver) releases the reference count. if we want to make the behavior consistent, the caller (sender) should pass a reference-counted bytebuf to bk (receiver), bk (receiver) would release the reference count.

but yes, agreed that either approach would work, we just need to achieve an agreement about the contract of bytebuf usage.

dlg99:
@merlimat @sijie any reasons why this should not be merged?
do we still have any disagreements on how ByteBuf ownership should work?

sijie:
@dlg99 let's wait one night for @merlimat . if there is no objections from him, we can merge tomorrow morning.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
jvrao: Hmm, netty does have weird recommendation practice. :)

dlg99:
@merlimat @sijie ping.

eolivelli:
This is ready to go please commit and cherry pick to 4.8

sijie:
@eolivelli @dlg99 merged to master and cherry-picked to 4.8.

@merlimat yes we will include this API change into release notes

dlg99:
Thank you!

PR #1586, opened 2018-08-07
Get currentEnsemble from ledger rather than directly from metadata

ivankelly:
This will allow us to override the currentEnsemble being used during
recovery without modifying the metadata itself.

Master issue: #281


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -1215,7 +1215,7 @@ private boolean isWritesetWritable(DistributionSchedule.WriteSet writeSet,
 
         int nonWritableCount = 0;
         for (int i = 0; i < sz; i++) {
-            if (!bk.getBookieClient().isWritable(getLedgerMetadata().currentEnsemble.get(i), key)) {
+            if (!bk.getBookieClient().isWritable(getCurrentEnsemble().get(i), key)) {
```
eolivelli: Sample currentEnsemble before starting the loop?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
```diff
@@ -138,7 +138,7 @@ long getEntryId() {
     void sendWriteRequest(int bookieIndex) {
         int flags = isRecoveryAdd ? FLAG_RECOVERY_ADD | FLAG_HIGH_PRIORITY : FLAG_NONE;
 
-        lh.bk.getBookieClient().addEntry(lh.getLedgerMetadata().currentEnsemble.get(bookieIndex),
+        lh.bk.getBookieClient().addEntry(lh.getCurrentEnsemble().get(bookieIndex),
```
eolivelli: Sample the list outside of this method ? in the loop I mean

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
```diff
@@ -266,7 +266,7 @@ public void writeComplete(int rc, long ledgerId, long entryId, BookieSocketAddre
         int bookieIndex = (Integer) ctx;
         --pendingWriteRequests;
 
-        if (!lh.getLedgerMetadata().currentEnsemble.get(bookieIndex).equals(addr)) {
+        if (!lh.getCurrentEnsemble().get(bookieIndex).equals(addr)) {
```
eolivelli: Same here

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadLacOp.java
```diff
@@ -62,12 +66,12 @@
...
         this.coverageSet = lh.distributionSchedule.getCoverageSet();
+        this.currentEnsemble = lh.getCurrentEnsemble();
     }
 
     public void initiate() {
-        LedgerMetadata metadata = lh.getLedgerMetadata();
-        for (int i = 0; i < metadata.currentEnsemble.size(); i++) {
-            lh.bk.getBookieClient().readLac(metadata.currentEnsemble.get(i),
+        for (int i = 0; i < currentEnsemble.size(); i++) {
+            lh.bk.getBookieClient().readLac(currentEnsemble.get(i),
```
eolivelli: This is perfect to me

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
sijie: I am also in favor of saving the ensemble to a variable before starting the loop. However I think the logic here needs to be revisited, just to make sure ensemble changes won't impact the logic here.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -2382,4 +2383,10 @@ public void closeComplete(int rc, LedgerHandle lh, Object ctx) {
         }
     }
 
+    List<BookieSocketAddress> getCurrentEnsemble() {
+        // Getting current ensemble from the metadata is only a temporary
+        // thing until metadata is immutable. At that point, current ensemble
+        // becomes a property of the LedgerHandle itself.
```
sijie: I know `LedgerHandle` is the mix of write and read handle. However this method is exposed to ReadOnlyHandle. It is a bit confused when we are saying `current ensemble` for a readonly handle. so I would suggest adding more comments here to clarify the meanings.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
ivankelly: changed. it won't affect the logic. this code is an advisary thing to apply backpressure. The ensemble could still change between a writeset being seen as writable, and the pendingAddOp being triggered, but this won't cause a problem.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
ivankelly: done

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
ivankelly: There'll be a big comment on the ReadOnlyLedgerHandle implementation. The whole class hierarchy here is broken. LedgerHandle should derive from ReadOnlyLedgerHandle, not vice versa. ReadOnlyLedgerHandle shouldn't do recovery. Recovery should be a separate operation that occurs before the ledger handle is even created. However, currently this is a big change because of the lack of separation between the *Op classes and LedgerHandle. This getCurrentEnsemble() is the ugliness I was referring to on dev@ when I said the solution isn't nice. The add op has to call back to the ledger handle to get the new ensemble after a failure. It would be better if, when an ensemble change occurs, all outstanding pending add ops are passed the new ensemble **by** the ledger handle. I might actually change it do that. The ops shouldn't depend on LedgerHandle. The objects they require should be passed in as parameters. There's so much janky stuff in this code, it can be painful to not fix it all at once.

sijie: @ivankelly we all understand the problem and motivation behind this. however that's a different topic to address all these. so let's not try to couple that here. 

Here what I am asking is adding a comment to make thing clear. That's it. I want people who might be working on this file or when people look back at the git commit, we know exactly what was the thought of `getCurrentEnsemble`.

eolivelli:
Checkstyle is not happy

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadLacOp.java
jvrao: @eolivelli can you elaborate why you made 'perfect' comment just for this? Just curious. :)

PR #1587, opened 2018-08-07
UpdateLedgerOp uses MetadataUpdateLoop

ivankelly:
Also refactored a lot of things around the operation that didn't make
sense, like the executor.

Master issue: #281


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/UpdateLedgerOp.java
```diff
@@ -81,174 +87,80 @@ public UpdateLedgerOp(final BookKeeper bkc, final BookKeeperAdmin admin) {
...
-        }
-
-        private void finishUpdateLedger() {
-            outstandings.remove(ledgerId);
-            if (outstandings.isEmpty() && stop.get()) {
-                LOG.info("Total number of ledgers issued={} updated={}", issuedLedgerCnt.get(), updatedLedgerCnt.get());
-                syncObj.countDown();
-            }
+        try {
+            CompletableFuture.allOf(outstanding.stream().toArray(CompletableFuture[]::new)).get();
```
eolivelli: Maybe we can add an utility method to FutureUtils in order to do this operation

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/UpdateLedgerOp.java
```diff
@@ -81,174 +87,80 @@ public UpdateLedgerOp(final BookKeeper bkc, final BookKeeperAdmin admin) {
...
-            this.toBookieAddr = toBookieAddr;
+    private static LedgerMetadata replaceBookieInEnsembles(LedgerMetadata metadata,
+                                                           BookieSocketAddress oldBookieId,
+                                                           BookieSocketAddress newBookieId) {
+        LedgerMetadataBuilder builder = LedgerMetadataBuilder.from(metadata);
+        for (Map.Entry<Long, ? extends List<BookieSocketAddress>> e : metadata.getEnsembles().entrySet()) {
+            List<BookieSocketAddress> newEnsemble = e.getValue().stream()
+                .map(b -> b.equals(oldBookieId) ? newBookieId : b)
+                .collect(Collectors.toList());
+            builder.withEnsembleEntry(e.getKey(), newEnsemble);
```
sijie: I have a comment around `withEnsembleEntry` in last PR but since I didn't know how it was used so I didn't left the comment.

but here it is:

- since it is used for updating an existing ensemble, i prefer renaming `withEnsembleEntry` to `replaceEnsembleEntry` to make things explicit and clear
- if it is a replace operation, let's add a validation before put to make sure the ensemble exists. https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadataBuilder.java#L94

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/UpdateLedgerOp.java
```diff
@@ -81,174 +87,80 @@ public UpdateLedgerOp(final BookKeeper bkc, final BookKeeperAdmin admin) {
...
+                                    .flatMap(Collection::stream)
+                                    .filter(b -> b.equals(oldBookieId))
+                                    .count() > 0;
+                            },
+                            (metadata) -> {
+                                return replaceBookieInEnsembles(metadata, oldBookieId, newBookieId);
+                            },
+                            ref::compareAndSet).run();
+                }).whenComplete((metadata, ex) -> {
+                        LOG.info("Update of {} finished", ledgerId);
```
sijie: `update` is ambiguous. can you make the statement more specific?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/UpdateLedgerOp.java
```diff
@@ -81,174 +87,80 @@ public UpdateLedgerOp(final BookKeeper bkc, final BookKeeperAdmin admin) {
...
-            if (outstandings.isEmpty() && stop.get()) {
-                LOG.info("Total number of ledgers issued={} updated={}", issuedLedgerCnt.get(), updatedLedgerCnt.get());
-                syncObj.countDown();
-            }
+        try {
+            CompletableFuture.allOf(outstanding.stream().toArray(CompletableFuture[]::new)).get();
+            LOG.info("Total number of ledgers issued={} updated={}",
+                     issuedLedgerCnt.get(), updatedLedgerCnt.get());
+        } catch (ExecutionException e) {
+            LOG.info("Error in execution", e);
```
sijie: can we improve the logging here?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/UpdateLedgerOp.java
```diff
@@ -81,174 +87,80 @@ public UpdateLedgerOp(final BookKeeper bkc, final BookKeeperAdmin admin) {
...
+            throw new IOException("Error executing update", e);
```
sijie: can we improve the error message?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/UpdateLedgerOp.java
```diff
@@ -81,174 +87,80 @@ public UpdateLedgerOp(final BookKeeper bkc, final BookKeeperAdmin admin) {
...
-                UpdateLedgerNotifier progressable) {
-            this.ledgerId = ledgerId;
-            this.stop = stop;
-            this.issuedLedgerCnt = issuedLedgerCnt;
-            this.updatedLedgerCnt = updatedLedgerCnt;
-            this.outstandings = outstandings;
-            this.syncObj = syncObj;
-            this.progressable = progressable;
+            outstanding.add(promise);
+            promise.whenComplete((metadata, ex) -> outstanding.remove(promise));
```
sijie: this is not 100% correct. 

1) you are attaching a complete closure to `promise`, not to the promise after `thenCompose` above.

2) remove is mutating the structure. I am not sure how it would impact `outstanding.stream()` below.

so at this point I am not 100% confident about the refactor here. what I would suggest:

1) move `outstanding.remove` to the `whenComplete` above.  we don't need a separate whenComplete. that means you also move `outstanding.add` after constructing `promise`.

2) I would change line 144 to use a CountDownLatch rather than `CompletableFuture.allOf`.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/UpdateLedgerOp.java
ivankelly: Will change to newEnsembleEntry and replaceEnsembleEntry with validation for each.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/UpdateLedgerOp.java
ivankelly: done.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/UpdateLedgerOp.java
ivankelly: done

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/UpdateLedgerOp.java
ivankelly: done

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/UpdateLedgerOp.java
ivankelly: 1. changed, though slightly different to what you suggested, as we need the write promise, it's the promise returned from thenCompose that is added to outstanding.

2. outstanding is a concurrenthashmap, wrapped in a set. the stream will be generated from an iterator on the ConcurrentHashMap, which is a view of the contents at or after the point of the iterator creation. These are safe from concurrentmodificationexception. At the point stream() is called, nothing new will be added to the set, though entries may be removed. This is fine, because remove will only remove a completed future, which is what we're waiting for anyhow. I'm not even sure how I could use CountDownLatch here, and we would end up having some messy rubbish to get any thrown error out.

ivankelly:
rerun integration tests

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/UpdateLedgerOp.java
```diff
@@ -81,174 +74,100 @@ public UpdateLedgerOp(final BookKeeper bkc, final BookKeeperAdmin admin) {
...
-
-                        FutureCallback<Void> updateLedgerCb = new UpdateLedgerCb(lId, stop, issuedLedgerCnt,
-                                updatedLedgerCnt, outstandings, syncObj, progressable);
-                        Futures.addCallback(readCb.getFutureListener(), updateLedgerCb);
-
-                        issuedLedgerCnt.incrementAndGet();
-                        if (limit != Integer.MIN_VALUE && issuedLedgerCnt.get() >= limit || !ledgerItr.hasNext()) {
-                            stop.set(true);
+        final CompletableFuture<Void> finalPromise = new CompletableFuture<>();
+        final Set<CompletableFuture<?>> outstanding =
```
sijie: nit: This can be shorten as `ConcurrentHashMap.newKeySet()`

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/UpdateLedgerOp.java
```diff
@@ -81,174 +74,100 @@ public UpdateLedgerOp(final BookKeeper bkc, final BookKeeperAdmin admin) {
...
+                    }
+                });
 
-        private void finishUpdateLedger() {
-            outstandings.remove(ledgerId);
-            if (outstandings.isEmpty() && stop.get()) {
-                LOG.info("Total number of ledgers issued={} updated={}", issuedLedgerCnt.get(), updatedLedgerCnt.get());
-                syncObj.countDown();
+        try {
+            finalPromise.get();
```
sijie: use `FutureUtils.result(finalPromise)`, which already wrap the logic for handling ExecutionException

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/UpdateLedgerOp.java
```diff
@@ -81,174 +74,100 @@ public UpdateLedgerOp(final BookKeeper bkc, final BookKeeperAdmin admin) {
...
+        while (ledgerItr.hasNext() && !finalPromise.isDone()
+               && (limit == Integer.MIN_VALUE || issuedLedgerCnt.get() < limit)) {
+            // throttler to control updates per second
+            throttler.acquire();
+
+            final long ledgerId = ledgerItr.next();
+            issuedLedgerCnt.incrementAndGet();
+
+            GenericCallbackFuture<LedgerMetadata> readPromise = new GenericCallbackFuture<>();
+            lm.readLedgerMetadata(ledgerId, readPromise);
```
sijie: I think it will be much clearer (and easier to maintain in future) if you call lm.readLedgerMetadata at the end of the sequence (after adding the promise to outstanding list).

```
GenericCallbackFuture<LedgerMetadata> readPromise = ...;
CompletableFuture<LedgerMetadata> writePromise = readPromise.thenCompose(...);
// register a completion callback when updating ledger metadata completes
writePromise.whenComplete(...);
// add outstanding write to outstanding list
outstanding.add(writePromise);
// trigger the read
lm.readLedgerMetadata(ledgerId, readPromise);
```

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/UpdateLedgerOp.java
```diff
@@ -81,174 +74,100 @@ public UpdateLedgerOp(final BookKeeper bkc, final BookKeeperAdmin admin) {
...
+            outstanding.add(writePromise);
+            writePromise.whenComplete((metadata, ex) -> {
+                        if (ex != null
+                            && !(ex instanceof BKException.BKNoSuchLedgerExistsException)) {
+                            String error = String.format("Failed to update ledger metadata %s, replacing %s with %s",
+                                                         ledgerId, oldBookieId, newBookieId);
+                            LOG.error(error, ex);
+                            finalPromise.completeExceptionally(new IOException(error, ex));
+                        } else {
+                            LOG.info("Updated ledger {} metadata, replacing {} with {}",
```
sijie: This line seems to be new. I think it changes the existing logging behavior. I would prefer converting it to DEBUG, since it is a refactor, better to keep behaviors unchanged

sijie:
@jvrao : can anyone from your side review this as well?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/UpdateLedgerOp.java
ivankelly: I disagree, and this is of the things that annoys me about callbacks. 

When reading code, it makes most sense that the action occur in the order they occur on the page. By putting readLedgerMetadata at the end, you are breaking that order. Consider how this would look if instead of taking a promise, readLedgerMetadata would return a future (something I've been thinking about doing).

sijie: sure

PR #1590, opened 2018-08-09
Typo in bookie request processor read handling

ivankelly:
The response in the case of overload should take the ledgerId and
entryId from the read request.


eolivelli:
How could it work before this change ?
Should we cherry pick to 4.7 ?

ivankelly:
@eolivelli it's a corner case in an error branch. I suspect the client doesn't even look at the ledgerId and entryId on return, only the txid.

eolivelli:
rerun integration tests

sijie:
Can we a small unit test for this case? I am a bit worried about this kind of bug :(

PR #1594, opened 2018-08-10
Make bookie watcher an interface

ivankelly:
So that it can be mocked out easily for testing.
This patch also contains a simple mock.


bookkeeper-server/src/test/java/org/apache/bookkeeper/client/MockBookieWatcher.java
```diff
@@ -0,0 +1,100 @@
...
+import org.apache.bookkeeper.client.BKException.BKNotEnoughBookiesException;
+import org.apache.bookkeeper.net.BookieSocketAddress;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Mock implementation of BookieWatcher.
+ */
+public class MockBookieWatcher implements BookieWatcher {
```
sijie: Why not just mock metadata driver (including RegistrationClient)? In that way, all BookieWatchImpl logic can actually be exercised during testing.

ivankelly: LedgerHandle doesn't use registration client, doesn't get one passed in. It does get a BookieWatcher though, which is why i went for that. It's a bit messy, because placement policy is also passed in to the LedgerHandle. The relationship between these classes needs to be cleaned up a bit. For now I was just going for what was easiest.

sijie: > LedgerHandle doesn't use registration client, doesn't get one passed in. 

sorry I don't quite get this point.

> It does get a BookieWatcher though, which is why i went for that. 

I think what you want is to provide a mock bookie watcher when you are constructing ledger handle. my comment is instead of having a mock bookie watcher, you can have a mock registration client, then you can construct a *real* watcher with mock reg client and a real policy. 

```
RegistratonClient mockRegClient = mockRegistrationClient();
EnsemblePlacementPolicy policy = new DefaultEnsemblePlacementPolicy();
BookieWatcher mockWatcher = new BookieWatcher(mockRegClient, ...);
```

I understand the motivation of providing mocks. however I don't think we should provide mock for everything, because mock means maintaining two different implementations, when you have two different implementations, they usually can not keep up with each other, then problems will start increasing when they diverge. in most of the time, we should just mock the components that interact with network (like metadata driver, bookie client). unless you need a mock implementation to control failure points, that can be an exception.

since we are doing refactor to improve testability, we should just think clearly about what we will do. having a mock metadata driver which can be used for providing a *mock* bookie watcher seems to me is a more sustainable approach.

sijie:
run pr validation
run integration tests
run build java8
run build java9

ivankelly:
Ooops, need to merge master

PR #1599, opened 2018-08-10
Fix master versions (4.8.0-SNAPSHOT->4.9.0-SNAPSHOT)

ivankelly:
Some pom versions and the versions in bk_test_bin_common.sh were
missed when the version was bumped to 4.9.0-SNAPSHOT, which is killing
all builds.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -699,15 +699,10 @@
               <artifactId>checkstyle</artifactId>
               <version>${puppycrawl.checkstyle.version}</version>
             </dependency>
-            <dependency>
-              <groupId>org.apache.bookkeeper</groupId>
-              <artifactId>buildtools</artifactId>
-              <version>${project.version}</version>
-            </dependency>
           </dependencies>
           <configuration>
-            <configLocation>bookkeeper/checkstyle.xml</configLocation>
-            <suppressionsLocation>bookkeeper/suppressions.xml</suppressionsLocation>
+            <configLocation>buildtools/src/main/resources/bookkeeper/checkstyle.xml</configLocation>
+            <suppressionsLocation>buildtools/src/main/resources/bookkeeper/suppressions.xml</suppressionsLocation>
             <encoding>UTF-8</encoding>
             <consoleOutput>true</consoleOutput>
             <failOnViolation>true</failOnViolation>
--- a/stream/distributedlog/core/pom.xml
+++ b/stream/distributedlog/core/pom.xml
@@ -146,7 +146,7 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-checkstyle-plugin</artifactId>
         <configuration>
-          <suppressionsLocation>distributedlog/suppressions.xml</suppressionsLocation>
+          <suppressionsLocation>buildtools/src/main/resources/distributedlog/suppressions.xml</suppressionsLocation>
           <excludes>**/thrift/**/*</excludes>
         </configuration>
       </plugin>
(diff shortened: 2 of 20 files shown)


eolivelli:
Oh thank you.
We have to change the release guide

pom.xml
```diff
@@ -699,15 +699,10 @@
...
-            <dependency>
-              <groupId>org.apache.bookkeeper</groupId>
-              <artifactId>buildtools</artifactId>
-              <version>${project.version}</version>
-            </dependency>
           </dependencies>
           <configuration>
-            <configLocation>bookkeeper/checkstyle.xml</configLocation>
-            <suppressionsLocation>bookkeeper/suppressions.xml</suppressionsLocation>
+            <configLocation>buildtools/src/main/resources/bookkeeper/checkstyle.xml</configLocation>
```
sijie: can you explain why this change is needed? if it is just a version problem, why does this work in 4.8.0, not working in 4.9.0?

tests/scripts/src/test/bash/bk_test_bin_common.sh
```diff
@@ -32,7 +32,8 @@ testDefaultVariables() {
   assertEquals "BINDIR is not set correctly" "${BK_BINDIR}" "${BINDIR}"
   assertEquals "BK_HOME is not set correctly" "${BK_HOMEDIR}" "${BK_HOME}"
   assertEquals "DEFAULT_LOG_CONF is not set correctly" "${BK_CONFDIR}/log4j.properties" "${DEFAULT_LOG_CONF}"
-  assertEquals "NETTY_LEAK_DETECTION_LEVEL is not set correctly" "disabled" "${NETTY_LEAK_DETECTION_LEVEL}"
+  assertEquals "NETTY_LEAK_DETECTION_LEVEL is not set correctly" "disabled" "${NETTY_LEAK
```
sijie: I am not sure why the changes in this file are needed. `4.8.0` here is just version number used in the testing code. the testing code isn't coupled with the build.

sijie:
> We have to change the release guide

ah, I missed one command last time. the creating-branch command should be provided with `-Dstream` option.

pom.xml
ivankelly: The checkstyle plugin has a dependency on org.apache.bookeeper:buildtools:${version}. If you bump the version and run checkstyle without running mvn install first, like we do in CI, CI will fail because it can't find the dependency.

tests/scripts/src/test/bash/bk_test_bin_common.sh
ivankelly: I've no idea why this file is needed, I just changed every place I saw 4.8.0

ivankelly:
@eolivelli @sijie Maybe we can make the stream profile always activate for the release plugin.

eolivelli:
I think we should drop that profile and consider 'stream' as part of the full build.
Most if the time I am working only of bookkeeper-server

ivankelly:
@eolivelli I'm the same, I only build the whole thing to make sure compilation is ok.

tests/scripts/src/test/bash/bk_test_bin_common.sh
sijie: do you mind reverting this change back? because the change will become misleading. in future if similar problem happens, people might check the gitlog and follow what the change was doing here.

sijie:
> Maybe we can make the stream profile always activate for the release plugin.

that works for me. if it is easy, we can do it. otherwise, we can file an issue to do it in 4.9.

tests/scripts/src/test/bash/bk_test_bin_common.sh
ivankelly: sure. what is this file for?

sijie: these are the tests for testing our bash scripts

stream/distributedlog/core/pom.xml
```diff
@@ -146,7 +146,7 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-checkstyle-plugin</artifactId>
         <configuration>
-          <suppressionsLocation>distributedlog/suppressions.xml</suppressionsLocation>
+          <suppressionsLocation>buildtools/src/main/resources/distributedlog/suppressions.xml</suppressionsLocation>
```
eolivelli: Is this path relative?
How can this work?
Does it work if you run
mvn validate
From a non root directory?

ivankelly: It works. No idea why, but it does work.

PR #1600, opened 2018-08-14
Provide consistent locking mechanism in EntryLogManagerForEntryLogPerLedger

reddycharan:

Assumption: The lock stored alongside the EntryLog in this map is meant
to be used to ensure that no two threads can be writing to the same
ledger at the same time.

The above invariant can be violated if the EntryLogAndLockTuple
object is evicted from the cache while in a critical section nominally
protected by the contained lock.

The conditions required for this to happen would be pretty odd --
there needs to be a huge amount of cache churn during one of
the protected operations.

The fix for this issue is to allocate in the constructor a fixed array of locks and select
for each EntryLogAndLockTuple a lock from that array
deterministically by ledgerId such that the same ledgerId
will always get the same lock.


reddycharan:
@athanatos fyi

PR #1601, opened 2018-08-14
Downgraded ZK version to 3.4.13

merlimat:
### Motivation

BK 4.7 has shipped with newer ZooKeeper client 3.5.3-beta. In many cases, there has been concerns regarding having a dependency on a "beta" release. Irrespective of the seriousness of these concerns for the specific case of ZK which has been in alpha/beta for a very long time, we should not have dependency on versions of software for which the team itself has marked as "not yet stable". 

Adding to that, there is no clear roadmap or ETA for when final 3.5.x stable release will be available.

### Changes

 * Downgraded ZK to latest stable version 3.4.13
 * Adjusted few usages of new client APIs that were introduced in 3.5.x

### DIFF
--- a/bookkeeper-benchmark/src/main/java/org/apache/bookkeeper/benchmark/BenchReadThroughputLatency.java
+++ b/bookkeeper-benchmark/src/main/java/org/apache/bookkeeper/benchmark/BenchReadThroughputLatency.java
@@ -35,6 +35,7 @@ import org.apache.bookkeeper.client.BookKeeper;
 import org.apache.bookkeeper.client.LedgerEntry;
 import org.apache.bookkeeper.client.LedgerHandle;
 import org.apache.bookkeeper.conf.ClientConfiguration;
+import org.apache.bookkeeper.zookeeper.ZooKeeperClient;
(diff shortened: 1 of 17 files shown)


sijie:
> I think we switched to 3.5 because DL needed the 3.5 client API.

dlog only needs the async version of multi. I think that has been ported back to 3.4. 

> Are you aware of serious ZK client side issues with 3.5 client?

I think the security concerns are from "-alpha" or "-beta".

I believe most of the bug fixes have been ported back from 3.5.x to 3.4.x. and none of the features in 3.5.x are really needed by bookkeeper.

merlimat:
run build java8
run build java9

eolivelli:
retest this please

sijie:
mark this for `4.7.2`

merlimat:
run bookie tests

sijie:
run integration tests

merlimat:
run integration tests // Failed to fetch jars from maven

merlimat:
Backported to `branche-4.8` (f84a87a8) and `branch-4.7` (603bdbfa)

Issue #1606, opened 2018-08-16
ArrayIndexOutOfBoundsException on ConcurrentLongHashMap


sijie:
*Problem*

```
00:34:54.922 [bookie-io-1-6] ERROR org.apache.bookkeeper.proto.WriteEntryProcessor - Unexpected exception while writing 30031@3 : 4611
java.lang.ArrayIndexOutOfBoundsException: 4611
        at org.apache.bookkeeper.util.collections.ConcurrentLongHashMap$Section.get(ConcurrentLongHashMap.java:245) ~[org.apache.bookkeeper-bookkeeper-server-4.7.1.jar:4.7.1]
        at org.apache.bookkeeper.util.collections.ConcurrentLongHashMap.get(ConcurrentLongHashMap.java:125) ~[org.apache.bookkeeper-bookkeeper-server-4.7.1.jar:4.7.1]
        at org.apache.bookkeeper.bookie.HandleFactoryImpl.getHandle(HandleFactoryImpl.java:45) ~[org.apache.bookkeeper-bookkeeper-server-4.7.1.jar:4.7.1]
        at org.apache.bookkeeper.bookie.Bookie.getLedgerForEntry(Bookie.java:1104) ~[org.apache.bookkeeper-bookkeeper-server-4.7.1.jar:4.7.1]
        at org.apache.bookkeeper.bookie.Bookie.addEntry(Bookie.java:1214) ~[org.apache.bookkeeper-bookkeeper-server-4.7.1.jar:4.7.1]
        at org.apache.bookkeeper.proto.WriteEntryProcessor.processPacket(WriteEntryProcessor.java:79) [org.apache.bookkeeper-bookkeeper-server-4.7.1.jar:4.7.1]
        at org.apache.bookkeeper.proto.PacketProcessorBase.safeRun(PacketProcessorBase.java:85) [org.apache.bookkeeper-bookkeeper-server-4.7.1.jar:4.7.1]
        at org.apache.bookkeeper.common.util.SafeRunnable.run(SafeRunnable.java:36) [org.apache.bookkeeper-bookkeeper-common-4.7.1.jar:4.7.1]
        at org.apache.bookkeeper.proto.BookieRequestProcessor.processAddRequest(BookieRequestProcessor.java:487) [org.apache.bookkeeper-bookkeeper-server-4.7.1.jar:4.7.1]
[...]

PR #1607, opened 2018-08-16
Issue 1606: Fixed race condition during expansion of concurrent open hash maps

merlimat:
### Motivation

As reported in #1606, there is a race condition in the concurrent open hash maps implementation. The race happens when the maps gets re-hashed after the expansion and the new arrays are substituting the old ones.

The race itself is that a thread doing a `get()` on the map is first checking the current `capacity` of the map, uses that to get the bucket and then tries to do optimistic read of the value in that bucket. 

This assumes `capacity` update is visible only after the `values` array is already swapped, but that is not always the case in current code.

### Changes
 * Use `volatile` qualifier for `capacity` and `values` arrays to ensure ordering of memory read is respected by compiler
 * In rehashing, update `capacity` after `values`


sijie:
run pr validation

merlimat:
Merged in master and ported to `branch-4.8` (07bb42ad) and `branch-4.7` (36edacab)

Issue #603, opened 2017-10-17
128 bit LedgerId


ivankelly:
JIRA: https://issues.apache.org/jira/browse/BOOKKEEPER-864

Reporter: Venkateswararao Jujjuri (JV) @jvrao

BookKeeper coordinates with ZooKeeper to generate an cluster wide LedgerId. This is a 64 bit number. This method works great because we have ZK acting as a centralized coordinator.  But this method may not scale as the cluster size and number of ledgers increase. 

GUIDs ( (https://en.wikipedia.org/wiki/Globally_unique_identifier) are preferred way to generate decentralized globally unique IDs and it takes 128 bits ; This method can scale well as it doesn’t need a centralized coordination. 
Current BK code used ledgerId as a primary object, any change to this (64 to128 bit) warrants across the board code changes. 

Opening this issue so we can solicit community input and track the progress.

### Comments from JIRA

---
*Sijie Guo* 2015-09-01T14:34:32.092+0000

I've been looking around the code base for a while. The current code base is kind of heavily tight with Long (64bits) ledger id.

- the bookie entry format is formed with ledger_id, entry_id & data.
- the bookie storage is using ledger id as the key on referencing entries. especially -1 (INVALID_LEDGER_ID) is used in the server for identifying as a special entry
- the network wire protocol is also tight with ledger id

It isn't a straightforward change to move ledger id from 64 bits to 128 bits. It involves changes from storage, network wire protocol to client interface. so if we are going to support 128 bit ledger id, I would suggest supporting byte[] as the 'ledger id'. the challenging part of this change is how to make backward compatibility. 

my thinking over this is:

- moving the ledger storage to support byte[] as the ledger id first. providing interfaces like addEntry(byte[] ledger_name, long entry_id, ByteBuffer data) and readEntry(byte[] ledger_name, long entry_id, ByteBuffer data).
- the original wire protocol (using long as the ledger id), could covert long ledger id to byte[] and adding/read entries
- adding a new wire protocol (typically a new AddRequest, ReadRequest), which use byte[] as the ledger id
- changing the ledger metadata manager to use byte[] as the ledger id
- for client side, we convert long to bytes[] to support backward clients who use long as the ledger id. we should propose a brand new client which uses byte[] as the ledger id (so which we could propose a clean API to get rid of some drawbacks of client API).

the most critical part of this change is to take care of backward compatibility. 

---
*Venkateswararao Jujjuri (JV)* 2015-09-01T15:20:05.019+0000

[...]

PR #1611, opened 2018-08-17
BP-35: 128 bits support

sijie:

### Motivation

BookKeeper coordinates with a metadata store to generate a cluster wide `ledgerId`.
Currently this is a signed `64 bit` number (effectively 63 bits). This method works
great because we have a centralized metadata store for coordinating the id generation.
However this method may not scale as the cluster size and number of ledgers grow.

[Universally unique identifier - Wikipedia](https://en.wikipedia.org/wiki/Globally_unique_identifier)
is a preferred way to generate decentralized globally unique IDs and it takes `128 bits`.
This method can scale well as it doesn't need a centralized coordination.

This BP proposes the changes for increasing ledger id from `63 bits` to `128 bits`.

Master Issue: #603

> ---
>
> - [x] [skip bookkeeper-server bookie tests]: skip testing `org.apache.bookkeeper.bookie` in bookkeeper-server module.
> - [x] [skip bookkeeper-server client tests]: skip testing `org.apache.bookkeeper.client` in bookkeeper-server module.
> - [x] [skip bookkeeper-server replication tests]: skip testing `org.apache.bookkeeper.replication` in bookkeeper-server module.
> - [x] [skip bookkeeper-server tls tests]: skip testing `org.apache.bookkeeper.tls` in bookkeeper-server module.
> - [x] [skip bookkeeper-server remaining tests]: skip testing all other tests in bookkeeper-server module.
> - [x] [skip integration tests]: skip docker based integration tests. if you make java code changes, you shouldn't skip integration tests.
> - [x] [skip build java8]: skip build on java8. *ONLY* skip this when *ONLY* changing files under documentation under `site`.
> - [x] [skip build java9]: skip build on java9. *ONLY* skip this when *ONLY* changing files under documentation under `site`.
> ---

> ---
> quickly and easily:
>
>
>
> Otherwise:
> 
> 
> ---

### DIFF
--- a/site/bps/BP-35-128-bits-support.md
+++ b/site/bps/BP-35-128-bits-support.md
@@ -1,0 +1,399 @@
+---
+title: "BP-34: 128 bits support"
+issue: https://github.com/apache/bookkeeper/603
+state: "Under Discussion"
+design doc: https://docs.google.com/document/d/1cu54dNSV2ZrdWCi40LcyX8NxXGRCW0609T_ewmK9BWM
+release: "4.9.0"
+---
+
+### Motivation
+
+BookKeeper coordinates with a metadata store to generate a cluster wide `ledgerId`.
+Currently this is a signed `64 bit` number (effectively 63 bits). This method works
+great because we have a centralized metadata store for coordinating the id generation.
+However this method may not scale as the cluster size and number of ledgers grow.
+
+[Universally unique identifier - Wikipedia](https://en.wikipedia.org/wiki/Globally_unique_identifier)
+is a preferred way to generate decentralized globally unique IDs and it takes `128 bits`.
+This method can scale well as it doesn't need a centralized coordination. 
+
+This BP proposes the changes for increasing ledger id from `63 bits` to `128 bits`.
+
+### 128 bits
+
+Since there is no native support for `128 bits` in both Java and
+[Protobuf](https://github.com/google/protobuf/issues/2180), we have to break `128 bits`
+into 2 `64 bits` numbers for representing the `128 bits` id:
+
+- ledger-id-msb: the most significant 64 bits, bit 64 - 127
+- ledger-id-lsb: the least significant 64 bits, bit 0 - 63
+
+For backward compatibility, the `ledger-id-lsb` is the current `64 bits` ledger-id.
+The `ledger-id-msb` will be added as a new field in both API and protocol. 
+
+I am proposing calling `ledger-id-msb` as `ledger-scope-id`. So the current 64bits `ledgerId` and
+the newly introduced 64bits `ledgerScopeId` together will be forming the new `128 bits` ledger id.
+
+The default `ledgerScopeId` is `0`. That means any ledgers created prior to this change are allocated
+under scope `0`. Hence it maintains backward compatibility during upgrade. 
+
+The combination of `ledgerScopeId` and `ledgerId` forms the `128 bits` ledger id. We can introduce a
+hex representation of this `128 bits` ledger id - `ledgerQualifiedName` . This `ledgerQualifiedName` can
+be useful for CLI tooling, REST api and troubleshooting purpose. The API internally can convert
+`ledgerQualifiedName` to `ledgerScopeId` and `ledgerId`.
+
+### Public Interfaces
+
+#### API Change
+
+The API will be introducing `ledgerScopeId` across the interfaces. This field will be optional and default to `0`. 
+
+##### Handle
+
+Introduce a new method `getScopeId` for representing the scope id (the most significant  `128 bits` ledger id).
+
+```java
+public interface Handle extends AutoCloseable {
+
+  ...
+
+  /**
+   * Return the ledger scope id. The most significant 64 bits of 128 bits.
+   */
+  long getScopeId();
+
+  /**
+   * Return the ledger id. The least significant 64 bits of 128 bits.
+   */ 
+  long getId();
+
+  ...
+
+}
+```
+
+##### Create LedgerAdv
+
+Introduce a new method `withLedgerScopeId` in `CreateAdvBuilder` for providing `scopeId`
+(the most significant 64 bits for 128 bits ledger id) on creating a ledger.
+
+```java
+public interface CreateAdvBuilder extends OpBuilder<WriteHandle> {
+  ...
+
+  /**
+   * Set the scope id for the newly created ledger.
+   * If no explicit scopeId is passed, the new ledger
+   * will be created under scope `0`.
+   */
+  CreateAdvBuilder withLedgerScopeId(long scopeId);	
+
+  ...
+}
+```
+
+##### Open Ledger
+
+Introduce a new method `withLedgerScopeId` in `OpenBuilder` for providing `scopeId`
+(the most significant 64 bits for 128 bits ledger id) on opening a ledger.
+
+```java
+public interface OpenBuilder extends OpBuilder<ReadHandle> {
+  ...
+  /**
+   * Set the scope id of the ledger to open.
+   */
+  OpenBuilder withLedgerScopeId(long scopeId);
+  ...
+}
+```
+
+##### Delete Ledger
+
+Introduce a new method `withLedgerScopeId` in `DeleteBuilder` for providing `scopeId`
+(the most significant 64 bits for 128 bits ledger id) on deleting a ledger.
+
+```java
+public interface DeleteBuilder extends OpBuilder<Void> {
+  ...
+  /**
+   * Set the scope id of the ledger to delete.
+   */
+  DeleteBuilder withLedgerScopeId(long scopeId);
+  ...
+}
+```
+
+#### CLI
+
+All BookKeeper CLI tools will be updated with additional option `—ledger-scope-id`.
+Optionally we can add option `—ledger-qualified-name` (the hex representation of 128 bits).
+Internally all the CLI tools will convert ledger qualified name to `ledgerId` and `ledgerScopeId`.
+
+#### REST
+
+1. All ledger related endpoints will be adding a new parameter `ledger_scope_id`. 
+2. `ListLedgerService`  only supports listing ledgers under a given ledger scope id.
+   If `ledger_scope_id` is missing, it will be listing ledgers under scope `0`.
+
+#### Wire Protocol
+
+> There will be no plan for supporting 128 bits in v2 protocol, due to the limitation in v2 protocol.
+> So any operations in v2 protocol with scope id not equal to 0  will be failed immediately with
+> `ILLEGAL_OP` exceptions.
+
+All the request and response messages will be adding an optional field `optional int64 ledgerScopeId`.
+
+#### Entry Format
+
+Currently all the entries written to bookies are encoded in a certain format, including `metadata`,
+`digest code` and `payload`. The entry format is not *versioned*.
+
+In order to support adding another field `ledgerScopeId` in the `metadata` section, we are introducing
+`version` in the entry format.
+
+##### Entry Format V1
+
+```json
+Entry Format V1
+===============
+--- header ---
+Bytes (0 - 7)                   : Ledger ID
+Bytes (8 - 15)                  : Entry ID
+Bytes (16 - 23)                 : LastAddConfirmed
+Bytes (24 - 31)                 : Length
+--- digest ---
+Bytes (32 - (32 + x - 1))       : Digest Code (e.g. CRC32)
+--- payload ---
+Bytes ((32 + x) - )             : Payload
+```
+
+> `x` is the length of digest code.
+
+>  Prior to introducing `ledgerScopeId`, ledgerId is assumed to be a positive value.
+
+##### Entry Format V2
+
+```json
+Entry Format V2
+===============
+--- header ---
+Bytes (0 - 7)                   : Metadata Flags
+Bytes (8 - 15)                  : Ledger Scope ID
+Bytes (16 - 23)                 : Ledger ID
+Bytes (24 - 31)                 : Entry ID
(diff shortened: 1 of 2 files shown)


site/bps/BP-35-128-bits-support.md
```diff
@@ -0,0 +1,399 @@
...
+###### EntryMemTable
+
+`ledgerScopeId` should be added as part of `EntryKey`.
+
+###### IndexPersistenceMgr
+
+Currently the ledger index files (64 bits) are stored into 2-level-hirechicy
+directories - `<msb-32bits-hex>/<lsb-32bits-hex>/<ledger-id-hex>.idx`.
+
+If `ledgerScopeId` is 0, it will be using existing scheme for storing and retrieving
```
ivankelly: Why bifurcate the code paths like this? ledgerScopeId 0 should be treated as special in as few places as possible.

site/bps/BP-35-128-bits-support.md
```diff
@@ -0,0 +1,399 @@
...
+
+All the file info caches should be updated to use `<ledgerScopeId, ledgerId>`
+as index keys.
+
+###### IndexInMemPageMgr
+
+The LRU pages map will be updated to use `<ledgerScopeId, ledgerId>` as index
+keys.
+
+###### DBLedgerStorage
```
ivankelly: Unless DBLedgerStorage is part of your intended usecase, I would exempt it from this change for now, especially since its primary usecase uses the V2 protocol which is also exempted.

site/bps/BP-35-128-bits-support.md
```diff
@@ -0,0 +1,399 @@
...
+- **Memtable**: additional 8 bytes will be added per indexed entry.
+- **FileInfo**: there is no change to the index file format itself.
+- **IndexPersistenceManager**: Files will be organized in more directory hierarchy.
+  It shouldn't be a big deal. 
+- **IndexInMemoryManager (LedgerCache)**: additional 8 bytes per index page.
+- **DbLedgerStorage**: additional 8 bytes per entry for entry location.
+- **Metadata**: on zookeeper, we need a 128 bit ledger manager, that means more znode
+  hierarchy than 64 bit ledger manager. Etcd like key/value metadata store is probably
+  more preferrable for 128 bit ledger manager.
+
```
ivankelly: Garbage collection?

site/bps/BP-35-128-bits-support.md
```diff
@@ -0,0 +1,399 @@
...
+However increasing ledger id from 64 bits to 128 bits can get rid of the only remaining
+central point, since we don't need to use zookeeper for ledger id generation. The id
```
ivankelly: "only remaining central point"

There's still a good few centrallized scalability squeeze points, namely Garbage collection (which pulls all ledger ids to each node) and Auditor (which also operates on all ids).

site/bps/BP-35-128-bits-support.md
```diff
@@ -0,0 +1,399 @@
...
+---
+
+### Motivation
+
+BookKeeper coordinates with a metadata store to generate a cluster wide `ledgerId`.
+Currently this is a signed `64 bit` number (effectively 63 bits). This method works
+great because we have a centralized metadata store for coordinating the id generation.
+However this method may not scale as the cluster size and number of ledgers grow.
+
+[Universally unique identifier - Wikipedia](https://en.wikipedia.org/wiki/Globally_unique_identifier)
```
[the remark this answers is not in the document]
jvrao: That's my understanding too. So in that way, should we just keep calling it lid_usb? or something? naming it ```scope``` might make folks see things differently (mostly in the wrong way)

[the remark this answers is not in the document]
sijie: `UUID` is one way for generating 128 bit. However it might be other ways to generate 128 bit ledger id. The angle I am coming from is adding another `64 bit` to existing `64 bit`, so that's why I call it `scopeId`. In that way, if people use UUID for ledger id generation, it can make msb to scopeId and lsb to existing ledger id; and it will also give the flexibility for people to decide how they want to generate the id and how they want to map it into scopeId and ledgerId.

call it "lid_msb" is fine to me.

site/bps/BP-35-128-bits-support.md
```diff
@@ -0,0 +1,399 @@
...
+- ledger-id-msb: the most significant 64 bits, bit 64 - 127
+- ledger-id-lsb: the least significant 64 bits, bit 0 - 63
+
+For backward compatibility, the `ledger-id-lsb` is the current `64 bits` ledger-id.
+The `ledger-id-msb` will be added as a new field in both API and protocol. 
+
+I am proposing calling `ledger-id-msb` as `ledger-scope-id`. So the current 64bits `ledgerId` and
+the newly introduced 64bits `ledgerScopeId` together will be forming the new `128 bits` ledger id.
+
+The default `ledgerScopeId` is `0`. That means any ledgers created prior to this change are allocated
```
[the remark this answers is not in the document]
sijie: > If you do go down the of generating scope from the UUID, the scope of 0 should be reserved (i.e. if UUID.randomUUID gives back a uuid with msb 0, discard and try again).

I am not sure we need to reserve scope 0 or Long.MAX_VALUE. because currently bookkeeper allows application managing id generation. 

for certain (or default) metadata driver implementation, we can reserve those scope ids if needed.

site/bps/BP-35-128-bits-support.md
```diff
@@ -0,0 +1,399 @@
...
+
+> There will be no plan for supporting 128 bits in v2 protocol, due to the limitation in v2 protocol.
+> So any operations in v2 protocol with scope id not equal to 0  will be failed immediately with
+> `ILLEGAL_OP` exceptions.
+
+All the request and response messages will be adding an optional field `optional int64 ledgerScopeId`.
+
+#### Entry Format
+
+Currently all the entries written to bookies are encoded in a certain format, including `metadata`,
```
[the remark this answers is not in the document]
sijie: it is not just wire protocol or journal format. it is the format of an entry:

- client is aware of this format (for serializing/deserializing an entry)
- it is transparent to wire protocol. so wire protocol doesn't care.
- journal needs to be aware of this format (for learning the header part, ledger id, entry id, lac and such).
- ledger storage needs to be aware of this format (same as journal).

that's why I kept this in its own section called "Entry Format".

site/bps/BP-35-128-bits-support.md
```diff
@@ -0,0 +1,399 @@
...
+--- payload ---
+Bytes ((37 + x) - )             : Payload
+``` 
+
+> `x` is the length of digest code.
+
+###### Metadata Flags
+
+```json
+Metadata: 1 Bytes (Long)
```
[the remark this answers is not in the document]
sijie: Nice catch. It was 8 bytes. However that means we are adding 16 bytes per entry (8 bytes flags + 8 bytes scopeId), so we reduced it to 1 byte. I need to update the above description.

> A byte is too small. At least give yourselves an Int

right now we just need version and digest type, which I think 1 byte is enough initially. once we have this version flag in place, it is easier to extend and change in future. since this is affecting every entry added to bookie, so I was thinking starting with 1 byte not one int or long.

However if you guys think it requires at least 4 bytes, I am fine with extending it.

site/bps/BP-35-128-bits-support.md
```diff
@@ -0,0 +1,399 @@
...
+ version
+
+----
+Bit 0 - 3: digest type (e.g. CRC32, CRC32C and such)
+Bit 4 - 7: version, the most significant bit of this byte will be always set to 1.
+it will be used for differentiating entry format v1 and v2.
+
+```
+
+We are setting the most significant bit to be `1`. So the first byte in entry v2 will
```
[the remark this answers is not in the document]
sijie: the format is different, because v1 doesn't have scope, v2 have scope.

site/bps/BP-35-128-bits-support.md
```diff
@@ -0,0 +1,399 @@
...
+```
+
+#### Bookie Storage
+
+##### Journal
+
+A new method should be added in journal `WriteCallback` to handle `ledgerScopeId`.
+
+```java
+public interface WriteCallback {
```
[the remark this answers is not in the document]
sijie: >  all the client side internal callbacks will also have to change to include scope.

all callback requires to change to take extra scope field.

> Change the callback type for writing the Journal completely. 

I would defer this to implementation decision, since it doesn't affect overall design.

site/bps/BP-35-128-bits-support.md
```diff
@@ -0,0 +1,399 @@
...
+```
+
+The journal should be changed to be able to retrieve `ledgerScopeId` from the entry
+payload based on [Entry Format](#entry-format).
+
+##### Ledger Storage
+
+###### EntryLogger
+
+1. Methods in `EntryLogger` should be able to accept `ledgerScopeId` as a parameter.
```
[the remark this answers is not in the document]
sijie: both write/read calls or related methods in `EntryLogger`. I didn't list the details and just use `methods`.

site/bps/BP-35-128-bits-support.md
sijie: This is for BC consideration. You already have a ledger manager running with ledgerScopeId == 0. you have to keep the logic functioning as it is needed.

site/bps/BP-35-128-bits-support.md
sijie: I include it for completeness. because DBLedgerStorage is not tied to V2 protocol, it can be used by V3 protocol users.

site/bps/BP-35-128-bits-support.md
sijie: GC should remain unchanged, right? All the performance impact should be on metadata side or bookie storage side.

site/bps/BP-35-128-bits-support.md
sijie: > Garbage collection

but it is no "central point", right? if you have a distributed metadata store, you should be able to scale out metadata operations.

ivankelly: It is though. When GC kicks off, we get a list of the ledgers locally stored, and a list of the ledgers in the metadata. This list of ledgers in the metadata has to be pulled to each bookie, which is inheritly unscalable. Anyhow, my point is, we need to work on GC and auditor to remove all the squeeze points.

site/bps/BP-35-128-bits-support.md
ivankelly: I'd prefer a migration script for this rather than keeping two independent code paths running at the same time. We need to keep the old path for people not using scopeId anyhow. the migration itself is easy (just a mv). If a bookie is going to be aware of the scopeId they need to upgrade their journal anyhow, so this could be part of the same upgrade.

site/bps/BP-35-128-bits-support.md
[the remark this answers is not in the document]
ivankelly: I'm fine with it being scopeID. I don't think we should tie ourselves to any one ID generation mechanism. I can think of many ways a scope ID mechanism could be used in a decentrallized manner without using UUID. Each client could even be given a scopeId on boot, given that there's 64bits to play with.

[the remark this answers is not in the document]
sijie: Great

site/bps/BP-35-128-bits-support.md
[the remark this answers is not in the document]
ivankelly: sure.

site/bps/BP-35-128-bits-support.md
sijie: > the migration itself is easy (just a mv).

a migration script can be problematic for a production system. it can become much complicated than keeping BC. 

I think we can start a brand new metadata driver as I commented in mailing list for Enrico's question, for existing metadata driver keeping the BC and maybe providing a migration script for people to cutover if needed.

site/bps/BP-35-128-bits-support.md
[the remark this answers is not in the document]
ivankelly: uff, this is going to touch a lot of places. When implementing, please make sure the v1 vs v2 stuff is well encapsulated, so that we don't end up with a load of ```if (v1) { doSomething } else if (v2) { doAnotherThing }```

site/bps/BP-35-128-bits-support.md
sijie: sure

site/bps/BP-35-128-bits-support.md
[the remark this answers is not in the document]
ivankelly: given you're above answer, it seems that this differentiation needs to be everywhere. The code shouldn't be aware of this differentiation, since there's a simple sane default of 0 for scope , so differentiation should only be in the code encapsulating the entry.

site/bps/BP-35-128-bits-support.md
ivankelly: From a performance PoV yes, but it will still need to take the scopeId into account.

site/bps/BP-35-128-bits-support.md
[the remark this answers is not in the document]
ivankelly: If you leave it as 8 bits, reserve one bit to allow us to extend later.

site/bps/BP-35-128-bits-support.md
sijie: I am not sure I understand your point here. 

Upgrading Journal is much easier as it doesn't require any coordination. However upgrading metadata is not, it includes coordination between clients (reader and writer) and bookies. A script on migrating metadata doesn't really work out in a production system.

sijie: oh sorry I misunderstood your comment. then a migration/upgrade script works

PR #1612, opened 2018-08-18
Allow to configure num of Netty IO threads in client and bookie

merlimat:
### Motivation

Currently the number of IO threads for client and bookie are set to `2 *  Runtime.getRuntime().availableProcessors()`. Added configuration options to tune it.


merlimat:
>  Do you have numbers or did some trials? 

I haven't tested the various numbers. Only scope for this PR is to allow flexibility to configure thread pool size. I might make sense to reduce the thread pool when there are other event loop thread pools running in same process.

sijie:
run integration tests

sijie:
run integration tests

PR #1616, opened 2018-08-20
Index for lockarray should be non-negative

reddycharan:

Make sure lockIndex is non-negative.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogManagerForEntryLogPerLedger.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogManagerForEntryLogPerLedger.java
@@ -94,7 +94,7 @@ class EntryLogManagerForEntryLogPerLedger extends EntryLogManagerBase {
         private BufferedLogChannelWithDirInfo entryLogWithDirInfo;
 
         private EntryLogAndLockTuple(long ledgerId) {
-            int lockIndex = Long.hashCode(ledgerId) % lockArrayPool.length();
+            int lockIndex = Math.abs(Long.hashCode(ledgerId) % lockArrayPool.length());
             if (lockArrayPool.get(lockIndex) == null) {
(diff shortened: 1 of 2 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogManagerForEntryLogPerLedger.java
```diff
@@ -93,7 +93,7 @@ BufferedLogChannel getLogChannel() {
         private BufferedLogChannelWithDirInfo entryLogWithDirInfo;
 
         private EntryLogAndLockTuple(long ledgerId) {
-            int lockIndex = Long.hashCode(ledgerId) % lockArrayPool.length();
+            int lockIndex = Math.abs(Long.hashCode(ledgerId) % lockArrayPool.length());
```
merlimat: If `hashCode()` returns a negative, the modulo operation will fail anyway. I think we should use `MathUtils.signSafeMod()` here.

reddycharan: https://docs.oracle.com/javase/specs/jls/se7/html/jls-15.html#jls-15.17.3

in java for modulo (remainder) operation both right and left operands can be negative.

reddycharan:
@sijie @eolivelli fyi this is critical fix

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogManagerForEntryLogPerLedger.java
merlimat: Sure, though I'd prefer to consistently use `signSafeMod()` like in rest of codebase

reddycharan: changed it to signSafeMod

Issue #1617, opened 2018-08-21
Remove sync calls from callbacks in AutoRecovery 


sijie:
Context: http://mail-archives.apache.org/mod_mbox/bookkeeper-dev/201411.mbox/%3CCAO2yDyZo5AzYgE%3D%3Dk8C5bifA3Miv-2A9w%2B_-6aS%2Bwa4zxRGcOw%40mail.gmail.com%3E

Related Issues: #1578

sijie:
This is done by #1619

PR #1619, opened 2018-08-22
Provide async version of markLedgerUnderreplicated for LedgerUnderreplicationManager

sijie:

 ### Motivation

Auditor has multiple places calling sync methods in async callbacks.
This raises the possibility hitting deadlock. Issue #1578 is one of the examples.

After looking into the `LedgerUnderreplicationManager`, `markLedgerUnderreplicated`
is the only interface that will be called in async callbacks. This change is
to provide an async version of `markLedgerUnderreplicated`.

 ### Changes

- add `markLedgerUnderreplicatedAsync` interface in `LedgerUnderreplicationManager`.
- implement the logic of `markLedgerUnderreplicated` using async callbacks
- use `markLedgerUnderreplicatedAsync` in the Auditor

Related Issues: #1578
Master Issue: #1617

> ---
>
> ---

> ---
> quickly and easily:
>
>
>
> Otherwise:
> 
> 
> ---

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BKException.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BKException.java
@@ -37,13 +37,7 @@ public abstract class BKException extends org.apache.bookkeeper.client.api.BKExc
         if (cause instanceof BKException) {
             return (BKException) cause;
         } else {
-            BKException ex;
-            if (cause instanceof InterruptedException) {
-                Thread.currentThread().interrupt();
-                ex = new BKInterruptedException();
-            } else {
-                ex = new BKUnexpectedConditionException();
-            }
+            BKException ex = new BKUnexpectedConditionException();
             ex.initCause(cause);
             return ex;
         }
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BKException.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BKException.java
@@ -20,6 +20,8 @@
  */
 package org.apache.bookkeeper.client;
 
+import java.util.function.Function;
+
 /**
  * Class the enumerates all the possible error conditions.
  *
@@ -27,6 +29,25 @@ package org.apache.bookkeeper.client;
  */
 @SuppressWarnings("serial")
 public abstract class BKException extends org.apache.bookkeeper.client.api.BKException {
+
+    public static final Function<Throwable, BKException> HANDLER = cause -> {
+        if (cause == null) {
+            return null;
+        }
+        if (cause instanceof BKException) {
+            return (BKException) cause;
+        } else {
+            BKException ex;
+            if (cause instanceof InterruptedException) {
+                Thread.currentThread().interrupt();
+                ex = new BKInterruptedException();
+            } else {
+                ex = new BKUnexpectedConditionException();
+            }
+            ex.initCause(cause);
+            return ex;
+        }
+    };
 
     BKException(int code) {
         super(code);
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/LedgerUnderreplicationManager.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/LedgerUnderreplicationManager.java
@@ -17,10 +17,14 @@
  */
 package org.apache.bookkeeper.meta;
 
+import com.google.common.collect.Lists;
+import java.util.Collection;
 import java.util.Iterator;
 import java.util.List;
+import java.util.concurrent.CompletableFuture;
 import java.util.function.Predicate;
 
+import org.apache.bookkeeper.common.concurrent.FutureUtils;
 import org.apache.bookkeeper.proto.BookkeeperInternalCallbacks.GenericCallback;
 import org.apache.bookkeeper.replication.ReplicationException;
 
@@ -28,12 +32,26 @@ import org.apache.bookkeeper.replication.ReplicationException;
  * Interface for marking ledgers which need to be rereplicated.
  */
 public interface LedgerUnderreplicationManager extends AutoCloseable {
+
     /**
      * Mark a ledger as underreplicated. The replication should
      * then check which fragments are underreplicated and rereplicate them
      */
-    void markLedgerUnderreplicated(long ledgerId, String missingReplica)
-            throws ReplicationException.UnavailableException;
+    default void markLedgerUnderreplicated(long ledgerId, String missingReplica) throws ReplicationException {
+        FutureUtils.result(
+            markLedgerUnderreplicatedAsync(
+                ledgerId, Lists.newArrayList(missingReplica)), ReplicationException.EXCEPTION_HANDLER);
+    }
+
+    /**
+     * Mark a ledger as underreplicated with missing bookies. The replication should then
+     * check which fragements are underreplicated and rereplicate them.
+     *
+     * @param ledgerId ledger id
+     * @param missingReplicas missing replicas
+     * @return a future presents the mark result.
+     */
+    CompletableFuture<Void> markLedgerUnderreplicatedAsync(long ledgerId, Collection<String> missingReplicas);
 
     /**
      * Mark a ledger as fully replicated. If the ledger is not
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/ZkLedgerUnderreplicationManager.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/ZkLedgerUnderreplicationManager.java
@@ -27,18 +27,21 @@ import com.google.protobuf.TextFormat.ParseException;
 
 import java.net.UnknownHostException;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Queue;
+import java.util.concurrent.CompletableFuture;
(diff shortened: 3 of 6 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
```diff
@@ -622,91 +611,54 @@ void checkAllLedgers() throws BKAuditException, BKException,
...
-                throw BKException.create(returnCode.get());
-            }
+                (rc, path, ctx) -> {
+                    if (Code.OK == rc) {
+                        FutureUtils.complete(processFuture, null);
+                    } else {
+                        FutureUtils.completeExceptionally(processFuture, BKException.create(rc));
+                    }
+                }, null, BKException.Code.OK, BKException.Code.ReadException);
+            FutureUtils.result(processFuture, BKException.HANDLER);
```
merlimat: If processing is done async, will get impacted by the below `admin.close()` and `client.close()`?

sijie: processFuture is completed only when finished processing the last ledger, so checkAllLedgers is still sync, which is fine, since checkAllLedgers is executing in Auditor's own executor.

sijie:
regarding tests: this PR only focuses on rewriting `markLedgerUnderreplicated` in async way. it doesn't change existing logic or adding new functionalities, so no tests are added.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/LedgerUnderreplicationManager.java
```diff
@@ -17,23 +17,41 @@
...
+
+    /**
+     * Mark a ledger as underreplicated with missing bookies. The replication should then
+     * check which fragements are underreplicated and rereplicate them.
+     *
+     * @param ledgerId ledger id
+     * @param missingReplicas missing replicas
+     * @return a future presents the mark result.
+     */
+    CompletableFuture<Void> markLedgerUnderreplicatedAsync(long ledgerId, Collection<String> missingReplicas);
```
reddycharan: nit: dont we need to have async version of "void markLedgerReplicated(long ledgerId)"

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BKException.java
```diff
@@ -28,6 +30,25 @@
...
 
+    public static final Function<Throwable, BKException> HANDLER = cause -> {
+        if (cause == null) {
+            return null;
+        }
+        if (cause instanceof BKException) {
+            return (BKException) cause;
+        } else {
+            BKException ex;
+            if (cause instanceof InterruptedException) {
```
reddycharan: FutureUtils.result would throw InterruptedException without applying the exceptionHandler function https://github.com/apache/bookkeeper/blob/master/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/concurrent/FutureUtils.java#L77

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/LedgerUnderreplicationManager.java
reddycharan: is there any need of making 'missingReplicas' collection, instead of just String missingReplica

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/ZkLedgerUnderreplicationManager.java
```diff
@@ -254,58 +258,95 @@ public UnderreplicatedLedgerFormat getLedgerUnreplicationInfo(long ledgerId)
...
-            List<ACL> zkAcls = ZkUtils.getACLs(conf);
-            String znode = getUrLedgerZnode(ledgerId);
-            while (true) {
-                UnderreplicatedLedgerFormat.Builder builder = UnderreplicatedLedgerFormat.newBuilder();
+            LOG.debug("markLedgerUnderreplicated(ledgerId={}, missingReplica={})", ledgerId, missingReplicas);
+        }
+        final List<ACL> zkAcls = ZkUtils.getACLs(conf);
+        final String znode = getUrLedgerZnode(ledgerId);
+        final CompletableFuture<Void> createFuture = new CompletableFuture<>();
+        tryMarkLedgerUnderreplicatedAsync(znode, missingReplicas, zkAcls, createFuture);
```
reddycharan: nit: why does caller has to create 'createFuture' (CompletableFuture) and pass it, why cann't method returns the CompletableFuture

sijie: this allows the method `tryMarkLedgerUnderreplicatedAsync` to be attempted with same `CompleteableFuture`.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/LedgerUnderreplicationManager.java
sijie: I changed this for https://github.com/apache/bookkeeper/pull/1619/files#diff-7525f06ad3a1ad0a00a462df4deb4698L581 . then it doesn't need to do multiple zk calls for creating an UL.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BKException.java
sijie: nice. I would remove this here

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/LedgerUnderreplicationManager.java
sijie: we don't. `markLedgerReplicated` is used by replication worker, which replication worker is single threaded and using sync methods. so `markLedgerReplicated` is fine at the context.
only `markLedgerUnderreplicated` is the issue.

ideally, LedgerUnderreplicationManager should be splitted into at least 2 interfaces, one for Auditor, the other one for ReplicationWorker. That would make things much clearer.

sijie:
@reddycharan I have addressed your comments.

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
```diff
@@ -571,45 +568,37 @@ private void publishSuspectedLedgers(String bookieIP, Set<Long> ledgers)
...
+                Set<BookieSocketAddress> bookies = Sets.newHashSet();
+                for (LedgerFragment f : fragments) {
+                    bookies.addAll(f.getAddresses());
                 }
+                publishSuspectedLedgersAsync(
+                    bookies.stream().map(BookieSocketAddress::toString).collect(Collectors.toList()),
+                    Sets.newHashSet(lh.getId())
+                ).whenComplete((result, cause) -> {
+                    if (null != cause) {
+                        callback.processResult(Code.ReplicationException, null, null);
```
reddycharan: just want to make sure that this 'cause' is not lost in log errors. will it not hide cause completely here?

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
```diff
@@ -571,45 +568,37 @@ private void publishSuspectedLedgers(String bookieIP, Set<Long> ledgers)
...
+                    } else {
+                        callback.processResult(Code.OK, null, null);
+                    }
+                });
+            } else {
+                callback.processResult(rc, null, null);
             }
-
-            callback.processResult(rc, null, null);
+            lh.closeAsync().whenComplete((result, cause) -> {
```
reddycharan: why is it responsibility of ProcessLostFragmentsCb to close the ledgerhandle, I'm not sure if it is appropriate to do it here, since ProcessLostFragmentsCb is not the owner of 'lh'

sijie: I didn’t change the behavior. The original logic is ProcessLostFramgmentsCb closing it.

reddycharan: checkLedgersProcessor 
 closes the 'lh' https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java#L679

sijie: check here: https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java#L584

reddycharan:
changes LGTM, but i understand that there is no functional change and they are not public API methods, but change of this magnitude needs some new test coverage for confidence.

sijie:
> change of this magnitude needs some new test coverage for confidence.

if there is no functionality change, what new tests do you expect?

sijie:

sijie:
This test is only rewriting sync calls to use async calls. Although `markLedgerUnderreplicatedAsync` is an async method, it is still used in sync ways across the places at Auditor. That says this PR doesn't change any existing logic how Auditor uses `markLedgerUnderreplicated`. from functionality wises, we don't need separate test cases.

Agreed that we need to have better coverage and since `makrLedgerUnderreplicatedAsync` is now at the interface, people might be using it in a different place. Those should be done in a separate PR rather than this PR. so I created a separate issue for tracking test coverage: https://github.com/apache/bookkeeper/issues/1626, not trying to combine changes with different purposes into one PR.

PR #1620, opened 2018-08-23
Avoid usage of RocksDB deleteRange() in DbLedgerStorage

merlimat:
### Motivation

There are few issues that are reconducible to a performance degradation in RocksDB when using deleteRange() feature (eg: https://github.com/apache/incubator-pulsar/issues/1737 and others).

There is some discussion going on RocksDB to address this issue: https://github.com/facebook/rocksdb/issues/3959

In the meantime, we should rollback the change and don't use deleteRange until these issues are resolved.

### Changes 

This PR is essentially reverting back the commit https://github.com/yahoo/bookkeeper/commit/4b849904bcd65b49cf963e6508dc7fb745f56294 from Yahoo branch (which was squashed when merging back to apache). 
The only addition here is to use `DELETE_ENTRIES_BATCH_SIZE` to amortize the cost of `batch.flush()` when there are many ledgers with few entries.


merlimat:
run integration tests

PR #1621, opened 2018-08-23
Recovery uses immutable metadata

ivankelly:
And MetadataUpdateLoop to update it.

Master issue: #281

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
@@ -2257,12 +2257,4 @@ public class LedgerHandle implements WriteHandle {
         // becomes a property of the LedgerHandle itself.
         return metadata.getCurrentEnsemble();
     }
-
-    static int getExceptionCode(Throwable t, int defaultCode) {
-        if (t instanceof BKException) {
-            return ((BKException) t).getCode();
-        } else {
-            return defaultCode;
-        }
-    }
 }
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
@@ -76,9 +76,7 @@ import org.apache.bookkeeper.common.concurrent.FutureUtils;
 import org.apache.bookkeeper.common.util.MathUtils;
 import org.apache.bookkeeper.net.BookieSocketAddress;
 import org.apache.bookkeeper.proto.BookieProtocol;
-import org.apache.bookkeeper.proto.BookkeeperInternalCallbacks;
 import org.apache.bookkeeper.proto.BookkeeperInternalCallbacks.GenericCallback;
-import org.apache.bookkeeper.proto.BookkeeperInternalCallbacks.TimedGenericCallback;
 import org.apache.bookkeeper.proto.DataFormats.LedgerMetadataFormat.State;
 import org.apache.bookkeeper.proto.checksum.DigestManager;
 import org.apache.bookkeeper.proto.checksum.MacDigestManager;
@@ -323,6 +321,10 @@ public class LedgerHandle implements WriteHandle {
             // ensure that we only update the metadata if it is the object we expect it to be
             if (metadata == expected) {
                 metadata = newMetadata;
+                if (metadata.isClosed()) {
+                    lastAddConfirmed = lastAddPushed = metadata.getLastEntryId();
+                    length = metadata.getLength();
+                }
                 return true;
             } else {
                 return false;
@@ -2222,112 +2224,6 @@ public class LedgerHandle implements WriteHandle {
         }
     }
 
-
-    void recover(GenericCallback<Void> finalCb) {
-        recover(finalCb, null, false);
-    }
-
-    /**
-     * Recover the ledger.
-     *
-     * @param finalCb
-     *          callback after recovery is done.
-     * @param listener
-     *          read entry listener on recovery reads.
-     * @param forceRecovery
-     *          force the recovery procedure even the ledger metadata shows the ledger is closed.
-     */
-    void recover(GenericCallback<Void> finalCb,
-                 final @VisibleForTesting BookkeeperInternalCallbacks.ReadEntryListener listener,
-                 final boolean forceRecovery) {
-        final GenericCallback<Void> cb = new TimedGenericCallback<Void>(
-            finalCb,
-            BKException.Code.OK,
-            clientCtx.getClientStats().getRecoverOpLogger());
-        boolean wasClosed = false;
-        boolean wasInRecovery = false;
-
-        LedgerMetadata metadata = getLedgerMetadata();
-        synchronized (this) {
-            if (metadata.isClosed()) {
-                if (forceRecovery) {
-                    wasClosed = false;
-                    // mark the ledger back to in recovery state, so it would proceed ledger recovery again.
-                    wasInRecovery = false;
-                    metadata.markLedgerInRecovery();
-                } else {
-                    lastAddConfirmed = lastAddPushed = metadata.getLastEntryId();
-                    length = metadata.getLength();
-                    wasClosed = true;
-                }
-            } else {
-                wasClosed = false;
-                if (metadata.isInRecovery()) {
-                    wasInRecovery = true;
-                } else {
-                    wasInRecovery = false;
-                    metadata.markLedgerInRecovery();
-                }
-            }
-        }
-
-        if (wasClosed) {
-            // We are already closed, nothing to do
-            cb.operationComplete(BKException.Code.OK, null);
-            return;
-        }
-
-        if (wasInRecovery) {
-            // if metadata is already in recover, dont try to write again,
-            // just do the recovery from the starting point
-            new LedgerRecoveryOp(LedgerHandle.this, clientCtx, cb)
-                    .setEntryListener(listener)
-                    .initiate();
-            return;
-        }
-
-        writeLedgerConfig(new OrderedGenericCallback<LedgerMetadata>(clientCtx.getMainWorkerPool(), ledgerId) {
-            @Override
-            public void safeOperationComplete(final int rc, LedgerMetadata writtenMetadata) {
-                if (rc == BKException.Code.MetadataVersionException) {
-                    rereadMetadata(new OrderedGenericCallback<LedgerMetadata>(clientCtx.getMainWorkerPool(),
-                                                                              ledgerId) {
-                        @Override
-                        public void safeOperationComplete(int rc, LedgerMetadata newMeta) {
-                            if (rc != BKException.Code.OK) {
-                                cb.operationComplete(rc, null);
-                            } else {
-                                LedgerHandle.this.metadata = newMeta;
-                                recover(cb, listener, forceRecovery);
-                            }
-                        }
-
-                        @Override
-                        public String toString() {
-                            return String.format("ReReadMetadataForRecover(%d)", ledgerId);
-                        }
-                    });
-                } else if (rc == BKException.Code.OK) {
-                    // we only could issue recovery operation after we successfully update the ledger state to
-                    // in recovery otherwise, it couldn't prevent us advancing last confirmed while the other writer is
-                    // closing the ledger, which will cause inconsistent last add confirmed on bookies & zookeeper
-                    // metadata.
-                    new LedgerRecoveryOp(LedgerHandle.this, clientCtx, cb)
-                        .setEntryListener(listener)
-                        .initiate();
-                } else {
-                    LOG.error("Error writing ledger {} config: {}", ledgerId, BKException.codeLogger(rc));
-                    cb.operationComplete(rc, null);
-                }
-            }
-
-            @Override
-            public String toString() {
-                return String.format("WriteLedgerConfigForRecover(%d)", ledgerId);
-            }
-        });
-    }
-
     static class NoopCloseCallback implements CloseCallback {
         static NoopCloseCallback instance = new NoopCloseCallback();
 
@@ -2358,5 +2254,13 @@ public class LedgerHandle implements WriteHandle {
         // thing until metadata is immutable. At that point, current ensemble
         // becomes a property of the LedgerHandle itself.
         return metadata.getCurrentEnsemble();
+    }
+
+    static int getExceptionCode(Throwable t, int defaultCode) {
+        if (t instanceof BKException) {
+            return ((BKException) t).getCode();
+        } else {
+            return defaultCode;
+        }
     }
 }
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerRecoveryOp.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerRecoveryOp.java
@@ -22,7 +22,6 @@ import com.google.common.annotations.VisibleForTesting;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.atomic.AtomicLong;
 import org.apache.bookkeeper.client.AsyncCallback.AddCallback;
-import org.apache.bookkeeper.common.concurrent.FutureUtils;
 import org.apache.bookkeeper.proto.BookkeeperInternalCallbacks.ReadEntryListener;
 import org.apache.bookkeeper.proto.checksum.DigestManager.RecoveryData;
 import org.slf4j.Logger;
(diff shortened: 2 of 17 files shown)


ivankelly:
I want to add more testing to this, but throwing it up so you all can see the shape of it beforehand.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadOnlyLedgerHandle.java
```diff
@@ -207,4 +206,190 @@ public void readLastConfirmedComplete(int rc, long lastConfirmed, Object ctx) {
...
+        for (int i = 0; i < e1.size(); i++) {
+            if (!e1.get(i).equals(e2.get(i))) {
+                diff.add(i);
+            }
+        }
+        return diff;
+    }
+
+    @Override
+    void handleBookieFailure(final Map<Integer, BookieSocketAddress> failedBookies) {
```
eolivelli: Can we add a comment explaining that we are in ReadOnlyLedgerHandle and this method will trigger only on recovery.
I am referring to @jvrao questions on slack

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadOnlyLedgerHandle.java
```diff
@@ -207,4 +206,190 @@ public void readLastConfirmedComplete(int rc, long lastConfirmed, Object ctx) {
...
+        }
+        LOG.info("Closing recovered ledger {} at entry {}", getId(), lac);
+        CompletableFuture<LedgerMetadata> f = new MetadataUpdateLoop(
+                clientCtx.getLedgerManager(), getId(),
+                this::getLedgerMetadata,
+                (metadata) -> metadata.isInRecovery(),
+                (metadata) -> {
+                    LedgerMetadataBuilder builder = LedgerMetadataBuilder.from(metadata);
+                    Optional<Long> lastEnsembleKey = metadata.getLastEnsembleKey();
+                    checkState(lastEnsembleKey.isPresent(),
```
eolivelli: How an error from this assertion will be handled from the MULoop ? 
Will the resulting CompletableFuture fail ?
Do we have test cases ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadOnlyLedgerHandle.java
```diff
@@ -207,4 +206,190 @@ public void readLastConfirmedComplete(int rc, long lastConfirmed, Object ctx) {
...
+                               "Metadata shouldn't have been created without at least one ensemble");
+                    newEnsemblesFromRecovery.entrySet().forEach(
+                            (e) -> {
+                                // Occurs when a bookie need to be replaced at very start of recovery
+                                if (lastEnsembleKey.get().equals(e.getKey())) {
```
eolivelli: Do we have a specific test case?

ivankelly: Ya, I want to test this more, hence the mock reg client (and some more).

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadOnlyLedgerHandle.java
ivankelly: checkState throws an IllegalStateException, which will be caught by the loop. The returned future will be completed with it.
MetadataUpdateLoopTest#testNewestValueCannotBeUsedAfterReadBack tests this functionality.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadOnlyLedgerHandle.java
ivankelly: Added

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadOnlyLedgerHandle.java
```diff
@@ -207,4 +206,190 @@ public void readLastConfirmedComplete(int rc, long lastConfirmed, Object ctx) {
...
+                                                       Map<Integer, BookieSocketAddress> failedBookies)
+            throws BKException.BKNotEnoughBookiesException {
+        List<BookieSocketAddress> newEnsemble = new ArrayList<BookieSocketAddress>(oldEnsemble);
+
+        int ensembleSize = metadata.getEnsembleSize();
+        int writeQ = metadata.getWriteQuorumSize();
+        int ackQ = metadata.getAckQuorumSize();
+        Map<String, byte[]> customMetadata = metadata.getCustomMetadata();
+
+        Set<BookieSocketAddress> exclude = new HashSet<BookieSocketAddress>(failedBookies.values());
```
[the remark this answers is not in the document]
ivankelly: changed

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -2359,4 +2255,12 @@ public void closeComplete(int rc, LedgerHandle lh, Object ctx) {
         // becomes a property of the LedgerHandle itself.
         return metadata.getCurrentEnsemble();
     }
+
+    static int getExceptionCode(Throwable t, int defaultCode) {
+        if (t instanceof BKException) {
```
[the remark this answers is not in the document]
ivankelly: I don't understand the question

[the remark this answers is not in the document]
eolivelli: Sorry.
This is a generic method, maybe the BKException is wrapped by another exception, like an ExecutionException.

Btw it is not so important

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadOnlyLedgerHandle.java
```diff
@@ -207,4 +206,197 @@ public void readLastConfirmedComplete(int rc, long lastConfirmed, Object ctx) {
...
+    /**
+     * For a read only ledger handle, this method will only ever be called during recovery,
+     * when we are reading forward from LAC and writing back those entries. As such,
+     * unlike with LedgerHandle, we do not want to persist changes to the metadata as they occur,
+     * but rather, we want to defer the persistence until recovery has completed, and do it all
+     * on the close.
+     */
+    @Override
+    void handleBookieFailure(final Map<Integer, BookieSocketAddress> failedBookies) {
+        blockAddCompletions.incrementAndGet();
```
athanatos: Do we need to do this on a ReadOnlyLedgerHandle?

ivankelly: Technically, I think not, as this should run in the same thread as the completions occur in. This is all implicit though, so we can't be 100% sure, and someone could change it in the future and break everything. Same thing for the synchronized. I have some other suggestions for cleaning up the thread model that I'll post to dev@ later this week, but for now I'd like to leave this here.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadOnlyLedgerHandle.java
```diff
@@ -207,4 +206,197 @@ public void readLastConfirmedComplete(int rc, long lastConfirmed, Object ctx) {
...
+
+        // handleBookieFailure should always run in the ordered executor thread for this
+        // ledger, so this synchronized should be unnecessary, but putting it here now
+        // just in case (can be removed when we validate threads)
+        synchronized (this) {
```
sijie: this seems to change to a broader lock scope. I would suggest to have a metadataLock for locking metadata during this refactor period, this would reduce the risk of unknowns. we can consider removing the metadata lock once you done with immutable ledger metadata change.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadOnlyLedgerHandle.java
```diff
@@ -207,4 +206,197 @@ public void readLastConfirmedComplete(int rc, long lastConfirmed, Object ctx) {
...
+        int ackQ = metadata.getAckQuorumSize();
+        Map<String, byte[]> customMetadata = metadata.getCustomMetadata();
+
+        Set<BookieSocketAddress> exclude = new HashSet<>(failedBookies.values());
+
+        int replaced = 0;
+        for (Map.Entry<Integer, BookieSocketAddress> entry : failedBookies.entrySet()) {
+            int idx = entry.getKey();
+            BookieSocketAddress addr = entry.getValue();
+            LOG.debug("[EnsembleChange-L{}] replacing bookie: {} index: {}", getId(), addr, idx);
```
sijie: `if (LOG.isDebugEnabled())`

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadOnlyLedgerHandle.java
```diff
@@ -207,4 +206,197 @@ public void readLastConfirmedComplete(int rc, long lastConfirmed, Object ctx) {
...
+
+            if (!newEnsemble.get(idx).equals(addr)) {
+                LOG.debug("[EnsembleChange-L{}] Not changing failed bookie {} at index {}, already changed to {}",
```
sijie: `if (LOG.isDebugEnabled())`

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadOnlyLedgerHandle.java
```diff
@@ -207,4 +206,197 @@ public void readLastConfirmedComplete(int rc, long lastConfirmed, Object ctx) {
...
+                                if (lastEnsembleKey.get().equals(e.getKey())) {
+                                    builder.replaceEnsembleEntry(e.getKey(), e.getValue());
+                                } else {
+                                    builder.newEnsembleEntry(e.getKey(), e.getValue());
+                                }
+                            });
+                    return builder.closingAt(lac, len).build();
+                },
+                this::setLedgerMetadata).run();
+        f.thenRun(() -> newEnsemblesFromRecovery.clear());
```
sijie: which thread does this run?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadOnlyLedgerHandle.java
```diff
@@ -207,4 +206,197 @@ public void readLastConfirmedComplete(int rc, long lastConfirmed, Object ctx) {
...
+        return f;
+    }
+
+    @Override
+    List<BookieSocketAddress> getCurrentEnsemble() {
+        if (newEnsemblesFromRecovery.size() > 0) {
```
sijie: no synchronization at `newEnsemblesFromRecovery`?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadOnlyLedgerHandle.java
```diff
@@ -207,4 +206,197 @@ public void readLastConfirmedComplete(int rc, long lastConfirmed, Object ctx) {
...
+        CompletableFuture<LedgerMetadata> f = new MetadataUpdateLoop(
+                clientCtx.getLedgerManager(), getId(),
+                this::getLedgerMetadata,
+                (metadata) -> metadata.isInRecovery(),
+                (metadata) -> {
+                    LedgerMetadataBuilder builder = LedgerMetadataBuilder.from(metadata);
+                    Optional<Long> lastEnsembleKey = metadata.getLastEnsembleKey();
+                    checkState(lastEnsembleKey.isPresent(),
+                               "Metadata shouldn't have been created without at least one ensemble");
+                    newEnsemblesFromRecovery.entrySet().forEach(
```
sijie: I would suggest being more strict here for safety concerns.

for check:

the first key in the `NewEnsembleFromRecovery` should exist in the LedgerMetadata read from zookeeper.

for updating:

- drop any ensembles before `lastEnsembleKey`
- replace ensemble `lastEnsembleKey`
- append ensembles after `lastEnsembleKey`

sijie: which thread run this? I don't think we have proper synchronization at `numEnsemblesFromRecovery`

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadOnlyLedgerHandle.java
sijie: nit: it is typically recommended to use `isEmpty` for size checking rather than calling `size`. because size on some data structures can be an expensive operation.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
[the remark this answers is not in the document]
sijie: add a unit test for this to make sure it capture the right exception code

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadOnlyLedgerHandle.java
ivankelly: First key in newEnsembleFromRecovery doesn't have to exist in the read metadata. For example, if a bookie fails during the recovery process.

I would exit screaming if there is any ensemble in newEnsembleFromRecovery before lastEnsembleKey. That would mean that, during the recovery process, something updated the last ensemble in ledger, yet didn't close the ledger. The only thing that can do that is a writing client. A writing client should *never* do this, as they should see that they are being fenced, and bail out. 

I'll add more sanity checks here though.

ivankelly: by the time this runs, no other thread will be adding anything to numEnsemblesFromRecovery. there could be cpu cache delays though, so will make it a synchronized treemap.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadOnlyLedgerHandle.java
ivankelly: Again, it doesn't actually matter. When this runs, no other thread will be accessing newEnsemblesFromRecovery. The iteration over it will be finished. getCurrentEnsemble() won't be called yet because the LedgerHandle isn't yet available to the client until after f completes.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadOnlyLedgerHandle.java
ivankelly: I don't think any synchronization is needed here either. newEnsemblesFromRecovery will only ever be greater than 0 during the recovery process, and during the recovery process, getCurrentEnsemble is only called in a callback (which run in the handle's executor, same as handleBookieFailure. That said, the whole threading model is a mess, so I'll add defensive synchronization until we clear it up.

ivankelly: sure, will change.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadOnlyLedgerHandle.java
ivankelly: will create the lock for locking this and the newEnsembles.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerRecoveryOp.java
```diff
@@ -125,51 +121,34 @@ public void readLastConfirmedDataComplete(int rc, RecoveryData data) {
...
-        if (!callbackDone.get()) {
+        if (!promise.isDone()) {
             startEntryToRead = endEntryToRead + 1;
             endEntryToRead = endEntryToRead + clientCtx.getConf().recoveryReadBatchSize;
             new RecoveryReadOp(lh, clientCtx, startEntryToRead, endEntryToRead, this, null)
                 .initiate();
         }
     }
 
-    private void closeAndCallback() {
```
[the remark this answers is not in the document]
ivankelly: I guess you found it below?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerRecoveryOp.java
```diff
@@ -96,7 +92,7 @@ LedgerRecoveryOp setEntryListener(ReadEntryListener entryListener) {
         return this;
     }
 
-    public void initiate() {
+    public CompletableFuture<Void> initiate() {
```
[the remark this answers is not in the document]
ivankelly: Sure

ivankelly:
rerun integration tests

sijie:
@athanatos @jvrao please review this

PR #1622, opened 2018-08-23
Add mock registration client

ivankelly:
For testing the client without having to boot zookeeper.

### DIFF
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/discover/MockRegistrationClient.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/discover/MockRegistrationClient.java
@@ -1,0 +1,164 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.bookkeeper.discover;
+
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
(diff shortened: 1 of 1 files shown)


bookkeeper-server/src/test/java/org/apache/bookkeeper/discover/MockRegistrationClient.java
```diff
@@ -0,0 +1,164 @@
...
+import org.apache.bookkeeper.versioning.Versioned;
+
+/**
+ * Mock implementation of registration client.
+ */
+public class MockRegistrationClient implements RegistrationClient {
+    /**
+     * Listener to receive changes from the registration service.
+     */
+    /*interface RegistrationListener {
```
eolivelli: Nit: this is to be removed

bookkeeper-server/src/test/java/org/apache/bookkeeper/discover/MockRegistrationClient.java
```diff
@@ -0,0 +1,164 @@
...
+        executor.submit(() -> {
+                readOnlyBookieWatchers.add(listener);
+                promise.complete(null);
+            });
+        return promise;
+    }
+
+    @Override
+    public void unwatchReadOnlyBookies(RegistrationListener listener) {
+        executor.submit(() -> {
```
eolivelli: Should we block until completion?

bookkeeper-server/src/test/java/org/apache/bookkeeper/discover/MockRegistrationClient.java
```diff
@@ -0,0 +1,164 @@
...
+
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import org.apache.bookkeeper.net.BookieSocketAddress;
+import org.apache.bookkeeper.versioning.LongVersion;
+import org.apache.bookkeeper.versioning.Versioned;
+
+/**
+ * Mock implementation of registration client.
```
eolivelli: What about adding a note that this implementation runs all of the callbacks in a single thread, to mimic how zk works?

bookkeeper-server/src/test/java/org/apache/bookkeeper/discover/MockRegistrationClient.java
ivankelly: Removed.

bookkeeper-server/src/test/java/org/apache/bookkeeper/discover/MockRegistrationClient.java
ivankelly: the interface indicates that we don't need to. This isn't used anywhere in bk, though dlog does use it in one place. for now I would prefer to leave it as is, and if in future the interface changes to allow waiting, change this too.

bookkeeper-server/src/test/java/org/apache/bookkeeper/discover/MockRegistrationClient.java
ivankelly: Added a comment.

eolivelli:
retest this please

Issue #1623, opened 2018-08-23
addEntryQuorumTimeout monitor callback leaks in ReadOnlyLedgerHandle and most LedgerHandle close calls


athanatos:
The LedgerHandle constructor schedules an addEntryQuorumTimeout check
    with the bk client scheduler. However, the only place this callback is
    canceled is in the closeAsync (the one which returns a future, not to be
    confused with asyncClose) method. asyncClose and close() both leak this
    callback. Moreover, ReadOnlyLedgerHandle invokes the LedgerHandle
    constructor and so also creates this callback, but it overrides close()
    and asyncClose() without passing them through.

Working on a fix.

PR #1624, opened 2018-08-23
ISSUE #1623: ReadOnlyLedgerHandle: don't schedule monitorPendingAddOps()

athanatos:
The LedgerHandle constructor schedules an addEntryQuorumTimeout check
with the bk client scheduler. However, the only place this callback is
canceled is in the closeAsync (the one which returns a future, not to be
confused with asyncClose) method. asyncClose and close() both leak this
callback. Moreover, ReadOnlyLedgerHandle invokes the LedgerHandle
constructor and so also creates this callback, but it overrides close()
and asyncClose() without passing them through.

ReadOnlyLedgerHandle already overrides
initializeExplicitLacFlushPolicy() to avoid write specific state.  This
patch generalizes that hack to initializeWriteHandleState() and the
cleanup to tearDownWriteHandleState().  tearDownWriteHandleState() is
moved into doAsyncClose(), which appears to be called for closes in
general.

(@rev cguttapalem@)
(@bug W-5362724@)

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
@@ -238,15 +238,25 @@ public class LedgerHandle implements WriteHandle {
                     return pendingAddOps.size();
                 }
             });
-        initializeExplicitLacFlushPolicy();
+
+        initializeWriteHandleState();
+    }
+
+    protected void initializeWriteHandleState() {
+        if (clientCtx.getConf().explicitLacInterval > 0) {
+            explicitLacFlushPolicy = new ExplicitLacFlushPolicy.ExplicitLacFlushPolicyImpl(
+                    this, clientCtx);
+        } else {
+            explicitLacFlushPolicy = ExplicitLacFlushPolicy.VOID_EXPLICITLAC_FLUSH_POLICY;
+        }
 
         if (clientCtx.getConf().addEntryQuorumTimeoutNanos > 0) {
             SafeRunnable monitor = new SafeRunnable() {
-                    @Override
-                    public void safeRun() {
-                        monitorPendingAddOps();
-                    }
-                };
+                @Override
+                public void safeRun() {
+                    monitorPendingAddOps();
+                }
+            };
             this.timeoutFuture = clientCtx.getScheduler().scheduleAtFixedRate(
                     monitor,
                     clientCtx.getConf().timeoutMonitorIntervalSec,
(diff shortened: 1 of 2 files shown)


athanatos:
Conflicted with 77b7fb870ef0fc2743dbe444c768c6c63bbe6875, rebasing.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -238,15 +238,27 @@ public Integer getSample() {
...
                 }
             });
-        initializeExplicitLacFlushPolicy();
+
+        initializeWriteHandleState();
+    }
+
+    protected void initializeWriteHandleState() {
+        if (!getLedgerMetadata().isClosed()
+            && !(this instanceof ReadOnlyLedgerHandle)
```
eolivelli: This is weird, can't we avoid?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -238,15 +238,27 @@ public Integer getSample() {
                     return pendingAddOps.size();
                 }
             });
-        initializeExplicitLacFlushPolicy();
+
+        initializeWriteHandleState();
+    }
+
+    protected void initializeWriteHandleState() {
+        if (!getLedgerMetadata().isClosed()
```
eolivelli: Is this possible? Do we have a test case?

ivankelly: It shouldn't be possible. in a write handle. In a read handle, it is.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
athanatos: Sure, might as well clean it up now that it doesn't get called in the other path.

athanatos:
@eolivelli Repushed with those guards removed.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
eolivelli: Seems to me that you didn't 'push'

athanatos:
@eolivelli Oops, actually pushed this time.  Sorry about that.

eolivelli:
rerun java8 tests

eolivelli:
retest this please

sijie:
rebuild java9

sijie:
@eolivelli this change might be needed to cherry-pick to branch-4.8. but there are quite a lot of conflicts, so I labelled it as `to-cherry-pick/4.8.0`. this need to manually cherry-pick and probably sent out a separate PR to review.

athanatos:
@sijie @eolivelli I can take care of cherry-picking it to 4.8, should be pretty close to the 4.7 version I need anyway.

sijie:
@athanatos that would be great! thank you!

PR #1627, opened 2018-08-25
Fix EntryLoggerAllocator.stop

reddycharan:

- make sure EntryLoggerAllocator.stop terminates pending allocatorExecutor's task.
Otherwise tests would flap, since even after EntryLoggerAllocator.stop, there is
possibility of new entrylog creation, which is not expected.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLoggerAllocator.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLoggerAllocator.java
@@ -206,11 +206,12 @@ class EntryLoggerAllocator {
         // wait until the preallocation finished.
         allocatorExecutor.shutdown();
         try {
-            if (!allocatorExecutor.awaitTermination(3, TimeUnit.SECONDS)) {
+            if (!allocatorExecutor.awaitTermination(5, TimeUnit.SECONDS)) {
                 log.warn("Timedout while awaiting for allocatorExecutor's termination, so force shuttingdown");
             }
         } catch (InterruptedException e) {
             log.warn("Got InterruptedException while awaiting termination of allocatorExecutor, so force shuttingdown");
+            Thread.currentThread().interrupt();
         }
         allocatorExecutor.shutdownNow();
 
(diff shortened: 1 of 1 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLoggerAllocator.java
```diff
@@ -203,6 +205,15 @@ private void setLastLogId(File dir, long logId) throws IOException {
     void stop() {
         // wait until the preallocation finished.
         allocatorExecutor.shutdown();
+        try {
+            if (!allocatorExecutor.awaitTermination(3, TimeUnit.SECONDS)) {
+                log.warn("Timedout while awaiting for allocatorExecutor's termination, so force shuttingdown");
+            }
+        } catch (InterruptedException e) {
+            log.warn("Got InterruptedException while awaiting termination of allocatorExecutor, so force shuttingdown");
```
eolivelli: Should be re-set the *interrupted* flag?.

I mean
Thread.currentThread().interrupt()

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLoggerAllocator.java
```diff
@@ -203,6 +205,15 @@ private void setLastLogId(File dir, long logId) throws IOException {
...
+        }
+        allocatorExecutor.shutdownNow();
```
dlg99: fun fact: shutdownNow() does not wait for termination, so awaitTermination() is still needed after.
So why don't we just do shutdownNow() on line 207 and wait after? 
there is nothing in that executor but prealloc tasks and we don't really need them to run at this point.

eolivelli: shutdownNow will *interrupt* all of the running threads. This will cause unpredictable logs/errors.

This way is cleaner

sijie:
@dlg99 @eolivelli please review it again.

sijie:

PR #1633, opened 2018-08-28
ISSUE #1623: ReadOnlyLedgerHandle: don't schedule monitorPendingAddOps()

### LINKED ISSUE #1623: addEntryQuorumTimeout monitor callback leaks in ReadOnlyLedgerHandle and most LedgerHandle close calls

athanatos:
The LedgerHandle constructor schedules an addEntryQuorumTimeout check
with the bk client scheduler. However, the only place this callback is
canceled is in the closeAsync (the one which returns a future, not to be
confused with asyncClose) method. asyncClose and close() both leak this
callback. Moreover, ReadOnlyLedgerHandle invokes the LedgerHandle
constructor and so also creates this callback, but it overrides close()
and asyncClose() without passing them through.

ReadOnlyLedgerHandle already overrides
initializeExplicitLacFlushPolicy() to avoid write specific state.  This
patch generalizes that hack to initializeWriteHandleState() and the
cleanup to tearDownWriteHandleState().  tearDownWriteHandleState() is
moved into doAsyncClose(), which appears to be called for closes in
general.

(rev cguttapalem)
(bug W-5362724)

Author: Samuel Just <sjust@salesforce.com>

Reviewers: Ivan Kelly <ivank@apache.org>, Enrico Olivelli <eolivelli@gmail.com>, Sijie Guo <sijie@apache.org>

This closes #1624 from athanatos/forupstream/wip-1623, closes #1623

(cherry picked from commit 9279dcc13b95406dbcb7a333fcb8a8961f52bf16)

Conflicts:
	bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java

Conflicts mainly revolved around refactors in LedgerHandle to remove the
explicit bkclient.

(cherry picked from commit e58e0747e93bed3c4bff5a2b866986fdba621c4e)

Conflicts:
	bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
@@ -217,31 +217,38 @@ public class LedgerHandle implements WriteHandle {
                                                   return pendingAddOps.size();
                                               }
                                           });
-        initializeExplicitLacFlushPolicy();
-
-        if (bk.getConf().getAddEntryQuorumTimeout() > 0) {
-            SafeRunnable monitor = new SafeRunnable() {
-                    @Override
-                    public void safeRun() {
-                        monitorPendingAddOps();
-                    }
-                };
-            this.timeoutFuture = bk.scheduler.scheduleAtFixedRate(monitor,
-                                                                  bk.getConf().getTimeoutMonitorIntervalSec(),
-                                                                  bk.getConf().getTimeoutMonitorIntervalSec(),
-                                                                  TimeUnit.SECONDS);
-        }
+        initializeWriteHandleState();
     }
 
     BookKeeper getBk() {
         return bk;
     }
 
-    protected void initializeExplicitLacFlushPolicy() {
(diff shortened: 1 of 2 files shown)


eolivelli:
@sijie as you are the release manager for 4.7, should we  cancel 4.7.2 and merge this one ?

sijie:
> @sijie as you are the release manager for 4.7, should we cancel 4.7.2 and merge this one ?

the vote has been out almost 3 days, there is no point to cancel 4.7.2 at this moment. also 4.8.0 is close to release, I don't think we should cancel 4.7.2. we can always cut 4.7.3 if needed.

sijie:
this PR is merged as dfe1232 in branch-4.7.

athanatos:
@sijie Does this look ok?

sijie:
@athanatos yes. this one was merged :)

PR #1641, opened 2018-09-04
Added module to enable CPU affinity

merlimat:
### Motivation

This is part of a set of changes aimed at reducing latency in BK at the expense of other aspects (eg: max throughput). While not intended to be used as default settings, they might be good to have whenever the latency becomes critical. 

Pinning a thread to a particular CPU will ensure no other process will execute on that CPU reducing all scheduler induced context switches that will cause latency jittery.

A given thread that wants to get pinned to a CPU just needs to call: 

```java
CpuAffinity.acquireCore();
```

It's called `acquireCore()` because it will also disable hyper-threading on the pinned cpu.

Subsequent PRs will use this module to have the option to pin critical threads to available CPUs.

### Changes
 * Added JNI module to call `sched_setaffinity()` to pin a thread to a particular CPU
 * Automatically discover available isolated CPUs
 * Acquire file-based locks to allow multiple processes on same machine to acquire CPUs independently.


cpu-affinity/src/main/java/org/apache/bookkeeper/utils/affinity/CpuAffinity.java
```diff
@@ -0,0 +1,48 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+
+package org.apache.bookkeeper.utils.affinity;
```
sijie: probably better to put it under the `org.apache.bookkeeper.common.utils.affinity` package. try to standardize common utils under `org.apache.bookkeeper.common`.

merlimat:
Ping. Please take a look again.

circe-checksum/src/main/java/com/scurrilous/circe/utils/NativeUtils.java
```diff
@@ -67,13 +67,16 @@ public static void loadLibraryFromJar(String path) throws Exception {
             throw new FileNotFoundException("Couldn't find file into jar " + path);
         }
 
-        OutputStream out = new FileOutputStream(temp);
         try {
-            while ((read = input.read(buffer)) != -1) {
-                out.write(buffer, 0, read);
+            OutputStream out = new FileOutputStream(temp);
```
eolivelli: Try-with-resources ? Or IoUtils or java.nio.files.Files

cpu-affinity/src/main/java/org/apache/bookkeeper/common/util/affinity/impl/CpuAffinityImpl.java
```diff
@@ -0,0 +1,210 @@
...
+
+    private static ProcessorsInfo processorsInfo = null;
+
+    public static synchronized void acquireCore() {
+        if (!SUPPORTED) {
+            throw new RuntimeException("CPU Affinity not supported in current environment");
+        }
+
+        if (!CpuAffinityJni.isRoot()) {
+            throw new RuntimeException("CPU Affinity can only be set if the process is running as root");
```
eolivelli: Is there any CAPABILITY to get this access without being root?

cpu-affinity/src/main/java/org/apache/bookkeeper/common/util/affinity/impl/CpuAffinityImpl.java
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    private static boolean checkSupported() {
+        try {
+            // Since this feature is only available in Linux, there's no point in checking for MacOS jnilib or Windows
+            // dll extensions
+            NativeUtils.loadLibraryFromJar("/lib/libcpu-affinity.so");
+            return CpuAffinityJni.isAvailable();
+        } catch (final Exception | UnsatisfiedLinkError e) {
+            return false;
```
eolivelli: Log some error?

circe-checksum/src/main/java/com/scurrilous/circe/utils/NativeUtils.java
merlimat: Fixed

cpu-affinity/src/main/java/org/apache/bookkeeper/common/util/affinity/impl/CpuAffinityImpl.java
merlimat: I have not been able to find a way other than being root in order to use `sched_setaffinity()`. The check here is just to provide more meaningful error message, if we can find a way to configure it we could remove this check.

cpu-affinity/pom.xml
```diff
@@ -0,0 +1,181 @@
...
+                <optimize>full</optimize>
+              </cpp>
+            </configuration>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
+
+    <profile>
+      <id>Linux</id>
```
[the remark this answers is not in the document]
merlimat: This pom was copied from the existing circe-checksum one. The profile name is not that important since it's automatically enabled based on the current OS.

cpu-affinity/src/main/affinity/cpp/affinity_jni.c
```diff
@@ -0,0 +1,98 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+
+// Use different error code to differentiate non-implemented error
+static const int NOT_IMPLEMENTED = -2;
```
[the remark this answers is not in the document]
merlimat: Yup, will reformat

cpu-affinity/src/main/java/org/apache/bookkeeper/common/util/affinity/impl/NativeUtils.java
```diff
@@ -0,0 +1,90 @@
...
+public class NativeUtils {
+    /**
+     * loads given library from the this jar. ie: this jar contains: /lib/pulsar-checksum.jnilib
+     *
+     * @param path
+     *            : absolute path of the library in the jar <br/>
+     *            if this jar contains: /lib/pulsar-checksum.jnilib then provide the same absolute path as input
+     * @throws Exception
+     */
+    public static void loadLibraryFromJar(String path) throws Exception {
```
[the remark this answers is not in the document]
merlimat: Yes, this code is similar to what we had imported with Circe-checksum, though I was trying to avoid interdependencies, dependencies on bk-common from this one.

merlimat:
@ivankelly @eolivelli @sijie Please take another look.

> However, I think we need an overarching document for these "performance" improvements, with benchmark results, detailed instructions to show how to reproduce the scenario in which performance changes, and the overall goals and approach for the perf improvements.

This is just the underlying implementation. I will work on the docs and instructions when the whole busy-wait change set is ready.

cpu-affinity/src/main/java/org/apache/bookkeeper/common/util/affinity/impl/CpuAffinityImpl.java
eolivelli: Okay

merlimat:
> Do you have some preliminary numbers then to show why this change is valuable?

@ivankelly 

This are the numbers on my local tests. It's not using the BK client lib directly, but rather through ManagedLedger (from Pulsar) which has one more thread handoff, to have callbacks serialized to 1 thread hashed on topic name.

Conditions:
 * Client/server on same machine with TCP connection
 * Tested on a bare-metal node
 * Client sends 1K rps (size 1KB) and waits for response
 * Multiple request outstanding, responses in order
 * Latency measured in millis

  | 50pct | 95pct | 99pct | 99.9pct | 99.99pct | 99.999pct | max
-- | -- | -- | -- | -- | -- | -- | --
Regular queue | 0.078 | 0.092 | 0.110 | 0.152 | 0.232 | 0.430 | 0.538
Spin-Wait Journal | 0.065 | 0.073 | 0.087 | 0.132 | 0.428 | 0.483 | 0.819
Spin-Wait Workers | 0.049 | 0.059 | 0.066 | 0.095 | 0.449 | 0.465 | 0.466
Spin-Wait IO Threads | 0.036 | 0.047 | 0.052 | 0.075 | 0.436 | 0.445 | 0.452
Disable HyperThread | 0.029 | 0.038 | 0.041 | 0.054 | 0.204 | 0.388 | 0.396
CPU-Affinity | 0.029 | 0.037 | 0.040 | 0.050 | 0.066 | 0.090 | 0.104

The remaining bulk of the median latency I think is related to SO_BUSY_POLL (netty/netty#8268) being ineffective on loopback interface.

merlimat:
run pr validation

ivankelly:
@merlimat thanks. Do you have any numbers for just this change? Like without the spin-wait stuff and HT disablement?

merlimat:
@ivankelly All the changes in the table above are incremental, so the last 2 lines show the effect of Cpu isolation: 

  | 50pct | 95pct | 99pct | 99.9pct | 99.99pct | 99.999pct | max
-- | -- | -- | -- | -- | -- | -- | --
Disable HyperThread | 0.029 | 0.038 | 0.041 | 0.054 | 0.204 | 0.388 | 0.396
CPU-Affinity | 0.029 | 0.037 | 0.040 | 0.050 | 0.066 | 0.090 | 0.104

I think it doesn't make sense to test cpu isolation without the other changes because the baseline noise will make the effects not visible. 

To summarize: cpu-affinity is used to remove interference from other tasks (OS thread and other processes) from a particular thread in order to avoid context switches and reduce the latency in the long tail. 

Without busy-poll, the bulk of context switches will still be there (eg: when sleeping while waiting for items in a queue), so the cpu-isolation won't provide much benefits.

PR #1642, opened 2018-09-04
Utility to rebuild interleaved storage index files

ivankelly:
We came across a case where the a ledger had been deleted from
zookeeper accidently. It was possible to recover the ledger metadata
from the zookeeper journal and old snapshots, but the bookies had
deleted the indices by this time. However, even if the index is
deleted, the data still exists in the entrylog.

This utility scans the entrylog to rebuild the index, thereby making
the ledger available again.


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -666,6 +659,15 @@ public Bookie(ServerConfiguration conf, StatsLogger statsLogger)
         this.ledgerMonitor = new LedgerDirsMonitor(conf, diskChecker, ledgerDirsManager);
         try {
             this.ledgerMonitor.init();
+
+            // if the index and ledger are sharing directories we can use the same monitor,
```
sijie: that change here is not correct.

if ledgerMonitor is throwing `NoWritableLedgerDirException` and transition to readonly mode, the idxMonitor will never be initialized.

tests/integration/smoke/src/test/java/org/apache/bookkeeper/tests/integration/TestBookieShellCluster.java
```diff
@@ -83,4 +93,98 @@ public void test002_ListROBookies() throws Exception {
...
+            try (WriteHandle writer = bk.newCreateLedgerOp().withEnsembleSize(1)
+                    .withWriteQuorumSize(1).withAckQuorumSize(1)
+                    .withPassword("".getBytes()).execute().get()) {
+                int i = 0;
+                for (; i < numEntries - 1; i++) {
+                    writer.appendAsync(("entry" + i).getBytes());
+                }
+                writer.append(("entry" + i).getBytes());
+
+                ledgerId = ledgerId;
```
sijie: this assignment is weird

tests/integration/smoke/src/test/java/org/apache/bookkeeper/tests/integration/TestBookieShellCluster.java
```diff
@@ -83,4 +93,98 @@ public void test002_ListROBookies() throws Exception {
...
+
+            log.info("Validate that we can read back");
+            try (ReadHandle reader = bk.newOpenLedgerOp()
+                    .withLedgerId(ledgerId)
+                    .withPassword("".getBytes()).execute().get();
+                 LedgerEntries entries = reader.read(0, numEntries - 1)) {
+                Assert.assertEquals(reader.getLastAddConfirmed(), numEntries - 1);
+
+                for (int i = 0; i < numEntries; i++) {
+                    Assert.assertEquals(new String(entries.getEntry(i).getEntryBytes()), "entry" + i);
```
sijie: add UTF_8; put expected value as the first param and actual value as the second param

tests/integration/smoke/src/test/java/org/apache/bookkeeper/tests/integration/TestBookieShellCluster.java
```diff
@@ -83,4 +93,98 @@ public void test002_ListROBookies() throws Exception {
...
+                ledgerId = ledgerId;
+            }
+
+            log.info("Restart bookies so index gets flushed");
+            assertTrue(BookKeeperClusterUtils.stopAllBookies(docker));
+            assertTrue(BookKeeperClusterUtils.startAllBookiesWithVersion(docker, currentVersion));
+
+            log.info("Find the bookie containing the index");
+            String indexFileName = String.format("/opt/bookkeeper/data/ledgers/current/0/0/%d.idx", ledgerId);
+            List<String> bookie = BookKeeperClusterUtils.allBookies().stream().filter(
```
sijie: nit: bookies

tests/integration/smoke/src/test/java/org/apache/bookkeeper/tests/integration/TestBookieShellCluster.java
```diff
@@ -83,4 +93,98 @@ public void test002_ListROBookies() throws Exception {
...
+
+            log.info("Validate that we can read back, after regeneration");
+            try (ReadHandle reader = bk.newOpenLedgerOp()
+                    .withLedgerId(ledgerId)
+                    .withPassword("".getBytes()).execute().get();
+                 LedgerEntries entries = reader.read(0, numEntries - 1)) {
+                Assert.assertEquals(reader.getLastAddConfirmed(), numEntries - 1);
+
+                for (int i = 0; i < numEntries; i++) {
+                    Assert.assertEquals(new String(entries.getEntry(i).getEntryBytes()), "entry" + i);
```
sijie: UTF_8

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/InterleavedStorageRegenerateIndexOp.java
```diff
@@ -0,0 +1,199 @@
...
+import org.apache.bookkeeper.util.DiskChecker;
+import org.apache.bookkeeper.util.SnapshotMap;
+import org.apache.commons.lang.time.DurationFormatUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Scan all entries in the entry log and rebuild the index file for one ledger.
+ */
+public class InterleavedStorageRegenerateIndexOp {
```
sijie: since we are providing a tool to mutate the state outside of main bookie process, we should provide a locking mechanism (e.g. using filelock) to prevent the tool can't mutate the state when the main bookie process is running. it doesn't necessarily to be done in this PR. but we should have a task to track that.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/InterleavedStorageRegenerateIndexOp.java
```diff
@@ -0,0 +1,199 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(InterleavedStorageRegenerateIndexOp.class);
+
+    private final ServerConfiguration conf;
+    private final long ledgerId;
+
+    public InterleavedStorageRegenerateIndexOp(ServerConfiguration conf, long ledgerId) {
```
sijie: I would suggest taking a `Set<Long>` of ledgers. because a single ledger is just one special case of a set with one element.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/InterleavedStorageRegenerateIndexOp.java
```diff
@@ -0,0 +1,199 @@
...
+
+        Set<Long> entryLogs = entryLogger.getEntryLogsSet();
+
+        int totalEntryLogs = entryLogs.size();
+        int completedEntryLogs = 0;
+        long startTime = System.nanoTime();
+
+        LOG.info("Scanning {} entry logs", totalEntryLogs);
+
+        AtomicLong firstEntry = new AtomicLong(Long.MAX_VALUE);
```
sijie: nit: since there is no concurrent requirements here, `firstEntry`, `lastEntry` and `numEntries` are just containers for primitive values. you can consider using `MutableLong` in this case. just fyi.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/InterleavedStorageRegenerateIndexOp.java
```diff
@@ -0,0 +1,199 @@
...
+
+                    if (LOG.isDebugEnabled()) {
+                        LOG.debug("Rebuilding {}:{} at location {} / {}", ledgerId, entryId, location >> 32,
+                                location & (Integer.MAX_VALUE - 1));
+                    }
+
+                    if (!ledgerCache.ledgerExists(ledgerId)) {
+                        // set dummy master key and set fenced.
+                        // master key is only used on write operations
+                        ledgerCache.setMasterKey(ledgerId, new byte[0]);
```
sijie: hmm. this would cause a problem, if we recover a ledger, but the writer comes in later and attempts to write entries, all the writes will be failing with auth failure. and the reader can't actually read them, right? do we have a test case to cover this negative case?

we might need to fix it or just add messages in the help messages of the tool, so people is aware of this situation.

sijie:
> We can cherry pick to 4.8 if you want

I don't think we need to cherry-pick this to 4.8. I think branch-4.8 is already cut and waiting for bug fixes only, so any feature-like issues should go to 4.9.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/InterleavedStorageRegenerateIndexOp.java
ivankelly: MutableLong from commons? I prefer to use JDK builtins if they do the same thing.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/InterleavedStorageRegenerateIndexOp.java
ivankelly: There's a little bit more hidden complexity here that I'm not sure how to resolve.

If a idx file has been deleted, but ledger metadata exists, when a client tries to open it and run recovery, it will see that it cannot read any entries, and close the ledger as empty.

I know DLog has a force option for this, but I think it would be better to provide another cli tool to allow users to mark the metadata from Closed -> In recovery, and then run the recovery again.

ivankelly: It looks like we don't actually check the password on fencing, we just blindly accept what the client passes in :/

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
sijie: it would be nice to create an issue to track your thought of the improvement. one single disk monitor is great.

tests/integration/smoke/src/test/java/org/apache/bookkeeper/tests/integration/TestBookieShellCluster.java
sijie: >  It's super annoying how junit and testng have different orders.

yeah, I felt annoyed as well

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/InterleavedStorageRegenerateIndexOp.java
sijie: @ivankelly yeah I think at twitter branch we actually have a FileLock change, not sure why I didn't port it back. Let me create an issue to track that. Created #1654 for this.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/InterleavedStorageRegenerateIndexOp.java
sijie: yeah. jdk builtins are fine

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/InterleavedStorageRegenerateIndexOp.java
sijie: > I think we need to pick up your effort to revamp the the tools, and 1) move them all into the same framework, 2) add tests & 3) add docs for each one.

let's pick up the tasks for #1000 

> If a idx file has been deleted, but ledger metadata exists, when a client tries to open it and run recovery, it will see that it cannot read any entries, and close the ledger as empty.

yeah. I think this is a rare corner case that we actuallly can do nothing for it.

> I know DLog has a force option for this, but I think it would be better to provide another cli tool to allow users to mark the metadata from Closed -> In recovery, and then run the recovery again.

sgtm

> It looks like we don't actually check the password on fencing, we just blindly accept what the client passes in :/

not sure we really need password for fencing, since it doesn't really append entries. and I think in general password should be removed eventually.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
ivankelly: #1655

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/InterleavedStorageRegenerateIndexOp.java
ivankelly: Hmm, yes, I'll push my other updates for now, and leave out the password stuff, since it has no effect.

ivankelly:
rerun validation tests

ivankelly:
rerun pr validation

PR #1667, opened 2018-09-07
Utility to dump/restore ledger metadata to/from file

### LINKED ISSUE #1604: Default ledger id formatter should be LongLedgerIdFormatter
Currently the default ledger id formatter in bookie shell is set to UUID formatter. I think that is only applicable when using LedgerAdv. so I think we need to change the default ledger id formatter to long ledger id formatter.

ivankelly:
Dumping will dump the serialized metadata to a file. Restoring will
parse the metadata from the file and write it to the metadata store.

Also changes the LedgerIdFormatter to default to long formatter (it
was UUID before, and we don't even currently support UUID ledger ids).

This fixes #1604

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
@@ -1107,6 +1107,10 @@ public class BookieShell implements Tool {
                 return -1;
             }
 
+            if (cmdLine.hasOption("dumptofile") && cmdLine.hasOption("restorefromfile")) {
+                System.err.println("Only one of --dumptofile and --restorefromfile can be specified");
+                return -2;
+            }
             runFunctionWithLedgerManagerFactory(bkConf, mFactory -> {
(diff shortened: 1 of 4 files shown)


sijie:
@ivankelly I linked this PR with #1604, since it fixes it

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
```diff
@@ -1115,6 +1095,8 @@ public void operationComplete(int rc, LedgerMetadata result) {
         LedgerMetadataCmd() {
             super(CMD_LEDGERMETADATA);
             lOpts.addOption("l", "ledgerid", true, "Ledger ID");
+            lOpts.addOption("dumptofile", true, "Dump metadata for ledger, to a file");
```
sijie: shall we add validation that `dumptofile` and `restorefromfile` are mutually exclusive

ivankelly: added

ivankelly:
rerun integration tests

Issue #1653, opened 2018-09-05
URI strings of the form zk://... should have the same behavior as strings of the form zk+null://


athanatos:
ZKMetadataDriverBase.resolveLedgerManagerFactory returns HierarchicalLedgerManagerFactory for uris of the form zk://...  This seems unintentional, zk:// should either be an error or should cause the client to infer the layout from zk.

athanatos:
Working on a patch.

PR #1668, opened 2018-09-07
ISSUE #1653: Clients with zk:// uri's should read layout from store

athanatos:
zk+null:// will cause the layout to be read from the store.  This patch
updates zk:// to do the same.  This matches past behavior with an
unspecified ledgerManagerFactoryClass.

Some tests needed to be updated with the new behavior.

(@bug W-5415120@)
(@rev cguttapalem@)


athanatos:
@sijie Ok to merge?

sijie:
yes

PR #1669, opened 2018-09-07
ISSUE 1662: LocalBookKeeper use layout manager and ledger path configs

athanatos:
The URI patch had a side effect of causing LocalBookKeeper to always
use a URI string of the form zk:// which defaults to HLM.  This
patch updates LocalBookKeeper to generate zk+<layout>:// based on
the legacy LEDGER_MANAGER_FACTORY_CLASS config when present.  +null
will be used otherwise and defaults to HLM.  Further, use the
(admittedly deprecated) zkLedgerRootPath config to determine the
ledger path (defaults to /ledger, so the behavior will be unchanged
if unspecified).

(@bug W-5415276@)
(@rev cguttapalem@)


athanatos:
@sijie Ok to merge?

sijie:
yes

PR #1670, opened 2018-09-07
Remove unnecessary log statement in LedgerDirsManager.getWritableLedgerDirs

reddycharan:

### Motivation

- when Bookie reaches diskusage threshold, LedgerDirsMonitor for every diskCheckInterval it logs NoWritableLedgerDirException call stack and message "All ledger directories are non writable".

### Changes

So remove unneccessary log statement in LedgerDirsManager.getWritableLedgerDirs and let the caller deal with the thrown exception.


eolivelli:
This is good to go

PR #1682, opened 2018-09-17
Added BlockingQueue implementation based on JCtools

merlimat:
### Motivation

Add a `BlockingQueue` implementation that is suitable for low latency and low contention. 

Key points:
 * Optimized for multiple producers and single consumer
 * When waiting for entries, the thread is blocked with busy wait to avoid context switch.

(This will be used in subsequent PRs to optionally enable it)


bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadEntryProcessorV3.java
```diff
@@ -236,6 +236,10 @@ protected ReadResponse getReadResponse() {
         } catch (IOException e) {
             LOG.error("IOException while reading entry: {} from ledger {} ", entryId, ledgerId, e);
             return buildResponse(readResponse, StatusCode.EIO, startTimeSw);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            LOG.error("InterruptedException while reading entry: {} from ledger {} ", entryId, ledgerId, e);
+            return buildResponse(readResponse, StatusCode.EIO, startTimeSw);
```
jvrao: Same EIO? What can cause interrupted exception here? other than obvious kill -9 <PID> ?

jvrao:
@merlimat any perf/measurements that you can share?

dlg99:
I am curious why you decided to implement blocking queue on top of JCtools' MPSC vs using disruptor that implements it. 
https://github.com/conversant/disruptor
https://github.com/conversant/disruptor/blob/master/src/test/java/com/conversantmedia/util/concurrent/DisruptorBlockingQueueTest.java

I do not have experience or stake in using either of those, but conversant claims to be faster and it already has blocking queue API.

eolivelli:
Same curiosity from me. Even Netty has its own MCSP queue

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadEntryProcessorV3.java
eolivelli: Kill -9 does not set interrupted flag on thread
This is a java artifact, it is a flag on the thread that mean 'please stop' and it is not a real signal from SO

sijie:
@merlimat can you answer the questions why we are choosing JCtools queues?

ivankelly:
@sijie @merlimat I suspect JCtools was chosen because this code was written before everyone jumped on the lock-free bandwagon ;)

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadEntryProcessorV3.java
merlimat: The interrupted exception could be caused when shutting down the bookie and, in this case, doing a fencing read. The fencing request will enqueued on the journal and the journal queue is now potentially blocking. If the read executor is being shutdown (when bookie exits) and the thread is blocked on enqueuing to the journal, it will get interrupted exception.

Previous behavior was to have always non-blocking queue that can grow indefinitely.

merlimat:
> @merlimat any perf/measurements that you can share?

@jvrao 
Not isolated to this implementation. It's tricky to properly microbenchmark a concurrency queue.

This really brings 2 benefits: 
 * Avoiding contention on the enqueuing threads (eg: IO threads passing items to journal, or submitting tasks the OrderedSafeExecutor)
 * Combined with a busy-poll strategy, it will avoid the latency penalty of context switch almost completely since the receiving thread will not "sleep" or wait for a new item in the queue, but it will be actively polling and never descheduled (this when also combining with CpuAffinity, see #1641).

Some test results were shared here: https://github.com/netty/netty/pull/8267#issuecomment-419538339 . Though this was including Pulsar in the mix, on top of BK, with same kind of optimizations applied.


> I am curious why you decided to implement blocking queue on top of JCtools' MPSC vs using disruptor that implements it.

@dlg99 @eolivelli I was not aware of conversant library. I was looking at Lmax disruptor but it forces to change all your code to fit into its model of dedicating a thread to it. I wanted to have a regular blocking queue so that this can also be directly plugged into the OrderedSafeExecutor (or any JDK executor), so that existing code doesn't have to be modified.

JCTool seems to be well maintained, the concurrency code is very scrupulously reviewed and is being used in several high profile projects, from Netty, Cassandra, Log4j, etc.

> Same curiosity from me. Even Netty has its own MCSP queue

Netty is actually importing JCTools queue :)

eolivelli:
Yes I noticed that Netty has its own copy of jctools queue.

I am fine with this idea.
Thank you for working on this topic

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/collections/BusyWait.java
```diff
@@ -0,0 +1,60 @@
...
+
+    private static final MethodHandle ON_SPIN_WAIT;
+
+    static {
+        MethodHandle handle = null;
+        try {
+            handle = MethodHandles.lookup().findStatic(Thread.class, "onSpinWait", MethodType.methodType(void.class));
+        } catch (Throwable t) {
+            // Ignore
+        }
```
eolivelli: We should log this error, so that it is clear that we are not able to leverage this feature

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDescriptorImpl.java
```diff
@@ -116,7 +117,7 @@ ByteBuf getExplicitLac() {
      * @param journal log the fence entry in the Journal
      * @return A future which will be satisfied when add entry to journal complete
      */
-    private SettableFuture<Boolean> logFenceEntryInJournal(Journal journal) {
+    private SettableFuture<Boolean> logFenceEntryInJournal(Journal journal) throws InterruptedException {
```
eolivelli: This method returns a Future, I wonder if it is okay to throw the Interrupted Exception or to return a failed Future

merlimat: Good point. will fix

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/collections/BusyWait.java
merlimat: This is not strictly critical for performance, it's just a way to consume less CPU power. 

Since it's not available in Java8, I'd rather keep this at debug level to avoid printing the warning every time (eg: even if busywait is not used).

merlimat:
run pr validation

merlimat:
run pr validation

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -1339,7 +1338,8 @@ public void writeComplete(int rc, long ledgerId, long entryId,
      * This method is idempotent. Once a ledger is fenced, it can
      * never be unfenced. Fencing a fenced ledger has no effect.
      */
-    public SettableFuture<Boolean> fenceLedger(long ledgerId, byte[] masterKey) throws IOException, BookieException {
+    public SettableFuture<Boolean> fenceLedger(long ledgerId, byte[] masterKey)
+            throws IOException, BookieException, InterruptedException {
```
eolivelli: Same problem here

merlimat: Good catch, fixed. I left the original IOException and BookieException as they were. We can argue they should be on the future as well, though that's unrelated with this PR.

merlimat:
@jvrao @dlg99 @sijie @ivankelly Are we good to merge this one?

merlimat:
run pr validation
run integrations tests

jvrao:
@dlg99 can you please review this? I wish we have some way to validate these perf changes. At lest to make sure it is not introducing perf regression. These changes are tricky - on paper they look like obvious win but in practice it is totally different picture. My comments are not specific to this change, but general category. @sijie

merlimat:
@jvrao all these changes are by default turned off

merlimat:
run integration tests

merlimat:
run integration tests

merlimat:
IGNORE IT CI

// Flaky TestCompatUpgrade integration test. Seen failing in other PRs as well

merlimat:
// Flaky TestCompatUpgrade integration test. Seen failing in other PRs as well

PR #1684, opened 2018-09-18
Fix license issues in master

ivankelly:
The only blocking error is that the GRPC NOTICE had not been bubbled
up to our notice.

The links for common-lang 3.6 and jackson 1.9.11 were also wrong, and
while these aren't required in license files, bad links makes it
harder to check if a dependency has been correct noted in the LICENSE
and NOTICE.

### DIFF
--- a/bookkeeper-dist/src/main/resources/LICENSE-all.bin.txt
+++ b/bookkeeper-dist/src/main/resources/LICENSE-all.bin.txt
@@ -275,30 +275,30 @@ Apache Software License, Version 2.
 - lib/com.yahoo.datasketches-sketches-core-0.8.3.jar [37]
 - lib/net.jpountz.lz4-lz4-1.3.0.jar [38]
 - lib/org.codehaus.jackson-jackson-core-asl-1.9.11.jar [39]
-- lib/org.codehaus.jackson-jackson-mapper-asl-1.9.11.jar [40]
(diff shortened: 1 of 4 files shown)


bookkeeper-dist/src/main/resources/LICENSE-all.bin.txt
```diff
@@ -337,18 +337,17 @@ Apache Software License, Version 2.
...
-[42] Source available at https://github.com/google/gson/tree/gson-parent-2.7
-[43] Source available at https://github.com/census-instrumentation/opencensus-java/tree/v0.11.0
-[44] Source available at https://github.com/square/okhttp/tree/parent-2.5.0
-[45] Source available at https://github.com/square/okio/tree/okio-parent-1.13.0
-[46] Source available at https://github.com/grpc/grpc-java/tree/v1.12.0
-[47] Source available at https://github.com/apache/curator/tree/apache-curator-4.0.1
-[48] Source available at https://github.com/inferred/FreeBuilder/tree/v1.14.9
-[49] Source available at https://github.com/google/error-prone/tree/v2.1.2
-[50] Source available at https://github.com/apache/yetus/tree/rel/0.5.0
+[39] Source available at https://github.com/codehaus/jackson/tree/d1c26ec538a8701cff7548009254fd6183b71873
```
eolivelli: good catch

eolivelli:

PR #1686, opened 2018-09-19
Shutdown appropriate EntryLogger instance.

reddycharan:

- EntryLogger's instance shutdown method should be called
- Otherwise test would flap, since in tearDown method, FileUtils.deleteDirectory can fail because of EntryLogFile PreAllocation task.


eolivelli:
retest this please

PR #1693, opened 2018-09-21
autorecovery-use-metadata-driver (part 1) : move AutoRecoveryMain to use MetadataBookieDriver

codingwangqi:

### Motivation

We are introducing Etcd as the metadata storage. However AutoRecovery currently is still tied to zookeeper. In order to use Etcd as the metadata storage, we have to move AutoRecovery related classes to use metadata driver API.

### Changes

This is the first change for changing AutoRecovery to use metadata driver. It changed AutoRecoveryMain to use metadata driver api and also removed the shutdown logic on session expired, which doesn't make any sense for current retryable zookeeper.

> ---
>
> ---

> ---
> quickly and easily:
>
>
>
> Otherwise:
> 
> 
> ---

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/AutoRecoveryMain.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/AutoRecoveryMain.java
@@ -20,6 +20,7 @@
  */
 package org.apache.bookkeeper.replication;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static org.apache.bookkeeper.replication.ReplicationStats.AUDITOR_SCOPE;
 import static org.apache.bookkeeper.replication.ReplicationStats.REPLICATION_WORKER_SCOPE;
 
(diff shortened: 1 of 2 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/AutoRecoveryMain.java
```diff
@@ -85,40 +85,36 @@ public AutoRecoveryMain(ServerConfiguration conf, StatsLogger statsLogger)
             throws IOException, InterruptedException, KeeperException, UnavailableException,
             CompatibilityException {
         this.conf = conf;
-        Set<Watcher> watchers = new HashSet<Watcher>();
-        // TODO: better session handling for auto recovery daemon  https://issues.apache.org/jira/browse/BOOKKEEPER-594
```
sijie: yeah, it doesn't really make sense to shutdown auto recovery when expired from zookeeper, especially if the underlying zookeeper is reconnectable. I am not worried about replication worker and the normal zookeeper metadata operations in auditor, however can you explain a bit after removing this part, how does Auditor elector handles the expire and retries?

codingwangqi: There is a logic in AuditorElector will shutdown itself when session expired : https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/AuditorElector.java#L203

when AuditorElector shuts itself down, the death watcher will quit and then AutoRecoveryMain will quit. so the session expire logic in AutoRecoveryMain is not really needed.

PR #1729, opened 2018-10-02
Decorate OrderedExecutor threads rather than runnables

ivankelly:
Actually, decorate the runnables too, but apply the decoration in a
decorator on the thread so that if a runnable/callable is passed to
the thread, the decoration will be applied.

This allows users of OrderedExecutor/Scheduler to use
executor.chooseThread() for CompletableFuture async handlers and still
get things like the MDC context.

### DIFF
--- a/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/OrderedExecutor.java
+++ b/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/OrderedExecutor.java
@@ -19,6 +19,7 @@ package org.apache.bookkeeper.common.util;
 
 import static com.google.common.base.Preconditions.checkArgument;
 
+import com.google.common.util.concurrent.ForwardingExecutorService;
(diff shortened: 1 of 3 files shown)


dlg99:
LGTM.
There is one related item to the executor story (not for this PR, of course): for each runnable we create new objects (depending on configuration, timed and/or context preserving runnables). Would be nice to use recycler for them.

sijie:
> for each runnable we create new objects (depending on configuration, timed and/or context preserving runnables). Would be nice to use recycler for them.

+1 for using recycler to improve them.

ivankelly:
:+1: Recycler sounds good. In general, the executor stuff could do with a bit of a refactor. The way OrderedScheduler inherits from OrderedExecutor rather than a common base is strange and means OrderedScheduler needs to do a bunch of janky casts. Once I clear my outstanding patch series I might take a look at it, along with other thread model stuff.

Issue #1737, opened 2018-10-04
EntryMemTable.newEntry retains reference to passed ByteBuffer array, can cause corruption on journal replay


athanatos:
Journal.scanJournal reuses the same ByteBuffer for each entry, but (if the entry is large enough) EntryMemTable retains a reference to that array for each affected entry.  The result is that if any entry is large enough, the bytes that get written to the entryLog upon flush for that entry will be a prefix of the last entry replayed from that journal file.  This will in general corrupt any such entries.  Fix seems relatively simple, PR will follow.

PR #1744, opened 2018-10-05
ISSUE #1737: EntryMemTable.newEntry: always make a copy

athanatos:
Retaining a reference to that array assumes that the caller
won't reuse the array for something else -- an assumption
violated by Journal.scanJournal and probably other callers.

(@bug W-5499346@)
(@rev cguttapalem@)


sijie:
run integration tests

sijie:
run integration tests

athanatos:
Will backport

PR #1746, opened 2018-10-08
ISSUE #1737: EntryMemTable.newEntry: always make a copy

### LINKED ISSUE #1737: EntryMemTable.newEntry retains reference to passed ByteBuffer array, can cause corruption on journal replay

athanatos:
Retaining a reference to that array assumes that the caller
won't reuse the array for something else -- an assumption
violated by Journal.scanJournal and probably other callers.

(bug W-5499346)
(rev cguttapalem)

Author: 

Reviewers: Enrico Olivelli <eolivelli@gmail.com>, Matteo Merli <mmerli@apache.org>, Sijie Guo <sijie@apache.org>

This closes #1744 from athanatos/forupstream/wip-1737, closes #1737

(cherry picked from commit df6595804cb1d78f5b531ae2db3a7c751439eb1c)

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryMemTable.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryMemTable.java
@@ -352,13 +352,8 @@ public class EntryMemTable {
         int offset = 0;
         int length = entry.remaining();
 
-        if (entry.hasArray()) {
-            buf = entry.array();
-            offset = entry.arrayOffset();
-        } else {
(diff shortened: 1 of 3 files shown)


sijie:
IGNORE IT CI

PR #1701, opened 2018-09-25
autorecovery-use-metadata-driver (part 4) : provide an `AuditorSelector` interface to abstract leader election logic for AuditorElector

codingwangqi:
Descriptions of the changes in this PR:

### Motivation

This is the last part for moving autorecovery to use metadata driver api.
Currently the only piece in autorecovery is tied to zookeeper is AuditorElector. This change
is to abstract the leader election logic and provide an interface.

### Changes

- abstract the leader election logic to an interface `AuditorSelector`
- move zookeeper related logic to `ZkAuditorSelector`

> ---
> In order to uphold a high standard for quality for code contributions, Apache BookKeeper runs various precommit
> checks for pull requests. A pull request can only be merged when it passes precommit checks. However running all
> the precommit checks can take a long time, some trivial changes don't need to run all the precommit checks. You
> can check following list to skip the tests that don't need to run for your pull request. Leave them unchecked if
> you are not sure, committers will help you:
>
> - [ ] [skip bookkeeper-server bookie tests]: skip testing `org.apache.bookkeeper.bookie` in bookkeeper-server module.
> - [ ] [skip bookkeeper-server client tests]: skip testing `org.apache.bookkeeper.client` in bookkeeper-server module.
> - [ ] [skip bookkeeper-server replication tests]: skip testing `org.apache.bookkeeper.replication` in bookkeeper-server module.
> - [ ] [skip bookkeeper-server tls tests]: skip testing `org.apache.bookkeeper.tls` in bookkeeper-server module.
> - [ ] [skip bookkeeper-server remaining tests]: skip testing all other tests in bookkeeper-server module.
> - [ ] [skip integration tests]: skip docker based integration tests. if you make java code changes, you shouldn't skip integration tests.
> - [ ] [skip build java8]: skip build on java8. *ONLY* skip this when *ONLY* changing files under documentation under `site`.
> - [ ] [skip build java9]: skip build on java9. *ONLY* skip this when *ONLY* changing files under documentation under `site`.
> ---

> ---
> Be sure to do all of the following to help us incorporate your contribution
> quickly and easily:
>
> If this PR is a BookKeeper Proposal (BP):
>
> - [ ] Make sure the PR title is formatted like:
>     `<BP-#>: Description of bookkeeper proposal`
>     `e.g. BP-1: 64 bits ledger is support`
> - [ ] Attach the master issue link in the description of this PR.
> - [ ] Attach the google doc link if the BP is written in Google Doc.
>
> Otherwise:
> 
[...]

PR #1747, opened 2018-10-09
Set ConnectionExpired Listener to MetadataClientDriver in AR

reddycharan:

- add setConnectionExpiredListener method to MetadataClientDriver interface.
- add listener to shutdown AR in the case of metadata connection expiry

### Motivation

This commit - 4f0d2a195bd9be3788876b47813cee1440cf005c, has removed the shutdown logic in AutoRecoveryMain incase of ZK client session expiry, with the following reason "which doesn't make any sense for current retryable zookeeper". But if the ZK session has expired then it is not completely correct to let AutoRecoveryMain to continue to run in that state.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/MetadataClientDriver.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/MetadataClientDriver.java
@@ -84,4 +84,23 @@ public interface MetadataClientDriver extends AutoCloseable {
     @Override
     void close();
 
+    /**
+     * Listener on listening the metadata client connection expiry event.
+     */
+    @FunctionalInterface
+    interface ConnectionExpiredListener {
+
+        /**
+         * Signal when connection is expired.
+         */
+        void onConnectionExpired();
+    }
+
+    /**
+     * sets listener for connection expiry event.
+     *
+     * @param connectionExpiredListener
+     *            listener listening on connection expiry.
+     */
+    void setConnectionExpiredListener(ConnectionExpiredListener connectionExpiredListener);
 }
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/MetadataClientDriver.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/MetadataClientDriver.java
@@ -85,22 +85,22 @@ public interface MetadataClientDriver extends AutoCloseable {
     void close();
 
     /**
-     * Listener on listening the metadata client connection expiry event.
+     * State Listener on listening the metadata client session states.
      */
     @FunctionalInterface
-    interface ConnectionExpiredListener {
+    interface SessionStateListener {
 
         /**
-         * Signal when connection is expired.
+         * Signal when client session is expired.
          */
-        void onConnectionExpired();
+        void onSessionExpired();
     }
 
     /**
-     * sets listener for connection expiry event.
+     * sets session state listener.
      *
-     * @param connectionExpiredListener
-     *            listener listening on connection expiry.
+     * @param sessionStateListener
+     *            listener listening on metadata client session states.
      */
-    void setConnectionExpiredListener(ConnectionExpiredListener connectionExpiredListener);
+    void setSessionStateListener(SessionStateListener sessionStateListener);
 }
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/AutoRecoveryMain.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/AutoRecoveryMain.java
@@ -93,7 +93,7 @@ public class AutoRecoveryMain {
         this.conf = conf;
         this.bkc = Auditor.createBookKeeperClient(conf);
         MetadataClientDriver metadataClientDriver = bkc.getMetadataClientDriver();
-        metadataClientDriver.setConnectionExpiredListener(() -> {
+        metadataClientDriver.setSessionStateListener(() -> {
             LOG.error("Client connection to the Metadata server has expired, so shutting down AutoRecoveryMain!");
             shutdown(ExitCode.ZK_EXPIRED);
         });
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/AutoRecoveryMain.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/AutoRecoveryMain.java
@@ -40,6 +40,7 @@ import org.apache.bookkeeper.common.component.ComponentStarter;
 import org.apache.bookkeeper.common.component.LifecycleComponent;
 import org.apache.bookkeeper.common.component.LifecycleComponentStack;
 import org.apache.bookkeeper.conf.ServerConfiguration;
+import org.apache.bookkeeper.meta.MetadataClientDriver;
 import org.apache.bookkeeper.replication.ReplicationException.CompatibilityException;
 import org.apache.bookkeeper.replication.ReplicationException.UnavailableException;
 import org.apache.bookkeeper.server.conf.BookieConfiguration;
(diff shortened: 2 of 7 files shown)


reddycharan:
@athanatos created WIP PR for this concern.

sijie:
I share same question as @eolivelli ; the underlying zookeeper client has infinite retry logic. if session expired on zookeeper, when auditor retries later, it will eventually succeed when established connections to zookeeper.

do you guys experience any problems on this? or do you have any tests that reproduce this issue?

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/MetadataClientDriver.java
```diff
@@ -84,4 +84,23 @@ LedgerManagerFactory getLedgerManagerFactory()
     @Override
     void close();
 
+    /**
+     * Listener on listening the metadata client connection expiry event.
+     */
+    @FunctionalInterface
```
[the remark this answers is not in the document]
athanatos: I actually think that a notification that events may have been dropped is a fairly general concept.  SessionStateListener might be better though.

reddycharan:
@eolivelli @sijie 

**Issue Summary:** I added the test case validating the need of this fix. As you can notice in this test case, with out this fix, when AR (with Auditor)’s ZK session expires, then as expected other AR will become Auditor and start running Auditor. But unexpectedly the previous Auditor also continue to run.

If AR is **not Auditor**, then because of the following ElectionWatcher - https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/AuditorElector.java#L224 in the case of ZK session expiry, AuditorElector will shutdown and hence AutoRecoveryMain will shutdown because of https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/AutoRecoveryMain.java#L230.

If AR is **Auditor**, then we are not setting ElectionWatcher in AuditorElector and hence above mentioned path wouldn’t execute. So AR and its Auditor will continue to run.

But the ‘myVote’ ephemeral node of the Auditor’s AR would be deleted in the case of ZK session expiry and hence some other AR will get this event and would become Auditor.

So effectively two AR’s are running Auditors now :( 

In summary just like what we are doing in the case of non-Auditor AR, we should shutdown the AR in the Auditor case as well. 

To be honest, I never fully grasped the semantics, guarantees and invariants of ZooKeeperClient wrapper class, especially regarding ’connectRetryPolicy’. @athanatos  has some strong concerns with how things (watchers, ephemeral nodes) are handled in ZooKeeperClient when connection is reestablished. Also, frequently I see people making misassumptions with what ZooKeeperClient offers.

@jvrao @codingwangqi fyi

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/MetadataClientDriver.java
```diff
@@ -84,4 +84,23 @@ LedgerManagerFactory getLedgerManagerFactory()
...
+    interface ConnectionExpiredListener {
+
+        /**
+         * Signal when connection is expired.
+         */
+        void onConnectionExpired();
```
eolivelli: I would call this simply 'onFatalError' or something that is simply telling that service should suicide.

Session expiration, and maybe session, is a ZK concept.

eolivelli:
Okay you convinced me.

Let's work on naming and then ship this fix

athanatos:
Generally speaking, the transparent reconnect logic in ZooKeeperClient is dangerous in two ways:
1) If the user isn't aware of what's going on, ephemeral nodes and watches can silently disappear.
2) Writes are generally non-idempotent, so retrying them can result in erroneous error conditions, as with node creation returning an EEXIST when it actually succeeded.

It's easy to reproduce errors like 2 by introducing intermittent packet drops while creating ledgers.

sijie:
@athanatos : yes. the users of ZooKeeperClient should be aware of the two points you listed.

reddycharan:
@sijie can you comment on validity of this fix and feedback on interface method name

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/AutoRecoveryMain.java
```diff
@@ -91,6 +92,11 @@ public AutoRecoveryMain(ServerConfiguration conf, StatsLogger statsLogger)
             CompatibilityException {
         this.conf = conf;
         this.bkc = Auditor.createBookKeeperClient(conf);
+        MetadataClientDriver metadataClientDriver = bkc.getMetadataClientDriver();
+        metadataClientDriver.setConnectionExpiredListener(() -> {
```
sijie: I think a cleaner solution is AuditorElector handling this session expires. because AuditorElector already has this session expires watcher to shutdown elector - https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/AuditorElector.java#L224 so all the leader election logic are all in one place, rather than spreading across multiple places. and it seems to be working better with #1701 (https://github.com/apache/bookkeeper/pull/1701/files#diff-7d01af204bb0a30c0193b46c737f8cbfR49)

However I am fine with this change as it is working well with the zookeeper concepts. We can go with this change for now, once the change #1701 is merged, we can considering moving this session expires logic in to the AuditorElector, which will provide a much cleaner abstraction.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/MetadataClientDriver.java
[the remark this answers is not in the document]
sijie: IMO SessionStateListener is a better name than ConnectionExpiredListener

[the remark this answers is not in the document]
eolivelli: +1 for SessionStateListener

metadata-drivers/etcd/src/main/java/org/apache/bookkeeper/metadata/etcd/EtcdMetadataClientDriver.java
```diff
@@ -72,4 +72,12 @@ public synchronized void close() {
         }
         super.close();
     }
+
+    @Override
+    public void setSessionStateListener(SessionStateListener sessionStateListener) {
+        /*
+         * TODO: EtcdMetadataClientDriver has to implement this method.
+         */
+        throw new UnsupportedOperationException();
```
reddycharan: @codingwangqi im just throwing exception for now. As part of metadataclientdriver/etcd changes can you please take care of this.

@sijie

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/MetadataClientDriver.java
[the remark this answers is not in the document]
reddycharan: changed interface/method name

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/AutoRecoveryMain.java
codingwangqi: @sijie thank you for pointing this out. yes, I think session expires is an implementation specific behavior, I will incorporate your proposal into #1701 once this change is merged.

metadata-drivers/etcd/src/main/java/org/apache/bookkeeper/metadata/etcd/EtcdMetadataClientDriver.java
codingwangqi: @reddycharan thank you for letting me know about this. I will take care of it.

reddycharan:
@eolivelli need your final approval

reddycharan:
@sijie i think you approved this. So it should be good to go.

eolivelli:
#shipit

PR #1751, opened 2018-10-16
Modules dependencies do not need to be included in checksum Jar

merlimat:
The current assembly configuration is including the jars of the modules dependency inside its own jar.

These jars are completely ignored, but they take up non trivial space.

With this change the size of `circe-checksum.jar` goes from 2.8 MB to 62 KB.


sijie:
run integration tests

sijie:
run integration tests

merlimat:
run integration tests

sijie:
IGNORE IT CI

PR #1753, opened 2018-10-19
Adding multi-node ZKCluster test util class

reddycharan:

- adding multi-node ZKCluster util class
- adding new interface for ZKCluster util and having separate implementaions for
single node ZKCluster and multi-node ZKCluster

### Motivation

This helps in adding more test coverage to Zookeeper aspects.

### DIFF
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/test/ZooKeeperCluster.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/test/ZooKeeperCluster.java
@@ -1,0 +1,79 @@
+/*
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+package org.apache.bookkeeper.test;
+
+import static org.apache.bookkeeper.util.BookKeeperConstants.AVAILABLE_NODE;
+import static org.apache.bookkeeper.util.BookKeeperConstants.READONLY;
+
+import java.io.IOException;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
+import org.apache.bookkeeper.zookeeper.ZooKeeperWatcherBase;
+import org.apache.zookeeper.CreateMode;
+import org.apache.zookeeper.KeeperException;
+import org.apache.zookeeper.Transaction;
+import org.apache.zookeeper.ZooDefs.Ids;
+import org.apache.zookeeper.ZooKeeper;
+
+/**
+ * Interface for ZooKeeperCluster.
+ */
+public interface ZooKeeperCluster {
+    ZooKeeper getZooKeeperClient();
+
+    String getZooKeeperConnectString();
+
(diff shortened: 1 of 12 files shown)


bookkeeper-server/src/test/java/org/apache/bookkeeper/test/ZooKeeperClusterUtil.java
```diff
@@ -0,0 +1,140 @@
...
+    protected ZooKeeper zkc; // zookeeper client
+
+    public static void enableZookeeperTestEnvVariables() {
+        /*
+         * org.apache.zookeeper.test.ClientBase uses FourLetterWordMain, from
+         * 3.5.3 four letter words are disabled by default due to security
+         * reasons
+         */
+        System.setProperty("zookeeper.4lw.commands.whitelist", "*");
+        System.setProperty("zookeeper.admin.enableServer", "false");
```
eolivelli: What about adding zookeeper.forceSync=no ?
We do not care about fsync in tests

bookkeeper-server/src/test/java/org/apache/bookkeeper/test/ZooKeeperCluster.java
```diff
@@ -0,0 +1,79 @@
...
+
+    String getZooKeeperConnectString();
+
+    String getMetadataServiceUri();
+
+    String getMetadataServiceUri(String zkLedgersRootPath);
+
+    String getMetadataServiceUri(String zkLedgersRootPath, String type);
+
+    void startServer() throws Exception;
```
eolivelli: What about renaming to startServers() or startCluster() ?

sijie: I would suggest `startCluster`. and also change `stopServer`, `restartServer`, `killServer` to `stopCluster`, `restartCluster` and `killCluster`.

reddycharan: yeah, but the issue is I've to change in 100's of places where these methods are called in existing testcode..will check

eolivelli: I hope your IDE will help....

bookkeeper-server/src/test/java/org/apache/bookkeeper/test/ZooKeeperClusterUtil.java
reddycharan: from doc -

"Unsafe Options
The following options can be useful, but be careful when you use them. The risk of each is explained along with the explanation of what the variable does.
forceSync
(Java system property: zookeeper.forceSync)
Requires updates to be synced to media of the transaction log before finishing processing the update. If this option is set to no, ZooKeeper will not require updates to be synced to the media."

maybe worth considering, but this must not be specific to multinode zk cluster. So I'll leave it now for this change. We can revisit if it is needed for both kind of zkclusters (singlenode and multinode).

bookkeeper-server/src/test/java/org/apache/bookkeeper/test/ZooKeeperCluster.java
reddycharan: done.

bookkeeper-server/src/test/java/org/apache/bookkeeper/test/ZooKeeperClusterUtil.java
eolivelli: This code is only for test cases. You don't really need to fsync in test cases, as processes do not crash.
Waiting/forcing an fsync at every zk write is useless in our tests corpus, isn't it?
I mean, this is not a production zk cluster

reddycharan: couple of things,

1) people may take snapshot of ZK for analysis of test failure, so not fsyncing might be an issue for them.
2) even if we decide not to do fsync, i think it should be in different commit with full details, since it affects the behavior of single noded ZKCluster utility class also. So I'm not inclined to make that change in this PR.

eolivelli: Okay.
I usually just add the sys prop on the command line.
We can think about changing the command line on CI.
Thank you for your explanations

sijie:
IGNORE IT CI

PR #1760, opened 2018-10-25
Kill LedgerMetadata#isConflictWith

ivankelly:
It doesn't make sense anymore, as local copies of the metadata are
never modified, only updated with the latest version from the metadata
store.

In effect, this logic has been broken out to the places where we try
to update the metadata store copy. Each time we try to update the
metadata store, we ensure that the update we are applying still makes
sense with regard to the copy of the metadata we are updating.

Master issue: #281


ivankelly:
rerun integration tests

ivankelly:
rerun integration tests

sijie:
IGNORE IT CI

PR #1762, opened 2018-10-26
Don't cache Bookie hostname DNS resolution forever

merlimat:
### Motivation

`BookieSocketAddress` is resolving the bookie DNS name in its constructor and then using the already resolved `InetSocketAddress` instance.

If the IP of a bookie changes, the BK client will continue to use the old IP address.

### Changes

Construct a new `InetSocketAddress` each time `getSocketAddress()` gets called (eg: each time we attempt to make a new connection) so that we're making sure to get the right IP.

I cannot think of a good way to add unit test for this at this point, suggestions are welcome. 

I think this should be included in a patch release as well 4.7.3 or 4.8.1

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/net/BookieSocketAddress.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/net/BookieSocketAddress.java
@@ -39,13 +39,10 @@ public class BookieSocketAddress {
     private final String hostname;
     private final int port;
 
-    private final InetSocketAddress socketAddress;
-
     // Constructor that takes in both a port.
     public BookieSocketAddress(String hostname, int port) {
         this.hostname = hostname;
         this.port = port;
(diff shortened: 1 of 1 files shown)


merlimat:
@eolivelli Added comment

merlimat:
@dlg99 `-Dnetworkaddress.cache.ttl` cannot be passed as system property from CLI but rather included in `java.security` property file. 

In any case, the current behavior is caching the DNS resolution forever in the BK client process. That comes before reaching the JVM DNS cache or the OS DNS cache. By creating new instances of `InetSocketAddress` we will rely on the JVM cache, that depends on the `networkaddress.cache.ttl` setting. 

(BTW: I think the JVM cache is a *bad* thing on its own, but that's a different story... :) )

merlimat:
run integration tests

sijie:
merged to master, branch-4.8 and branch-4.7

PR #1765, opened 2018-10-27
Use default metric registry in Prometheus exporter

merlimat:
### Motivation

Currently the Prometheus exporter is using a private metrics register. This doesn't work well for BK client where we want to expose in the same registry of application, or when we want to interact with other tools that use the default static registry (eg: log4j, jetty integrations for prometheus).


bookkeeper-stats-providers/prometheus-metrics-provider/src/main/java/org/apache/bookkeeper/stats/prometheus/PrometheusMetricsProvider.java
```diff
@@ -128,10 +128,17 @@ public void start(Configuration conf) {
...
-        new StandardExports().register(registry);
-        new MemoryPoolsExports().register(registry);
-        new GarbageCollectorExports().register(registry);
-        new ThreadExports().register(registry);
+        try {
+            new StandardExports().register(registry);
+            new MemoryPoolsExports().register(registry);
+            new GarbageCollectorExports().register(registry);
+            new ThreadExports().register(registry);
+        } catch (Exception e) {
```
sijie: this approach will cause some exports are not registered. for example, if `StandardExports` was registered but no the others, this approach will cause the others being skipped.

sijie: I would suggest having a common method that catches exception. 

```
public void registerExports(Exports exports, CollectorRegistry registry) {
     try {
            exports.register(registry);
     } catch (Exception e) {
            ....
     }

}
```

merlimat: @sijie Fixed

merlimat:
run java9 tests

merlimat:
rebuild java9

sijie:
run integration tests

merlimat:
run integration tests

eolivelli:
@sijie. Do you want this on branch 4.8?

sijie:
@eolivelli yes. I was in the process of merging it. :)

Issue #1781, opened 2018-11-02
Add traces/zipkin support to simplify troubleshooting of slow requests


dlg99:
**FEATURE REQUEST**

1. Please describe the feature you are requesting.

Add traces/zipkin support to simplify troubleshooting of slow requests

2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have). Are you currently using any workarounds to address this issue?

Should have.

3. Provide any additional detail on your proposed use case for this feature.

Tracing compliments metrics framework by enabling one's deep dive into performance details of single request. Zipkin provides UI to view such traces though there are other implementations.
Opentracing provides framework for tracing that can be used with various tracing backends, zipkin included.

PR #1782, opened 2018-11-02
Issue #1781: Add traces/zipkin support to simplify troubleshooting of slow requests

dlg99:

Instrumented bookie and the client with tracing, added configuration to enable tracing.

### Motivation

Tracing compliments metrics framework by enabling one's deep dive into performance details of single request. Zipkin provides UI to view such traces though there are other implementations.
Opentracing provides framework for tracing that can be used with various tracing backends, zipkin included.

### Changes

Instrumented bookie and the client with tracing, added configuration to enable tracing.
Added unit tests.

Master Issue: #1781 

> ---
>
> ---

> ---
> quickly and easily:
>
>
>
> Otherwise:
> 
> 
> ---

### DIFF
--- a/bookkeeper-common/pom.xml
+++ b/bookkeeper-common/pom.xml
@@ -65,6 +65,37 @@
       <artifactId>commons-lang3</artifactId>
       <scope>test</scope>
     </dependency>
+    <!-- tracing dependencies -->
+    <dependency>
+      <groupId>io.opentracing</groupId>
+      <artifactId>opentracing-api</artifactId>
+      <version>${opentracing.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>io.opentracing</groupId>
+      <artifactId>opentracing-util</artifactId>
+      <version>${opentracing.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>io.zipkin.brave</groupId>
+      <artifactId>brave</artifactId>
+      <version>${zipkin.brave.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>io.zipkin.reporter2</groupId>
+      <artifactId>zipkin-sender-urlconnection</artifactId>
+      <version>${zipkin.reporter.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>io.opentracing.brave</groupId>
+      <artifactId>brave-opentracing</artifactId>
+      <version>${opentracing.version}</version>
+    </dependency>
+
+    <dependency>
+      <groupId>com.google.protobuf</groupId>
+      <artifactId>protobuf-java</artifactId>
+    </dependency>
   </dependencies>
   <build>
     <plugins>
--- a/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/OrderedExecutor.java
+++ b/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/OrderedExecutor.java
@@ -26,6 +26,11 @@ import com.google.common.util.concurrent.ThreadFactoryBuilder;
 
 import io.netty.util.concurrent.DefaultThreadFactory;
 
+import io.opentracing.Scope;
+import io.opentracing.Span;
+import io.opentracing.Tracer;
+import io.opentracing.util.GlobalTracer;
+
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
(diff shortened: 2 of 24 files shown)


merlimat:
I've seen in many places this is using multiple builder object to track operations in different places. I would like to make sure there is no overhead (like in objects allocations) if the tracing feature is disabled.

dlg99:
@merlimat I agree with this concern.
NOOP tracer should not create an overhead (it does not create span/context objects and uses same noop instance as far as I can tell, it does not set active trace as thread local etc.)
If actual tracer is used but sample is set to 0 we'll end up with traces created but not reported (unless somewhere along the way app created active trace that had "sample flag" set.
There are some places where tags are set and i.e. string concatenation is used to build the tag. I'll  move these out to make sure string concatenation does not happen if tracing is a NOOP. Otherwise tags on noop spans/context do nothing. 

@sijie Let me make sure that I understood you correctly.
Pretty much we need better way to specify tracer, so something similar to metrics in BK where one can drop in a jar and specify class name to use. API along the lines of createTracer, contextToByteString, byteStringToContext.
I agree, I'll do that. I short-circuited this part to get review out sooner.
Opentracing stays as the tracing API. I treat it same way as we do slf4j for logging. I figured since other apache projects use it directly we can do too, i.e. https://github.com/apache/hadoop/compare/trunk...carlosalberto:ot_initial_integration#diff-1f7d0efa26604bfd48d1754534a77becR262
Users can implement their version of Tracer (i.e. opencensus) and register it as GlobalTracer.
i.e. in current implementation any application that uses bk client will have to register tracer on its own, it does not happen as part of client creation/builder.

dlg99:
@sijie 
> more generic to intercept bookkeeper operations

as in any possible step? i.e. traces pass context through executor etc, not sure what value it provides outside of the context of higher level trace. Either way, one can create Tracer that does some internal magic with traces, register that Tracer as a global Tracer.

as in very specific points of bookie/client process (add start-finish, remote call start-finish etc)? 
This is related to tracing but not 100% the same. I.e. in traces I am interested to see wait/run times on executor etc., in case of such specific points executors are noise.

I am not familiar with pulsar, I took a very brief look at https://github.com/apache/pulsar/blob/7a92111fe7824d422d1739abd06d85955b652212/pulsar-broker/src/test/java/org/apache/pulsar/client/api/InterceptorsTest.java where interceptors seem to be events that happen on start/finish of some operations. 
Using Storm as analogy, interceptors seem like a good idea to customize something when a bolt receives/sends a message letting whoever implements the bolt to deal with it internally. If I wanted to trace storm's internals the same way, interceptors seem to become nuisance. 
Or using servlet analogy, interceptors are like servlet filters that allow one to i.e. change headers or do something at specific points. 
I.e. if we are not interested in too many details abut the message or in changing the message and only need start/finish points of some events, we can use existing bookie's metrics framework to intercept such events. It did not work for me because I wanted to have flexibility in setting parent/child relationships.
Interceptors could be a good feature to add but not necessarily as replacement or something dependent on tracing.

TL;DR:
There is probably some specific usecase that I am failing to catch right away but it feels like interceptors vs traces have some overlap but not complete and my immediate needs closer to traces.
It also feels like interceptors are a great idea for i.e. streaming processing systems where big chunk of the code dealing with data is created by external user but that's not bookkeeper's case as far as I see it.

sijie:
@dlg99 : 

interceptor is usually a pattern that keep the core business logic separated from any other argumented logics (e.g. tracing, filtering, monitoring). so argumented logic can be easier to unit test rather than coupling with the core runtime. 

All what we are caring here is client operations (e.g. metadata operations, data operations) and server side operations (e.g. storage operations, journal operation) and executor operations. If we define some sort of interceptor interfaces for those key operations, it is easier for us to add a tracing interceptor. 

There are a few benefits have interceptors:

1) we can easily add more types of interceptors in future.
2) people can implement their own tracing interceptors. people is not enforced to use open tracing, and be able to connect with their in-house tracing solution.
3) even more, when bookkeeper is used in other services (e.g. databases, messaging systems), what people really want is not just bookkeeper's own tracing events. people want to do end-to-end tracing, in that sense, if we provide interceptor interface, people can build their tracing interceptor to push their upstreaming trace context all the wya down to bookkeeper. it is hard to do so if we just use tracer here.

Based on these considerations, I think a better approach will be:

- define a set of interceptor interfaces, providing methods to intercept key operations.
- the open tracing implementation should be one interceptor implementation.

Other comments on your comment:

>  It did not work for me because I wanted to have flexibility in setting parent/child relationships.

An interceptor interface doesn't limit the flexibility. You are still using open tracing api to set parent/child relationship.

> Interceptors could be a good feature to add but not necessarily as replacement or something dependent on tracing.

A better way to think about interceptor vs tracing. Interceptor is the interface, while tracing is the implementation. I am not saying interceptors replacing tracing. what I am suggesting is a common interceptor interface is allowing more extensibility, rather than directly sticking to one tracing api.

> There is probably some specific usecase that I am failing to catch right away but it feels like interceptors vs traces have some overlap but not complete and my immediate needs closer to traces.

End-to-end tracing is one of the use cases when bookkeeper is used as one piece in the whole system.

 Context propagation will be important in such case, providing interceptors rather than directly implementing tracing will be more flexible for people to do proper context propagation and use proper tracing framework (e.g. open source, in house).

Allow any in-house implementations that do tracing but doesn't use any open source tracing api.

>

dlg99:
I think I have an idea what to do, it won't be exactly interceptors fw but rather an extension of current metrics. I'll do the POC as time allows and we'll look at it.

PR #1788, opened 2018-11-06
Fix bugs in DefaultEnsemblePlacementPolicy

reddycharan:

- bookieInfoMap is not initialized and newEnsemble will throws BKNotEnoughBookiesException if
diskWeightBasedPlacement is enabled
- add test coverage for DefaultEnsemblePlacementPolicy with diskWeightBasedPlacement enabled

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/DefaultEnsemblePlacementPolicy.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/DefaultEnsemblePlacementPolicy.java
@@ -22,6 +22,7 @@ import io.netty.util.HashedWheelTimer;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -58,6 +59,7 @@ public class DefaultEnsemblePlacementPolicy implements EnsemblePlacementPolicy {
     private final ReentrantReadWriteLock rwLock;
 
     DefaultEnsemblePlacementPolicy() {
+        bookieInfoMap = new HashMap<BookieSocketAddress, WeightedObject>();
         rwLock = new ReentrantReadWriteLock();
     }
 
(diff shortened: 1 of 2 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/DefaultEnsemblePlacementPolicy.java
```diff
@@ -92,6 +94,9 @@
                     }
                     newBookies.add(b);
                     --ensembleSize;
+                    if (ensembleSize == 0) {
+                        return newBookies;
```
dlg99: simple ```return newBookies;``` after the while loop should be enough.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/DefaultEnsemblePlacementPolicy.java
```diff
@@ -92,6 +94,9 @@
                     }
```
dlg99: is it possible for ```weightedSelection.getNextRandom()``` to return exactly the same one bookie which i.e. is already in the newBookies or excludeBookies thus resulting in the infinite loop?

reddycharan: getNextRandom should eventually return different bookie. 

check the logic here - https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/WeightedRandomSelection.java#L149

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/DefaultEnsemblePlacementPolicy.java
reddycharan: may be, but i just want to follow the existing code format here, like how it is dealt in else block

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/DefaultEnsemblePlacementPolicy.java
dlg99: what if all the bookies already in either in excludeBookies or in a newBookies?

reddycharan: there is check for this condition - https://github.com/apache/bookkeeper/pull/1788/files#diff-b8389e8cd72c81923a6c9aa82e119310L86

sijie:
run integration tests
rebuild java8

PR #1789, opened 2018-11-06
Cache InetSocketAddress if hostname is IPAddress

reddycharan:

- in BookieSocketAddress if IPAddress is hostname then it is okay
to cache InetSocketAddress, since the canonicalhostname of the node
dont change.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/net/BookieSocketAddress.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/net/BookieSocketAddress.java
@@ -22,6 +22,7 @@ package org.apache.bookkeeper.net;
 
 import static org.apache.bookkeeper.util.BookKeeperConstants.COLON;
 
+import com.google.common.net.InetAddresses;
 import io.netty.channel.local.LocalAddress;
 
 import java.net.InetSocketAddress;
@@ -38,11 +39,19 @@ public class BookieSocketAddress {
     // Member fields that make up this class.
     private final String hostname;
     private final int port;
+    private final boolean isHostnameIPAddress;
+    private final InetSocketAddress socketAddress;
 
     // Constructor that takes in both a port.
     public BookieSocketAddress(String hostname, int port) {
         this.hostname = hostname;
         this.port = port;
+        isHostnameIPAddress = InetAddresses.isInetAddress(hostname);
+        if (isHostnameIPAddress) {
+            socketAddress = new InetSocketAddress(hostname, port);
+        } else {
+            socketAddress = null;
+        }
     }
 
     // Constructor from a String "serialized" version of this class.
@@ -57,6 +66,12 @@ public class BookieSocketAddress {
         } catch (NumberFormatException nfe) {
             throw new UnknownHostException(addr);
         }
+        isHostnameIPAddress = InetAddresses.isInetAddress(hostname);
+        if (isHostnameIPAddress) {
+            socketAddress = new InetSocketAddress(hostname, port);
+        } else {
+            socketAddress = null;
+        }
     }
 
     // Public getters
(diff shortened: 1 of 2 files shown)


merlimat:
@reddycharan One other option could be to use `InetSocketAddress.createUnresolved()`, though I haven't tested if the cached gets updated when the name is finally resolved. If it doesn't, it might be a good solution too.

reddycharan:
@merlimat we cannt use InetSocketAddress.createUnresolved() here, because InetSocketAddress.getAddress would return null if it is unresolved https://docs.oracle.com/javase/7/docs/api/java/net/InetSocketAddress.html#getAddress()

in many places in our code we would use address of InetSocketAddress - 
https://github.com/apache/bookkeeper/blob/318893399b8040a7d291b8094f0d218a3b4e4f1e/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java#L578
https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/helpers/CommandHelpers.java#L44

reddycharan:
@eolivelli regarding testing I'm in the same boat as Matteo here, https://github.com/apache/bookkeeper/commit/3d8bad44a40ce5cb173df5dfd7cb06d81bed26e1

for ipv4/ipv6 i can confirm that InetAddresses.isInetAddress handles both ipv4/ipv6 - https://github.com/google/guava/blob/master/guava/src/com/google/common/net/InetAddresses.java#L186

bookkeeper-server/src/main/java/org/apache/bookkeeper/net/BookieSocketAddress.java
```diff
@@ -38,11 +39,19 @@
     // Member fields that make up this class.
     private final String hostname;
     private final int port;
+    private final boolean isHostnameIPAddress;
+    private final InetSocketAddress socketAddress;
```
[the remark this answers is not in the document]
reddycharan: fixed it

reddycharan:
@eolivelli added basic testcases validating cache functionality.

bookkeeper-server/src/main/java/org/apache/bookkeeper/net/BookieSocketAddress.java
```diff
@@ -38,11 +40,26 @@
     // Member fields that make up this class.
     private final String hostname;
     private final int port;
+    private final boolean isHostnameIPAddress;
```
merlimat: With “optional” the Boolean shouldn’t be needed anymore

reddycharan: removed boolean variable

reddycharan:
@eolivelli my bad. fixed it.

sijie:
run integration tests

sijie:
run integration tests

bookkeeper-server/src/main/java/org/apache/bookkeeper/net/BookieSocketAddress.java
```diff
@@ -70,12 +92,15 @@ public int getPort() {
...
-        // hostnames and the IP change, the BK client will keep forever to try to connect
-        // to the old IP.
-        return new InetSocketAddress(hostname, port);
+        /*
+         * Return each time a new instance of the InetSocketAddress if hostname
+         * is used as bookieid. If we keep using the same InetSocketAddress
+         * instance, if bookies are advertising hostnames and the IP change, the
+         * BK client will keep forever to try to connect to the old IP.
+         */
+        return socketAddress.orElseGet(() -> {
```
dlg99: nit:
```
        return socketAddress.orElseGet(() -> new InetSocketAddress(hostname, port));
```

reddycharan: multiline was suggestion from code formatter.

Issue #1791, opened 2018-11-06
Read Submission should bypass OSE Threads


nicmichael:
**BUG REPORT**

1. Please describe the issue you observed:

Profiling of our Bookkeeper Client code for read requests shows that client threads spend half of their time in dispatching requests to OrderedExecutors (just the dispatch itself, not the execution inside OSE): 54% of their CPU time is spent in OrderedExecutor.executeOrdered() (called by LedgerHandle.readEntriesInternalAsync()). The high time spend in request submission to OSE is largely caused by Linux scheduling cost, that is the cost of dispatching the OSE thread to CPU: 42% of total time (3/4th of executeOrdered() time), threads spend in Unsafe.unpark(), which is essentially Linux scheduling/dispatching of another thread.

It seems unnecessary for all reads to have to go through OrderedExecutors. Reads that do not require ordering (reads to read-only ledger handles?) should bypass OrderedExecutors, and read submission instead be executed directly in the context of the client thread without any context-switching to another thread (meaning, the client thread dispatches the read to Netty). This should both reduce overall CPU consumption of Bookkeeper client code, and reduce latency of reads by eliminating one context-switch and serialization/queuing of reads on OSEs.

Experiments with a prototype of reads to read-only ledger handles have shown significant improvements in both overall CPU consumption as well as read latency. The additional work client threads have to do (the dispatch of the read requests to netty) is roughly the same as the (saved) dispatch cost to OSE, so the change turns out to be neutral for CPU consumption of client threads. In some experiments, the savings even exceed the additional work, and client threads consume less cpu even though they "do more". It also frees up lots of resources in OSE threads. Since it eliminates one context-switch in read submission and also avoids serialization of reads to the same ledger (or ledgers hashing to the same OSE), it also reduces read latency. For a mixed read-write workload (14,000 reads/sec on read-only ledgers, 4,000 writes/sec on another set of ledgers), this change has reduced CPU consumption of OSE threads by 25%, kept CPU consumption of client (and Netty) threads the same, and yielded a 6% improvement of read latency (as measured by BK Client).

nicmichael:
![profiling_executeordered](https://user-images.githubusercontent.com/27270046/48089349-20c85780-e1b9-11e8-800f-4f63b97dfff9.png)

merlimat:
I think pegging requests to thread should avoid the mutex contention. 
[...]

PR #1792, opened 2018-11-06
Issue #1791: Read Submission should bypass OSE Threads

nicmichael:
### Motivation

Profiling of our Bookkeeper Client code for read requests shows that client threads spend half of their time in dispatching requests to OrderedExecutors (just the dispatch itself, not the execution inside OSE): 54% of their CPU time is spent in OrderedExecutor.executeOrdered() (called by LedgerHandle.readEntriesInternalAsync()). The high time spend in request submission to OSE is largely caused by Linux scheduling cost, that is the cost of dispatching the OSE thread to CPU: 42% of total time (3/4th of executeOrdered() time), threads spend in Unsafe.unpark(), which is essentially Linux scheduling/dispatching of another thread.

### Changes

This change executes read submissions (PendingReadOp) on read-only ledger handles directly inside the client thread instead of submitting them to Ordered Executors.

Tests with a prototype have shown significant improvements in both overall CPU consumption as well as read latency. The additional work client threads have to do (the dispatch of the read requests to netty) is roughly the same as the (saved) dispatch cost to OSE, so the change turns out to be neutral for CPU consumption of client threads. In some experiments, the savings even exceed the additional work, and client threads consume less cpu even though they "do more". It also frees up lots of resources in OSE threads. Since it eliminates one context-switch in read submission and also avoids serialization of reads to the same ledger (or ledgers hashing to the same OSE), it also reduces read latency. For a mixed read-write workload (14,000 reads/sec on read-only ledgers, 4,000 writes/sec on another set of ledgers), this change has reduced CPU consumption of OSE threads by 25%, kept CPU consumption of client (and Netty) threads the same, and yielded a 6% improvement of read latency (as measured by BK Client).

Master Issue: #1791: Read Submission should bypass OSE Threads

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
@@ -840,7 +840,14 @@ public class LedgerHandle implements WriteHandle {
                 ws.recycle();
             }
 
-            clientCtx.getMainWorkerPool().executeOrdered(ledgerId, op);
+            if (isHandleWritable()) {
+                // Ledger handle in read/write mode: submit to OSE for ordered execution.
+                clientCtx.getMainWorkerPool().executeOrdered(ledgerId, op);
+            } else {
+                // Read-only ledger handle: bypass OSE and execute read directly in client thread.
+                // This avoids a context-switch to OSE thread and thus reduces latency.
+                op.run();
+            }
         } else {
             op.future().completeExceptionally(BKException.create(ClientClosedException));
         }


sijie:
@eolivelli 

> Can we add some test case?

don't think we need test cases. read op is a common enough operation, which should be covered by all existing test cases.

> I am not sure we can able this by default,

I think the change is enabling only on readonly handles. that seems to be fine.

nicmichael:
@eolivelli I looked at the existing test cases. As sijie pointed out, there's quite a few existing ones that test reads on both read-only as well as read-write ledger handles. I hope that would be sufficient. Or were you thinking of any kind of special tests?

eolivelli:
I was talking about adding a test about the fact that if the handlr is read-only the operation is not submitted to the OSE, otherwise it uses the OSE.

Not blocker

eolivelli:
No need for a flag

sijie:
@merlimat please review this since you have comment at #1791

lhotari:
In Pulsar, https://github.com/apache/pulsar/issues/14436 seems to be caused by a thread safety issue in Bookkeeper client. It looks like it is caused by the changes in this PR. 

The reason why there's a thread safety issue is that a network request-response call doesn't ensure happens-before. Happens-before visibility guarantee is needed since there's no synchronization in the PendingReadOp handling.

lhotari:
BookKeeper uses LinkedBlockingQueue for executors by default which isn't very efficient. 

https://github.com/apache/bookkeeper/blob/3db4de9da1447e2a5135ac9330d4347b9d220a0d/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/OrderedExecutor.java#L308

A better Queue implementation would be something that Jetty uses for it's thread pools https://github.com/eclipse/jetty.project/blob/jetty-10.0.x/jetty-util/src/main/java/org/eclipse/jetty/util/BlockingArrayQueue.java .

By improving the queue implementation, there would be better performance without breaking thread safety.

lhotari:
I created #3104 about the thread safety issue.

lhotari:
#3104 turned out to be a clear state handling issue in PendingReadOp class (and included embedded classes), fix is #3110

PR #1797, opened 2018-11-08
Move version out of LedgerMetadata

ivankelly:
The version is a property of the metadata store and not of the
LedgerMetadata object itself. Putting it in the LedgerMetadata forces
that LedgerMetadata to be mutable, as it needs to be updated before
being sent to the LedgerManager.

This change moves version out of LedgerMetadata and modifies the
LedgerManager to use Versioned<LedgerMetadata>.

This is a large change, but almost purely mechanical.

TestHttpService has been modified to remove a preexisting flake in a
test which had to be modified for this change.

Master issue: #281


ivankelly:
I didn't do it in this change, but I want to change LedgerManager to use CompletableFutures instead of callbacks, because everywhere that uses it is converting to Futures anyhow.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -190,7 +190,9 @@
         // password, so that the bookie can avoid processing the keys for each entry
         this.ledgerKey = DigestManager.generateMasterKey(password);
         distributionSchedule = new RoundRobinDistributionSchedule(
-                metadata.getWriteQuorumSize(), metadata.getAckQuorumSize(), metadata.getEnsembleSize());
+                metadata.getValue().getWriteQuorumSize(),
```
eolivelli: Nit: call getValue() only once?

ivankelly:
rerun bookie replication tests

ivankelly:

ivankelly:
rerun pr validation

PR #1804, opened 2018-11-10
[client] read submission for readonly handle should bypass OSE threads.

sijie:

*Motivation*

apache/bookkeeper#1791 improves performance by bypassing OSE threads for read request submission.
however the change doesn't handle `tailing` ReadOnlyLedgerHandle

*Changes*

ReadOnlyLedgerHandle overrides `isHandleWritable` and return `false` always.

> ---
>
> ---

> ---
> quickly and easily:
>
>
>
> Otherwise:
> 
> 
> ---

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
@@ -471,7 +471,12 @@ public class LedgerHandle implements WriteHandle {
         return getLedgerMetadata().isClosed();
     }
 
-    boolean isHandleWritable() {
+    /**
+     * Return if this handle is writable or not.
+     *
+     * @return true if this handle is writable, otherwise false
+     */
+    protected boolean isHandleWritable() {
         return !getLedgerMetadata().isClosed() && handleState == HandleState.OPEN;
     }
 
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadOnlyLedgerHandle.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadOnlyLedgerHandle.java
@@ -109,7 +109,7 @@ class ReadOnlyLedgerHandle extends LedgerHandle implements LedgerMetadataListene
     }
 
     @Override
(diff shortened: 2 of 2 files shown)


sijie:
@nicmichael can you please review this PR to see if that matches your intention at #1791 ?

eolivelli:
In client tests there are a lot of failure around fencing.
This is quite scary

sijie:
retest this please

eolivelli:
retest this please

eolivelli:
there are still a lot of errors

nicmichael:
@sijie Yes that was the intention. 
I'm looking at the test failures now. Adding "|| isRecoveryRead" to isHandleWritable() in readEntriesInternalAsync seems to address some, but not all of the failures. I don't quite understand yet what's causing the failures.

eolivelli:
maybe the errors on recovery are due to the fact that  reads should strictly respect the order

nicmichael:
During recovery I suppose they should...
In the absence of a recovery on a read-only LH, they don't need to, and getting rid of the ordering (and the context-switch) is a noticeable performance win. So the trick is to correctly distinguish these scenarios. Guess that's still not right.

sijie:
yeah. I think we don't this change. we should just close it.

nicmichael:
I *think* what we need is something like this:

            if (isHandleWritable() || metadata.isInRecovery()) { 
                // Ledger handle in read/write mode: submit to OSE for ordered execution.
                clientCtx.getMainWorkerPool().executeOrdered(ledgerId, op);
            } else {
                // Read-only ledger handle: bypass OSE and execute read directly in client thread.
                // This avoids a context-switch to OSE thread and thus reduces latency.
                op.run();
            }

It passes tests on my workstation. I'm just not quite sure whether this check needs any synchronization or can go without.

sijie:
@nicmichael I think it is a bit risky to take `recovery` state into account. so I would rather rely on whether metadata is closed or not. that says, the check in current master is probably good enough for the cases we are targeting.

nicmichael:
Sounds good to me. The most common case is probably addressed by the earlier change.

eolivelli:
Okay, we don't need this.

PR #1805, opened 2018-11-11
Fixed Auth with v2 protocol

merlimat:
### Motivation

BK auth framework is currently broken when using v2 protocol.

### Changes

 * Fixed auth when using V2 protocol
 * Made sure a client with authentication enabled can talk to a bookie without authentication. This is required in any case when enabling/disabling authentication on a live cluster.
 * Run all auth tests against both v2 and v3 protocol.

This should be included in 4.7.2 to give a path to upgrade.

cc/ @rdhabalia


bookkeeper-server/src/test/java/org/apache/bookkeeper/auth/TestAuthMixed.java
```diff
@@ -0,0 +1,150 @@
...
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameters;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Test authentication when enabled in client and disabled in bookie.
+ */
+@RunWith(Parameterized.class)
+public class TestAuthMixed extends BookKeeperClusterTestCase {
```
eolivelli: Why is test can't be with the other ones? (TestAuth).

bookkeeper-server/src/test/java/org/apache/bookkeeper/auth/TestAuthMixed.java
```diff
@@ -0,0 +1,150 @@
...
+
+    private final ProtocolVersion protocolVersion;
+
+    public TestAuthMixed(ProtocolVersion protocolVersion) {
+        super(1);
+        this.protocolVersion = protocolVersion;
+    }
+
+    /**
+     * Test with a client that sends auth and bookie that has no auth provider.
```
eolivelli: This case should have been already covered in the other class, which is the difference ?

merlimat: Oh, I didn't see that. I started investigating this condition and that's why I wrote the test. No need to keep it, since it's the same test.

merlimat:
> The very interesting fact is that when I worked on auth only Yahoo had auth in place and they used v2 :)

@eolivelli I think this got broken (in v2) after the changes to remove the protobuf extension mechanism for the auth message. Core problem is the test was not running with v2.

sijie:
merged to master, branch-4.8, and branch 4.7

PR #1807, opened 2018-11-12
[bookie] Fix sorted ledger storage rotating entry log files too frequent

sijie:

*Motivation*

A strong behavior was observed when using sorted ledger storage with single entry log manager on production:

"the entry log files are rotated very frequently and small entry log files are produced".

The problem was introduced due to #1410.

At current entry logger, when a new entry log file is created, EntryLogger will notify its listeners
that a new entry log file is rotated via [`onRotateEntryLog`](https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogManagerBase.java#L152).

Before the change in #1410, `SortedLedgerStorage` inherits from `InterleavedLedgerStorage`.
So when a new entry log file is rotated, `SortedLedgerStorage` is notified.

However after the change in #1410, `SortedLedgerStorage` doesn't inherits `InterleavedLedgerStorage` anymore.
Instead, the relationship is changed to composition. `SortedLedgerStorage` is composed using an interleaved ledger
storage. So the entrylog listener contract was broken. `SortedLedgerStorage` will not receive any `onRotateEntryLog`
notification any more.

*Changes*

When `SortedLedgerStorage` initializes, it passes its own entry log listener down to the interleaved ledger storage.
So entry logger can notify the right person for entry log rotations.

*Tests*

Existing tests should cover most of the case. Looking for how to add new test cases.

> ---
>
> ---

> ---
> quickly and easily:
>
>
>
> Otherwise:
> 
> 
> ---

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogManagerForEntryLogPerLedger.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogManagerForEntryLogPerLedger.java
@@ -642,7 +642,7 @@ class EntryLogManagerForEntryLogPerLedger extends EntryLogManagerBase {
                 }
                 createNewLog(ledgerId,
                     ": diskFull = " + diskFull + ", allDisksFull = " + allDisksFull
-                        + ", reachEntryLogLimit = " + reachEntryLogLimit);
+                        + ", reachEntryLogLimit = " + reachEntryLogLimit + ", logChannel = " + logChannel);
             }
 
             return getCurrentLogForLedger(ledgerId);
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogManagerForEntryLogPerLedger.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogManagerForEntryLogPerLedger.java
@@ -484,7 +484,7 @@ class EntryLogManagerForEntryLogPerLedger extends EntryLogManagerBase {
                 try {
                     if (reachEntryLogLimit(currentLog, 0L)) {
                         log.info("Rolling entry logger since it reached size limitation for ledger: {}", ledgerId);
-                        createNewLog(ledgerId);
+                        createNewLog(ledgerId, "after entry log file is rotated");
                     }
                 } finally {
                     lock.unlock();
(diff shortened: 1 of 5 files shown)


sijie:
(not sure if that's the best fix or not. any suggestions are welcome)

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/InterleavedLedgerStorage.java
```diff
@@ -98,11 +98,32 @@ public void initialize(ServerConfiguration conf,
...
+            ledgerDirsManager,
+            indexDirsManager,
+            stateManager,
+            checkpointSource,
+            checkpointer,
+            this,
+            statsLogger);
+    }
+
+    void initializeWithEntryLogListener(ServerConfiguration conf,
```
reddycharan: instead of adding a new initialize method, you should consider changing the signature of the existing initialize method, which takes EntryLogListener argument. In ILS initialize method it will check if it receives the listener, if it is not null use it otherwise use ‘this’.

reddycharan:
@sijie in this change can you add log files in the stackTrace of createNewLog call hierarchy. As we observed in the log file, it was not evident what triggered the creation of new entrylog files. Since createNewLog is infrequent thing, it should be ok to add logs as much as we want.

sijie:
@reddycharan I added a logging message with a reason field. let me know if this approach works for you.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogManagerForEntryLogPerLedger.java
```diff
@@ -640,7 +640,9 @@ BufferedLogChannel getCurrentLogForLedgerForAddEntry(long ledgerId, int entrySiz
                 if (logChannel != null) {
                     logChannel.flushAndForceWriteIfRegularFlush(false);
                 }
-                createNewLog(ledgerId);
+                createNewLog(ledgerId,
+                    ": diskFull = " + diskFull + ", allDisksFull = " + allDisksFull
+                        + ", reachEntryLogLimit = " + reachEntryLogLimit);
```
reddycharan: we can get to this block even if logChannel is null. so add that aswell to the reason string.

reddycharan:
LGTM other than that minor log statement issue - https://github.com/apache/bookkeeper/pull/1807/files#r233123238

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogManagerForEntryLogPerLedger.java
sijie: add logChannel in the reason string.

sijie:
run integration tests

sijie:
IGNORE IT CI

PR #1808, opened 2018-11-13
Allow to configure sticky reads

merlimat:
### Motivation

Currently the BK client is issuing the read requests in round-robin fashion across all the bookies in the write set.

One issue with this approach is that it's not taking full advantage of the read-ahead cache, either explicit (like in `DbLedgerStorage`) or implicit (by reading data through Linux page cache which will do some prefetching).

With `e=2`, `w=2`, when we read `e-0` from `bookie-1` and `e-1` from `bookie-2`, we fail to take advantage of the fact that `bookie-1` will have already `e-1` in memory.

Effectively with `e-2`, `w-2` the disk read IO will be doubled, compared to the amount of data served to BK clients. The larger the quorum, the bigger will be overhead (eg: `e=5`, `w=5` will lead to 5x reads from disk).

### Changes

Added a BK client flag for "sticky reads". When reading from a ledger that has `E=W` (every bookie has all the entries), the sticky reads will direct all read request to 1 single bookie in the ensemble.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
@@ -126,10 +126,10 @@ public class LedgerHandle implements WriteHandle {
      * selected as "sticky" for this ledger. The bookie is chosen at random when the
      * LedgerHandle is created.
      *
-     * In case of failures, the bookie index will be updated (to the next bookie in
+     * <p>In case of failures, the bookie index will be updated (to the next bookie in
      * the ensemble) to avoid continuing to attempt to read from a failed bookie.
      *
-     * If the index is -1, it means the sticky reads are disabled.
+     * <p>If the index is -1, it means the sticky reads are disabled.
      */
     private int stickyBookieIndex;
 
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
@@ -42,13 +42,13 @@ import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Queue;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.ScheduledFuture;
-import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -196,7 +196,7 @@ public class LedgerHandle implements WriteHandle {
 
         if (clientCtx.getConf().enableStickyReads
                 && getLedgerMetadata().getEnsembleSize() == getLedgerMetadata().getWriteQuorumSize()) {
-            stickyBookieIndex = ThreadLocalRandom.current().nextInt() % getLedgerMetadata().getEnsembleSize();
+            stickyBookieIndex = clientCtx.getPlacementPolicy().getStickyReadBookieIndex(metadata, Optional.empty());
         } else {
             stickyBookieIndex = -1;
         }
@@ -267,7 +267,8 @@ public class LedgerHandle implements WriteHandle {
         if (stickyBookieIndex != -1) {
             // This will be idempotent when we have multiple read errors on the
             // same bookie. The net result is that we just go to the next bookie
-            stickyBookieIndex = (bookieIndex + 1) % getLedgerMetadata().getEnsembleSize();
+            stickyBookieIndex = clientCtx.getPlacementPolicy().getStickyReadBookieIndex(getLedgerMetadata(),
+                    Optional.of(bookieIndex));
         }
     }
 
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
@@ -48,6 +48,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -58,6 +59,7 @@ import org.apache.bookkeeper.client.AsyncCallback.ReadCallback;
 import org.apache.bookkeeper.client.AsyncCallback.ReadLastConfirmedCallback;
 import org.apache.bookkeeper.client.BKException.BKIncorrectParameterException;
 import org.apache.bookkeeper.client.BKException.BKReadException;
+import org.apache.bookkeeper.client.DistributionSchedule.WriteSet;
 import org.apache.bookkeeper.client.SyncCallbackUtils.FutureReadLastConfirmed;
 import org.apache.bookkeeper.client.SyncCallbackUtils.FutureReadLastConfirmedAndEntry;
 import org.apache.bookkeeper.client.SyncCallbackUtils.SyncAddCallback;
@@ -118,6 +120,18 @@ public class LedgerHandle implements WriteHandle {
       * in order to have an ordered sequence of addEntry ackknowledged to the writer
       */
     volatile long pendingAddsSequenceHead;
+
+    /**
+     * If bookie sticky reads are enabled, this will contain the index of the bookie
+     * selected as "sticky" for this ledger. The bookie is chosen at random when the
+     * LedgerHandle is created.
+     *
+     * In case of failures, the bookie index will be updated (to the next bookie in
+     * the ensemble) to avoid continuing to attempt to read from a failed bookie.
+     *
+     * If the index is -1, it means the sticky reads are disabled.
+     */
+    private int stickyBookieIndex;
 
     long length;
     final DigestManager macManager;
@@ -180,6 +194,13 @@ public class LedgerHandle implements WriteHandle {
 
         this.ledgerId = ledgerId;
 
+        if (clientCtx.getConf().enableStickyReads
+                && getLedgerMetadata().getEnsembleSize() == getLedgerMetadata().getWriteQuorumSize()) {
+            stickyBookieIndex = ThreadLocalRandom.current().nextInt() % getLedgerMetadata().getEnsembleSize();
+        } else {
+            stickyBookieIndex = -1;
+        }
+
         if (clientCtx.getConf().throttleValue > 0) {
             this.throttler = RateLimiter.create(clientCtx.getConf().throttleValue);
         } else {
@@ -234,6 +255,20 @@ public class LedgerHandle implements WriteHandle {
             });
 
         initializeWriteHandleState();
+    }
+
+    /**
+     * Notify the LedgerHandle that a read operation was failed on a particular bookie.
+     */
+    void recordReadErrorOnBookie(int bookieIndex) {
+        // If sticky bookie reads are enabled, switch the sticky bookie to the
+        // next bookie in the ensemble so that we avoid to keep reading from the
+        // same failed bookie
+        if (stickyBookieIndex != -1) {
+            // This will be idempotent when we have multiple read errors on the
+            // same bookie. The net result is that we just go to the next bookie
+            stickyBookieIndex = (bookieIndex + 1) % getLedgerMetadata().getEnsembleSize();
+        }
     }
 
     protected void initializeWriteHandleState() {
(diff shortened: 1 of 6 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
```diff
@@ -96,13 +96,38 @@
...
+                // e-1   X     X
+                // e-2   X     X
+                //
+                // In this case we want all the requests to be issued to B-1 (by
+                // preference), so that cache hits will be maximized.
+                //
+                // We can only enable sticky reads if the ensemble==writeQuorum
+                // otherwise the same bookie will not have all the entries
+                // stored
+                entryIdToConsiderForWriteSet = 0;
```
eolivelli: How do you deal with ensemble changes?

eolivelli:
We also need tests.

Is this a WIP ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
sijie: hmm I don't think it is a good idea to sticky to ensemble 0. if the first bookie of ensemble 0 went down, all the reads will have huge penalties. I think "sitckiness" here should be at ledger handle level, when a ledger handle is constructed, one bookie is chosen as the sticky bookie, most of the requests would be sent to this sticky bookie. until failure occurs, a new sticky bookie would be choose. that is being said, ledger handle should probably implement a `Supplier<WriteSet>` which supplies the `WriteSet` for pending read ops to read. 

so the pending read ops only talks to this supplier to get write set to read entries.

- for normal settings, the supplier will return `lh.distributionSchedule.getWriteSet(entryId)`.
- if reorder is enabled, the supplier will return reordered write set.
- if sticky read is enabled, the supplier will return a fix writeset most of the time, if errors occurs, the supplier will update the write set, so the subsequent reads will choose a different sticky bookie to read.

sijie: the change is to pick a "sticky" bookie. the idea here is to pick a fixed "write set" (a certain  read order). the write set is a sequence of bookie index not fixed bookie addresses, so there is nothing to worry about about "ensemble changes".

jvrao:
@nicmichael  can you please look into this too?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
jvrao: I believe this change needs to be more elaborate. Sticking to a fixed set is a good idea, but it should be driven with a feedback loop rather than strictly fixing. The bookie can be down, or overloaded, or it could be in a different zone in public cloud etc.

eolivelli: Thinking out loud
Maybe the PlacementPolicy should decide about which bookie is to be used for the 'sticky read'

sijie: > Maybe the PlacementPolicy should decide about which bookie is to be used for the 'sticky read'

PlacementPolicy should be used for choosing bookies for locality. but it doesn't have to distinguish sticky reads or normal reads. The "stickiness" should not be as part of placement policy, "stickiness" should be part of ledger handle.

eolivelli: Okay

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ClientConfiguration.java
```diff
@@ -1135,6 +1136,34 @@ public ClientConfiguration setReorderReadSequenceEnabled(boolean enabled) {
         return this;
     }
 
+    /**
+     * If read operation should be sticky to a single bookie or not.
+     *
+     * @return true if reorder read sequence is enabled, otherwise false.
+     */
+    public boolean isStickyReadsEnabled() {
```
dlg99: can it be a per read request flag rather than per client?
i.e. we have various cases, in some cases reads are random and sticky reads probably won't help.
In other cases reads are strictly sequential and sticky read may help. "may" is because we can hit bottleneck on bookie NIC side and spreading reads across multiple bookies might actually help more then hurt.
Having two clients for these cases would mean 2x connections to bookies, executors/threads etc.

sijie: you mean per ledger handle flag, not per read request flag, right?

merlimat: > can it be a per read request flag rather than per client?

@dlg99 It could work on per-request as long as the mechanism works across multiple requests. Eg: `read(0, 99)` and `read(100, 199)`, all the 200 requests should go to same bookie.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
```diff
@@ -96,13 +96,38 @@
             this.ensemble = ensemble;
             this.eId = eId;
 
+            long entryIdToConsiderForWriteSet = eId;
+
+            if (clientCtx.getConf().enableStickyReads
+                    && lh.getLedgerMetadata().getWriteQuorumSize() == lh.getLedgerMetadata().getEnsembleSize()) {
```
dlg99: WQ == ES is quite limiting requirement. I.e. we run with ES = 7 and WQ = 3.
I'd suggest picking bookie with ensemble.id % WQ == 0.
i.e in case of 7 bookies in the ensemble and WQ 3, regular round-robin placement:

entry 0 is on bookies 0,1,2 (from ensemble) - reads stick to 0 (0 % 3 == 0)
entry 1 is on bookies 1,2,3 - reads stick to 3
entry 2 on 2,3,4 - reads sticks to 3
entry 3 on 3,4,5 - stick to 3
entry 4 on 4,5,6 - stick to 6
entry 5 on 5,6,7 - stick to 6
entry 6 on 6,7,0 - stick to 6 (choice of 6 and 0, got to the first one in order)
entry 7 on 7,0,1 - stick to 0.
etc

with WQ == ES it will stick to 0 all the time.

Plus it has to support read reordering (as I understand, it did not change) - don't try to use bookie if its netty channel is not writable.

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ClientConfiguration.java
dlg99: > read(0, 99) and read(100, 199), all the 200 requests should go to same bookie.

won't happen for us, ES > WQ. but with https://github.com/apache/bookkeeper/pull/1808/files#r233679775 in mind that should not make any difference.
also we don't use read(0, 99) option. it happened so that we issue read(entryId) requests individually.

dlg99:
do you have any perf numbers to share to show what kind of perf improvements we should expect with this feature enabled?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
merlimat: I thought of it but I think the IO reads saving will be just a fraction compared to E=W case. 

Eg, in your example, Bookie-0 will read e-0 and prefetch e-5 and e-6 though these 2 entries will be read on bookie 6

merlimat:
> do you have any perf numbers to share to show what kind of perf improvements we should expect with this feature enabled?

I haven't completed full-scale tests yet. Though I was able to measure the fact that read-ahead cache hit-miss rate was being driven down by the fact that the prefetching was happening on all 3 the bookies for the same entries, driving the IO reads to 3x of what it should be.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
dlg99: but OS page cache will load more than e5 and e6 so all and all it will help.
I mean, it is speculations without perf tests anyways.

merlimat: Sure, but reading of e-5 and e-6 will be anyway "wasted".

eolivelli:
We could have a ReadFlag instead of a global client wise config parameter.

The discussion seems broader than the expected, should we start a BP? Or any way discuss on an 'issue' and then go back to 'code'

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
dlg99: From strictly theoretical POV:
It works the same as picking bookie 0 for ES == WQ case.
For ES > WQ case it pins reads to limited subset of bookies and yes, something will get wasted but what else we can do other than round robin? 
in a full sequential read of ledger will this waste more or less than not handling ES > WQ case?

From practical POV, even with ES=WQ something will get wasted.
There is no guarantee that entries strictly sequential in the file. 1st, they can interleave with other ledgers.
Even with entry log per ledger *and* sorted ledger we can end up with out of order writes and out of order entries in entry log.

also, when you say prefetched do you mean OS file system cache or bufferedchannel in the bookie?

sijie:
> The discussion seems broader than the expected, should we start a BP? Or any way discuss on an 'issue' and then go back to 'code'

@eolivelli there is already an implementation in progress (it is working although not perfect). so I don't think it is necessarily to this heavy-lifting bp discussion loop. Lets stick the discussion here and get the ball move.

nicmichael:
I actually like @dlg99's suggestion: if each W adjacent entries have the same primary bookie, but the next W entries another primary bookie, we'd get W times better data locality but still evenly distribute reads across all bookies in E. Plus it would work for E>W. Ideally E and W are prime (e.g. E=3, W=7), then the primary bookie would be easy to calculate and guaranteed to be evenly distributed.

If even distribution is guaranteed, this policy wouldn't even need a flag to turn it on/off. It could be the default policy as it wouldn't harm random reads (due to even distribution across all bookies in E) but benefit sequential reads (at least in the absence of out-of-order writes and entry logs per ledger...).

The benefits of read-ahead anyways depend on the size of the entries and file system configuration. Unless entries are tiny, a policy like this that benefits from read-ahead of W entries might bring most of the performance gain (and a stickiness of the entire ledger to just one bookie might be marginally better for sequential reads).

I suppose reordering of reads (e.g. due to slow bookies or number of outstanding requests to a bookie, Issue #1489) would still kick in if enabled, and rearrange the order if a primary bookie is found to be slow?

merlimat:
@eolivelli @jvrao  @dlg99 @sijie @ivankelly I have updated the PR with the following: 

 * When enabled, `LedgerHandle` has the notion of a sticky bookie
 * The sticky bookie is initially selected at random (this helps when there are multiple readers that are independently reading different portions of the ledger concurrently).
 * If there are read failures, the ledger handle will pick a new sticky bookie, to avoid read impact.
 * Added tests

I haven't addressed the case where E > W because I believe that the benefit in that scenario are not that big, compared to the case where E=W. It would be easy to add that afterward in any case.

Please take a look.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -180,6 +194,13 @@
 
         this.ledgerId = ledgerId;
 
+        if (clientCtx.getConf().enableStickyReads
+                && getLedgerMetadata().getEnsembleSize() == getLedgerMetadata().getWriteQuorumSize()) {
+            stickyBookieIndex = ThreadLocalRandom.current().nextInt() % getLedgerMetadata().getEnsembleSize();
```
sijie: it might be better to add a method in `EnsemblePlacementPolicy`

```
int getStickyReadBookieIndex(LedgerMetadata metadata);
```

it can pick up a bookie randomly by default. for rack-aware and region-aware placement, it can be implemented using network locations.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -1972,4 +2007,33 @@ public void closeComplete(int rc, LedgerHandle lh, Object ctx) {
...
+            // e-1 X X
+            // e-2 X X
+            //
+            // In this case we want all the requests to be issued to B-1 (by
+            // preference), so that cache hits will be maximized.
+            //
+            // We can only enable sticky reads if the ensemble==writeQuorum
+            // otherwise the same bookie will not have all the entries
+            // stored
+            return distributionSchedule.getWriteSet(stickyBookieIndex);
```
eolivelli: Is this an entryId or a reference to a bookie.
I am a bit confused

sijie:
@merlimat are we including this for 4.9.0?

merlimat:
@sijie Added the `getStickyReadBookieIndex()` in `EnsemblePlacementPolicy` so it can be overridden. I haven't done the region-aware selection since I'm not too familiar on that part.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
merlimat: The write set is typically dependent on the entry id. If a `stickyBookieIndex` was set, it means we're try to override that distribution and instead send all reads to a particular bookie. `Bookie-i ` will be the "master" bookie for the `entry-i`. Therefore, a write set with `bookie-i` will first try to read from `bookie-i`, falling back to other bookies in case of failure.

merlimat: @eolivelli Does this answer the question?

eolivelli: Okay. It works.
Sorry for late reply

PR #1809, opened 2018-11-13
Change LedgerManager to use CompletableFuture

ivankelly:
All new uses of LedgerManager are creating GenericCallbackFutures to
pass in as the callback. This patch cuts out the middleman, and makes
LedgerManager return CompletableFutures itself.

The patch only touches the CRUD operations, and not
LedgerManager#asyncProcessLedgers, which I may tackle in a later PR.

Master issue: #281

### DIFF
--- a/tests/backward-compat/recovery-no-password/src/test/groovy/org/apache/bookkeeper/tests/backwardcompat/TestCompatRecoveryNoPassword.groovy
+++ b/tests/backward-compat/recovery-no-password/src/test/groovy/org/apache/bookkeeper/tests/backwardcompat/TestCompatRecoveryNoPassword.groovy
@@ -59,16 +59,7 @@ class TestCompatRecoveryNoPassword {
     DockerClient docker
 
     private LedgerMetadata getLedgerMetadata(BookKeeper bookkeeper, long ledgerId) throws Exception {
-        CompletableFuture<Versioned<LedgerMetadata>> future = new CompletableFuture<>()
-        bookkeeper.getLedgerManager().readLedgerMetadata(
-            ledgerId, { rc, result ->
-                if (rc != BKException.Code.OK) {
-                    future.completeExceptionally(BKException.create(rc))
-                } else {
-                    future.complete(result)
-                }
-            })
-        return future.get().getValue()
+        return bookkeeper.getLedgerManager().readLedgerMetadata(ledgerId).get().getValue()
     }
 
     private static class ReplicationVerificationCallback implements ReadEntryCallback {


jvrao:
@reddycharan Please review this.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BKException.java
```diff
@@ -439,12 +439,22 @@ public BKLedgerIdOverflowException() {
      * Extract an exception code from an BKException, or use a default if it's another type.
      */
     public static int getExceptionCode(Throwable t, int defaultCode) {
-        if (t instanceof BKException) {
+        if (t == null) {
+            return BKException.Code.OK;
+        } else if (t instanceof BKException) {
             return ((BKException) t).getCode();
         } else if (t.getCause() != null) {
```
eolivelli: not directly related, but I have started working with CompletableFuture in other projects, and I see that dependency on race conditions sometimes CompletableFuture notify CompletionException which wrap the actual exception.
Did you ever seen such problems in BK ?

is this branch addressing such issue ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
```diff
@@ -1024,41 +1022,45 @@ public int runCmd(CommandLine cmdLine) throws Exception {
...
-                                };
-                                ledgerManager.readLedgerMetadata(ledgerId, gencb);
+                            @Override
+                            public void process(Long ledgerId, VoidCallback cb) {
+                                if (!printMeta && (bookieAddress == null)) {
+                                    printLedgerMetadata(ledgerId, null, false);
+                                    cb.processResult(BKException.Code.OK, null, null);
+                                } else {
+                                    ledgerManager.readLedgerMetadata(ledgerId).whenComplete(
+                                            (metadata, exception) -> {
```
sijie: I think the exception would be `CompletionException` here. So I am not sure how `if (exception instanceof BKException.BKNoSuchLedgerExistsException)` will  be triggered. (unless I misunderstand `whenComplete`)

see: https://github.com/apache/bookkeeper/blob/master/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/concurrent/FutureEventListener.java#L35

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BKException.java
```diff
@@ -439,12 +439,22 @@ public BKLedgerIdOverflowException() {
      * Extract an exception code from an BKException, or use a default if it's another type.
      */
     public static int getExceptionCode(Throwable t, int defaultCode) {
-        if (t instanceof BKException) {
+        if (t == null) {
```
sijie: can you explain why we are adding this? and please add a comment here

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperAdmin.java
```diff
@@ -491,26 +489,24 @@ public void asyncGetLedgersContainBookies(final Set<BookieSocketAddress> bookies
...
-                            return;
-                        } else if (BKException.Code.OK != rc) {
-                            cb.processResult(rc, null, null);
-                            return;
-                        }
-                        Set<BookieSocketAddress> bookiesInLedger = metadata.getValue().getBookiesInThisLedger();
-                        Sets.SetView<BookieSocketAddress> intersection =
+                bkc.getLedgerManager().readLedgerMetadata(lid)
+                    .whenComplete((metadata, exception) -> {
+                            if (exception instanceof BKException.BKNoSuchLedgerExistsException) {
```
sijie: isn't exception here a `CompletionException`?

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractZkLedgerManager.java
```diff
@@ -100,34 +102,31 @@ public void run() {
...
                                 for (LedgerMetadataListener listener : listenerSet) {
                                     listener.onChanged(ledgerId, result);
                                 }
                             }
-                        }
-                    });
+                        });
                 }
-            } else if (BKException.Code.NoSuchLedgerExistsException == rc) {
+            } else if (exception instanceof BKException.BKNoSuchLedgerExistsException) {
```
sijie: I will suggest not using `instanceof` here. We should just use `getExceptionCode(exception) == BKException.Code.NoSuchLedgerExistsException`

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractZkLedgerManager.java
```diff
@@ -144,7 +143,7 @@ public void run() {
                     }
                 }
             } else {
-                LOG.warn("Failed on read ledger metadata of ledger {} : {}", ledgerId, rc);
+                LOG.warn("Failed on read ledger metadata of ledger {}", ledgerId, exception);
```
sijie: I don't think the exception stack trace is useful here (in async callbacks). so I think we should still use `rc` here.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/CleanupLedgerManager.java
```diff
@@ -75,11 +76,17 @@ public LedgerRange next() throws IOException {
         }
     }
 
+    private static final CompletableFuture<Versioned<LedgerMetadata>> closedPromise = new CompletableFuture<>();
```
sijie: don't do this.  you can have a static method creating the future, but never keep it as a static variable.
every future should be a brand new instance, otherwise, all the callbacks will be registered here.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/CleanupLedgerManager.java
```diff
@@ -108,64 +115,69 @@ private GenericCallback removeCallback(GenericCallback callback) {
         return callbacks.remove(callback);
     }
 
+    private void recordPromise(CompletableFuture<?> promise) {
```
sijie: I am not sure we need this or not. The underlying LM should error about its complete futures, so all the complete futures at this class will be errored out correctly. The change here is masking any potentially errors in underlying LM. I would suggest not adding such logic for now. or at least don't do it in a refactor PR.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/MSLedgerManagerFactory.java
```diff
@@ -238,34 +239,31 @@ public void run() {
...
-                                        listener.onChanged(ledgerId, result);
+                                        listener.onChanged(ledgerId, metadata);
                                     }
                                 }
-                            }
-                        });
+                            });
                     }
-                } else if (BKException.Code.NoSuchLedgerExistsException == rc) {
+                } else if (exception instanceof BKException.BKNoSuchLedgerExistsException) {
```
sijie: use `getExceptionCode` to check exception code.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/MSLedgerManagerFactory.java
```diff
@@ -275,7 +273,7 @@ public void run() {
                         }
                     }
                 } else {
-                    LOG.warn("Failed on read ledger metadata of ledger {} : {}", ledgerId, rc);
+                    LOG.warn("Failed on read ledger metadata of ledger {}", ledgerId, exception);
```
sijie: the exception stack trace is not useful.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractZkLedgerManager.java
ivankelly: Ya, I'm going to change any place I have instanceof to getExceptionCode. The CompletionException stuff is too unclear.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/CleanupLedgerManager.java
ivankelly: https://github.com/bpupadhyaya/openjdk-8/blob/45af329463a45955ea2759b89cb0ebfe40570c3f/jdk/src/share/classes/java/util/concurrent/CompletableFuture.java#L2015

If the future is complete, nothing is queued. It's executed straight away. I've no problem creating a future each time, this isn't fast path code, but it is safe to use a future like this.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/CleanupLedgerManager.java
ivankelly: Are you saying the underlying LM should keep track of all futures they return? 
This change is to basically do the same thing as the CleanupLedgerManager did for callbacks. i.e. If close is called, all outstanding futures should be closed, and all future calls also closed. Moving this into the underlying LM would basically mean making an AbstractLM with this recordPromise and pulling that into the other LMs.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/CleanupLedgerManager.java
sijie: Yeah. I would prefer creating a future each time, that would be a clean path, and will be no surprise from gc perspective.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/CleanupLedgerManager.java
sijie: No, what I mean is underlying implementation will probably already handle that. we don't need to worry about that at this layer.

it is a good feature to have at any case. but let's do it in a separate PR rather than doing it in a large refactor.

ivankelly: the underlying implementations don't handle it though. for them to handle it, each would have to have this recording logic. so it's cleaner to have it abstracted somewhere else.

The reason I did it in this patch is so this patch doesn't remove functionality. Currently, the cleanup lm ensures that if the lm is closed, then all outstanding requests are completed. If we don't implement this here, then this functionality is lost.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
ivankelly: in this case, it would be a BKException, though when exceptions are wrapped and when they are not is not straightforward. 

If I have an future X and I call completeExceptionally on it, all handlers for that exception will get the raw exception. Any futures which have been created when installing the handlers will get the wrapped exception. 

In any case, I'm going remove all the instanceof and just pull the error codes out

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BKException.java
ivankelly: Will add to javadoc

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperAdmin.java
ivankelly: No, shouldn't be, but it's dependent on how readLedgerMetadata is implemented, so will change to getCode

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BKException.java
ivankelly: You'll get CompletionException wrapping, if the future you put your handler is not the future on which completeExceptionally is called.

sijie:
run bookkeeper replication tests
run integration tests

sijie:

sijie:
run integration tests

PR #1810, opened 2018-11-14
Remove duplication logic for 'minNumRacksPerWriteQuorum' handling in RackawareEnsemblePlacementPolicyImpl

reddycharan:

With this change
https://github.com/apache/bookkeeper/commit/9ba4c4e0d8be770e03110a958fb8b75a65ae0f59
we introduced 'minNumRacksPerWriteQuorum' config option. To handle that, logic is added in
RackawareEnsemblePlacementPolicyImpl.newEnsembleInternal and RackQuorumCoverageSet. But in
retrospect, this is kind of duplicate logic and changes made in RackawareEnsemblePlacementPolicyImpl.newEnsembleInternal
is not really needed, because anyhow 'apply' of predicate (RRTopologyAwareCoverageEnsemble) is called, which calls
'apply' method on all the concerned RackQuorumCoverageSet.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
@@ -554,9 +554,7 @@ public class RackawareEnsemblePlacementPolicyImpl extends TopologyAwareEnsembleP
                 }
                 return addrs;
             }
-            // pick nodes by racks, to ensure there is at least write quorum number of racks.
-            int idx = 0;
-            String[] racks = new String[ensembleSize];
+
             for (int i = 0; i < ensembleSize; i++) {
                 String curRack;
                 if (null == prevNode) {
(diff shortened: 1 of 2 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
```diff
@@ -566,64 +564,11 @@ public void handleBookiesThatJoined(Set<BookieSocketAddress> joinedBookies) {
...
-                            if (racks[nextIndex] != null) {
-                                if (!((sb.length() == 1) && (sb.charAt(0) == '~'))) {
-                                    sb.append(NetworkTopologyImpl.NODE_SEPARATOR);
-                                }
-                                sb.append(racks[nextIndex]);
-                            }
-                        }
-                    }
-                    curRack = sb.toString();
+                    curRack = "~" + prevNode.getNetworkLocation();
```
athanatos: I'm actually not even sure this part is necessary.

reddycharan: we have been making sure that node selected will be atleast different from the rack of the previous node. I don't want to revert this existing behavior.

PR #1813, opened 2018-11-15
Set default sizes of DbLedgerStorage read and write cache to be propo…

merlimat:
…rtional to JVM direct memory

### Motivation

To simplify Bookie configuration when using `DbLedgerStorage`, set the memory size defaults for WriteCache, ReadCache and RocksDB block cache to be pegged to the available direct memory configured in the JVM.

User can always configure specific values and override this behavior.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
@@ -28,6 +28,9 @@ import com.google.common.collect.Lists;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.util.concurrent.DefaultThreadFactory;
+//CHECKSTYLE.OFF: IllegalImport
+import io.netty.util.internal.PlatformDependent;
+//CHECKSTYLE.ON: IllegalImport
 
 import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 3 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
```diff
@@ -28,6 +28,9 @@
 
 import io.netty.buffer.ByteBuf;
 import io.netty.util.concurrent.DefaultThreadFactory;
+//CHECKSTYLE.OFF: IllegalImport
```
eolivelli: This one?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageRocksDB.java
```diff
@@ -24,6 +24,10 @@
 
 import com.google.common.primitives.UnsignedBytes;
 
+//CHECKSTYLE.OFF: IllegalImport
```
eolivelli: This comment can be removed as well

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
eolivelli: Sorry, why is it illegal?

merlimat: It was marked here: 

https://github.com/apache/bookkeeper/blob/28dee8464764b0edceeb31ce24424f0947b5789b/buildtools/src/main/resources/bookkeeper/checkstyle.xml#L104-L106

eolivelli:
Mmm
But if we wat to use it we should remove it from the blacklist.
Otherwise who reads checkstyle config thinks that we are not using it...

What do you think?
Do you know why it was blacklisted?

merlimat:
> What do you think?
Do you know why it was blacklisted?

I have no idea why it was blacklisted. I guess `io.netty.util.internal` could be regarded as non-public API package. Maybe it's because there are also some shaded classes under that package.

In my opinion, the override here is fine. We should only using things from `io.netty.util.internal` if there's a good reason for it and no other way around.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
eolivelli: Okay

sijie:
the reason why it is in blacklist is internal classes are not encouraged to use in general, if there are exceptions, those imports should be highlighted with these `checkstyle` annotations. so we know these packages are used intentionally.

merlimat:
@sijie  Updated

merlimat:
run integration tests

PR #1787, opened 2018-11-05
 [stats] introduce `StatsDoc` annotation for better documenting metrics exposed by bookkeeper

sijie:
Descriptions of the changes in this PR:

### Motivation

A common ask from people using bookkeeper is how they can monitor bookies and bookkeeper clients, what kind of metrics that bookkeeper exposes
and what are the important metrics. Currently bookkeeper doesn't provide any metrics page for guiding people on monitoring bookkeeper services.

In order to help people on this, we need to provide a few documentation pages about metrics. However if we just write such pages, those pages
can quickly get out-of-dated when code is changed.

### Changes

- Introduce an annotation `StatsDoc` for annotating the counters/gauges/opstats in the source code.
- Provide a tool to generate the stats and their documentation into a yaml file.

The yaml file will be used by the website for rendering a metrics reference page.

### Results

```
"server":
  "bookie_BOOKIE_READ_ENTRY_BYTES":
    "description": |-
      bytes stats of ReadEntry on a bookie
    "type": |-
      OPSTATS
  "bookie_WRITE_BYTES":
    "description": |-
      total bytes written to a bookie
    "type": |-
      COUNTER
  "bookie_BOOKIE_ADD_ENTRY":
    "description": |-
      operations stats of AddEntry on a bookie
    "type": |-
      OPSTATS
  "bookie_BOOKIE_RECOVERY_ADD_ENTRY":
    "description": |-
      operation stats of RecoveryAddEntry on a bookie
    "type": |-
      OPSTATS
  "bookie_BOOKIE_ADD_ENTRY_BYTES":
    "description": |-
      bytes stats of AddEntry on a bookie
    "type": |-
      OPSTATS
  "bookie_BOOKIE_FORCE_LEDGER":
    "description": |-
      total force operations occurred on a bookie
    "type": |-
      COUNTER
  "bookie_READ_BYTES":
    "description": |-
      total bytes read from a bookie
    "type": |-
      COUNTER
  "bookie_BOOKIE_READ_ENTRY":
    "description": |-
      operation stats of ReadEntry on a bookie
    "type": |-
      OPSTATS
```

Master Issue: #1786

> ---
> In order to uphold a high standard for quality for code contributions, Apache BookKeeper runs various precommit
> checks for pull requests. A pull request can only be merged when it passes precommit checks. However running all
> the precommit checks can take a long time, some trivial changes don't need to run all the precommit checks. You
> can check following list to skip the tests that don't need to run for your pull request. Leave them unchecked if
[...]

PR #1817, opened 2018-11-19
Rollback snakeyaml version (Fixes IT tests)

ivankelly:
The stats gen (#1787) change introduced a version of snakeyaml which
is incompatible with arquillian-cube (between 1.19 and 1.23 they
remove some methods).

This was picked up by the integration tests, but overriden to submit.

This change pins the snakeyaml version at 1.19, thereby allowing the
integration tests to run again.


ivankelly:
5 other changes have gone in since the breaking change, most of them also ignoring IT

ivankelly:

PR #1819, opened 2018-11-19
ISSUE-1770: Add local checker for Sorted/InterleavedLedgerStorage

athanatos:
The main goal of this patch is the ScrubberService LifecycleComponent
which runs in the background periodically verifying the internal
consistency of the LedgerStorage. To get that to work:

- LedgerStorage now has a localConsistencyCheck call with
implementations in Interleaved and Sorted.

- In service of that implementation, LedgerCache gains an interface for
iterating safely over the entries of a ledger with a way of handling
concurrently modified or deleted ledgers with corresponding
modifications to LedgerEntryPage for detecting deletion.

- EntryLogger has been refactored to permit checking the correctness
(and throwing a descriptive exception in case of a problem) of an entry
without actually reading it for use within localConsistencyCheck.

- The two mechanisms for iterating over a ledger's entries in
BookieShell have both been replaced with the new single implementation
(InterleavedLedgerStorageTest now has a test checking the output of the
affected command.)

- Misc changes to *LogCompactor to support tests in
InterleavedLedgerStorageTest.

Because the consistency check needs to run in the background and hold
LEP instances potentially for a relatively long time, a delete may
overlap with the scan of an LEP page. As part of this patch,
IndexInMemPageMgr and LedgerEntryPage now permit an LEP to be marked
deleted and not added back to the set of free pages until released.

This patch also adds an option to run the checker on startup (defaults
to false).

(@bug W-5188823@)
(@bug W-5153309@)
(@rev cguttapalem@)

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
@@ -47,6 +47,7 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
@@ -602,6 +603,64 @@ public class Bookie extends BookieCriticalThread {
         this(conf, NullStatsLogger.INSTANCE);
     }
 
+    private static LedgerStorage buildLedgerStorage(ServerConfiguration conf) throws IOException {
+        // Instantiate the ledger storage implementation
+        String ledgerStorageClass = conf.getLedgerStorageClass();
+        LOG.info("Using ledger storage: {}", ledgerStorageClass);
+        return LedgerStorageFactory.createLedgerStorage(ledgerStorageClass);
+    }
+
+    /**
+     * Initialize LedgerStorage instance without checkpointing for use within the shell
+     * and other RO users.  ledgerStorage must not have already been initialized.
+     *
+     * @param conf Bookie config.
+     * @param ledgerStorage Instance to initialize.
+     * @return Passed ledgerStorage instance
+     * @throws IOException
+     */
+    static LedgerStorage mountLedgerStorageOffline(
+            ServerConfiguration conf,
+            LedgerStorage ledgerStorage) throws IOException {
+        StatsLogger statsLogger = NullStatsLogger.INSTANCE;
+        DiskChecker diskChecker = new DiskChecker(conf.getDiskUsageThreshold(), conf.getDiskUsageWarnThreshold());
+
+        LedgerDirsManager ledgerDirsManager = createLedgerDirsManager(
+                conf, diskChecker, statsLogger.scope(LD_LEDGER_SCOPE));
+        LedgerDirsManager indexDirsManager = createIndexDirsManager(
+                conf, diskChecker, statsLogger.scope(LD_INDEX_SCOPE), ledgerDirsManager);
+
+        if (null == ledgerStorage) {
+            ledgerStorage = buildLedgerStorage(conf);
+        }
+
+        CheckpointSource checkpointSource = new CheckpointSource() {
+            @Override
+            public Checkpoint newCheckpoint() {
+                return Checkpoint.MAX;
+            }
+
+            @Override
+            public void checkpointComplete(Checkpoint checkpoint, boolean compact)
+                    throws IOException {
+            }
+        };
+
+        Checkpointer checkpointer = Checkpointer.NULL;
+
+        ledgerStorage.initialize(
+                conf,
+                null,
+                ledgerDirsManager,
+                indexDirsManager,
+                null,
+                checkpointSource,
+                checkpointer,
+                statsLogger);
+
+        return ledgerStorage;
+    }
+
     public Bookie(ServerConfiguration conf, StatsLogger statsLogger)
             throws IOException, InterruptedException, BookieException {
         super("Bookie-" + conf.getBookiePort());
@@ -677,10 +736,7 @@ public class Bookie extends BookieCriticalThread {
         this.entryLogPerLedgerEnabled = conf.isEntryLogPerLedgerEnabled();
         CheckpointSource checkpointSource = new CheckpointSourceList(journals);
 
-        // Instantiate the ledger storage implementation
-        String ledgerStorageClass = conf.getLedgerStorageClass();
-        LOG.info("Using ledger storage: {}", ledgerStorageClass);
-        ledgerStorage = LedgerStorageFactory.createLedgerStorage(ledgerStorageClass);
+        ledgerStorage = buildLedgerStorage(conf);
 
         boolean isDbLedgerStorage = ledgerStorage instanceof DbLedgerStorage;
 
@@ -872,6 +928,25 @@ public class Bookie extends BookieCriticalThread {
             LOG.error("Error on executing a fully flush after replaying journals.");
             shutdown(ExitCode.BOOKIE_EXCEPTION);
         }
+
+        if (conf.isLocalConsistencyCheckOnStartup()) {
+            LOG.info("Running local consistency check on startup prior to accepting IO.");
+            List<LedgerStorage.DetectedInconsistency> errors = null;
+            try {
+                errors = ledgerStorage.localConsistencyCheck(Optional.empty());
+            } catch (IOException e) {
+                LOG.error("Got a fatal exception while checking store", e);
+                shutdown(ExitCode.BOOKIE_EXCEPTION);
+            }
+            if (errors != null && errors.size() > 0) {
+                LOG.error("Bookie failed local consistency check:");
+                for (LedgerStorage.DetectedInconsistency error : errors) {
+                    LOG.error("Ledger {}, entry {}: ", error.getLedgerId(), error.getEntryId(), error.getException());
+                }
+                shutdown(ExitCode.BOOKIE_EXCEPTION);
+            }
+        }
+
         LOG.info("Finished reading journal, starting bookie");
 
 
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
@@ -614,7 +614,7 @@ public class Bookie extends BookieCriticalThread {
      * Initialize LedgerStorage instance without checkpointing for use within the shell
      * and other RO users.  ledgerStorage must not have already been initialized.
      *
-     * The caller is responsible for disposing of the ledgerStorage object.
+     * <p>The caller is responsible for disposing of the ledgerStorage object.
      *
      * @param conf Bookie config.
      * @param ledgerStorage Instance to initialize.
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
@@ -614,6 +614,8 @@ public class Bookie extends BookieCriticalThread {
      * Initialize LedgerStorage instance without checkpointing for use within the shell
      * and other RO users.  ledgerStorage must not have already been initialized.
      *
+     * The caller is responsible for disposing of the ledgerStorage object.
+     *
      * @param conf Bookie config.
      * @param ledgerStorage Instance to initialize.
      * @return Passed ledgerStorage instance
@@ -927,6 +929,7 @@ public class Bookie extends BookieCriticalThread {
         } catch (ExecutionException e) {
             LOG.error("Error on executing a fully flush after replaying journals.");
             shutdown(ExitCode.BOOKIE_EXCEPTION);
+            return;
         }
 
         if (conf.isLocalConsistencyCheckOnStartup()) {
@@ -937,6 +940,7 @@ public class Bookie extends BookieCriticalThread {
             } catch (IOException e) {
                 LOG.error("Got a fatal exception while checking store", e);
                 shutdown(ExitCode.BOOKIE_EXCEPTION);
+                return;
             }
             if (errors != null && errors.size() > 0) {
                 LOG.error("Bookie failed local consistency check:");
@@ -944,6 +948,7 @@ public class Bookie extends BookieCriticalThread {
                     LOG.error("Ledger {}, entry {}: ", error.getLedgerId(), error.getEntryId(), error.getException());
                 }
                 shutdown(ExitCode.BOOKIE_EXCEPTION);
+                return;
             }
         }
 
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
@@ -26,7 +26,6 @@ import static org.apache.bookkeeper.tools.cli.helpers.CommandHelpers.getBookieSo
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Lists;
-import com.google.common.collect.Maps;
 import com.google.common.util.concurrent.UncheckedExecutionException;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufUtil;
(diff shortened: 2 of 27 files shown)


athanatos:
Fixing the spotbugs problems.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -724,12 +807,42 @@ public ByteBuf internalReadEntry(long ledgerId, long entryId, long location)
...
 
-        ByteBuf data = PooledByteBufAllocator.DEFAULT.directBuffer(entrySize, entrySize);
-        int rc = readFromLogChannel(entryLogId, fc, data, pos);
-        if (rc != entrySize) {
+    public ByteBuf internalReadEntry(long ledgerId, long entryId, long location)
+            throws IOException, Bookie.NoEntryException {
+        long entryLogId = logIdForOffset(location);
+        long pos = location & 0xffffffffL;
+
+        final EntryLogEntry entry;
```
sijie: it seems that now we generate a new object for each read?

bookkeeper-server/src/main/java/org/apache/bookkeeper/server/service/ScrubberService.java
```diff
@@ -0,0 +1,141 @@
...
+import org.apache.bookkeeper.server.conf.BookieConfiguration;
+import org.apache.bookkeeper.stats.Counter;
+import org.apache.bookkeeper.stats.OpStatsLogger;
+import org.apache.bookkeeper.stats.StatsLogger;
+import org.apache.bookkeeper.util.MathUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A {@link org.apache.bookkeeper.common.component.LifecycleComponent} that runs stats provider.
```
sijie: can you update the comment?

eolivelli:
Awesome work.
It is a big patch, it needs time for review. I will finish tomorrow

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -863,6 +919,25 @@ public synchronized void start() {
...
+            try {
+                errors = ledgerStorage.localConsistencyCheck(Optional.empty());
+            } catch (IOException e) {
+                LOG.error("Got a fatal exception while checking store", e);
+                shutdown(ExitCode.BOOKIE_EXCEPTION);
+            }
+            if (errors != null && errors.size() > 0) {
+                LOG.error("Bookie failed local consistency check:");
+                for (LedgerStorage.DetectedInconsistency error : errors) {
+                    LOG.error("Ledger {}, entry {}: ", error.getLedgerId(), error.getEntryId(), error.getException());
```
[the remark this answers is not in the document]
athanatos: I actually want the whole stack trace there.  This layer doesn't know anything about the nature of the underlying implementation, and more information is better.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
```diff
@@ -740,19 +748,69 @@ public int runCmd(CommandLine cmdLine) throws Exception {
...
+            } else if ((bkConf.getLedgerStorageClass().equals(SortedLedgerStorage.class.getName())
+                    || bkConf.getLedgerStorageClass().equals(InterleavedLedgerStorage.class.getName()))) {
+                ServerConfiguration conf = new ServerConfiguration(bkConf);
+                InterleavedLedgerStorage interleavedStorage = new InterleavedLedgerStorage();
+                Bookie.mountLedgerStorageOffline(conf, interleavedStorage);
+
                 if (printMeta) {
                     // print meta
-                    readLedgerMeta(ledgerId);
+                    printInfoLine("===== LEDGER: " + ledgerIdFormatter.formatLedgerId(ledgerId) + " =====");
```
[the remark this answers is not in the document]
athanatos: There weren't any other users and I found this to be clearer.  I don't feel strongly about it and could be convinced otherwise.

athanatos:
@eolivelli Had a chance to take a look?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/IndexInMemPageMgr.java
```diff
@@ -318,7 +308,11 @@ public void onSetInUse(LedgerEntryPage lep) {
 
         @Override
         public void onResetInUse(LedgerEntryPage lep) {
-            addToCleanPagesList(lep);
+            if (!lep.isDeleted()) {
```
eolivelli: Is this a bug fix?

bookkeeper-server/src/main/java/org/apache/bookkeeper/server/service/ScrubberService.java
```diff
@@ -0,0 +1,141 @@
...
+
+        this.ledgerStorage = ledgerStorage;
+
+        this.scrubCounter = logger.getOpStatsLogger(RUN_DURATION);
+        this.errorCounter = logger.getCounter(DETECTED_SCRUB_ERRORS);
+        this.fatalErrorCounter = logger.getCounter(DETECTED_FATAL_SCRUB_ERRORS);
+    }
+
+    private long getNextPeriodMS() {
+        return (long) (((double) scrubPeriod) * (1.5 - rng.nextDouble()) * 1000);
```
eolivelli: If scrubPeriod <= 0 this will break

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/InterleavedLedgerStorageTest.java
```diff
@@ -0,0 +1,305 @@
...
+
+    @Before
+    public void setUp() throws Exception {
+        File tmpDir = File.createTempFile("bkTest", ".dir");
+        tmpDir.delete();
+        tmpDir.mkdir();
+        File curDir = Bookie.getCurrentDirectory(tmpDir);
+        Bookie.checkDirectoryStructure(curDir);
+
+        System.out.println(tmpDir);
```
eolivelli: Is this useful?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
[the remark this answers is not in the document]
athanatos: Oh, it was touched.  That's the reason for the new unit tests.  I reworked them to use the new entry iteration machinery.

athanatos:
Thanks for the review, I'll get an updated version out this week.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -601,6 +602,64 @@ public Bookie(ServerConfiguration conf)
...
+     * @param conf Bookie config.
+     * @param ledgerStorage Instance to initialize.
+     * @return Passed ledgerStorage instance
+     * @throws IOException
+     */
+    static LedgerStorage mountLedgerStorageOffline(
+            ServerConfiguration conf,
+            LedgerStorage ledgerStorage) throws IOException {
+        StatsLogger statsLogger = NullStatsLogger.INSTANCE;
+        DiskChecker diskChecker = new DiskChecker(conf.getDiskUsageThreshold(), conf.getDiskUsageWarnThreshold());
```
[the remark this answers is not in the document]
athanatos: Probably, but I'd like to leave that refactor for another time.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -863,6 +919,25 @@ public synchronized void start() {
...
         }
+
+        if (conf.isLocalConsistencyCheckOnStartup()) {
+            LOG.info("Running local consistency check on startup prior to accepting IO.");
+            List<LedgerStorage.DetectedInconsistency> errors = null;
+            try {
+                errors = ledgerStorage.localConsistencyCheck(Optional.empty());
+            } catch (IOException e) {
+                LOG.error("Got a fatal exception while checking store", e);
+                shutdown(ExitCode.BOOKIE_EXCEPTION);
```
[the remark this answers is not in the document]
athanatos: Yeah, fixing.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -863,6 +919,25 @@ public synchronized void start() {
...
+            }
+            if (errors != null && errors.size() > 0) {
+                LOG.error("Bookie failed local consistency check:");
+                for (LedgerStorage.DetectedInconsistency error : errors) {
+                    LOG.error("Ledger {}, entry {}: ", error.getLedgerId(), error.getEntryId(), error.getException());
+                }
+                shutdown(ExitCode.BOOKIE_EXCEPTION);
```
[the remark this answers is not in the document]
athanatos: Yep, fixing.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedChannelBase.java
```diff
@@ -25,6 +25,12 @@
  * to buffer the input and output data. This class is a base class for wrapping the {@link FileChannel}.
  */
 public abstract class BufferedChannelBase {
+    static class ChannelClosed extends IOException {
```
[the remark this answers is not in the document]
athanatos: Ok

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -693,26 +701,101 @@ static long logIdForOffset(long offset) {
...
 
-        if (readFromLogChannel(entryLogId, fc, sizeBuff, pos) != sizeBuff.capacity()) {
-            throw new Bookie.NoEntryException("Short read from entrylog " + entryLogId,
-                                              ledgerId, entryId);
+        try {
+            if (readFromLogChannel(entryLogId, fc, sizeBuff, pos) != sizeBuff.capacity()) {
+                throw new EntryLookupException.MissingEntryException(ledgerId, entryId, entryLogId, pos);
+            }
+        } catch (BufferedChannelBase.ChannelClosed | AsynchronousCloseException e) {
+            throw new EntryLookupException.MissingLogFileException(ledgerId, entryId, entryLogId, pos);
```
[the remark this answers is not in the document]
athanatos: AFAICT, it only happens if the deletion of the EntryLogger races with that read.  It's not really common prior to this patch, but this implementation can easily be checking an EntryLogger as it gets deleted.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
athanatos: Yeah, but I wanted to unify the getFCForEntryInternal logic and you can't return a compound structure without it being on the heap.  We could pool it, or use a per-thread preallocated holder, or w/e, but this is simple enough for now unless you think it's worth adding additional complexity.

bookkeeper-server/src/main/java/org/apache/bookkeeper/server/service/ScrubberService.java
athanatos: Ok.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -601,6 +602,64 @@ public Bookie(ServerConfiguration conf)
...
+        StatsLogger statsLogger = NullStatsLogger.INSTANCE;
+        DiskChecker diskChecker = new DiskChecker(conf.getDiskUsageThreshold(), conf.getDiskUsageWarnThreshold());
+
+        LedgerDirsManager ledgerDirsManager = createLedgerDirsManager(
+                conf, diskChecker, statsLogger.scope(LD_LEDGER_SCOPE));
+        LedgerDirsManager indexDirsManager = createIndexDirsManager(
+                conf, diskChecker, statsLogger.scope(LD_INDEX_SCOPE), ledgerDirsManager);
+
+        if (null == ledgerStorage) {
+            ledgerStorage = buildLedgerStorage(conf);
```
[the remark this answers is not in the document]
athanatos: Sure it is, the caller of this method is responsible for dealing with it.  I'll clarify that in the comment.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/IndexInMemPageMgr.java
athanatos: Not really, the isDeleted() concept came in with this patch.  It's there to deal with the inevitability of the consistency checker holding an LEP reference as the corresponding ledger is deleted.

bookkeeper-server/src/main/java/org/apache/bookkeeper/server/service/ScrubberService.java
athanatos: I'll update the ServerConfiguration comments to clarify that the service is only enabled if > 0 and add an assert to the constructor to check.

athanatos: The service won't be instantiated otherwise.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/InterleavedLedgerStorageTest.java
athanatos: It is not.  I'll remove it.

athanatos:
Rebase to address conflict and added commit addressing review comments.  I responded on the others.

eolivelli:
Thank you
Looks good to me, a part last nit about assert in not test code.
Then the patch is ready

athanatos:
@eolivelli Ok, pushed a fix for the assert.

athanatos:
Should I go ahead and merge?

eolivelli:
Sure

athanatos:
rerun pr validation

athanatos:
The last two changes angered checkstyle, I'll fix that and then merge.

eolivelli:
Now you should wait for CI before merging

athanatos:
Yup.  Looks like we're good now, merging.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -724,12 +807,42 @@ public ByteBuf internalReadEntry(long ledgerId, long entryId, long location)
...
         if (entrySize < MIN_SANE_ENTRY_SIZE) {
             LOG.error("Read invalid entry length {}", entrySize);
-            throw new IOException("Invalid entry length " + entrySize);
+            throw new EntryLookupException.InvalidEntryLengthException(ledgerId, entryId, entryLogId, pos);
+        }
+
+        long thisLedgerId = sizeBuff.getLong(4);
+        long thisEntryId = sizeBuff.getLong(12);
+        if (thisLedgerId != ledgerId || thisEntryId != entryId) {
+            throw new EntryLookupException.WrongEntryException(
```
merlimat: This is problematic because it happens even from the `internalReadEntry()`. The purpose of `internalReadEntry()` was to avoid the sanity checks from the regular `readEntry()` and to be called when doing read-ahead. 

In read-ahead mode, we don't know whether the next entry is from the same ledger or not. Therefore we keep reading until we find an entry of a different ledger. 

With this change in 4.9.2 that means each read-ahead incurs in creating an exception, which is very expensive (since it will walk through the stack trace each time).

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -724,12 +807,42 @@ public ByteBuf internalReadEntry(long ledgerId, long entryId, long location)
...
-        int rc = readFromLogChannel(entryLogId, fc, data, pos);
-        if (rc != entrySize) {
+    public ByteBuf internalReadEntry(long ledgerId, long entryId, long location)
+            throws IOException, Bookie.NoEntryException {
+        long entryLogId = logIdForOffset(location);
+        long pos = location & 0xffffffffL;
+
+        final EntryLogEntry entry;
+        try {
+            entry = getFCForEntryInternal(ledgerId, entryId, entryLogId, pos);
```
merlimat: It seems to be that a less expensive way would to be to just an enum code to signal a particular error and also avoid the `EntryLogEntry` allocation for each read.

PR #1824, opened 2018-11-21
AbstractZkLedgerManager should catch all exceptions from parseConfig

ivankelly:
And always complete the request if an exception does occur, rather
than throwing it out to the higher level, and hanging the request
forever.


ivankelly:
rerun integration tests

PR #1825, opened 2018-11-21
LedgerMetadata#parseConfig uses Metadata Builder

ivankelly:
parseConfig is one of the main places that mutates member fields of
LedgerMetadata. This change removes these mutates to make it use
LedgerMetadataBuilder instead.

Master issue: #281

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
@@ -22,7 +22,6 @@ import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkState;
 
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.base.Optional;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Maps;
 import com.google.protobuf.ByteString;
@@ -38,9 +37,11 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.NavigableMap;
+import java.util.Optional;
 import java.util.Set;
 import java.util.SortedMap;
 import java.util.TreeMap;
+import java.util.stream.Collectors;
 import lombok.EqualsAndHashCode;
 import org.apache.bookkeeper.client.api.DigestType;
 import org.apache.bookkeeper.net.BookieSocketAddress;
@@ -72,14 +73,13 @@ public class LedgerMetadata implements org.apache.bookkeeper.client.api.LedgerMe
     public static final String VERSION_KEY = "BookieMetadataFormatVersion";
 
     private int metadataFormatVersion = 0;
-
     private int ensembleSize;
     private int writeQuorumSize;
     private int ackQuorumSize;
     private long length;
     private long lastEntryId;
-    private long ctime;
-    boolean storeSystemtimeAsLedgerCreationTime = false;
+    final Optional<Long> ctime;
+    final Optional<Long> metadataStoreCtime;
 
     private LedgerMetadataFormat.State state;
     private TreeMap<Long, ImmutableList<BookieSocketAddress>> ensembles =  new TreeMap<>();
@@ -102,13 +102,14 @@ public class LedgerMetadata implements org.apache.bookkeeper.client.api.LedgerMe
         this.writeQuorumSize = writeQuorumSize;
         this.ackQuorumSize = ackQuorumSize;
         if (storeSystemtimeAsLedgerCreationTime) {
-            this.ctime = System.currentTimeMillis();
+            this.ctime = Optional.of(System.currentTimeMillis());
+            this.metadataStoreCtime = Optional.empty();
         } else {
             // if client disables storing its system time as ledger creation time, there should be no ctime at this
             // moment.
-            this.ctime = -1L;
+            this.ctime = Optional.empty();
+            this.metadataStoreCtime = Optional.empty();
         }
-        this.storeSystemtimeAsLedgerCreationTime = storeSystemtimeAsLedgerCreationTime;
 
         /*
          * It is set in PendingReadOp.readEntryComplete, and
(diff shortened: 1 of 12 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
```diff
@@ -486,89 +482,94 @@ public static LedgerMetadata parseConfig(byte[] bytes, Optional<Long> msCtime) t
...
+            builder.withPassword(data.getPassword().toByteArray())
+                .withDigestType(protoToApiDigestType(data.getDigestType()));
         }
 
         for (LedgerMetadataFormat.Segment s : data.getSegmentList()) {
-            ArrayList<BookieSocketAddress> addrs = new ArrayList<BookieSocketAddress>();
-            for (String member : s.getEnsembleMemberList()) {
-                addrs.add(new BookieSocketAddress(member));
+            List<BookieSocketAddress> addrs = new ArrayList<>();
+            for (String a : s.getEnsembleMemberList()) {
```
sijie: using a variable 'a' is generally not a good practice. why are you changing 'member' to 'a'?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
```diff
@@ -691,4 +690,23 @@ Long getLastEnsembleKey() {
         checkState(!ensembles.isEmpty(), "Metadata should never be created with no ensembles");
         return ensembles.lastKey();
     }
+
+    int getMetadataFormatVersion() {
+        return metadataFormatVersion;
+    }
+
+    private static DigestType protoToApiDigestType(LedgerMetadataFormat.DigestType digestType) {
```
sijie: I would suggest pulling this method into a util class. so it can be used across the project.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
```diff
@@ -72,14 +73,13 @@
...
-
     private int ensembleSize;
     private int writeQuorumSize;
     private int ackQuorumSize;
     private long length;
     private long lastEntryId;
-    private long ctime;
-    boolean storeSystemtimeAsLedgerCreationTime = false;
+    final Optional<Long> ctime;
+    final Optional<Long> metadataStoreCtime;
```
sijie: this doesn't sound a right refactor to me. it is also very confusing and hard to track, when you are mapping a write behavior (whether to store system time as ledger cation time) to a read behavior (whether metadata store ctime is set). 

If we are doing a refactor, I would suggest : 

- keep only one ctime in the metadata
- when building ledger metadata, passing it the `storeSystemtimeAsLedgerCreationTime` when building proto ledger metadata. simply because the flag is controlling write time only.

ivankelly: This whole thing is a hacky abomination. I made the change because passing storeSystemtimeAsLedgerCreationTime is passing client configuration into metadata object, where the scope of the lifetime of these objects is completely different. Still looks hacky though.

I think the correct way to do this is to pass a flag to the serialize() type method. This is where this storage decision should be taken, just as we take a decision on which version we want to store (which will be part of the binary metadata change).

For now, I'll revert this part of the change, and will deal with it in a later PR.

sijie: works for me

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
ivankelly: changed it to stream() implementation, compiled and it complained about BookieSocketAddress exceptions so changed it back, but kept variable names. Will revert.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
```diff
@@ -260,7 +255,11 @@ boolean hasPassword() {
 
     @VisibleForTesting
     public byte[] getPassword() {
-        return Arrays.copyOf(password, password.length);
+        if (!hasPassword()) {
```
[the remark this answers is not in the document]
ivankelly: Was debugging an issue, and it involved adding a call to this, and this throws a NPE without the change if called.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
```diff
@@ -240,12 +240,7 @@ public int getAckQuorumSize() {
 
     @Override
     public long getCtime() {
-        return ctime;
-    }
-
-    @VisibleForTesting
-    void setCtime(long ctime) {
-        this.ctime = ctime;
+        return ctime.orElse(metadataStoreCtime.orElse(-1L));
```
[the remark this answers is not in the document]
ivankelly: will revert for now. See other comment.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
```diff
@@ -153,9 +157,7 @@ public LedgerMetadata(int ensembleSize,
         length.ifPresent((l) -> this.length = l);
         setEnsembles(ensembles);
 
-        if (state != LedgerMetadataFormat.State.CLOSED) {
-            currentEnsemble = this.ensembles.lastEntry().getValue();
-        }
+        currentEnsemble = this.ensembles.lastEntry().getValue();
```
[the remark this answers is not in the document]
ivankelly: ParallelLedgerRecoveryTest expect it, but I guess it's better to fix the test.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
```diff
@@ -72,14 +73,13 @@
...
     private int metadataFormatVersion = 0;
-
     private int ensembleSize;
     private int writeQuorumSize;
     private int ackQuorumSize;
     private long length;
     private long lastEntryId;
-    private long ctime;
-    boolean storeSystemtimeAsLedgerCreationTime = false;
+    final Optional<Long> ctime;
```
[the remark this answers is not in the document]
ivankelly: it is exposed for builder to copy.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
ivankelly: Not sure where it would fit, so leaving it here for now. We have too many of these conversions though. We should push them all into a DigestConvertor class at some point, but I don't want to do that in this patch as it would mean touching other classes which are completely unrelated.

PR #1826, opened 2018-11-22
Fix flake in AuditorPeriodicBookieCheckTest

ivankelly:
The flake was occurring because the bookie checker was trying to
connect to a TCP endpoint that didn't respond (1.1.1.1) and was
waiting for the whole tcp connect timeout (10sec), while the test was
waiting 10 seconds for the success condition to occur. The success
condition often occurred just after the test fails.

The solution was to bring the connect timeout for the client down to
500ms, giving the auditor plenty of time to register the
underreplicated ledger.

I guess this used to pass in the past because maybe 1.1.1.1 was
actually rejecting packets instead of dropping them, and these changed
when cloudflare took over the address. I changed the endpoint to
192.0.2.0/24 range (TEST-NET) to avoid a similar scenario in the
future.


sijie:

PR #1827, opened 2018-11-22
Remove LedgerMetadata constructors except one for builder

ivankelly:
LedgerMetadata shouldn't generally be created outside the client,
except for testing. In the case where it is, the builder should be
used to construct it.

I've left the copy constructor for now, as I don't feel comfortable
removing it until all fields are final, which will occur soon.

Master issue: #281


eolivelli:
There are failures on CI

ivankelly:
rerun integration tests

PR #1828, opened 2018-11-22
Make most LedgerMetadata fields final

ivankelly:
Make most fields in LedgerMetadata final. The ones which I have not
changed in this patch require a more involved change, so they'll come
in separate PRs.

Master issue: #281


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadata.java
```diff
@@ -126,11 +126,13 @@
...
-        password.ifPresent((pw) -> {
-                this.password = pw;
-                this.hasPassword = true;
-            });
-
+        if (password.isPresent()) {
+            this.password = password.get();
+            this.hasPassword = true;
+        } else {
+            this.password = new byte[0];
```
eolivelli: Won't 'null' be better?

ivankelly: changed

PR #1829, opened 2018-11-22
Move utility method from LedgerMetadata to test

ivankelly:
The method is only used in the test, so it should live there.

Master issue: #281


ivankelly:

ivankelly:
rerun integration tests

PR #1831, opened 2018-11-23
Make LedgerMetadata fields for closing immutable

ivankelly:
Master issue: #281

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadataBuilder.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadataBuilder.java
@@ -73,14 +73,9 @@ public class LedgerMetadataBuilder {
         builder.ackQuorumSize = other.getAckQuorumSize();
 
         builder.state = other.getState();
-
-        long lastEntryId = other.getLastEntryId();
-        if (lastEntryId != LedgerHandle.INVALID_ENTRY_ID) {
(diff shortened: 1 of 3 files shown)


ivankelly:
I need to fix up some tests, they seem to be making assumptions about unclosed ledgers

ivankelly:
I need to fix up some tests, they seem to be making assumptions about unclosed ledgers

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadataBuilder.java
```diff
@@ -73,14 +73,9 @@ public static LedgerMetadataBuilder from(LedgerMetadata other) {
...
         builder.state = other.getState();
-
-        long lastEntryId = other.getLastEntryId();
-        if (lastEntryId != LedgerHandle.INVALID_ENTRY_ID) {
-            builder.lastEntryId = Optional.of(lastEntryId);
-        }
-        long length = other.getLength();
-        if (length > 0) {
-            builder.length = Optional.of(length);
+        if (builder.state == LedgerMetadataFormat.State.CLOSED) {
```
eolivelli: Is this a behavior change?
Do you think it is worth to add a comment?

ivankelly: The previous behaviour was incorrect. length should never be set if the state is anything but closed. I don't think it needs a comment. the intent is clear.

PR #1833, opened 2018-11-23
Make custom metadata final

ivankelly:
Master issue: #281


ivankelly:
Ah, copy constructor change needs to go in before this

PR #1837, opened 2018-11-24
Remove MathUtils.now()

sijie:
Descriptions of the changes in this PR:

*Motivation*

`MathUtils.now()` using System.nanoTime doesn't provide accurate time. It is used
for measuring pupose. However since `nanoTime` can have numeric overflow, so when
it is used for converting into milliseconds, it is misleading and error-prone. A
lot of places using `MathUtis.now` can all replaced with `System.currentTimeMillis()`.

It leads to unknown behavior on compactions. examples is shown in the following figure.

*Changes*

Remove `MathUtils.now()` and replace it with `System.currentTimeMillis()`.

> ---
> In order to uphold a high standard for quality for code contributions, Apache BookKeeper runs various precommit
> checks for pull requests. A pull request can only be merged when it passes precommit checks. However running all
> the precommit checks can take a long time, some trivial changes don't need to run all the precommit checks. You
> can check following list to skip the tests that don't need to run for your pull request. Leave them unchecked if
> you are not sure, committers will help you:
>
> - [ ] [skip bookkeeper-server bookie tests]: skip testing `org.apache.bookkeeper.bookie` in bookkeeper-server module.
> - [ ] [skip bookkeeper-server client tests]: skip testing `org.apache.bookkeeper.client` in bookkeeper-server module.
> - [ ] [skip bookkeeper-server replication tests]: skip testing `org.apache.bookkeeper.replication` in bookkeeper-server module.
> - [ ] [skip bookkeeper-server tls tests]: skip testing `org.apache.bookkeeper.tls` in bookkeeper-server module.
> - [ ] [skip bookkeeper-server remaining tests]: skip testing all other tests in bookkeeper-server module.
> - [ ] [skip integration tests]: skip docker based integration tests. if you make java code changes, you shouldn't skip integration tests.
> - [ ] [skip build java8]: skip build on java8. *ONLY* skip this when *ONLY* changing files under documentation under `site`.
> - [ ] [skip build java9]: skip build on java9. *ONLY* skip this when *ONLY* changing files under documentation under `site`.
> ---

> ---
> Be sure to do all of the following to help us incorporate your contribution
> quickly and easily:
>
> If this PR is a BookKeeper Proposal (BP):
>
> - [ ] Make sure the PR title is formatted like:
>     `<BP-#>: Description of bookkeeper proposal`
>     `e.g. BP-1: 64 bits ledger is support`
> - [ ] Attach the master issue link in the description of this PR.
[...]

PR #1838, opened 2018-11-26
Add rest endpoint trigger_gc to trigger GC on Bookie

jiazhai:

Add rest endpoint trigger_gc to trigger GC on Bookie

### Motivation

Some times user would like to trigger GC manually instead of waiting to the timeout or disk full. 

> ---
>
> ---

> ---
> quickly and easily:
>
>
>
> Otherwise:
> 
> 
> ---

### DIFF
--- a/bookkeeper-http/http-server/src/main/java/org/apache/bookkeeper/http/HttpRouter.java
+++ b/bookkeeper-http/http-server/src/main/java/org/apache/bookkeeper/http/HttpRouter.java
@@ -45,6 +45,7 @@ public abstract class HttpRouter<Handler> {
     public static final String LAST_LOG_MARK                = "/api/v1/bookie/last_log_mark";
     public static final String LIST_DISK_FILE               = "/api/v1/bookie/list_disk_file";
     public static final String EXPAND_STORAGE               = "/api/v1/bookie/expand_storage";
+    public static final String TRIGGER_GC                   = "/api/v1/bookie/trigger_gc";
     // autorecovery
     public static final String RECOVERY_BOOKIE              = "/api/v1/autorecovery/bookie";
     public static final String LIST_UNDER_REPLICATED_LEDGER = "/api/v1/autorecovery/list_under_replicated_ledger";
@@ -73,6 +74,7 @@ public abstract class HttpRouter<Handler> {
         this.endpointHandlers.put(LAST_LOG_MARK, handlerFactory.newHandler(HttpServer.ApiType.LAST_LOG_MARK));
         this.endpointHandlers.put(LIST_DISK_FILE, handlerFactory.newHandler(HttpServer.ApiType.LIST_DISK_FILE));
         this.endpointHandlers.put(EXPAND_STORAGE, handlerFactory.newHandler(HttpServer.ApiType.EXPAND_STORAGE));
+        this.endpointHandlers.put(TRIGGER_GC, handlerFactory.newHandler(HttpServer.ApiType.TRIGGER_GC));
 
         // autorecovery
         this.endpointHandlers.put(RECOVERY_BOOKIE, handlerFactory.newHandler(HttpServer.ApiType.RECOVERY_BOOKIE));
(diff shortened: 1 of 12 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -382,6 +382,13 @@ public void runWithFlags(boolean force, boolean suspendMajor, boolean suspendMin
             lastMinorCompactionTime = System.currentTimeMillis();
             minorCompactionCounter.inc();
         }
+
+        if (force) {
+            if (forceGarbageCollection.compareAndSet(true, false)) {
+                LOG.info("{} Set forceGarbageCollection to true after force GC to make it forceGC-able again.", Thread
```
sijie: @jiazhai this is "set forcrGarbageCollection to false after force GC", no?

sijie: Also doesn't https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java#L338 already do that?

bookkeeper-http/http-server/src/main/java/org/apache/bookkeeper/http/HttpRouter.java
```diff
@@ -45,6 +45,7 @@
     public static final String LAST_LOG_MARK                = "/api/v1/bookie/last_log_mark";
     public static final String LIST_DISK_FILE               = "/api/v1/bookie/list_disk_file";
     public static final String EXPAND_STORAGE               = "/api/v1/bookie/expand_storage";
+    public static final String TRIGGER_GC                   = "/api/v1/bookie/trigger_gc";
```
sijie: I think it is better to just call it `/api/v1/bookie/gc`. a `PUT` method is to trigger gc and a `GET` method is to query the gc status. we can leave implementing `GET` in a separate pull request. but at least it allows extensibility.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/InterleavedLedgerStorage.java
```diff
@@ -191,6 +191,11 @@ public void diskJustWritable(File disk) {
         };
     }
 
+    @Override
+    public void forceGC() {
+        gcThread.enableForceGC();
```
sijie: this is not actually implementing a "trigger" behavior. it only sets flag, however the actual gc will be happening after `gcWaitInterval`. so I think a better solution would be:

```suggestion
        gcThread.enableForceGC();
        // submit a task immediately
        gcExecutor.submit(this);
```

sijie:
mark this for 4.7.3 as well, since this is needed for handling production issue introduced by #1837 without restarting bookies.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
jiazhai: Thanks. This depends. If `runWithFlags` is called in above safeRun(), this part of change is not needed.
In this PR it is force called by triggerGC, and will not use safeRun.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/InterleavedLedgerStorage.java
jiazhai: Thanks, Maybe the original name is misleading. Here is what enableForceGC does in gcThread:
```
    public void enableForceGC() {
        if (forceGarbageCollection.compareAndSet(false, true)) {
            LOG.info("Forced garbage collection triggered by thread: {}", Thread.currentThread().getName());
            triggerGC(true, suspendMajorCompaction.get(),
                      suspendMinorCompaction.get());
        }
    }

    Future<?> triggerGC(final boolean force,
                        final boolean suspendMajor,
                        final boolean suspendMinor) {
        return gcExecutor.submit(() -> {
                runWithFlags(force, suspendMajor, suspendMinor);
            });
    }

```

sijie: i see

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
sijie: okay

sijie:
@jiazhai can you fix the endpoint name? once it is fixed, we can move forward with 4.7.3 release.

PR #1849, opened 2018-11-29
Make Spotbugs pass on JDK11

eolivelli:
- upgrade spotbugs to 3.1.8
- disable spotbugs on JDK11 (due to for https://github.com/spotbugs/spotbugs/issues/756)

### Motivation

This change is on the way to give full "official" support do JDK11

### Changes

- upgrade spotbugs to 3.1.8
- fix new spotbugs issues
- disable spotbugs on JDK11 (due to for https://github.com/spotbugs/spotbugs/issues/756)

> ---
>
> ---

> ---
> quickly and easily:
>
>
>
> Otherwise:
> 
> 
> ---


bookkeeper-common/src/main/java/org/apache/bookkeeper/common/collections/RecyclableArrayList.java
```diff
@@ -64,4 +64,15 @@ public void recycle() {
             handle.recycle(this);
         }
     }
+
+    @Override
+    public boolean equals(Object obj) {
```
sijie: why do we need these? it seems to call `super.equals`?

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/collections/RecyclableArrayList.java
```diff
@@ -64,4 +64,15 @@ public void recycle() {
...
+        return super.equals(obj);
+    }
+
+    @Override
+    public int hashCode() {
```
sijie: why do we need this? if it is needed, add a comment.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
```diff
@@ -154,6 +155,8 @@
 /**
  * Bookie Shell is to provide utilities for users to administer a bookkeeper cluster.
  */
+@SuppressFBWarnings(value = "RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE",
```
sijie: I see these everywhere. I don't think it is good to disable just because it doesn't work with jdk11. 

I think we should just disable spotbugs for jdk11

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/collections/RecyclableArrayList.java
eolivelli: Spot bugs is not happy

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
eolivelli: jdk11  Is the current major java version, and it is LTS.
We should run spotbugs.
Javac is different, even while compiling for jdk8 release. I am spending a lot of time on this kind of issues in many projects.
If you prefer I can disable this rules on jdk11 but not completely spotbugs

sijie: please disable these rules on jdk11 only.

eolivelli: Okay, it will be tricky but it is worth.

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/collections/RecyclableArrayList.java
eolivelli: Spot bugs is not happy

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
eolivelli: @sijie 
I am re-thinking about disabling this rule only on JDK11+....
I feel it is not worth such complication. We should have two different findbugs configuration files, and with JDK11+ the problem will stay forever.

In Apache Kafka they have disabled it at all
https://github.com/apache/kafka/pull/5943/files

eolivelli: @sijie 
The same we did in Pravega and in ZooKeeper
https://github.com/pravega/pravega/commit/fbaa1577385bab730e6c2f10d9fcda48478a2162

https://github.com/apache/zookeeper/commit/b752ef66876a141035a42f30aad69e3166cad746#diff-b4a4d2c7fb2ac879fb126283ee65fd07

I am asking on JDK compiler dev mailing list for solutions, but any solution won't be feasible until jdk13, if any solution will be ever adopted

sijie: @eolivelli I would still prefer not disabling those rules for non-jdk11 environments. we can disable spotbugs for jdk11 only.

eolivelli: @sijie okay, I will update the patch.

Let's disable spotbugs from JDK11+ at all

eolivelli:
@sijie  I have addressed all of your comments.
This should be good to go now

eolivelli:
@sijie this one could go in 4.9 ?
I would like to setup CI for JDK11 and drop jdk9

I am working on support for JDK12 on my downstream applications now.
and this week Oracle made available early access versions of JDK13
:-)

sijie:
yes this one can go to 4.9

eolivelli:
@sijie do you approve this change?

eolivelli:
retest this please

eolivelli:
@sijie I have triggered a full rebuild in order to check the patch against current master.
I will merge and then send patch for new CI jobs on JDK11, then I will send another patch for removing jdk9

PR #1857, opened 2018-12-03
changingEnsemble should be negated before calling unset success

ivankelly:
If the first pending add op is completed, but does not have the
replaced bookie in its write set, callbacks are triggered straight
away.

Previously this would then hang forever, as the changingEnsemble would
be true. This patch sets changingEnsemble to false before calling
unsetSuccessAndSendWriteRequest so that if callbacks are triggered
straight away, they can actually complete. It also moves the call to
unsetSuccessAndSendWriteRequest outside of the metadataLock so that
the callbacks don't run inside the lock.

### DIFF
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/HandleFailuresTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/HandleFailuresTest.java
@@ -446,4 +446,50 @@ public class HandleFailuresTest {
         Assert.assertEquals(lh.getLedgerMetadata().getAllEnsembles().get(0L), Lists.newArrayList(b1, b2, b3));
         Assert.assertEquals(lh.getLedgerMetadata().getAllEnsembles().get(1L), Lists.newArrayList(b1, b2, b4));
     }
+
+    @Test
+    public void testHandleFailureBookieNotInWriteSet() throws Exception {
+        MockClientContext clientCtx = MockClientContext.create();
+        Versioned<LedgerMetadata> md = ClientUtil.setupLedger(clientCtx, 10L,
+                LedgerMetadataBuilder.create()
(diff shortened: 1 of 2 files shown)


ivankelly:
@athanatos actually flagged this in original review: https://github.com/apache/bookkeeper/pull/1646#discussion_r216768869

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/HandleFailuresTest.java
```diff
@@ -446,4 +446,50 @@ public void testNoAddsAreCompletedWhileFailureHandlingInProgress() throws Except
...
+
+        log.info("write second entry, should have enough bookies, but blocks completion on failure handling");
+        CompletableFuture<?> e2 = lh.appendAsync("entry2".getBytes());
+        changeInProgress.get();
+        assertEventuallyTrue("e2 should eventually complete", () -> lh.pendingAddOps.peek().completed);
+        Assert.assertFalse("e2 shouldn't be completed to client", e2.isDone());
+        blockEnsembleChange.complete(null); // allow ensemble change to continue
+
+        log.info("e2 should complete");
+        e2.get(10, TimeUnit.SECONDS);
```
eolivelli: Nit: we can use 'get()'. The test will fail due to timeout  if hanging endlessly

ivankelly: @eolivelli I'm trying with raw get() now, not timing out even after 6 minutes.

ivankelly: looks like the default timeout is 30 minutes, which is too much :/

PR #1858, opened 2018-12-03
Add max ledger metadata format version to layout

ivankelly:
The max ledger metadata format version is the maximum format version
that will be used to write ledger metadata. By setting it in the
ledger layout it becomes a cluster-wide configuration which is
initialized along with the cluster. Any cluster initialized with the
current code will end up with version 2. For this cluster serde will
only ever serialize with up to version 2 of the ledger metadata
format, so all clients that understand version 2 will continue to
work, even as the software version increases and new metadata formats
become available (such as the binary metadata format coming soon).

Currently there is no handling in LedgerMetadataSerDe, because we
don't have a new version, but when there is, it will use
```
Math.min(maxLedgerMetadataFormatVersion,
         metadata.getMetadataFormatVersion())
```
to decide which serialization method to use.


ivankelly:
@eolivelli there's no need in 4.8. This is only useful once we add a modification to the ledger metadata, which won't happen until we add the binary stuff

ivankelly:
rerun integration tests

Issue #723, opened 2017-11-14
Storing ledger metadata in binary format


sijie:
**FEATURE REQUEST**

1. Please describe the feature you are requesting.

Currently we stores ledger metadata in TextFormat protobuf not in binary format. Text format doesn't have a good backwards and forwards compatibility story as binary format. We should switch from text format to binary format. 

In order to keep backward compatibility and maintain good debuggability as text format, we can consider writing both text format and binary format.

2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have). Are you currently using any workarounds to address this issue?

*must-have* in 4.7

3. Provide any additional detail on your proposed use case for this feature.

sijie:
All the necessary changes are in master and will be shipped as part of 4.9.0.

PR #1864, opened 2018-12-05
 HTTP GetMetaService returns JSON rather than calling #serialize

ivankelly:
Once the binary ledger metadata serializing is available, the output
of GetMetaService would no longer be understandable. In preparation
for this, make GetMetaService use output JSON, rather than directly
calling #serialize.

Master issue: #723

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerCreateOp.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerCreateOp.java
@@ -132,7 +132,7 @@ class LedgerCreateOp {
             metadataBuilder.withCustomMetadata(customMetadata);
         }
         if (bk.getConf().getStoreSystemtimeAsLedgerCreationTime()) {
-            metadataBuilder.withCreationTime(System.currentTimeMillis());
+            metadataBuilder.withCreationTime(System.currentTimeMillis()).storingCreationTime(true);
         }
 
         // select bookies for first ensemble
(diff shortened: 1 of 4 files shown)


ivankelly:
If tools are relying on the format of this, we should use something that has some schema to it, like json.

eolivelli:
serialize() as it is part of our metadata protocol is well defined.

Maybe we should keep serializing in protobuf text format, even when we will switch to binary format.
How does it sounds to you?

sijie:
my take here is we should have a method `serializeAsJson` that dump the metadata as json (using protobuf TextFormat)

ivankelly:
It shouldn't expose the protobuf. the serialize format is an internal format. what is should expose is the jsonification of the api.LedgerMetadata interface.

sijie:
> what is should expose is the jsonification of the api.LedgerMetadata interface.

that's fine. my point is we should expose a json result in the http response. how the json result looks like can be aligned with the public api exposes.

ivankelly:
Seems we already pull in jackson, so it's pretty easy to do.

eolivelli:
Ok for json

ivankelly:
Ok, I have a json version for this, but it would be better to have the binary metadata in first (ready to go when #1865 in), due to ctime annoyances on json comparison

sijie:
@ivankelly what should we do for this issue?

ivankelly:
@sijie yes. This one doesn't even need the binary stuff anymore. Will rekick tests

ivankelly:
rerun integration tests

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerCreateOp.java
```diff
@@ -132,7 +132,7 @@ public void initiate() {
             metadataBuilder.withCustomMetadata(customMetadata);
         }
         if (bk.getConf().getStoreSystemtimeAsLedgerCreationTime()) {
-            metadataBuilder.withCreationTime(System.currentTimeMillis());
+            metadataBuilder.withCreationTime(System.currentTimeMillis()).storingCreationTime(true);
```
eolivelli: This sounds like an unrelated bug.
Okay for keeping.
I wonder why we did not catch this bug on tests

Issue #1863, opened 2018-12-05
EtcdLedgerManager doesn't support layout versioning


ivankelly:
So we can't set the max ledger metadata format version, among other things.

sijie:
@codingwangqi can you take a look at this?

codingwangqi:
@ivankelly @sijie I don't quite understand the question here. are you talking about LedgerLayout? or something else?

ivankelly:
@codingwangqi yes LedgerLayout. When a client or bookie writes and reads from a ledger manager they need to know the format. There's two parts to this. They need to know the layout of tree which they are reading, and they need to know the format of the contents of the nodes of the tree. In etcd, this is all implicit for now. Initially it was the same in ZK. Problems occur when you want to make a change. Some clients/bookies will understand the change while others will not (assuming not all client upgrade at the same instance), which can cause the clients to break each other. 

This is how it was with ZK. We went from a flat layout, with each ledger getting a znode under /ledgers to a hierarchical, where the ledger ID was segmented and the ledger metadata being stored in /ledgers/<XX>/<YY>/LXXYY. If a client assumed flat and added a ledger under /ledgers/LXXYY, stuff like garbage collection on the bookie (using hierarchical) would miss it and data would be lost.

LedgerLayout avoids this situation. When a client connects, it sees the layout which the client has been using and continues to use that layout, or gives up if it doesn't understand it. 

Recently I've extended this to include a version for the metadata contained within the nodes, which is how I spotted it was missing from etcd. Is anyone using etcd in production for this?

codingwangqi:
@ivankelly : 

There is a EtcdLayoutManager which is used for storing and retrieving LedgerLayout. However my understand is the ledger layout was used to handle different variants of zookeeper-based ledger managers (as what you said, from flat to hierarchical). 

However Etcd is a key-value system, instead of tree system. so we don't think we are going to change how we structure key/values in Etcd. Even we have to do so in future, we can add that later (treating layout doesn't exist as v0).

We are getting rid of all the baggages in zookeeper and start with a clean design for Etcd. 

> Is anyone using etcd in production for this?

We are using it in staging environment.

ivankelly:
@codingwangqi 

> However Etcd is a key-value system, instead of tree system. so we don't think we are going to change how we structure key/values in Etcd. Even we have to do so in future, we can add that later (treating layout doesn't exist as v0).

It's happening now. If you want to be able to use binary metadata with etcd you'll need to manage this.
[...]

PR #1865, opened 2018-12-05
Ledger manager factories initialized with max metadata version

ivankelly:
In dce4fd4 the max metadata format version was put in the
configuration which was then used to create ledger managers. This is
not safe though, as it there are multiple paths which through which
ledger manager factories are created and some may not see this
configuration modification. For example, on a new cluster where there
is no preexisting layout, the layout isn't written until after the
ledger manager factory has been created.

This patch changes LedgerManagerFactory#initialize to explicitly
require that the max ledger metadata format is specified in the call.

Master issue: #723

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
@@ -85,7 +85,6 @@ public abstract class AbstractConfiguration<T extends AbstractConfiguration>
     protected static final String STORE_SYSTEMTIME_AS_LEDGER_UNDERREPLICATED_MARK_TIME =
             "storeSystemTimeAsLedgerUnderreplicatedMarkTime";
     protected static final String STORE_SYSTEMTIME_AS_LEDGER_CREATION_TIME = "storeSystemTimeAsLedgerCreationTime";
-    protected static final String MAX_LEDGER_METADATA_FORMAT_VERSION = "maxLedgerMetadataFormatVersion";
 
     // Metastore settings, only being used when LEDGER_MANAGER_FACTORY_CLASS is MSLedgerManagerFactory
     protected static final String METASTORE_IMPL_CLASS = "metastoreImplClass";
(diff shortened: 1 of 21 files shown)


ivankelly:
Etcd Ledger Manager doesn't seem to support any layout versioning at all.I've created an issue (#1863 )

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
```diff
@@ -85,7 +85,6 @@
     protected static final String STORE_SYSTEMTIME_AS_LEDGER_UNDERREPLICATED_MARK_TIME =
             "storeSystemTimeAsLedgerUnderreplicatedMarkTime";
     protected static final String STORE_SYSTEMTIME_AS_LEDGER_CREATION_TIME = "storeSystemTimeAsLedgerCreationTime";
-    protected static final String MAX_LEDGER_METADATA_FORMAT_VERSION = "maxLedgerMetadataFormatVersion";
```
eolivelli: I miss one piece of this new story.
How does the use configure this field? In order to force an older version because of the presence of a mix of old and new clients in the cluster?

ivankelly:
rerun pr validation

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
eolivelli: But if I start a cluster with some v3 client, and then I need to add v2 clients (which cannot upgrade bk client library version) I will fail.
We should have a way to start a cluster and make it use v2 for some time and eventually update to v3.

I have clusters shared by different applications and they cannot be upgraded all together for instance

ivankelly: So, this is something that we need to decide when we push the PR which bumps the version. Should new clusters default to version 3 or default to version 2. I'm inclined to think they should default to version 2 for a couple of versions anyhow, but have version 3 available.

eolivelli: I can leave with default = v2 and optional version 3 by now.
We should have a story like the journal format version and storage version

PR #1866, opened 2018-12-06
Binary metadata format

ivankelly:
This patch adds a binary metadata format and bumps the metadata format
version to 3. The contents of the binary metadata is the same as the
contents of the text format for now. The difference is that the binary
is more compact, and the fields can be added to the metadata when
using the binary format, which isn't possible with the text
format. With the text format, parsing with a client that didn't
recognise the new field would fail.

For now, the text format (version 2) is still used by default. We will
provide a tool to allow administrators to bump to version 3.

Some tests have been modified to provide digest and password to the
builder. All protobuf metadata in released versions has had digest and
password (first protobuf metadata was in release-4.2.0). So if new
metadata is created or read with version 2, it will have this two
fields set.

Master issue: #723

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/LedgerMetadataSerDe.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/LedgerMetadataSerDe.java
@@ -183,11 +183,8 @@ public class LedgerMetadataSerDe {
 
             builder.setCtime(metadata.getCtime());
             builder.setDigestType(apiToProtoDigestType(metadata.getDigestType()));
-            if (metadata.getPassword() == null || metadata.getPassword().length == 0) {
-                builder.setPassword(ByteString.EMPTY);
-            } else {
-                builder.setPassword(ByteString.copyFrom(metadata.getPassword()));
-            }
+
+            serializePassword(metadata.getPassword(), builder);
 
             Map<String, byte[]> customMetadata = metadata.getCustomMetadata();
             if (customMetadata.size() > 0) {
@@ -252,11 +249,7 @@ public class LedgerMetadataSerDe {
                 }
 
                 builder.setDigestType(apiToProtoDigestType(metadata.getDigestType()));
-                if (metadata.getPassword() == null || metadata.getPassword().length == 0) {
-                    builder.setPassword(ByteString.EMPTY);
-                } else {
-                    builder.setPassword(ByteString.copyFrom(metadata.getPassword()));
-                }
+                serializePassword(metadata.getPassword(), builder);
 
                 Map<String, byte[]> customMetadata = metadata.getCustomMetadata();
                 if (customMetadata.size() > 0) {
@@ -317,6 +310,14 @@ public class LedgerMetadataSerDe {
                 throw new RuntimeException("UTF_8 should be supported everywhere");
             }
             return os.toByteArray();
+        }
+    }
+
+    private static void serializePassword(byte[] password, LedgerMetadataFormat.Builder builder) {
+        if (password == null || password.length == 0) {
+            builder.setPassword(ByteString.EMPTY);
+        } else {
+            builder.setPassword(ByteString.copyFrom(password));
         }
     }
 


bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/LedgerMetadataSerDe.java
```diff
@@ -142,15 +163,9 @@ public LedgerMetadataFormat buildProtoFormat(LedgerMetadata metadata) {
...
-            builder.setCtime(metadata.getCtime());
-        }
-
-        if (metadata.hasPassword()) {
-            builder.setDigestType(apiToProtoDigestType(metadata.getDigestType()))
-                .setPassword(ByteString.copyFrom(metadata.getPassword()));
-        }
+        builder.setCtime(metadata.getCtime());
+        builder.setDigestType(apiToProtoDigestType(metadata.getDigestType()))
+            .setPassword(ByteString.copyFrom(metadata.getPassword()));
```
merlimat: If password is empty, we could use `ByteString.EMPTY`

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/LedgerMetadataSerDe.java
```diff
@@ -66,59 +76,70 @@
...
     }
 
-    private byte[] serializeVersion1(LedgerMetadata metadata) {
-        StringBuilder s = new StringBuilder();
-        s.append(VERSION_KEY).append(FIELD_SPLITTER)
-            .append(metadata.getMetadataFormatVersion()).append(LINE_SPLITTER);
-        s.append(metadata.getWriteQuorumSize()).append(LINE_SPLITTER)
-            .append(metadata.getEnsembleSize()).append(LINE_SPLITTER).append(metadata.getLength());
+    private static int readHeader(InputStream is) throws IOException {
+        checkState(LINE_SPLITTER_BYTES.length == 1, "LINE_SPLITTER must be single byte");
```
eolivelli: This should be in a static{}  block.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/LedgerMetadataSerDe.java
```diff
@@ -66,59 +76,70 @@
...
-
-        return s.toString().getBytes(UTF_8);
+        return serialized;
     }
 
-    @VisibleForTesting
-    public LedgerMetadataFormat buildProtoFormat(LedgerMetadata metadata) {
+    private static byte[] serializeVersion3(LedgerMetadata metadata) throws IOException {
+        ByteArrayOutputStream os = new ByteArrayOutputStream();
+        writeHeader(os, 3);
```
eolivelli: Use constant for '3' ?

ivankelly: what would you name such a constant?

eolivelli: Isn't it MAXIMUM_METADATA_FORMAT_VERSION ?

Or we can have 
METADATA_FORMAT_VERSION_3

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractZkLedgerManager.java
```diff
@@ -266,9 +266,17 @@ public void processResult(int rc, String path, Object ctx, String name) {
                 }
             }
         };
+        final byte[] data;
+        try {
+            data = serDe.serialize(metadata);
+        } catch (IOException ioe) {
+            promise.completeExceptionally(new BKException.ZKException(ioe));
```
sijie: I would suggest defining a new error code for this type of exception. `BKException.ZKException` is a bit confusing.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractZkLedgerManager.java
```diff
@@ -423,8 +431,16 @@ public void processResult(int rc, String path, Object ctx, byte[] data, Stat sta
             return promise;
         }
         final LongVersion zv = (LongVersion) currentVersion;
+
+        final byte[] data;
+        try {
+            data = serDe.serialize(metadata);
+        } catch (IOException ioe) {
```
sijie: same comment as above

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/LedgerMetadataSerDe.java
```diff
@@ -66,59 +76,70 @@
...
+            }
         }
+        throw new IOException("Unable to find end of version number, metadata appears corrupt");
+    }
 
+    public byte[] serialize(LedgerMetadata metadata) throws IOException {
+        int formatVersion = Math.min(maxLedgerMetadataFormatVersion, metadata.getMetadataFormatVersion());
+        final byte[] serialized;
+        switch (formatVersion) {
+        case 3:
```
sijie: can you define constants for these versions?

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/LedgerMetadataSerDe.java
```diff
@@ -187,55 +301,61 @@ public LedgerMetadataFormat buildProtoFormat(LedgerMetadata metadata) {
...
+        try (ByteArrayInputStream is = new ByteArrayInputStream(bytes)) {
+            int metadataFormatVersion = readHeader(is);
+            if (log.isDebugEnabled()) {
+                log.debug("Format version {} detected", metadataFormatVersion);
+            }
 
-        if (metadataFormatVersion == 1) {
-            return parseVersion1Config(reader);
+            switch (metadataFormatVersion) {
+            case 3:
```
sijie: constants

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/LedgerMetadataSerDe.java
ivankelly: I had it in static{} originally. Will do so again.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/LedgerMetadataSerDe.java
ivankelly: Uncomfortably close to ```#define THREE=3```. Will add them all the same.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractZkLedgerManager.java
ivankelly: Will add

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/LedgerMetadataSerDe.java
ivankelly: done

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/LedgerMetadataSerDe.java
ivankelly: done

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/LedgerMetadataSerDe.java
ivankelly: done

sijie:
@ivankelly are we still targeting this for 4.9? if not, I would move it out of 4.9 to unblock the release.

ivankelly:
@sijie this can go in now that the ledger layout change has been removed. It'll sit dormant in the tree but it will still be possible to read binary metadata if it is enabled in the future. I need to rework for the layout change though. Will push something later.

ivankelly:
rerun integration tests

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/BKException.java
```diff
@@ -271,6 +273,13 @@ public static String getMessage(int code) {
          */
         int LedgerIdOverflowException = -106;
 
+        /**
+         * Failure to serialize/deserialize.
+         *
+         * @since 4.9
+         */
+        int SerializationException = -107;
```
sijie: SerializationException seems to be a very generic exception. Shall we call it a "MetadataCorruptedException" indicating it is a corrupted metadata?

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/LedgerMetadataSerDe.java
```diff
@@ -51,68 +58,100 @@
...
+            break;
+        case METADATA_FORMAT_VERSION_1:
+            serialized = serializeVersion1(metadata);
+            break;
+        default:
+            throw new IllegalArgumentException("Invalid format version " + formatVersion);
+        }
         if (log.isDebugEnabled()) {
-            log.debug("Serialized config: {}", s);
+            log.debug("Serialized with format {}: {}", formatVersion, Base64.getEncoder().encodeToString(serialized));
```
sijie: how does `encodeToString` work for v1 and v2 metadata? since v1 and v2 are already printable strings.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/LedgerMetadataSerDe.java
```diff
@@ -164,9 +201,112 @@ public LedgerMetadataFormat buildProtoFormat(LedgerMetadata metadata) {
...
             builder.addSegment(segmentBuilder.build());
         }
-        return builder.build();
+
+        builder.build().writeDelimitedTo(os);
+        return os.toByteArray();
+    }
+
+    private static byte[] serializeVersion2(LedgerMetadata metadata) throws IOException {
+        ByteArrayOutputStream os = new ByteArrayOutputStream();
```
sijie: don't we need to put os in a try-with-resource block to release os at the end?

merlimat:
> It'll sit dormant in the tree but it will still be possible to read binary metadata if it is enabled in the future.

It would be nice to have this in 4.9, so that enabling binary format in 4.10 will be easier (while maintaining rollback capability)

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/BKException.java
ivankelly: renamed to metadata serialization exception. it's not for corrupted metadata. This exception occurs when there's and exception in the call to #serialize, which can really only be an IOException when doing the protobuf binary encoding, which should never happen as the IO is all on memory.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/LedgerMetadataSerDe.java
ivankelly: Changed to print string if version <= VERSION_2

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/LedgerMetadataSerDe.java
ivankelly: added, but not strictly needed. there's no resource held here.

ivankelly:
@sijie can you take another look?

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/LedgerMetadataSerDe.java
```diff
@@ -51,122 +58,267 @@
...
+        try (ByteArrayOutputStream os = new ByteArrayOutputStream()) {
+            writeHeader(os, METADATA_FORMAT_VERSION_3);
+            LedgerMetadataFormat.Builder builder = LedgerMetadataFormat.newBuilder();
+            builder.setQuorumSize(metadata.getWriteQuorumSize())
+                .setAckQuorumSize(metadata.getAckQuorumSize())
+                .setEnsembleSize(metadata.getEnsembleSize())
+                .setLength(metadata.getLength())
+                .setLastEntryId(metadata.getLastEntryId());
+
+            switch (metadata.getState()) {
```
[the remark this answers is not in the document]
ivankelly: All these duplicates are on purpose. For example, if we add a state in future, but the max metadata version is 2, and we are using shared code, it will happily write the new state to the metadata and break clients that can only read V2. It's ugly, but it's here defensively.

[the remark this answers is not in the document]
sijie: For ‘state’, it is probably the case. However I am not sure about the other comments.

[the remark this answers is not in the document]
ivankelly: A field could be added to LedgerMetadataFormat.Segment or LedgerMetadataFormat.cMetadataMapEntry, which could cause the same problem. Will factor out password though.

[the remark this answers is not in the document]
sijie: ack

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/LedgerMetadataSerDe.java
```diff
@@ -181,55 +333,65 @@ public LedgerMetadataFormat buildProtoFormat(LedgerMetadata metadata) {
...
-            reader = new BufferedReader(new StringReader(config));
-        }
-
-        if (metadataFormatVersion < LOWEST_COMPAT_METADATA_FORMAT_VERSION
-            || metadataFormatVersion > CURRENT_METADATA_FORMAT_VERSION) {
-            throw new IOException(
-                    String.format("Metadata version not compatible. Expected between %d and %d, but got %d",
-                                  LOWEST_COMPAT_METADATA_FORMAT_VERSION, CURRENT_METADATA_FORMAT_VERSION,
-                                  metadataFormatVersion));
+            log.debug("Deserializing {}", Base64.getEncoder().encodeToString(bytes));
```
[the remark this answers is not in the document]
sijie: @ivankelly I saw your refactored the password part change. just wondering are you addressing this comment or not?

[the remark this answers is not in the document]
ivankelly: Yes, i did, but not here. Before starting to deserialize, it's impossible to know the version, so I log the base64 here (for cases where it's entirely corrupt so would blow up before reading the version), and then if the version is <= 2, I print the string version on line 346.

PR #1883, opened 2018-12-13
Enhance EnsemblePlacementPolicy and DNSResolverDecorator to log relevant metrics.

reddycharan:

Make changes to EnsemblePlacementPolicy so that it would return boolean value indicating
if the return value of newEnsemble and replaceBookie are strictly adhering to corresponding
PlacementPolicy or it fell back to random.

Similarly DNSResolverDecorator should log a metric when it was unable to resolve rack info and
it is using default rack.


reddycharan:
hey @sijie @athanatos before I get into full implementation, I would like to get your feedback for this change.

sijie:
@reddycharan looks good to me

athanatos:
I don't really get why isEnsembleFellbackToRandom needs to work like that.  Can't we just propagate that information from where the fallback actually happens?

Also, probably better create an object with named fields rather than using a Pair.

reddycharan:
@sijie @athanatos this is ready for review.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RegionAwareEnsemblePlacementPolicy.java
```diff
@@ -550,4 +564,14 @@ protected BookieNode replaceFromRack(BookieNode bookieNodeToReplace,
         finalList.addMissingIndices(ensemble.size());
         return finalList;
     }
+
+    @Override
+    public boolean isEnsembleAdheringToPlacementPolicy(List<BookieSocketAddress> ensembleList, int writeQuorumSize,
+            int ackQuorumSize) {
+        /**
+         * TODO: have to implement actual logic for this method for
```
reddycharan: will create an Issue to track this.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookKeeperServerStats.java
```diff
@@ -69,6 +69,11 @@
     String WATCHER_SCOPE = "bookie_watcher";
     String REPLACE_BOOKIE_TIME = "REPLACE_BOOKIE_TIME";
     String NEW_ENSEMBLE_TIME = "NEW_ENSEMBLE_TIME";
+    String FAILED_TO_RESOLVE_NETWORK_LOCATION_COUNTER = "FAILED_TO_RESOLVE_NETWORK_LOCATION_COUNTER";
+    String NEW_ENSEMBLE_NOT_ADHERING_TO_PLACEMENT_POLICY_COUNTER =
+            "NEW_ENSEMBLE_NOT_ADHERING_TO_PLACEMENT_POLICY_COUNTER";
+    String REPLACE_BOOKIE_NOT_ADHERING_TO_PLACEMENT_POLICY_COUNTER =
+            "REPLACE_BOOKIE_NOT_ADHERING_TO_PLACEMENT_POLICY_COUNTER";
```
jvrao: What is the benefit of tracking these things separately? Do you plan to act differently? If not, I would make it one instead of cluttering further.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookieWatcherImpl.java
```diff
@@ -213,19 +223,33 @@ public void initialBlockingBookieRead() throws BKException {
...
-            socketAddresses = placementPolicy.newEnsemble(ensembleSize,
+            newEnsembleResponse = placementPolicy.newEnsemble(ensembleSize,
                     writeQuorumSize, ackQuorumSize, customMetadata, new HashSet<BookieSocketAddress>(
                             quarantinedBookies.asMap().keySet()));
+            socketAddresses = newEnsembleResponse.getLeft();
+            isEnsembleAdheringToPlacementPolicy = newEnsembleResponse.getRight();
+            if (!isEnsembleAdheringToPlacementPolicy) {
+                newEnsembleNotAdheringToPlacementPolicy.inc();
+                log.warn("New ensemble: {} is not adhering to Placement Policy", socketAddresses);
+            }
```
jvrao: I guess I am confused. Why do we need to push up these metrics? why can't we do it at the lowest level? i.e placementPolicy.newEnsemble itself can log/update the stat. What is the point in pushing this up?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookieWatcherImpl.java
```diff
@@ -239,22 +263,40 @@ public BookieSocketAddress replaceBookie(int ensembleSize, int writeQuorumSize,
...
         BookieSocketAddress addr = existingBookies.get(bookieIdx);
+        Pair<BookieSocketAddress, Boolean> replaceBookieResponse;
         BookieSocketAddress socketAddress;
+        boolean isEnsembleAdheringToPlacementPolicy = false;
         try {
             // we exclude the quarantined bookies also first
-            Set<BookieSocketAddress> existingAndQuarantinedBookies = new HashSet<BookieSocketAddress>(existingBookies);
-            existingAndQuarantinedBookies.addAll(quarantinedBookies.asMap().keySet());
-            socketAddress = placementPolicy.replaceBookie(
+            Set<BookieSocketAddress> excludeBookiesAndQuarantinedBookies = new HashSet<BookieSocketAddress>(
```
jvrao: you mean 'exclude[d]' bookies?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookieWatcherImpl.java
```diff
@@ -88,6 +92,8 @@
         help = "operation stats of replacing bookie in an ensemble"
     )
     private final OpStatsLogger replaceBookieTimer;
+    private final Counter newEnsembleNotAdheringToPlacementPolicy;
```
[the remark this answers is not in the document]
reddycharan: will do

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/EnsemblePlacementPolicy.java
```diff
@@ -351,4 +352,23 @@ BookieSocketAddress replaceBookie(int ensembleSize,
...
+     *
+     * @param ensembleList
+     *            list of BookieSocketAddress of bookies in the ensemble
+     * @param writeQuorumSize
+     *            writeQuorumSize of the ensemble
+     * @param ackQuorumSize
+     *            ackQuorumSize of the ensemble
+     * @return
+     */
+    boolean isEnsembleAdheringToPlacementPolicy(List<BookieSocketAddress> ensembleList, int writeQuorumSize,
```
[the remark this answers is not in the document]
reddycharan: will do

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookKeeperServerStats.java
reddycharan: will make it one

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookieWatcherImpl.java
reddycharan: I would see methods of BookieWatcher / BookieWatcherImpl - newEnsemble, replaceBookie as the interface methods for BookKeeper client operations and it is appropriate to deal with metrics and log lines here rather than in EnsemblePlacementPolicy classes for few reasons

- first BookieWatcher / BookieWatcherImpl methods have more context, so having log line with appropriate info. along with metric statement would be helpful in diagnosing/debugging any noticed issue.
- it would make things simpler and easier to have metric and log line in these methods instead of having similar metric and log line in all of the EnsemblePlacementPolicy implementations and in several overloaded methods in EnsemblePlacementPolicy implementation classes.
- moreover I don't see advantage in moving these metrics and log lines down the stack, especially considering the complexity of each EnsemblePlacementPolicy implementations

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookieWatcherImpl.java
reddycharan: followed existing terminology - 'Set<BookieSocketAddress> excludeBookies', anyhow dont have strong opinion, will change

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
```diff
@@ -1206,4 +1240,32 @@ static void shuffleWithMask(DistributionSchedule.WriteSet writeSet,
...
+    public boolean isEnsembleAdheringToPlacementPolicy(List<BookieSocketAddress> ensembleList, int writeQuorumSize,
+            int ackQuorumSize) {
+        int ensembleSize = ensembleList.size();
+        int minNumRacksPerWriteQuorumForThisEnsemble = Math.min(writeQuorumSize, minNumRacksPerWriteQuorum);
+        HashSet<String> racksOrRegionsInQuorum = new HashSet<String>();
+        BookieSocketAddress bookie;
+        for (int i = 0; i < ensembleList.size(); i++) {
+            racksOrRegionsInQuorum.clear();
+            for (int j = 0; j < writeQuorumSize; j++) {
+                bookie = ensembleList.get((ensembleSize + i + j) % ensembleSize);
```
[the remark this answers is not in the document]
reddycharan: not really if 'i' and 'j' are positive numbers, will remove it from numerator to avoid confusion

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookieWatcherImpl.java
```diff
@@ -88,6 +92,8 @@
...
+    private final Counter replaceBookieNotAdheringToPlacementPolicy;
```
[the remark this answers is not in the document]
reddycharan: done

athanatos:
LGTM

PR #1888, opened 2018-12-19
Bring back statslogger to BookKeeper client in ReplicationWorker process.

reddycharan:

- https://github.com/apache/bookkeeper/commit/2837f6257baf15dc9dd9eb4bcac34596b442be33
had inadvertently removed StatsLogger to BookKeeper client instance in ReplicationWorker process.
So restore StatsLogger in BookKeeper client object.
- Also introduce new config called - 'limitStatsLogging', which would be used to limit statslogging
as and when needed.
- currently this config is used to limit the stats from PCBC. Because if AR process is running in
each Bookie node, then for each AR there will be n number of PCBCs and totally it comes out to
n^2 PCBCs in the cluster. Which is unmanageable from stats collector perspective. So this config
value can be set to true in AR config.

### DIFF
--- a/site/_data/config/bk_server.yaml
+++ b/site/_data/config/bk_server.yaml
@@ -538,6 +538,9 @@ groups:
         - Twitter Ostrich   : org.apache.bookkeeper.stats.twitter.ostrich.OstrichProvider
         - Twitter Science   : org.apache.bookkeeper.stats.twitter.science.TwitterStatsProvider
     default: org.apache.bookkeeper.stats.prometheus.PrometheusMetricsProvider
+  - param: limitStatsLogging
+    description: option to limit stats logging
+    default: 'false'
 
 - name: Prometheus Metrics Provider Settings
   params:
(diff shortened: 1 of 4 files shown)


reddycharan:
@pasha-kuznetsov fyi.

site/_data/config/bk_server.yaml
```diff
@@ -538,6 +538,9 @@ groups:
         - Twitter Ostrich   : org.apache.bookkeeper.stats.twitter.ostrich.OstrichProvider
         - Twitter Science   : org.apache.bookkeeper.stats.twitter.science.TwitterStatsProvider
     default: org.apache.bookkeeper.stats.prometheus.PrometheusMetricsProvider
+  - param: limitStatsLogging
+    description: option to limit stats logging
```
eolivelli: Honestly from this description I can't understand what 'stats logging' really means in this context.
It seems that we are disabling stats at all

@sijie @reddycharan

Issue #1712, opened 2018-09-27
Potential for TOAB violation with multiple calls to close


ivankelly:
I came across this when doing the immutable work. It exists in both pre and post immutable implementations.

The problem occurs if a client calls close twice concurrently with some writes outstanding. Take the local LAC in the writer to be 10, and 11-20 are outstanding writes. In this case, in the pre-immutable world, the first close will do the actual work, update the state in the metadata to closed and try to do the write to metadata with lastEntry as 10. The second local call to close sees that state is closed, so returns immediately, so the caller thinks that the ledger has been successfully closed with 10 as the last entry. However, if another client recovers the ledger, and sets the lastEntry to 20, then we have a potential for a TOAB violation.

The solution would be to have 3 handle states, OPEN, CLOSING, and CLOSED, and when the state is CLOSING, only wait for the outstanding close operation to complete before returning to the client. It's out of scope for this change though.

sijie:
Can you provide a repo for this? 

> However, if another client recovers the ledger, and sets the lastEntry to 20, then we have a potential for a TOAB violation.

I don't fully understand this sentence here. 

from what I understand, when another client recovers the ledger, if it sees the ledger is closed, it does nothing; if the recovery starts before the first client updated the metadata, the second recovery will set the metadata to IN_RECOVERY, the first client will see metadata version and attempt to resolve the conflict. when resolving the conflicts, there should be one client successfully close the ledger and the other one fails. the current logic give up at closing when hitting metadata version conflicts, so there will not be any metadata overwritten. so I am not sure how this would happen in current code though.


regarding the state, we already have a state called IN_RECOVERY which is the "CLOSING" in some sense. Can't we reuse that?

ivankelly:
I don't have a repro yet.

But this is local handle state, not remote metadata state. AFAICS the remote metadata state handling is safe.

The core of the problem is that a second call to #close on a handle, will return successfully before the first one is even finished. If that thread then takes any action based on what it thinks the state of the ledger is, it risks being wrong, since the first call to #close could fail.

sijie:
okay now the problem makes sense to me.

ivankelly:
I don't have a repro yet.

But this is local handle state, not remote metadata state. AFAICS the remote metadata state handling is safe.

The core of the problem is that a second call to #close on a handle, will return successfully before the first one is even finished. If that thread then takes any action based on what it thinks the state of the ledger is, it risks being wrong, since the first call to #close could fail.

[...]

PR #1889, opened 2018-12-19
Multiple calls to LedgerHandle#close should wait on actual close

ivankelly:
The previous behaviour was to complete successfully immediately if
close had already been called on the handle. This allowed for
potential consistency violations, as the caller could believe that the
ledger was closed, when in fact the close operation was still in
progress and could still potentially fail.

Issue: #1712

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
@@ -106,7 +106,7 @@ public class LedgerHandle implements WriteHandle {
     };
 
     private HandleState handleState = HandleState.OPEN;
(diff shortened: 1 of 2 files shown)


ivankelly:
run integration tests

ivankelly:
run pr validation

ivankelly:
run pr validation

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -106,6 +106,7 @@
     };
 
     private HandleState handleState = HandleState.OPEN;
+    private CompletableFuture<Void> closePromise = new CompletableFuture<>();
```
sijie: final?

PR #1902, opened 2019-01-14
Metadata checker validating EnsemblePlacementpolicy

reddycharan:

- As described in BP-34,
https://github.com/apache/bookkeeper/blob/master/site/bps/BP-34-cluster-metadata-checker.md,
this change introduces new checker - cluster-metadata-checker with initial responsibility
of validating the EnsemblePlacementpolicy of segments of ledgers (which are writeclosed/fenced).


reddycharan:
created WIP in change. Before adding testcases, I just want to make sure you guys don't have any major issue with this implementation.

reddycharan:
@sijie @athanatos can you take a quick look to see if you are fine with the code layout.

bookkeeper-proto/src/main/proto/DataFormats.proto
```diff
@@ -102,3 +102,10 @@ message AuditorVoteFormat {
 message CheckAllLedgersFormat {
     optional int64 checkAllLedgersCTime = 1;
 }
+
+/**
+ * information of MetadataCheck execution
+ */
+message MetadataCheckFormat {
+    optional int64 metadataCheckCTime = 1;
+}
```
jvrao: Hmm format issue at the end of the curly brace?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperAdmin.java
```diff
@@ -1643,4 +1643,24 @@ private static boolean areEntriesOfSegmentStoredInTheBookie(LedgerMetadata ledge
...
         return firstStoredEntryId != LedgerHandle.INVALID_ENTRY_ID;
     }
+
+    /**
+     * returns boolean value specifying if the ensemble of the segment is
+     * adhering to the ensemble placement policy for the given writeQuorumSize
+     * and ackQuorumSize.
+     *
+     * @param ensembleList
+     *            ensemble of the segment
```
jvrao: comment reads as if it is one segment. List of ensembles of the ledger or something

jvrao: ensembleList confused me as if it is a list of ensembles. May be you should consider renaming it to ensembleBookiesList or something

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
```diff
@@ -2188,6 +2189,29 @@ public long getAuditorPeriodicBookieCheckInterval() {
         return getLong(AUDITOR_PERIODIC_BOOKIE_CHECK_INTERVAL, 86400);
     }
 
+    /**
+     * Set the regularity at which the auditor will run a metadata check of all
```
jvrao: s/regularity/interval

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
```diff
@@ -2188,6 +2189,29 @@ public long getAuditorPeriodicBookieCheckInterval() {
...
+     * write closed/fenced ledgers. This should not be run very often, and at
```
jvrao: No need to say write closed -> as there is no read open / close ledger state. 
Also fenced ledger may not have final consistent state yet. Why do you need to include
fenced ledgers, it is just a transient state right?

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
```diff
@@ -2188,6 +2189,29 @@ public long getAuditorPeriodicBookieCheckInterval() {
...
+     * most, once a day. Setting this to 0 will completely disable the periodic
```
jvrao: s/and most/ at most.

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
```diff
@@ -2188,6 +2189,29 @@ public long getAuditorPeriodicBookieCheckInterval() {
...
+     * @param interval
+     *            The interval in seconds. e.g. 86400 = 1 day, 604800 = 1 week
+     */
+    public void setAuditorPeriodicMetadataCheckInterval(long interval) {
+        setProperty(AUDITOR_PERIODIC_METADATA_CHECK_INTERVAL, interval);
+    }
+
+    /**
+     * Get the regularity at which the auditor does metadata check of all write
+     * closed/fenced ledgers.
```
jvrao: Same comment as above.

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
```diff
@@ -161,6 +174,12 @@
         help = "the number of delayed-bookie-audits cancelled"
     )
     private final Counter numDelayedBookieAuditsCancelled;
+    @StatsDoc(
+        name = METADATA_CHECK_ENSEMBLE_NOT_ADHERING_TO_PLACEMENT_POLICY_COUNTER,
+        help = "total number of "
+            + "segments of ledgers failed to adhere to EnsemblePlacementPolicy found in metadata check"
+    )
```
jvrao: What is the advantage we are getting counting each segment rather than one count per ledger? Just thinking aloud here. If we count by segment, we have no way of knowing what percentage of the store is wack. @reddycharan  do you see any advantage of counting segments instead of just one increment per ledger?

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
```diff
@@ -466,80 +487,141 @@ public void start() {
...
+                    } catch (ReplicationException.UnavailableException ue) {
+                        LOG.error("Underreplication manager unavailable running periodic check", ue);
                     }
-                    }, initialDelay, interval, TimeUnit.SECONDS);
+                }
+                }, initialDelay, interval, TimeUnit.SECONDS);
+        } else {
+            LOG.info("Periodic checking disabled");
+        }
+    }
```
jvrao: @reddycharan I am guessing above two routines are just refactoring. Let me know if they need a close look for logic change.

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
```diff
@@ -466,80 +487,141 @@ public void start() {
...
+
+            executor.scheduleAtFixedRate(new Runnable() {
+                public void run() {
+                    try {
+                        Stopwatch stopwatch = Stopwatch.createStarted();
+                        LOG.info("Starting MetadataCheck");
+                        metadataCheck();
+                        long metadataCheckDuration = stopwatch.stop().elapsed(TimeUnit.MILLISECONDS);
+                        LOG.info("Completed metadataCheck in {} milliSeconds", metadataCheckDuration);
+                        checkAllLedgersTime.registerSuccessfulEvent(metadataCheckDuration, TimeUnit.MILLISECONDS);
```
jvrao: checkAllLedgersTime?? Cut and Paste issue?

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
```diff
@@ -466,80 +487,141 @@ public void start() {
...
+                            metadataCheckLastExecutedCTime, durationSinceLastExecutionInSecs, initialDelay, interval);
+
+            executor.scheduleAtFixedRate(new Runnable() {
+                public void run() {
+                    try {
+                        Stopwatch stopwatch = Stopwatch.createStarted();
+                        LOG.info("Starting MetadataCheck");
+                        metadataCheck();
+                        long metadataCheckDuration = stopwatch.stop().elapsed(TimeUnit.MILLISECONDS);
+                        LOG.info("Completed metadataCheck in {} milliSeconds", metadataCheckDuration);
```
jvrao: Log message should have the time stamp right? Anyway it won't hurt to have the duration...

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
```diff
@@ -466,80 +487,141 @@ public void start() {
...
+                    "metadataCheck scheduling info.  metadataCheckLastExecutedCTime: {} "
+                            + "durationSinceLastExecutionInSecs: {} initialDelay: {} interval: {}",
+                            metadataCheckLastExecutedCTime, durationSinceLastExecutionInSecs, initialDelay, interval);
+
+            executor.scheduleAtFixedRate(new Runnable() {
+                public void run() {
+                    try {
+                        Stopwatch stopwatch = Stopwatch.createStarted();
+                        LOG.info("Starting MetadataCheck");
+                        metadataCheck();
```
jvrao: How do you know if the previous instance of the metadataCheck is running or not? I guess we need to make sure that we have only one instance running at any given time.

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
jvrao: Open a visibility item to monitor/alert on this if not running.

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
```diff
@@ -803,8 +885,72 @@ void checkAllLedgers() throws BKException, IOException, InterruptedException, Ke
...
+        final CountDownLatch metadataCheckLatch = new CountDownLatch(1);
+        Processor<Long> ledgerProcessor = new Processor<Long>() {
+            @Override
+            public void process(Long ledgerId, AsyncCallback.VoidCallback iterCallback) {
+                ledgerManager.readLedgerMetadata(ledgerId).whenComplete((metadataVer, exception) -> {
+                    if (exception == null) {
+                        LedgerMetadata metadata = metadataVer.getValue();
+                        int writeQuorumSize = metadata.getWriteQuorumSize();
+                        int ackQuorumSize = metadata.getAckQuorumSize();
+                        if (metadata.isClosed()) {
```
jvrao: Ah good you are just checking closed ledgers; the above (fence) is just a comment issue; please take care of the comment.

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
```diff
@@ -803,8 +885,72 @@ void checkAllLedgers() throws BKException, IOException, InterruptedException, Ke
...
+        ledgerManager.asyncProcessLedgers(ledgerProcessor, new AsyncCallback.VoidCallback() {
+
+            @Override
+            public void processResult(int rc, String s, Object obj) {
+                resultCode.add(rc);
+                metadataCheckLatch.countDown();
+            }
+        }, null, BKException.Code.OK, BKException.Code.ReadException);
+        try {
+            metadataCheckLatch.await();
```
jvrao: Some debugging around this wait will be good; I am worried about indefinite wait scenarios.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperAdmin.java
```diff
@@ -1643,4 +1643,24 @@ private static boolean areEntriesOfSegmentStoredInTheBookie(LedgerMetadata ledge
...
+     *            writeQuorumSize of the ledger
+     * @param ackQuorumSize
+     *            ackQuorumSize of the ledger
+     * @return <tt>true</tt> if the ledger is adhering to
+     *         EnsemblePlacementPolicy
+     */
+    public boolean isEnsembleAdheringToPlacementPolicy(List<BookieSocketAddress> ensembleList, int writeQuorumSize,
+            int ackQuorumSize) {
+        return bkc.getPlacementPolicy().isEnsembleAdheringToPlacementPolicy(ensembleList, writeQuorumSize,
+                ackQuorumSize);
```
jvrao: I might be missing something; I don't see where this isEnsembleAdheringToPlacementPolicy implemented

reddycharan: it is implemented in each EnsemblePlacementPolicy

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
reddycharan: actually even I'm concerned about this indefinite wait. It is indefinite wait for other checks as well. Have to think about it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
reddycharan: ok will change comment to 'closed ledgers'

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
reddycharan: log line will inherently has timestamp

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
reddycharan: yep copy error..
yep it needs monitoring/alert

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
reddycharan: yes it is simple refactoring, didn't change anything

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
reddycharan: by definition ensemble of bookies is associated to segment, so I opted for per segment. Being said that I see point in your concern. Probably its best to log for each segment failure and have metric per ledger.

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
```diff
@@ -2188,6 +2189,29 @@ public long getAuditorPeriodicBookieCheckInterval() {
         return getLong(AUDITOR_PERIODIC_BOOKIE_CHECK_INTERVAL, 86400);
     }
 
+    /**
+     * Set the regularity at which the auditor will run a metadata check of all
+     * write closed/fenced ledgers. This should not be run very often, and at
+     * most, once a day. Setting this to 0 will completely disable the periodic
+     * metadata check.
```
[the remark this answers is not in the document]
reddycharan: yes this metadatacheck will have two parts. this is just first part. I'm planning to do both of these checks in the same scrutiny. So there will be only one config variable. So I'm not planning to separate these two scrutinies and have two different config variables.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperAdmin.java
```diff
@@ -1643,4 +1643,24 @@ private static boolean areEntriesOfSegmentStoredInTheBookie(LedgerMetadata ledge
...
+     * @param ensembleList
+     *            ensemble of the segment
+     * @param writeQuorumSize
+     *            writeQuorumSize of the ledger
+     * @param ackQuorumSize
+     *            ackQuorumSize of the ledger
+     * @return <tt>true</tt> if the ledger is adhering to
+     *         EnsemblePlacementPolicy
+     */
+    public boolean isEnsembleAdheringToPlacementPolicy(List<BookieSocketAddress> ensembleList, int writeQuorumSize,
```
eolivelli: What about using the word 'Consistent' instead of 'Adhering' ? 

Cc @jvrao

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
```diff
@@ -743,11 +825,11 @@ public void operationComplete(int rc, Set<LedgerFragment> fragments) {
      * be run very often.
      */
     void checkAllLedgers() throws BKException, IOException, InterruptedException, KeeperException {
-        final BookKeeper client = createBookKeeperClient(conf);
-        final BookKeeperAdmin admin = new BookKeeperAdmin(client, statsLogger);
+        final BookKeeper localClient = createBookKeeperClient(conf);
```
eolivelli: Better to use try-with-resources otherwise you could fail to close the client if the admin close fails

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
eolivelli: We could put an upperbound to the ovetall processing.
The downside is that ledgers with greater id could starve

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
reddycharan: yeah..but it is existing code, probably it would be better to do it in different commit. Otherwise it shows as if complete code is changed.

bookkeeper-proto/src/main/proto/DataFormats.proto
reddycharan: added new line

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperAdmin.java
reddycharan: changed

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperAdmin.java
reddycharan: isEnsembleAdheringToPlacementPolicy method name is already used in EnsemblePlacementPolicy, so continuing to use it. Also, 'adhere' sounds more appropriate.

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
reddycharan: changed it to "all ledgers which are closed"

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
reddycharan: changed it

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
```diff
@@ -20,10 +20,14 @@
  */
 package org.apache.bookkeeper.replication;
 
+
 import static org.apache.bookkeeper.replication.ReplicationStats.AUDITOR_SCOPE;
 import static org.apache.bookkeeper.replication.ReplicationStats.AUDIT_BOOKIES_TIME;
 import static org.apache.bookkeeper.replication.ReplicationStats.BOOKIE_TO_LEDGERS_MAP_CREATION_TIME;
 import static org.apache.bookkeeper.replication.ReplicationStats.CHECK_ALL_LEDGERS_TIME;
+import static org.apache.bookkeeper.replication.ReplicationStats.
```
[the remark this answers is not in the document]
reddycharan: hmm..otherwise it exceeds 120 characters line limit.

eolivelli:
Thanks for your replies Reddy.

I will approve this path once it is no more WIP

reddycharan:
@eolivelli @jvrao @sijie requesting review. rebased to master.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperAdmin.java
jvrao: Is that not part of this pull request?

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
eolivelli: Ok

bookkeeper-server/src/test/java/org/apache/bookkeeper/replication/AuditorPeriodicCheckTest.java
```diff
@@ -454,7 +457,100 @@ void validateInitialDelayOfCheckAllLedgers(LedgerUnderreplicationManager urm, lo
...
+
+        urm.setMetadataCheckCTime(metadataCheckCTime);
+        auditor.start();
+        /*
+         * since auditorPeriodicMetadataCheckInterval are higher values (in the
+         * order of 100s of seconds), its ok bufferTimeInMsecs to be ` 10 secs.
+         */
+        assertTrue("metadataCheck should have executed with initialDelay " + initialDelayInMsecs,
+                latch.await(initialDelayInMsecs + bufferTimeInMsecs, TimeUnit.MILLISECONDS));
+        for (int i = 0; i < 10; i++) {
```
eolivelli: This would be a new flaky test on slow machines
Maybe it is better to have a longer timeout

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
eolivelli: This series of comments seems not addressed

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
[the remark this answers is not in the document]
reddycharan: @eolivelli @sijie 

in this bp - https://github.com/apache/bookkeeper/blob/master/site/bps/BP-34-cluster-metadata-checker.md I mentioned that there would be one single checker which would validate placement policy, underreplicated state and availability of entries in bookies. But @jvrao is of the opinion that since PlacementPolicy validation would be light weight it makes sense to run more frequently and have different config intervals for placement policy checker and availability of entries checker. So I'm planning to split this into two different checkers, this one would be placement policy checker and the other (yet to be implemented) one would be for availability of entries (and underreplication state) checker. So I'll rename this metadatachecker to placementpolicychecker and other corresponding method names and config variables.

eolivelli:
Okay let's split

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
reddycharan: @eolivelli @jvrao @sijie i just want to make sure we are on the same page for this.

reddycharan:
@sijie @eolivelli @jvrao can you please take a look. I would like to merge once I get approval from atleast couple of you guys.

site/_data/config/bk_server.yaml
```diff
@@ -652,6 +652,9 @@ groups:
 
       Set this to 0 to disable the periodic check completely. Note that periodic checking will put extra load on the cluster, so it should not be run more frequently than once a day.
     default: 604800
+  - param: auditorPeriodicPlacementPolicyCheckInterval
+    description: The time interval between auditor placement policy checks, in seconds. The auditor placement policy check validates if the ensemble of segments of all the closed ledgers is adhering to the placement policy. It is just monitoring scrutiny but doesn't take any corrective measure other than logging error and reporting metrics. The default is once for every 30 hours.
+    default: 108000
```
eolivelli: Very last question:
 Shouldn't we keep this feature disabled by default?

reddycharan: valid point. will disable by default. will send the quick fix.

sijie:
@reddycharan have you addressed @eolivelli 's last review comment?

@eolivelli can you take a look at the latest patch?

reddycharan:
@sijie @eolivelli just pushed the commit, disabling the check by default.

reddycharan:

Issue #1914, opened 2019-01-22
EnsemblePlacementPolicy exposes third party API "Pair" from commons-lang3 in a public API


eolivelli:
We are going to release a public API which depends on a third party library. This will introduce compatibility problems in the future.

EnsemblePlacementPolicy API is an extension point on the client side, we should not depend on APIs from third party providers.

The issue was introduced in #1883

This is very like to #67, #300 and #804

reddycharan:
@eolivelli @sijie will take care of this in couple of days. Thanks for creating an issue for this.

@jvrao fyi.

eolivelli:
If you prefer @reddycharan I can send a patch.
The proposal from @sijie is to create a Pair class inside bookkeeper-common package.

I would like to move the release forward, because new  Pulsar release depends on this patch (AFAIK)

sijie:
@eolivelli agreed with you. I think it is a straightforward task, @reddycharan if you agree with approach, @eolivelli can help with the code change.

reddycharan:
@eolivelli @sijie I'm fine, you may proceed. Thanks.

sijie:
This has been fixed by #1916 in 4.9.0 release

PR #1916, opened 2019-01-23
Add semantically meaningful return values to placement policy

ivankelly:
Currently newEnsemble and replaceBookie in EnsemblePlacementPolicy
return a apache commons Pair<> with the second argument being a
boolean to denote whether the placement conforms strictly to the
policy. From calling code, the meaning of this second value is
unclear.

This patch replaces Pair<> with an PlacementResult object, in which
the strict conformity argument is clearly labels. This will also allow
extension in the future to return more metadata about particular
placements.

Also, we shouldn't put third party library classes in interfaces.

Issue:  #1914

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/EnsemblePlacementPolicy.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/EnsemblePlacementPolicy.java
@@ -411,11 +411,11 @@ public interface EnsemblePlacementPolicy {
     /**
      * Result of a placement calculation against a placement policy.
      */
-    class PlacementResult<T> {
+    final class PlacementResult<T> {
         private final T result;
         private final boolean adhering;
 
-        public PlacementResult(T result, boolean adhering) {
+        PlacementResult(T result, boolean adhering) {
             this.result = result;
             this.adhering = adhering;
         }
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/EnsemblePlacementPolicy.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/EnsemblePlacementPolicy.java
@@ -415,7 +415,11 @@ public interface EnsemblePlacementPolicy {
         private final T result;
         private final boolean adhering;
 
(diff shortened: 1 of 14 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/EnsemblePlacementPolicy.java
```diff
@@ -408,4 +407,25 @@ default boolean isEnsembleAdheringToPlacementPolicy(List<BookieSocketAddress> en
...
     }
+
+    /**
+     * Result of a placement calculation against a placement policy.
+     */
+    class PlacementResult<T> {
+        private final T result;
+        private final boolean adhering;
+
+        public PlacementResult(T result, boolean adhering) {
```
eolivelli: what about making the constructor private and making the class 'final' ?

We can open to extensions on the future if needed

ivankelly: I can't make it private because LocalBookieEnsemblePlacementPolicy is not in the same package. I've made it final though.

eolivelli: And what about making constructor private and adding a public static 'of' method to build instances?

sijie:
run pr validation

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/EnsemblePlacementPolicy.java
eolivelli: @sijie what about this comment of mine?

sijie: I am fine with either approach.

ivankelly:

eolivelli:
run integration tests

merlimat:
run integration tests

reddycharan:
Hey @sijie / @eolivelli / @ivankelly can you guys merge this PR. I have to rebase one of my PR on top of this change and waiting for it.

PR #1917, opened 2019-01-24
Ignore usage of localNode in PlacementPolicy.

reddycharan:

'ignoreLocalNodeInPlacementPolicy' specifies whether to ignore localnode
in the internal logic of placement policy. If it is not possible or useful
to use Bookkeeper client node's (or AutoReplicator) rack/region info. for
placement policy then it is better to ignore localnode instead of false
alarming with error/warn log lines and metrics.

It is not valid to expect rack mapping information to be available for Bookkeeper
client and ReplicationWorker nodes. So if rackmapping info. is not available
then it is better to ignore creation of BookieNode by resolving rack information
for the client localnode. Otherwise it would log unnecessary warn/error log lines
and rack resolution failure metrics.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicy.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicy.java
@@ -56,18 +56,20 @@ public class RackawareEnsemblePlacementPolicy extends RackawareEnsemblePlacement
                                                           int maxWeightMultiple,
                                                           int minNumRacksPerWriteQuorum,
                                                           boolean enforceMinNumRacksPerWriteQuorum,
+                                                          boolean ignoreLocalNodeInPlacementPolicy,
             StatsLogger statsLogger) {
         if (stabilizePeriodSeconds > 0) {
             super.initialize(dnsResolver, timer, reorderReadsRandom, 0, reorderThresholdPendingRequests, isWeighted,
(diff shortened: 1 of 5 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicy.java
```diff
@@ -56,18 +56,20 @@ protected RackawareEnsemblePlacementPolicy initialize(DNSToSwitchMapping dnsReso
                                                           int maxWeightMultiple,
                                                           int minNumRacksPerWriteQuorum,
                                                           boolean enforceMinNumRacksPerWriteQuorum,
+                                                          boolean ignoreLocalNodeInPlacementPolicy,
```
eolivelli: In the near future we should refactor this method and introduce a single bean or anyhow reduce the number of parameters.
Not blocker for this change

jvrao:
Just to generalize this, we are wondering if this is a replication worker running on bookie or not. Unless it is a bookie running on bookie, everyone else must ignore this right? Don't we have a configuration parameter for that (where replication can happen with any bookie)

reddycharan:
@jvrao it is not mainly about whether ReplicationWorker is running on bookie or not. This is generic change which says if 'localnode' of BookKeeper client (be it Application using Bookkeeper Client or ReplicationWorker) should be used in PlacementPolicy logic in any form or not.

sijie:

PR #1931, opened 2019-02-01
GetListOfEntriesOfLedger implementation

reddycharan:

As described in this BP - https://github.com/apache/bookkeeper/blob/master/site/bps/BP-34-cluster-metadata-checker.md, this request returns list of entries bookie contains for the given ledger in an encoded format. The returned list provides weakly consistent state, of the entries of the ledger.


bookkeeper-proto/src/main/proto/BookkeeperProtocol.proto
```diff
@@ -64,6 +64,7 @@ enum OperationType {
     GET_BOOKIE_INFO = 8;
     START_TLS = 9;
     FORCE_LEDGER = 10;
+    GET_LIST_OF_ENTRIES_OF_A_LEDGER = 11;
```
reddycharan: will change this to GET_LIST_OF_ENTRIES_OF_LEDGER and all the corresponding names will be changed to GetListOfEntriesOfLedger

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -1635,4 +1635,8 @@ static LedgerDirsManager createIndexDirsManager(ServerConfiguration conf, DiskCh
             return new LedgerDirsManager(conf, idxDirs, diskChecker, statsLogger);
         }
     }
+
+    public ByteBuf getListOfEntriesOfALedger(long ledgerId) throws IOException, NoLedgerException {
+        return null;
```
reddycharan: have to add corresponding method to LedgerDescriptor/LedgerDescriptorImpl and call that method here.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
```diff
@@ -335,4 +335,10 @@ public boolean isInForceGC() {
         return ledgerStorageList.stream()
             .map(single -> single.getGarbageCollectionStatus().get(0)).collect(Collectors.toList());
     }
+
+    @Override
+    public byte[] getEntriesOfLedger(long ledgerId) {
+        // TODO Auto-generated method stub
+        return null;
```
eolivelli: not sure how can we deal with this

I guess we have to implement this one as well, or have some kind of protocol in which the bookie can report to the client something like 'I don't know which are the entries I have'

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/GetListOfEntriesOfALedgerProcessorV3.java
```diff
@@ -0,0 +1,113 @@
...
+            if (listOfEntriesOfALedger != null) {
+                getListOfEntriesOfALedgerResponse
+                        .setAvailabilityOfEntriesOfLedger(ByteString.copyFrom(listOfEntriesOfALedger.nioBuffer()));
+            }
+        } catch (Bookie.NoLedgerException e) {
+            status = StatusCode.ENOLEDGER;
+            LOG.error("No ledger found while performing getListOfEntriesOfALedger from ledger: {}", ledgerId, e);
+        } catch (IOException e) {
+            status = StatusCode.EIO;
+            LOG.error("IOException while performing getListOfEntriesOfALedger from ledger: {}", ledgerId);
```
eolivelli: any reason for not reporting the stacktrace of the exception
IOException are very generic and having the detail will ease debug

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/InterleavedLedgerStorage.java
```diff
@@ -641,4 +649,233 @@ public void onRotateEntryLog() {
...
+     * sequences are grouped as a SequenceGroup. - ‘firstSequenceStart’ is the
+     * first entry in the first sequence of the SequenceGroup. -
+     * ‘lastSequenceStart’ is the first entry in the last sequence of the
+     * SequenceGroup.
+     *
+     * To represent a SequenceGroup, two long values and two int values are
+     * needed, so each SequenceGroup can be represented with (2 * 8 + 2 * 4 = 24
+     * bytes).
+     */
+    private static class SequenceGroup {
```
eolivelli: we should have specific unit tests around these classes, they can be tested without starting a bookie

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryMemTable.java
```diff
@@ -456,4 +459,95 @@ boolean isEmpty() {
...
+                }
+                hasToPreFetch = false;
+                return (curKVMapEntry != null || curSnapshotEntry != null);
+            }
+
+            @Override
+            public EntryKey next() {
+                if (hasNext()) {
+                    EntryKey returnEntryKey = null;
+                    if (curKVMapEntry != null && curSnapshotEntry != null) {
```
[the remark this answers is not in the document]
athanatos: Agreed.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/InterleavedLedgerStorage.java
```diff
@@ -641,4 +649,233 @@ public void onRotateEntryLog() {
...
+                seqGroup.serializeSequenceGroup(serializedSequenceGroupByteArray);
+                System.arraycopy(serializedSequenceGroupByteArray, 0, serializedStateByteArray,
+                        HEADER_SIZE + ((seqNum++) * SequenceGroup.SEQUENCEGROUP_BYTES),
+                        SequenceGroup.SEQUENCEGROUP_BYTES);
+            }
+            return serializedStateByteArray;
+        }
+    }
+
+    protected byte[] getEntriesOfLedger(long ledgerId, Iterator<EntryKey> entriesInMemtable) throws Exception {
```
athanatos: This shouldn't take entriesInMemtable as an input -- InterleavedLedgerStorage doesn't necessarily require a memtable.  Separate the above representation classes out into their own files with unit tests.  Modify this method to return an Iterator.  SortedMemTable would be responsible for merging this iterator with the one returned from the memtable.  The SequenceGroup above could simply take an interator in its constructor.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/InterleavedLedgerStorage.java
```diff
@@ -641,4 +649,233 @@ public void onRotateEntryLog() {
...
+     * the sake of future extensibility it would be helpful to have reserved
+     * space for header at the beginning. So the first 64 bytes will be used for
+     * header, with the first four bytes specifying the int version number, next
+     * four bytes specifying the number of sequencegroups for now and the rest
+     * of the bytes in the reserved space will be 0's. The encoded format will
+     * be represented after the first 64 bytes. The ordered collection of
+     * SequenceGroups will be appended sequentially to this byte array, with
+     * each SequenceGroup taking 24 bytes.
+     */
+    static class AvailabilityOfEntriesOfLedger {
```
athanatos: Call this something more general -- LedgerEntrySet or something.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SortedLedgerStorage.java
```diff
@@ -354,4 +355,10 @@ public boolean isInForceGC() {
     public List<GarbageCollectionStatus> getGarbageCollectionStatus() {
         return interleavedLedgerStorage.getGarbageCollectionStatus();
     }
+
+    @Override
+    public byte[] getEntriesOfLedger(long ledgerId) throws Exception {
```
athanatos: See: this shouldn't return a byte[] -- it should return an iterator representing the merged streams from the memtable and interleavedLedgerStorage.  The caller should be responsible for encoding it before it gets sent back.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
athanatos: An error code would suffice -- getEntriesOfLedger should therefore be able to raise a NotImplemented exception or whatever is most idiomatic.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/GetListOfEntriesOfALedgerProcessorV3.java
```diff
@@ -0,0 +1,113 @@
...
+            return getListOfEntriesOfALedgerResponse.build();
+        }
+
+        if (LOG.isDebugEnabled()) {
+            LOG.debug("Received new getListOfEntriesOfALedger request: {}", request);
+        }
+        StatusCode status = StatusCode.EOK;
+        ByteBuf listOfEntriesOfALedger = null;
+        try {
+            listOfEntriesOfALedger = requestProcessor.bookie.getListOfEntriesOfALedger(ledgerId);
```
athanatos: The encoding should happen here.

eolivelli: totally agree ! good idea @athanatos

eolivelli: I am not really sure getListOfEntriesOfALedger should return an iterator.
what about leaving the scan inside bookie or the LedgerStorage impl

getListOfEntriesOfALedger(long ledgerId, Consumer<Long> entryId)

this will ease handling of errors and it may help LedgerStorage implementations which already have that piece of information

athanatos: An Iterator based interface is likely to be easier for using a generic method for combining streams.

reddycharan: @athanatos @eolivelli I don't think it is valid to do here for the following reasons -

1) getListOfEntriesOfALedger implementation is LedgerStorage specific, I don't see how can we abstract this out at LedgerStorage specific level and deal with it here. I don’t think it is optimal to expect all the Storage implementation to return Iterator of long values (entry numbers), because in each storage implementation, things might be different for how entry details can be read. Having iterator based approach is not ideal in all cases.
2) From what I understand, GetListOfEntriesOfALedgerProcessorV3 is a request/response processor and it is not appropriate to deal with business logic in this class.
3) Also this class is in proto package, we don’t want the Encoding class details / code to be in this package and also this class won’t have access to ‘bookie’ package protected classes.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SortedLedgerStorage.java
reddycharan: as mentioned in other comment - 

https://github.com/apache/bookkeeper/pull/1931/files/a9d00a25d29bd034c7ee070f7a0d75bac29a94d2#r254091827

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/GetListOfEntriesOfALedgerProcessorV3.java
athanatos: For 2. I'd argue that it's not business logic, it's logic specific to the particular binary format you are using to ship it to the client.

For 1. I'm mainly arguing that you need to merge streams of sorted entries in at least two places, I'd like it to be uniform and an iterator seems the natural way to do that.  A visitor could also work.

3. That's a good argument for moving the class representing the encoding up into the proto package.

athanatos: Also, listing the entries of a ledger seems like it would be applicable to any ledger manager.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
sijie: I would suggest throw UnsupportedOperationException for now.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookkeeperInternalCallbacks.java
```diff
@@ -107,6 +107,13 @@
         void startTLSComplete(int rc, Object ctx);
     }
 
+    /**
+     * A callback interface for GetListOfEntriesOfALedger command.
+     */
+    public interface GetListOfEntriesOfALedgerCallback {
+        void getListOfEntriesOfALedgerComplete(int rc, long ledgerId, ByteBuf buffer, Object ctx);
```
sijie: I would suggest trying to use `CompletableFuture` on newly introduced features.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerStorage.java
```diff
@@ -248,4 +248,20 @@ default boolean isInForceGC() {
...
+     * LedgerStorage by the time this method is called will be available but
+     * modifications made after method invocation may not be available.
+     *
+     * @param ledgerId
+     *            - id of the ledger
+     * @return the list of entries of the ledger available in this
+     *         ledgerstorage.
+     * @throws Exception
+     */
+    byte[] getEntriesOfLedger(long ledgerId) throws Exception;
```
sijie: I would suggest returning a stream/iterator of entries. The encoding should be separated from LedgerStorage implementation.

reddycharan:
@sijie @eolivelli @jvrao this is good for final review.

eolivelli:
I left some minor comments, not addresses yet.
And CI is red

bookkeeper-proto/src/main/proto/BookkeeperProtocol.proto
reddycharan: fixed it

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
reddycharan: fixed it

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryMemTable.java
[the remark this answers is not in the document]
reddycharan: moved merging of iterator logic to bookkeeper-server/src/main/java/org/apache/bookkeeper/util/IteratorUtility.java

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
reddycharan: fixed it

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/InterleavedLedgerStorage.java
reddycharan: moved logic to AvailabilityOfEntriesOfLedger.java class and added relevant tests for this class

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/InterleavedLedgerStorage.java
reddycharan: new logic should take care of this

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SortedLedgerStorage.java
reddycharan: new logic should take care of this

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/GetListOfEntriesOfALedgerProcessorV3.java
reddycharan: new logic should take care of this

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerStorage.java
reddycharan: new logic should take care of this

bookkeeper-proto/src/main/proto/BookkeeperProtocol.proto
```diff
@@ -152,6 +154,10 @@ message GetBookieInfoRequest {
     optional int64 requested = 1;
 }
 
+message GetListOfEntriesOfALedgerRequest {
+	required int64 ledgerId = 1;
```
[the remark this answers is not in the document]
reddycharan: will fix indentation issues.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryMemTable.java
```diff
@@ -44,7 +47,7 @@
  */
 public class EntryMemTable implements AutoCloseable{
     private static Logger logger = LoggerFactory.getLogger(Journal.class);
-
+    static final long LONG_MAX_VALUE = Long.MAX_VALUE - 1;
```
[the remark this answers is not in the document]
reddycharan: hmm..I could've directly used 'Long.MAX_VALUE', but I'm worried if anywhere they are incrementing the entryid value (be it for comparison or something), thats why to be on safe side I'm creating this static variable.

I dont see any issue in naming it LONG_MAX_VALUE.

bookkeeper-proto/src/main/proto/BookkeeperProtocol.proto
reddycharan: already fixed it

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/InterleavedLedgerStorage.java
```diff
@@ -641,4 +649,233 @@ public void onRotateEntryLog() {
     public List<GarbageCollectionStatus> getGarbageCollectionStatus() {
         return Collections.singletonList(gcThread.getGarbageCollectionStatus());
     }
+
+    @Override
+    public byte[] getEntriesOfLedger(long ledgerId) throws Exception {
+        return getEntriesOfLedger(ledgerId, Collections.emptyIterator());
+    }
+
+    public static final long INVALID_ENTRYID = -1;
```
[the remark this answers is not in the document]
reddycharan: moved this logic to AvailabilityOfEntriesOfLedger and yeah I should make it private.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/GetListOfEntriesOfALedgerProcessorV3.java
reddycharan: will fix it

bookkeeper-proto/src/main/proto/BookkeeperProtocol.proto
```diff
@@ -213,5 +220,11 @@ message GetBookieInfoResponse {
     optional int64 freeDiskSpace = 3;
 }
 
+message GetListOfEntriesOfLedgerResponse {
+    required StatusCode status = 1;
+    required int64 ledgerId = 2;
+    optional bytes availabilityOfEntriesOfLedger = 3; // treated as array of bits indicating availability of entry at that particular index location
```
jvrao: I think comment is not accurate. Please update accordingly

reddycharan:
run integration tests

reddycharan:

reddycharan:
rebuild java11

reddycharan:

reddycharan:

reddycharan:
rebuild java11

reddycharan:

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/InterleavedLedgerStorage.java
```diff
@@ -643,4 +655,57 @@ public void onRotateEntryLog() {
...
+    @Override
+    public OfLong getListOfEntriesOfLedger(long ledgerId) throws IOException {
+        Iterator<LedgerCache.PageEntries> pageEntriesIteratorNonFinal = null;
+        try {
+            pageEntriesIteratorNonFinal = ledgerCache.listEntries(ledgerId).iterator();
+        } catch (Bookie.NoLedgerException noLedgerException) {
+            pageEntriesIteratorNonFinal = Collections.emptyIterator();
+        }
+        final Iterator<LedgerCache.PageEntries> pageEntriesIterator = pageEntriesIteratorNonFinal;
+        return new OfLong() {
```
sijie: It might be worth moving this class to a separate class so it is easy to write tests for it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/InterleavedLedgerStorage.java
```diff
@@ -643,4 +655,57 @@ public void onRotateEntryLog() {
...
+            long[] entriesInCurrentLedgerPage = new long[entriesPerLedgerEntryPage];
```
sijie: This seems to be an unnecessary allocation. I am wondering why can't you just use the LEP.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
```diff
@@ -352,4 +353,10 @@ static long getLongVariableOrDefault(ServerConfiguration conf, String keyName, l
             return conf.getLong(keyName);
         }
     }
+
+    @Override
+    public OfLong getListOfEntriesOfLedger(long ledgerId) throws IOException {
```
sijie: If you don't implement this, please create a github issue and comment it here.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/InterleavedLedgerStorage.java
reddycharan: adding new method in LEP to return iterator

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/InterleavedLedgerStorage.java
reddycharan: moved it to LedgerCache class

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
reddycharan: created https://github.com/apache/bookkeeper/issues/2078 and adding comment in the code

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookkeeperInternalCallbacks.java
reddycharan: introducing CompletableFuture

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryMemTable.java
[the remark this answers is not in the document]
reddycharan: removed this static long value

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryMemTable.java
```diff
@@ -456,4 +459,95 @@ boolean isEmpty() {
...
+                            returnEntryKey = curKVMapEntry;
+                            curKVMapEntry = null;
+                        } else {
+                            returnEntryKey = curSnapshotEntry;
+                            curSnapshotEntry = null;
+                        }
+                    } else if (curKVMapEntry != null) {
+                        returnEntryKey = curKVMapEntry;
+                        curKVMapEntry = null;
+                    } else {
```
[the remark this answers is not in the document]
reddycharan: i dont want to over complicate the utility iterator merger. I would leave a comment saying that IteratorUtility.mergePrimitiveLongIterator and IteratorUtility.mergeIteratorsForPrimitiveLongIterator expects input iterators to be sorted.

reddycharan:
@sijie @eolivelli addressed review comments and this should be good to go. Replication tests failure isn't related to my change, not sure why they are flapping.

reddycharan:

reddycharan:
rebuild java11

reddycharan:
Thanks @sijie @eolivelli @athanatos @jvrao for the review. I'm merging this PR.

PR #1937, opened 2019-02-06
[bk-gc] Fix GC thread gets blocked

rdhabalia:
### Motivation

It addresses below thread-stuck while performing gc in bookie.

```
Thread 3363: (state = BLOCKED)
 - java.lang.Object.wait(long) @bci=0 (Compiled frame; information may be imprecise)
 - java.lang.Object.wait() @bci=2, line=502 (Compiled frame)
 - org.apache.bookkeeper.util.ZkUtils.getChildrenInSingleNode(org.apache.zookeeper.ZooKeeper, java.lang.String) @bci=34, line=243 (Compiled frame)
 - org.apache.bookkeeper.meta.LongHierarchicalLedgerManager$LongHierarchicalLedgerRangeIterator.getChildrenAt(java.lang.String) @bci=8, line=165 (Compiled fr
ame)
 - org.apache.bookkeeper.meta.LongHierarchicalLedgerManager$LongHierarchicalLedgerRangeIterator$LeafIterator.<init>(org.apache.bookkeeper.meta.LongHierarchic
alLedgerManager$LongHierarchicalLedgerRangeIterator, java.lang.String) @bci=11, line=187 (Compiled frame)
 - org.apache.bookkeeper.meta.LongHierarchicalLedgerManager$LongHierarchicalLedgerRangeIterator$InnerIterator.advance() @bci=137, line=261 (Compiled frame)
 - org.apache.bookkeeper.meta.LongHierarchicalLedgerManager$LongHierarchicalLedgerRangeIterator$InnerIterator.next() @bci=28, line=281 (Compiled frame)
 - org.apache.bookkeeper.meta.LongHierarchicalLedgerManager$LongHierarchicalLedgerRangeIterator$InnerIterator.next() @bci=4, line=278 (Compiled frame)
 - org.apache.bookkeeper.meta.LongHierarchicalLedgerManager$LongHierarchicalLedgerRangeIterator$InnerIterator.next() @bci=4, line=278 (Compiled frame)
 - org.apache.bookkeeper.meta.LongHierarchicalLedgerManager$LongHierarchicalLedgerRangeIterator$InnerIterator.next() @bci=4, line=278 (Compiled frame)
 - org.apache.bookkeeper.meta.LongHierarchicalLedgerManager$LongHierarchicalLedgerRangeIterator.next() @bci=8, line=304 (Compiled frame)
 - org.apache.bookkeeper.meta.HierarchicalLedgerManager$HierarchicalLedgerRangeIterator.next() @bci=26, line=117 (Compiled frame)
 - org.apache.bookkeeper.bookie.ScanAndCompareGarbageCollector.gc(org.apache.bookkeeper.bookie.GarbageCollector$GarbageCleaner) @bci=195, line=168 (Compiled frame)
 - org.apache.bookkeeper.bookie.GarbageCollectorThread.doGcLedgers() @bci=8, line=393 (Compiled frame)
 - org.apache.bookkeeper.bookie.GarbageCollectorThread.runWithFlags(boolean, boolean, boolean) @bci=39, line=355 (Compiled frame)
 - org.apache.bookkeeper.bookie.GarbageCollectorThread.safeRun() @bci=28, line=333 (Compiled frame)
 - org.apache.bookkeeper.common.util.SafeRunnable.run() @bci=1, line=36 (Compiled frame)
 - java.util.concurrent.Executors$RunnableAdapter.call() @bci=4, line=511 (Compiled frame)
 - java.util.concurrent.FutureTask.runAndReset() @bci=47, line=308 (Compiled frame)
 - java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask) @bci=1, line=180 (Compiled frame)
 - java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run() @bci=37, line=294 (Compiled frame)
 - java.util.concurrent.ThreadPoolExecutor.runWorker(java.util.concurrent.ThreadPoolExecutor$Worker) @bci=95, line=1142 (Compiled frame)
 - java.util.concurrent.ThreadPoolExecutor$Worker.run() @bci=5, line=617 (Interpreted frame)
 - io.netty.util.concurrent.FastThreadLocalRunnable.run() @bci=4, line=30 (Interpreted frame)
 - java.lang.Thread.run() @bci=11, line=748 (Interpreted frame)
```

### Changes

add time-out to zk operation to avoid GC thread blocking.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
@@ -104,7 +104,6 @@ public class ServerConfiguration extends AbstractConfiguration<ServerConfigurati
     protected static final String GC_OVERREPLICATED_LEDGER_WAIT_TIME = "gcOverreplicatedLedgerWaitTime";
     protected static final String USE_TRANSACTIONAL_COMPACTION = "useTransactionalCompaction";
     protected static final String VERIFY_METADATA_ON_GC = "verifyMetadataOnGC";
-    protected static final String ZK_OP_TIME_OUT = "zkOperationTimeout";
     // Scrub Parameters
     protected static final String LOCAL_SCRUB_PERIOD = "localScrubInterval";
     protected static final String LOCAL_SCRUB_RATE_LIMIT = "localScrubRateLimit";
@@ -432,27 +431,6 @@ public class ServerConfiguration extends AbstractConfiguration<ServerConfigurati
         return this;
     }
 
-    /**
-     * Get zk-operation timeout in seconds used by GC while performing zk
-     * operation (set 0 to ignore timeout. default = 0).
-     *
-     */
-    public long getZkOperationTimeOut() {
-        return this.getLong(ZK_OP_TIME_OUT, 0);
-    }
-
-    /**
-     * Set zk-operation timeout in seconds used by GC while performing zk
-     * operation, set 0 to ignore timeout.
-     * 
-     * @param zkOpTimeout
-     * @return server configuration
-     */
-    public ServerConfiguration setZkOperationTimeOut(long zkOpTimeout) {
-        this.setProperty(ZK_OP_TIME_OUT, zkOpTimeout);
-        return this;
-    }
-    
     /**
      * Get whether local scrub is enabled.
      *
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
@@ -104,6 +104,7 @@ public class ServerConfiguration extends AbstractConfiguration<ServerConfigurati
     protected static final String GC_OVERREPLICATED_LEDGER_WAIT_TIME = "gcOverreplicatedLedgerWaitTime";
     protected static final String USE_TRANSACTIONAL_COMPACTION = "useTransactionalCompaction";
     protected static final String VERIFY_METADATA_ON_GC = "verifyMetadataOnGC";
+    protected static final String ZK_OP_TIME_OUT = "zkOperationTimeout";
     // Scrub Parameters
     protected static final String LOCAL_SCRUB_PERIOD = "localScrubInterval";
     protected static final String LOCAL_SCRUB_RATE_LIMIT = "localScrubRateLimit";
@@ -431,6 +432,27 @@ public class ServerConfiguration extends AbstractConfiguration<ServerConfigurati
         return this;
     }
 
+    /**
+     * Get zk-operation timeout in seconds used by GC while performing zk
+     * operation (set 0 to ignore timeout. default = 0).
+     *
+     */
+    public long getZkOperationTimeOut() {
+        return this.getLong(ZK_OP_TIME_OUT, 0);
+    }
+
+    /**
+     * Set zk-operation timeout in seconds used by GC while performing zk
+     * operation, set 0 to ignore timeout.
+     * 
+     * @param zkOpTimeout
+     * @return server configuration
+     */
+    public ServerConfiguration setZkOperationTimeOut(long zkOpTimeout) {
+        this.setProperty(ZK_OP_TIME_OUT, zkOpTimeout);
+        return this;
+    }
+    
     /**
      * Get whether local scrub is enabled.
      *
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/FlatLedgerManager.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/FlatLedgerManager.java
@@ -103,7 +103,8 @@ class FlatLedgerManager extends AbstractZkLedgerManager {
 
                 try {
                     zkActiveLedgers = ledgerListToSet(
-                            ZkUtils.getChildrenInSingleNode(zk, ledgerRootPath), ledgerRootPath);
+                            ZkUtils.getChildrenInSingleNode(zk, ledgerRootPath, ZkUtils.OP_TIME_OUT_SEC),
+                            ledgerRootPath);
                     nextRange = new LedgerRange(zkActiveLedgers);
                 } catch (KeeperException.NoNodeException e) {
                     throw new IOException("Path does not exist: " + ledgerRootPath, e);
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/FlatLedgerManager.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/FlatLedgerManager.java
@@ -89,7 +89,7 @@ class FlatLedgerManager extends AbstractZkLedgerManager {
     }
 
     @Override
-    public LedgerRangeIterator getLedgerRanges() {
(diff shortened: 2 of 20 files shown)


jvrao:
@athanatos

athanatos:
Fwiw, we saw a similar hang which turned out to be caused by https://github.com/apache/zookeeper/pull/787

I think this patch may be a good idea nevertheless.  I'd like to see the timeout passed as a param, however.

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/ZkUtils.java
```diff
@@ -240,7 +242,12 @@ public void operationComplete(int rc, List<String> ledgers) {
 
         synchronized (ctx) {
             while (!ctx.done) {
-                ctx.wait();
+                try {
+                    ctx.wait(TimeUnit.SECONDS.toMillis(OP_TIME_OUT_SEC));
```
merlimat: In this case, if no one calls `ctx.done = true`,  the thread will just keep waiting in 2sec blocks, but it will never exit, right?

rdhabalia: umm.. if no one calls `ctx.done = true; ctx.notifyAll();` at line#238 then `ctx.wait(timeout)` will throw `InterruptedException` and `ctx.done will be true;` at line#249 which takes out thread out of while loop. So, it will not keep waiting but exit. right?

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/FlatLedgerManager.java
```diff
@@ -103,7 +103,8 @@ private synchronized void preload() throws IOException {
 
                 try {
                     zkActiveLedgers = ledgerListToSet(
-                            ZkUtils.getChildrenInSingleNode(zk, ledgerRootPath), ledgerRootPath);
+                            ZkUtils.getChildrenInSingleNode(zk, ledgerRootPath, ZkUtils.OP_TIME_OUT_SEC),
```
athanatos: What I mean is, it should probably be passed in from the gc thread.  The gc caller should probably get it's value from a config.  Doing it here doesn't allow different users with different tolerances to provide different bounds.  Also, passing in 0 should mean no limit.

reddycharan:
@rdhabalia without fixing the real issue (taking forever to fetch ledgerznodes), how would this fix help? It can run into the same issue again, right?

rdhabalia:
@reddycharan 
> without fixing the real issue (taking forever to fetch ledgerznodes), how would this fix help? It can run into the same issue again, right?

the real issue is when zk does zxid rollover, it restarts the leader and zk-client doesn't receive the callback for in transit requests. however, new outgoing requests will not have the issue once zk quorum will come back. so, with this fix, blocked zk-call will be timeout and next time it will succeed.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/FlatLedgerManager.java
sijie: +1 to what @athanatos suggested. I would like to see a setting configurable.

athanatos:
@rdhabalia No, the patch I linked above addresses a problem with the zk client side cxid (32 bits), not the zk server side zxid (which is 64 bit iirc).  With that patch, this call won't hang in the first place.  However, that may not be the only such bug and there are other ways for zk calls to hang, so this patch may nevertheless be useful for some auxiliary zk users like the gc scan.

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
```diff
@@ -431,6 +432,27 @@ public ServerConfiguration setVerifyMetadataOnGc(boolean verifyMetadataOnGC) {
         return this;
     }
 
+    /**
+     * Get zk-operation timeout in seconds used by GC while performing zk
```
eolivelli: If this is only for GC we should provide a more meaningful name.
Can you change it please?

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/ZkUtils.java
```diff
@@ -240,7 +241,12 @@ public void operationComplete(int rc, List<String> ledgers) {
 
         synchronized (ctx) {
             while (!ctx.done) {
-                ctx.wait();
+                try {
+                    ctx.wait(timeOutSec > 0 ? TimeUnit.SECONDS.toMillis(timeOutSec) : 0);
```
eolivelli: What happens in case  of timeout ?

rdhabalia: it throws IOException that will be caught at the very top and gc metadata-iteration fails. so, current gc task fails and next gc task will again run to continue.

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
rdhabalia: i think this could be generic configuration for timeout and can be used at anywhere where bookie needs blocking zk call. However, I think we can also use existing `zkTimeout` which can genuinely guarantees that zk-session is alive and callback couldn't complete. or let me know if you guys think we need different name?

eolivelli:
> can you add the settings to `conf/bk_server.conf`?
@sijie  there is no new setting any more

sijie:
@eolivelli sorry I didn't follow the discussion. I just checked the conversation. I am not sure if using 2 * zkTimeout is a good idea. I would prefer a separate setting called zkOpTimeout, so people can really configure op timeout. We can set the default `zkOpTimeout` as 2 * `zkTimeout`.

eolivelli:
> @eolivelli sorry I didn't follow the discussion. I just checked the conversation. I am not sure if using 2 * zkTimeout is a good idea. I would prefer a separate setting called zkOpTimeout, so people can really configure op timeout. We can set the default `zkOpTimeout` as 2 * `zkTimeout`.

Ok @sijie 
The initial idea was to have a tunable parameter, and @rdhabalia introduced a generic timeout, used only in GC. So I proposed to cal that parameter with a name with an explicit reference to GC.
Then we ended with dropping the parameter.

A new parameter works for me as well but I think it should be stated clearly that it is only for GC operations

sijie:
> A new parameter works for me as well but I think it should be stated clearly that it is only for GC operations

if it is only for GC operations, I am fine with current approach.

rdhabalia:
@sijie 
> if it is only for GC operations, I am fine with current approach.

Yes, in previous commit, I have added separate config but then we realize that for now, it only needs for GC and GC should expect response before zk-session time out happens so, `zkTimeout` can be used to set this timeout for now so, we moved forward with this change.

rdhabalia:
un integration tests

rdhabalia:
run integration tests

rdhabalia:
run integration tests

rdhabalia:
@eolivelli  can we merge this PR?

eolivelli:
Merging now

PR #1938, opened 2019-02-06
[bk-gc] Prevent OOM: Restrict number of entry-loggers extraction

rdhabalia:
### Motivation

Due to #1937, bookies couldn't run GC for a week that accumulates > 10K entry-logger files on the disk. After restarting bookie, GC gets trigger and GC tries to extract and loads all data into memory which eventually causes OOM in bookie and bookies keep restarting.

### Changes

Bookie should extract entry-loggers in chunk to avoid memory pressure and OOM issues.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
@@ -63,6 +63,13 @@ public class GarbageCollectorThread extends SafeRunnable {
 
     // This is how often we want to run the Garbage Collector Thread (in milliseconds).
     final long gcWaitTime;
+    // Max number of entry-logger files should be extracted concurrently while
+    // performing GC
+    final int maxEntryLoggersPerScan;
+    // flag to iterate over chunk of of entry-logger files
+    boolean moreEntryLoggers = true;
+    long lastIterationLogId = 0;
+    private final boolean verifyMetadataOnGc;
 
     // Compaction parameters
     boolean enableMinorCompaction = false;
@@ -146,6 +153,9 @@ public class GarbageCollectorThread extends SafeRunnable {
         this.entryLogger = ledgerStorage.getEntryLogger();
         this.ledgerStorage = ledgerStorage;
         this.gcWaitTime = conf.getGcWaitTime();
+        
+        this.maxEntryLoggersPerScan = conf.getMaxEntryLoggersScanOnGc();
+        this.verifyMetadataOnGc = conf.getVerifyMetadataOnGC();
 
         this.numActiveEntryLogs = 0;
         this.totalEntryLogSize = 0L;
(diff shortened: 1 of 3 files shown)


merlimat:
I think a better long-term fix here is to remove the entyLogMetadata maps from the bookie memory and instead store them on disk (eg: in rocksdb). 

That will ensure that the amount of memory required by a bookie is completely independent from the amount of data stored in it.

rdhabalia:
> I think a better long-term fix here is to remove the entyLogMetadata maps from the bookie memory and instead store them on disk (eg: in rocksdb).

could be good idea. I will make the change.

reddycharan:
@rdhabalia to begin with, this is a major change touching one of the core component - gc. Functional tests validating the correctness and tests proving there wont be regression would be helpful.

reddycharan:
@rdhabalia correct me if I'm wrong with the following math

first of all I'm surprised that you have 10K entrylog files in a Bookie. What is your 'logSizeLimit'? the default value is 1 GB. If it is the default value then 10K entrylog file would take 10 TB.

lets assume you have 10 K entrylog files, for each entrylog file there will be EntryLogMetadata. Each EntryLogMetadata mainly contains Map of <long ledgerid, long ledgersize>. Lets assume on average in each entrylog file there will be entries of 100 ledgers. So it would be 100 * (8+8) = 1.6 KB. Including other overhead lets make it 4 KB. So 10 K * 4 KB = 40 MB. Even if we assume that there are entries of 1000 ledgers in an entrylog, including overhead memory size of EntryLogMetaData should be 40 KB, so 10K * 40KB = 400 MB.

So I'm not sure how memory needed for entryLogMetaMap could become bottleneck.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -322,55 +332,64 @@ public void runWithFlags(boolean force, boolean suspendMajor, boolean suspendMin
...
+        long startTime = System.currentTimeMillis();
+        boolean isIteration = false;
+        do {
+            // Extract all of the ledger ID's that comprise all of the entry
+            // logs
+            // (except for the current new one which is still being written to).
+            entryLogMetaMap = extractMetaFromEntryLogs(entryLogMetaMap, isIteration);
+
+            // gc inactive/deleted ledgers
+            doGcLedgers();
```
reddycharan: this is very expensive operation. It iterates over all the ledger ranges in the metadata server/all the ledgers in the bookie, this shouldn't be called multiple times

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -322,55 +332,64 @@ public void runWithFlags(boolean force, boolean suspendMajor, boolean suspendMin
...
+
+            // gc entry logs
+            doGcEntryLogs();
```
reddycharan: this method iterates over all the entries in entryLogMetaMap, again calling this multiple times is not sensible.

rdhabalia:
@reddycharan 
Bookie reached to 10K entryLog files due to this #1937 bug and based on bookie stats, bookie had > 20M ledgers across all entryLog files. So, as soon as GC triggers, bookie was dying due to OOM and dump gave the root cause where entryLogMetadata-map data structure causing the issues which holds a map of entryLogId->ledgerId->ledgerIdSize. and It ends up taking up over 14GB total to the bookie node. I ran this patch on the node with `maxEntryLoggersPerScan=2K` which prevented bookie going OOM and GC/Compaction successfully cleaned up all 10K entryLogger files.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
rdhabalia: yes, that's correct. therefore, have provided option `maxEntryLoggersPerScan=2000`. In normal case, bookie will have < 500 entryLog files so, it will not require multiple iteration but it will target issue where bookie is not able to recover with large number of entrylog files. and therefore, @merlimat suggested to persist entryLogmetadataMap into rocksDB so, I have created PR: #1949

reddycharan:
if you are planning to do - https://github.com/apache/bookkeeper/pull/1949/ , then is this change needed?

eolivelli:
@merlimat  @reddycharan  what's the status of this patch ?

Issue #1884, opened 2018-12-13
BUG: dir_*_usage stats are reported as 0 for read-only bookies after a restart


pasha-kuznetsov:
When a read-only bookie is restarted it keeps reporting `dir_*_usage` stats as `0` until it becomes writable again.

This is caused by the `LedgerDirsMonitor.check` only updating `diskUsages` if there are any writable dirs, or if the total usage goes below the low water mark, otherwise relying on previously filled values which are `0` after a bookie is restarted.

[Source](https://github.com/apache/bookkeeper/blob/052908ef30818d9f1e6be56bf3a3e3083c5bd6fb/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDirsMonitor.java#L133):
```
    private void check() {
        try {
            List<File> writableDirs = ldm.getWritableLedgerDirs();
            // Check all writable dirs disk space usage.
            for (File dir : writableDirs) {
                try {
                    diskUsages.put(dir, diskChecker.checkDir(dir));
                    ...
            }
        ...
        List<File> fullfilledDirs = new ArrayList<File>(ldm.getFullFilledLedgerDirs());
        boolean hasWritableLedgerDirs = ldm.hasWritableLedgerDirs();
        float totalDiskUsage = 0;

        // When bookie is in READONLY mode .i.e there are no writableLedgerDirs:
        // - Check if the total disk usage is below DiskLowWaterMarkUsageThreshold.
        // - If So, walk through the entire list of fullfilledDirs and add them back to writableLedgerDirs list if
        // their usage is < conf.getDiskUsageThreshold.
        try {
            if (hasWritableLedgerDirs
                    || (totalDiskUsage = diskChecker.getTotalDiskUsage(ldm.getAllLedgerDirs())) < conf
                            .getDiskLowWaterMarkUsageThreshold()) {
                // Check all full-filled disk space usage
                for (File dir : fullfilledDirs) {
                    try {
                        diskUsages.put(dir, diskChecker.checkDir(dir));
                        ...
```

PR #1941, opened 2019-02-08
Exclude defaultrack bookies when enforceMinNumRacksPerWriteQuorum is enabled

reddycharan:

- enforceMinNumRacksPerWriteQuorum is meant to be used for strict placement policy. So when
it is enabled, bookies which belong to default faultzone/rack (because of failure in resolving
network location) should be excluded from bookie selection.
- add gauge for number of bookies in default faultzone/rack. It will be helpful to create alerts
based on this gauge.
- add gauge for number of ledgers found not adhering to strict placement policy in Auditor's
placement policy check. This gauge will be more helpful in creating alert instead of using
monotonously increasing alert.


reddycharan:
@pasha-kuznetsov fyi.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperClientStats.java
```diff
@@ -93,6 +93,9 @@
     String FAILED_CONNECTION_COUNTER = "FAILED_CONNECTION_COUNTER";
     String FAILED_TLS_HANDSHAKE_COUNTER = "FAILED_TLS_HANDSHAKE_COUNTER";
 
+    // placementpolicy stats
+    String NUM_BOOKIES_IN_DEFAULT_FAULTZONE = "NUM_BOOKIES_IN_DEFAULT_FAULTZONE";
```
pasha-kuznetsov: Shouldn't we have similar stats for each fault zone, including number of writable bookies there? Then we could alert on a fault zone not having enough (writable) bookies in a fault zone.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
```diff
@@ -282,8 +288,24 @@ protected RackawareEnsemblePlacementPolicyImpl initialize(DNSToSwitchMapping dns
...
+            public Integer getSample() {
+                rwLock.readLock().lock();
+                try {
+                    return topology.countNumOfAvailableNodes(getDefaultRack(), Collections.emptySet());
+                } finally {
+                    rwLock.readLock().unlock();
+                }
+            }
+        };
+        this.statsLogger.registerGauge(NUM_BOOKIES_IN_DEFAULT_FAULTZONE, numBookiesInDefaultFaultZone);
```
pasha-kuznetsov: @reddycharan @jvrao registering the gauge before producing the actual numbers will result in issues similar to #1884 BUG: dir_*_usage stats are reported as 0 for read-only bookies after a restart — see the PR #1885 on the way it was addressed there

bookkeeper-server/src/test/java/org/apache/bookkeeper/replication/AuditorPlacementPolicyCheckTest.java
```diff
@@ -362,11 +361,10 @@ public void testPlacementPolicyCheckWithLedgersNotAdheringToPolicyWithMultipleSe
...
         try {
             TestStatsLogger statsLogger = startAuditorAndWaitForPlacementPolicyCheck(servConf, auditorRef);
-            TestCounter placementPolicyCheckEnsembleNotAdheringToPlacementPolicy = (TestCounter) statsLogger.getCounter(
-                    ReplicationStats.PLACEMENT_POLICY_CHECK_ENSEMBLE_NOT_ADHERING_TO_PLACEMENT_POLICY_COUNTER);
-            assertEquals("PLACEMENT_POLICY_CHECK_ENSEMBLE_NOT_ADHERING_TO_PLACEMENT_POLICY_COUNTER SuccessCount",
-                    (long) numOfLedgersNotAdheringToPlacementPolicy,
-                    placementPolicyCheckEnsembleNotAdheringToPlacementPolicy.get().longValue());
+            Gauge<? extends Number> ledgersNotAdheringToPlacementPolicyGuage = statsLogger
+                    .getGauge(ReplicationStats.NUM_LEDGERS_NOT_ADHERING_TO_PLACEMENT_POLICY);
+            assertEquals("NUM_LEDGERS_NOT_ADHERING_TO_PLACEMENT_POLICY guage value",
```
pasha-kuznetsov: nit: g**au**ge

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
```diff
@@ -240,7 +237,16 @@ public void reloadCachedMappings() {
...
+    @StatsDoc(
+            name = FAILED_TO_RESOLVE_NETWORK_LOCATION_COUNTER,
+            help = "Counter for number of times DNSResolverDecorator failed to resolve Network Location"
+    )
     protected Counter failedToResolveNetworkLocationCounter = null;
+    @StatsDoc(
+            name = NUM_BOOKIES_IN_DEFAULT_FAULTZONE,
+            help = "Gauge for the number of Bookies in default Fault Zone"
+    )
+    protected Gauge<Integer> numBookiesInDefaultFaultZone;
```
sijie: we rarely use "fault zone" in bookkeeper's concept. I would prefer using a consistent name in the same file, e.g. calling it default rack.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
reddycharan: @pasha-kuznetsov reporting '0' in this case is valid. It means that so far we haven't come across bookie which would fall in default rack. This is appropriate understanding.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
reddycharan: ok will rename it to defaultrack

reddycharan:
@pasha-kuznetsov @sijie i'm little surprised both of you suggested for metrics/stats for num of bookies (writable) available per rack. I have multiple concerns with this

1) what should be the name of the metric for this stat? It has to be different metric/stat for each rack so different names for each metric/stat. Moreover new Bookies can be added/removed to the cluster so racks can be added/deleted to the cluster. Effectively name of these metrics should be dynamical. If that is the case how can we create actionable alerts without knowing name of the metric/stats upfront?
2) I can see it might serve as info. purpose but creating an alert effectively for each rack is overkill. For. eg getting an alert saying that there is no writable bookie in a rack of 3 nodes, when there are multiple racks with enough writable bookies is not helpful.
3) In any case, even if we decide to have such metrics, I feel it should be treated as a separate work item. Because it needs new code to maintain the specific rack details and corresponding stats/metrics (dynamic in nature).

bookkeeper-server/src/test/java/org/apache/bookkeeper/replication/AuditorPlacementPolicyCheckTest.java
reddycharan: will change

sijie:
> 1. what should be the name of the metric for this stat?

I am thinking of `placementpolicy/<rack>/num_writable_bookie`.

> 1. If that is the case how can we create actionable alerts without knowing name of the metric/stats upfront?

different metrics/altering system might have different behaviors. some of the alerting systems are allowed '*' queries. so you can write alert rules on using `placementpolicy/*/num_writable_bookie` (e.g. count("placementpolicy/*/num_writble_bookies") => return the number of racks)

> 2\. I can see it might serve as info. purpose

I agree with this. 

> 3\. In any case, even if we decide to have such metrics, I feel it should be treated as a separate work item.

agree

> 3\. Because it needs new code to maintain the specific rack details and corresponding stats/metrics (dynamic in nature).

in current stats library, you can use `scope("rack")` for this.

reddycharan:

reddycharan:
@sijie @pasha-kuznetsov added testcases.

reddycharan:
@sijie @pasha-kuznetsov @jvrao @eolivelli waiting for final review.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
```diff
@@ -531,26 +561,40 @@ public void handleBookiesThatJoined(Set<BookieSocketAddress> joinedBookies) {
...
+    protected Set<BookieSocketAddress> addDefaultZoneBookiesIfMinNumRacksIsEnforced(
+            Set<BookieSocketAddress> excludeBookies) {
+        Set<BookieSocketAddress> comprehensiveExclusionBookiesSet;
+        if (enforceMinNumRacksPerWriteQuorum) {
+            comprehensiveExclusionBookiesSet = new HashSet<BookieSocketAddress>(excludeBookies);
+            topology.getLeaves(getDefaultRack()).forEach((Node node) -> {
+                if (node instanceof BookieNode) {
+                    BookieNode bookieNode = (BookieNode) node;
+                    comprehensiveExclusionBookiesSet.add(bookieNode.getAddr());
+                }
```
eolivelli: Else?

sijie: yeah we should handle `else` branch.

reddycharan: I think I added this if block just to be on safe side. Leaves of "getDefaultRack()" should be BookieNode. In no normal circumstances it should be non-BookieNode (InnerNode). Anyhow do you have any strong suggestion for what needs to be in else block?

@eolivelli @sijie

eolivelli: No I was only pointing here, maybe something that I missed was missing.

reddycharan:
@sijie it should be good to go.

reddycharan:

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperClientStats.java
jvrao: +1

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/EnsemblePlacementPolicy.java
```diff
@@ -251,6 +251,9 @@ EnsemblePlacementPolicy initialize(ClientConfiguration conf,
      * <p>{@code customMetadata} is the same user defined data that user provides
      * when {@link BookKeeper#createLedger(int, int, int, BookKeeper.DigestType, byte[], Map)}.
      *
+     * <p>If 'enforceMinNumRacksPerWriteQuorum' config is enabled then the bookies belonging to default
+     * faultzone (rack) will be excluded while selecting bookies. This is to enable strict placementpolicy.
```
jvrao: Not sure if the "This is to enable strict placementpolicy." is needed. Rest of the wording gives enough context.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
jvrao: @reddycharan  if we don't expect anything otherthan bookie node, we should just remove the if condition. Or add an else with exception. It really doesn't make sense to leave it like that.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
```diff
@@ -531,26 +561,40 @@ public void handleBookiesThatJoined(Set<BookieSocketAddress> joinedBookies) {
...
     }
 
+    /*
+     * this method should be called in readlock scope of 'rwLock'
+     */
+    protected Set<BookieSocketAddress> addDefaultZoneBookiesIfMinNumRacksIsEnforced(
+            Set<BookieSocketAddress> excludeBookies) {
+        Set<BookieSocketAddress> comprehensiveExclusionBookiesSet;
+        if (enforceMinNumRacksPerWriteQuorum) {
+            comprehensiveExclusionBookiesSet = new HashSet<BookieSocketAddress>(excludeBookies);
```
jvrao: Looks like you have put this inside the if condition to avoid unnecessary object allocation. To serve that purpose, may be you should check if the defaultRack is empty or not before allocating new object.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
```diff
@@ -531,26 +561,40 @@ public void handleBookiesThatJoined(Set<BookieSocketAddress> joinedBookies) {
...
+            topology.getLeaves(getDefaultRack()).forEach((Node node) -> {
+                if (node instanceof BookieNode) {
+                    BookieNode bookieNode = (BookieNode) node;
+                    comprehensiveExclusionBookiesSet.add(bookieNode.getAddr());
```
jvrao: I would like to have INFO message here with the list of all bookies that are in the default rack and are being excluded. I understand it could amplify logging but I am OK to take it when the configuration is asking for strict placement.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
```diff
@@ -531,26 +561,40 @@ public void handleBookiesThatJoined(Set<BookieSocketAddress> joinedBookies) {
...
-        return newEnsembleInternal(
-                ensembleSize,
-                writeQuorumSize,
-                writeQuorumSize,
-                excludeBookies,
-                parentEnsemble,
-                parentPredicate);
+        rwLock.readLock().lock();
+        try {
+            Set<BookieSocketAddress> comprehensiveExclusionBookiesSet = addDefaultZoneBookiesIfMinNumRacksIsEnforced(
```
jvrao: Somehow I am not a big fan of the name. It is too specific to what it is doing. Tomorrow we might add more criteria to exclusion of bookies (Say based on load etc); i am OK with it as-is but why not the name to me "compileComprehensiveExclusionSet" So it is generic; Anyway.. just a suggestion.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
```diff
@@ -1253,28 +1298,31 @@ static void shuffleWithMask(DistributionSchedule.WriteSet writeSet,
...
-                     * placement policy should be swallowed.
+                     * any issue/exception in analyzing whether ensemble is
+                     * strictly adhering to placement policy should be
+                     * swallowed.
                      */
                     LOG.warn("Received exception while trying to get network location of bookie: {}", bookie, e);
                 }
             }
-            if (racksOrRegionsInQuorum.size() < minNumRacksPerWriteQuorumForThisEnsemble) {
+            if ((racksInQuorum.size() < minNumRacksPerWriteQuorumForThisEnsemble)
```
jvrao: Don't you need to check for enforceMinNumRacksPerWriteQuorum for this if condition too? If this is implicit why not the defaultRack is not implicit?

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
```diff
@@ -612,12 +627,33 @@ public void run() {
...
+                             * placementPolicyCheck, it found few ledgers not
+                             * adhering to placement policy. So reporting it.
+                             */
+                            ledgersNotAdheringToPlacementPolicyGuageValue
+                                    .set(numOfLedgersFoundInPlacementPolicyCheckValue);
+                        }
+                        LOG.error(
+                                "BKAuditException running periodic placementPolicy check."
+                                        + "numOfLedgersNotAdheringToPlacementPolicy {}",
+                                numOfLedgersFoundInPlacementPolicyCheckValue, e);
```
jvrao: @reddycharan Why are you checking for non zero value of numOfLedgersFoundInPlacementPolicyCheckValue to set the stat/guage but not for error message? Do you expect to have an exception but a zero numOfLedgersFoundInPlacementPolicyCheckValue?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperClientStats.java
reddycharan: responded already - https://github.com/apache/bookkeeper/pull/1941#issuecomment-462951665

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
reddycharan: will add error logging

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
reddycharan: didn't get you, here i'm making sure that number of racks in quorum is not less than minNumRacksPerWriteQuorumForThisEnsemble and if enforceMinNumRacksPerWriteQuorum is enabled then no bookie is from defaultRack.

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
reddycharan: _Do you expect to have an exception but a zero numOfLedgersFoundInPlacementPolicyCheckValue?_

yes in the case of exception if we have not found faulty ledgers so far, we don't want to update the gauge value with zero. it would be misleading to report 0. so to avoid false negative i added this if condition in exception case.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
jvrao: If enforceMinNumRacksPerWriteQuorum is not enabled, (racksInQuorum.size() < minNumRacksPerWriteQuorumForThisEnsemble) is not an issue right? So I am thinking the check should be enforceMinNumRacksPerWriteQuorum && ((racksInQuorum.size() < minNumRacksPerWriteQuorumForThisEnsemble) || (enforceMinNumRacksPerWriteQuorum && racksInQuorum.contains(getDefaultRack()))

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
jvrao: I see; what kind of exception do you expect even when numOfLedgersFoundInPlacementPolicyCheckValue is zero.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
reddycharan: dont see an issue. will keep it as it is

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
reddycharan: @jvrao If enforceMinNumRacksPerWriteQuorum is not enabled, (racksInQuorum.size() < minNumRacksPerWriteQuorumForThisEnsemble is **an** issue . for writequorum to adhere to placement policy it should have bookies from atleast minNumRacksPerWriteQuorumForThisEnsemble. if enforceMinNumRacksPerWriteQuorum is enabled then on top of the existing condition, no bookie should be from default rack.

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
reddycharan: for placementPolicyCheck, it has to iterate over all the ledgers in the metadataserver and check the placement policy. So half way though the process there could be an exception and no faulty ledger is found so far. So in this case we want to log error but dont want to report gauge with zero value.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/EnsemblePlacementPolicy.java
reddycharan: will remove

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
reddycharan: will do

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
reddycharan: will do

reddycharan:
@jvrao addressed your comments.

reddycharan:
rebuild java11

reddycharan:

reddycharan:

reddycharan:
@sijie @jvrao @eolivelli since it is been approved and have successful checks, I'm going to merge this.

PR #1944, opened 2019-02-12
BP-38: Bypass journal ledger

ArvinDevel:

### Motivation

To guarantee high durability, BK write journal before flush data to persistent device which will cause two write of data.
At the presence of replicating and auto-recovery mechanism, the two-write is a bit waste of the persistent device bandwidth,
especially on the [scenarios](https://cwiki.apache.org/confluence/display/BOOKKEEPER/BP-14+Relax+durability) which prefer weak durability guarantee.
This proposal is aimed at providing bypass journal ledger, this feature includes these parts work:
 - add new write flag `BYPASS_JOURNAL` to existing protocol
 - impl the newly write flag at the client side and server side

### Changes

Master Issue: #1945

> ---
>
> - [ ] [skip build java11]: skip build on java11. *ONLY* skip this when *ONLY* changing files under documentation under `site`.
> ---

> ---
> quickly and easily:
>
>
>
> Otherwise:
> 
> 
> ---

### DIFF
--- a/site/bps/BP-38-bypass-journal-ledger.md
+++ b/site/bps/BP-38-bypass-journal-ledger.md
@@ -8,8 +8,7 @@ release: "N/A"
 ### Motivation
 
 To guarantee high durability, BK write journal before flush data to persistent device which will cause two write of data.
-At the presence of replicating and auto-recovery mechanism, the two-write is a bit waste of the persistent device bandwidth,
-especially on the [scenarios](https://cwiki.apache.org/confluence/display/BOOKKEEPER/BP-14+Relax+durability) which prefer weak durability guarantee.
+But we may not need this level of persistence under [scenarios](https://cwiki.apache.org/confluence/display/BOOKKEEPER/BP-14+Relax+durability) which prefer weak durability guarantee.
 This proposal is aimed at providing bypass journal ledger, this feature includes these parts work:
  - add new write flag `BYPASS_JOURNAL` to existing protocol
  - impl the newly write flag at the client side and server side
--- a/site/bps/BP-38-bypass-journal-ledger.md
+++ b/site/bps/BP-38-bypass-journal-ledger.md
@@ -12,7 +12,22 @@ But we may not need this level of persistence under [scenarios](https://cwiki.ap
 This proposal is aimed at providing bypass journal ledger, this feature includes these parts work:
  - add new write flag `BYPASS_JOURNAL` to existing protocol
  - impl the newly write flag at the client side and server side
- 
+
+The usage will looks like this:
+```java
+WriteHandle writeHandle = newCreateLedgerOp()
+                        .withEnsembleSize(3)
+                        .withWriteQuorumSize(3)
+                        .withAckQuorumSize(3)
+                        .withPassword(PASSWORD)
+                        .withWriteFlags(WriteFlag.BYPASS_JOURNAL)
+                        .execute().get();
+// append data as usual, but the lac won't advance
+writeHandle.appendAsync(...);
+// force to advance lac
+writeHandle.force();
+writeHandle.closeAsync();
+``` 
 The details of changes are listed in Section “[Proposed Changes](#proposed-changes)”.
  
 ### Public Interfaces
@@ -27,7 +42,7 @@ The detailed compatibility related stuff is in Section “[CDMP](#compatibility-
 While the main approach is based on [`WRITE_FLAG` impl](https://github.com/apache/bookkeeper/pull/742),
 the actual implementation is more tricky. In my opinion, there are three different solution:
 
-1. Relax LAC protocol
+1. Relax LAC protocol(rejected due to violate LAC protocol)
 
     Modify server side code mostly and don't change legerHandle's LAC advance logic, if the write flag is `BYPASS_JOURNAL`, after write to `LegerStorage`(the data maybe in the memTable, or the buffer of File, or the os cache),
 bookie return result to the client directly.
@@ -39,19 +54,19 @@ Note, the user shall know the weak durability for his use case when using this i
     Like `DEFERRED_SYNC`, the normal 'bypass-journal' operation don't advance LAC on the client.
 Only advance LAC using `force` api. To support this new 'bypass-journal' option, we need theses changes:
 
-    - Add persistent callback to LedgerStorage
+    - Add force interface to `LedgerDescriptor` and LedgerStorage
     
-        Maintain non-persistent entry list(necessary for out of order entry arriving) and `maxPersistentEntryId` on `LedgerDescriptor`
-If the entry is 'bypass-journal', the callback for this entry is not null, and it will receives notification after persistent through legerStorage.
+        Used to flush the data on memTable to persistent device for specific ledger, and this make senses with "one entry logger per ledger" feature.
+        For general SortedLedgerStorage, this "force" operation may has some overhead due to flush data too early.
     
     - Extend the force request
         
-        If the force request contains bypass journal option, get `maxPersistentEntryId` from `LedgerDescriptor`
- or impl force method on `LedgerDescriptor`, which force the entry to persistent device through `LedgerStorage`.
+        If the force request contains bypass journal option, execute force method of `LedgerDescriptor`, which will call LedgerStorage's force interface to flush received entries to persistent device.
 
     In addition to these changes, we should consider who is responsible for the `force` execution? How often to schedule 'force'?
+    To simply the design and give more choices to apps, these choices are up to user.
 
-3. Add NonPersistentLAC semantic to existing LAC protocol
+3. Add NonPersistentLAC semantic to existing LAC protocol(rejected due to infeasible and complex)
 
     To not violate LAC semantics like method 1, we can add `NonPersistentLAC` semantic, this includes these changes:
 
@@ -68,6 +83,8 @@ If the entry is 'bypass-journal', the callback for this entry is not null, and i
         Add `nonPersistentLAC` to WriteHandle. WriteHandle with 'bypass-journal' option updates the LAC using `maxPersistentEntryId`, and update `nonPersistentLAC` if receives enough ack.
         For normal WriteHandle, the `nonPersistentLAC` is set to null to indicate to keep original LAC advance logic. 
 
+As we only introduce `BYPASS_JOURNAL` write flag to add request, it's better not change existing fencing logic. So fence operation shall falls into journal io path.
+
 ### Compatibility, Deprecation, and Migration Plan
 
 As this new feature only add one ledger creating option and corresponding ledger implementation. It has no effect
@@ -77,7 +94,7 @@ When user use the latest client which has the `BYPASS_JOURNAL` write flag, but t
 
 ### Test Plan
 
-- unit tests for newly introduced write_flag at client side and server side
+- unit tests for newly introduced write_flag and extended force at client side and server side
 - end-to-end tests for the new Protocol request/response
 - compat tests (client with this new writeflags will not be able to write on old bookies)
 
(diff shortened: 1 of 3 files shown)


jvrao:
Glad to see this is being worked on. Will review the proposal.

site/bps/BP-38-bypass-journal-ledger.md
```diff
@@ -0,0 +1,91 @@
...
+issue: https://github.com/apache/bookkeeper/issues/1918
+state: "Under Discussion"
+release: "N/A"
+---
+
+### Motivation
+
+To guarantee high durability, BK write journal before flush data to persistent device which will cause two write of data.
+At the presence of replicating and auto-recovery mechanism, the two-write is a bit waste of the persistent device bandwidth,
+especially on the [scenarios](https://cwiki.apache.org/confluence/display/BOOKKEEPER/BP-14+Relax+durability) which prefer weak durability guarantee.
```
jvrao: I don't believe  our auto-recovery and replication alone can address our persistence needs. I would request you to reword it saying that - "we may not need this level of persistence under scenarios like week durability....." something like that.

site/bps/BP-38-bypass-journal-ledger.md
```diff
@@ -0,0 +1,91 @@
...
+This proposal is aimed at providing bypass journal ledger, this feature includes these parts work:
+ - add new write flag `BYPASS_JOURNAL` to existing protocol
```
jvrao: I assume BYPASS_JOURNAL is applicable only in the week durability case. We have that in the protocol, so can't BYPASS_JOURNAL be just a configuration parameter on bookie, which will used only (if enabled) on week durability case?

site/bps/BP-38-bypass-journal-ledger.md
```diff
@@ -0,0 +1,91 @@
...
+ - impl the newly write flag at the client side and server side
```
jvrao: Do you mean client visible API change ? or where exactly is this flag?

site/bps/BP-38-bypass-journal-ledger.md
```diff
@@ -0,0 +1,91 @@
...
+### Proposed Changes
+
+While the main approach is based on [`WRITE_FLAG` impl](https://github.com/apache/bookkeeper/pull/742),
+the actual implementation is more tricky. In my opinion, there are three different solution:
+
+1. Relax LAC protocol
+
+    Modify server side code mostly and don't change legerHandle's LAC advance logic, if the write flag is `BYPASS_JOURNAL`, after write to `LegerStorage`(the data maybe in the memTable, or the buffer of File, or the os cache),
+bookie return result to the client directly.
+This impl is like [disable syncData](https://github.com/apache/bookkeeper/issues/753), once all the replica fails, the BK cluster can't recovery from it.
```
jvrao: With this the above will be obsolete?

site/bps/BP-38-bypass-journal-ledger.md
ArvinDevel: Yes. What I mean is extending WriteFlags, the user who wants using this feature only needs pass WriteFlags to the WriteHandle. looks like this:
``` 
newCreateLedgerOp()
                .withEnsembleSize(3)
                .withWriteQuorumSize(3)
                .withAckQuorumSize(3)
                .withPassword(PASSWORD)
                .withWriteFlags(WriteFlag.BYPASS_JOURNAL)
                .execute()
```

site/bps/BP-38-bypass-journal-ledger.md
ArvinDevel: When user create LedgerHandle, he/she can construct it with `WriteFlag.BYPASS_JOURNAL`

site/bps/BP-38-bypass-journal-ledger.md
ArvinDevel: The bookie level config `syncData` example is just used to compare. We still needs to modify server side to sense client's WriteFlag.BYPASS_JOURNAL option, and extend WriteFlag to give users this choice.

site/bps/BP-38-bypass-journal-ledger.md
ArvinDevel: done

site/bps/BP-38-bypass-journal-ledger.md
```diff
@@ -0,0 +1,90 @@
...
+The detailed compatibility related stuff is in Section “[CDMP](#compatibility-deprecation-and-migration-plan)”.
+
+### Proposed Changes
+
+While the main approach is based on [`WRITE_FLAG` impl](https://github.com/apache/bookkeeper/pull/742),
+the actual implementation is more tricky. In my opinion, there are three different solution:
+
+1. Relax LAC protocol
+
+    Modify server side code mostly and don't change legerHandle's LAC advance logic, if the write flag is `BYPASS_JOURNAL`, after write to `LegerStorage`(the data maybe in the memTable, or the buffer of File, or the os cache),
```
eolivelli: > don't change legerHandle's LAC advance logic

I think we should use the DEFERRED_SYNC way of handling LAC, that is that a regular write will not advance LAC, but you need a "force" (like you state below).

site/bps/BP-38-bypass-journal-ledger.md
```diff
@@ -0,0 +1,90 @@
...
+
+    In addition to these changes, we should consider who is responsible for the `force` execution? How often to schedule 'force'?
+
+3. Add NonPersistentLAC semantic to existing LAC protocol
+
+    To not violate LAC semantics like method 1, we can add `NonPersistentLAC` semantic, this includes these changes:
+
+    - Add persistent callback to LedgerStorage
+
+        Maintain non-persistent entry list and `maxPersistentEntryId` on `LedgerDescriptor`.
```
eolivelli: I tried to have a 'maxPersistentEntryId' while working on DEFERRED_SYNC but actually it was not possible because if you use "LedgerHandleAdv" entries won't get to the bookie with a specific order, so you will have to keep track of holes.

site/bps/BP-38-bypass-journal-ledger.md
```diff
@@ -0,0 +1,90 @@
...
+        
+    - Wire protocol changes
+    
+        Add optional `maxPersistentEntryId` field to `AddResponse`. When the `AddRequest` has 'bypass-journal' option, the server get `maxPersistentEntryId` from `LedgerDescriptor` to construct response.
+
+    - Client side changes
+    
+        Add `nonPersistentLAC` to WriteHandle. WriteHandle with 'bypass-journal' option updates the LAC using `maxPersistentEntryId`, and update `nonPersistentLAC` if receives enough ack.
```
eolivelli: nonPersistentLAC is more like 'pendingAddsSequenceHead' that we introduced for DEFERRED_SYNC

eolivelli:
When you will go thru the implementation remember to create patches for the Server-side first and then we will be able to work on the client side, because client-side changes are "public API" and we cannot release an API which is missing the server side counterpart

site/bps/BP-38-bypass-journal-ledger.md
ArvinDevel: you're right, thanks

site/bps/BP-38-bypass-journal-ledger.md
ArvinDevel: Yes, and I forgot to state something to complete nonPersistentLAC :
-  carry `nonPersistentLAC` to bookie, and LedgerDesciptor record it
- extend readEntry of ReadHandle, so that it can read up to `nonPersistentLAC`

site/bps/BP-38-bypass-journal-ledger.md
ArvinDevel: The difficult is how to execute "force", leave to app or client lib, and how often to schedule force if WriteHandle is responsible for that. Do you have any ideas?

site/bps/BP-38-bypass-journal-ledger.md
eolivelli: Why persisting an non-persistent value ?
The value makes sense only on the client (writer), you won't ever read that value from bookies

site/bps/BP-38-bypass-journal-ledger.md
eolivelli: force() RPC may carry a flag on the wire which tells that the clients wants to fait for a flush of the memtable for the given ledger.
force() with DEFERRED_SINC -> wait for/force a write to the journal
force() with BYPASS_JOURNAL -> wait for/force a flush on the EntryLogger

It makes sense only in conjunction with "multiple entry loggers"/"one entry logger per ledger" feature.

site/bps/BP-38-bypass-journal-ledger.md
eolivelli: @jvrao I think we cannot have a bookie configuration parameter because the client must be aware of relaxed durability and advance LAC accordingly (like we do with DEFERRED_SYNC)

eolivelli:
@ArvinDevel  this doc does not deal with **fencing**.
We should state how with this new feature we are providing this core feature of BK.

With  DEFERRED_SYNC for instance we are not changing how recovery/fencing works.
Should we store the fence flag on journal ? or does the fence operation need a flush of the FileInfo to disk ?

site/bps/BP-38-bypass-journal-ledger.md
ArvinDevel: yes, but I'm more worried about that is the synchronized "force" applicable for bypass-journal weak durability. On the "deferred_sync" scenario, the consumer/ReadHandle can tolerant reading after "force" or close. If we still restrict the reader can only read up to the point where writer "force", will this limit the applied scenarios of bypass-journal weak durability?

site/bps/BP-38-bypass-journal-ledger.md
ArvinDevel: What I want is relaxing the reader to read up to latest entry as early as possible.

ArvinDevel:
> We should state how with this new feature we are providing this core feature of BK.

To be honest, I'm not familiar with fencing stuff. To simplify the design, can we keep the simple design which not changing recovery/fencing? Since the fence info is small enough, store it to journal has little effect.

site/bps/BP-38-bypass-journal-ledger.md
sijie: @ArvinDevel it would be good if you can put the example code in the BP.

site/bps/BP-38-bypass-journal-ledger.md
sijie: I agree with @eolivelli. `BYPASS_JOURNAL` should have same LAC semantics as `DEFERRED_SYNC`. We only advanced LAC when `force` happen. `force` will carry the write flag and decide to force a write to the journal, or force a flush on ledger storage.

For bypass journal use cases, I think applications are usually when a ledger is closed/forced. So the above assumptions would make implementation easier.

site/bps/BP-38-bypass-journal-ledger.md
sijie: @jvrao If I remember this correctly, we were thinking of adding BYPASS_JOURNAL as a WriteFlag, no? Applications can decide whether to use journal or not.

ArvinDevel:
@jvrao @eolivelli @sijie @merlimat 
Most of you are agreed to implement BYPASS_JOURNAL using WRITE_FLAG, and use "force" api to advance LAC, then I'll focus on this method on the doc, ok?

site/bps/BP-38-bypass-journal-ledger.md
ArvinDevel: will do that

PR #1946, opened 2019-02-12
(WIP) Have min time interval for force compaction.

reddycharan:

- minimum interval needed to trigger force compaction by LedgerDirsListener,
since the last main compaction. Otherwise if isForceGCAllowWhenNoSpace is enabled
and when disk reaches storage threshold, it will do busy force gc for every
'diskCheckInterval' (10 secs) untill the disk/bookie becomes writable again.
gc operation is an expensive operation and it puts load on metadataserver for
iterating over all ledgers metadata. So we should call forcegc judiciously when
disk/bookie reaches storage threshold.

- here main compaction will be Major compaction, if majorcompaction is
disabled/suspended then it will be minorcompaction. If both major and minor are
disabled/suspended then it will be just gc run.

- in LedgerDirsMonitor and in GarbageCollectorThread we should schedule tasks
with fixed delay instead of at fixed rate. This would ensure that there would be
uniform delay between executions.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
@@ -74,6 +74,8 @@ public class GarbageCollectorThread extends SafeRunnable {
     final double majorCompactionThreshold;
     final long majorCompactionInterval;
     long lastMajorCompactionTime;
+    final long forceCompactionInterval;
+    final AtomicLong lastMainCompactionTime;
 
     final boolean isForceGCAllowWhenNoSpace;
 
@@ -175,6 +177,7 @@ public class GarbageCollectorThread extends SafeRunnable {
         majorCompactionThreshold = conf.getMajorCompactionThreshold();
         majorCompactionInterval = conf.getMajorCompactionInterval() * SECOND;
         isForceGCAllowWhenNoSpace = conf.getIsForceGCAllowWhenNoSpace();
+        forceCompactionInterval = conf.getForceCompactionInterval();
 
         AbstractLogCompactor.LogRemovalListener remover = new AbstractLogCompactor.LogRemovalListener() {
             @Override
@@ -227,7 +230,9 @@ public class GarbageCollectorThread extends SafeRunnable {
         LOG.info("Major Compaction : enabled=" + enableMajorCompaction + ", threshold="
                + majorCompactionThreshold + ", interval=" + majorCompactionInterval);
 
-        lastMinorCompactionTime = lastMajorCompactionTime = System.currentTimeMillis();
+        long curSysTime = System.currentTimeMillis();
+        lastMinorCompactionTime = lastMajorCompactionTime = curSysTime;
+        lastMainCompactionTime = new AtomicLong(curSysTime);
     }
 
     public void enableForceGC() {
@@ -242,6 +247,20 @@ public class GarbageCollectorThread extends SafeRunnable {
         if (forceGarbageCollection.compareAndSet(true, false)) {
             LOG.info("{} disabled force garbage collection since bookie has enough space now.", Thread
                     .currentThread().getName());
+        }
+    }
+
+    public void enableForceGCIfMinIntervalElapsed() {
+        long curTime = System.currentTimeMillis();
+        if ((lastMainCompactionTime.longValue() - curTime) > forceCompactionInterval) {
+            enableForceGC();
+        } else {
+            if (LOG.isDebugEnabled()) {
+                LOG.debug(
+                        "lastMainCompactionTime: {}, curTime: {}. Interval is less than forceCompactionInterval: {},"
+                                + " so skipping enableForceGC",
+                        lastMainCompactionTime.longValue(), curTime, forceCompactionInterval);
+            }
         }
     }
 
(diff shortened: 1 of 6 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -297,7 +316,7 @@ public void start() {
         if (scheduledFuture != null) {
             scheduledFuture.cancel(false);
         }
-        scheduledFuture = gcExecutor.scheduleAtFixedRate(this, gcWaitTime, gcWaitTime, TimeUnit.MILLISECONDS);
+        scheduledFuture = gcExecutor.scheduleWithFixedDelay(this, gcWaitTime, gcWaitTime, TimeUnit.MILLISECONDS);
```
sijie: any reason why did you change this to `FixedDelay`?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -245,6 +250,20 @@ public void disableForceGC() {
         }
     }
 
+    public void enableForceGCIfMinIntervalElapsed() {
+        long curTime = System.currentTimeMillis();
+        if ((lastMainCompactionTime.longValue() - curTime) > forceCompactionInterval) {
```
jiazhai: Will `curTime` always bigger than `lastMainCompactionTime`?

jvrao: yeah curTime will be >= lastMain isn't it @reddycharan ?

reddycharan: bummer..yes, it should be other way round. In next iteration I'm planning to add testcases.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -74,6 +74,8 @@
     final double majorCompactionThreshold;
     final long majorCompactionInterval;
     long lastMajorCompactionTime;
+    final long forceCompactionInterval;
+    final AtomicLong lastMainCompactionTime;
```
jvrao: why are we calling it 'main' compaction? Don't we need to delay as long as we run *any* compaction?

reddycharan: @jvrao - here main compaction will be Major compaction, if majorcompaction is
  disabled/suspended then it will be minorcompaction. If both major and minor are
  disabled/suspended then it will be just gc run.

I dont want to delay if major compaction is enabled and just minor compaction has run recently. Because if forcegc is triggered immediately without delay, then there is good chance of reclaiming enough space to become writable again.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -359,8 +379,21 @@ public void runWithFlags(boolean force, boolean suspendMajor, boolean suspendMin
             minorCompacting.set(true);
             doCompactEntryLogs(minorCompactionThreshold);
             lastMinorCompactionTime = System.currentTimeMillis();
+            if (!enableMajorCompaction || suspendMajor) {
```
jvrao: Given the if condition above, we enter this block only if MajorCompaction is enabled right?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -359,8 +379,21 @@ public void runWithFlags(boolean force, boolean suspendMajor, boolean suspendMin
...
+                lastMainCompactionTime.set(lastMinorCompactionTime);
+            }
             gcStats.getMinorCompactionCounter().inc();
             minorCompacting.set(false);
+        } else if (((!enableMajorCompaction) || suspendMajor) && ((!enableMinorCompaction) || suspendMinor)) {
+            /*
+             * if both compactions (major and minor) are disabled / suspended,
+             * then consider regular gc run as main compaction
+             */
+            lastMainCompactionTime.set(curTime);
```
jvrao: What is the downside of any major/minor/gc as last compaction time? This  to make a decision if we need to run force GC or not right?

reddycharan: I dont want to delay forcegc, if major compaction is enabled and just minor compaction has run recently. Because, if forcegc is triggered immediately without delay, then there is good chance of reclaiming enough space to become writable again.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
karanmehta93: @sijie `FixedDelay` would ensure that fixed amount of time passes between consecutive runs. This will prevent from GC process overburdening the bookie. With `FixedRate`, the duration of GC run is counted as part of the delay time.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
karanmehta93: yes @jvrao That's right.
I will rename this as `lastCompactionTime` and add a comment there.

karanmehta93:
I will be taking over this PR. Major task is to add tests and fix the bug. 
@sijie @jvrao Any other concerns here before I start?

karanmehta93:

eolivelli:
Are you using this patch in production at SF ?
cc @jvrao @karanmehta93

karanmehta93:
@eolivelli We aren't using this patch in prod. While its a actual issue, the frequency of it in production is rare for us. Hence we didn't prioritize it.

eolivelli:
I am closing this PR.

Feel free to open it again Charan if you want to complete the work.

eolivelli:
closing in favour of #2094

PR #1949, opened 2019-02-15
[bookie-gc] add option to cache entry-log metadata map into rocksDB

rdhabalia:
### Motivation

It addresses #1938, where it provides interface for `entyLogMetadataMap` so, `entrylogMetadata` can be cached into RocksDB instead main-memory.

### Modification
- introduced RocksDB entryLogMetadataMap to cache entry-log metadata.
- it can fallback to in-memory implementation using  `gcPersistentEntrylogMetadataMapEnabled` flag and that can be removed in future release and keep only rocksDB implementation in future.

### Result
- It will help to avoid storing entryLogMetadata into main-memory.

cc @merlimat


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogMetadata.java
```diff
@@ -96,4 +100,49 @@ public String toString() {
...
+        out.writeLong(ledgersMap.size());
+        ledgersMap.forEach((ledgerId, size) -> {
+            try {
+                out.writeLong(ledgerId);
+                out.writeLong(size);
+            } catch (IOException e) {
+                throw new IllegalStateException("Failed to serialize entryLogMetadata", e);
+            }
+        });
+        out.writeLong(remainingSize);
```
eolivelli: is this a duplicate of the third item ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogMetadata.java
```diff
@@ -96,4 +100,49 @@ public String toString() {
...
+     * {@link DataOutputStream}.
+     * 
+     * @param out
+     * @throws IOException
+     *             throws if it couldn't serialize metadata-fields
+     * @throws IllegalStateException
+     *             throws if it couldn't serialize ledger-map
+     */
+    public void serialize(DataOutputStream out) throws IOException, IllegalStateException {
+        out.writeLong(entryLogId);
```
eolivelli: should we have some versioning flag ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogMetadataMap.java
```diff
@@ -0,0 +1,44 @@
...
+import java.io.IOException;
+import java.util.function.BiConsumer;
+
+/**
+ *   Map-store to store Entrylogger metadata. 
+ *
+ */
+public interface EntryLogMetadataMap extends Closeable {
+
+    boolean containsKey(long entryLogId) throws IOException;
```
eolivelli: please add minimal javadoc

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -127,35 +127,41 @@ public GarbageCollectorThread(ServerConfiguration conf, LedgerManager ledgerMana
                 Executors.newSingleThreadScheduledExecutor(new DefaultThreadFactory("GarbageCollectorThread")));
     }
 
+    private EntryLogMetadataMap createEntryLogMetadataMap(boolean gcPersistentEntrylogMetadataMapEnabled) {
+        final String baseDir = this.entryLogger.getLedgerDirsManager().getAllLedgerDirs().get(0).toString();
```
eolivelli: what happens if you have more than one "ledgerDir" ? should this directory be configurable ?

maybe you can move this line inside the *if* block at line 133

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -127,35 +127,41 @@ public GarbageCollectorThread(ServerConfiguration conf, LedgerManager ledgerMana
...
+        try {
+            if (gcPersistentEntrylogMetadataMapEnabled) {
+                return new PersistentEntryLogMetadataMap(baseDir, conf);
+            }
+        } catch (IOException e) {
+            LOG.error("Failed to initialize persistent-metadata-map , clean up {}", baseDir, e);
```
eolivelli: I think it is not safe to fallback to "inMemory" in case of errors,
if you are using the persistent memory you don't want to use  the in memory one (will the bookie crash with OutOfMemory errors?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -190,67 +196,63 @@ public void removeEntryLog(long logToRemove) {
...
-                      suspendMinorCompaction.get());
+            triggerGC(true, suspendMajorCompaction.get(), suspendMinorCompaction.get());
         }
     }
 
     public void disableForceGC() {
         if (forceGarbageCollection.compareAndSet(true, false)) {
-            LOG.info("{} disabled force garbage collection since bookie has enough space now.", Thread
-                    .currentThread().getName());
+            LOG.info("{} disabled force garbage collection since bookie has enough space now.",
```
eolivelli: it would be better not to reformat code, as this changes makes the diff harder to review.

Can you revert the formatting ?
as soon as checkstyle is happy we are okay

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -306,7 +312,11 @@ public void safeRun() {
         boolean suspendMajor = suspendMajorCompaction.get();
         boolean suspendMinor = suspendMinorCompaction.get();
 
-        runWithFlags(force, suspendMajor, suspendMinor);
+        try {
+            runWithFlags(force, suspendMajor, suspendMinor);
+        } catch (Exception e) {
+            LOG.warn("Failed to complete gc {}", e.getMessage(), e);
```
eolivelli: if we are catching "Exception" any time you call runWithFlags....
why don't catch the error inside runWithFlags and leave the previous signature ?

catch (Exception) is a code smell, do we really need it ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogMetadataMap.java
```diff
@@ -0,0 +1,44 @@
...
+
+    void put(long entryLogId, EntryLogMetadata entryLogMeta) throws IOException;
```
eolivelli: IOException is very generic, what about having a specific EntryLogMetadataMapException
it will ease understanding the source of the Exception (in catch/rethrow clauses I mean)

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -415,64 +436,69 @@ private void removeIfLedgerNotExists(EntryLogMetadata meta) {
...
-            int bucketIndex = Math.min(
-                    numBuckets - 1,
-                    (int) Math.ceil(meta.getUsage() * numBuckets));
+        entryLogMetaMap.forEach((entryLogId, meta) -> {
+            int bucketIndex = Math.min(numBuckets - 1, (int) Math.ceil(meta.getUsage() * numBuckets));
             entryLogUsageBuckets[bucketIndex]++;
 
             if (meta.getUsage() >= threshold) {
-                continue;
+                return;
```
eolivelli: this is a behaviour change
is this needed in this patch ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/InMemoryEntryLogMetadataMap.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.BiConsumer;
+
+import org.apache.bookkeeper.bookie.EntryLogMetadata;
+import org.apache.bookkeeper.bookie.EntryLogMetadataMap;
+
+public class InMemoryEntryLogMetadataMap implements EntryLogMetadataMap {
+
+    private Map<Long, EntryLogMetadata> entryLogMetaMap = new ConcurrentHashMap<Long, EntryLogMetadata>();
```
eolivelli: final

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/InMemoryEntryLogMetadataMap.java
```diff
@@ -0,0 +1,65 @@
...
+
+import java.io.IOException;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.BiConsumer;
+
+import org.apache.bookkeeper.bookie.EntryLogMetadata;
+import org.apache.bookkeeper.bookie.EntryLogMetadataMap;
+
+public class InMemoryEntryLogMetadataMap implements EntryLogMetadataMap {
```
eolivelli: javadoc ?

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/storage/ldb/PersistentEntryLogMetadataMapTest.java
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    /**
+     * Validates PersistentEntryLogMetadataMap functionalities.
+     * 
+     * @throws Exception
+     */
+    @Test
+    public void simple() throws Exception {
+        Path tmpDir = Files.createTempDirectory("bookie");
```
eolivelli: can we use junit TemporaryFolder rule ?

jvrao:
@reddycharan  can you please review this? This is larger change and we don't use rocksDb, lot of code reorg. @sijie @merlimat  do we seriously have an issue where we need to handle large number of entrylogs ? This all came out of ZK/GC issue and it will be taken care.

jvrao:
>    keep only rocksDB implementation in future.

I don't think that is a good option. Can we talk about this in the next week's meeting? 
For now I am not a fan of this change.

merlimat:
> @sijie @merlimat do we seriously have an issue where we need to handle large number of entrylogs ? This all came out of ZK/GC issue and it will be taken care.

I haven't checked this PR yet (will do soon :) ). 

The main issue is that since the entryLogMaps for each entryLog are kept in memory. The amount of memory needed by a bookie is directly proportional to the amount of data stored. Your milage might vary: if you have all your data in 10 ledgers this will require less memory that if you have it in 10M ledgers, because for each entryLog we track (ledgerId -> sizeOfEntriesInThisLog).. 

In any case, I believe this is a fundamental issue with the current approach, in that we need to be able to have a fixed amount of memory used, independent of the amount of data.

In many cases, with bookies holding >10TB we saw this very high mem usage, leading to issue where the heap size is not enough and bookie basically restart and goes into a loop of GC leading to OOM.

eolivelli:
I am ok with Rocks DB.

In other projects where I need huge CuncurrentMap I am using MapDB

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
rdhabalia: > what happens if you have more than one "ledgerDir" ? should this directory be configurable ?

actually, that doesn't matter because here we use it to just get the base dir path of ledger so, we can create dedicated dir to store metadata and that is independent of multi-ledger dir usecase. also we do similar at [multiple places](https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java#L2869) to get base-dir path for the entry-logger and stores rocksDB dir under it.
However, we can also add config to get dir path and fallback to this logic if config is not provided. I will add it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
rdhabalia: > (will the bookie crash with OutOfMemory errors?

No, it's not necessary that it will throw OOM because using inMemory is same as current logic where GC stores metadata-map into hashmap. in order to solve in-memory scale issue, we can also add this patch #1938.

> I think it is not safe to fallback to "inMemory" in case of errors,

I think it can be helpful in case of rocksDB corruption and user will not be impacted in such cases and GC can still continue. It will not create any issue because entryLogMetadataMap is any way transient and recovered by extracting entry-log files.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
rdhabalia: yes, formatting was not intentional change but IDE performed automatically. i will try to revert these line changes.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
rdhabalia: not behavior change but in `forEach(..)`, we need to return if we want to skip the record processing in the lamda. that's same as `continue` in `for(..)`

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogMetadata.java
rdhabalia: good idea. it will be helpful in future if we want to add/change definition.

rdhabalia:
@eolivelli
addressed your all comments.

rdhabalia:

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/InMemoryEntryLogMetadataMap.java
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+package org.apache.bookkeeper.bookie.storage.ldb;
```
reddycharan: org.apache.bookkeeper.bookie.storage.ldb package is for "Classes related to DB based ledger storage.", so not sure why this class has to be in this package - InMemoryEntryLogMetadataMap.java

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
```diff
@@ -431,6 +433,47 @@ public ServerConfiguration setVerifyMetadataOnGc(boolean verifyMetadataOnGC) {
         return this;
     }
 
+    /**
+     * Get whether the bookie is configured to use persistent
+     * entrylogMetadataMap.
+     * @return use persistent entry-log metadata map
+     */
+    public boolean isGcPersistentEntrylogMetadataMapEnabled() {
+        return this.getBoolean(GC_PERSISTENT_ENTRYLOGMETADATA_MAP_ENABLED, true);
```
reddycharan: i'm not sure if it is correct to set it to true by default. This will have perf impact, so I dont think making it default behavior is not ideal for new users/ who are unaware of usage characteristics.

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
```diff
@@ -104,6 +104,8 @@
     protected static final String GC_OVERREPLICATED_LEDGER_WAIT_TIME = "gcOverreplicatedLedgerWaitTime";
     protected static final String USE_TRANSACTIONAL_COMPACTION = "useTransactionalCompaction";
     protected static final String VERIFY_METADATA_ON_GC = "verifyMetadataOnGC";
+    protected static final String GC_PERSISTENT_ENTRYLOGMETADATA_MAP_ENABLED = "gcPersistentEntrylogMetadataMapEnabled";
```
reddycharan: you missed adding this config to bk_server.yaml

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
rdhabalia: sure. I will change it to false.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -127,6 +127,21 @@ public GarbageCollectorThread(ServerConfiguration conf, LedgerManager ledgerMana
                 Executors.newSingleThreadScheduledExecutor(new DefaultThreadFactory("GarbageCollectorThread")));
     }
 
+    private EntryLogMetadataMap createEntryLogMetadataMap(boolean gcPersistentEntrylogMetadataMapEnabled) {
+        final String baseDir = this.conf.getGcPersistentEntrylogMetadataMapPath() != null
```
reddycharan: reading of getGcPersistentEntrylogMetadataMapPath should be done only if gcPersistentEntrylogMetadataMapEnabled is enabled. so move this line to the 
`if (gcPersistentEntrylogMetadataMapEnabled){..}` block

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -127,6 +127,21 @@ public GarbageCollectorThread(ServerConfiguration conf, LedgerManager ledgerMana
...
+                ? this.conf.getGcPersistentEntrylogMetadataMapPath()
```
reddycharan: i think conf.getGcPersistentEntrylogMetadataMapPath (method in ServerConfiguration.java) should handle this logic. I mean if the config value is null then use ledgerDirs().get(0).

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -127,6 +127,21 @@ public GarbageCollectorThread(ServerConfiguration conf, LedgerManager ledgerMana
...
+                : this.entryLogger.getLedgerDirsManager().getAllLedgerDirs().get(0).toString();
```
reddycharan: since you are adding new persistence state to the bookie - 'PersistentEntrylogMetadataMapPath', have you thought through all the persistence scenarios? like any changes required to cookie, what it means for bookie formatting, backward compatibility / forward compatibility scenarios,...

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -127,6 +127,21 @@ public GarbageCollectorThread(ServerConfiguration conf, LedgerManager ledgerMana
                 Executors.newSingleThreadScheduledExecutor(new DefaultThreadFactory("GarbageCollectorThread")));
     }
 
+    private EntryLogMetadataMap createEntryLogMetadataMap(boolean gcPersistentEntrylogMetadataMapEnabled) {
```
reddycharan: if this method is anyhow going to access "this.conf" for EntrylogMetadataMapPath, then is it needed for this method to receive 'gcPersistentEntrylogMetadataMapEnabled' as an argument to this method.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -142,20 +157,21 @@ public GarbageCollectorThread(ServerConfiguration conf,
...
         this.numActiveEntryLogs = 0;
         this.totalEntryLogSize = 0L;
         this.garbageCollector = new ScanAndCompareGarbageCollector(ledgerManager, ledgerStorage, conf, statsLogger);
         this.gcStats = new GarbageCollectorStats(
-            statsLogger,
-            () -> numActiveEntryLogs,
-            () -> totalEntryLogSize,
-            () -> garbageCollector.getNumActiveLedgers()
-        );
+                statsLogger,
```
reddycharan: it would be easier on reviewers if you are not making formatting changes on the existing code

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
athanatos: I don't think that's a good idea.  As a general rule, rocksdb is fairly robust.  If rocksdb is corrupt, likely there is a hardware problem.  The user should have to manually switch the config value back to in-memory if they want to bypass the problem.

It would also be nice to have a command to rebuild the rocksdb instance.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieException.java
```diff
@@ -254,6 +260,23 @@ public CookieExistException(Throwable cause) {
         }
     }
 
+    /**
+     * Signal that error while accessing entry-log metadata map.
+     */
+    public static class EntryLogMetadataMapException extends BookieException {
+        public EntryLogMetadataMapException() {
```
athanatos: This doesn't seem to be used.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieException.java
```diff
@@ -254,6 +260,23 @@ public CookieExistException(Throwable cause) {
...
+            this("");
+        }
+
+        public EntryLogMetadataMapException(String reason) {
```
athanatos: This doesn't seem to be used.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -547,11 +585,14 @@ protected void compactEntryLog(EntryLogMetadata entryLogMeta) {
...
-                // Read through the entry log file and extract the entry log meta
+                // Read through the entry log file and extract the entry log
+                // meta
                 EntryLogMetadata entryLogMeta = entryLogger.getEntryLogMetadata(entryLogId);
-                removeIfLedgerNotExists(entryLogMeta);
+                removeIfLedgerNotExists(entryLogId, entryLogMeta, false);
                 if (entryLogMeta.isEmpty()) {
                     entryLogger.removeEntryLog(entryLogId);
+                    // remove it from entrylogmetadata-map if it presents into map
+                    removeEntryLogRecordSafely(entryLogId);
```
athanatos: I don't understand why you want to swallow the exception here.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -571,10 +613,23 @@ protected void compactEntryLog(EntryLogMetadata entryLogMeta) {
         return entryLogMetaMap;
```
athanatos: This only gets called afaict with this.entryLogMetaMap.  I don't think it needs to be passed as a parameter or returned, particularly since removeEntryLogRecordSafely above assumes that it's using this.entryLogMetaMap.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -323,54 +336,60 @@ public void runWithFlags(boolean force, boolean suspendMajor, boolean suspendMin
...
+            if (force) {
+                if (forceGarbageCollection.compareAndSet(true, false)) {
+                    LOG.info("{} Set forceGarbageCollection to false after force GC to make it forceGC-able again.",
+                            Thread.currentThread().getName());
+                }
             }
+            gcStats.getGcThreadRuntime().registerSuccessfulEvent(MathUtils.nowInNano() - threadStart,
+                    TimeUnit.NANOSECONDS);
+        } catch (EntryLogMetadataMapException e) {
+            LOG.error("Error in entryLog-metadatamap, Failed to complete GC/Compaction due to entry-log {}",
```
reddycharan: i dont get why are you swallowing the exception explicitly. If you look at the callers of runWithFlags, it is called from 'safeRun' or this method is submitted to executor, which would recreate the thread if it is terminated because of a failure

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -381,31 +400,43 @@ private void doGcLedgers() {
...
+            totalEntryLogSizeAcc.getAndAdd(meta.getRemainingSize());
         });
 
         this.totalEntryLogSize = totalEntryLogSizeAcc.get();
-        this.numActiveEntryLogs = entryLogMetaMap.keySet().size();
+        this.numActiveEntryLogs = entryLogMetaMap.size();
     }
 
-    private void removeIfLedgerNotExists(EntryLogMetadata meta) {
+    private void removeIfLedgerNotExists(long entryLogId, EntryLogMetadata meta, boolean updateMetadata)
```
reddycharan: meta (EntryLogMetadata) contains entryLogId, so you dont have to pass 'entryLogId' explicitly

athanatos:
I'd like to suggest an alternate approach.  We already in fact have the ledger map for each entry log compiled at the end of each file.  Suppose instead of adding rocksdb, you use some bounded auto-loading cache (evict newest entry rather than oldest to avoid thrashing the cache during scan) which falls back to simply re-reading the index for the given EntryLog?  It would smoothly transition between fully in-memory and spilling to disk behaviors as needed.  By introducing rocksdb, you introduce a new way for bookie state to be corrupted -- inconsistency between entry log and cached ledger map.  You also introduce rocksdb itself as a source of overhead -- it's got (among other things) its own writer thread and journal.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -415,64 +446,69 @@ private void removeIfLedgerNotExists(EntryLogMetadata meta) {
                 return false;
             }
         });
+        // update entryMetadta to persistent-map
+        if (updateMetadata) {
```
reddycharan: im not sure how you are deciding to update entrylogmetadata. Isn't it supposed to be updated only if there is change in the entrylogmetadata, right? In the current code, if the current argument - 'updateMetadata' value is true, isn't it going to update even if there is no change in entrylogmetadata?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
reddycharan: what is the expectation here with 'updateMetadata', when is caller expected to pass true/false?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -381,31 +400,43 @@ private void doGcLedgers() {
...
-               // This means the entry log is not associated with any active ledgers anymore.
-               // We can remove this entry log file now.
-               LOG.info("Deleting entryLogId " + entryLogId + " as it has no active ledgers!");
-               removeEntryLog(entryLogId);
-               gcStats.getReclaimedSpaceViaDeletes().add(meta.getTotalSize());
-           }
-
-           totalEntryLogSizeAcc.getAndAdd(meta.getRemainingSize());
+            try {
+                removeIfLedgerNotExists(entryLogId, meta, true);
```
reddycharan: i guess to begin with, this call 'removeIfLedgerNotExists' is redundant here. 'removeIfLedgerNotExists' should be already taken care by "entryLogMetaMap = extractMetaFromEntryLogs(entryLogMetaMap);" in 'runWithFlags' method. 'removeIfLedgerNotExists' call should be removed from here.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/PersistentEntryLogMetadataMap.java
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    @Override
+    public void put(long entryLogId, EntryLogMetadata entryLogMeta) throws EntryLogMetadataMapException {
+        LongWrapper key = LongWrapper.get(entryLogId);
+        try {
+            baos.get().reset();
+            try {
+                entryLogMeta.serialize(dataos.get());
+                dataos.get().flush();
```
reddycharan: output of serialize is expected to be flushed. so move this flush call to serialize method.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/PersistentEntryLogMetadataMap.java
```diff
@@ -0,0 +1,185 @@
...
+                metadataMapDB.put(key.array, baos.get().toByteArray());
+            } catch (IllegalStateException | IOException e) {
+                LOG.error("Failed to deserialize entrylog-metadata, entryLogId {}", entryLogId);
```
reddycharan: nit: failed to serialize

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/PersistentEntryLogMetadataMap.java
```diff
@@ -0,0 +1,185 @@
...
+                    localDatais.close();
+                    ByteArrayInputStream newBais = new ByteArrayInputStream(entry.getValue());
+                    bais.set(newBais);
+                    datais.set(new DataInputStream(newBais));
+                } else {
+                    localBais.read(entry.getValue(), 0, entry.getValue().length);
+                }
+                localBais.reset();
+                localDatais.reset();
+                action.accept(entryLogId, EntryLogMetadata.deserialize(datais.get()));
```
reddycharan: this 'forEach' call is going to be quite expensive on jvm gc call. Since it will create EntryLogMetadata object for each entrylog file. You should consider optimizing it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
rdhabalia: I think bookie-formatting is the only operation that requires metadata cleanup for which I made the change. also, GC has fall-back mechanism to use InMemoryMap in case of PersistentMap is not accessible and PersistentMap is recoverable by extracting entryLog so, that should take care of forward/backward compatibility concerns.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
rdhabalia: I think we can avoid failing entire gc-task due to single record failure and gc should continue by ignoring that record because that will not cause any harm except it will not remove that specific entry/ledger in that gc-run.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
rdhabalia: at the beginning GC creates `EntryLogMetadata` by reading and extracting entryLog file and at that time entry doesn't exist into map so, there is no need to update rocksDB. However, this flag was confusing so, I have removed the flag and updating metadata-map where it requires to be updated at ln#414.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
rdhabalia: yes, I think thats' correct. However, this change is not related to this PR because even previously GC used to call this method. So, can we add this change in separate PR to make sure we don't miss any usecase while making this change.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/PersistentEntryLogMetadataMap.java
rdhabalia: yes, I had the same concern but I forgot to make this change. but, now I have added recyclable entryMetadata class rather making any change in existing `EntryLogMetadata` because I don't want to change life-cycle semantic of existing `EntryLogMetadata`

rdhabalia:
@reddycharan @athanatos addressed all the comments.

rdhabalia:

athanatos:
I had a comment above about an alternate approach entirely that wouldn't require further ondisk state.  What are your thoughts on that?

rdhabalia:
> Suppose instead of adding rocksdb, you use some bounded auto-loading cache..It would smoothly transition between fully in-memory and spilling to disk behaviors as needed.

I have created a PR #1938 to mainly avoid in memory pressure when we have large number of entryLog files while performing GC. It does similar thing which you mentioned and it performs compaction with bounded and limited number of entryLog files. I have run the same patch when I had faced bookie OOM due to large number of entryLoggers and it works fine for me. However, it pays more performance penalty because it might require multiple iterations while checking against zk-nodes. Also, GC runs on the same bookie jvm-process and this GC-entryLogMetadata-map sits in memory even when bookie is not running compaction which again adds memory pressure to bookie. Therefore, @merlimat also suggested to store metadata-map into rocksDB to avoid consuming heap-memory.

> By introducing rocksdb, you introduce a new way for bookie state to be corrupted -- inconsistency between entry log and cached ledger map.

rocksDB map will always be updated when gc-task gets trigger so, it will not have inconsistency while gc/compaction is running. Also, rocksDB-map is again recoverable from entrylog files and any corruption will fallback to inmemory map which will not cause gc-failure as well.

rdhabalia:
ping @eolivelli @reddycharan @merlimat

rdhabalia:
@sijie addressed comments.

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/BookKeeperConstants.java
```diff
@@ -31,6 +31,7 @@
     public static final String BOOKIE_STATUS_FILENAME = "BOOKIE_STATUS";
     public static final String PASSWD = "passwd";
     public static final String CURRENT_DIR = "current";
+    public static final String METADATA_DIR = "metadata";
```
sijie: I would suggest calling this a `entrylog_index_cache` directory. `metadata` is too ambitious and it indicates that it is a directory for storing persistent information.

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
```diff
@@ -104,6 +105,8 @@
...
```
sijie: I think a better name for these two settings would be:

`entryLogMetadataCacheEnabled` and `entryLogMetadataCachePath`

rdhabalia: @sijie sorry for late response. I think I missed this notification. I addressed all the comments.

rdhabalia:
rebuild java8
run pr validation

rdhabalia:

rdhabalia:
rebuild java11

rdhabalia:

rdhabalia:
rebuild java8

rdhabalia:

rdhabalia:

rdhabalia:

rdhabalia:

rdhabalia:
run integration tests

rdhabalia:

rdhabalia:
run integration tests

rdhabalia:

rdhabalia:

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -1560,6 +1560,15 @@ public static boolean format(ServerConfiguration conf,
                     }
                 }
             }
+
+            // Clean up metadata directories if they are separate from the
+            // ledger dirs
+            File metadataDir = new File(conf.getGcEntryLogMetadataCachePath());
+            if (!cleanDir(metadataDir)) {
+                LOG.error("Formatting ledger metadata directory " + metadataDir + " failed");
```
vzhikserg: Please use a parametrized message https://www.slf4j.org/faq.html#logging_performance

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -323,54 +333,60 @@ public void runWithFlags(boolean force, boolean suspendMajor, boolean suspendMin
...
+                lastMinorCompactionTime = System.currentTimeMillis();
+                gcStats.getMinorCompactionCounter().inc();
+                minorCompacting.set(false);
+            }
 
-        if (force) {
-            if (forceGarbageCollection.compareAndSet(true, false)) {
-                LOG.info("{} Set forceGarbageCollection to false after force GC to make it forceGC-able again.", Thread
-                    .currentThread().getName());
+            if (force) {
```
vzhikserg: Two if statements can be merged into one

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -381,31 +397,45 @@ private void doGcLedgers() {
     }
 
     /**
-     * Garbage collect those entry loggers which are not associated with any active ledgers.
+     * Garbage collect those entry loggers which are not associated with any
+     * active ledgers.
+     *
+     * @throws IOException
```
vzhikserg: Should not it be `EntryLogMetadataMapException`?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -381,31 +397,45 @@ private void doGcLedgers() {
...
+                // Ignore and continue because ledger will not be cleaned up
+                // from entry-logger in this pass and will be taken care in next
+                // schedule task
+                LOG.warn("Failed to remove ledger from entry-log metadata {}", entryLogId, e);
+            }
+            if (meta.isEmpty()) {
+                // This means the entry log is not associated with any active
+                // ledgers anymore.
+                // We can remove this entry log file now.
+                LOG.info("Deleting entryLogId " + entryLogId + " as it has no active ledgers!");
```
vzhikserg: Please use a parametrized message

rdhabalia:
@vzhikserg addressed all the comments. @sijie @eolivelli @reddycharan already addressed other comments as well. Can we merge this PR as I have to use explicit patch while upgrading the bookie.

rdhabalia:
run integration tests

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -1560,6 +1560,15 @@ public static boolean format(ServerConfiguration conf,
                     }
                 }
             }
+
+            // Clean up metadata directories if they are separate from the
+            // ledger dirs
+            File metadataDir = new File(conf.getGcEntryLogMetadataCachePath());
```
eolivelli: is this code to be executed only if the feature is enabled?

eolivelli:
What about this work?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
rdhabalia: yes, it uses rocksDB to cache metadata map only if the feature is enabled else it continues to use existing logic.

merlimat:
I had given some thoughts to this problem a long time back. Let me regather that into a document.

rdhabalia:
run integration tests

rdhabalia:
@eolivelli 
can we merge this change because default configuration will not execute this code path and won't impact existing setup. and if we have any thoughts then we can add on top of this changes. we have to run patched version in our prod env as we have seen this issue multiple times and it can actually bring down our entire cluster.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
eolivelli: Palese address this comment

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
eolivelli: Oh yes, I see

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
eolivelli: Did you ever see this case in production?
If we do not have a strong motivation O prefer  to leave the code as before here

rdhabalia:
@eolivelli addressed above two comments as well..

rdhabalia:
ping.. can we merge this or #1938.. we will be fine with either option as far as we have option to prevent this issue.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
eolivelli: This comment looks unresolved

eolivelli:
@rdhabalia  can you please rebase and fix the last comment ?
we have enough approvals to merge this patch

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
eolivelli: @rdhabalia  please address this comment from me and from @athanatos 
cc @merlimat

conf/bk_server.conf
```diff
@@ -534,6 +534,12 @@ ledgerDirectories=/tmp/bk-data
 # True if the bookie should double check readMetadata prior to gc
 # verifyMetadataOnGC=false
 
+# True if bookie should persist entrylog file metadata and avoid in-memory object allocation
+gcEntryLogMetadataCacheEnabled=false
+
+# Directory to persist Entrylog metadata if gcPersistentEntrylogMetadataMapEnabled is true
+# gcEntryLogMetadataCachePath=
```
eolivelli, line 541: we should document the behavior of this configuration parameter
in ServerConfiguration we are using a computed default value....
we should have a consistent and documented behavior

https://github.com/apache/bookkeeper/pull/1949/files#diff-ca6cec8ce70c1bb7a79d3aab8ce752deb12731a33348510c2f9472f5d8ef53eeR464

eolivelli (changes requested):
@rdhabalia this is a very nice feature, I wonder if you have time to address the remaining comments

eolivelli:
@merlimat can you please take a look ?

eolivelli:
@rdhabalia @merlimat @Ghatage 
I hope that we can resume this work and make it land to BK 4.13.0

rdhabalia:
I will address the comments soon.

conf/bk_server.conf
rdhabalia: good point. addressed it.

rdhabalia:
@eolivelli rebased the PR and also addressed both the comments.

eolivelli (approved):
LGTM

@sijie you left feedback in the past, do you mind taking another look ?

dlg99 (commented):
Overall LGTM.
A few comments but the idea lgtm.
Now we have to figure out how to deal with case of bookie running out of disk space and not having disk and memory to handle all the metadata for compaction :)

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -127,6 +126,20 @@ public GarbageCollectorThread(ServerConfiguration conf, LedgerManager ledgerMana
                 Executors.newSingleThreadScheduledExecutor(new DefaultThreadFactory("GarbageCollectorThread")));
     }
 
+    private EntryLogMetadataMap createEntryLogMetadataMap() throws IOException {
+        if (conf.isGcEntryLogMetadataCacheEnabled()) {
+            String baseDir = this.conf.getGcEntryLogMetadataCachePath();
+            try {
+                return new PersistentEntryLogMetadataMap(baseDir, conf);
```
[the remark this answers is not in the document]
rdhabalia: > Are you trying to preserve data on the disk between bookie restarts?

yes. and we should prevent auto-cleanup because this can be also failed if by mistake multiple processes try to read rocks-db same time then it may fail intermittently. so, to avoid such scenarios, let user cleanup rocksDB cache manually if needed. generally rocksDB is robust and it should not fail in normal conditions.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogMetadata.java
```diff
@@ -96,4 +109,111 @@ public String toString() {
...
+     *             throws if it couldn't serialize ledger-map
+     */
+    public void serialize(DataOutputStream out) throws IOException, IllegalStateException {
+        out.writeShort(DEFAULT_SERIALIZATION_VERSION);
+        out.writeLong(entryLogId);
+        out.writeLong(totalSize);
+        out.writeLong(remainingSize);
+        out.writeLong(ledgersMap.size());
+        ledgersMap.forEach((ledgerId, size) -> {
+            try {
```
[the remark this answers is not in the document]
rdhabalia: `serialize` method anyway throws `IOException` so, no need to catch `IOException` for `out.flush()`. the only reason of try/catch and throwing runtime-exception is because it's in lambda and it requires runtime-exception.

rdhabalia:
@eolivelli @sijie can we merge this PR? we can also create separate PRs if we need any enhancement or improvement on top of it.?

eolivelli:
@sijie I would like to merge this feature, WDYT?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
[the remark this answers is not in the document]
hangc0276: If we configured multi ledger directories, all the ledger directories' PersistentEntryLogMetadataMap use the same RocksDB and located in the first ledger directory. 

When we do `forEach` operation for PersistentEntryLogMetadataMap, it will iterate all the items in RocksDB. However most of items we iterated are not belongs to current ledger storage, it will wast of time and cpu resource.  

We'd better separate the PersistentEntryLogMetadataMap's RocksDB for each ledger directory. Do you have any ideas? @rdhabalia

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogMetadata.java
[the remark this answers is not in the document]
pkumar-singh: Probably this is one of the reason to avoid lamba here. Just a normal iteration should be just fine.

rdhabalia:
@hangc0276 sure, we can make that enhancement, but first I would like to merge this PR then we can make any enhancement as this PR is sitting for a while. we are patching our internal release to protect ourselves. so, it would be great if we can merge it first and then make any required enhancement.

@eolivelli @sijie can we merge this PR?

eolivelli:
It looks like this patch needs a rebase @rdhabalia 

Also @sijie  left 'request changes', so we have to check if his concerns have been fixed

rdhabalia:
I think I addressed all the comments. Actually, I had rebased it multiple times so, if it looks fine for two +1 then I will rebase it and then we can merge it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -381,31 +403,44 @@ private void doGcLedgers() {
...
-               LOG.info("Deleting entryLogId " + entryLogId + " as it has no active ledgers!");
-               removeEntryLog(entryLogId);
-               gcStats.getReclaimedSpaceViaDeletes().add(meta.getTotalSize());
-           }
-
-           totalEntryLogSizeAcc.getAndAdd(meta.getRemainingSize());
+            try {
+                removeIfLedgerNotExists(meta);
+                // update entryMetadta to persistent-map
+                entryLogMetaMap.put(meta.getEntryLogId(), meta);
```
zymap, line 420: Maybe we can update it only when the meta is changed?

rdhabalia:
@sijie @eolivelli can you please let me know if all the changes are addressed to unblock the PR. it's hard to keep rebasing and keep addressing new comments for a year.

sijie (approved):

zymap:
Hi @rdhabalia, could you please resolve the conflicts?

PR #1953, opened 2019-02-19
[bookie-http-server] Fix: broken vertx rest endpoints

rdhabalia:
### Motivation
Right now, vertx-http-server is not serving any rest endpoint except get because vertx server doesn't add put/post/delete http-methods into routing rules.

### Modification
Add put/post/delete http-methods into routing rules.

### DIFF
--- a/bookkeeper-http/vertx-http-server/src/main/java/org/apache/bookkeeper/http/vertx/VertxHttpServer.java
+++ b/bookkeeper-http/vertx-http-server/src/main/java/org/apache/bookkeeper/http/vertx/VertxHttpServer.java
@@ -70,6 +70,9 @@ public class VertxHttpServer implements HttpServer {
             @Override
             public void bindHandler(String endpoint, VertxAbstractHandler handler) {
                 router.get(endpoint).handler(handler);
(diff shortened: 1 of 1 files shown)


bookkeeper-http/vertx-http-server/src/main/java/org/apache/bookkeeper/http/vertx/VertxHttpServer.java
```diff
@@ -70,6 +70,9 @@ public boolean startServer(int port) {
             @Override
             public void bindHandler(String endpoint, VertxAbstractHandler handler) {
                 router.get(endpoint).handler(handler);
+                router.put(endpoint).handler(handler);
```
sijie: interesting... I am wondering why it was not captured by our tests. /cc @jiazhai can you check do we have test coverage of `gc` endpoint?

rdhabalia: I think test only covers get and not the put. I will add test for triggering gc.

sijie:
run integration tests

PR #1958, opened 2019-02-23
Use automatic resource management to close streams

liketic:
We can use automatic resource management introduced in java 7, which makes code shorter.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
@@ -123,16 +123,9 @@ public class BookieStatus {
      * @throws IOException
      */
     private static void writeToFile(File file, String body) throws IOException {
-        FileOutputStream fos = new FileOutputStream(file);
-        BufferedWriter bw = null;
-        try {
-            bw = new BufferedWriter(new OutputStreamWriter(fos, UTF_8));
+        try (FileOutputStream fos = new FileOutputStream(file);
+             BufferedWriter bw = new BufferedWriter(new OutputStreamWriter(fos, UTF_8))) {
             bw.write(body);
-        } finally {
-            if (bw != null) {
-                bw.close();
-            }
-            fos.close();
         }
     }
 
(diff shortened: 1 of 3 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStatus.java
```diff
@@ -123,16 +123,9 @@ synchronized void writeToDirectories(List<File> directories) {
      * @throws IOException
      */
     private static void writeToFile(File file, String body) throws IOException {
-        FileOutputStream fos = new FileOutputStream(file);
-        BufferedWriter bw = null;
-        try {
-            bw = new BufferedWriter(new OutputStreamWriter(fos, UTF_8));
+        try (FileOutputStream fos = new FileOutputStream(file);
+             BufferedWriter bw = new BufferedWriter(new OutputStreamWriter(fos, UTF_8))) {
```
sijie:
```suggestion
             try (BufferedWriter bw = new BufferedWriter(new OutputStreamWriter(fos, UTF_8)))) {
```

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Cookie.java
```diff
@@ -217,16 +217,9 @@ public void writeToDirectory(File directory) throws IOException {
     }
 
     public void writeToFile (File versionFile) throws IOException {
-        FileOutputStream fos = new FileOutputStream(versionFile);
-        BufferedWriter bw = null;
-        try {
-            bw = new BufferedWriter(new OutputStreamWriter(fos, UTF_8));
+        try (FileOutputStream fos = new FileOutputStream(versionFile);
+             BufferedWriter bw = new BufferedWriter(new OutputStreamWriter(fos, UTF_8))) {
```
sijie: put this in a try block as well

sijie:

sijie:

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Cookie.java
liketic: Thanks @sigee , are you suggest we add a try block for each closable resource here? I think that's won't be necessary. As this [doc](https://www.oracle.com/technetwork/articles/java/trywithresources-401775.html) said: ```Any resource declared within a try block opening will be closed.```, so one try-block is enough. WDYT?

sijie: I misread the code. you are right. +1 #shipit

PR #1963, opened 2019-02-25
Add http method test for vertx-http server

rdhabalia:
### Motivation

As discussed at #1953, adding test to verify all http-method works for vertx-http-server.


rdhabalia:

rdhabalia:

Issue #634, opened 2017-10-17
Shell cmd to audit the ledgers and list bookie id requires updation


ivankelly:
JIRA: https://issues.apache.org/jira/browse/BOOKKEEPER-808

Reporter: Rakesh R 

The idea is to expose shell command to list the ledgers which needs to be updated using the 'updateledgers' command. This will be helpful for the administrators to ensure that the bookie id updation is completed.

### Comments from JIRA

---
*Sijie Guo* 2016-03-14T05:56:03.209+0000

[~rakesh_r] what do you think about this? shall we move to 4.5.0?

---
*Rakesh R* 2016-03-14T06:10:36.798+0000

Apart from the doc update this is the only known pending task for this feature. Can we move this jira to {{4.4.1}} minor release, considering that most of the major code changes for this feature is in and is available in 4.4.0(trunk).

---
*Venkateswararao Jujjuri (JV)* 2016-03-15T16:39:14.287+0000

anyone actively working on this?? [~reddycharan18@gmail.com] can you check if this is something you are looking into?

---
*Rakesh R* 2016-03-16T07:25:51.023+0000

bq. anyone actively working on this?? 
[~jujjuri] Would you be interested in proposing a patch. I'm happy to help in reviews.

ivankelly:
bumping to 4.8.0
maybe this'll be done as part of the update to the cli (BP-27)

sijie:
This has been fixed by #1968

PR #1968, opened 2019-03-01
Tool to search and replace bookie ids in ledger metadata

ivankelly:
To use:
```
bin/bkctl bookieid searchreplace --from <from> --to <to>
```

To be used in cases where the DNS name of the bookie has to change, and you don't want all the data to have to be moved by autorecovery.

### DIFF
--- a/tools/ledger/src/main/java/org/apache/bookkeeper/tools/cli/commands/bookieid/SearchReplaceBookieIdCommand.java
+++ b/tools/ledger/src/main/java/org/apache/bookkeeper/tools/cli/commands/bookieid/SearchReplaceBookieIdCommand.java
@@ -79,50 +79,51 @@ public class SearchReplaceBookieIdCommand extends ClientCommand<SearchReplaceBoo
 
     @Override
     protected void run(BookKeeper bk, Flags flags) throws Exception {
-        BookKeeperAdmin admin = new BookKeeperAdmin((org.apache.bookkeeper.client.BookKeeper) bk);
-        LedgerManager ledgerManager = ((org.apache.bookkeeper.client.BookKeeper) bk).getLedgerManager();
-        long i = 0;
+        try (BookKeeperAdmin admin = new BookKeeperAdmin((org.apache.bookkeeper.client.BookKeeper) bk)) {
+            LedgerManager ledgerManager = ((org.apache.bookkeeper.client.BookKeeper) bk).getLedgerManager();
+            long i = 0;
 
-        BookieSocketAddress fromAddr = new BookieSocketAddress(flags.from);
-        BookieSocketAddress toAddr = new BookieSocketAddress(flags.to);
-        System.out.println(String.format("Replacing bookie id %s with %s in metadata", fromAddr, toAddr));
-        RateLimiter limiter = RateLimiter.create(flags.rate);
-        for (Long lid : admin.listLedgers()) {
-            Versioned<LedgerMetadata> md = ledgerManager.readLedgerMetadata(lid).get();
-            if (md.getValue().getAllEnsembles().entrySet().stream().anyMatch(e -> e.getValue().contains(fromAddr))) {
-                limiter.acquire();
+            BookieSocketAddress fromAddr = new BookieSocketAddress(flags.from);
+            BookieSocketAddress toAddr = new BookieSocketAddress(flags.to);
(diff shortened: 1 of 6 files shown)


tools/ledger/src/main/java/org/apache/bookkeeper/tools/cli/commands/bookieid/SearchReplaceBookieIdCommand.java
```diff
@@ -0,0 +1,128 @@
...
+        super(CliSpec.<Flags>newBuilder()
+            .withName(NAME)
+            .withDescription(DESC)
+            .withFlags(flags)
+            .build());
+    }
+
+    @Override
+    protected void run(BookKeeper bk, Flags flags) throws Exception {
+        BookKeeperAdmin admin = new BookKeeperAdmin((org.apache.bookkeeper.client.BookKeeper) bk);
```
sijie: might be worth closing admin at the end?

tools/ledger/src/main/java/org/apache/bookkeeper/tools/cli/commands/bookieid/SearchReplaceBookieIdCommand.java
```diff
@@ -0,0 +1,128 @@
...
+                                    } else {
+                                        return a;
+                                    }
+                                });
+                            builder.replaceEnsembleEntry(e.getKey(), ensemble);
+                        });
+                LedgerMetadata newMeta = builder.build();
+                if (flags.verbose) {
+                    System.out.println("Replacing ledger " + lid + " metadata ...");
+                    System.out.println(md.getValue().toString());
```
sijie: use `toSafeString()`?

tools/ledger/src/main/java/org/apache/bookkeeper/tools/cli/commands/bookieid/SearchReplaceBookieIdCommand.java
```diff
@@ -0,0 +1,128 @@
...
+                    System.out.println("with ...");
+                    System.out.println(newMeta.toString());
```
sijie: use `toSafeString()`?

jvrao:
@ivankelly  @sijie  what is the use case of this utility? can you please give little more info ?

sijie:
@jvrao @eolivelli I thought it is related to #634 , but after looking into BookieShell, there is already a comment `updateledgers`. so I am also a bit confused about this PR. @ivankelly can you explain more about the context?

ivankelly:
@jvrao @sijie @eolivelli The usecase is that a client needed to update the DNS addresses of their bookies. The data already existed on the bookie, but the DNS and advertised address was changing (I had to update the cookies manually with sed, it was a stateful set in k8s). I wanted to avoid autorecovery trying to recovery every existing ledger, since the data was still where it should have been.

```updateledgers``` does exist, but it's very limited. It only allows changing from ip to hostname or vice versa based on what is in the configuration file. Even on baremetal, this is awkward to use, but on k8s it's completely non-viable.

**Comment by ivankelly** [comment]
run pr validation

**Comment by ivankelly** [comment]
> Can you also provide a command to update the cookies (as you did with 'sed')?

I don't have the cycles to do so right now (with good testing) but i'll put it on the backlog.

**Comment by sijie** [comment]
@ivankelly 

there is one checkstyle issue, can you fix it?

```
[ERROR] /home/jenkins/jenkins-slave/workspace/bookkeeper_precommit_pullrequest_validation/tools/ledger/src/main/java/org/apache/bookkeeper/tools/cli/commands/bookieid/SearchReplaceBookieIdCommand.java:92: Line is longer than 120 characters (found 121). [LineLength]
```

**Comment by jiazhai** [comment]
@ivankelly 
Is it good to add your use case into the PR description? 
```
The usecase is that a client needed to update the DNS addresses of their bookies. The data already existed on the bookie, but the DNS and advertised address was changing (I had to update the cookies manually with sed, it was a stateful set in k8s). I wanted to avoid autorecovery trying to recovery every existing ledger, since the data was still where it should have been.

updateledgers does exist, but it's very limited. It only allows changing from ip to hostname or vice versa based on what is in the configuration file. Even on baremetal, this is awkward to use, but on k8s it's completely non-viable.
```

sijie:
run pr validation

PR #1971, opened 2019-03-01
DNM/RFC ISSUE-1967: make ledger creation and removal robust to zk connection loss

athanatos:
The bookkeeper project ZooKeeperClient wrapper for the ZooKeeper client
will resend zk node creations and removals upon reconnect after a
ConnectionLoss event. In the event that the original succeeded, the
resent operation will erroneously return LedgerExistException or
NoSuchLedgerExistsException for creation and removal respectively.

For removal, this patch limits the operation by allowing it to always
succeed if the ledger does not exist in order to make it idempotent.
This is appears to be the simplest solution as exclusive removal isn't
important.

**Note, the above is an actual change to the bk client semantics**

For creation, exclusive creation is cleary important for correctness,
so this patch adds a creatorId field to the LedgerMetdata to
disambiguate the above race from a real race. For
AbstractZkLedgerManager, this is simply the zk session id.

There's an oportunity for optimization with the above if exclusive
ledger creation failures are expected to be common.  You only actually
need to perform this check if the operation was really resent.  I chose
not to go this route yet because it would require messing with the
ZooKeeperClient interface to surface that information without burdening
other callers.

If the client is set to version 2 or older, this field will be ignored
and the old behavior will be retained.  If the client is version 3 or
newer but creation races with an older client, the new client will
interpret the nonce to be BLANK and thereby detect the race correctly.

DNM: this patch still needs tests.

(@bug W-4077048@)

### DIFF
--- a/bookkeeper-proto/src/main/proto/DataFormats.proto
+++ b/bookkeeper-proto/src/main/proto/DataFormats.proto
@@ -60,6 +60,8 @@ message LedgerMetadataFormat {
         optional bytes value = 2;
     }
     repeated cMetadataMapEntry customMetadata = 11;
+
+    optional int64 creatorId = 12;
 }
 
 message LedgerRereplicationLayoutFormat {
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadataBuilder.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadataBuilder.java
@@ -63,6 +63,9 @@ public class LedgerMetadataBuilder {
     private boolean storeCtime = false;
     private Map<String, byte[]> customMetadata = Collections.emptyMap();
 
+    private static final long BLANK_CREATOR_ID = 0;
+    private long creatorId = BLANK_CREATOR_ID;
+
     public static LedgerMetadataBuilder create() {
         return new LedgerMetadataBuilder();
     }
@@ -181,6 +184,11 @@ public class LedgerMetadataBuilder {
         return this;
     }
 
+    public LedgerMetadataBuilder withCreatorId(long id) {
+        this.creatorId = id;
+        return this;
+    }
+
     public LedgerMetadata build() {
         checkArgument(ensembleSize >= writeQuorumSize, "Write quorum must be less or equal to ensemble size");
         checkArgument(writeQuorumSize >= ackQuorumSize, "Write quorum must be greater or equal to ack quorum");
@@ -189,6 +197,7 @@ public class LedgerMetadataBuilder {
                                       ensembleSize, writeQuorumSize, ackQuorumSize,
                                       state, lastEntryId, length, ensembles,
                                       digestType, password, ctime, storeCtime,
+                                      creatorId,
                                       customMetadata);
     }
 
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadataImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadataImpl.java
@@ -69,6 +69,8 @@ class LedgerMetadataImpl implements LedgerMetadata {
 
     private final Map<String, byte[]> customMetadata;
 
+    private long creatorId;
+
     LedgerMetadataImpl(int metadataFormatVersion,
                        int ensembleSize,
                        int writeQuorumSize,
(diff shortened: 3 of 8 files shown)


athanatos:
This patch is a crack at hardening ledger creation and removal semantics against zk connection loss events.  There are some particular drawbacks I'd like feedback on before going further:
1) Ledger removal here is modified to be idempotent by always succeeding if the ledger does not exist after the call whether it existed before or not.  *This is a change to the bk client semantics.*
2) The fix for ledger creation involves a modification to the LedgerMetadata.  Version 3 permits this kind of modification easily enough since old clients will ignore the field, but it makes backports impractical.

asfgit:
SUCCESS 
 
--none--

athanatos:
run integration tests

bookkeeper-proto/src/main/proto/DataFormats.proto
```diff
@@ -60,6 +60,8 @@ message LedgerMetadataFormat {
         optional bytes value = 2;
     }
     repeated cMetadataMapEntry customMetadata = 11;
+
+    optional int64 creatorId = 12;
```
sijie: we need to a flag to control whether to add `creatorId` or not. since we are still using text format for storing ledger metadata.

athanatos: This patch only added it to Version 3 which afaict is using the normal protobuf format, so old clients should ignore it.  Is that right?

sijie: @athanatos 

sorry I misread that. you are right.

athanatos: @sijie ^ ?

sijie: I thought I have approved the pull request since you have answered my question :)

athanatos: Oh, I think I added my comment without reloading the page.  Sorry!

reddycharan:
@athanatos @sijie there is one fundamental flaw with this fix, if bookkeeper client tries to create a new ledger with ledgerid already existing, which was successfully created by this client earlier, then it will succeed because of the same creatorid (zk.sessionId) in this case and it is not expected. So we should rethink on creatorId and it has to be some thing more than zk.sessionid.

athanatos:
@reddycharan is correct.  Today is my last day at SF, so @reddycharan will be taking over this PR.

jiazhai:

jiazhai:
@reddycharan Are we going to fix the flaw in this PR?

reddycharan:
@jiazhai i'm taking over this change. Probably will create a new PR with fix and testcode. For now, I would leave this PR as it is and keep it open for comments.

sijie:
Close this pull request since @reddycharan already addressed it in #2006

ivankelly:
@sijie @athanatos This change changed the default metadata from text to binary. This should have been mentioned loudly in the release notes for 4.10.0 as it has a major impact on whether a cluster can roll back after an upgrade.

eolivelli:
@ivankelly you are totaly right.
I am sorry I wrote those release notes.
Unfortunately from the title of the PR it was not clear that we were doing the change you noticed.

I wonder if it is better to release a 4.10.1 version with a more conservative default.
But it may create more mess

ivankelly:
I don't think it would be a bad thing to switch CURRENT version back to 2, but then we need a way for clients to change it to 3.

I think a method for clients to change it to change the default metadata version is necessary anyhow. Lets say you want to upgrade from 4.9 to 4.10, and you're well aware of this change. You want to upgrade in a way that you can possibly rollback. There's a lot of changes in 4.10, so it would be nice to be able upgrade to 4.10 without default metadata version at 2, burn in the new version for a while, and then when you see 4.10 is ok, flip the switch up start using binary metadata.

I think a user configuration for this is the best option. Clusterwise upgrade to 2->3 creates another big jumping off point. We would also need a tool to roll back binary ledgers to format 2 in case everything goes to hell.

eolivelli:
Yes
It would have been better to have some knob.
In my opinion we always take care about the 'rollback  story'

@jvrao @reddycharan

ivankelly:
Actually, rollback should still be possible as long as there's no client using anything less than 4.9.0 which is when the binary stuff went in. It's still a very unexpected change though and should have been called out loudly.

Issue #1970, opened 2019-03-01
Exception when sending read requests


merlimat:
**BUG REPORT**

***Describe the bug***

Got reports from Pulsar users (with BK 4.9.0) of these errors: 

```
17:24:16.302 [BookKeeperClientWorker-OrderedExecutor-0-0] ERROR org.apache.bookkeeper.common.util.SafeRunnable - Unexpected throwable caught 
java.lang.ArrayIndexOutOfBoundsException: -2
    at com.google.common.collect.RegularImmutableList.get(RegularImmutableList.java:60) ~[com.google.guava-guava-21.0.jar:?]
    at org.apache.bookkeeper.client.PendingReadOp$SequenceReadRequest.sendNextRead(PendingReadOp.java:400) ~[org.apache.bookkeeper-bookkeeper-server-4.9.0.jar:4.9.0]
    at org.apache.bookkeeper.client.PendingReadOp$SequenceReadRequest.read(PendingReadOp.java:382) ~[org.apache.bookkeeper-bookkeeper-server-4.9.0.jar:4.9.0]
    at org.apache.bookkeeper.client.PendingReadOp.initiate(PendingReadOp.java:526) ~[org.apache.bookkeeper-bookkeeper-server-4.9.0.jar:4.9.0]
    at org.apache.bookkeeper.client.PendingReadOp.safeRun(PendingReadOp.java:536) ~[org.apache.bookkeeper-bookkeeper-server-4.9.0.jar:4.9.0]
    at org.apache.bookkeeper.common.util.SafeRunnable.run(SafeRunnable.java:36) [org.apache.bookkeeper-bookkeeper-common-4.9.0.jar:4.9.0]
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149) [?:1.8.0_181]
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624) [?:1.8.0_181]
    at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30) [io.netty-netty-all-4.1.32.Final.jar:4.1.32.Final]
    at java.lang.Thread.run(Thread.java:748) [?:1.8.0_181]
```

eolivelli:
Do you already have a reproducer?

merlimat:
No, I haven't seen this error myself yet

grantwwu:
[...]

PR #1973, opened 2019-03-04
[DLOG] Avoid double read in readahead

ivankelly:
There was a bug in the ReadAheadReader whereby, if it was in a paused
state and the last segment it had read had been closed, then when a
writer started writing new segments, the ReadAheadReader would issue a
double read. This would end up triggering an error when it got to the
end of the segment as the issued one of the issued reads would then be
cancelled when the segment was closed.

This fix adds an explicit state machine to the reader and moves all
state transitions to run on the ordered executor, so only one read can
ever be issued at a time.

### DIFF
--- a/stream/distributedlog/core/src/main/java/org/apache/distributedlog/ReadAheadEntryReader.java
+++ b/stream/distributedlog/core/src/main/java/org/apache/distributedlog/ReadAheadEntryReader.java
@@ -67,6 +67,15 @@ class ReadAheadEntryReader implements
 
     private static final Logger logger = LoggerFactory.getLogger(ReadAheadEntryReader.class);
 
+    enum State {
+        IDLE,
+        READING,
+        PAUSED,
+        CLOSED,
+        ERROR
+    }
+    private State state = State.IDLE;
+
     //
     // Static Functions
     //
@@ -460,6 +469,11 @@ class ReadAheadEntryReader implements
     }
 
     private void unsafeAsyncClose(CompletableFuture<Void> closePromise) {
+        if (logger.isDebugEnabled()) {
+            logger.debug("[{}][state:{}] Closing read ahead", streamName, state);
+        }
+        state = State.CLOSED;
+
         List<CompletableFuture<Void>> closeFutures = Lists.newArrayListWithExpectedSize(
                 segmentReaders.size() + segmentReadersToClose.size() + 1);
         if (null != currentSegmentReader) {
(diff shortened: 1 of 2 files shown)


eolivelli:
@ivankelly did you see some error which is getting fixed by this patch ? I can't see any issue attached.

Is this problem affecting production with 4.9 or is only a case that you found analysing the code ?

cc @diegosalvi FYI

eolivelli:
is it https://github.com/apache/bookkeeper/issues/1970 ?

ivankelly:
@eolivelli the issue was observed by a customer running 4.7.3, but the code hasn't changed much since. The attached unit test will trigger it if you want to see the problem in action.

stream/distributedlog/core/src/main/java/org/apache/distributedlog/ReadAheadEntryReader.java
```diff
@@ -67,6 +67,15 @@
 
     private static final Logger logger = LoggerFactory.getLogger(ReadAheadEntryReader.class);
 
+    enum State {
```
eolivelli: private ?

stream/distributedlog/core/src/main/java/org/apache/distributedlog/ReadAheadEntryReader.java
```diff
@@ -942,25 +948,60 @@ private void unsafeMoveToNextLogSegment() {
...
+                    case PAUSED:
+                        if (!cacheFull && hasMoreSegments) {
+                            reader.readNext().whenComplete(ReadAheadEntryReader.this);
+                            state = State.READING;
+                        }
+                        break;
+                    case READING:
+                    case ERROR:
+                    case CLOSED:
+                        // do nothing
```
eolivelli: shall a log at trace level be useful ?

ivankelly: nah, we already know if we're going to hit this branch from the debug log above.

stream/distributedlog/core/src/main/java/org/apache/distributedlog/ReadAheadEntryReader.java
ivankelly: updated

sijie:
please also cherry-pick the change to 4.9.1 and 4.8.3

ivankelly:

Issue #1974, opened 2019-03-05
Migrate command `bookieformat` from shell to bkctl


zymap:
**FEATURE REQUEST**

1. Please describe the feature you are requesting.

- Migrate command `bookieformat` from shell to bkctl.

2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have).
   Are you currently using any workarounds to address this issue?

- nice-to-have.

3. Provide any additional detail on your proposed use case for this feature.

```
bin/bkctl bookie format -h
Usage:  bkctl bookie format [flags]

Flags:

    -d, --deletecookie
        Delete its cookie on metadata store.

    -f, --force
        If [noninteractive] is specified, then whetherto force delete the old
        data without prompt?

    -n, --noninteractive
        Whether to confirm if old data exists?


    -h, --help
        Display help information

```

PR #1975, opened 2019-03-05
Migrate command `bookieformat`

zymap:

Migrate command `bookieformat`.

### Motivation

#1974 

### Changes

- Add command `bookieformat` to bkctl
- Replace command in shell

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/commands/bookie/FormatCommand.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/commands/bookie/FormatCommand.java
@@ -1,0 +1,113 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/commands/bookie/FormatCommand.java
```diff
@@ -0,0 +1,111 @@
...
+            try {
+                runFunctionWithRegistrationManager(conf, rm -> {
+
+                    try {
+                        Versioned<Cookie> cookie = Cookie.readFromRegistrationManager(rm, bfconf);
+                        cookie.getValue().deleteFromRegistrationManager(rm, bfconf, cookie.getVersion());
+                    } catch (BookieException.CookieNotFoundException nne) {
+                        LOG.warn("No cookie to remove : ", nne);
+                    } catch (BookieException e) {
+                        e.printStackTrace();
```
sijie: this will shallow the exception. I would suggest `throw new UncheckedExecutionException(be.getMessage(), be);`

bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/commands/bookie/FormatCommand.java
```diff
@@ -0,0 +1,111 @@
...
+                    }
+
+                    return null;
+                });
+            } catch (MetadataException e) {
+                e.printStackTrace();
```
sijie: can you throw exception?

bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/commands/bookie/FormatCommand.java
```diff
@@ -0,0 +1,111 @@
...
+            } catch (ExecutionException e) {
+                e.printStackTrace();
```
sijie: throw exception?

zymap:
@sijie  ok . I will fix it. Thank you for your review. :)

PR #1995, opened 2019-03-19
Added HTTP handler to expose bookie state

merlimat:
### Motivation

Allow to check the state of the bookie through HTTP.

The main reason is to allow light-weight probes that can be run frequently to establish whether the bookie is running and meets a minimum readiness criteria. One example is to automatically wait until the bookie has completed the startup sequence.

### Changes

Added new HTTP handler that exposes the values from the bookie `StateManager`.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/server/http/service/BookieIsReadyService.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/server/http/service/BookieIsReadyService.java
@@ -20,51 +20,22 @@ package org.apache.bookkeeper.server.http.service;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
-import lombok.Data;
-import lombok.NoArgsConstructor;
-
 import org.apache.bookkeeper.bookie.Bookie;
 import org.apache.bookkeeper.bookie.StateManager;
-import org.apache.bookkeeper.common.util.JsonUtil;
 import org.apache.bookkeeper.http.HttpServer;
 import org.apache.bookkeeper.http.service.HttpEndpointService;
 import org.apache.bookkeeper.http.service.HttpServiceRequest;
 import org.apache.bookkeeper.http.service.HttpServiceResponse;
 
 /**
- * HttpEndpointService that exposes the current state of the bookie.
- *
- * <p>Get the current bookie status:
- *
- * <pre>
- * <code>
- * {
- *  "running" : true,
- *  "readOnly" : false,
- *  "shuttingDown" : false,
- *  "availableForHighPriorityWrites" : true
- *}
- * </code>
- * </pre>
+ * HttpEndpointService that returns 200 if the bookie is ready.
  */
-public class BookieStateService implements HttpEndpointService {
+public class BookieIsReadyService implements HttpEndpointService {
 
     private final Bookie bookie;
 
-    public BookieStateService(Bookie bookie) {
+    public BookieIsReadyService(Bookie bookie) {
         this.bookie = checkNotNull(bookie);
-    }
-
-    /**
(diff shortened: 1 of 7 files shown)


eolivelli:
@jvrao fyi

merlimat:
run integration tests

bookkeeper-server/src/main/java/org/apache/bookkeeper/server/http/service/BookieStateService.java
```diff
@@ -0,0 +1,94 @@
...
+        private boolean availableForHighPriorityWrites;
+    }
+
+    @Override
+    public HttpServiceResponse handle(HttpServiceRequest request) throws Exception {
+        HttpServiceResponse response = new HttpServiceResponse();
+
+        if (HttpServer.Method.GET != request.getMethod()) {
+            response.setCode(HttpServer.StatusCode.NOT_FOUND);
+            response.setBody("Only support GET method to retrieve GC details."
```
sijie: the content in the body doesn't match the endpoint.

merlimat:
Updated based on comments. I also added a separate handler `/api/v1/bookie/is_ready`. The reason is that to use this a readyness probe, systems like K8S expects a simple 200 response or error if the process is not ready. Having the JSON response makes it difficult to integrate with that.

bookkeeper-server/src/test/java/org/apache/bookkeeper/server/http/TestHttpService.java
```diff
@@ -826,4 +829,30 @@ public void testGCDetailsService() throws Exception {
...
+    }
+
+    @Test
+    public void testGetBookieIsReady() throws Exception {
+        HttpEndpointService bookieStateServer = bkHttpServiceProvider
+                .provideHttpEndpointService(HttpServer.ApiType.BOOKIE_IS_READY);
+
+        HttpServiceRequest request1 = new HttpServiceRequest(null, HttpServer.Method.GET, null);
+        HttpServiceResponse response1 = bookieStateServer.handle(request1);
+        assertEquals(HttpServer.StatusCode.OK.getValue(), response1.getStatusCode());
```
eolivelli: Can you add a test case for the not-ready case?

bookkeeper-server/src/main/java/org/apache/bookkeeper/server/http/service/BookieIsReadyService.java
```diff
@@ -0,0 +1,60 @@
...
+            response.setBody("Only support GET method check if bookie is ready.");
+            return response;
+        }
+
+        StateManager sm = bookie.getStateManager();
+        if (sm.isRunning()) {
+            response.setCode(HttpServer.StatusCode.OK);
+        } else {
+            response.setCode(HttpServer.StatusCode.SERVICE_UNAVAILABLE);
+            response.setBody("Bookie is not fully started yet");
```
sijie: if a bookie is shutting down, `isRunning` is also false. then the message is not correct. Can we check if the state manager is shutting down or not, and return the proper error message?

bookkeeper-server/src/test/java/org/apache/bookkeeper/server/http/TestHttpService.java
merlimat: Added

bookkeeper-server/src/main/java/org/apache/bookkeeper/server/http/service/BookieIsReadyService.java
merlimat: Fixed

merlimat:
@sijie @eolivelli Address the comments. Please take a final look.

merlimat:

PR #1996, opened 2019-03-19
In DbLedgerStorage use default values when config key is present but empty

merlimat:
### Motivation

Currently setting the `dbStorage_writeCacheMaxSizeMb=` with empty value is making the
DbLedgerStorage initialize to fail since the empty string is being parsed as long. 

Instead, we should just apply the default value as in the case where the config key is not there.


sijie:

PR #1998, opened 2019-03-24
fix potential NPE when releasing entry that is null

jerrypeng:

### Motivation
A interrupt exception can occur during the poll operation of the blocking and cause a NPE to be thrown

### Changes
Check if entry is null before trying to release it

### DIFF
--- a/stream/distributedlog/core/src/main/java/org/apache/distributedlog/impl/logsegment/BKLogSegmentEntryReader.java
+++ b/stream/distributedlog/core/src/main/java/org/apache/distributedlog/impl/logsegment/BKLogSegmentEntryReader.java
@@ -800,7 +800,9 @@ public class BKLogSegmentEntryReader implements SafeRunnable, LogSegmentEntryRea
                         return;
                     }
                 } finally {
-                    removedEntry.release();
+                    if (removedEntry != null) {
+                        removedEntry.release();
+                    }
                 }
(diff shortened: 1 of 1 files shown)


jerrypeng:
@merlimat @ivankelly @sijie @jiazhai please review. Thanks!

stream/distributedlog/core/src/main/java/org/apache/distributedlog/impl/logsegment/BKLogSegmentEntryReader.java
```diff
@@ -800,7 +800,9 @@ private void readEntriesFromReadAheadCache(PendingReadRequest nextRequest) {
                         return;
                     }
                 } finally {
-                    removedEntry.release();
+                    if (removedEntry != null) {
+                        removedEntry.release();
+                    }
```
merlimat: Alternately: `ReferenceCountUtils.safeRelease(removedEntry)`

jerrypeng: will change

sijie:
@jerrypeng can you fix the checkstyle error?

ivankelly:
@eolivelli I didn't cherrypick this to 4.9 branch as there's a release in progress. if there's another RC this should be pulled into 4.9.1, otherwise 4.9.2.

eolivelli:
@ivankelly sure.

PR #1999, opened 2019-03-24
Fix IDE complain as there are multi choices for error code

liketic:
There are two ```Code``` in ```org.apache.bookkeeper.client. BKException```:

The first is https://github.com/apache/bookkeeper/blob/f26a4cae0e9205ad391c6d4d79f2937871864c28/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BKException.java#L136

The second is https://github.com/apache/bookkeeper/blob/f26a4cae0e9205ad391c6d4d79f2937871864c28/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/BKException.java#L180

BKException's subclasses cannot auto choice which ```Code```.


jiazhai:
@eolivelli It may be idea, Seems in my idea, it also complaining it.

image

PR #2006, opened 2019-03-27
ISSUE-1967: make ledger creation and removal robust to zk connectionloss

reddycharan:

The bookkeeper project ZooKeeperClient wrapper for the ZooKeeper client
will resend zk node creations and removals upon reconnect after a
ConnectionLoss event. In the event that the original succeeded, the
resent operation will erroneously return LedgerExistException or
NoSuchLedgerExistsException for creation and removal respectively.

For removal, this patch limits the operation by allowing it to always
succeed if the ledger does not exist in order to make it idempotent.
This is appears to be the simplest solution as exclusive removal isn't
important.

**Note, the above is an actual change to the bk client semantics**

For creation, exclusive creation is cleary important for correctness,
so this patch adds a creator token field to the LedgerMetdata to
disambiguate the above race from a real race. For
AbstractZkLedgerManager, this is simply a random long value.

There's an oportunity for optimization with the above if exclusive
ledger creation failures are expected to be common.  You only actually
need to perform this check if the operation was really resent.  I chose
not to go this route yet because it would require messing with the
ZooKeeperClient interface to surface that information without burdening
other callers.

If the client is set to version 2 or older, this field will be ignored
and the old behavior will be retained.  If the client is version 3 or
newer but creation races with an older client, the new client will
interpret the nonce to be BLANK and thereby detect the race correctly.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractZkLedgerManager.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractZkLedgerManager.java
@@ -33,9 +33,11 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.TimeUnit;
 
 import org.apache.bookkeeper.client.BKException;
+import org.apache.bookkeeper.client.LedgerMetadataBuilder;
 import org.apache.bookkeeper.client.api.LedgerMetadata;
 import org.apache.bookkeeper.common.concurrent.FutureUtils;
 import org.apache.bookkeeper.conf.AbstractConfiguration;
@@ -247,8 +249,19 @@ public abstract class AbstractZkLedgerManager implements LedgerManager, Watcher 
     }
 
     @Override
-    public CompletableFuture<Versioned<LedgerMetadata>> createLedgerMetadata(long ledgerId, LedgerMetadata metadata) {
+    public CompletableFuture<Versioned<LedgerMetadata>> createLedgerMetadata(long ledgerId,
+                                                                             LedgerMetadata inputMetadata) {
         CompletableFuture<Versioned<LedgerMetadata>> promise = new CompletableFuture<>();
+        /*
+         * Create a random number and use it as creator token.
+         */
+        final long cToken = ThreadLocalRandom.current().nextLong(Long.MAX_VALUE);
+        final LedgerMetadata metadata;
+        if (inputMetadata.getMetadataFormatVersion() > 2) {
+            metadata = LedgerMetadataBuilder.from(inputMetadata).withCToken(cToken).build();
+        } else {
+            metadata = inputMetadata;
+        }
         String ledgerPath = getLedgerPath(ledgerId);
         StringCallback scb = new StringCallback() {
             @Override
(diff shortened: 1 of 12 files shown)


reddycharan:
@sijie I'm taking over this PR - https://github.com/apache/bookkeeper/pull/1971. So creating new PR.

Fixed creator token logic and added required functional testcases.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/LedgerMetadataSerDe.java
```diff
@@ -78,7 +78,7 @@
     public static final int METADATA_FORMAT_VERSION_3 = 3;
 
     public static final int MAXIMUM_METADATA_FORMAT_VERSION = METADATA_FORMAT_VERSION_3;
-    public static final int CURRENT_METADATA_FORMAT_VERSION = METADATA_FORMAT_VERSION_2;
+    public static final int CURRENT_METADATA_FORMAT_VERSION = METADATA_FORMAT_VERSION_3;
```
reddycharan: for now I changed the metadata version. I'm open for feedback, regarding if it is ok to bump hardcoded metadata version, make it configurable or something else..

reddycharan:
@sijie @jvrao @eolivelli requesting review.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/LedgerMetadataSerDe.java
reddycharan: @sijie just want to confirm that you are fine with this hardcoded value change.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractZkLedgerManager.java
```diff
@@ -247,17 +249,59 @@ public void process(WatchedEvent event) {
...
-    public CompletableFuture<Versioned<LedgerMetadata>> createLedgerMetadata(long ledgerId, LedgerMetadata metadata) {
+    public CompletableFuture<Versioned<LedgerMetadata>> createLedgerMetadata(long ledgerId,
+                                                                             LedgerMetadata inputMetadata) {
         CompletableFuture<Versioned<LedgerMetadata>> promise = new CompletableFuture<>();
+        /*
+         * Create a random number and use it as creator token.
+         */
+        final long cToken = ThreadLocalRandom.current().nextLong(Long.MAX_VALUE);
+        final LedgerMetadata metadata;
+        if (inputMetadata.getMetadataFormatVersion() > 2) {
```
eolivelli: Why aren't you using a constant for '2'?

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractZkLedgerManager.java
```diff
@@ -247,17 +249,59 @@ public void process(WatchedEvent event) {
...
+                        CompletableFuture<Versioned<LedgerMetadata>> readFuture = readLedgerMetadata(ledgerId);
+                        readFuture.handle((readMetadata, exception) -> {
+                            if (exception == null) {
+                                if (readMetadata.getValue().getCToken() == cToken) {
+                                    FutureUtils.complete(promise, new Versioned<>(metadata, new LongVersion(0)));
+                                } else {
+                                    LOG.warn("Failed to create ledger metadata for {} which already exists", ledgerId);
+                                    promise.completeExceptionally(new BKException.BKLedgerExistException());
+                                }
+                            } else if (exception instanceof KeeperException.NoNodeException) {
```
eolivelli: Can we see here a CompletionException?

reddycharan:
run integration tests

reddycharan:

reddycharan:

reddycharan:
run integration tests

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractZkLedgerManager.java
reddycharan: there is a else block for all other kind of exceptions

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractZkLedgerManager.java
reddycharan: fixed it

reddycharan:

reddycharan:
@sijie @eolivelli finally im able to fix backward compat test failures. It should be good to go. Can you please sign-off on it, so I can merge it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractZkLedgerManager.java
eolivelli: CompletionException may wrap one of the exceptions you are checking, like KeeperExceptio.NoNodeException.
Are you sure when don't need to handle it explicitly, by unwrapping the hidden exception?

tests/backward-compat/yahoo-custom-version/src/test/groovy/org/apache/bookkeeper/tests/backwardcompat/TestCompatUpgradeYahooCustom.groovy
```diff
@@ -170,11 +170,16 @@ class TestCompatUpgradeYahooCustom {
...
+            // Since METADATA_VERSION is upgraded and it is using binary format, the older
+            // clients which are expecting text format would fail to read ledger metadata.
             def ledger5 = createAndWrite(currentCL, currentBK)
             ledgers.add(ledger5.getId())
-            openAndVerifyEntries(yahooCL, yahooBK, ledger5.getId())
-            assertCantWrite(currentCL, ledger5)
+            try {
+                openAndVerifyEntries(yahooCL, yahooBK, ledger5.getId())
+            } catch (Exception exc) {
+                Assert.assertEquals(exc.getClass().getName(),
```
eolivelli: Do we still need this test?
This is telling now that we are not compatible by default

reddycharan: I would prefer to keep this. I would see it as more like with this change it is expected to fail with the specific exception. Also, it would serve as record (history) that this change had broke backward compatibility.

reddycharan:
Thanks @sijie @eolivelli for your feedback. I'm going to merge this PR.

reddycharan:
IGNORE IT CI

reddycharan:
IGNORE CI

PR #2008, opened 2019-03-28
Migrate command `whoisauditor`

zymap:

- Using `bkctl` run command `whoisauditor`

### Motivation

#2007


zymap:
run integration tests

tools/ledger/src/main/java/org/apache/bookkeeper/tools/cli/commands/ToggleCommandGroup.java
```diff
@@ -0,0 +1,46 @@
...
+
+import org.apache.bookkeeper.tools.cli.commands.toggle.WhoIsAuditorCommand;
+import org.apache.bookkeeper.tools.common.BKFlags;
+import org.apache.bookkeeper.tools.framework.CliCommandGroup;
+import org.apache.bookkeeper.tools.framework.CliSpec;
+
+/**
+ * Commands on some specific operation.
+ */
+public class ToggleCommandGroup extends CliCommandGroup<BKFlags> {
```
sijie: I think this would be part of `recovery` group?

bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/commands/toggle/WhoIsAuditorCommand.java
```diff
@@ -0,0 +1,90 @@
...
+    @Override
+    public boolean apply(ServerConfiguration conf, CliFlags cmdFlags) {
+        try {
+            return handler(conf);
+        } catch (Exception e) {
+            throw new UncheckedExecutionException(e.getMessage(), e);
+        }
+    }
+
+    public boolean handler(ServerConfiguration conf)
```
sijie: private boolean

also `handler` is a too generic name. it would be better to call it `getAuditor`?

zymap: fixed.

PR #2017, opened 2019-03-29
Docker autobuild hook

ivankelly:
This hook will allow docker autobuild to build from the offical tag
and removes the need to update the dockerfile after the release has
been made.

It also adds a build arg, DISTRO_URL which can be used for testing
release candidates.

```
docker build --build-arg BK_VERSION=4.9.1
    --build-arg DISTRO_URL=<rc-tarball-url> .
```


eolivelli:
cc @andianmo

PR #2051, opened 2019-04-02
Cancel LAC watch when longpoll LAC times out

ivankelly:
Otherwise the watch objects will accumulate and eventually cause an
OOM on the bookie, if LAC doesn't progress.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/LongPollReadEntryProcessorV3.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/LongPollReadEntryProcessorV3.java
@@ -163,9 +163,10 @@ class LongPollReadEntryProcessorV3 extends ReadEntryProcessorV3 implements Watch
                 }
                 synchronized (this) {
                     expirationTimerTask = requestTimer.newTimeout(timeout -> {
-                        // When the timeout expires just get whatever is the current
-                        // readLastConfirmed
-                        LongPollReadEntryProcessorV3.this.scheduleDeferredRead(true);
+                            requestProcessor.bookie.cancelWaitForLastAddConfirmedUpdate(ledgerId, this);
+                            // When the timeout expires just get whatever is the current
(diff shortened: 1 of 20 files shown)


ivankelly:
It's hard to test an OOM type thing in a unit test, but if you want to see the problem in action, run a local bookie and run the following against it.

https://gist.github.com/ivankelly/65c999a1a3a280eea7e5b27e0a5a361b

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/LongPollReadEntryProcessorV3.java
```diff
@@ -163,9 +163,10 @@ private ReadResponse getLongPollReadResponse() {
                 }
                 synchronized (this) {
                     expirationTimerTask = requestTimer.newTimeout(timeout -> {
-                        // When the timeout expires just get whatever is the current
-                        // readLastConfirmed
-                        LongPollReadEntryProcessorV3.this.scheduleDeferredRead(true);
+                            requestProcessor.bookie.cancelWaitForLastAddConfirmedUpdate(ledgerId, this);
```
eolivelli: this is the core of the fix, isn't it ?

ivankelly: @eolivelli yes

ivankelly:
@eolivelli regarding mockito test, what would you assert on?

ivankelly:
Must have neglected to clean before doing full build. Tests compile now

eolivelli:
@ivankelly 
I was talking about ensuring that `LongPollReadEntryProcessorV3.java` actually calls `requestProcessor.bookie.cancelWaitForLastAddConfirmedUpdate`
which is the real fix.

This way we want drop that code in the future by mistake

ivankelly:
@eolivelli Added test,

PR #2052, opened 2019-04-02
Wait for LAC update even if ledger fenced

ivankelly:
Previous behaviour was to return straight away. However, LAC can
change when the ledger is fenced and there is no guarantee that a
fenced ledger will turn into a closed ledger (fencing client may
crash), which would cause the client tailing with longpoll LAC to go
into a tight loop.


eolivelli:
Any test?
At least about long poll + fence
I didn't check, so we already have such test?

ivankelly:

ivankelly:
@eolivelli this would be a test on timing, which I don't think is very useful.

Issue #2053, opened 2019-04-02
FastCodahale Timers miscalculate percentiles


nicmichael:
**BUG REPORT**

***Describe the bug***

FastCodahale Timer implementation may miscalculate percentiles if snapshots of values are slightly out of sync, and if only few events have been recorded.

FastCodahale Timers use fine-grained locking and are meant to tolerate that (some) values change while being recorded or while snapshots are created. Currently, the total count of requests is not synchronized with the number of requests recorded in percentile buckets. If a snapshot is created while the total count of the timer has been incremented beyond the sum of values in the percentile buckets, the percentile calculation may produce wrong values.

For example, if 3 percentile values have been recorded, but the overall count is 4, then the percentile calculation would be based on 4 values. This becomes most obvious if a percentile > .75 (e.g. p95) is being calculated. For this, the implementation will try to find 0.95 * 4 values, which is more than the 3 values recorded in the buckets. Since no bucket fulfills the criteria, the bound of the last (overflow) bucket will be returned, i.e. Long.MAX_VALUE.

***To Reproduce***

I added a test case in my pull request to fix this bug which demonstrates the behavior.

***Expected behavior***

Snapshots should base the percentile calculation on the sum of all values contained in the buckets, not on any value provided by a caller (which might be wrong or out of sync).

PR #2054, opened 2019-04-02
Issue 2053: Bugfix for Percentile Calculation in FastCodahale Timer Implementation

nicmichael:
This bugfix for the FastCodahale timer implementation ensures that percentiles provided by a FastSnapshot are calculated correctly even if the total count of events (provided by FastTimer) is out of sync with the recorded events in the percentile buckets.

### Motivation

FastCodahale Timer implementation may miscalculate percentiles if snapshots of values are slightly out of sync, and if only few events have been recorded.

FastCodahale Timers use fine-grained locking and are meant to tolerate that (some) values change while being recorded or while snapshots are created. Currently, the total count of requests is not synchronized with the number of requests recorded in percentile buckets. If a snapshot is created while the total count of the timer has been incremented beyond the sum of values in the percentile buckets, the percentile calculation may produce wrong values.

For example, if 3 percentile values have been recorded, but the overall count is 4, then the percentile calculation would be based on 4 values. This becomes most obvious if a percentile > .75 (e.g. p95) is being calculated. For this, the implementation will try to find 0.95 * 4 values, which is more than the 3 values recorded in the buckets. Since no bucket fulfills the criteria, the bound of the last (overflow) bucket will be returned, i.e. Long.MAX_VALUE.

### Changes

FastSnapshots now bases the percentile calculation on the sum of values in the percentile buckets rather than a count provided by the caller (i.e. FastTimer). This ensures that percentiles are calculated correctly without the need of having all counters fully synchronized.

Master Issue: #2053

### DIFF
--- a/bookkeeper-stats-providers/codahale-metrics-provider/src/main/java/org/apache/bookkeeper/stats/codahale/FastSnapshot.java
+++ b/bookkeeper-stats-providers/codahale-metrics-provider/src/main/java/org/apache/bookkeeper/stats/codahale/FastSnapshot.java
@@ -108,13 +108,15 @@ public class FastSnapshot extends Snapshot {
     }
 
     /**
-     * Calculates the sum of values of an array
+     * Calculates the sum of values of an array.
      * @param a an array of values
      * @return the sum of all array values
      */
     private long sumOf(long[] a) {
         long sum = 0;
-        for (long x : a) sum += x;
+        for (long x : a) {
+            sum += x;
+        }
         return sum;
     }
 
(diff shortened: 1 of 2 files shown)


sijie:
run pr validation
run integration tests

sijie:
there are two checkstyle violations. can you check and fix them?

```
[ERROR] /home/jenkins/jenkins-slave/workspace/bookkeeper_precommit_pullrequest_validation@2/bookkeeper-stats-providers/codahale-metrics-provider/src/main/java/org/apache/bookkeeper/stats/codahale/FastSnapshot.java:110: First sentence should end with a period. [JavadocStyle]
[ERROR] /home/jenkins/jenkins-slave/workspace/bookkeeper_precommit_pullrequest_validation@2/bookkeeper-stats-providers/codahale-metrics-provider/src/main/java/org/apache/bookkeeper/stats/codahale/FastSnapshot.java:117: 'for' construct must use '{}'s. [NeedBraces]
```

nicmichael:
I've fixed the checkstyle violations, but cluster test fails. I don't see why...?

sijie:
run integration tests

PR #2069, opened 2019-04-20
Use pure python implementation of MurmurHash

merlimat:
### Motivation

BK table python client is depending on `mmh3` library for MurmurHash. This library wraps a C based function but there are no binaries published on PyPI. Therefore users need to have a GCC installed in order to install the BK client lib, since it compiles it at install time. GCC is typically not available in Docker containers.

### Modifications

Include a pure python implementation of MurmurHash to use if the C based is not found. 

Notes:
 * I couldn't find any published pure-python MurmurHash implementations on PyPI
 * Importing public-domain code is permitted by ASF


vicaya:
Instead of creating a python impl that needs to be thoroughly tested and maintained (I don't see any unit tests in this patch). Why can't you guys fork mmh3 and create corresponding binary packages on pypi?

merlimat:
> Why can't you guys fork mmh3 and create corresponding binary packages on pypi?

I'd rather just not use MurmurHash than maintaining a fork

> There is no test case and this is dangerous.
 
We could import tests from that same repo

> If it is possible to use a third party maintaned lib it would be better

Of course but I couldn't find any library available on PyPI that was either: 
 * Pure python
 * Has the binaries available

ivankelly:
> Importing public-domain code is permitted by ASF

You should still add something in the NOTICE. https://apache.org/legal/resolved.html#handling-public-domain-licensed-works

grantwwu:
What about https://github.com/wc-duck/pymmh3 ?

merlimat:
@eolivelli @ivankelly The pymmh3 are now available on PyPI. Just referencing those instead of importing the source code.

grantwwu:
What's the performance impact of this?  Want to know if installing Cython mmh3 does anything significant.

vicaya:
@merlimat, where is your pymmh3 source repo? I just tried it. It works on python 2.7.x but fails on python 3.7.x:
```
Python 3.7.3 (default, Apr  8 2019, 12:02:14) 
[Clang 10.0.1 (clang-1001.0.46.3)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import pymmh3
>>> pymmh3.hash64("foo")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'pymmh3' has no attribute 'hash64'
>>> 
```
This is due to a bug in `pymmh3/__init__.py`: `from pymmh3 import *` missing a dot.

A quick microbenchmark with `perf` shows that the pure python impl ranges from being 16x slower for a short message ("foo"), 44x slower for a medium message (42 chars), to 710x slower for a long message (1512 chars (The Gettysburg Address)) than mmh3:
```
$ python3 simple.py 
.....................
short mmh3: Mean +- std dev: 429 ns +- 26 ns
.....................
short pymmh3: Mean +- std dev: 6.85 us +- 0.08 us
.....................
medium mmh3: Mean +- std dev: 426 ns +- 8 ns
.....................
medium pymmh3: Mean +- std dev: 18.6 us +- 0.4 us
.....................
long mmh3: Mean +- std dev: 705 ns +- 14 ns
.....................
long pymmh3: Mean +- std dev: 501 us +- 21 us
```
Since the bk client only impacts pulsar functions, which is deployed inside official broker containers, which would always have mmh3 installed, there should be no perf impact in usual deployment. The slow down only happens when people what to custom build smaller broker container images.

aahmed-se:
@vicaya here is the source
https://github.com/aahmed-se/pymmh3

aahmed-se:
I have pushed a new version to fix the issue.

PR #2072, opened 2019-04-26
DLog Namespace#openLog should declare LogNotFoundException

ivankelly:
Other methods such as delete do, so openLog should declare it to be
consistent and to give users a hint that it can be caught.


ivankelly:

Issue #1927, opened 2019-02-01
localbookie cannot be started with a ledger path which isn't on the root level


ivankelly:
74bd92605bc4986705031aea8943905dbc9f9f85 allowed the client to overwrite localbookie configurations from conf/bk_server.conf

However, if you set the zkLedgersRootPath to something like /messaging/bookkeeper/ledgers, it will fail because it doesn't create the parent znodes (and there's no other way to do it since this is localbookie, and it only lives for the lifetime of the process).

zhaohaidao:
Is there anyone working on it ? 
I want to give it a try.

sijie:
@zhaohaidao I don't think so. Please go ahead. If you need any help, please let me know.

PR #2082, opened 2019-05-04
Issue 1927: fix NoNodeException in LocalBookeeper

zhaohaidao:
### Motivation
It addresses #1927 where it provides a conf check at the beginning of startLocalBookiesInternal function.
### Changes

- Add a conf check at the beginning of startLocalBookiesInternal function where non-default zkLedgersRootPath is not allowed.

cc @sijie

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/util/LocalBookKeeper.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/util/LocalBookKeeper.java
@@ -134,17 +134,16 @@ public class LocalBookKeeper {
                     .connectString(zkHost + ":" + zkPort)
                     .sessionTimeoutMs(zkSessionTimeOut)
                     .build()) {
-            List<Op> multiOps = Lists.newArrayListWithExpectedSize(3);
             String zkLedgersRootPath = ZKMetadataDriverBase.resolveZkLedgersRootPath(baseConf);
-            multiOps.add(
-                Op.create(zkLedgersRootPath, new byte[0], Ids.OPEN_ACL_UNSAFE, CreateMode.PERSISTENT));
+            ZkUtils.createFullPathOptimistic(zkc, zkLedgersRootPath, new byte[0], Ids.OPEN_ACL_UNSAFE,
+                    CreateMode.PERSISTENT);
+            List<Op> multiOps = Lists.newArrayListWithExpectedSize(2);
             multiOps.add(
                 Op.create(zkLedgersRootPath + "/" + AVAILABLE_NODE,
                     new byte[0], Ids.OPEN_ACL_UNSAFE, CreateMode.PERSISTENT));
             multiOps.add(
                 Op.create(zkLedgersRootPath + "/" + AVAILABLE_NODE + "/" + READONLY,
                     new byte[0], Ids.OPEN_ACL_UNSAFE, CreateMode.PERSISTENT));
-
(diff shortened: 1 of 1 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/util/LocalBookKeeper.java
```diff
@@ -345,15 +345,18 @@ static void startLocalBookiesInternal(ServerConfiguration conf,
                                           String zkDataDir,
                                           String localBookiesConfigDirName)
             throws Exception {
-
+        if (conf.getZkLedgersRootPath() != BookKeeperConstants.DEFAULT_ZK_LEDGERS_ROOT_PATH) {
```
eolivelli: This is a String, so you have to use 'equals'

zhaohaidao:
Failed replication tests  finished with timeout in my laptop(macOS).
It seems that it blocked at test env setup step according the test log.
My test command is as followed.
`mvn clean install -DskipTests; mvn -pl bookkeeper-server clean test -Dtest=BookieAutoRecoveryTest#testEmptyLedgerLosesQuorumEventually`

Did I execute the right command ?  Or I missed something. The test log is attached as followed. Thanks

[org.apache.bookkeeper.replication.BookieAutoRecoveryTest-output.txt](https://github.com/apache/bookkeeper/files/3144538/org.apache.bookkeeper.replication.BookieAutoRecoveryTest-output.txt)

sijie:

zhaohaidao:
Looks like my commit affects case-BookieAutoRecoveryTest#testEmptyLedgerLosesQuorumEventually.
However, this case failed with timeout in my laptop(macOS). Should I do some preparation for test in macOS?

> Failed replication tests finished with timeout in my laptop(macOS).
> It seems that it blocked at test env setup step according the test log.
> My test command is as followed.
> `mvn clean install -DskipTests; mvn -pl bookkeeper-server clean test -Dtest=BookieAutoRecoveryTest#testEmptyLedgerLosesQuorumEventually`
> 
> Did I execute the right command ? Or I missed something. The test log is attached as followed. Thanks
> 
> [org.apache.bookkeeper.replication.BookieAutoRecoveryTest-output.txt](https://github.com/apache/bookkeeper/files/3144538/org.apache.bookkeeper.replication.BookieAutoRecoveryTest-output.txt)

ivankelly:
This doesn't address #1927. The problem in #1927 is that if you have something like /streaming/bookkeeper/ledgers as your root path, localbookie will fail to start. With this fix, it still fails to start. To address the issue, you need to create the path on boot.

zhaohaidao:
@ivankelly Thanks for your advice.
My fix logic is that it is simpler to constrain the input range than dealing with arbitrary input.
Although this fix will still fail with /streaming/bookkeeper/ledgers as the root path,  the error message means it is an illegal parameter error rather than a runtime error like NoNodeException.
Then if users see the error message, they will update their zkLedgerRootPath.

ivankelly:
@zhaohaidao the fix to permit the paths is one line, using ```ZkUtils#createFullPathOptimistic```

zhaohaidao:
@ivankelly Thanks for your reminder. I submitted a fix as per your suggestion. Please have a look at it.

sijie:

sijie:
@eolivelli can you please take a look at this PR again?

sijie:
ping @eolivelli

Issue #2071, opened 2019-04-24
A bookkeeper should fallback to use v3 protocol for some types of requests if they are not implemented in v2


sijie:
**FEATURE REQUEST**

1. Please describe the feature you are requesting.

Currently not all the features are available in v2 protocol. Such as readLac, writeLac, getBookieInfo. The gap is increasing as well.  Moving forward, it is really hard to maintain what features are not available when `useV2WireProtocol` is used.

When a bookkeeper client is configured with `useV2WireProtocol = true`, it should fallback to use v3 protocol for the types of request that v2 doesn't support. This would allow use have a fully functional client even when using `useV2WireProtocol = true`.

2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have).

This is a blocker for apache/pulsar#3828

3. Provide any additional detail on your proposed use case for this feature.

For addressing the problem in apache/pulsar#3828

sijie:
This has been implemented and fixed by #2085

PR #2085, opened 2019-05-06
Fallback to use v3 protocol for some types of requests if they are not implemented in v2

zymap:

### Motivation

#2071 

### Changes

- Add a client pool use v3 wire protocol
- Obtain client by version
- Currently only support `writeLac` and `readLac`

Master Issue: #2071

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieClientImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieClientImpl.java
@@ -21,6 +21,8 @@
 package org.apache.bookkeeper.proto;
 
 import static com.google.common.base.Charsets.UTF_8;
+import static org.apache.bookkeeper.proto.BookkeeperProtocol.ProtocolVersion;
+import static org.apache.bookkeeper.proto.BookkeeperProtocol.ProtocolVersion.VERSION_THREE;
 import static org.apache.bookkeeper.util.SafeRunnable.safeRun;
 
 import com.google.common.collect.Lists;
@@ -93,6 +95,7 @@ public class BookieClientImpl implements BookieClient, PerChannelBookieClientFac
     private final ExtensionRegistry registry;
 
     private final ClientConfiguration conf;
+    private final ClientConfiguration v3Conf;
     private volatile boolean closed;
     private final ReentrantReadWriteLock closeLock;
     private final StatsLogger statsLogger;
@@ -105,6 +108,8 @@ public class BookieClientImpl implements BookieClient, PerChannelBookieClientFac
                             OrderedExecutor executor, ScheduledExecutorService scheduler,
                             StatsLogger statsLogger) throws IOException {
         this.conf = conf;
+        this.v3Conf = new ClientConfiguration(conf);
+        this.v3Conf.setUseV2WireProtocol(false);
         this.eventLoopGroup = eventLoopGroup;
         this.allocator = allocator;
         this.executor = executor;
@@ -179,14 +184,14 @@ public class BookieClientImpl implements BookieClient, PerChannelBookieClientFac
 
     @Override
     public PerChannelBookieClient create(BookieSocketAddress address, PerChannelBookieClientPool pcbcPool,
-            SecurityHandlerFactory shFactory, int bookieProtocolVersion) throws SecurityException {
+            SecurityHandlerFactory shFactory, ProtocolVersion bookieProtocolVersion) throws SecurityException {
         StatsLogger statsLoggerForPCBC = statsLogger;
         if (conf.getLimitStatsLogging()) {
             statsLoggerForPCBC = NullStatsLogger.INSTANCE;
         }
         ClientConfiguration clientConfiguration = conf;
-        if (BookieProtocol.PROTOCOL_VERSION3 == bookieProtocolVersion) {
-            clientConfiguration.setUseV2WireProtocol(false);
+        if (bookieProtocolVersion == VERSION_THREE) {
+            clientConfiguration = v3Conf;
         }
         return new PerChannelBookieClient(clientConfiguration, executor, eventLoopGroup, allocator, address,
                                    statsLoggerForPCBC, authProviderFactory, registry, pcbcPool, shFactory);
@@ -271,7 +276,7 @@ public class BookieClientImpl implements BookieClient, PerChannelBookieClientFac
             }
 
             toSend.release();
-        }, ledgerId, BookieProtocol.PROTOCOL_VERSION3);
+        }, ledgerId, VERSION_THREE);
     }
 
     private void completeAdd(final int rc,
@@ -471,7 +476,7 @@ public class BookieClientImpl implements BookieClient, PerChannelBookieClientFac
             } else {
                 pcbc.readLac(ledgerId, cb, ctx);
             }
-        }, ledgerId, BookieProtocol.PROTOCOL_VERSION3);
+        }, ledgerId, VERSION_THREE);
     }
 
     public void readEntry(BookieSocketAddress addr, long ledgerId, long entryId,
@@ -550,7 +555,7 @@ public class BookieClientImpl implements BookieClient, PerChannelBookieClientFac
             } else {
                 pcbc.getBookieInfo(requested, cb, ctx);
             }
-        }, requested, BookieProtocol.PROTOCOL_VERSION3);
+        }, requested, VERSION_THREE);
     }
 
     private void monitorPendingOperations() {
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieClientImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieClientImpl.java
@@ -179,13 +179,17 @@ public class BookieClientImpl implements BookieClient, PerChannelBookieClientFac
 
     @Override
     public PerChannelBookieClient create(BookieSocketAddress address, PerChannelBookieClientPool pcbcPool,
-            SecurityHandlerFactory shFactory) throws SecurityException {
+            SecurityHandlerFactory shFactory, int bookieProtocolVersion) throws SecurityException {
         StatsLogger statsLoggerForPCBC = statsLogger;
         if (conf.getLimitStatsLogging()) {
             statsLoggerForPCBC = NullStatsLogger.INSTANCE;
         }
-        return new PerChannelBookieClient(conf, executor, eventLoopGroup, allocator, address, statsLoggerForPCBC,
-                authProviderFactory, registry, pcbcPool, shFactory);
+        ClientConfiguration clientConfiguration = conf;
+        if (BookieProtocol.PROTOCOL_VERSION3 == bookieProtocolVersion) {
+            clientConfiguration.setUseV2WireProtocol(false);
+        }
+        return new PerChannelBookieClient(clientConfiguration, executor, eventLoopGroup, allocator, address,
+                                   statsLoggerForPCBC, authProviderFactory, registry, pcbcPool, shFactory);
     }
 
     public PerChannelBookieClientPool lookupClient(BookieSocketAddress addr) {
@@ -467,7 +471,7 @@ public class BookieClientImpl implements BookieClient, PerChannelBookieClientFac
             } else {
                 pcbc.readLac(ledgerId, cb, ctx);
             }
-        }, ledgerId);
+        }, ledgerId, BookieProtocol.PROTOCOL_VERSION3);
     }
 
     public void readEntry(BookieSocketAddress addr, long ledgerId, long entryId,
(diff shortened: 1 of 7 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtocol.java
```diff
@@ -47,6 +47,11 @@
      */
     byte CURRENT_PROTOCOL_VERSION = 2;
 
+    /**
```
sijie: This file is only for v2 protocol. I would suggest not touching this file. You can use `ProtocolVersion` instead. `ProtocolVersion` already defines the version numbers from 1 to 3.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClientFactory.java
```diff
@@ -36,6 +36,6 @@
      * @return the client connected to address.
      * @throws SecurityException
      */
-    PerChannelBookieClient create(BookieSocketAddress address,
-            PerChannelBookieClientPool pcbcPool, SecurityHandlerFactory shFactory) throws SecurityException;
+    PerChannelBookieClient create(BookieSocketAddress address, PerChannelBookieClientPool pcbcPool,
```
sijie: I would suggest using `ProtocolVersion` in this method.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClientPool.java
```diff
@@ -40,6 +40,16 @@
...
 
+    /**
+     * Obtain a channel from channel pool by version to execute operations.
+     *
+     * @param callback
+     *          callback to return channel from channel pool
+     * @param version
+     *          get specify version channel from pool
+     */
+    void obtain(GenericCallback<PerChannelBookieClient> callback, long key, int version);
```
sijie:
```suggestion
    void obtain(GenericCallback<PerChannelBookieClient> callback, long key, ProtocolVersion version);
```

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieClientImpl.java
```diff
@@ -179,13 +179,17 @@ public long getNumPendingRequests(BookieSocketAddress address, long ledgerId) {
...
     public PerChannelBookieClient create(BookieSocketAddress address, PerChannelBookieClientPool pcbcPool,
-            SecurityHandlerFactory shFactory) throws SecurityException {
+            SecurityHandlerFactory shFactory, int bookieProtocolVersion) throws SecurityException {
         StatsLogger statsLoggerForPCBC = statsLogger;
         if (conf.getLimitStatsLogging()) {
             statsLoggerForPCBC = NullStatsLogger.INSTANCE;
         }
-        return new PerChannelBookieClient(conf, executor, eventLoopGroup, allocator, address, statsLoggerForPCBC,
-                authProviderFactory, registry, pcbcPool, shFactory);
+        ClientConfiguration clientConfiguration = conf;
```
sijie: Oh, you need to clone the clientConfiguration.

So I would suggest having a variable in BookieClientImpl `v3Conf` then use v3Conf if `ProtocolVersion` is V3.

```
if (v3Conf == null) {
    v3Conf = new ClientConfiguration(conf);
    v3Conf.setUseV2WriteProtocol(false);
}
```

zymap: Yeah, that my mistake. I shouldn't copy it directly. Thanks, fix it later

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtocol.java
zymap: Sure

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClientFactory.java
zymap: Sure

zymap:

zymap:
run integration tests

zymap:
@sijie All test is passed, PTAL

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/DefaultPerChannelBookieClientPool.java
```diff
@@ -63,12 +66,16 @@
         this.address = address;
         this.conf = conf;
 
-        this.shFactory = SecurityProviderFactoryFactory
-                .getSecurityProviderFactory(conf.getTLSProviderFactoryClass());
+        this.shFactory = SecurityProviderFactoryFactory.getSecurityProviderFactory(conf.getTLSProviderFactoryClass());
 
         this.clients = new PerChannelBookieClient[coreSize];
```
eolivelli: Do we have to always create v2 clients even if I am using only v3 ?

tests/integration-tests-topologies/src/main/java/org/apache/bookkeeper/tests/containers/BKStandaloneContainer.java
```diff
@@ -31,7 +31,7 @@
 @Slf4j
 public class BKStandaloneContainer<SelfT extends BKStandaloneContainer<SelfT>> extends ChaosContainer<SelfT> {
 
-    private static final int ZK_PORT = 2181;
+    public static final int ZK_PORT = 2181;
```
eolivelli: Why are you making this variable 'public'?

zymap: Thanks for your review. oh, That's my mistake. I will fix it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/DefaultPerChannelBookieClientPool.java
zymap: It doesn't. I should make it as optional. Fix it later

zymap:
@eolivelli  Fixed, PTAL

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieClientImpl.java
```diff
@@ -267,7 +274,7 @@ public void writeLac(final BookieSocketAddress addr, final long ledgerId, final
             }
 
             toSend.release();
-        }, ledgerId);
+        }, ledgerId, conf.getUseV2WireProtocol());
```
sijie: I don't think the value here should be `conf.getUseV2WireProtocol()`. It should be `true /* forceUseV3 *`, right?

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieClientImpl.java
```diff
@@ -467,7 +474,7 @@ public void readLac(final BookieSocketAddress addr, final long ledgerId, final R
             } else {
                 pcbc.readLac(ledgerId, cb, ctx);
             }
-        }, ledgerId);
+        }, ledgerId, conf.getUseV2WireProtocol());
```
sijie: I don't think the value here should be `conf.getUseV2WireProtocol()`. It should be `true /* forceUseV3 *`, right?

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieClientImpl.java
```diff
@@ -546,7 +553,7 @@ public void getBookieInfo(final BookieSocketAddress addr, final long requested,
             } else {
                 pcbc.getBookieInfo(requested, cb, ctx);
             }
-        }, requested);
+        }, requested, conf.getUseV2WireProtocol());
```
sijie: I don't think the value here should be `conf.getUseV2WireProtocol()`. It should be `true /* forceUseV3 *`, right?

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/DefaultPerChannelBookieClientPool.java
```diff
@@ -48,6 +48,7 @@
     final BookieSocketAddress address;
 
     final PerChannelBookieClient[] clients;
+    final PerChannelBookieClient[] clientsV3;
```
sijie: It is better to call it `clientsV3Enforced`. otherwise it becomes a bit confused, because `clients` can be v3 as well.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/DefaultPerChannelBookieClientPool.java
```diff
@@ -63,12 +64,20 @@
...
+            this.clients[i] = factory.create(address, this, shFactory, false);
+        }
+
+        if (conf.getUseV2WireProtocol()) {
+            this.clientsV3 = new PerChannelBookieClient[coreSize];
+            for (int i = 0; i < coreSize; i++) {
+                this.clientsV3[i] = factory.create(address, this, shFactory, true);
+            }
+        } else {
+            this.clientsV3 = null;
```
sijie: assign `clients` to `clientsV3` here. try to avoid having a field to be null.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/DefaultPerChannelBookieClientPool.java
```diff
@@ -106,6 +130,9 @@ public boolean isWritable(long key) {
     public void checkTimeoutOnPendingOperations() {
         for (int i = 0; i < clients.length; i++) {
             clients[i].checkTimeoutOnPendingOperations();
+            if (null != clientsV3) {
```
sijie:
```suggestion
            if (clients != clientsV3) {
```

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/DefaultPerChannelBookieClientPool.java
```diff
@@ -116,15 +143,21 @@ public void recordError() {
 
     @Override
     public void disconnect(boolean wait) {
-        for (PerChannelBookieClient pcbc : clients) {
-            pcbc.disconnect(wait);
+        for (int i = 0; i < clients.length; i++) {
+            clients[i].disconnect();
+            if (null != clientsV3) {
```
sijie:
```suggestion
            if (clients != clientsV3) {
```

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/DefaultPerChannelBookieClientPool.java
```diff
@@ -116,15 +143,21 @@ public void recordError() {
...
         }
     }
 
     @Override
     public void close(boolean wait) {
-        for (PerChannelBookieClient pcbc : clients) {
-            pcbc.close(wait);
+        for (int i = 0; i < clients.length; i++) {
+            clients[i].close(wait);
+            if (null != clientsV3) {
```
sijie:
```suggestion
            if (clients != clientsV3) {
```

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/DefaultPerChannelBookieClientPool.java
```diff
@@ -134,6 +167,11 @@ public long getNumPendingCompletionRequests() {
         for (PerChannelBookieClient pcbc : clients) {
             numPending += pcbc.getNumPendingCompletionRequests();
         }
+        if (null != clientsV3) {
```
sijie:
```suggestion
        if (clients != clientsV3) {
```

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/DefaultPerChannelBookieClientPool.java
zymap: good idea

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/DefaultPerChannelBookieClientPool.java
zymap: ok

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieClientImpl.java
zymap: Yeah. will fix

zymap:
@sijie  @eolivelli  I do some changes PTAL. :)

zymap:

Issue #2087, opened 2019-05-06
Update lastLogMark when replaying journal


karanmehta93:
**QUESTION**

Consider the following scenario.

Bookie 1 started cleanly.
Bookie replays journals (there aren't any) and flushed `lastLogMark` to disk `(logFileId, logFileOffset) -> (0, 0)`
5 entries are written to it (Journal fsync done but entry logs are not flushed, hence `lastLogMark` is not updated on disk) `(logFileId, logFileOffset) -> (15156416854, 4096)`
Bookie crashes.
Bookie restarts in read only mode.
Bookie reads the persisted `lastLogMark` as `(logFileId, logFileOffset) -> (0, 0)`
Bookie replays journals but doesn't update `lastLogMark` while doing it (since in current behavior it gets updated only on journal writes)
Bookie flushes the ledger and flushes the `lastLogMark`
However the value written again is `(logFileId, logFileOffset) -> (0, 0)` and not the updated one.

Is this behavior expected? 
In my understanding, the `lastLogMark` field should get updated even when replaying the journals so that the new value can be persisted to disk.

@jvrao @reddycharan @sijie Thoughts?

This is not a huge issue if the bookie restarts in RW mode, since the `lastLogMark` will get updated as soon as bookie receives its first write and `syncThread` work will go on.

karanmehta93:
I wrote up this test for `BookieInitializationTest` class to corroborate my understanding of the code. There are 3 different lines that print the checkpoint to console.

```
    @Test
    public void someTest() throws Exception {
        File tmpDir = createTempDir("DiskCheck", "test");

        final ServerConfiguration conf = newServerConfiguration()
                .setJournalDirName(tmpDir.getPath())
                .setLedgerDirNames(new String[] { tmpDir.getPath() })
                .setDiskCheckInterval(1000)
                .setLedgerStorageClass(SortedLedgerStorage.class.getName())
                .setAutoRecoveryDaemonEnabled(false)
                .setZkTimeout(5000);

        BookieServer server = new MockBookieServer(conf);
        server.start();
        Checkpoint checkpoint = server.getBookie().journals.get(0).newCheckpoint();
        System.out.println("Checkpoint: " + checkpoint);
        ClientConfiguration clientConf = new ClientConfiguration();
        clientConf.setMetadataServiceUri(metadataServiceUri);
        BookKeeper bkClient = new BookKeeper(clientConf);
[...]

PR #2090, opened 2019-05-09
Update and flush lastLogMark when replaying journal

karanmehta93:

Update `lastLogMark` in memory after replaying each journal
Check for writable ledger dirs with `minUsableSizeForEntryLogCreation` to flush the `lastMark` file for bookies in ReadOnlyMode
Log line changes

### Motivation

Master Issue: #2087

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
@@ -924,7 +924,7 @@ public class Bookie extends BookieCriticalThread {
             journal.scanJournal(id, logPosition, scanner);
             // Update LastLogMark to Long.MAX_VALUE position after replaying journal
             // After LedgerStorage flush, SyncThread should persist this to disk
-            journal.getLastLogMark().setCurLogMark(id, Long.MAX_VALUE);
+            journal.setLastLogMarkToEof(id);
         }
     }
 
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
@@ -872,6 +872,7 @@ public class Bookie extends BookieCriticalThread {
                         recBuff.rewind();
                         handle.addEntry(Unpooled.wrappedBuffer(recBuff));
                     }
+
                 } catch (NoLedgerException nsle) {
                     if (LOG.isDebugEnabled()) {
                         LOG.debug("Skip replaying entries of ledger {} since it was deleted.", ledgerId);
@@ -883,10 +884,49 @@ public class Bookie extends BookieCriticalThread {
         };
 
         for (Journal journal : journals) {
-            journal.replay(scanner);
+            replay(journal, scanner);
         }
         long elapsedTs = System.currentTimeMillis() - startTs;
         LOG.info("Finished replaying journal in {} ms.", elapsedTs);
+    }
+
+    /**
+     * Replay journal files and updates journal's in-memory lastLogMark object.
+     *
+     * @param journal Journal object corresponding to a journalDir
+     * @param scanner Scanner to process replayed entries.
+     * @throws IOException
+     */
+    private void replay(Journal journal, JournalScanner scanner) throws IOException {
+        final LogMark markedLog = journal.getLastLogMark().getCurMark();
+        List<Long> logs = Journal.listJournalIds(journal.getJournalDirectory(), journalId -> {
+            if (journalId < markedLog.getLogFileId()) {
+                return false;
+            }
+            return true;
+        });
+        // last log mark may be missed due to no sync up before
+        // validate filtered log ids only when we have markedLogId
+        if (markedLog.getLogFileId() > 0) {
+            if (logs.size() == 0 || logs.get(0) != markedLog.getLogFileId()) {
+                throw new IOException("Recovery log " + markedLog.getLogFileId() + " is missing");
+            }
+        }
+
+        // TODO: When reading in the journal logs that need to be synced, we
+        // should use BufferedChannels instead to minimize the amount of
+        // system calls done.
+        for (Long id : logs) {
+            long logPosition = 0L;
+            if (id == markedLog.getLogFileId()) {
+                logPosition = markedLog.getLogFileOffset();
+            }
+            LOG.info("Replaying journal {} from position {}", id, logPosition);
+            journal.scanJournal(id, logPosition, scanner);
+            // Update LastLogMark to Long.MAX_VALUE position after replaying journal
+            // After LedgerStorage flush, SyncThread should persist this to disk
+            journal.getLastLogMark().setCurLogMark(id, Long.MAX_VALUE);
+        }
     }
 
     @Override
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
@@ -872,7 +872,6 @@ public class Bookie extends BookieCriticalThread {
                         recBuff.rewind();
                         handle.addEntry(Unpooled.wrappedBuffer(recBuff));
                     }
-
                 } catch (NoLedgerException nsle) {
                     if (LOG.isDebugEnabled()) {
                         LOG.debug("Skip replaying entries of ledger {} since it was deleted.", ledgerId);
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
@@ -710,6 +710,15 @@ public class Journal extends BookieCriticalThread implements CheckpointSource {
     }
 
     /**
+     * Update lastLogMark of the journal,
+     * Indicates that the file has been processed.
+     * @param id
+     */
+    void setLastLogMarkToEof(Long id) {
+        lastLogMark.getCurMark().setLogMark(id, Long.MAX_VALUE);
+    }
+
+    /**
      * Application tried to schedule a checkpoint. After all the txns added
      * before checkpoint are persisted, a <i>checkpoint</i> will be returned
      * to application. Application could use <i>checkpoint</i> to do its logic.
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
@@ -78,7 +78,7 @@ public class Journal extends BookieCriticalThread implements CheckpointSource {
     /**
      * Filter to pickup journals.
      */
-    private interface JournalIdFilter {
+    public interface JournalIdFilter {
         boolean accept(long journalId);
     }
 
@@ -197,7 +197,9 @@ public class Journal extends BookieCriticalThread implements CheckpointSource {
             // persisted to disk (both index & entry logger)
             lastMark.getCurMark().writeLogMark(bb);
 
-            LOG.info("RollLog to persist last marked log : {}", lastMark.getCurMark());
+            if (LOG.isDebugEnabled()) {
+                LOG.debug("RollLog to persist last marked log : {}", lastMark.getCurMark());
+            }
 
             List<File> writableLedgerDirs = ledgerDirsManager
                     .getWritableLedgerDirsForNewLog();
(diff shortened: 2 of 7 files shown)


karanmehta93:
@jvrao @reddycharan Please review.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -196,11 +196,11 @@ void rollLog(LastLogMark lastMark) throws NoWritableLedgerDirException {
...
             lastMark.getCurMark().writeLogMark(bb);
-            if (LOG.isDebugEnabled()) {
-                LOG.debug("RollLog to persist last marked log : {}", lastMark.getCurMark());
-            }
+
+            LOG.info("RollLog to persist last marked log : {}", lastMark.getCurMark());
+
             List<File> writableLedgerDirs = ledgerDirsManager
-                    .getWritableLedgerDirs();
+                    .getWritableLedgerDirsForNewLog();
```
jvrao: I think this is crucial fix. May be reflect in the commit summary?

karanmehta93: Done.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -851,13 +854,10 @@ public boolean accept(long journalId) {
             }
         }
 
-        if (LOG.isDebugEnabled()) {
```
karanmehta93: Removing this since we have better log line at 865

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -196,11 +196,11 @@ void rollLog(LastLogMark lastMark) throws NoWritableLedgerDirException {
             // which is safe since records before lastMark have been
             // persisted to disk (both index & entry logger)
             lastMark.getCurMark().writeLogMark(bb);
-            if (LOG.isDebugEnabled()) {
-                LOG.debug("RollLog to persist last marked log : {}", lastMark.getCurMark());
-            }
+
+            LOG.info("RollLog to persist last marked log : {}", lastMark.getCurMark());
```
karanmehta93: @reddycharan 
I switched this log line from `DEBUG` to `INFO` since I believe that this is an important event to be logged. However given that we use ELPL, which flushes every 10 seconds, is this a problem?

reddycharan: yes, having log line for every 10 secs is not ideal. Keep it at debug level.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -821,6 +821,9 @@ public void scanJournal(long journalId, long journalPos, JournalScanner scanner)
                     scanner.process(journalVersion, offset, recBuff);
                 }
             }
+            // Update LastLogMark to EOF position after replaying journal
+            // After force flush of LedgerStorage, SyncThread will persist this to disk
+            lastLogMark.setCurLogMark(journalId, recLog.fc.position());
```
reddycharan: @karanmehta93 in retrospect, i think this may not be right place to update the lastLogMark. It should be part of the scanner (JournalScanner) logic. Not all scanners add entries to LedgerStorage, so this update logic should be part of JournalScanner, which is created in Bookie.readJournal method.

bookkeeper-server/src/test/java/org/apache/bookkeeper/test/BookKeeperClusterTestCase.java
```diff
@@ -833,4 +834,13 @@ public TestStatsProvider getStatsProvider(int index) throws Exception {
         return getStatsProvider(bs.get(index).getLocalAddress());
     }
 
+    public static int countNumOfFiles(File[] folderNames, String... extensions) {
```
reddycharan: move this utility method to bookkeeper/bookkeeper-server/src/test/java/org/apache/bookkeeper/util/TestUtils.java

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieJournalTest.java
```diff
@@ -640,6 +637,15 @@ public void testTruncatedInEntryJournal() throws Exception {
         }
     }
 
+    private Bookie startBookieReadJournal(ServerConfiguration conf) throws IOException, InterruptedException, BookieException {
```
reddycharan: nit: technically you are not starting Bookie (not calling start method of Bookie), so better rename it to createBookieAndReadJournal.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieJournalTest.java
```diff
@@ -640,6 +637,15 @@ public void testTruncatedInEntryJournal() throws Exception {
...
+        Bookie b = new Bookie(conf);
+        Journal journal = b.journals.get(0);
```
reddycharan: wondering if this testsuite is configured to use multiple journal directories, will it work correctly?

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieInitializationTest.java
```diff
@@ -136,6 +137,126 @@ public void tearDown() throws Exception {
         super.tearDown();
     }
 
+    @Test
+    public void testOneJournalReplayForBookieRestartInReadOnlyMode() throws Exception {
+        testJournalReplayForBookieRestartInReadOnlyMode(1);
+    }
+
+    @Test
+    public void testFourJournalReplayForBookieRestartInReadOnlyMode() throws Exception {
```
reddycharan: nit: testMultipleJournalReplayForBookieRestartInReadOnlyMode

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieInitializationTest.java
```diff
@@ -136,6 +137,126 @@ public void tearDown() throws Exception {
...
+
+            int txnBefore = countNumOfFiles(conf.getJournalDirs(), "txn");
+            int logBefore = countNumOfFiles(conf.getLedgerDirs(), "log");
+            int idxBefore = countNumOfFiles(conf.getLedgerDirs(), "idx");
+
+            server.start();
+
+            for (int j = 0; j < journalDirs.length; j++) {
+                Journal journal = server.getBookie().journals.get(j);
+                assertTrue(journal.getLastLogMark().getCurMark().compare(lastLogMarkList.get(j).getCurMark()) >= 0);
```
reddycharan: in which case it will be > 0 and == 0? can you separate these cases.

reddycharan:
also there are checkstyle violations, check them 

[ERROR] /home/jenkins/jenkins-slave/workspace/bookkeeper_precommit_pullrequest_validation/bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieInitializationTest.java:150: First sentence should end with a period. [JavadocStyle]
[ERROR] /home/jenkins/jenkins-slave/workspace/bookkeeper_precommit_pullrequest_validation/bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieJournalTest.java:640: Line is longer than 120 characters (found 127). [LineLength]
Audit done.

you can try running following commands locally

mvn clean install -DskipTests apache-rat:check spotbugs:check
mvn checkstyle:check

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieInitializationTest.java
karanmehta93: It should be always `> 0`, updated the test.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieJournalTest.java
karanmehta93: Updated the method to make it generic.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
karanmehta93: Yes @reddycharan That's a good observation. As discussed offline, I am refactoring the code to move the `replay()` method from Journal to Bookie, since it's bookie's task to replay the journal. In that manner we can simplify the logic of updating `LastLogMark`. 
By doing this, we lose the information of the file pointer end, hence I am arbitrarily updating it to `LONG.MAX_VALUE` since the FileChannel treats both those cases similarly.

karanmehta93:
Thanks @jvrao  and @reddycharan for the review. I have addressed the comments.

karanmehta93:
@sijie Can you please review?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -872,6 +872,7 @@ public void process(int journalVersion, long offset, ByteBuffer recBuff) throws
                         recBuff.rewind();
                         handle.addEntry(Unpooled.wrappedBuffer(recBuff));
                     }
+
```
reddycharan: nit: remove this new line

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -883,12 +884,51 @@ public void process(int journalVersion, long offset, ByteBuffer recBuff) throws
...
+        for (Long id : logs) {
+            long logPosition = 0L;
+            if (id == markedLog.getLogFileId()) {
+                logPosition = markedLog.getLogFileOffset();
+            }
+            LOG.info("Replaying journal {} from position {}", id, logPosition);
+            journal.scanJournal(id, logPosition, scanner);
+            // Update LastLogMark to Long.MAX_VALUE position after replaying journal
+            // After LedgerStorage flush, SyncThread should persist this to disk
+            journal.getLastLogMark().setCurLogMark(id, Long.MAX_VALUE);
```
reddycharan: I missed noticing this, I'm not sure if it is completely ok to store Long.MAX_VALUE as its position.

karanmehta93: I referred this: https://docs.oracle.com/javase/7/docs/api/java/nio/channels/FileChannel.html#position(long)

Since we don't have a clear way to know the exact EOF, this is a work around for that.

reddycharan: there are multiple things which we have to make sure about this
1) under no circumstance, we would try to write to this journal file beyond this curLogMark (Long.MAX_VALUE position).
2) What if a reader tries to read this journal file upto this Long.MAX_VALUE position? say for eg - readJournal BookieShell command.

In summary we have to be sure that all the components in our code deals correctly with this special position.

karanmehta93: > under no circumstance, we would try to write to this journal file beyond this curLogMark (Long.MAX_VALUE position).

Bookie restarts always create a new journal file with default size of 16 MB. This ensures that we never write to old journal file. This scenario is covered by the test case.

> What if a reader tries to read this journal file upto this Long.MAX_VALUE position? say for eg - readJournal BookieShell command

Some text from the link in my previous comment, `Setting the position to a value that is greater than the file's current size is legal but does not change the size of the file. A later attempt to read bytes at such a position will immediately return an end-of-file indication`, should help resolve it.

I understand that this is a critical fix and we have to think through all possibilities. IMHO, the test case (although a bit long), tries to cover all such potential scenarios.

reddycharan: @sijie @jvrao I don't have strong case to say no to this change, but I would like to get second opinion on this. Since it falls under critical section, I just want to double confirm.

karanmehta93: I have fixed the required tests and checkstyle violations. Other tests are failing b/c they are flaky. I have verified `testEmptyLedgerLosesQuorumEventually` by running locally.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -883,12 +883,51 @@ public void process(int journalVersion, long offset, ByteBuffer recBuff) throws
...
+        // system calls done.
+        for (Long id : logs) {
+            long logPosition = 0L;
+            if (id == markedLog.getLogFileId()) {
+                logPosition = markedLog.getLogFileOffset();
+            }
+            LOG.info("Replaying journal {} from position {}", id, logPosition);
+            journal.scanJournal(id, logPosition, scanner);
+            // Update LastLogMark to Long.MAX_VALUE position after replaying journal
+            // After LedgerStorage flush, SyncThread should persist this to disk
```
eolivelli: What about creating a method *moveLastMaskToEof()* in Journal?

This we will notice better this new case in the future and we will have a better encapsulation of the behaviour of the Journal (I don't like 'x.get.modifyYourInternals)

karanmehta93: I like the idea, have updated it.

reddycharan:
run integration tests

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
reddycharan: @karanmehta93 i'm approving this PR. @sijie please let us know if we missed anything with respect to setting CurLogMark to Long.MAX_VALUE.

reddycharan:
merging this commit.

reddycharan:
IGNORE IT CI

reddycharan:
IGNORE CI

karanmehta93:
The merge script has accidentally added an additional commit id as a part of commit message https://github.com/apache/bookkeeper/commit/36be8362399341022c8de64f9319270726df2cb3

`d35aa22 [Charan Reddy Guttapalem] Move common placementpolicy components to TopologyAwareEnsemblePlacementPolicy.`

Any ideas @sijie @eolivelli ?

sijie:
@karanmehta93 I think the merge script does a diff to list all the commits involved in this PR. We can probably remove "including the list commits" from the merge script since we are using github merge api to merge the PR.

ivankelly:
> IGNORE CI

@reddycharan You broke the build merging this. This change is causing integration tests to fail consistently.

cc: @jvrao @sijie @eolivelli @karanmehta93

karanmehta93:
@ivankelly Thanks for bringing this up. I believed that those failures were unrelated to this PR. AFAIK those were docker related tests and I was unable to reproduce them locally. If you could point me to any doc on running type of tests, that would be helpful. Thanks!

ivankelly:
@karanmehta93 The integration test jenkins config should show you how it's run:
https://github.com/apache/bookkeeper/blob/master/.test-infra/jenkins/job_bookkeeper_precommit_integrationtests.groovy

If you believe a failure is unrelated to the PR, you can rerun the test by commenting "rerun integration tests" on the PR.

karanmehta93:
rerun integration tests

ivankelly:
@karanmehta93 it only works for PRs that are open. and in any case, this change broke those tests. It's in the same area of code.

PR #2094, opened 2019-05-14
(WIP) Have min time interval for force compaction

karanmehta93:

Work continued from https://github.com/apache/bookkeeper/pull/1946

Descriptions of the changes in this PR (WIP):

- Added support for scheduling tasks with fixed delay in `MockExecutorController`
  - Can be used for adding unit tests for GC threads, LedgerDirsMonitor threads etc
- Fixed some broken tests
- Re-based the PR

Descriptions of the changes in PR 1946:

- minimum interval needed to trigger force compaction by LedgerDirsListener, 
since the last main compaction. Otherwise if isForceGCAllowWhenNoSpace is enabled
and when disk reaches storage threshold, it will do busy force gc for every
'diskCheckInterval' (10 secs) untill the disk/bookie becomes writable again.
gc operation is an expensive operation and it puts load on metadataserver for
iterating over all ledgers metadata. So we should call forcegc judiciously when
disk/bookie reaches storage threshold.

- here main compaction will be Major compaction, if majorcompaction is 
disabled/suspended then it will be minorcompaction. If both major and minor are
disabled/suspended then it will be just gc run.

- in LedgerDirsMonitor and in GarbageCollectorThread we should schedule tasks 
with fixed delay instead of at fixed rate. This would ensure that there would be
uniform delay between executions.

### DIFF
--- a/bookkeeper-common/src/test/java/org/apache/bookkeeper/common/testing/executors/MockExecutorController.java
+++ b/bookkeeper-common/src/test/java/org/apache/bookkeeper/common/testing/executors/MockExecutorController.java
@@ -28,6 +28,7 @@ import com.google.common.util.concurrent.SettableFuture;
 import java.time.Duration;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Random;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Delayed;
 import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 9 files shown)


karanmehta93:
As discussed with @reddycharan offline, due to certain race conditions, it might be simply preferable to update `lastCompactionTime` for every GC run, irrespective of who the caller is (forceGC run by `LedgerDirsMonitor` or scheduled GC Thread run). 

Without this PR, when disks get full, it will try running force GC every 10 seconds (since the monitor thread runs every 10 seconds), causing lot of burden on system. With this pause introduced by the PR, we would ensure some wait between consecutive GC runs. I also propose to remove the config parameter `FORCE_COMPACTION_INTERVAL` and fix it to `getGcWaitTime() / 2`, which is the current default.

Thoughts? @sijie @jvrao

sijie:
@karanmehta93 looks good to me.

shoothzj:
@karanmehta93 closed by stale and conflicts. If you still want to work on `min time interval for force compaction triggered by disk usage`, feel free to reopen and rebase it.

PR #2097, opened 2019-05-15
[Bk-Client] Check empty ledger-parent node while deleting ledger

rdhabalia:
### Motivation

As discussed at [#4276](https://github.com/apache/pulsar/issues/4276), while deleting ledger, bk-client should check parent node is empty before issuing delete request for parent znode.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/util/ZkUtils.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/util/ZkUtils.java
@@ -134,8 +134,21 @@ public class ZkUtils {
             public void processResult(int rc, String path, Object ctx) {
                 if (rc == Code.OK.intValue()) {
                     String parent = new File(originalPath).getParent().replace("\\", "/");
-                    asyncDeleteFullPathOptimistic(zk, parent, -1, callback, leafNodePath);
+                    zk.getData(parent, false, (dRc, dPath, dCtx, data, stat) -> {
+                        if (Code.OK.intValue() == dRc && (stat != null && stat.getNumChildren() == 0)) {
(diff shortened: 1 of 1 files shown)


rdhabalia:
@eolivelli @merlimat  can you please help to merge this one as well.

rdhabalia:
I verified that are couple of testcases that validate this change: eg: `TestZkUtils`, `AbstractZkLedgerManagerTest`.. so, i think we don't need same type of test again.

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/ZkUtils.java
```diff
@@ -134,8 +134,21 @@ public static void asyncDeleteFullPathOptimistic(final ZooKeeper zk, final Strin
...
                 if (rc == Code.OK.intValue()) {
                     String parent = new File(originalPath).getParent().replace("\\", "/");
-                    asyncDeleteFullPathOptimistic(zk, parent, -1, callback, leafNodePath);
+                    zk.getData(parent, false, (dRc, dPath, dCtx, data, stat) -> {
+                        if (Code.OK.intValue() == dRc && (stat != null && stat.getNumChildren() == 0)) {
+                            asyncDeleteFullPathOptimistic(zk, parent, -1, callback, leafNodePath);
+                        } else {
+                            // parent node is not empty so, complete the
+                            // callback
+                            if (path.equals(leafNodePath)) {
```
reddycharan: wondering what is the difference between the if/else cases here? since rc is OK here, will it not be the same?

rdhabalia: @reddycharan  that's true.. rc will be always OK here. fixed it.

rdhabalia:
@reddycharan addressed it.

PR #2066, opened 2019-04-18
Explicit error message if extent is not present on ZK

jvrao:
Client can get no extent error in the following two scenarios.
1. When client attempted to read/open an extent that
   is not on ZK (metadata server)
2. When client attempted to read a Ledger which is on metadata
   server, but somehow missing on bookies (Data server)

It is quite confusing to get the same error, NoSuchLedgerExists
for both these cases. This checkin introduced a new error at the
BookKeeper client level, NoSuchLedgerExistsOnMetadataServer if
it is not available on ZK.

For errors related to Ledger is not available on bookie,
I left NoSuchLedgerExists as-is(instead of changing it to
NoSuchLedgerExistsOnDataServer) to minimize code changes.

This change alone provides enough distinction between these
two types of errors.

Signed-off-by: Venkateswararao Jujjuri (JV) <vjujjuri@salesforce.com>

Descriptions of the changes in this PR:

### Motivation

(Explain: why you're making that change, what is the problem you're trying to solve)

### Changes

(Describe: what changes you have made)

Master Issue: #<master-issue-number>

> ---
> In order to uphold a high standard for quality for code contributions, Apache BookKeeper runs various precommit
> checks for pull requests. A pull request can only be merged when it passes precommit checks. However running all
> the precommit checks can take a long time, some trivial changes don't need to run all the precommit checks. You
> can check following list to skip the tests that don't need to run for your pull request. Leave them unchecked if
> you are not sure, committers will help you:
>
> - [ ] [skip bookkeeper-server bookie tests]: skip testing `org.apache.bookkeeper.bookie` in bookkeeper-server module.
> - [ ] [skip bookkeeper-server client tests]: skip testing `org.apache.bookkeeper.client` in bookkeeper-server module.
> - [ ] [skip bookkeeper-server replication tests]: skip testing `org.apache.bookkeeper.replication` in bookkeeper-server module.
> - [ ] [skip bookkeeper-server tls tests]: skip testing `org.apache.bookkeeper.tls` in bookkeeper-server module.
> - [ ] [skip bookkeeper-server remaining tests]: skip testing all other tests in bookkeeper-server module.
> - [ ] [skip integration tests]: skip docker based integration tests. if you make java code changes, you shouldn't skip integration tests.
> - [ ] [skip build java8]: skip build on java8. *ONLY* skip this when *ONLY* changing files under documentation under `site`.
> - [ ] [skip build java11]: skip build on java11. *ONLY* skip this when *ONLY* changing files under documentation under `site`.
[...]

PR #2099, opened 2019-05-20
[DLOG] Delete empty inprogress segment on recovery

ivankelly:
This addresses an issue, whereby if a user is using getLastTxId() to
find the first txnid when starting to write, they will continue to be
able to write if the last segment written by the previous writer was
empty. If the last segment was empty, then the maxTxId would be higher
than the result of getLastTxId(). maxTxId is read from a znode, while
getLastTxId() reads the txid of the last persisted record. In the case
of an empty inprogress segment, the maxTxId znode was being updated
with the expected first transaction id of the segment.

This patch addresses the issue with the following changes:
1. The maxTxId znode is only updated when inprogress segment is
   completed, so it's value always refers to a transaction that
   exists.
2. On recovery, if the inprogress segment is empty, delete it. There
   was a TODO comment to do this already there.
3. When generating the sequence number, allow a potential sequence
   number which is equal to the current max sequence number, as this
   can be the case where recovery deleted an inprogress empty segment.

### DIFF
--- a/stream/distributedlog/core/src/test/java/org/apache/distributedlog/TestDistributedLogBase.java
+++ b/stream/distributedlog/core/src/test/java/org/apache/distributedlog/TestDistributedLogBase.java
@@ -104,7 +104,6 @@ public class TestDistributedLogBase {
         Pair<ZooKeeperServerShim, Integer> serverAndPort = LocalDLMEmulator.runZookeeperOnAnyPort(zkTmpDir);
         zks = serverAndPort.getLeft();
         zkPort = serverAndPort.getRight();
-        LOG.info("IKDEBUG num bookies is {}", numBookies);
         bkutil = LocalDLMEmulator.newBuilder()
                 .numBookies(numBookies)
                 .zkHost("127.0.0.1")
--- a/stream/distributedlog/core/src/test/java/org/apache/distributedlog/TestDistributedLogBase.java
+++ b/stream/distributedlog/core/src/test/java/org/apache/distributedlog/TestDistributedLogBase.java
@@ -95,15 +95,16 @@ public class TestDistributedLogBase {
     protected static String zkServers;
     protected static int zkPort;
     protected static int numBookies = 3;
-    private static final List<File> tmpDirs = new ArrayList<File>();
+    protected static final List<File> TMP_DIRS = new ArrayList<File>();
 
     @BeforeClass
     public static void setupCluster() throws Exception {
         File zkTmpDir = IOUtils.createTempDir("zookeeper", "distrlog");
-        tmpDirs.add(zkTmpDir);
+        TMP_DIRS.add(zkTmpDir);
         Pair<ZooKeeperServerShim, Integer> serverAndPort = LocalDLMEmulator.runZookeeperOnAnyPort(zkTmpDir);
         zks = serverAndPort.getLeft();
         zkPort = serverAndPort.getRight();
+        LOG.info("IKDEBUG num bookies is {}", numBookies);
         bkutil = LocalDLMEmulator.newBuilder()
                 .numBookies(numBookies)
                 .zkHost("127.0.0.1")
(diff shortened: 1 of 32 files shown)


stream/distributedlog/core/src/test/java/org/apache/distributedlog/TestDistributedLogBase.java
```diff
@@ -95,15 +95,16 @@
...
 
     @BeforeClass
     public static void setupCluster() throws Exception {
         File zkTmpDir = IOUtils.createTempDir("zookeeper", "distrlog");
-        tmpDirs.add(zkTmpDir);
+        TMP_DIRS.add(zkTmpDir);
         Pair<ZooKeeperServerShim, Integer> serverAndPort = LocalDLMEmulator.runZookeeperOnAnyPort(zkTmpDir);
         zks = serverAndPort.getLeft();
         zkPort = serverAndPort.getRight();
+        LOG.info("IKDEBUG num bookies is {}", numBookies);
```
eolivelli: Nit: remove Ivan Kelly's debug?

stream/distributedlog/core/src/test/java/org/apache/distributedlog/TestTxnId.java
```diff
@@ -0,0 +1,107 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Test Cases for RollLogSegments.
+ */
+public class TestTxnId extends TestDistributedLogBase {
+    private static final Logger logger = LoggerFactory.getLogger(TestRollLogSegments.class);
+
+    @Test(timeout = 600000)
```
eolivelli: Nit: we banned test timeouts IIRC

stream/distributedlog/core/src/test/java/org/apache/distributedlog/TestTxnId.java
```diff
@@ -0,0 +1,107 @@
...
+
+            try (BKDistributedLogManager dlm = (BKDistributedLogManager) createNewDLM(conf, name);
+                 BKAsyncLogWriter writer = dlm.startAsyncLogSegmentNonPartitioned()) {
+                writer.write(DLMTestUtil.getLogRecordInstance(1, 100000)).join();
+                writer.write(DLMTestUtil.getLogRecordInstance(2, 100000)).join();
+
+                extraBookies.forEach(b -> b.shutdown());
+
+                writer.write(DLMTestUtil.getLogRecordInstance(3, 100000));
+                Thread.sleep(1000);
```
eolivelli: Is there any safer (not flaky) way of waiting?

stream/distributedlog/core/src/test/java/org/apache/distributedlog/TestDistributedLogBase.java
ivankelly: oops, will remove

stream/distributedlog/core/src/test/java/org/apache/distributedlog/TestTxnId.java
ivankelly: they're all over the place in dlog. Will remove though.

stream/distributedlog/core/src/test/java/org/apache/distributedlog/TestTxnId.java
ivankelly: I can try. This sleep exists, to give time for for the write to error and the error handling to try to create a new ledger and then fail and mark the current log segment as null. Looks like it's exposed.

ivankelly:
@eolivelli addressed comments. ptal

ivankelly:
run integration tests

ivankelly:
rerun integration tests

ivankelly:

sijie:
@ivankelly I have a separate pull request (#2102) for fixing the issue introduced by #2066

ivankelly:

ivankelly:

ivankelly:
rerun integration tests

sijie:
run integration tests

ivankelly:
run integration tests

PR #2105, opened 2019-05-30
Update lastLogMark to EOF when replaying journal

karanmehta93:

### Motivation
 
The [commit](https://github.com/apache/bookkeeper/commit/36be8362399341022c8de64f9319270726df2cb3) caused integration test failure `test101_RegenerateIndex`, with the exception
```
```java.io.IOException: Invalid argument
    at sun.nio.ch.FileDispatcherImpl.read0(Native Method)
    at sun.nio.ch.FileDispatcherImpl.read(FileDispatcherImpl.java:46)
    at sun.nio.ch.IOUtil.readIntoNativeBuffer(IOUtil.java:223)
    at sun.nio.ch.IOUtil.read(IOUtil.java:197)
    at sun.nio.ch.FileChannelImpl.read(FileChannelImpl.java:159)
    at org.apache.bookkeeper.bookie.JournalChannel.read(JournalChannel.java:257)
    at org.apache.bookkeeper.bookie.Journal.fullRead(Journal.java:1171)
    at org.apache.bookkeeper.bookie.Journal.scanJournal(Journal.java:792)
    at org.apache.bookkeeper.bookie.Bookie.replay(Bookie.java:924)
    at org.apache.bookkeeper.bookie.Bookie.readJournal(Bookie.java:886)
    at org.apache.bookkeeper.bookie.Bookie.start(Bookie.java:943)
    at org.apache.bookkeeper.proto.BookieServer.start(BookieServer.java:141)
    at org.apache.bookkeeper.server.service.BookieService.doStart(BookieService.java:58)
    at org.apache.bookkeeper.common.component.AbstractLifecycleComponent.start(AbstractLifecycleComponent.java:78)
    at org.apache.bookkeeper.common.component.LifecycleComponentStack.lambda$start$2(LifecycleComponentStack.java:113)
    at com.google.common.collect.ImmutableList.forEach(ImmutableList.java:408)
    at org.apache.bookkeeper.common.component.LifecycleComponentStack.start(LifecycleComponentStack.java:113)
    at org.apache.bookkeeper.common.component.ComponentStarter.startComponent(ComponentStarter.java:80)
    at org.apache.bookkeeper.server.Main.doMain(Main.java:229)
    at org.apache.bookkeeper.server.Main.main(Main.java:203)
```

As discussed on slack, it is hard to figure out an exact reason as to why the native JNI call fails with an invalid argument. Hence this PR proposes that the `lastLogMark` is updated to journal EOF instead of an arbitrary LONG.MAX_VALUE. The FileChannel interface defines that the implementors can pass in any long offset and the file handler should return EOF immediately when trying to read it. However it doesn't seem to be working as expected.

### Changes

Updated `Journal#setLastLogMark()` method to accept an `scanOffset` instead of constant `LONG.MAX_VALUE`.

@ivankelly @eolivelli

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
@@ -710,12 +710,13 @@ public class Journal extends BookieCriticalThread implements CheckpointSource {
     }
 
     /**
-     * Update lastLogMark of the journal,
+     * Update lastLogMark of the journal
      * Indicates that the file has been processed.
      * @param id
+     * @param scanOffset
      */
-    void setLastLogMarkToEof(Long id) {
-        lastLogMark.getCurMark().setLogMark(id, Long.MAX_VALUE);
+    void setLastLogMark(Long id, long scanOffset) {
+        lastLogMark.setCurLogMark(id, scanOffset);
     }
 
     /**
@@ -771,7 +772,7 @@ public class Journal extends BookieCriticalThread implements CheckpointSource {
      * @param scanner Scanner to handle entries
      * @throws IOException
      */
-    public void scanJournal(long journalId, long journalPos, JournalScanner scanner)
+    public long scanJournal(long journalId, long journalPos, JournalScanner scanner)
         throws IOException {
         JournalChannel recLog;
         if (journalPos <= 0) {
@@ -832,6 +833,7 @@ public class Journal extends BookieCriticalThread implements CheckpointSource {
                     scanner.process(journalVersion, offset, recBuff);
                 }
             }
+            return recLog.fc.position();
         } finally {
             recLog.close();
         }
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
@@ -770,6 +770,7 @@ public class Journal extends BookieCriticalThread implements CheckpointSource {
      * @param journalId Journal Log Id
      * @param journalPos Offset to start scanning
      * @param scanner Scanner to handle entries
+     * @return scanOffset - represents the byte till which journal was read
      * @throws IOException
      */
     public long scanJournal(long journalId, long journalPos, JournalScanner scanner)
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
@@ -921,10 +921,11 @@ public class Bookie extends BookieCriticalThread {
                 logPosition = markedLog.getLogFileOffset();
             }
             LOG.info("Replaying journal {} from position {}", id, logPosition);
-            journal.scanJournal(id, logPosition, scanner);
-            // Update LastLogMark to Long.MAX_VALUE position after replaying journal
+            long scanOffset = journal.scanJournal(id, logPosition, scanner);
+            // Update LastLogMark after completely replaying journal
+            // scanOffset will point to EOF position
             // After LedgerStorage flush, SyncThread should persist this to disk
-            journal.setLastLogMarkToEof(id);
+            journal.setLastLogMark(id, scanOffset);
         }
     }
 


karanmehta93:
I retried the test failure at this [link](https://builds.apache.org/job/bookkeeper_precommit_pullrequest_validation/org.apache.distributedlog$distributedlog-core/1257/testReport/junit/org.apache.distributedlog/TestBKDistributedLogManager/testDeleteLog/) locally with and without my change, however it fails with the same error. IMHO, I don't think this one is related. 

The integration test suite failed at the very end with the error `ExecutionException There was an error in the forked process` which seems unrelated too. Please advice more. Thanks!

karanmehta93:
run integration tests

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -771,7 +772,7 @@ public void checkpointComplete(Checkpoint checkpoint, boolean compact) throws IO
      * @param scanner Scanner to handle entries
      * @throws IOException
```
eolivelli: Add @return

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -832,6 +833,7 @@ public void scanJournal(long journalId, long journalPos, JournalScanner scanner)
                     scanner.process(journalVersion, offset, recBuff);
                 }
             }
+            return recLog.fc.position();
```
eolivelli: Do we have any corner case ? Empty file, more then one file..

 it better to add a method in JournalChannel, isn't it?
We could then add specific unit tests about this crucial method

eolivelli:
I wonder why this approach wasn't take at the first implementation and we opted for MAX_VALUE

karanmehta93:
@eolivelli Checkout [this](https://docs.oracle.com/javase/7/docs/api/java/nio/channels/FileChannel.html#position(long))

When I first looking at this, I felt that it is much simpler to do LONG.MAX_VALUE instead of finding the exact spot in this way. The interface said nothing wrong about assuming that. Little did I know that the underlying JNI call can still mess it up and fail to obey it (for example, if its a 32 bit system or any specific hardware or OS). Tests passed on my machine too, so it seemed okay. I wonder if someone else can help run these tests locally and help me validate my assumption. Is it just the Jenkins box or there is something really wrong here?

This also raises two more concerns for me. @reddycharan pointed out the second one.
1. What if we allow max journal size to a larger long size than the one value which underlying JNI call wouldn't mess with it? The defaults seem to be conservative enough I guess.
2. We are swallowing the `IOException` when we try to position the pointer at the LONG.MAX_VALUE during initialization. What should be the right approach here?

FYI @jvrao

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
karanmehta93: According to my understanding, every bookie restart creates a new journal file on start of size 16 MB and the first 512 bytes contains header. The method also processes a single file only. IMHO, this is covering all the cases.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
karanmehta93: Done

ivankelly:
rerun integration tests

eolivelli:
For 1) I think we should fail the bookie, it would be a java runtime bug.

For 2) we should fail the bookie

We should also file a bug againts openjdk for the case of the error we are seeing.
You could start a discussion on nio-dev list

ivankelly:
@karanmehta93 
Fix looks good. I have a couple of questions regarding the original patch.
1. What's the motivation for updating the mark? So that if a bookie crashes during recovery the work that has already been done is retained?
2. Why did you move replay out of Journal? The method is calling a bunch of journal methods. It belongs in Journal.

ivankelly:

eolivelli:
Merging

eolivelli:
thank you @karanmehta93 committed to master branch

karanmehta93:
Thanks for the quick turnaround, review and merge. Appreciate it. 
@eolivelli @ivankelly Answers inline

> For 1) I think we should fail the bookie, it would be a java runtime bug.

The problem is that we wouldn't actually know if this will be an actual issue or not. We don't explicitly move the file pointer while writing but we update the in-memory lastLogMark. Not sure what would happen in a scenario when the bookie crashes later and we are trying to replay from a value which cannot be replayed anymore. Potentially dangerous data loss scenario here. If the write itself fails, then we are good. 

> For 2) we should fail the bookie

Let me file an issue for that, indicate the code line. We can discuss further on fixing it there.

> What's the motivation for updating the mark? So that if a bookie crashes during recovery the work that has already been done is retained?

Yes, however this bug is only triggered in case of RO bookie. The scenario is described [here](https://github.com/apache/bookkeeper/issues/2087) and checkout this [comment](https://github.com/apache/bookkeeper/issues/2087#issuecomment-489924525). More relevant discussion [here.](https://github.com/apache/bookkeeper/pull/2090#discussion_r283997469)

> Why did you move replay out of Journal? The method is calling a bunch of journal methods. It belongs in Journal.

Some relevant discussion [here](https://github.com/apache/bookkeeper/pull/2090#discussion_r282595592)

ivankelly:
>>    Why did you move replay out of Journal? The method is calling a bunch of journal methods. It belongs in Journal.

> Some relevant discussion here

@karanmehta93 ok, so it was moved out because some scanners don't update the mark. I think there are better ways to do that than breaking the encapsulation of Journal. This whole area needs a refactor at some stage, because it's all tangled up. For example, entries going to the ledger storage should carry their position in the journal, so that ledger storage itself doesn't need to know anything about the journal. It can decide the mark based on what it's flushing. And journal itself should be an interface, with only 2-3 methods. This would allow other possible implementations, like something backed by an NVRAM disk, etc. All future work though.

karanmehta93:
@eolivelli @ivankelly 
Thanks for your suggestion, I started a nio-dev discussion [here](https://mail.openjdk.java.net/pipermail/nio-dev/2019-June/006226.html). I realized that it is dependent on underlying File system and hence I submitted this [PR](https://github.com/apache/bookkeeper/pull/2107) to know the details. The key difference I found out here is that the local docker service (on my machine) was using `overlay2` storage driver instead of `aufs`. When I changed it, I am able to repro the issue as in Jenkins. According to nio-dev email, the problem is visible on ext4 FS and not on XFS. 

I will investigate further in this and will provide updates here. Meanwhile, feel free to add any info you have.

eolivelli:
@karanmehta93 good work !
Thank you

PR #2107, opened 2019-06-07
Display filesystem and docker info for all jenkins runs

karanmehta93:

Changes to shell commands in few scripts

### Motivation

It was difficult to reproduce jenkins test failures locally and the cause might be due to differences in underlying file system. These commands will output more information to spot the exact differences.

### Changes

Running `df` command with an additional `-T` parameter to output filesystem information
Added `docker info` command to print docker details before every test run


karanmehta93:
@jvrao @sijie Quick review please?

PR #2109, opened 2019-06-08
Testcase for limitStatsLogging.

reddycharan:

bd2b16e880d172d4761461fdbf85c1bd19b24e36 had introduced
'limitStatsLogging' feature, which would be used to limit statslogging,
where it is needed. eg - we would like to suppress logs from PCBC
in Auditor/ReplicationWorker process.

In this commit I'm adding testcases for this feature.


reddycharan:
@pasha-kuznetsov fyi

PR #2113, opened 2019-06-20
Implementation of ZoneawareEnsemblePlacementPolicy

reddycharan:

Implementation of ZoneawareEnsemblePlacementPolicy

- This is meant to be used in public cloud infrastructure in which
Availability zones (AZs) are isolated locations within data center
regions from which public cloud services originate and operate.
- ZoneawareEnsemblePlacementPolicy is a placement policy implementation
which uses zone information for placing ensembles.
- In this PlacementPolicy, it is assumed the networktopology/networklocation
 is going to be two level tree, where first part indicates zoneid
and the second part indicates upgradedomainid.
- Here upgrade domain is a set of bookies in an AZ that can be brought
down together for short term maintenance with no impact to the service.
This would help in enabling parallel patching.
- Upgrade domain is a logic concept/division and it may be mapped to
cloud provider native cluster/grouping like Placement Groups in AWS.

Introduce soft PlacementPolicyAdherence

- for ZoneawarePlacementPolicy if in a writeset the number
of zones are in between minnumofzones and desirednumofzones
then it is considered MEETS_SOFT.
- if all of the writesets have zones >= desirednumofzones
then it is considered MEETS_STRICT
- if a writeset contains < minnumofzones then it is considered
FAIL
- add corresponding metric for PlacementPolicyAdherence.MEETS_SOFT


reddycharan:
@jvrao @sijie @eolivelli 

requesting review.

eolivelli:
Sorry.
I started to review but I did not finish.
I will do today

reddycharan:
Thanks @sijie @eolivelli @jvrao for the review. Since all checks have passed, I'm merging this PR.

hangc0276:
@reddycharan  hi, i am researching for how to use `ZoneawareEnsemblePlacementPolicy`, does it has any guides for new users of how to use the policy?

PR #2119, opened 2019-06-28
[bk-server] Clean up over-replicated ledgers owned by different bookies

rdhabalia:
### Motivation
As described at: https://github.com/apache/pulsar/issues/4632
- Sometimes due to overreplication, bookie contains ledgers which are not owned by that bookie anymore and that bookie is not part of the ensemble-list of those ledgers. In this case, GC finds out those overreplicated ledgers and 
- deletes their index from dbStorage (rocksDB) and 
- tries to delete them from entrylog files.

However, bookie doesn't delete them from entry-log files due to change made in [#870](https://github.com/apache/bookkeeper/issues/870) where bookie avoids deleting ledger if znode of that ledger exists.

Because of that bookie ends up storing large number entrylog files with ledgers which are owned by different bookies. It also cause OOM when GC tries to deal with large number of entry log files.

### Modification

Delete the ledgers if bookie is not part of ensemble list of over-replicated ledgers


rdhabalia:
run pr validation

rdhabalia:
rebuild java11
rebuild java8
run pr validation

rdhabalia:
run integration tests

Issue #2103, opened 2019-05-28
Ledger directory is 100% used


codelipenghui:
**Describe the bug**
A single ledger directory is 100% used, but other directories are used very low

**Screenshots**
![image](https://user-images.githubusercontent.com/12592133/58229119-674f5f00-7d63-11e9-8710-f97142fd7748.png)

**Heap dump**
https://drive.google.com/open?id=11J9OQjCj-pYgJxu0MLzEWg8ge63H71Te

**Bookkeeper version**
4.7.2

hrsakai:
Similar issue has happened in our environment.


Compaction suddenly stoped with following exception.
```
09:23:42.310 [GarbageCollectorThread-11-1] INFO  o.a.b.bookie.GarbageCollectorThread  - Enter minor compaction, suspendMinor false
09:23:42.310 [GarbageCollectorThread-11-1] INFO  o.a.b.bookie.GarbageCollectorThread  - Do compaction to compact those files lower than 0.2
.
.
09:25:17.681 [GarbageCollectorThread-11-1] INFO  o.a.b.bookie.EntryLogCompactor       - Removing entry log 27135 after compaction
09:25:17.810 [GarbageCollectorThread-11-1] INFO  o.a.b.bookie.GarbageCollectorThread  - Removing entry log metadata for 27135
09:25:17.810 [GarbageCollectorThread-11-1] DEBUG o.a.b.bookie.GarbageCollectorThread  - Compacting entry log 6743 below threshold 0.2
09:25:17.984 [GarbageCollectorThread-11-1] ERROR o.a.b.common.util.SafeRunnable       - Unexpected throwable caught
java.lang.IllegalArgumentException: Keys and values must be >= 0
        at org.apache.bookkeeper.util.collections.ConcurrentLongLongHashMap.checkBiggerEqualZero(ConcurrentLongLongHashMap.java:731)
        at org.apache.bookkeeper.util.collections.ConcurrentLongLongHashMap.get(ConcurrentLongLongHashMap.java:141)
        at org.apache.bookkeeper.util.collections.ConcurrentLongLongHashMap.containsKey(ConcurrentLongLongHashMap.java:147)
        at org.apache.bookkeeper.bookie.EntryLogMetadata.containsLedger(EntryLogMetadata.java:51)
        at org.apache.bookkeeper.bookie.EntryLogCompactor$CompactionScannerFactory$1.accept(EntryLogCompactor.java:85)
[...]

PR #2121, opened 2019-07-04
Issue #2103: Avoid stop of entry log compaction

massakam:
### Motivation

As mentioned in #2103, if an exception occurs during compaction of a specific entry log, `GarbageCollectorThread` does not perform compaction of other entry logs until the bookie server is restarted. As a result, the number of entry logs continues to increase and eventually it will run out of disk space.

### Changes

The cause of the compaction stop is that the `compacting` flag remains true if `compactor.compact(entryLogMeta)` throws some exception.
https://github.com/apache/bookkeeper/blob/b2e099bbc7b13f13825fe78ab009ca132cb3a9ba/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java#L504-L519

Therefore, fixed `GarbageCollectorThread` so that it set the compaction flag to false even if compaction of a specific entry log fails.

Master Issue: #2103

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
@@ -512,10 +512,16 @@ public class GarbageCollectorThread extends SafeRunnable {
             // indicates that compaction is in progress for this EntryLogId.
             return;
         }
-        // Do the actual compaction
-        compactor.compact(entryLogMeta);
-        // Mark compaction done
-        compacting.set(false);
+
+        try {
+            // Do the actual compaction
+            compactor.compact(entryLogMeta);
+        } catch (Exception e) {
+            LOG.error("Failed to compact entry log {} due to unexpected error", entryLogMeta.getEntryLogId(), e);
+        } finally {
+            // Mark compaction done
+            compacting.set(false);
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -512,10 +512,16 @@ protected void compactEntryLog(EntryLogMetadata entryLogMeta) {
...
         }
-        // Do the actual compaction
-        compactor.compact(entryLogMeta);
-        // Mark compaction done
-        compacting.set(false);
+
+        try {
+            // Do the actual compaction
+            compactor.compact(entryLogMeta);
+        } catch (Exception e) {
```
eolivelli: Do we need to catch *Exception*
Isn't it enough to use try/finally to fix you problem?

massakam: If we don't catch the exception here, I think that the following for loop is interrupted at that time.
https://github.com/apache/bookkeeper/blob/b2e099bbc7b13f13825fe78ab009ca132cb3a9ba/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java#L441-L465
In our case, there is a problem with a particular entry log, so this loop will always stop when `GarbageCollectorThread` try to compact that log. Because of that, I think that the subsequent entry logs are not compacted every time.

massakam:
@eolivelli Added a unit test.

eolivelli:

PR #2137, opened 2019-07-30
Simplify if statements

vzhikserg:
### Changes

* Simplify `if` statements where it is possible (mainly for `if` statements returning boolean values)
* Remove redundant semicolons
* Remove the `close` method where the resource is used withing the try-with-resources block

### DIFF
--- a/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/OrderedScheduler.java
+++ b/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/OrderedScheduler.java
@@ -377,6 +377,6 @@ public class OrderedScheduler extends OrderedExecutor implements ScheduledExecut
             public void execute(Runnable command) {
                 super.execute(timedRunnable(command));
             }
-        };
+        }
 
 }
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieDeferredSyncTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieDeferredSyncTest.java
@@ -68,7 +68,7 @@ public class BookieDeferredSyncTest extends BookKeeperClusterTestCase {
                 .withLedgerId(ledgerId)
                 .withRecovery(true)
                 .withPassword(new byte[0])
-                .execute());) {
+                .execute())) {
 
             try (LedgerEntries entries = readLh.read(0, n - 1)) {
                 for (int i = 0; i < n; i++) {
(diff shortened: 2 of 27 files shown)


bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/OrderedScheduler.java
```diff
@@ -377,6 +377,6 @@ protected ListeningExecutorService delegate() {
             public void execute(Runnable command) {
                 super.execute(timedRunnable(command));
             }
-        };
+        }
```
sijie: just wondering why our checkstyle rule doesn't catch this?

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieDeferredSyncTest.java
```diff
@@ -121,7 +121,7 @@ private void testClose(boolean force) throws Exception {
                     .withLedgerId(ledgerId)
                     .withRecovery(true)
                     .withPassword(new byte[0])
-                    .execute());) {
+                    .execute())) {
```
sijie: why checkstyle doesn't catch this?

vzhikserg: Currently, there is no check defined to catch this violation. Since the [8.22 version](https://checkstyle.sourceforge.io/releasenotes.html#Release_8.22) of checkstyle it is possible to add the `UnnecessarySemicolonInTryWithResources` module which should prevent this type of violations. I will try to update the version of checkstyle and add this module to the checkstyle.xml in a separate pull request.

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/OrderedScheduler.java
vzhikserg: I could not find any check which can catch this violation.

vzhikserg: I could not find any check which can catch this violation.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieDeferredSyncTest.java
vzhikserg: Currently, there is no check defined to catch this violation. Since the [8.22 version](https://checkstyle.sourceforge.io/releasenotes.html#Release_8.22) of checkstyle it is possible to add the `UnnecessarySemicolonInTryWithResources` module which should prevent this type of violations. I will try to update the version of checkstyle and add this module to the checkstyle.xml in a separate pull request.

sijie:

sijie:

sijie:

eolivelli:
this is good to go, merging now

vzhikserg:
@eolivelli great! thank you

Issue #2145, opened 2019-08-16
Using shell generate an unexpected cookie


lsy180829:
**BUG REPORT**

***Describe the bug***

When I run the order "bookkeeper shell cookie_generate", I got a new cookie, but this new cookie is different from the other cookie in zk.

Order( It belongs to pulsar ) :
./bin/bookkeeper shell cookie_generate -j data2/bookkeeper/journal -l data2/bookkeeper/ledgers -i 9285da1b-e1be-471b-8d11-d30d80212e68 -o /tmp/cookie.x 10.254.3.60:3181

The generated cookie is this:
4
bookieHost: "10.254.17.8:3181"
journalDir: "data2/bookkeeper/journal"
ledgerDirs: "data2/bookkeeper/ledgers"
instanceId: "3ed785b5-eb93-498c-9d57-b38ab8846e7b"

The cookie in zk is this :
4
bookieHost: "10.254.17.8:3181"
journalDir: "data/bookkeeper/journal"
ledgerDirs: "1\tdata/bookkeeper/ledgers"
instanceId: "3ed785b5-eb93-498c-9d57-b38ab8846e7b"

We can find the difference is "1\t" in two ledgerDirs, because of it, After I changed the dir and others operation, starting bookie is fail.

09:26:07.829 [main] ERROR org.apache.bookkeeper.server.Main - Failed to build bookie server
org.apache.bookkeeper.bookie.BookieException$InvalidCookieException: Cookie [4
bookieHost: "10.254.17.8:3181"
journalDir: "data2/bookkeeper/journal"
ledgerDirs: "1\tdata2/bookkeeper/ledgers"
instanceId: "3ed785b5-eb93-498c-9d57-b38ab8846e7b"
] is not matching with [4
bookieHost: "10.254.17.8:3181"
journalDir: "data/bookkeeper/journal"
ledgerDirs: "1\tdata/bookkeeper/ledgers"
instanceId: "3ed785b5-eb93-498c-9d57-b38ab8846e7b"

***To Reproduce***

As BUG says

***Expected behavior***

When I generate the cookie by the shell way, It should include "1\t"

eolivelli:
this is a nasty problem, I am sorry.
maybe you can work around the problem just by using a sym link from data2 to data.
if your original problem is to expand bookie capacity you can add the new directory to the configuration and enable 'storage expansion', this way the bookie will start using the new directory as well

sijie:
@eolivelli I think this is a bug we should fix.

[...]

PR #1794, opened 2018-11-07
[tools] add cookie related commands

sijie:
Descriptions of the changes in this PR:

*Motivation*

In some use cases, you need cookie related tools to create/delete/update/get cookie when handling production issues.
Currently bookkeeper doesn't provide such commands.

*Changes*

Add cookie related commands

- create
- delete
- get
- update
- generate

> ---
> In order to uphold a high standard for quality for code contributions, Apache BookKeeper runs various precommit
> checks for pull requests. A pull request can only be merged when it passes precommit checks. However running all
> the precommit checks can take a long time, some trivial changes don't need to run all the precommit checks. You
> can check following list to skip the tests that don't need to run for your pull request. Leave them unchecked if
> you are not sure, committers will help you:
>
> - [ ] [skip bookkeeper-server bookie tests]: skip testing `org.apache.bookkeeper.bookie` in bookkeeper-server module.
> - [ ] [skip bookkeeper-server client tests]: skip testing `org.apache.bookkeeper.client` in bookkeeper-server module.
> - [ ] [skip bookkeeper-server replication tests]: skip testing `org.apache.bookkeeper.replication` in bookkeeper-server module.
> - [ ] [skip bookkeeper-server tls tests]: skip testing `org.apache.bookkeeper.tls` in bookkeeper-server module.
> - [ ] [skip bookkeeper-server remaining tests]: skip testing all other tests in bookkeeper-server module.
> - [ ] [skip integration tests]: skip docker based integration tests. if you make java code changes, you shouldn't skip integration tests.
> - [ ] [skip build java8]: skip build on java8. *ONLY* skip this when *ONLY* changing files under documentation under `site`.
> - [ ] [skip build java9]: skip build on java9. *ONLY* skip this when *ONLY* changing files under documentation under `site`.
> ---

> ---
> Be sure to do all of the following to help us incorporate your contribution
> quickly and easily:
>
> If this PR is a BookKeeper Proposal (BP):
>
> - [ ] Make sure the PR title is formatted like:
>     `<BP-#>: Description of bookkeeper proposal`
>     `e.g. BP-1: 64 bits ledger is support`
> - [ ] Attach the master issue link in the description of this PR.
> - [ ] Attach the google doc link if the BP is written in Google Doc.
>
> Otherwise:
> 
> - [ ] Make sure the PR title is formatted like:
>     `<Issue #>: Description of pull request`
>     `e.g. Issue 123: Description ...`
[...]

PR #2146, opened 2019-08-20
Issue #2145: Make ledgerDirs in Cookie consistent across generating a…

Ghatage:
…nd setting

### Motivation

In PR #1794 we brought in a way to generate a cookie from command line.
However during generation, we have to also encode the ledger dirs by default.

### Changes

Added encoding of ledger dir paths while setting the ledger dirs in a generated cookie in `GenerateCookieCommand.java`.
Tweaked existing test to include more than one ledger path for test coverage.

Master Issue: #2145

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Cookie.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Cookie.java
@@ -82,7 +82,7 @@ public class Cookie {
         this.instanceId = instanceId;
     }
 
-    private static String encodeDirPaths(String[] dirs) {
+    public static String encodeDirPaths(String[] dirs) {
         StringBuilder b = new StringBuilder();
         b.append(dirs.length);
         for (String d : dirs) {
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/commands/cookie/GenerateCookieCommand.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/commands/cookie/GenerateCookieCommand.java
@@ -116,7 +116,7 @@ public class GenerateCookieCommand extends CookieCommand<Flags> {
             builder.setInstanceId(instanceId);
         }
         builder.setJournalDirs(cmdFlags.journalDirs);
-        builder.setLedgerDirs(Cookie.encodeDirPaths(new String[] { cmdFlags.ledgerDirs }));
+        builder.setLedgerDirs(Cookie.encodeDirPaths(cmdFlags.ledgerDirs.split(",")));
 
         Cookie cookie = builder.build();
         cookie.writeToFile(new File(cmdFlags.outputFile));
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/commands/cookie/GenerateCookieCommand.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/commands/cookie/GenerateCookieCommand.java
@@ -116,7 +116,7 @@ public class GenerateCookieCommand extends CookieCommand<Flags> {
             builder.setInstanceId(instanceId);
         }
         builder.setJournalDirs(cmdFlags.journalDirs);
-        builder.setLedgerDirs(cmdFlags.ledgerDirs);
+        builder.setLedgerDirs(Cookie.encodeDirPaths(new String[] { cmdFlags.ledgerDirs }));
 
         Cookie cookie = builder.build();
         cookie.writeToFile(new File(cmdFlags.outputFile));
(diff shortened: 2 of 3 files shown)


Ghatage:
rebuild java11
rebuild java8

eolivelli:
You have test failure. Please check

Ghatage:
run integration tests

Ghatage:

eolivelli:
We also have problems in integration tests, please check it out

Ghatage:
run integration tests

Ghatage:
@eolivelli All checks have passed.
cc @reddycharan

eolivelli:
@sijie @reddycharan any seconder for this change?

reddycharan:
will review it tonight.

sijie:
Also please add tests for the code change

Ghatage:
> I think you need to split the ledgerDirs by , into a String[] and then pass the array to builder.setLedgerDirs.

@sijie Do you think it would be a good idea if we just pass the string to `builder.setLedgerDirs()`, and then inside it split it by ","  and use that `String[]` to generate and set a single encoded string instead?

sijie:
> Do you think it would be a good idea if we just pass the string to builder.setLedgerDirs(), and then inside it split it by "," and use that String[] to generate and set a single encoded string instead?

I don't think that is a good idea. from a builder view, we should just take`String[]` and not assume the string is a comma-separated string.

Ghatage:
@sijie I feel the current changes which we have here are appropriate because If we pass it a `String[]` we will have to encode it inside setLedgers() and we can't do that because the [read path for the cookie also uses the builder for parsing](https://github.com/apache/bookkeeper/blob/0034fe684136e318d1a7e8b384968d703d2dd8c4/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Cookie.java#L204) and it will tack on an extra `'1\t'` at the beginning when it parses the cookie it read from ZK which already has the encoded ledgerDir string set.

Hence, I feel the current changes which I have done are the [same as what we do when generating cookie when the bookie comes up the first time.](https://github.com/apache/bookkeeper/blob/0034fe684136e318d1a7e8b384968d703d2dd8c4/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Cookie.java#L297)

Please advise.

sijie:
@Ghatage I don't think your analysis is correct. If there are two directories, you should encode the list of ledger dirs with `2/t` not `1/t`. Your current approach will cause it always attach `1/t` which is incorrect.

An easy way to understand the problem here. Please check `Cookie.generateCookie`. GenerateCookieCommand should generate a cookie same as how `Cookie.generateCookie` generate one.

```
    static Builder generateCookie(ServerConfiguration conf)
            throws UnknownHostException {
        Builder builder = Cookie.newBuilder();
        builder.setLayoutVersion(CURRENT_COOKIE_LAYOUT_VERSION);
        builder.setBookieHost(Bookie.getBookieAddress(conf).toString());
        builder.setJournalDirs(Joiner.on(',').join(conf.getJournalDirNames()));
        builder.setLedgerDirs(encodeDirPaths(conf.getLedgerDirNames()));
        return builder;
    }
```

Ghatage:
Thanks @sijie for the explanation, I have now followed the same approach as `Cookie.generateCookie` and added another test for this change.

Ghatage:
run integration tests
run pr validation

eolivelli:
run pr validation
run integration tests

eolivelli:
@sijie is would be good to have this fix in 4.10 and maybe port it to 4.9 branch as well

Ghatage:
run pr validation
run integration tests

eolivelli:
run pr validation

Ghatage:
run pr validation

eolivelli:
run pr validation

eolivelli:
@sijie    it it worth to have this change in 4.10 please take a look

eolivelli:
run pr validation

eolivelli:
@Ghatage we have two checkstyle issues, please fix them, then we are good to go and merge this patch:
```
 
2019-10-25T16:18:27.918 [INFO] --- maven-checkstyle-plugin:3.0.0:check (default-cli) @ bookkeeper-tools-ledger ---
2019-10-25T16:18:29.872 [INFO] Starting audit...
[ERROR] /home/jenkins/jenkins-slave/workspace/bookkeeper_precommit_pullrequest_validation/tools/ledger/src/test/java/org/apache/bookkeeper/tools/cli/commands/cookie/GenerateCookieCommandTest.java:201: First sentence should end with a period. [JavadocStyle]
[ERROR] /home/jenkins/jenkins-slave/workspace/bookkeeper_precommit_pullrequest_validation/tools/ledger/src/test/java/org/apache/bookkeeper/tools/cli/commands/cookie/GenerateCookieCommandTest.java:231:50: ',' is not followed by whitespace. [WhitespaceAfter]



```

Ghatage:

Ghatage:
@eolivelli we're good to go!

PR #2156, opened 2019-09-03
[bookie-mTLS] add support of hostname verification

rdhabalia:
### Motivation

Right now, bookkeeper-client is not able to perform [hostname-verification](https://tersesystems.com/blog/2014/03/23/fixing-hostname-verification/) when it connects to broker over tls. Hostname-verification feature is already implemented in almost all [http-client](https://github.com/apache/httpcomponents-client/blob/master/httpclient5/src/main/java/org/apache/hc/client5/http/ssl/DefaultHostnameVerifier.java) but it's not supported by [netty](https://stackoverflow.com/questions/13315623/netty-ssl-hostname-verification-support) yet. therefore, client should be able to perform hostname-verification as per [RFC-2181](https://tools.ietf.org/html/rfc2818#section-3.1)

### Modifications

- added client configuration to enable hostname-verification (default disable)
- use [http-client](https://github.com/apache/httpcomponents-client/blob/master/httpclient5/src/main/java/org/apache/hc/client5/http/ssl/DefaultHostnameVerifier.java) but it's not supported by [netty](https://stackoverflow.com/questions/13315623/netty-ssl-hostname-verification-support) to perform hostname-validation rather adding custom logic.
- add httpclient-apache dependency into LICENSE and NOTICE files.

### Result

Bookkeeper client will be able to perform hostname verification while creating ssl session with bookie.

### DIFF
--- a/bookkeeper-dist/src/main/resources/LICENSE-all.bin.txt
+++ b/bookkeeper-dist/src/main/resources/LICENSE-all.bin.txt
@@ -282,8 +282,8 @@ Apache Software License, Version 2.
 - lib/com.google.errorprone-error_prone_annotations-2.1.2.jar [40]
 - lib/org.apache.yetus-audience-annotations-0.5.0.jar [41]
 - lib/org.jctools-jctools-core-2.1.2.jar [42]
-- lib/org.apache.httpcomponents-httpclient-4.4.1.jar [43]
-- lib/org.apache.httpcomponents-httpcore-4.4.1.jar [44]
+- lib/org.apache.httpcomponents-httpclient-4.5.5.jar [43]
+- lib/org.apache.httpcomponents-httpcore-4.4.9.jar [44]
 - lib/org.apache.thrift-libthrift-0.9.3.jar [45]
 
 [1] Source available at https://github.com/FasterXML/jackson-annotations/tree/jackson-annotations-2.9.7
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/AuthHandler.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/AuthHandler.java
@@ -457,8 +457,8 @@ class AuthHandler {
                 return true;
             }
             if (LOG.isDebugEnabled()) {
-                LOG.debug("Verifying HostName for {}, Cipher {}, Protocols {}, on {}", hostname, sslSession.getCipherSuite(),
-                        sslSession.getProtocol(), channel);
+                LOG.debug("Verifying HostName for {}, Cipher {}, Protocols {}, on {}", hostname,
+                        sslSession.getCipherSuite(), sslSession.getProtocol(), channel);
             }
             boolean verification = HOSTNAME_VERIFIER.verify(hostname, sslSession);
             if (!verification) {
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/AuthHandler.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/AuthHandler.java
@@ -29,12 +29,16 @@ import io.netty.channel.ChannelDuplexHandler;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelInboundHandlerAdapter;
 import io.netty.channel.ChannelPromise;
+import io.netty.handler.ssl.SslHandler;
 
 import java.io.IOException;
+import java.net.InetSocketAddress;
 import java.net.SocketAddress;
 import java.util.Queue;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.atomic.AtomicLong;
+
+import javax.net.ssl.SSLSession;
 
 import org.apache.bookkeeper.auth.AuthCallbacks;
 import org.apache.bookkeeper.auth.AuthToken;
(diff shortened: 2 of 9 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/AuthHandler.java
```diff
@@ -26,27 +26,34 @@
...
 import org.apache.bookkeeper.auth.ClientAuthProvider;
 import org.apache.bookkeeper.client.BKException;
 import org.apache.bookkeeper.proto.BookkeeperProtocol.AuthMessage;
+import org.apache.http.conn.ssl.DefaultHostnameVerifier;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 class AuthHandler {
     static final Logger LOG = LoggerFactory.getLogger(AuthHandler.class);
+    private static final DefaultHostnameVerifier HOSTNAME_VERIFIER = new DefaultHostnameVerifier();
```
eolivelli: I think it is not worth to add such a dependency for only this class.

Doesn't netty have already this feature?

rdhabalia: yes, that's correct that's why I wanted to exclude all the transitive dependencies from it but then somehow I had some unknown compilation issue. and regarding netty, last time I had some issue with netty hostname-verification feature so, skipped to use it. But I am not able to recollect exact issue so, let me give a try and see alternative.

rdhabalia: I have added `EndpointIdentificationAlgorithm` as mentioned at [netty-doc](https://github.com/netty/netty/blob/4.1/handler/src/main/java/io/netty/handler/ssl/SslContext.java#L952)
```
if (conf.getHostnameVerificationEnabled()) {
            SSLEngine sslEngine = handler.engine();
            SSLParameters sslParameters = sslEngine.getSSLParameters();
            sslParameters.setEndpointIdentificationAlgorithm("HTTPS");
            sslEngine.setSSLParameters(sslParameters);
        }
```

but doesn't seem working in all usecases as I see NPE while hostname verification because `sslSession.getPeerHost()` is null when client side handshake is happening. it's also reproducible in unit-test.

```
Caused by: java.lang.NullPointerException
	at sun.net.util.IPAddressUtil.textToNumericFormatV4(IPAddressUtil.java:49)
	at sun.net.util.IPAddressUtil.isIPv4LiteralAddress(IPAddressUtil.java:241)
	at sun.security.util.HostnameChecker.isIpAddress(HostnameChecker.java:124)
	at sun.security.util.HostnameChecker.match(HostnameChecker.java:92)
	at sun.security.ssl.X509TrustManagerImpl.checkIdentity(X509TrustManagerImpl.java:455)
	at sun.security.ssl.X509TrustManagerImpl.checkIdentity(X509TrustManagerImpl.java:436)
	at sun.security.ssl.X509TrustManagerImpl.checkTrusted(X509TrustManagerImpl.java:252)
	at sun.security.ssl.X509TrustManagerImpl.checkServerTrusted(X509TrustManagerImpl.java:136)
	at io.netty.handler.ssl.OpenSslTlsv13X509ExtendedTrustManager.checkServerTrusted(OpenSslTlsv13X509ExtendedTrustManager.java:239)
	at io.netty.handler.ssl.ReferenceCountedOpenSslClientContext$ExtendedTrustManagerVerifyCallback.verify(ReferenceCountedOpenSslClientContext.java:242)
	at io.netty.handler.ssl.ReferenceCountedOpenSslContext$AbstractCertificateVerifier.verify(ReferenceCountedOpenSslContext.java:655)
```

So, I have to spend some more time to look into it.

bookkeeper-dist/src/main/resources/LICENSE-all.bin.txt
```diff
@@ -315,6 +315,8 @@ Apache Software License, Version 2.
 - lib/com.google.errorprone-error_prone_annotations-2.1.2.jar [48]
 - lib/org.apache.yetus-audience-annotations-0.5.0.jar [49]
 - lib/org.jctools-jctools-core-2.1.2.jar [50]
+- lib/org.apache.httpcomponents-httpclient-4.5.5.jar [51]
```
eolivelli: It turns out in current master that we already have these jars on the server.

Maybe we can accept them now even for the client

conf/bk_server.conf
```diff
@@ -260,6 +260,9 @@ httpServerClass=org.apache.bookkeeper.http.vertx.VertxHttpServer
 # Tls certificate files refresh duration in seconds.
 # tlsCertFilesRefreshDurationSeconds=0
 
+# Enable/Disable hostname verification for tls connection.
+#tlsHostnameVerificationEnabled=false
```
eolivelli: Is this a client parameter or a server parameter? Or both?

sijie:
@rdhabalia can you rebase this pull request?

eolivelli:
@rdhabalia  what's the status of this patch?

conf/bk_server.conf
rdhabalia: yes, it's only client config so, it shouldn't be here. Let me remove it.

rdhabalia:
@eolivelli @sijie I rebased the PR, addressed the comment.

PR #2157, opened 2019-09-05
[bookie] provide option to skip adding entry to journal

rdhabalia:
### Motivation
As described at : https://github.com/apache/pulsar/issues/5128

Right now, bookie serves WAL and provides durability by writing entry to journal before acking the message and it persist the entry to entry log file. bookie also provides below options to improve throughput
- to persist entry with fsync disable 
- writing entry to separate devices for journal and ledger. 

So, persisting entry to journal before acking gives durability but sometimes it becomes bottleneck (mostly due to number of io ops) when user wants to achieve higher throughput in system.

Few users have usecase to achieve high throughput with low publish latency, can compromise with durability and mostly reads from the cache. In that case, if we skip adding entry to journal and acks the message as soon as it's added to ledger cache then user can achieve comparatively very high writes in bookie. and it also helps in pulsar to reduce footprints of bookie while serving such usecases.

So, we want to add option to skip adding entry to journal. We also want to make sure that bookie should be able to handle change in this configuration and should come back gracefully whenever this configuration has been changed.

### Changes
Provide option `journalSkipEntryEnable` using which bookie can skip adding entry to journal while serving addEntry request.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
@@ -592,6 +592,7 @@ public class Journal extends BookieCriticalThread implements CheckpointSource {
     final int maxBackupJournals;
 
     final File journalDirectory;
+    final boolean journalSkipEntryEnable;
     final ServerConfiguration conf;
     final ForceWriteThread forceWriteThread;
     // Time after which we will stop grouping and issue the flush
@@ -655,6 +656,7 @@ public class Journal extends BookieCriticalThread implements CheckpointSource {
         this.ledgerDirsManager = ledgerDirsManager;
         this.conf = conf;
         this.journalDirectory = journalDirectory;
+        this.journalSkipEntryEnable = conf.getJournalSkipEntryEnable();
         this.maxJournalSize = conf.getMaxJournalSizeMB() * MB;
         this.journalPreAllocSize = conf.getJournalPreAllocSizeMB() * MB;
         this.journalWriteBufferSize = conf.getJournalWriteBufferSizeKB() * KB;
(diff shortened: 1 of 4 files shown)


sijie:
Agreed to what @eolivelli commented. In order to implement this feature properly, we have to follow what we have discussed in the BP before. Since this change touches the semantic and guarantees that BookKeeper is provided, a discussion and a BP are required to get the community onboard with this change.

rdhabalia:
Thanks @eolivelli for your inputs. Actually, we have a flexibility to create a small bookie cluster where we can compromise persistent durability so, I have provided option to introduce flag at bookie-server level.

But I think your proposal is better because in that case client can decide durability level and bookie can serve both durable and non-durable usecase at the same time. 

I have also gone through with [BP-14](https://cwiki.apache.org/confluence/display/BOOKKEEPER/BP-14+Relax+durability) but our usecase is more related to skipping add-entry to journal to achieve higher throughput. So, we will need implementation with flag`BYPASS_JOURNAL` as you fully described in your previous reply. So, if you have not worked on it then do you think I can create a PR with those changes so, we can use this client-side feature instead compromising durability at bookie-server side.?

rdhabalia:
@eolivelli I think [BP-14](https://cwiki.apache.org/confluence/display/BOOKKEEPER/BP-14+Relax+durability) is definitely a good proposal but I think we don't want to mix requests coming from both the usecases to one bookie and want bookie to serve one kind of usecase mainly multiple specifications of usecases to achieve highest throughput. 
So, I think adding this server-side option would be worth to have and we can always add BP-14 as an enhancement. any thoughts?
+ @merlimat

sijie:
@rdhabalia 

I don't think the main point of BP-14 is to have client flag and allow mixing workloads. The main point of BP-14 to adjust our protocol to make sure it stick to LAC protocol even bookies are running in a less-durable mode. The change you are proposing is different from adding a flag in journal to bypass fsync. A process crash can easily cause data loss which can mess up LAC protocol, which is much much weaker than skipping fsync on journal. 

At the minimal, we should do the followings:

- not advance LAC for non-durable writes
- add flush method to flush writes and advance LAC only after flush is called.

rdhabalia:
>> At the minimal, we should do the followings: not advance LAC for non-durable writes, add flush method to flush writes and advance LAC only after flush is called.

@sijie 
Ok.. so, if we can make LAC related changes in this PR to make sure bookie doesn't mess up with LAC protocol then we should be fine with the PR. is that correct?

sijie:
>  so, if we can make LAC related changes in this PR to make sure bookie doesn't mess up with LAC protocol then we should be fine with the PR. is that correct?

If you are not taking the approach proposed in BP-14, it would be good to write a BP for your approach before actually implementing the pull request. Because it is a piece that the community has discussed before and people has something already in their mind. Writing a BP for your approach would help align everyone on this.

eolivelli:
We must do the implementation on the bookie and then deliver it to the public client API

eolivelli:
@rdhabalia  what's the status of this patch?

cc @atris

rdhabalia:
@eolivelli I thought #1944 will take care it so, I was waiting for that PR to be merged. We can close this PR and I can create another new one if #1944 will not address it.

eolivelli:
this change relates to #2401

danielnaumau:
I've been searching for a while and I cannot find a proper way how to disable journal in pulsar. What flag should I use now to disable it?

PR #2166, opened 2019-09-23
Enhance deferLedgerLockReleaseOfFailedLedger in ReplicationWorker

reddycharan:

**Issue:** In the past, ReplicationWorker (RW) retrial logic is enhanced to backoff
replication after threshold number of replication failures of a ledger. This is
to help in a pathological situation where data (ledger/entry) is unavailable.
But this is sub-optimal solution, since there is possibility that each RW can
try recovering a ledger threshold number of times, before a RW defers
ledgerLockRelease. Also each time a RW tries to recover it would read entry/fragment
sequentially and writes to new bookies until it finds a missing entry (completely
unavailable) before failing on replication of that ledger. This is done for
each retrial and it bloats the storage and overreplication need to detect and
delete it, which runs once a day by default. So because of this cluster can
run out of storage space and may become RO cluster. Also this puts quite a bit of
load on cluster in vain.

**So the new proposal is to**
- On each RW. remember the state in addition to the counter. State must include the entries which RW failed to read.
- Counter and state must kept around in each RW node. And exponential backup should be used for deferLedgerLockReleaseOfFailedLedger
- During next attempt by RW, try to read the failed entries which is noted in the state. Read must be successful before proceeding replicating.
- With this model we avoid duplicate copies on each attempt. At the most each RW will create only one copy


bookkeeper-server/src/test/java/org/apache/bookkeeper/client/TestLedgerFragmentReplication.java
```diff
@@ -111,7 +111,7 @@ public void testReplicateLFShouldCopyFailedBookieFragmentsToTargetBookie()
         // 0-9 entries should be copy to new bookie
 
         for (LedgerFragment lf : result) {
-            admin.replicateLedgerFragment(lh, lf);
+            admin.replicateLedgerFragment(lh, lf, null);
```
eolivelli: Isn't it better to always pass a non null value?
It will make it clearer the flow of that variable where it is used, we can prevent bugs

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerFragmentReplicator.java
```diff
@@ -328,6 +334,9 @@ public void readComplete(int rc, LedgerHandle lh,
                 if (rc != BKException.Code.OK) {
                     LOG.error("BK error reading ledger entry: " + entryId,
                             BKException.create(rc));
+                    if (entriesUnableToReadForThisLedger != null) {
```
eolivelli: I would require that this variable is already non null, in order to not miss any possible code path that reaches this point

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/TestLedgerFragmentReplication.java
reddycharan: @eolivelli even I've mixed opinions about this null thing. In retrospect, instead of passing ConcurrentSet it would be better if I pass function (callback) to the bottom of stack and call this function incase of read failure. So callers can decide what they want to have in their callback logic. Here instead of passing null, it can pass no-op function.

eolivelli: Sounds good

reddycharan:
rerun tests

reddycharan:
@sijie @eolivelli not sure why the test runs are failing with following exceptions

ERROR: Error cloning remote repo 'origin'
hudson.plugins.git.GitException: Could not init /home/jenkins/jenkins-slave/workspace/bookkeeper_precommit_replication_tests

Suppressed: hudson.remoting.Channel$CallSiteStackTrace: Remote call to H37
		at hudson.remoting.Channel.attachCallSiteStackTrace(Channel.java:1741)

Caused by: java.lang.OutOfMemoryError: unable to create new native thread
	at java.lang.Thread.start0(Native Method)

ERROR: Error cloning remote repo 'origin'
Retrying after 10 seconds
No credentials specified

reddycharan:
@eolivelli @jvrao @sijie 

Addressed comments and added testcases. It is for full review now.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperAdmin.java
```diff
@@ -95,6 +96,7 @@
 
     private static final Logger LOG = LoggerFactory.getLogger(BookKeeperAdmin.class);
     private static final Logger VERBOSE = LoggerFactory.getLogger("verbose");
+    public static final BiConsumer<Long, Long> NOOP_BICONSUMER = (l, e) -> { };
```
eolivelli: nit: this one should be private

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/ReplicationWorker.java
```diff
@@ -270,6 +294,42 @@ private void logBKExceptionAndReleaseLedger(BKException e, long ledgerIdToReplic
...
+        }
+        final CountDownLatch numEntriesToRead = new CountDownLatch(entriesOfThisFragmentUnableToRead.size());
+        final AtomicInteger returnRCValue = new AtomicInteger(BKException.Code.OK);
+        for (long entryIdToRead : entriesOfThisFragmentUnableToRead) {
+            lh.asyncReadEntries(entryIdToRead, entryIdToRead, (rc, ledHan, seq, ctx) -> {
+                long thisEntryId = (Long) ctx;
+                if (rc == BKException.Code.OK) {
+                    entriesUnableToReadForThisLedger.remove(thisEntryId);
+                } else {
+                    LOG.error("Received error: {} while trying to read entry in ReplicationWorker", rc);
```
eolivelli: can we write the LedgerId (and entryId ?)

bookkeeper-server/src/test/java/org/apache/bookkeeper/replication/TestReplicationWorker.java
```diff
@@ -502,10 +506,11 @@ public void testBookiesNotAvailableScenarioForReplicationWorker() throws Excepti
                 @Override
                 public void run() {
                     try {
-                        Thread.sleep(4000);
+                        Thread.sleep(3000);
                         isBookieRestarted.set(true);
+                        Thread.sleep(500);
                         /*
-                         * after sleeping for 4000 msecs, restart one of the
+                         * after sleeping for 3000 msecs, restart one of the
```
eolivelli: do we have a better to wait ?
Just I don't want to see another flaky test on CI

not blocker, the test is already using Thraed.sleep

eolivelli:
It would be better to have an eye from @sijie  and @jvrao  as I am not an user of the re-replication subsystem

reddycharan:
run pr validation

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperAdmin.java
reddycharan: this is used in testcases as well, so I changed it from private to public.

bookkeeper-server/src/test/java/org/apache/bookkeeper/replication/TestReplicationWorker.java
reddycharan: basically here we are starting Bookie after sometime (3 secs), till then the ledger is expected to remain underreplicated. So we should be good with this sleep, since we are not depending on this sleep for an event to happen.

reddycharan:

reddycharan:
run integration tests

reddycharan:

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperAdmin.java
eolivelli: I prefer not to add such kind of generic constants to a public API (BookKeeperAdmin is used on downstream projects).

can you please move it in some Utility class in tests ?

eolivelli:
@reddycharan I have left my last comment about BookKeeperAdmin

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperAdmin.java
reddycharan: Hey @eolivelli I made it private as you suggested.

bookkeeper-server/src/test/java/org/apache/bookkeeper/replication/TestReplicationWorker.java
eolivelli: ok

reddycharan:
run integration tests

reddycharan:

eolivelli:
@merlimat  @ivankelly  @jvrao @jiazhai @sijie PTAL

reddycharan:
@jvrao reviewed this PR in our internal repo, will ask him to approve here as well. Thanks @eolivelli for the follow up.

reddycharan:
Thanks @eolivelli and @jvrao for the review. Will merge this PR.

eolivelli:
Cool

We are ready to roll 4.10 out!

PR #2160, opened 2019-09-12
Make the project buildable with JDK12 and JDK13

eolivelli:
Upgrade Lombok to 1.18.10
Upgrade Groovy and Groovy Compiler and Maven Compiler plugin
Fix an inverted if that prevented **bkctl** to work with JDK11+ and it made GC LOGGING opts not applied with JDK8 (in bkctl command), a regression introduced by #2132

PR #2169, opened 2019-09-26
Fix tests and upgrade libs to make tests pass on JDK13

eolivelli:
- Upgrade Mockito and PowerMock with a version that support JDK12+
- Fix bash script tests


eolivelli:
rebuild java11
rebuild java8

eolivelli:
rebuild java8
rebuild java11

eolivelli:
@ivankelly  can you please take a look to this patch as well ?
it is a follow up of #2160

PR #2172, opened 2019-10-02
Avoid throwing exception when doing EntryLogger.internalReadEntry

merlimat:
### Motivation

In the refactoring part of #1819, the `internalReadEntry()` behavior was changed into throwing an exception when reading an entry from a different ledger. 

This is causing a big performance issue when doing read-head from the ledger storage, because we keep reading from the current entry log until we find an entry from a different ledger.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
@@ -756,37 +756,43 @@ public class EntryLogger {
         }
     }
 
-    private static class EntryLogEntry {
-        final int entrySize;
-        final BufferedReadChannel fc;
-
-        EntryLogEntry(int entrySize, BufferedReadChannel fc) {
-            this.entrySize = entrySize;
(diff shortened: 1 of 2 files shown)


merlimat:

eolivelli:
CI passed at first round!

ivankelly:
Also, the commit message is misleading. It throws the exception when both the ledger ID and entry ID don't match, the latter being the actual problem here.

merlimat:
The check is actually for either of ledgerid or entryid: `if (thisLedgerId != ledgerId || thisEntryId != entryId)`

It would be a problem even when the  ledgerid is different, since we’re scanning on the entry log until we find and entry for a different ledger.

ivankelly:
But for the purposes of readahead, ledger Id being different it ok. You only want to read ahead entries from the same ledger. Throwing on different ledger removes the need for the check in the read ahead code even.

merlimat:
> Throwing on different ledger removes the need for the check in the read ahead code even. 

Sure, that would be functionally correct, though the exception will be still expensive since it will be thrown in a "no-error" code path that gets executed a lot of times.

PR #2173, opened 2019-10-05
Call exceptionHandler if Bookie.start fails with exception.

reddycharan:

When main thread of Bookie/BookieServer has exited with exception while starting Bookie,
Bookie process shouldn't be alive because of any non-daemon thread that has already
started.

### DIFF
--- a/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/component/ComponentStarter.java
+++ b/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/component/ComponentStarter.java
@@ -47,7 +47,7 @@ public class ComponentStarter {
                 component.close();
                 log.info("Closed component {} in shutdown hook successfully. Exiting.", component.getName());
                 FutureUtils.complete(future, null);
-            } catch (Exception e) {
+            } catch (Throwable e) {
                 log.error("Failed to close component {} in shutdown hook gracefully, Exiting anyway",
                     component.getName(), e);
                 future.completeExceptionally(e);
--- a/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/component/ComponentStarter.java
+++ b/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/component/ComponentStarter.java
@@ -19,6 +19,7 @@
 package org.apache.bookkeeper.common.component;
 
 import java.util.concurrent.CompletableFuture;
+
 import lombok.extern.slf4j.Slf4j;
 import org.apache.bookkeeper.common.concurrent.FutureUtils;
 
@@ -72,6 +73,8 @@ public class ComponentStarter {
 
         // register a component exception handler
         component.setExceptionHandler((t, e) -> {
+            log.error("Triggered exceptionHandler of Component: {} because of Exception in Thread: {}",
+                    component.getName(), t, e);
             // start the shutdown hook when an uncaught exception happen in the lifecycle component.
             shutdownHookThread.start();
         });
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/server/service/AutoRecoveryService.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/server/service/AutoRecoveryService.java
@@ -26,6 +26,8 @@ import org.apache.bookkeeper.replication.ReplicationException.UnavailableExcepti
 import org.apache.bookkeeper.server.component.ServerLifecycleComponent;
 import org.apache.bookkeeper.server.conf.BookieConfiguration;
 import org.apache.bookkeeper.stats.StatsLogger;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * A {@link ServerLifecycleComponent} that runs autorecovery.
@@ -33,6 +35,7 @@ import org.apache.bookkeeper.stats.StatsLogger;
 public class AutoRecoveryService extends ServerLifecycleComponent {
 
     public static final String NAME = "autorecovery";
+    private static final Logger LOG = LoggerFactory.getLogger(AutoRecoveryService.class);
 
     private final AutoRecoveryMain main;
 
@@ -45,6 +48,7 @@ public class AutoRecoveryService extends ServerLifecycleComponent {
 
     @Override
     public void setExceptionHandler(UncaughtExceptionHandler handler) {
+        super.setExceptionHandler(handler);
         main.setExceptionHandler(handler);
     }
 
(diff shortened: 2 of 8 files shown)


reddycharan:
@karanmehta93 fyi. 

I'm not able to add you as reviewer.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieInitializationTest.java
```diff
@@ -640,6 +644,67 @@ public void testBookieServiceExceptionHandler() throws Exception {
...
+        /*
+         * Create LifecycleComponent for BookieServer and start it.
+         */
+        LifecycleComponent server = Main.buildBookieServer(bkConf);
+        CompletableFuture<Void> startFuture = ComponentStarter.startComponent(server);
+
+        /*
+         * Since Bookie/BookieServer/BookieService is expected to fail to start
+         */
+        startFuture.get();
```
eolivelli: Isn't this line expected to throw an exception?

reddycharan: No. When exceptionHandler is called, which would trigger [shutdownHookThread](https://github.com/apache/bookkeeper/blob/master/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/component/ComponentStarter.java#L76 ) .

in ComponentShutdownHook.run method, component is closed and then Future is [marked complete](https://github.com/apache/bookkeeper/blob/master/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/component/ComponentStarter.java#L48). Unless, component.close failed with exception this future would complete normally.

eolivelli: That's was my suspect but the test is not clear.
Can we verify that the handler has been called?

eolivelli: The comment reads 'it is expected to fail' but the future is not failing, it is not very clear

reddycharan: > Can we verify that the handler has been called?

It is little tricky to test that since the creation and setting of handler and starting of the component are happening in ComponentStarter.startComponent method. Anyhow I came up with other solution where we check if the particular log line saying "exceptionHandler is triggered" is getting logged or not. as part of this scenario.

> The comment reads 'it is expected to fail' but the future is not failing, it is not very clear

added more detailed description for what we are doing here.

reddycharan:

reddycharan:

reddycharan:
@eolivelli @sijie @jvrao this should be good to go. All checks have passed as well.

karanmehta93:
@eolivelli I am taking a look now, will send the review soon. Thanks!

bookkeeper-server/src/main/java/org/apache/bookkeeper/server/service/BookieService.java
```diff
@@ -56,8 +59,14 @@ public BookieServer getServer() {
     protected void doStart() {
         try {
             this.server.start();
-        } catch (IOException | UnavailableException | InterruptedException | BKException e) {
-            throw new RuntimeException("Failed to start bookie server", e);
+        } catch (Throwable exc) {
+            LOG.error("Got unexpected exception while starting BookieServer", exc);
+            if (uncaughtExceptionHandler != null) {
+                LOG.error("Calling uncaughtExceptionHandler");
```
karanmehta93: nit: add more details to this log line?

bookkeeper-server/src/main/java/org/apache/bookkeeper/server/service/BookieService.java
```diff
@@ -35,6 +35,8 @@
     public static final String NAME = "bookie-server";
 
     private final BookieServer server;
+    //private UncaughtExceptionHandler uncaughtExceptionHandler = null;
```
karanmehta93: nit: remove

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/component/ComponentStarter.java
```diff
@@ -72,6 +73,8 @@ public void run() {
 
         // register a component exception handler
         component.setExceptionHandler((t, e) -> {
+            log.error("Triggered exceptionHandler of Component: {} because of Exception in Thread: {}",
```
karanmehta93: Since we hook up exception handler on every component, in case if multiple of them fail, we might see IllegalStateException when we try to start a thread which is already started.

bookkeeper-server/src/main/java/org/apache/bookkeeper/server/service/AutoRecoveryService.java
```diff
@@ -56,8 +59,14 @@ public AutoRecoveryMain getAutoRecoveryServer() {
     protected void doStart() {
         try {
             this.main.start();
-        } catch (UnavailableException e) {
-            throw new RuntimeException("Can't not start '" + NAME + "' component.", e);
+        } catch (Throwable exc) {
+            LOG.error("Got unexpected exception while starting AutoRecoveryMain", exc);
```
karanmehta93: Lets move this to AbstractLifecycle#start() method, since exceptions can happen in any component start.

eolivelli:
@reddycharan please take a look to @karanmehta93 comments

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/component/ComponentStarter.java
karanmehta93: @reddycharan Even if we share the same reference across all components, multiple of them can fail at the same time, causing the hook to trigger (even though it isn't correctly implemented yet). I am fine with addressing this later since this is not very likely to happen.

bookkeeper-server/src/main/java/org/apache/bookkeeper/server/service/AutoRecoveryService.java
reddycharan: changing that

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/component/ComponentStarter.java
reddycharan: yes, as we deiscussed it would be shared instance.

> multiple of them can fail at the same time, causing the hook to trigger
All we should care about is 'atleast-once' semantics. As long as 'shutdownHookThread' is executed once we should be fine. Anyhow in `LifecycleComponentStack.start` Components are started sequentially, so multiple of them failing at the same time is not going to happen.

karanmehta93: > All we should care about is 'atleast-once' semantics. As long as 'shutdownHookThread' is executed once we should be fine.

Yes this is not a major concern, all I was saying was that it would throw a bunch of `IllegalThreadStateException` at the end of logs, which can be potentially confusing.

> Anyhow in LifecycleComponentStack.start Components are started sequentially, so multiple of them failing at the same time is not going to happen.

It's not just during the start. Since we are passing the handler to the component, component may trigger it anytime during its lifecycle. It's OK to be not overthink much about it for now.

reddycharan:
run integration tests

reddycharan:

eolivelli:
Good.
Merging as soon as possible

eolivelli:
merged to master branch, thank you @reddycharan

PR #2185, opened 2019-10-21
[DLOG] LogReader shouldn't be added to pending if not locking

ivankelly:
The pendingReaders set in BKDistributedLogManager exists so that if
the manager is closed which the lock is being acquired for a reader,
that reader will be closed (even though it hasn't been returned to the
client).

In the case that the reader is opened without a lock, there
is not async action being performed. Previously we were also adding
these readers to the pendingReaders, but they were never being removed
from the pendingReaders, causing a memory leak. This change avoids
adding no-locking readers to pendingReaders.


eolivelli:
This code is new to me, but the change looks good to me.

Unfortunately ASF CI is is bad state.
We need to wait for it to be fixed before merging this change

eolivelli:
rebuild java8

eolivelli:
rebuild java8

eolivelli:

PR #2198, opened 2019-11-06
Fix the log level of not support Sse42Crc32C

wolfstudy:
Signed-off-by: xiaolong.ran <ranxiaolong716@gmail.com>

### Motivation

When users are using Apache Pulsar, they will encounter the following error about `Sse42Crc32C`. But in fact it does not affect the correctness of Apache Pulsar use. But the error level log information will cause confusion for the user.
```
10:12:49.119 [pulsar-ordered-OrderedExecutor-0-0-EventThread] ERROR org.apache.bookkeeper.proto.checksum.CRC32CDigestManager - Sse42Crc32C is not supported, will use a slower CRC32C implementation.
```

### Changes

Replace the log level of `not support Sse42Crc32C` from error to warn.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/CRC32CDigestManager.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/CRC32CDigestManager.java
@@ -45,7 +45,7 @@ class CRC32CDigestManager extends DigestManager {
         super(ledgerId, useV2Protocol, allocator);
 
         if (!Sse42Crc32C.isSupported() && !nonSupportedMessagePrinted) {
-            log.error("Sse42Crc32C is not supported, will use a slower CRC32C implementation.");
+            log.warn("Sse42Crc32C is not supported, will use a slower CRC32C implementation.");
             nonSupportedMessagePrinted = true;
         }
     }
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/CRC32CDigestManager.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/CRC32CDigestManager.java
@@ -32,6 +32,8 @@ import org.apache.commons.lang3.mutable.MutableInt;
 @Slf4j
 class CRC32CDigestManager extends DigestManager {
 
+    private static boolean isSupported = true;
+
     private static final FastThreadLocal<MutableInt> currentCrc = new FastThreadLocal<MutableInt>() {
         @Override
         protected MutableInt initialValue() throws Exception {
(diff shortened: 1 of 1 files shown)


wolfstudy:
@sijie @jiazhai PTAL

sijie:
+1 to what @eolivelli suggested.

wolfstudy:
Thanks @sijie @eolivelli 

Fixed done, PTAL again.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/CRC32CDigestManager.java
```diff
@@ -41,8 +43,14 @@ protected MutableInt initialValue() throws Exception {
 
     public CRC32CDigestManager(long ledgerId, boolean useV2Protocol, ByteBufAllocator allocator) {
         super(ledgerId, useV2Protocol, allocator);
+
+        if (!isSupported) {
```
sijie: I don't think `isSupported` is a good name. it is actually a confusing name in this case. because the flag isn't used for telling whether CRC32C is support. The flag should be used for controlling whether the error log message is printed or not.

I would suggest renaming it to "nonSupportedMessagePrinted" and make the default value as `false`.

Then you can change the logic to:

```
if (!Sse42Crc32C.isSupported() && !nonSupportedMessagePrinted) {
    log.error("Sse42Crc32C is not supported, will use a slower CRC32C implementation.");
    nonSupportedMessagePrinted = true;
}
```

eolivelli:
@sijie let's cherry pick to branch-4.10

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/CRC32CDigestManager.java
```diff
@@ -41,8 +43,10 @@ protected MutableInt initialValue() throws Exception {
 
     public CRC32CDigestManager(long ledgerId, boolean useV2Protocol, ByteBufAllocator allocator) {
         super(ledgerId, useV2Protocol, allocator);
-        if (!Sse42Crc32C.isSupported()) {
+
+        if (!Sse42Crc32C.isSupported() && !nonSupportedMessagePrinted) {
             log.error("Sse42Crc32C is not supported, will use a slower CRC32C implementation.");
```
sijie: Can we still make this logging statement to `warn`?

wolfstudy:
run replication tests

wolfstudy:
run replication tests

wolfstudy:

wolfstudy:
@sijie Done, PTAL again, thanks.

Issue #609, opened 2017-10-17
Cookie should include index directory


ivankelly:
JIRA: https://issues.apache.org/jira/browse/BOOKKEEPER-824

Reporter: Charles X 

after introduced index directory, the cookie isn't upgraded to include the index directories, which potentially break data integrity  

### Comments from JIRA

---
*Sijie Guo* 2016-03-14T05:52:49.500+0000

[~agarg1986] are you able to produce a patch before end of the month?

ivankelly:
@sijie we currently add the index directories to the ledger dirs part of the cookie. This isn't 100% safe though, as if someone swapped around their index directories and ledger directories in the conf, they would lose data.

Bumping to release/4.8.0 in any case.

sijie:
@ivankelly agreed.

Ghatage:
Hi @ivankelly,

If I understand correctly, as of now, we are adding index directories directly to the ledgerDirs config option and instead we want to add another field to the cookie for `indexDirs` and populate it with the entries mentioned for the config option `indexDirectories` ?

sijie:
@Ghatage correct

PR #2203, opened 2019-11-12
Issue #609 Add indexDirs field to cookie

Ghatage:
### Motivation
 Currently users add indexDirectories to ledgerDirectories, if they swap their configs it may result in data loss.

 ### Changes
 - Add indexDirectories field to cookie. Bump up Cookie layout version.
 - Add indexDirs initialization to some test suites
 - Add config option 'enforceCookieIndexDirCheck' to disable checking cookie for indexDirs (for backward compat). Keeping this to default to `false` right now. 
 - Add basic tests to check for compatibility of v5 cookie in v4 cluster.

Once review passes:
- Future plan to create BookieShell tool to update existing cookie to v5 and add indexDir to it.

Opening this as a draft PR right now as this is a WIP.
There are many tests which are going to fail with `InvalidCookieException` because we need to `.setIndexDirName()` in the ServerConfig being passed around in the test.
I will get to them in another following commit once the team approves of the underlying changes.

Master Issue: #609 

Signed off by: Anup Ghatage ghatageanup@gmail.com


Ghatage:
@sijie @ivankelly @eolivelli @jvrao @reddycharan 
This is a draft PR as it's a WIP. Please provide feedback and once underlying code changes are good, I will get to fixing the tests (which will require manual addition of `indexDir` (just like `journalDir` and `ledgerDirs`) to the configs being passed around in the test for them to pass.

tools/ledger/src/test/java/org/apache/bookkeeper/tools/cli/commands/cookie/GetCookieCommandTest.java
```diff
@@ -99,6 +99,7 @@ public void testGetCookieFromExistentCookieFile() throws Exception {
             .setInstanceId("test-instance-id")
             .setJournalDirs("/path/to/journal/dir")
             .setLedgerDirs("/path/to/ledger/dirs")
+                .setIndexDirs("/path/to/index/dirs")
```
eolivelli: nit: it looks like we have some leading spaces

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Cookie.java
```diff
@@ -124,14 +127,28 @@ private boolean verifyLedgerDirs(Cookie c, boolean checkIfSuperSet) {
...
+        }
+
+        if (!checkIfSuperSet) {
+            return indexDirs.equals(c.indexDirs);
+        } else {
+            return isSuperSet(decodeDirPathFromCookie(indexDirs), decodeDirPathFromCookie(c.indexDirs));
+        }
+    }
+
+    private void verifyInternal(Cookie c, boolean checkIfSuperSet, ServerConfiguration conf) throws BookieException.InvalidCookieException {
```
eolivelli: can't we simply pass 'conf.isEnforceCookieIndexDirCheck' instead of the full object ?
why did you choose to pass the reference to the full ServerConfiguration object ?

Ghatage: I passed a conf object for future cookie changes that we bring in, viz. network location for 3AZ work we are doing. 
A conf object would make it easier than to refactor the function definition to accept yet another boolean.

tools/ledger/src/test/java/org/apache/bookkeeper/tools/cli/commands/cookie/GetCookieCommandTest.java
Ghatage: Hmmm looks like a problem with my IntelliJ.. Good catch, will fix it.

eolivelli:
What's the status of this patch?

Ghatage:
Hi @eolivelli,
Thanks for following up!
Sorry for the delay, have been busy - have a few local commits which should be pushed.
Will post an update soon!

eolivelli:
news ?

eolivelli:
rerun failure tests

Ghatage:
Ugh. Ignore. Rebased it and now its a mess. Will open a new PR with the changes.

Issue #901, opened 2017-12-21
Add more test cases on LongZkLedgerIdGenerator


sijie:
#900 fix a bug in LongZkLedgerIdGenerator. However it is a bit difficult to test with a real zookeeper client. It would be good to think about add a test case with mock zookeeper client to cover as many cases as possible.

merlimat:
@sijie in Pulsar we have a `MockZookeeper` class that can be used to trigger ZK failures at a specific point, eg: "fail after 3 steps" https://github.com/apache/incubator-pulsar/blob/f21cb67fcaccd4e77d47a2e1171cf5506dd8ec57/managed-ledger/src/test/java/org/apache/zookeeper/MockZooKeeper.java#L844

sijie:
awesome

PR #2206, opened 2019-11-13
Remove lock from URManager’s state only if lock znode deletion has succeeded

reddycharan:

- in ZkLedgerUnderreplicationManager.releaseUnderreplicatedLedger remove ‘lock’
 from ‘heldLocks’ only if lock znode deletion has succeeded.
- This is needed because, if RW.logBKExceptionAndReleaseLedger fails to delete
 the lock znode, then it needs to be tried once more in the finally block of
 'rereplicate(long ledgerIdToReplicate)’ before giving up and shutting down RW.

### DIFF
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/replication/TestReplicationWorker.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/replication/TestReplicationWorker.java
@@ -892,7 +892,7 @@ public class TestReplicationWorker extends BookKeeperClusterTestCase {
             super(connectString, sessionTimeoutMs, watcher,
                     new BoundExponentialBackoffRetryPolicy(sessionTimeoutMs, sessionTimeoutMs, Integer.MAX_VALUE),
                     new BoundExponentialBackoffRetryPolicy(sessionTimeoutMs, sessionTimeoutMs, 0),
-                    NullStatsLogger.INSTANCE, 1, 0);
+                    NullStatsLogger.INSTANCE, 1, 0, false);
             this.connectString = connectString;
             this.sessionTimeoutMs = sessionTimeoutMs;
             this.watcherManager = watcher;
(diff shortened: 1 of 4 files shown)


reddycharan:
How much this patch is critical?
Should be roll a 4.10.1 hotfix release?

This is not critical fix. Probability of happening is low, we internally configured number of zkOperations retry count to 3, but where as in the community version, it is set to infinite number.

Ghatage:
HI @reddycharan is the `MockZooKeeperClient` class in this change a dupe of [this already existing MockZooKeeperClient class](https://github.com/apache/bookkeeper/blob/9e34c9a7e75b87441c2455139a715e39ff985f43/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookKeeperTest.java#L844)?
It seems there might be some functions which are missing from either, do you think it will be a good idea to consolidate them and use them across test cases? Potentially even for #901 ?
@eolivelli what do you think?

reddycharan:
@Ghatage it is not dupe. I'm creating new class to mimic (mock/inject) certain behavior needed for this particular testcase. Scope of this Mock class is local to this testsuite. So it is not meant to be merged.

eolivelli:
Yes, we can have two classes, they are doing different things.
Having a single MockZookeeperClient can complicate things and maybe affect each other test suites.
So I'd stay as Charan is proposing

bookkeeper-server/src/test/java/org/apache/bookkeeper/replication/TestReplicationWorker.java
```diff
@@ -857,4 +874,205 @@ private void verifyRecoveredLedgers(LedgerHandle lh, long startEntryId,
...
+         * 2, we should be able to replicate to the other bookie.
+         */
+        BookieSocketAddress replicaToKill = LedgerHandleAdapter.getLedgerMetadata(lh).getEnsembles().get(0L).get(0);
+        LOG.info("Killing Bookie", replicaToKill);
+        killBookie(replicaToKill);
+
+        /*
+         * Start RW.
+         */
+        ReplicationWorker rw = new ReplicationWorker(baseConf, bkWithMockZK, false, NullStatsLogger.INSTANCE);
```
eolivelli: Do we need to shutdown this object and the other ones in a finallyblock in order to teardown properly the test?

eolivelli:
@reddycharan do you have time to move forward this work?
It would be awesome to have this fix in 4.11

bookkeeper-server/src/test/java/org/apache/bookkeeper/replication/TestReplicationWorker.java
reddycharan: @eolivelli fixed it

reddycharan:
hey @eolivelli can you approve this PR. 

Also, btw, can I use this "Squash and Merge" option for merging code to master?

eolivelli:
We are using the script not this button.
I will merge your patch.
It is better to not self merge if not needed

Thank you very much!

eolivelli:
I can't merge it now because I don't have my laptop. I will do it as soon as possible

PR #2207, opened 2019-11-15
[Docs] Modify “Using Autoecovery”,recover does not support the specified target bookie

SunDapeng1:
### Motivation 
`bin/bookkeeper shell recover` only receives one parameter i.e. failed bookie. According to [recovery](https://bookkeeper.apache.org/docs/latest/admin/autorecovery/#manual-recovery) it is useless to specify ZooKeeper address and target bookie.
image


sijie:

PR #2211, opened 2019-12-06
bin/bookkeeper shell supports the -conf

SunDapeng1:
### Motivation
`bin/bookkeeper shell ` support `-conf configuration`, but it doesn't work. bk_server.conf is loaded by default, even if the configuration is specified
image
Modify `bin/bookkeeper` to specify the configuration file as an absolute path

### DIFF
--- a/bin/bookkeeper
+++ b/bin/bookkeeper
@@ -108,16 +108,30 @@ shift
 LOCALBOOKIES_CONFIG_DIR="${LOCALBOOKIES_CONFIG_DIR:-/tmp/localbookies-config}"
 if [ ${COMMAND} == "shell" ]; then
   DEFAULT_LOG_CONF=${BK_HOME}/conf/log4j.shell.properties
-  if [[ $1 == "-localbookie"  ]]; then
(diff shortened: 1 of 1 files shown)


sijie:
run integration tests

SunDapeng1:
run integration tests

eolivelli:
run integration tests

eolivelli:
run integration tests

SunDapeng1:
run integration tests

SunDapeng1:
run integration tests

sijie:
ignore ci

eolivelli:
@sijie we have  pushed this PR with 'ignore CI' but this change actually broke "bookkeeper shell" and integration tests are no more able to pass.

I am going to push a revert for this fix
https://github.com/apache/bookkeeper/pull/2265

here you can see that integration tests are passing again.

@SunDapeng1 do you want to restart this work and check why this change does not work ?
In my findings it seems that with this fix there is an endless loop in the bash script in case of "bookkeeper shell" command

PR #2214, opened 2019-12-12
BP-38 Publish Bookie Service Info on Metadata Service

eolivelli:
This is the BP page for BP-38 Publish Bookie Service Info on Metadata Service

Master Issue: #2215 

> ---
>
> - [ ] [skip build java11]: skip build on java11. *ONLY* skip this when *ONLY* changing files under documentation under `site`.
> ---

> ---
> quickly and easily:
>
>
>
> Otherwise:
> 
> 
> ---

### DIFF
--- a/site/bps/BP-38-bookie-endpoint-discovery.md
+++ b/site/bps/BP-38-bookie-endpoint-discovery.md
@@ -18,46 +18,82 @@ to store on the Metadata service a set of name value pairs that describe *availa
 
 ### Public Interfaces
 
-client:
+On the client side, we introduce a new data structure that describes the services
+exposed by a Bookie:
+
+```
 inteface BookieServiceInfo {
     Iterable<String> keys();
     String get(String key, String defaultValue);
 }
 
-in RegistrationClient interface:
+```
 
+In RegistrationClient interface we expose a new method:
+
+```
 BookieServiceInfo getBookieServiceInfo(String bookieId)
+```
 
-in RegistrationManager
+You can derive bookieId from a BookieSocketAddress, in the future it would be possible
+to store bookie ids in Ledger Metadata but this change should be covered in the scope
+of another BP.
+
+On the Bookie side we change the RegistrationManager interface in order to let the Bookie
+advertise the services:
+
+in RegistrationManager class this method 
+```
 void registerBookie(String bookieId, boolean readOnly)
+```
+
 becomes
-void registerBookie(String bookieId, Map<String, String> bookieServieInfo, boolean readOnly)
 
-Notable known keys are:
-- bookie.binary.address: hostname:port
-- bookie.tls.supported: true|false
-- boolie.auth.type: node|sasl
-- bookie.admin.url: protocol://hostname:port
-- bookie.metrics.url:  protocol://hostname:port/metrics
+```
+void registerBookie(String bookieId, BookieServiceInfo bookieServieInfo, boolean readOnly)
+```
+
+It will be up to the implementation of RegistrationManager and RegistrationClient to serialize
+the BookieServiceInfo structure.
+
+For the ZooKeeper based implementation we are going to store such information as a JSON Object
+
+```
+{
+    "property": "value",
+    "property2": "value2",
+}
+```
+Such information will be stored inside the '/available' znode (or 'available_readonline' in case of readonly bookie).
+The rationale around this choice is that the client is watching over these znodes, in the future the client will read the endpoint information (BookieSocketAddress) from this node
+that is updated whenever the Bookie becomes available (like after a restart).
 
 
-_Briefly list any new interfaces that will be introduced as part of this proposal or any existing interfaces that will be removed or changed. The purpose of this section is to concisely call out the public contract that will come along with this feature._
+Notable known keys will be:
+- bookie.binary.url: bk://hostname:port - address of the main Bookie RPC interface
+- bookie.binary.tls.supported: true|false - information about the availability of TLS
+- bookie.binary.auth.type: node|sasl - information about available authentication mechanism
+- bookie.admin.url: protocol://hostname:port - information about the HTTP endpoint
+- bookie.metrics.url:  protocol://hostname:port/metrics - information about Metrics endpoint
 
-A public interface is any change to the following:
+Which kind of properties should be stored in BookieServiceInfo ?
 
-- Data format, Metadata format
-- The wire protocol and api behavior
-- Any class in the public packages
-- Monitoring
-- Command line tools and arguments
-- Anything else that will likely break existing users in some way when they upgrade
+We should store here only the minimal set of information useful to reach the bookie.
+For instance internal state of the Bookie should be exposed by the HTTP endpoint
+so having the *bookie.admin.url* is enough.
+Having information about authentication or security will be useful for PlacementPolicies or for readers in order to select bookies to use.
+It is expected that these properties change when a bookie is restarted, like after a reconfiguration (change auth type, enable TLS).
+It is better not to expose the version of the Bookie in order not to open the way to protocol usage based on the Bookie version (it is better to implement 'feature detection' instead of using the version).
 
 ### Proposed Changes
 
-_Describe the new thing you want to do in appropriate detail. This may be fairly extensive and have large subsections of its own. Or it may be a few sentences. Use judgement based on the scope of the change._
+See the 'Public intefaces' section.
 
 ### Compatibility, Deprecation, and Migration Plan
 
+The proposed change will be compatible with existing clients.
+
+It is possible that future client will need these new properties exposed by the Bookie
 - What impact (if any) will there be on existing users? 
 - If we are changing behavior how will we phase out the older behavior? 
 - If we need special migration tools, describe them here.
--- a/site/bps/BP-38-bookie-endpoint-discovery.md
+++ b/site/bps/BP-38-bookie-endpoint-discovery.md
@@ -18,8 +18,28 @@ to store on the Metadata service a set of name value pairs that describe *availa
 
 ### Public Interfaces
 
-TODO: RegistrationManager, registerBookie(String bookieId, Map\<String, String\>)....
-      RegistrationClient: getBookieEndpoints(String bookidId).
+client:
+inteface BookieServiceInfo {
+    Iterable<String> keys();
+    String get(String key, String defaultValue);
+}
+
+in RegistrationClient interface:
+
+BookieServiceInfo getBookieServiceInfo(String bookieId)
+
+in RegistrationManager
+void registerBookie(String bookieId, boolean readOnly)
+becomes
+void registerBookie(String bookieId, Map<String, String> bookieServieInfo, boolean readOnly)
+
+Notable known keys are:
+- bookie.binary.address: hostname:port
+- bookie.tls.supported: true|false
+- boolie.auth.type: node|sasl
+- bookie.admin.url: protocol://hostname:port
+- bookie.metrics.url:  protocol://hostname:port/metrics
+
 
 _Briefly list any new interfaces that will be introduced as part of this proposal or any existing interfaces that will be removed or changed. The purpose of this section is to concisely call out the public contract that will come along with this feature._
 
--- a/site/bps/BP-38-bookie-endpoint-discovery.md
+++ b/site/bps/BP-38-bookie-endpoint-discovery.md
@@ -1,7 +1,7 @@
 ---
 title: "BP-38: Publish Bookie Service Info on Metadata Service"
 issue: https://github.com/apache/bookkeeper/2215
-state: 'Under Discussion'
+state: 'Accepted'
 release: "4.11.0"
 ---
 
--- a/site/bps/BP-38-bookie-endpoint-discovery.md
+++ b/site/bps/BP-38-bookie-endpoint-discovery.md
@@ -30,9 +30,7 @@ interface BookieServiceInfo {
         String name;
         String hostname;
         int port;
-        String protocol; // "bookie-rpc", "http", "https"....
-        String[] auth; // "sasl-gsapi", "tls-client-auth", "sasl-md5"....
-        String[] extensions; // "starttls"
+        String protocol; // "bookie-rpc", "http", "https"....      
     }
     List<Endpoint> endpoints;
     Map<String, String> properties; // additional properties
@@ -74,17 +72,13 @@ For the ZooKeeper based implementation we are going to store such information in
          "name": "bookie",
          "hostname": "localhost",
          "port": 3181,
-         "protocol": "bookie-rpc",
-         "auth": ["sasl-gsapi"],
-         "extensions": ["starttls"]
+         "protocol": "bookie-rpc"
         },
         {
          "name": "bookie-http-server",
          "hostname": "localhost",
          "port": 8080,
-         "protocol": "http",
-         "auth": [],
-         "extensions": []
+         "protocol": "http"
         }
     ],
     properties {
@@ -106,14 +100,10 @@ It is out of the scope of this proposal to change semantics of ledger metadata, 
 - **hostname**: the hostname (or a raw IP address)
 - **port**: the TCP port (int)
 - **protocol**: the type of service (bookie-rpc, http, https)
-- **auth**: supported authentication types (tls-auth, sasl-gssapi...)
-- **extensions**: protocol extensions, like 'starttls'
 
 **Well known additional properties**
-- **metrics.type**: the type of metrics, for instance 'prometheus' for Prometheus.io metrics
 - **autorecovery.enabled**: 'true' case of the presence of the auto recovery daemon
 
-In the future we could also leverage this information from the PlacementPolicy, in order to select only bookies that publish a TLS endpoint or a particular authentication mechanism.
 The GRPC service would use this mechanism as well.
 
 #### Which kind of properties should be stored in BookieServiceInfo ?
@@ -121,7 +111,7 @@ The GRPC service would use this mechanism as well.
 We should store here only the minimal set of information useful to reach the bookie in an efficient way.
 So we are not storing on the Metadata Service information about the internal state of the server: if you know the address of an HTTP endpoint you can use the REST API to query the Bookie for its state.
 
-These properties may change during the lifetime of a Bookie, think about a configuration (change auth type, enable TLS, change network address) or a dynamic assigned DNS name.
+These properties may change during the lifetime of a Bookie, think about a configuration (change network address) or a dynamic assigned DNS name.
 
 It is better not to expose the version of the Bookie, if the client wants to use particular features of the Bookie this should be implemented on the protocol itself, not just by using the version. The version of the Bookie could be exposed on the HTTP endpoint.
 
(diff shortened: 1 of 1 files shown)


site/bps/BP-38-bookie-endpoint-discovery.md
```diff
@@ -0,0 +1,117 @@
...
+```
+Such information will be stored inside the '/REGISTRATIONPATH/available' znode (or /REGISTRATIONPATH/available/readonline' in case of readonly bookie).
+
+The rationale around this choice is that the client is already using these znodes in order to discover available bookies services.
+
+It is possible that such endpoint information change during the lifetime of a Bookie, like after rebooting a machine and the machine gets a new network address.
+
+It is out of the scope of this proposal to change semantics of ledger metadata, in which we  are currently storing directly the network address of the bookies, but with this proposal we are preparing the way to have an indirection and separate the concept of Bookie ID from the Network address.
+
+**Well known** keys will be:
```
sijie: I think the information should be designed in a way for extensibility rather than just putting key/value pairs in the znode.

Here is something I have in my mind:

```
listeners:
    - name: 'internal'
       hostname: internal_ip
       port: port
       type: bookie
    - name: 'external'
       hostname: external_ip
       port: port
       type: bookie
    - name: 'external_tls'
       hostname: external_ip
       port: port
       type: bookie+tls
    - name: 'http'
       hostname: ip
       port: port
       type: http
network_address:
       path: "/region/rack/host"
properties:
       'user-key': 'user-val'
```

site/bps/BP-38-bookie-endpoint-discovery.md
```diff
@@ -0,0 +1,117 @@
...
+
+For the ZooKeeper based implementation we are going to store such information in JSON format.
+
+```
+{
+    "property": "value",
+    "property2": "value2",
+}
+```
+Such information will be stored inside the '/REGISTRATIONPATH/available' znode (or /REGISTRATIONPATH/available/readonline' in case of readonly bookie).
```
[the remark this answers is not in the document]
eolivelli: @sijie my idea is to use the existing znode.
I have copied current paths, maybe I was wrong with copy and paste.

You can see in the attached PR with a proof of concept.

If we use existing paths the change is minimal and we are still compatible with old versions

site/bps/BP-38-bookie-endpoint-discovery.md
```diff
@@ -0,0 +1,117 @@
...
+The rationale around this choice is that the client is already using these znodes in order to discover available bookies services.
+
+It is possible that such endpoint information change during the lifetime of a Bookie, like after rebooting a machine and the machine gets a new network address.
+
+It is out of the scope of this proposal to change semantics of ledger metadata, in which we  are currently storing directly the network address of the bookies, but with this proposal we are preparing the way to have an indirection and separate the concept of Bookie ID from the Network address.
+
+**Well known** keys will be:
+- **bookie.port**: the TCP port for the Bookie RPC interface
+- **bookie.host**: the TCP network hostname for the Bookie RPC interface
+- **bookie.tls.enabled**: 'true' in case of TLS endpoint support
```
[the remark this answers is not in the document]
eolivelli: I would like to expose such info because this way a client can decide to write (or read) only to bookies that implement TLS.

site/bps/BP-38-bookie-endpoint-discovery.md
```diff
@@ -0,0 +1,117 @@
...
+- **bookie.http.port**: the TCP port for the HTTP Service, if enabled
+- **bookie.stats.provider**: the name of the Stats Provider
+- **bookie.autorecovery.enabled**: 'true' case of the presence of the auto recovery daemon
+- **bookie.auth.provider**: the name of the Authentication provider in use
```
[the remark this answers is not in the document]
eolivelli: The stats provider is to inform about the type of metrics, for instance if you want to know if stats are exported in Prometheus format.
I will use this in Bookkeeper Visual Manager

The auth provider is in order to let the client choose to write (or read) to a bookie with a given auth type, or to choose the right mechanism for the bookie.

site/bps/BP-38-bookie-endpoint-discovery.md
eolivelli: So you are thinking about a more strongly typed structure ?

```
class BookieServiceInfo {
      class Endpoint {
           String name;
           String hostname;
           int port;
           String type;
      }
      Map<String, Endpoint> endpoints; (name => endpoint, no duplicate names allowed)
      String networkAddress;
      Map<String, String > properties;
     
     public void addEndpoint(Endpoint endpoint) ....
     public void setProperty(String key, String value)...
     public void setNetworkAddress(String networkAddress) .....
}
```

I like this idea.
I am not sure about the "network_address" example, 
I see value in "region" and "rack", maybe as separate values

Anyway I would defer introducing those concepts in a first implementation, because it deserves a discussion about how to migrate from the current script based approach

Is it okay for use to use JSON format ?

eolivelli:
@sijie  I have addressed your comments with a new proposal for the data structure, please take a look.

I will update the implementation as soon as we agree on the final proposal.

Thank you

site/bps/BP-38-bookie-endpoint-discovery.md
sijie: I was suggesting JSON or YAML.

eolivelli: @sijie okay.
I prefer JSON, in order not to introduce some new library for writing/reading YAML (and this library will be used by clients as well)

site/bps/BP-38-bookie-endpoint-discovery.md
```diff
@@ -0,0 +1,149 @@
...
+becomes
+
+```
+void registerBookie(String bookieId, boolean readOnly, BookieServiceInfo bookieServieInfo)
+```
+
+It will be up to the implementation of RegistrationManager and RegistrationClient to serialize
+the BookieServiceInfo structure.
+
+For the ZooKeeper based implementation we are going to store such information in JSON format.
```
ivankelly: Can we use protobuf instead to (a) have schema evolution (b) avoid a random mix of serialization formats in the metadata?

site/bps/BP-38-bookie-endpoint-discovery.md
ivankelly: We are already using protobuf for serialization in ZK. We shouldn't mix and match. We should definitely not use YAML.

site/bps/BP-38-bookie-endpoint-discovery.md
```diff
@@ -0,0 +1,149 @@
...
+         "extensions": []
+        }
+    ],
+    properties {
+        "property": "value",
+        "property2": "value2"
+    }
+}
+```
+Such information will be stored inside the '/REGISTRATIONPATH/available' znode (or /REGISTRATIONPATH/available/readonline' in case of readonly bookie), these paths are the same used in 4.10, but in 4.10 we are writing empty z-nodes.
```
fpj: I think you meant to say `/REGISTRATIONPATH/available/readonly`. Is it confusing to have two possible nodes storing it? Why don't you want it to always be in '/REGISTRATIONPATH/available'?

site/bps/BP-38-bookie-endpoint-discovery.md
fpj: It is a good comment, why did you choose JSON?

eolivelli: @ivankelly @fpj
Currently we are using Protobuf for all of the metadata, so it makes sense to use Protobuf even in this case.

Initially this work started with a simpler key/value pair, then  @sijie suggested to introduce a more structured model, and I ended up in JSON.

I will change to Protobuf

site/bps/BP-38-bookie-endpoint-discovery.md
eolivelli: Sorry for the typo.
We are already using these znodes:
`/REGISTRATIONPATH/available`
`/REGISTRATIONPATH/available/readonly`

I don't want to change this logic, it will make harder backward  compatibility

see:
https://github.com/apache/bookkeeper/blob/163db4496d50cae198b80c9d88c14579f6d2e9f3/bookkeeper-server/src/main/java/org/apache/bookkeeper/discover/ZKRegistrationClient.java#L188

site/bps/BP-38-bookie-endpoint-discovery.md
```diff
@@ -0,0 +1,149 @@
...
+
+The rationale around this choice is that the client is already using these znodes in order to discover available bookies services.
+
+It is possible that such endpoint information change during the lifetime of a Bookie, like after rebooting a machine and the machine gets a new network address.
```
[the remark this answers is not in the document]
eolivelli: Yes, I will reword the sentence.

Some cases:
- The new machine may have a new DNS name.
- You added a new TLS endpoint
- You changed the list of available authentication types

site/bps/BP-38-bookie-endpoint-discovery.md
```diff
@@ -0,0 +1,149 @@
...
+
+It is out of the scope of this proposal to change semantics of ledger metadata, in which we  are currently storing directly the network address of the bookies, but with this proposal we are preparing the way to have an indirection and separate the concept of Bookie ID from the Network address.
```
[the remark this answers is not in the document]
eolivelli: We can split the work in two tasks.

Currently I am working on this very first task because I need bookies to advertise their capabilities on the discovery service as in my company we are building a BookKeeper Management system.

To me the very important point is to be aware of future directions and prepare our way for the separation of **BookieID** from **Bookie Network Address.**

I will be happy to follow up and work on the second part.
I am already trying to see how client side components like the PlacementPolicy will be in the second iteration.

site/bps/BP-38-bookie-endpoint-discovery.md
```diff
@@ -0,0 +1,149 @@
...
+We should store here only the minimal set of information useful to reach the bookie.
+For instance internal state of the Bookie should be exposed by the HTTP endpoint
+so having the *bookie.http.port* is enough to know that the Bookie has an HTTP endpoint and how to reach it (up to 4.10 we can assume that the hostname is the same as the bookie.host).
+It is expected that these properties change when a bookie is restarted, like after a reconfiguration (change auth type, enable TLS, change network address).
+
+It is better not to expose the version of the Bookie, if the client wants to use particular features of the Bookie this should be implemented on the protocol itself, not just by using the version. The version of the Bookie could be exposed on the HTTP endpoint.
+
+### Proposed Changes
+
+See the 'Public intefaces' section.
```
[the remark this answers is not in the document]
eolivelli: thanks

site/bps/BP-38-bookie-endpoint-discovery.md
```diff
@@ -0,0 +1,149 @@
...
+
+Bookie server exposes several services and some of them are optional: the binary BookKeeper protocol endpoint, the HTTP service, the StreamStorage service, a Metrics endpoint.
+
+Currently (in BookKeeper 4.10) the client can only discover the main Bookie endpoint:
+the main BookKeeper binary RPC service.
+Discovery of the TCP address is implicit, because the *id* of the bookie is made of the same host:port that point to the TCP address of the Bookie service.
+
+With this proposal we are now introducing a way for the Bookie to advertise the services it exposes, basically with this change the Bookie will be able to store on the Metadata Service a set of name-value pairs that describe the *available services*.
+
+We will also define a set of well know properties that will be useful for futher implementations.
```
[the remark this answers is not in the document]
eolivelli: I have reworked this paragraph, initially I started with a flat list of key-value pair.

The list is below at 'Well known additional properties', like _metrics.type_ and _autorecovery.enabled_

site/bps/BP-38-bookie-endpoint-discovery.md
```diff
@@ -0,0 +1,149 @@
...
+
+This information will be also useful for Monitoring and Management services as it will enable full discovery of the capabilities of all of the Bookies in a cluster just by having read access to the Metadata Service.
+
+### Public Interfaces
+
+On the Registration API, we introduce a new data structure that describes the services
+exposed by a Bookie:
+
+```
+inteface BookieServiceInfo {
```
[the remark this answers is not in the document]
eolivelli: done

eolivelli:
> Looks good to me. One comment on using protobuf vs json.
> 
> It would be good to to remove the direct zookeeper access for this, as it involves a bunch of zookeeper accesses to get this information. Previously it was 1 getChildren to get all writable bookies. This changes to 1 + n. If we had a rest endpoint on the bookies, it becomes a lot more cachable.

@ivankelly I see this cost, but actually we are adding an indirection between the BookieID and the information about how to reach the bookie.

`If we had a rest endpoint on the bookies, it becomes a lot more cachable.`
We would end up in storing the HTTP endpoint of the bookie and so we won't be able to move the bookie to another DNS name

site/bps/BP-38-bookie-endpoint-discovery.md
```diff
@@ -0,0 +1,149 @@
...
+It is possible that such endpoint information change during the lifetime of a Bookie, like after rebooting a machine and the machine gets a new network address.
+
+It is out of the scope of this proposal to change semantics of ledger metadata, in which we  are currently storing directly the network address of the bookies, but with this proposal we are preparing the way to have an indirection and separate the concept of Bookie ID from the Network address.
+
+**Endpoint structure**
+- **name**: this is an id for the service type, like 'bookie'
+- **hostname**: the hostname (or a raw IP address)
+- **port**: the TCP port (int)
+- **protocol**: the type of service (bookie-rpc, http, https)
+- **auth**: supported authentication types (tls-auth, sasl-gssapi...)
```
sijie: What is the difference between `auth` and `extensions`? Can you give an example of how they are used?

ivankelly:
> @eolivelli  If we had a rest endpoint on the bookies, it becomes a lot more cachable.
We would end up in storing the HTTP endpoint of the bookie and so we won't be able to move the bookie to another DNS name

Sorry, I wasn't clear. I meant (and it's out of scope for this), that we have a http rest endpoint on each bookie, something like /bookies which returns the list of bookies available and all their available endpoints. When a client is creating a ledger, it doesn't need to query ZK (1 + n times), it just hits this endpoint to get the list. The client shouldn't have to know the actual individual bookie DNS addresses. This would be behind a VIP of some kind (like a loadbalancer pointing to a statefulset in k8s).

site/bps/BP-38-bookie-endpoint-discovery.md
```diff
@@ -0,0 +1,149 @@
...
+
+It is better not to expose the version of the Bookie, if the client wants to use particular features of the Bookie this should be implemented on the protocol itself, not just by using the version. The version of the Bookie could be exposed on the HTTP endpoint.
+
+### Proposed Changes
+
+See the 'Public interfaces' section.
+
+### Compatibility, Deprecation, and Migration Plan
+
+The proposed change will be compatible with existing clients up to 4.10.x.
```
ivankelly: @eolivelli Where is the break? 4.10.x is a very recent version.

eolivelli: There is no break, we will be compatible with every version of BK, up to 4.3 (personally I don't know how it worked before 4.3).
Maybe the English is not good ?
I meant to say that this change will be compatible with every client from xxx (4.3?) up to 4.10

eolivelli:
> > @eolivelli  If we had a rest endpoint on the bookies, it becomes a lot more cachable.
> > We would end up in storing the HTTP endpoint of the bookie and so we won't be able to move the bookie to another DNS name
> 
> Sorry, I wasn't clear. I meant (and it's out of scope for this), that we have a http rest endpoint on each bookie, something like /bookies which returns the list of bookies available and all their available endpoints. When a client is creating a ledger, it doesn't need to query ZK (1 + n times), it just hits this endpoint to get the list. The client shouldn't have to know the actual individual bookie DNS addresses. This would be behind a VIP of some kind (like a loadbalancer pointing to a statefulset in k8s).

Got it

site/bps/BP-38-bookie-endpoint-discovery.md
ivankelly: It's a little unclear. The fact that you state "up to 4.10.x", implies that anything after 4.10.x is _not_ compatible. So then I took it to mean the inverse, 4.10.x onwards. This should state that it's compatible with all clients from 4.3 onwards. Maybe a note on what the existing clients will do (i.e. just read the list of znodes, which doesn't change). I think the last break in this area was to use hostnames, and I don't recall that even broke anything.

site/bps/BP-38-bookie-endpoint-discovery.md
eolivelli: @sijie 
**Auth** from my point of view describes how the bookie is able to perform *Authentication*, that is how to bind an user to the connection/request, it is useful in order to select Bookie for which the client has the request auth type available and for monitoring purposes (I want to see if all of the bookies are exposing the expected auth type)

**Extensions** is more generic and it will depends on the protocol, in the case of bookie-rpc we have the ability to use TLS, but it is only an _extension_ to the base protocol. In fact we are using the StartTLS mechanism, if we had a "pure" TLS connection I would add a new *protocol* named "bookie-rpcs" or something like that.

site/bps/BP-38-bookie-endpoint-discovery.md
eolivelli: @ivankelly I have removed the reference to the version.
Now it reads 'The proposed change will be compatible with existing clients'

site/bps/BP-38-bookie-endpoint-discovery.md
sijie: regarding `Auth`, if it is only for monitoring purpose. it should be done by exposing metrics for that. If we are exposing this to the bookie service info, we should describe how clients are going to use this. If there is no clear use cases for using this, we should try to avoid adding them right now.

if you have a new protocol, that would be new endpoint, the endpoint name should already  be used for distinguishing different endpoints. I don't know we need an extension.

sijie:
@eolivelli I am not convinced by the `Auth` and `Extensions`. Unless there is a strong reason to add them, I would suggest removing them for now to push this change forward.

eolivelli:
@sijie 
Okay.
I will remove them.
We do not need them currently

site/bps/BP-38-bookie-endpoint-discovery.md
ivankelly: Sure, sgtm.

site/bps/BP-38-bookie-endpoint-discovery.md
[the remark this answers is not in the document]
eolivelli: removed

site/bps/BP-38-bookie-endpoint-discovery.md
eolivelli: Dropped *Auth* and *Extensions* from the document

eolivelli:
@sijie @fpj 
I have addressed all of your comments.

I think we are on the same page.
I will mark this BP as "Accepted" and commit it to the website once I get your approvals.

In the meantime I can start the implementation.

eolivelli:
thank you guys.
I will proceed with the implementation.
I am marking this BP as "Accepted" and I will merge to the website

PR #2240, opened 2020-01-10
fix the problem that the method InetAddress.getLocalHost().getHostAddress() gets ip as 127.0.0.1

liudezhi2098:
### Motivation
In some cases, Bookie obtained the local IP of 127.0.0.1, which caused an abnormal report status.
get ip method:
`InetAddress.getLocalHost().getHostAddress();`
host
```
 127.0.0.1   localhost localhost.localdomain localhost4 localhost4.localdomain4
  ::1         localhost localhost.localdomain localhost6 localhost6.localdomain6
```
_Although it can be solved by setting up the host, but the customer does not necessarily set it when using it, so if you do not rely on this when obtaining the IP, it will be much more friendly_

fix the problem that the method InetAddress.getLocalHost().getHostAddress() gets ip as 127.0.0.1

### Changes

Added a method to get the local IP in NetUtils, and replaced the other places to get the IP


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
```diff
@@ -466,7 +466,7 @@ private void fillReadAheadCache(long orginalLedgerId, long firstEntryId, long fi
             long size = 0;
 
             while (count < readAheadCacheBatchSize && currentEntryLogId == firstEntryLogId) {
-                ByteBuf entry = entryLogger.internalReadEntry(orginalLedgerId, -1, currentEntryLocation,
+                ByteBuf entry = entryLogger.internalReadEntry(orginalLedgerId, firstEntryId, currentEntryLocation,
```
eolivelli: This change is not part of this task.

bookkeeper-server/src/main/java/org/apache/bookkeeper/net/NetUtils.java
```diff
@@ -85,4 +85,52 @@ public static String resolveNetworkLocation(DNSToSwitchMapping dnsResolver, Inet
...
+                return ip;
+            }
+            throw new UnknownHostException("InetAddress java.net.InetAddress.getLocalHost() throws UnknownHostException,"
+                    + e.getMessage());
+        }
+    }
+    
+    private static boolean isboolIp(String ipAddress) {
+        String ip = "([1-9]|[1-9]\\d|1\\d{2}|2[0-4]\\d|25[0-5])(\\.(\\d|[1-9]\\d|1\\d{2}|2[0-4]\\d|25[0-5])){3}";
+        Pattern pattern = Pattern.compile(ip);
```
eolivelli: You can make this variable static

bookkeeper-server/src/main/java/org/apache/bookkeeper/net/NetUtils.java
```diff
@@ -85,4 +85,52 @@ public static String resolveNetworkLocation(DNSToSwitchMapping dnsResolver, Inet
...
     }
 
+   public static String getLocalHost() throws UnknownHostException {
+        try {
+            String ip = InetAddress.getLocalHost().getHostAddress();
+            if (ip.equals("127.0.0.1")) {
+                return getAddress();
+            } else
+                return ip;
+        } catch (Throwable e) {
```
eolivelli: Please catch the exact expected exception type

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
liudezhi2098: ok , removed

bookkeeper-server/src/main/java/org/apache/bookkeeper/net/NetUtils.java
liudezhi2098: You are right

bookkeeper-server/src/main/java/org/apache/bookkeeper/net/NetUtils.java
liudezhi2098: This exception is superfluous and has been removed

bookkeeper-server/src/main/java/org/apache/bookkeeper/net/NetUtils.java
eolivelli: You are missing this change

bookkeeper-server/src/main/java/org/apache/bookkeeper/net/NetUtils.java
```diff
@@ -85,4 +86,44 @@ public static String resolveNetworkLocation(DNSToSwitchMapping dnsResolver, Inet
         return rNames.get(0);
     }
 
+    public static String getLocalHost() throws UnknownHostException {
+        String ip = InetAddress.getLocalHost().getHostAddress();
+        if (ip.equals("127.0.0.1")) {
```
eolivelli: We should use a better way, this approach does not deal with the whole 127.0.0.0/24 class and with ipv6

I think there is some InteAddress.isLoopback method

bookkeeper-server/src/main/java/org/apache/bookkeeper/net/NetUtils.java
```diff
@@ -85,4 +86,44 @@ public static String resolveNetworkLocation(DNSToSwitchMapping dnsResolver, Inet
...
+                    while (addresses.hasMoreElements()) {
+                        String ip = addresses.nextElement().getHostAddress();
+                        if (isboolIp(ip)) {
+                            return ip;
+                        }
+                    }
+                }
+            }
+            throw new RuntimeException("InetAddress java.net.InetAddress.getLocalHost() throws UnknownHostException");
+        } catch (Throwable e) {
```
eolivelli: Please catch only specific exceptions

bookkeeper-server/src/main/java/org/apache/bookkeeper/net/NetUtils.java
liudezhi2098: sorry, already change

bookkeeper-server/src/main/java/org/apache/bookkeeper/net/NetUtils.java
liudezhi2098: ok，catch  SocketException

bookkeeper-server/src/main/java/org/apache/bookkeeper/net/NetUtils.java
liudezhi2098: You are right

liudezhi2098:
retest this please

bookkeeper-server/src/main/java/org/apache/bookkeeper/net/NetUtils.java
```diff
@@ -85,4 +92,46 @@ public static String resolveNetworkLocation(DNSToSwitchMapping dnsResolver, Inet
...
+        String ip = InetAddress.getLocalHost().getHostAddress();
+        if (InetAddress.getLocalHost().isLoopbackAddress()) {
+            return getAddress();
+        } else {
+            return ip;
+        }
+
+    }
+
+    private static boolean isboolIp(String ipAddress) {
```
eolivelli: please call this method **acceptIpAddress** or **isIpAddressValid** or **canUseIpAddress**
please make it visibility "package" and add unit tests.

from the regexp I see that we won't allow IPv6 addresses

bookkeeper-server/src/main/java/org/apache/bookkeeper/net/NetUtils.java
```diff
@@ -85,4 +92,46 @@ public static String resolveNetworkLocation(DNSToSwitchMapping dnsResolver, Inet
...
+        Matcher matcher = pattern.matcher(ipAddress);
+        return matcher.matches();
+    }
+
+    private static String getAddress() throws UnknownHostException {
```
eolivelli: please call this method **getValidAddress** or something like that

eolivelli:
@liudezhi2098  do you have time to fix my comments.
It would be good to have this patch delivered in 4.11.0

bookkeeper-server/src/main/java/org/apache/bookkeeper/net/NetUtils.java
liudezhi2098: I will call this method isIpAddressValid

bookkeeper-server/src/main/java/org/apache/bookkeeper/net/NetUtils.java
liudezhi2098: Already dealt with at  910f6c7

liudezhi2098:
retest this please

bookkeeper-server/src/main/java/org/apache/bookkeeper/net/NetUtils.java
```diff
@@ -85,4 +88,45 @@ public static String resolveNetworkLocation(DNSToSwitchMapping dnsResolver, Inet
         return rNames.get(0);
     }
 
-}
+    public static String getLocalHost() throws UnknownHostException {
+        String ip = InetAddress.getLocalHost().getHostAddress();
+        if (InetAddress.getLocalHost().isLoopbackAddress()) {
```
pkumar-singh, line 93: How about something like this?
return (InetAddress.getLocalHost().isLoopbackAddress()) ?  getValidAddress() : InetAddress.getLocalHost().getHostAddress();

PR #2257, opened 2020-01-29
2178: IOException Should Be Thrown When Journal Is Corrupted

### LINKED ISSUE #2178: JournalChannel should throw IOException if it cannot seek to desired position
**BUG REPORT**

***Describe the bug***

Here's a piece of code from `JournalChannel.java`
```
try {
    if (position == START_OF_FILE) {
        if (formatVersion >= V5) {
            fc.position(HEADER_SIZE);
        } else if (formatVersion >= V2) {
            fc.position(VERSION_HEADER_SIZE);
        } else {
            fc.position(0);
        }
    } else {
        fc.position(position);
    }
} catch (IOException e) {
    LOG.error("Bookie journal file can seek to position :", e); <-- Swallows IOException and moves on
}
```

***To Reproduce***

Create a corrupt journal file. Try seeking to the desired position.

***Expected behavior***

IOException should be thrown and bookie should be shutdown. Not sure if a retry would help. It's not really recoverable since journal file is corrupted. 

@reddycharan @jvrao @eolivelli Another potential problem?

atris:
Fixes #2178

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/JournalChannel.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/JournalChannel.java
@@ -160,7 +160,7 @@ class JournalChannel implements Closeable {
                         + " suddenly appeared, is another bookie process running?");
             }
             randomAccessFile = new RandomAccessFile(fn, "rw");
-            fc = randomAccessFile.getChannel();
+            fc = openFileChannel(randomAccessFile);
             formatVersion = formatVersionToWrite;
 
             int headerSize = (V4 == formatVersion) ? VERSION_HEADER_SIZE : HEADER_SIZE;
@@ -178,7 +178,7 @@ class JournalChannel implements Closeable {
             fc.write(zeros, nextPrealloc - journalAlignSize);
         } else {  // open an existing file
             randomAccessFile = new RandomAccessFile(fn, "r");
-            fc = randomAccessFile.getChannel();
+            fc = openFileChannel(randomAccessFile);
             bc = null; // readonly
 
             ByteBuffer bb = ByteBuffer.allocate(VERSION_HEADER_SIZE);
@@ -290,5 +290,13 @@ class JournalChannel implements Closeable {
             }
             this.lastDropPosition = newDropPos;
         }
+    }
+
+    public static FileChannel openFileChannel(RandomAccessFile randomAccessFile) {
+        if (randomAccessFile == null) {
+            throw new IllegalArgumentException("Input cannot be null");
+        }
+
+        return randomAccessFile.getChannel();
     }
 }
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/JournalChannel.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/JournalChannel.java
@@ -23,6 +23,8 @@ package org.apache.bookkeeper.bookie;
 
 import static com.google.common.base.Charsets.UTF_8;
 
+import com.google.common.annotations.VisibleForTesting;
+
 import java.io.Closeable;
 import java.io.File;
 import java.io.IOException;
@@ -30,6 +32,7 @@ import java.io.RandomAccessFile;
 import java.nio.ByteBuffer;
 import java.nio.channels.FileChannel;
 import java.util.Arrays;
+
 import org.apache.bookkeeper.util.NativeIO;
 import org.apache.bookkeeper.util.ZeroBuffer;
 import org.slf4j.Logger;
@@ -292,6 +295,7 @@ class JournalChannel implements Closeable {
         }
     }
 
+    @VisibleForTesting
     public static FileChannel openFileChannel(RandomAccessFile randomAccessFile) {
         if (randomAccessFile == null) {
             throw new IllegalArgumentException("Input cannot be null");
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/JournalChannel.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/JournalChannel.java
@@ -224,6 +224,7 @@ class JournalChannel implements Closeable {
                 }
             } catch (IOException e) {
                 LOG.error("Bookie journal file can seek to position :", e);
+                throw e;
             }
         }
         if (fRemoveFromPageCache) {
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieJournalTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieJournalTest.java
@@ -47,12 +47,19 @@ import org.apache.bookkeeper.util.IOUtils;
 import org.apache.commons.io.FileUtils;
 import org.junit.After;
 import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mockito;
+import org.powermock.api.mockito.PowerMockito;
+import org.powermock.core.classloader.annotations.PrepareForTest;
+import org.powermock.modules.junit4.PowerMockRunner;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
  * Test the bookie journal.
  */
+@RunWith(PowerMockRunner.class)
+@PrepareForTest(JournalChannel.class)
 public class BookieJournalTest {
     private static final Logger LOG = LoggerFactory.getLogger(BookieJournalTest.class);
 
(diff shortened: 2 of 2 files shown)


atris:
> Can you please referente the originale issue in the description of this pr.
Done
> Also we need a test that demonstrate the fix and an explanation about how it works

I wanted to check on that. Looking in BookieJournalTest, can I use writeV5Journal API to construct a corrupted file and test that the bookie closes?

eolivelli:
Yes it can be a good place to put the test

atris:
@eolivelli Need a bit of help with the test. I have created a test such as:
https://gist.github.com/atris/773e1074edd1d50786ff35e300495984

However, I was not able to trigger the IOException that this PR added. What am I doing wrong, please?

sijie:
@atris I think the issue was thrown when attempting to seek to a position doesn't exist. so I guess you can truncate the file to reproduce.

atris:
> @atris I think the issue was thrown when attempting to seek to a position doesn't exist. so I guess you can truncate the file to reproduce.

@sijie Thanks, I updated the test but FileChannel.position then moves the file pointer to the asked location. What am I missing, please?

https://gist.github.com/atris/773e1074edd1d50786ff35e300495984

atris:
@sijie @eolivelli Updated the PR, please see and comment

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/JournalChannel.java
```diff
@@ -160,7 +176,13 @@ private JournalChannel(File journalDirectory, long logId,
                         + " suddenly appeared, is another bookie process running?");
             }
             randomAccessFile = new RandomAccessFile(fn, "rw");
-            fc = randomAccessFile.getChannel();
+
+            if (dummyFC == null) {
+                fc = randomAccessFile.getChannel();
```
eolivelli: What about using Powermock in order to hack the result of getChannel ?

I feel this way of testing is quite invasive

atris:
The issue there would be that the RandomFileAccess instance used to create
the FileChannel is also created in the same context so we will probably
have to powermock up the call stack.

I felt this way is a longer term way to test different cases of FileChannel
implementations especially since FileChannel is an external class not owned
by our codebase.

Thoughts?
-- 
Regards,

Atri
*l'apprenant*

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/JournalChannel.java
eolivelli: You can add a static method like:

FileChannel openFileChannel (RandomAccessFile)

Then you can override it with PowerMock

atris: @eolivelli Fixed, thanks

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieJournalTest.java
```diff
@@ -794,4 +801,50 @@ private void testPartialFileInfoPostV3Journal(boolean truncateMasterKey)
...
+        }
+    }
+
+    private class DummyJournalScan implements Journal.JournalScanner {
+        boolean printJournalVersion = false;
+
+        @Override
+        public void process(int journalVersion, long offset, ByteBuffer entry) throws IOException {
+            if (!printJournalVersion) {
+                System.out.println("Journal Version : " + journalVersion);
```
eolivelli: Use Logger please

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieJournalTest.java
```diff
@@ -794,4 +801,50 @@ private void testPartialFileInfoPostV3Journal(boolean truncateMasterKey)
...
+                .thenThrow(new IOException());
+
+        PowerMockito.mockStatic(JournalChannel.class);
+        PowerMockito.when(JournalChannel.openFileChannel(Mockito.any(RandomAccessFile.class))).thenReturn(fileChannel);
+
+        Bookie b = new Bookie(conf);
+        for (Journal journal : b.journals) {
+            List<Long> journalIds = journal.listJournalIds(journal.getJournalDirectory(), null);
+
+            assert journalIds.size() == 1;
```
eolivelli: Use assertEquals

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/JournalChannel.java
```diff
@@ -290,4 +291,12 @@ public void forceWrite(boolean forceMetadata) throws IOException {
             this.lastDropPosition = newDropPos;
         }
     }
+
+    public static FileChannel openFileChannel(RandomAccessFile randomAccessFile) {
```
eolivelli: Add VisibleForTesting comment

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieJournalTest.java
```diff
@@ -794,4 +801,50 @@ private void testPartialFileInfoPostV3Journal(boolean truncateMasterKey)
             // correct behaviour
         }
     }
+
+    @Test(expected = IOException.class)
```
eolivelli: Is is too ambiguous.
We ate not sure that only the last line threw this exception.

Please move the assertion exactly around the expected invocation

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieJournalTest.java
```diff
@@ -794,4 +801,50 @@ private void testPartialFileInfoPostV3Journal(boolean truncateMasterKey)
...
+        Journal.JournalScanner journalScanner = new DummyJournalScan();
+        FileChannel fileChannel = PowerMockito.mock(FileChannel.class);
+
+        PowerMockito.when(fileChannel.position(Mockito.anyLong()))
+                .thenThrow(new IOException());
+
+        PowerMockito.mockStatic(JournalChannel.class);
+        PowerMockito.when(JournalChannel.openFileChannel(Mockito.any(RandomAccessFile.class))).thenReturn(fileChannel);
+
+        Bookie b = new Bookie(conf);
```
eolivelli: Shall we stop the Bookie in a finally block?

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieJournalTest.java
```diff
@@ -794,4 +801,50 @@ private void testPartialFileInfoPostV3Journal(boolean truncateMasterKey)
...
+            List<Long> journalIds = journal.listJournalIds(journal.getJournalDirectory(), null);
+
+            assert journalIds.size() == 1;
+
+            journal.scanJournal(journalIds.get(0), Long.MAX_VALUE, journalScanner);
+        }
+    }
+
+    private class DummyJournalScan implements Journal.JournalScanner {
+        boolean printJournalVersion = false;
```
eolivelli: This flag looks useless

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieJournalTest.java
atris: This annotation works at a class level -- not sure how I can make it to the specific line?

atris:
@eolivelli Updated, please see and let me know your thoughts

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieJournalTest.java
eolivelli: It is not at **class** level, but it works around this **method**

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieJournalTest.java
```diff
@@ -794,4 +801,53 @@ private void testPartialFileInfoPostV3Journal(boolean truncateMasterKey)
...
+                .thenThrow(new IOException());
+
+        PowerMockito.mockStatic(JournalChannel.class);
+        PowerMockito.when(JournalChannel.openFileChannel(Mockito.any(RandomAccessFile.class))).thenReturn(fileChannel);
+
+        Bookie b = null;
+
+        try {
+            b = new Bookie(conf);
+        } finally {
```
eolivelli: you should shut down the bookie after the " for (Journal journal : b.journals) {" loop

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieJournalTest.java
```diff
@@ -794,4 +801,53 @@ private void testPartialFileInfoPostV3Journal(boolean truncateMasterKey)
...
+            b.shutdown();
+        }
+
+        for (Journal journal : b.journals) {
+            List<Long> journalIds = journal.listJournalIds(journal.getJournalDirectory(), null);
+
+            assertEquals(journalIds.size(), 1);
+
+            journal.scanJournal(journalIds.get(0), Long.MAX_VALUE, journalScanner);
```
eolivelli: try {
    journal.scanJournal(journalIds.get(0), Long.MAX_VALUE, journalScanner) ;
   fail("should not be able to scan the journal")
}  catch (IOException err) {
   // expected
   // maybe you can perform some assertion about the err.getMessage() result, but maybe it depends on java version or locale
}

atris: I tried that but it seems too brittle a check for depending on the error message.

atris:
@eolivelli Updated, please see and let me know your thoughts

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieJournalTest.java
```diff
@@ -794,4 +801,63 @@ private void testPartialFileInfoPostV3Journal(boolean truncateMasterKey)
...
+
+        PowerMockito.mockStatic(JournalChannel.class);
+        PowerMockito.when(JournalChannel.openFileChannel(Mockito.any(RandomAccessFile.class))).thenReturn(fileChannel);
+
+        Bookie b = null;
+
+        try {
+            b = new Bookie(conf);
+        } finally {
+            b.shutdown();
```
eolivelli: Why are you shutting down the bookie twice?

atris: Oops, sorry, slipped through the cracks. Updated

atris:
@eolivelli Updated,  please see. On another note, I see that TestEntryLog has just started failing again (the PR checks passed on my last commit 2 hours ago). Did something change?

eolivelli:
LGTM

eolivelli:
@sijie do you have comments?
@merlimat you may be interested in this fix

eolivelli:
@atris please rebase to latest master, we have fixed a problem with git checkout

eolivelli:
@atris please rebase to latest master, we have updated the github action "checkout"

atris:
@eolivelli DOne

eolivelli:
Hopefully we can commit this patch

Pinging @sijie @ivankelly @jiazhai

atris:
Can we merge now?

karanmehta93:
Thank you @atris for fixing this.

eolivelli:
Merged to master branch, thank you @atris 
sorry that it took so much time.

I hope that our CI will be in better shape soon

PR #2278, opened 2020-03-05
Add forkCount option to parallelize build and tests

Ghatage:
### Motivation
Current code forks a new JVM per module. (bookkeeper-server, bookkeeper-proto etc)
This means one fork per module for build and testing, no parallelism within the module where majority of the time goes.
We need parallelism within a module during the test execution so we can have the builds complete faster and have the artifacts shipped out quicker.

### Changes
We use the maven surefire plugin but don't define the `forkCount` and hence set it to default of 1.
This means it executes each module with one thread.
This change sets `forkCount` to 5, enabling parallelism in testing and drastically reducing total turnaround time. (by about 2/3rds!)

*Total build+test time without this change*
```
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  01:06 h
[INFO] Finished at: 2020-03-05T02:01:29-08:00
[INFO] ------------------------------------------------------------------------
```
*Total build+test time with this change*
```
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  18:23 min
[INFO] Finished at: 2020-03-05T02:38:22-08:00
[INFO] ------------------------------------------------------------------------
```

### Things to watch
Added parallelism may cause some flappers but with much trial and error I have come to the number `5`. The flappers are usually only from conflict in obtaining the same port number.
If needed, we can increase the retryCount, but as of now I consistently don't see any flappers at a `forkCount` of 5

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -195,6 +195,7 @@
     <protobuf-maven-plugin.version>0.5.0</protobuf-maven-plugin.version>
     <puppycrawl.checkstyle.version>6.19</puppycrawl.checkstyle.version>
     <spotbugs-maven-plugin.version>3.1.8</spotbugs-maven-plugin.version>
+    <forkCount.variable>5</forkCount.variable>
   </properties>
 
   <!-- dependency definitions -->
@@ -838,7 +839,7 @@
         <configuration>
           <argLine>-Xmx2G -Djava.net.preferIPv4Stack=true -Dio.netty.leakDetection.level=paranoid</argLine>
           <redirectTestOutputToFile>${redirectTestOutputToFile}</redirectTestOutputToFile>
-          <forkCount>5</forkCount>
+          <forkCount>${forkCount.variable}</forkCount>
           <reuseForks>false</reuseForks>
           <forkedProcessTimeoutInSeconds>1800</forkedProcessTimeoutInSeconds>
           <rerunFailingTestsCount>${testRetryCount}</rerunFailingTestsCount>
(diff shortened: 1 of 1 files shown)


Ghatage:
@eolivelli Looks like we've halved the total execution time required to run most CI tests on Github Actions! (it's even better locally, on my MBP its reduced by 2/3rds)
Can you please re-run the flapper? (Since this is Github actions only committers can re-execute the tests)

Ghatage:
Hi @eolivelli,

Do you think that is necessary? Only asking as [Pulsar doesn't do that either.](https://github.com/apache/pulsar/blob/ca7bb998436a32f7f919567d4b62d42b994363f8/pom.xml#L1238)

Ghatage:
Will do. We can reduce it as you suggest if it causes too much ruckus.

eolivelli:
I have restarted bookie tests and integration tests.
Is this patch affecting integration tests ?
I think we cannot run more than one IT at a time, it will use many resources

Ghatage:
Fair enough @eolivelli. I have moved it to 1. 
Please note that folks will be able to take benefit of this addition by changing `<forkCount.variable>` locally in their pom.xml
Hopefully his should help the Github Actions Integration tests.

eolivelli:
I have restarted the failing Bookie Tests

Issue #2279, opened 2020-03-05
NullPointerException when execute `bin/bookkeeper shell readjournal`


aloyszhang:
**BUG REPORT**

***Describe the bug***
Run
`./bin/bookkeeper shell readjournal -msg /data1/bookkeeper/journal/current/1709f92d291.txn   -dir /data1/bookkeeper/journal`

It throws exception
```
Exception in thread "main" java.lang.NullPointerException
        at org.apache.bookkeeper.tools.cli.commands.bookie.ReadJournalCommand.handler(ReadJournalCommand.java:127)
        at org.apache.bookkeeper.tools.cli.commands.bookie.ReadJournalCommand.apply(ReadJournalCommand.java:105)
        at org.apache.bookkeeper.bookie.BookieShell$ReadJournalCmd.runCmd(BookieShell.java:1046)
        at org.apache.bookkeeper.bookie.BookieShell$MyCommand.runCmd(BookieShell.java:223)
        at org.apache.bookkeeper.bookie.BookieShell.run(BookieShell.java:1976)
        at org.apache.bookkeeper.bookie.BookieShell.main(BookieShell.java:2067)
```
I found when  `ReadJournalCmd` initialized,  the "dir " `Option`  added with `hasArg` flag is false. So, we can't get the option's value of ”dir“ from shell command parameters.
Here is the related code
   ```
ReadJournalCmd() {
            super(CMD_READJOURNAL);
            rjOpts.addOption("dir", false, "Journal directory (needed if more than one journal configured)");
            rjOpts.addOption("m", "msg", false, "Print message body");
        }
```

```
ReadJournalCommand.ReadJournalFlags flags = new ReadJournalCommand.ReadJournalFlags().msg(printMsg)
                                                            .fileName(filename).journalId(journalId)
                                                            .dir(cmdLine.getOptionValue("dir"));
```

I think set the `hasArg`  to true will resolve this problem.

***To Reproduce***
Just run `bin/bookkeeper shell readjournal `

sijie:
This is fixed by #2280

PR #2280, opened 2020-03-05
#2279 fix NPE when exec readjournal command

aloyszhang:
This PR fix NullPointerException when exec readjournal command
See issue #2279

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
@@ -1013,7 +1013,7 @@ public class BookieShell implements Tool {
 
         ReadJournalCmd() {
             super(CMD_READJOURNAL);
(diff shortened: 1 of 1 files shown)


aloyszhang:
Parameter `-dir` must be specified when bookie has more than one directories for `journalDirectories` while not need when only one journal directory.

aloyszhang:
cc @eolivelli

eolivelli:
Yes but if you only have one directory the parameter is not required.
With this patch you are making it always required.

aloyszhang:
As I understand, if `hasArg` of an `Option`  set to true,  it means that an argument is required after this option other than this `Option`  is must be specified.
We still can run the command without the parameter `-dir` like this 
 ```
./bin/bookkeeper shell readjournal /data1/bookkeeper/journal/current/170ab39159f.txn 
```
and this command runs well.

aloyszhang:
cc @eolivelli please have a look.

eolivelli:
I am sorry.
I clicked on 'close and comment' instead of 'comment'

Issue #2273, opened 2020-02-24
Bookie does not try to download ledger from another bookie


hamadodene:
**BUG REPORT**

***Describe the bug***

We have a cluster of Bookkeeper with 3 servers. 
For example we can considered (bk1,bk2 and bk3). We writing only inside bk1 and bk2 , and bk3 read from bk2.  
When bk2 come down  bk3 continue to search reading from bk2. 
We expected if bk2 come down, bk3 should try to read from bk1 and not continue to  try with bk2.

```20-02-24-12-35-15       While readLastConfirmed ledger: 19734 did not hear success responses from all quorums
20-02-24-12-35-21       While readLastConfirmed ledger: 19734 did not hear success responses from all quorums
20-02-24-12-35-27       While readLastConfirmed ledger: 19734 did not hear success responses from all quorums
20-02-24-12-35-33       While readLastConfirmed ledger: 19734 did not hear success responses from all quorums
20-02-24-12-35-39       While readLastConfirmed ledger: 19734 did not hear success responses from all quorums
```
```
20-02-24-12-33-30       Could not connect to bookie: [id: 0xa6f88d56, L:/ip-bk3:37224]/bk2:3181, current state CONNECTING :
20-02-24-12-33-30       io.netty.channel.AbstractChannel$AnnotatedConnectException: finishConnect(..) failed: No route to host: bk2/ip-address:3181
io.netty.channel.AbstractChannel$AnnotatedConnectException: finishConnect(..) failed: No route to host: bk2/ip-address:3181
        at io.netty.channel.unix.Errors.throwConnectException(Errors.java:112)
        at io.netty.channel.unix.Socket.finishConnect(Socket.java:269)
        at io.netty.channel.epoll.AbstractEpollChannel$AbstractEpollUnsafe.doFinishConnect(AbstractEpollChannel.java:670)
        at io.netty.channel.epoll.AbstractEpollChannel$AbstractEpollUnsafe.finishConnect(AbstractEpollChannel.java:647)
        at io.netty.channel.epoll.AbstractEpollChannel$AbstractEpollUnsafe.epollOutReady(AbstractEpollChannel.java:527)
        at io.netty.channel.epoll.EpollEventLoop.processReady(EpollEventLoop.java:422)
        at io.netty.channel.epoll.EpollEventLoop.run(EpollEventLoop.java:333)
        at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:906)
[...]

PR #2281, opened 2020-03-07
[bookie-ledger-recovery] Fix bookie recovery stuck even with enough ack-quorum response

rdhabalia:
### Motivation
As discussed at https://github.com/apache/pulsar/issues/6505 

Bk-client was not able to recover ledger which has 2 write/ack quorum and one of the bookie went down and recovery was kept failing and bookkeeper client was not able to recover the ledger.

**BK-Client log**

```
20:44:43.721 [BookKeeperClientWorker-OrderedExecutor-1-0] ERROR org.apache.bookkeeper.client.ReadLastConfirmedOp - While readLastConfirmed ledger: 1234567 did not hear success responses from all quorums
20:44:43.721 [bookkeeper-io-12-27] ERROR org.apache.bookkeeper.proto.PerChannelBookieClient - Could not connect to bookie: [id: 0xb8b97441, L:/1.1.1.1:1234]/1.1.1.2:3181, current s
tate CONNECTING : 
io.netty.channel.AbstractChannel$AnnotatedConnectException: finishConnect(..) failed: No route to host: /1.1.1.2:3181
        at io.netty.channel.unix.Errors.throwConnectException(Errors.java:112) ~[netty-all-4.1.32.Final.jar:4.1.32.Final]
        at io.netty.channel.unix.Socket.finishConnect(Socket.java:269) ~[netty-all-4.1.32.Final.jar:4.1.32.Final]
        at io.netty.channel.epoll.AbstractEpollChannel$AbstractEpollUnsafe.doFinishConnect(AbstractEpollChannel.java:665) [netty-transport-native-epoll-4.1.31.Final.jar:4.1.31.Final]
        at io.netty.channel.epoll.AbstractEpollChannel$AbstractEpollUnsafe.finishConnect(AbstractEpollChannel.java:642) [netty-transport-native-epoll-4.1.31.Final.jar:4.1.31.Final]
        at io.netty.channel.epoll.AbstractEpollChannel$AbstractEpollUnsafe.epollOutReady(AbstractEpollChannel.java:522) [netty-transport-native-epoll-4.1.31.Final.jar:4.1.31.Final]
        at io.netty.channel.epoll.EpollEventLoop.processReady(EpollEventLoop.java:423) [netty-transport-native-epoll-4.1.31.Final.jar:4.1.31.Final]
        at io.netty.channel.epoll.EpollEventLoop.run(EpollEventLoop.java:330) [netty-transport-native-epoll-4.1.31.Final.jar:4.1.31.Final]
        at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:897) [netty-common-4.1.31.Final.jar:4.1.31.Final]
        at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30) [netty-common-4.1.31.Final.jar:4.1.31.Final]
        at java.lang.Thread.run(Thread.java:834) [?:?]
Caused by: java.net.ConnectException: finishConnect(..) failed: No route to host
```

**Ledger metadata**

```
BookieMetadataFormatVersion 2
quorumSize: 2
ensembleSize: 2
length: 0
lastEntryId: -1
state: IN_RECOVERY
segment {
  ensembleMember: "1.1.1.1:3181"
  ensembleMember: "1.1.1.2:3181"
  firstEntryId: 0
}
digestType: CRC32
```

**Root cause:**
Bookie should be able to recover ledger once it receives the response from total N (`(Qw - Qa)+1`) bookies. But it was waiting for a successful response from both quorums.
Reference: https://bookkeeper.apache.org/docs/4.5.0/development/protocol/

### Modification
Bookie should be able to recover ledger once it receives the response from total N (`(Qw - Qa)+1`) bookies.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadLastConfirmedOp.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadLastConfirmedOp.java
@@ -38,6 +38,7 @@ class ReadLastConfirmedOp implements ReadEntryCallback {
     LedgerHandle lh;
     BookieClient bookieClient;
     int numResponsesPending;
+    int numSuccessfulResponse;
     RecoveryData maxRecoveredData;
     volatile boolean completed = false;
     int lastSeenError = BKException.Code.ReadException;
@@ -59,6 +60,7 @@ class ReadLastConfirmedOp implements ReadEntryCallback {
         this.bookieClient = bookieClient;
         this.maxRecoveredData = new RecoveryData(LedgerHandle.INVALID_ENTRY_ID, 0);
         this.lh = lh;
+        this.numSuccessfulResponse = 0;
         this.numResponsesPending = lh.getLedgerMetadata().getEnsembleSize();
         this.coverageSet = lh.distributionSchedule.getCoverageSet();
         this.currentEnsemble = ensemble;
@@ -99,6 +101,7 @@ class ReadLastConfirmedOp implements ReadEntryCallback {
                     maxRecoveredData = recoveryData;
                 }
                 heardValidResponse = true;
+                numSuccessfulResponse++;
             } catch (BKDigestMatchException e) {
                 // Too bad, this bookie didn't give us a valid answer, we
                 // still might be able to recover though so continue
@@ -137,8 +140,15 @@ class ReadLastConfirmedOp implements ReadEntryCallback {
         }
 
         if (numResponsesPending == 0 && !completed) {
+            int totalExepctedResponse = lh.getLedgerMetadata().getWriteQuorumSize()
+                    - lh.getLedgerMetadata().getAckQuorumSize() + 1;
+            if (numSuccessfulResponse >= totalExepctedResponse) {
+                cb.readLastConfirmedDataComplete(BKException.Code.OK, maxRecoveredData);
+                return;
+            }
             // Have got all responses back but was still not enough, just fail the operation
-            LOG.error("While readLastConfirmed ledger: {} did not hear success responses from all quorums", ledgerId);
+            LOG.error("While readLastConfirmed ledger: {} did not hear success responses from all quorums {}", ledgerId,
+                    lastSeenError);
             cb.readLastConfirmedDataComplete(lastSeenError, maxRecoveredData);
         }
 
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/ParallelLedgerRecoveryTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/ParallelLedgerRecoveryTest.java
@@ -25,6 +25,7 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 
 import io.netty.buffer.ByteBuf;
+import io.netty.buffer.ByteBufAllocator;
 import io.netty.buffer.Unpooled;
 
 import java.io.IOException;
@@ -66,6 +67,7 @@ import org.apache.bookkeeper.test.BookKeeperClusterTestCase;
 import org.apache.bookkeeper.util.ByteBufList;
 import org.apache.bookkeeper.versioning.Version;
 import org.apache.bookkeeper.versioning.Versioned;
+import org.apache.commons.lang3.mutable.MutableInt;
 import org.apache.zookeeper.AsyncCallback.VoidCallback;
 import org.apache.zookeeper.KeeperException;
 import org.junit.After;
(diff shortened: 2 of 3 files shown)


eolivelli:
This is related to #2273

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/ParallelLedgerRecoveryTest.java
```diff
@@ -693,4 +695,55 @@ public void readLastConfirmedDataComplete(int rc, DigestManager.RecoveryData dat
...
+                        responseLatch.countDown();
+                    }
+                });
+
+        // complete first successful write.
+        byte[] lac = new byte[DigestManager.METADATA_LENGTH * 2];
+        try {
+            readLCOp.readEntryComplete(BKException.Code.OK, 0, 0, Unpooled.wrappedBuffer(lac, 0, lac.length), 0);
+        } catch (Exception e) {
+            // Ok
```
eolivelli: What is this expected error?
Can we add an assertion about this exception?

eolivelli:
Okay.
We are waiting for  WQ - AQ +1 so:
- WQ=2 AQ=2 wait for 1, it is okay as all bookies have every entry
- WQ=2  AQ=1 wait for 2, it is okay as we have to check every bookie

So your fix work well according to the protocol .

Please add test cases

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/ParallelLedgerRecoveryTest.java
rdhabalia: actually it was failing due to digest verification because of mock data but I have fixed it after spending some time understanding it correctly.

rdhabalia:
@eolivelli @diegosalvi  I have spent some time to cover most of the usecases with multiple groups of tests.

sijie:
@eolivelli please take a look

eolivelli:
I will merge today if @diegosalvi  do not have other comments

eolivelli:
@rdhabalia I have merged this patch to master branch.

Thank you !

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadLastConfirmedOp.java
```diff
@@ -137,8 +140,15 @@ public synchronized void readEntryComplete(final int rc, final long ledgerId, fi
         }
 
         if (numResponsesPending == 0 && !completed) {
```
ivankelly, line 142: @eolivelli @sijie 

 This fix is the wrong fix. This problem is the same problem as #2273 , and we've also seen internally. The root cause is that quorum coverage is broken. 5e399df broke it. 5e399df was a change in the write path, so why it touched the quorum coverage in the read path is a mystery. dcdd1e88 supplemented the break which and added a test case. The test case tests exactly the wrong behaviour Pre-5e399df it worked correctly. 

It seems there is confusion as to what the quorum coverage is supposed to do.

The quorum coverage set is there to check if we have received a valid response from enough nodes that there cannot exist an entry which has been acknowledged to the client which has not touched at least one of these nodes. In other words, there cannot exist an entry which formed an ack quorum without touching one of the nodes we have talked to.

The quorum coverage set should only have two states for a node, known or unknown. The known state means that we know definitely than an entry does or does not exist on that node. So, an OK, NoSuchLedger and NoSuchEntry response will move the node into a known state. Anything else is an unknown state. No response from node is unknown state. Node had an IOException is an unknown state.

Take the test supplied with dcdd1e88:
```
        RoundRobinDistributionSchedule schedule = new RoundRobinDistributionSchedule(
            5, 3, 5);
        DistributionSchedule.QuorumCoverageSet covSet = schedule.getCoverageSet();
        covSet.addBookie(0, BKException.Code.NoSuchLedgerExistsException);
        covSet.addBookie(1, BKException.Code.NoSuchEntryException);
        covSet.addBookie(2, BKException.Code.NoSuchLedgerExistsException);
        covSet.addBookie(3, BKException.Code.UNINITIALIZED);
        covSet.addBookie(4, BKException.Code.UNINITIALIZED);
        assertFalse(covSet.checkCovered());
```
For clarify, w is 5, a is 3, e is 5. RoundRobinDistributionSchedule parameters are messed up.

This test is simply wrong. The coverage set is covered. There is no way an entry can have ack quorum in any write quorum. There is no set of 3 bookies in the unknown state.

Anyhow, I have a fix going through CI internally. I'll push up a PR on monday.

rdhabalia: yes makes sense.  provided test in https://github.com/apache/bookkeeper/commit/dcdd1e887ef01f7515053f7eb06f8479de8700ff  is not correct and it should cover the coverage and if that covers the coverage then we won't need this fix.

PR #2287, opened 2020-03-13
Add Journal PageCache flush interval control when journalSyncData is disabled

hangc0276:
### Motivation
There was a weird behavior observed: when journalSyncData is disabled, it will trigger callbacks when the data is flushed to filesystem(PageCache) but still enqueue a fsync request to force write queue to fsync PageCache data to disk frequently. Though the fsync thread dequeue fsync request and execute fsync operation asynchronously, it still cause disk io util increase when fsync frequency high enough. when disk io util increase, especially reach 100%, it will block new journal file creation and lead to journal sync time increases to mutiple seconds.

### Changes
when journalSyncData is disabled, we introduce `journalPageCacheFlushIntervalMSec` to control journal PageCache flush frequency to reduce disk io util, default value is 1000ms

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
@@ -1062,11 +1062,12 @@ public class Journal extends BookieCriticalThread implements CheckpointSource {
                             // 3. if journalSyncData disabled and shouldRolloverJournal is false, we can use
                             //   journalPageCacheFlushIntervalMSec to control sync frequency, preventing disk
                             //   synchronize frequently, which will increase disk io util.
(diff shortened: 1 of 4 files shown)


hangc0276:
ok, i will add test cases for this change.

hangc0276:
I add a test case, please take a look @jiazhai @eolivelli @sijie @merlimat 

when an entry add into Journal and after journalPageCacheFlushInterval passed, no new entry added into journal, the last added entry will not be flushed to disk from PageCache.

In the test case `BookieJournalPageCacheFlushTest#testAckBeforeUnSyncPageCacheFlush`, I just sleep twice journalPageCacheFlushInterval, and add another entry into journal, then check whether foreWriteRequest add into forceWriteRequestQueue.

PR #2317, opened 2020-04-21
Force to use Python grpcio less then 1.26.0

eolivelli:
- it looks like a newer version of grpcio has been published and we were using open version range
- by forcing a closed ended range of version we workaround the error in integration tests


eolivelli:
This is a fix for the integration tests,
if integration tests pass we are unblocking the stack of pending pull requests that we cannot accept due to problems with integration tests

eolivelli:
@mino181295 PTAL

mino181295:
Nice! LGTM 👍🏻

eolivelli:
Integration tests passed ! Hurray !

Issue #2285, opened 2020-03-10
Spammy log when one bookie of ensemble is down


hamadodene:
When one bookie of the ensemble is down, the others bookies client  spam in the log ***org.apache.bookkeeper.proto.PerChannelBookieClient*** this exception: 

```
20-03-10-10-42-01       Could not connect to bookie: [id: 0x823ef32e, L:/10.162.1.12:42330]/es-dev1.sviluppo.dna:3181, current state CONNECTING :
20-03-10-10-42-01       io.netty.channel.AbstractChannel$AnnotatedConnectException: syscall:getsockopt(..) failed: Connection refused: es-dev1.sviluppo.dna/10.162.1.11:3181
io.netty.channel.AbstractChannel$AnnotatedConnectException: syscall:getsockopt(..) failed: Connection refused: es-dev1.sviluppo.dna/10.162.1.11:3181
        at io.netty.channel.unix.Socket.finishConnect(..)(Unknown Source)
Caused by: io.netty.channel.unix.Errors$NativeConnectException: syscall:getsockopt(..) failed: Connection refused
        ... 1 more 

```
In our case, in 8m it produced 1,5gb log.
For riproduced it, i simple shutdown a bookie and checked **org.apache.bookkeeper.proto.PerChannelBookieClient**

sijie:
I think we can reduce the logging level to avoid the noisiness.

PR #2319, opened 2020-04-22
Issue 2285: Spammy log when one bookie of ensemble is down

mino181295:
### Motivation

"Could not connect to bookie" spammed too much times

### Changes

Debug instead of trace logging level

(Previous issue that added this log #1776)

Master issue: #2285


bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
```diff
@@ -2447,12 +2447,16 @@ public void operationComplete(ChannelFuture future) throws Exception {
                     Throwable cause = future.cause();
                     if (cause instanceof UnknownHostException || cause instanceof NativeIoException) {
                         // Don't log stack trace for common errors
-                        LOG.warn("Could not connect to bookie: {}/{}, current state {} : {}",
```
eolivelli, line 2450: If you remove this line how do you know that the client is not able to connect to the bookie?

PR #2321, opened 2020-04-29
[bookie-shell] Add cmd to change bookie-address into ledger metadata

rdhabalia:
### Motivation
In a large bookie cluster environment, we frequently require a utility to change bookie-ip with a different bookie-ip in ledger's metadata in different scenarios such as:
1. Host Re-IP: requires updating ledger metadata and reuse the host without losing/copying data
2. Backup-restore usecase while doing cluster level data migration.
Therefore, we frequently need a tool to update bookie-ip with a new bookie-ip in existing ledgers' ensemble metadata.

### Modification
Add CLI command to update bookie-id in ledger metadata.


rdhabalia:
> But can you please add a documentation page about how to perform this operation safely?

Sure. Is [shell.yaml](https://github.com/apache/bookkeeper/blob/master/site/_data/cli/shell.yaml) the right place to document? This CLI already validates if src-bookie is down before initiating the process and for usecase such as: host-reip or data-migration to other bookie require updating cookies. So, If `shell.yaml` is the right place then I can add the document.

eolivelli:
rerun failure tests

eolivelli:
@rdhabalia I don't know why GitHub actions doesn't want to retest this patch.
Can you please rebase onto current master ? this will trigger CI

rdhabalia:
@eolivelli can you merge this change

Issue #2294, opened 2020-03-20
Full build download distributedlog from apache snapshots repository


eolivelli:
**BUG REPORT**

***Describe the bug***

If you fresh clone bookkeeper from github, with an empty Maven repository, Maven tries to download distributedlog-core from apache snapshots and it is not using the binaries built during local compilation.

This can lead to weird behaviour and non reproducible builds

```
[INFO] -------------------< org.apache.bookkeeper:statelib >-------------------
[INFO] Building Apache BookKeeper :: Stream Storage :: State Library 4.11.0-SNAPSHOT [26/78]
[INFO] --------------------------------[ jar ]---------------------------------
Downloading from apache.snapshots: https://repository.apache.org/snapshots/org/apache/distributedlog/distributedlog-core/4.11.0-SNAPSHOT/maven-metadata.xml
Downloaded from apache.snapshots: https://repository.apache.org/snapshots/org/apache/distributedlog/distributedlog-core/4.11.0-SNAPSHOT/maven-metadata.xml (1.2 kB at 499 B/s)
Downloading from apache.snapshots: https://repository.apache.org/snapshots/org/apache/distributedlog/distributedlog-core/4.11.0-SNAPSHOT/distributedlog-core-4.11.0-20200307.150610-126.pom
Downloaded from apache.snapshots: https://repository.apache.org/snapshots/org/apache/distributedlog/distributedlog-core/4.11.0-SNAPSHOT/distributedlog-core-4.11.0-20200307.150610-126.pom (12 kB at 9.0 kB/s)
Downloading from apache.snapshots: https://repository.apache.org/snapshots/org/apache/distributedlog/distributedlog/4.11.0-SNAPSHOT/maven-metadata.xml
Downloaded from apache.snapshots: https://repository.apache.org/snapshots/org/apache/distributedlog/distributedlog/4.11.0-SNAPSHOT/maven-metadata.xml (622 B at 504 B/s)
Downloading from apache.snapshots: https://repository.apache.org/snapshots/org/apache/distributedlog/distributedlog/4.11.0-SNAPSHOT/distributedlog-4.11.0-20200307.144315-126.pom
Downloaded from apache.snapshots: https://repository.apache.org/snapshots/org/apache/distributedlog/distributedlog/4.11.0-SNAPSHOT/distributedlog-4.11.0-20200307.144315-126.pom (5.6 kB at 5.5 kB/s)
[...]

PR #2323, opened 2020-05-05
Do not download SNAPSHOT during GitHub actions checks

eolivelli:
### Motivation

Sometimes GitHub actions checks are stuck in downloading SNAPSHOTs, there is not need for it

### Changes

Add "-nsu" option to every Maven invokations and remove "-Dstream" that is now useless.
I am also adding "-B" that suppresses the "Progress" noise in logs

This fix relates to #2294


eolivelli:
@zymap can you please take a look ?

mino181295:
Ok, +1

dmercuriali:
+1

PR #2303, opened 2020-04-06
QuorumCoverage should only count unknown nodes

ivankelly:
Quorum coverage checks if we have heard from enough nodes to know that
there is no entry that can have been written to enough nodes that we
haven't heard from to have formed an ack quorum.

The coverage algorithm was correct pre-5e399df.

5e399df(BOOKKEEPER-759: Delay Ensemble Change & Disable Ensemble
Change) broke this, but it still seems to have worked because they had
a broken else statement at the end. Why a change which is 100% about
the write-path changed something in the read-path is a mystery.

dcdd1e(Small fix wrong nodesUninitialized count when checkCovered)
went on to fix the broken fix, so the whole thing ended up broke.

The change also modifies ReadLastConfirmedOp to make it testable.

PR #2333, opened 2020-05-17
QuorumCoverage should only count unknown nodes

eolivelli:
The original patch was contributed by @ivankelly in PR #2303, I have only fixed checkstyle and removed two tests that were wrong.

Quorum coverage checks if we have heard from enough nodes to know that
there is no entry that can have been written to enough nodes that we
haven't heard from to have formed an ack quorum.

The coverage algorithm was correct pre-5e399df.

5e399df(BOOKKEEPER-759: Delay Ensemble Change & Disable Ensemble
Change) broke this, but it still seems to have worked because they had
a broken else statement at the end. Why a change which is 100% about
the write-path changed something in the read-path is a mystery.

dcdd1e(Small fix wrong nodesUninitialized count when checkCovered)
went on to fix the broken fix, so the whole thing ended up broke.

The change also modifies ReadLastConfirmedOp to make it testable.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
@@ -1407,7 +1407,13 @@ public class LedgerHandle implements WriteHandle {
                 }
             };
 
-        new ReadLastConfirmedOp(this, clientCtx.getBookieClient(), getCurrentEnsemble(), innercb).initiate();
+        new ReadLastConfirmedOp(clientCtx.getBookieClient(),
+                                distributionSchedule,
+                                macManager,
+                                ledgerId,
+                                getCurrentEnsemble(),
+                                ledgerKey,
+                                innercb).initiate();
     }
 
     /**
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerRecoveryOp.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerRecoveryOp.java
@@ -93,7 +93,12 @@ class LedgerRecoveryOp implements ReadEntryListener, AddCallback {
     }
 
     public CompletableFuture<LedgerHandle> initiate() {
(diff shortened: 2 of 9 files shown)


eolivelli:
@sijie @rdhabalia @merlimat you already approved the original patch

Please any of you take a look and confirm your approval
my addition is
https://github.com/apache/bookkeeper/pull/2333/commits/55607a874738b02457d309291b87713a077b3e33

eolivelli:
It looks like I still have a test to fix:
```
[ERROR] testZoneawarePlacementPolicyCheck(org.apache.bookkeeper.replication.AuditorPlacementPolicyCheckTest)  Time elapsed: 61.135 s  <<< FAILURE!
java.lang.AssertionError: placementPolicyCheck should have executed
	at org.apache.bookkeeper.replication.AuditorPlacementPolicyCheckTest.startAuditorAndWaitForPlacementPolicyCheck(AuditorPlacementPolicyCheckTest.java:684)
	at org.apache.bookkeeper.replication.AuditorPlacementPolicyCheckTest.testZoneawarePlacementPolicyCheck(AuditorPlacementPolicyCheckTest.java:628)
```

eolivelli:
@sijie did all CI tests pass this time ?
last time I saw that testZoneawarePlacementPolicyCheck did not pass.
I wouldn't want that we have introduced a new annoying  flaky test

sijie:
@eolivelli all CI tests passed.

lamber-ken:
hi @eolivelli, I have a question that why remove following code snippet? thanks

image

eolivelli:
That test was wrong.
It used ensemble size = 4 and write quorum size = 1.
Then it shutdown a bookie.
If it shuts down the only bookie that contains the entry the test will fail

lamber-ken:
> That test was wrong.
> It used ensemble size = 4 and write quorum size = 1.
> Then it shutdown a bookie.
> If it shuts down the only bookie that contains the entry the test will fail

right, 💯

rdhabalia:
just fyi: fix: https://github.com/apache/bookkeeper/pull/2333

PR #2346, opened 2020-05-22
 [CLEAN] Replace com.google.common.base.Optional with java.util.Optional

mino181295:
### Motivation

`com.google.common.base.Optional` and `java.util.Optional` exist in the code base at the same time.

### Changes
Replace `com.google.common.base.Optional` with `java.util.Optional`.

### DIFF
--- a/site/docs/latest/api/distributedlog-api.md
+++ b/site/docs/latest/api/distributedlog-api.md
@@ -222,7 +222,7 @@ logConf.setRetentionPeriodHours(12);
 DistributedLogManager logManager = namespace.openLog(
         "test-log",
         Optional.of(logConf),
-        Optional.absent()
+        Optional.empty()
 );
 ```
 
--- a/stream/distributedlog/core/src/main/java/org/apache/distributedlog/BookKeeperClientBuilder.java
+++ b/stream/distributedlog/core/src/main/java/org/apache/distributedlog/BookKeeperClientBuilder.java
@@ -20,9 +20,9 @@ package org.apache.distributedlog;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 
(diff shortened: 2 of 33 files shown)


eolivelli:
@mino181295 awesome work.

One question:
did you check if you are changing public API ?
org.apache.bookkeeper.client
org.apache.bookkeeper.client.api
and DLog API

site/docs/latest/api/distributedlog-api.md
```diff
@@ -222,7 +222,7 @@ logConf.setRetentionPeriodHours(12);
 DistributedLogManager logManager = namespace.openLog(
         "test-log",
         Optional.of(logConf),
-        Optional.absent()
+        Optional.empty()
```
eolivelli, line 225: this is documentation but it looks like an API change

stream/distributedlog/core/src/main/java/org/apache/distributedlog/BookKeeperClientBuilder.java
```diff
@@ -59,7 +59,7 @@ public static BookKeeperClientBuilder newBuilder() {
     // request timer
     private HashedWheelTimer requestTimer = null;
     // feature provider
-    private Optional<FeatureProvider> featureProvider = Optional.absent();
+    private Optional<FeatureProvider> featureProvider = Optional.empty();
```
eolivelli, line 62: this change looks like an API change

eolivelli:
@diegosalvi @sijie PTAL

mino181295:
To answer your questions:
- The guava Optional was marked as @deprecated since 0.5 in favor of 
`
loadStreamConf(java.util.Optional<DistributedLogConfiguration> streamConfiguration)
`
- The `Namespace` class is marked as `@Evolving` and I think this change it's not a big deal to fix

cc @sijie

Issue #2341, opened 2020-05-21
Add a way to configure bookkeeper parameters


coolbeevip:
**FEATURE REQUEST**

1. Please describe the feature you are requesting.

In docker, the environment variable parameters are used applying-config-from-env.py to modify bk_server.conf. This inconsistency can be confusing for non-container deployments

2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have).
   Are you currently using any workarounds to address this issue?

I think you can add a new parse method parseEnvironment and this method is executed after parseCommandLine. Environment variable parameters take precedence over command line parameters

3. Provide any additional detail on your proposed use case for this feature.

 code snippet is shown below, any suggestions？

Main.java
```
        // 0. parse command line
        try {
            conf = parseCommandLine(args);
        } catch (IllegalArgumentException iae) {
            return ExitCode.INVALID_CONF;
        }

        try {
            conf = parseEnvironment(conf);
        } catch (IllegalArgumentException iae) {
            return ExitCode.INVALID_CONF;
        }
```

Read the variable with the prefix BK_ from the environment variable, and overwrite the variable value read from the command line through reflection

```
    private static ServerConfiguration parseEnvironment(ServerConfiguration conf) {
        System.getenv().entrySet().stream().filter(entry -> entry.getKey().startsWith("BK_"))
            .forEach(entry -> {
                String name = entry.getKey().substring("BK_".length());
                String value = entry.getValue();
                Arrays.stream(conf.getClass().getDeclaredMethods())
                    .filter(m -> m.getName().startsWith("set")).forEach(m -> {
                    try {
                        if (m.getName().toLowerCase().equals("set" + name.toLowerCase())
                            && m.getParameterCount() == 1) {
                            if (m.getParameterTypes()[0] == boolean.class) {
                                m.invoke(conf, Boolean.valueOf(value).booleanValue());
                            } else if (m.getParameterTypes()[0] == int.class) {
                                m.invoke(conf, Integer.valueOf(value).intValue());
                            } else if (m.getParameterTypes()[0] == long.class) {
                                m.invoke(conf, Long.valueOf(value).longValue());
                            } else if (m.getParameterTypes()[0] == double.class) {
                                m.invoke(conf, Double.valueOf(value).doubleValue());
                            } else if (m.getParameterTypes()[0] == float.class) {
                                m.invoke(conf, Float.valueOf(value).floatValue());
                            } else if (m.getParameterTypes()[0] == String.class) {
[...]

PR #2347, opened 2020-05-24
ISSUE-2341 Load bookkeeper parameters from environment variables

coolbeevip:
Load bookkeeper parameters from environment variables prefixed with BK_

### Motivation

Add the way to configure bookkeeper through environment variables to simplify deployment

### Changes

Load bookkeeper parameters from environment variables with the prefix BK_ before load bk_server.conf

Master Issue: #2341

### DIFF
--- a/bookkeeper-server/pom.xml
+++ b/bookkeeper-server/pom.xml
@@ -170,11 +170,6 @@
       <version>${project.parent.version}</version>
       <scope>test</scope>
     </dependency>
-    <dependency>
-      <groupId>com.github.stefanbirkner</groupId>
-      <artifactId>system-rules</artifactId>
-      <scope>test</scope>
-    </dependency>
   </dependencies>
   <build>
     <plugins>
--- a/bookkeeper-server/pom.xml
+++ b/bookkeeper-server/pom.xml
@@ -170,6 +170,11 @@
       <version>${project.parent.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.github.stefanbirkner</groupId>
+      <artifactId>system-rules</artifactId>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
   <build>
     <plugins>
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
@@ -236,7 +236,7 @@ public abstract class AbstractConfiguration<T extends AbstractConfiguration>
                 String name = entry.getKey().substring(ENV_PREFIX.length());
                 String value = entry.getValue();
                 log.info("load environment {}={}", name, value);
-                setProperty(name,value);
+                setProperty(name, value);
             });
     }
 
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
@@ -72,6 +72,9 @@ public abstract class AbstractConfiguration<T extends AbstractConfiguration>
         DEFAULT_LOADER = loader;
     }
 
+    // Environment Parameters Prefix
+    protected static final String ENV_PREFIX = "BK_";
+
     // Zookeeper Parameters
     protected static final String ZK_TIMEOUT = "zkTimeout";
     protected static final String ZK_SERVERS = "zkServers";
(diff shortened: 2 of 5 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/server/Main.java
```diff
@@ -125,6 +135,8 @@ private static ServerConfiguration parseArgs(String[] args)
 
             ServerConfiguration conf = new ServerConfiguration();
 
+            loadConfEnv(conf);
+
             if (cmdLine.hasOption('c')) {
                 String confFile = cmdLine.getOptionValue("c");
                 loadConfFile(conf, confFile);
```
mino181295, line 142: Do you expect `loadConfFile` overrides `loadConfFile` if the same property is defined?

bookkeeper-server/src/main/java/org/apache/bookkeeper/server/Main.java
```diff
@@ -112,6 +112,16 @@ private static void loadConfFile(ServerConfiguration conf, String confFile)
         log.info("Using configuration file {}", confFile);
     }
 
+    private static void loadConfEnv(ServerConfiguration conf) throws IllegalArgumentException {
```
mino181295, line 115: What about this in `LocalBookKeeper` ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/server/Main.java
coolbeevip: Thank you for reminding me, my intention is that the environment variable has a higher priority than the configuration file.

I will modify it to call LoadConfEnv after loadConfFile

eolivelli: Please add a command line option.

Otherwise with this change we are introducing unexpected behaviour and possibly a security issue.

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
```diff
@@ -224,6 +227,19 @@ public void loadConf(URL confURL) throws ConfigurationException {
         }
     }
 
+    /**
+     * Load configurations from environment variables prefixed with BK_.
+     */
+    public void loadEnv() {
```
eolivelli, line 233: I am not sure this is the best place to put this code.
This is the configuration for the client as well.
I suggest to put this block into Main.java? This way it is specific to the bookie bootstrap procedure.

coolbeevip: I changed the loadConfEnv method of Main.java to the following

```
    private static void loadConfEnv(ServerConfiguration conf) throws IllegalArgumentException {
        try {
            System.getenv().entrySet().stream().filter(entry -> entry.getKey().startsWith(ENV_PREFIX))
                .forEach(entry -> {
                    String name = entry.getKey().substring(ENV_PREFIX.length());
                    String value = entry.getValue();
                    log.info("load environment {}={}", name, value);
                    conf.setProperty(name, value);
                });
            conf.validate();
        } catch (ConfigurationException e) {
            log.error("Malformed configuration file: {}", e);
            throw new IllegalArgumentException();
        }
    }
```

Do you have any suggestions about defining the location of constant ENV_PREFIX?

```
    // Environment Parameters Prefix
    protected static final String ENV_PREFIX = "BK_";
```

coolbeevip: > I am not sure this is the best place to put this code.
> This is the configuration for the client as well.
> I suggest to put this block into Main.java? This way it is specific to the bookie bootstrap procedure.

If this block is put into Main.java, it cannot be reused with LocalBookKeeper :(

bookkeeper-server/pom.xml
```diff
@@ -170,6 +170,11 @@
       <version>${project.parent.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.github.stefanbirkner</groupId>
+      <artifactId>system-rules</artifactId>
+      <scope>test</scope>
```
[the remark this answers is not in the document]
coolbeevip: done

bookkeeper-server/src/main/java/org/apache/bookkeeper/server/Main.java
coolbeevip: > Please add a command line option.
> 
> Otherwise with this change we are introducing unexpected behaviour and possibly a security issue.

Use the command-line option `env`, What do you think?

michaeljmarshall:
If this gets any more attention in the future, we should make sure that the PR also adds documentation for this feature.

hsaputra:
This PR has been open for a while with requests for changes that has not been completed.
Closing this PR for now, but please do reopen it when the issues and concerns have been addressed. Thanks.

PR #2353, opened 2020-06-03
Prototype patch about non resending writes in case of "disable ensemble change"

eolivelli:


eolivelli:
this is only a prototype,
original patch contributed by @jvrao 

I have created this Pr in order to run tests on CI

This patch is not meant to be committed

eolivelli:
@jvrao all tests passed with your patch.
So it looks like that actually resending the write is useless (or we do not have tests that covered that scenario)

In my opinion it makes sense to complete the work, by adding specific tests cases around your change.

cc @Ghatage @karanmehta93 @mino181295

Issue #2354, opened 2020-06-04
Certificate role based authorization for Bookkeeper


Ghatage:
This is the master ticket for tracking BP-41.

Even with TLS/mTLS authentication, any other service with rootCA is able to access Bookkeeper which is unacceptable as Bookkeeper is the source or truth.

This feature allows a predefined set of roles to be 'whitelisted' to be able
to access bookkeeper based on their client certificates.
It also introduces a structure for these roles to adhere to in order to scale.

Proposal PR - [Link](https://docs.google.com/document/d/15atmnl3pS4HrhQ6fV-gSY7faIVlmU91KoApBaXPjEfg/edit?usp=sharing)

EronWright:
Just a comment on the BP itself, it is described as BP-41 but another proposal uses that identifier ([BP-41: Separate BookieId from Bookie Network Address](http://bookkeeper.apache.org/bps/BP-41-bookieid/)).  Meanwhile this proposal is not listed on the community site ([ref](http://bookkeeper.apache.org/community/bookkeeper_proposals/)).

shoothzj:
closed by #2429

PR #2355, opened 2020-06-04
Certificate role based authorization in Apache Bookkeeper.

Ghatage:
This feature allows a predefined set of services to be 'whitelisted' to be able
to access bookkeeper based on their client certificates.

### Motivation
As BookKeeper and its supported services move to a cloud friendly service based architecture, it becomes of utmost importance to monitor and allow only certain qualified services to be able to access the data in BK.
We have TLS based authentication, however, any service with the rootCA can access Bookkeeper clusters which is not desirable.

### Changes
So this feature can be broken down into two parts:
1. Certificate and roles
2. Server configuration for authorized roles

Details:
1. Certificate and roles:
Here is an example of how the SUBJECT field of a final certificate for Apache Pulsar running in the cloud would look like:
    CN=apache.bookkeeper.org
O=apache-pulsar
OU=0:pulsar-broker-role;1:cluster-1
L=San Francisco
S=CA
C=US
This shows that this bookkeeper client certificate is owned by the apache pulsar service has the role ‘pulsar-broker-role’ for entities in ‘cluster-1’.
Only those services with pulsar-broker-role should be able to access it.
We can add more fields separated by commas to increase the upstream application clusters to be able to access this bookkeeper cluster.

For example: OU=0:herddb-readonlyNode,herddb-readwriteNode;1:herddb-cluster2

Such separation of access based on services is paramount to keeping this secure as many upstream users of BookKeeper are financial institutions, databases and other services.

2. Server configuration for authorized roles
Once we have a certificate whose SUBJECT field has the OU attribute with the roles we want to authorize, on the Bookie side, we need to specify which roles are allowed.
We make this happen by introducing a server configuration option called ‘authorizedRoles’.
Since we have only static options, this will be set in stone as long as the bookie booted up with it.
If in case we need to change the allowed roles, we’ll need to stop the bookie, update the configuration file and then restart the bookie.
We can have multiple roles which are authorized as the OU field can have multiple comma separated values for roles.

Master Issue: #2354


bookkeeper-server/src/main/java/org/apache/bookkeeper/tls/BookieAuthZFactory.java
```diff
@@ -0,0 +1,133 @@
...
+import org.apache.bookkeeper.conf.ServerConfiguration;
+import org.apache.bookkeeper.proto.BookieConnectionPeer;
+import org.apache.bookkeeper.util.CertUtils;
+
+
+/**
+ * Authorization factory class.
+ */
+@Slf4j
+public class BookieAuthZFactory implements BookieAuthProvider.Factory {
```
eolivelli, line 42: what about BookieX509CertificateAuthZFactory ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/tls/BookieAuthZFactory.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    @Override
+    public void init(ServerConfiguration conf) throws IOException {
+        // Read from config
+        allowedRoles = conf.getAuthorizedRoles();
+
+        if (allowedRoles == null || allowedRoles.length == 0) {
+            throw new RuntimeException("Configuration option \'bookieAuthProviderFactoryClass\' is set to"
+                    + " \'BookieAuthZFactory\' but no roles set for configuration field \'authorizedRoles\'.");
```
eolivelli, line 58: what about usiing this.class.getName() instead of an hardcoded "BookieAuthZFactory" ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/CertUtils.java
```diff
@@ -0,0 +1,117 @@
...
+import javax.naming.ldap.LdapName;
+import javax.naming.ldap.Rdn;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Certificate parsing utilities.
+ */
+public class CertUtils {
```
eolivelli, line 40: abstract ? 
with a private default constructor

bookkeeper-server/src/test/java/org/apache/bookkeeper/tls/TestTLS.java
```diff
@@ -635,6 +635,47 @@ public void testBookieAuthPluginRequireClientTLSAuthenticationLocal() throws Exc
         assertTrue(cert instanceof X509Certificate);
     }
 
+    /**
+     * Verify that given role in client certificate is checked when BookieAuthZFactory is set.
+     * Positive test case where all given roles are present.
```
eolivelli, line 640: this test does look like a "Positive test"
you are asserting that the client is not able to work

conf/bk_server.conf
```diff
@@ -226,6 +226,15 @@ httpServerClass=org.apache.bookkeeper.http.vertx.VertxHttpServer
 #
 # permittedStartupUsers=
 
+# Certificate role based authorization for Bookie
+# To enable this option, bookieAuthProviderFactoryClass should be set to org.apache.bookkeeper.tls.BookieAuthZFactory
+# comma separated values of roles from the certificates OU field which we want to authorize for access
+#
+# Example settings:
+#   authorizedRoles=apache-pulsar-cluster-12,apache-kafka-cluster-2
```
eolivelli, line 234: looks like 'kafka' is not useful in BK context :-)

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
```diff
@@ -564,10 +564,19 @@ public void operationComplete(Future<Channel> future) throws Exception {
                     AuthHandler.ServerSideHandler authHandler = c.pipeline()
                             .get(AuthHandler.ServerSideHandler.class);
                     authHandler.authProvider.onProtocolUpgrade();
-                    if (future.isSuccess()) {
+
+                    /*
+                     * Success of the future doesn't guarantee success in authentication
+                     * future.isSuccess() only checks if the result field is not null
+                     */
+                    if (future.isSuccess() && authHandler.authenticated) {
```
eolivelli, line 572: can we add a "isAuthenticated" method to authHandler ?
this way we want access directly the field

bookkeeper-server/src/test/java/org/apache/bookkeeper/tls/TestTLS.java
Ghatage: If `testClient()` fails, it will be because of an exception thrown during Authorization.
And hence we catch that and fail the test.
Otherwise, if the test passes, it never reaches the `catch` block and finishes, passing the test
Sorry for the confusing comment, I can change that in a follow up.

lamber-ken:
rerun failure checks

rdhabalia:
rerun failure checks

rdhabalia:
@Ghatage do we need this patch for 4.11?

Ghatage:
> @Ghatage do we need this patch for 4.11?

@rdhabalia No, we can leave this for later.

Ghatage:
Hey @eolivelli,

I've made the following changes as requested:
* Added a test class for `BookieAuthZFactory` which doesn't spin up a bookie cluster.
* Added all negative test cases to the above new test class.
* Removed the 'heavy' negative test case from `TestTLS` class.
* Modified comment to explain the success case better.
* Converted `CertUtils` class to abstract with a private default constructor.
* Removed unrelated comments from `bk_server.conf`
* Added `isAuthenticated()` method to authProvider

Feature wise this is complete and switched off by default.
Lastly, once the CI tests pass, it's best to hold off on merging this before 4.11 is cut and released as discussed in the mailing list.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
Ghatage: Fixed in [7276947](https://github.com/apache/bookkeeper/commit/72769477f52e4982081ae87a8e43da0a0e34e2d1)

conf/bk_server.conf
Ghatage: Fixed in [7276947](https://github.com/apache/bookkeeper/commit/72769477f52e4982081ae87a8e43da0a0e34e2d1)

bookkeeper-server/src/test/java/org/apache/bookkeeper/tls/TestTLS.java
Ghatage: Fixed in [7276947](https://github.com/apache/bookkeeper/commit/72769477f52e4982081ae87a8e43da0a0e34e2d1)

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/CertUtils.java
Ghatage: Fixed in [7276947](https://github.com/apache/bookkeeper/commit/72769477f52e4982081ae87a8e43da0a0e34e2d1)

bookkeeper-server/src/main/java/org/apache/bookkeeper/tls/BookieAuthZFactory.java
Ghatage: Gonna let this be, as the message is specific to this class.

bookkeeper-server/src/main/java/org/apache/bookkeeper/tls/BookieAuthZFactory.java
Ghatage: I don't know about this class, I checked the code and it doesn't seem to be there..

Ghatage:
rerun failure checks

Ghatage:
rerun failure checks

Ghatage:
rerun failure checks

Ghatage:
The test suite runs fine locally and the test I've written doesn't fail either, two other tests randomly fail with `Not Enough Bookies`.
The option set is only for that test and by default it's false.
@sijie, @eolivelli  any idea why `TestTLS` keeps failing?

Ghatage:
Messed up the PR, closing this and reopening under new PR.

Ghatage:
Reopened under https://github.com/apache/bookkeeper/pull/2429

PR #2362, opened 2020-06-20
Change Python client version to 4.11.0

rdhabalia:
According to the release guide, before cutting a release we have to remove the 'alpha' qualifier from Python client version


eolivelli:
Use the script

rdhabalia:
sure. thanks

PR #2364, opened 2020-06-22
Fix distributedlog Maven reactor module

eolivelli:
- fix version
- add explicitly distributedlog to the modules of Stream project
- fix checkstyle on distributedlog, it wasn't executed


eolivelli:
same fix as #2365 but for master branch

PR #2372, opened 2020-07-11
[bookkeeper-site] fix latest version of bookkeeper website

rdhabalia:
### Motivation

After generating website using [script](http://bookkeeper.apache.org/community/release_guide/#update-website), it seems somhow it has updated site with incorrect version. I still have to find out where was the issue while generating the website. But fixing it for now to correct the version on website.


eolivelli:
Thanks

Issue #2264, opened 2020-02-13
Bookie cannot perform Garbage Collection: EntryLogger.scanEntryLog: java.lang.IllegalArgumentException: newCapacity: -610402302 (expected: 0-2147483647)


hamadodene:
**BUG REPORT**

***Describe the bug***

I got this stackstrace while BK was running  (BK version: 4.10.0)
```
20-02-13-16-33-01       Unexpected throwable caught
20-02-13-16-33-01       java.lang.IllegalArgumentException: newCapacity: -610402302 (expected: 0-2147483647)
java.lang.IllegalArgumentException: newCapacity: -610402302 (expected: 0-2147483647)
        at io.netty.buffer.AbstractByteBuf.checkNewCapacity(AbstractByteBuf.java:1419)
        at io.netty.buffer.PooledByteBuf.capacity(PooledByteBuf.java:87)
        at org.apache.bookkeeper.bookie.EntryLogger.scanEntryLog(EntryLogger.java:1014)
        at org.apache.bookkeeper.bookie.EntryLogger.extractEntryLogMetadataByScanning(EntryLogger.java:1132)
        at org.apache.bookkeeper.bookie.EntryLogger.getEntryLogMetadata(EntryLogger.java:1041)
        at org.apache.bookkeeper.bookie.GarbageCollectorThread.extractMetaFromEntryLogs(GarbageCollectorThread.java:557)
        at org.apache.bookkeeper.bookie.GarbageCollectorThread.runWithFlags(GarbageCollectorThread.java:328)
        at org.apache.bookkeeper.bookie.GarbageCollectorThread.safeRun(GarbageCollectorThread.java:309)
        at org.apache.bookkeeper.common.util.SafeRunnable.run(SafeRunnable.java:36)
        at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
        at java.base/java.util.concurrent.FutureTask.runAndReset(FutureTask.java:305)
        at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:305)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
[...]

PR #2373, opened 2020-07-13
Issue 2264: Bookie cannot perform Garbage Collection in case of corrupted EntryLogger file

eolivelli:
- in case of corrupted entry log file the bookie cannot perform GC
- handle the case of an unexpected negative entrysize

### Changes

Stop scanning the file

Master Issue: #2264

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
@@ -1011,6 +1011,11 @@ public class EntryLogger {
                 // read the entry
 
                 data.clear();
+                if (entrySize < 0) {
+                    LOG.warn("bad read for ledger entry from entryLog {}@{} (entry size {})",
+                            entryLogId, pos, entrySize);
+                    return;
+                }
                 data.capacity(entrySize);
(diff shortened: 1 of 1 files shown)


eolivelli:
please cherry pick to branch-4.10

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -1011,6 +1011,11 @@ public void scanEntryLog(long entryLogId, EntryLogScanner scanner) throws IOExce
                 // read the entry
 
                 data.clear();
+                if (entrySize < 0) {
```
rdhabalia, line 1014: as we don't know root cause, does it make sense to add a flag to skip entry-log on such failure? It can be helpful to users who don't want to skip and validate actual corruption.?

eolivelli:
@rdhabalia in my opinion this fix is like the current behavior that tries to save what looks good in the file
https://github.com/apache/bookkeeper/blob/9c34adb6b6b5033483fda973242ee363b6836ac7/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java#L996

if you are still okay can you please merge this patch ?
@rdhabalia @jiazhai 

thanks

Issue #2378, opened 2020-07-17
Build fail after updating "rocksdb" and "netty-tcnative-boringssl-static" package


odidev:
**BUG REPORT**

***Describe the bug***
After updating the latest version of netty-tcnative-boringssl-static (2.0.31.Final) and rocksdb(6.10.2) package in pom.xml, build of bookkeeper-server package is failing due to compilation warning. 
And after removing "Werror" option from compilerArguments, build passed successfully. 

***To Reproduce***

1. Update netty-tcnative-boringssl-static  to "2.0.31.Final" in pom.xml
2. Update rocksdb to "6.10.2" in pom.xml
3. Build the bookkeeper-server package by executing the "mvn -pl bookkeeper-server install -DskipTests" command
4. See error:

```
[WARNING] COMPILATION WARNING :
[INFO] -------------------------------------------------------------
[WARNING] /netty/bookkeeper/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageRocksDB.java:[128,24] setMaxBackgroundCompactions(int) in org.rocksdb.Options has been deprecated
[WARNING] /netty/bookkeeper/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageRocksDB.java:[129,24] setMaxBackgroundFlushes(int) in org.rocksdb.Options has been deprecated
[WARNING] /netty/bookkeeper/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageRocksDB.java:[138,29] setBlockCacheSize(long) in org.rocksdb.BlockBasedTableConfig has been deprecated
[WARNING] /netty/bookkeeper/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageRocksDB.java:[142,33] setFilter(org.rocksdb.Filter) in org.rocksdb.BlockBasedTableConfig has been deprecated
[INFO] 4 warnings
[INFO] -------------------------------------------------------------
[INFO] -------------------------------------------------------------
[ERROR] COMPILATION ERROR :
[INFO] -------------------------------------------------------------
[ERROR] /netty/bookkeeper/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageRocksDB.java: warnings found and -Werror specified
[INFO] 1 error
[...]

PR #2379, opened 2020-07-17
Updated netty,netty-boringssl and rocksdb

odidev:
### Descriptions of the changes in this PR:

-Updated netty,netty-boringssl and rocksdb to latest version for aarch64 support
-Updated deprecated methods of rocksdb to remove compilation warning and resolve build failure of bookkeeper-server on amd64 and aarch64 platforms as Werror flag is enabled.

### Motivation
Build of the bookkeeper-server package was failing on amd64 and aarch64 platforms.

### Changes
There are many methods that are marked deprecated in rocksdb, but in use by bookkeeper package.
So updated those methods in bookkeeper package according to current rocksdb implementation.
Also updated netty,netty-boringssl and rocksdb to latest version as they are having aarch64 support.

Master Issue: #2378

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -150,8 +150,8 @@
     <lombok.version>1.18.10</lombok.version>
     <lz4.version>1.3.0</lz4.version>
     <mockito.version>3.0.0</mockito.version>
-    <netty.version>4.1.32.Final</netty.version>
-    <netty-boringssl.version>2.0.20.Final</netty-boringssl.version>
+    <netty.version>4.1.50.Final</netty.version>
+    <netty-boringssl.version>2.0.31.Final</netty-boringssl.version>
     <ostrich.version>9.1.3</ostrich.version>
     <powermock.version>2.0.2</powermock.version>
     <prometheus.version>0.8.1</prometheus.version>
@@ -161,7 +161,7 @@
     <protoc3.version>3.5.1-1</protoc3.version>
     <protoc-gen-grpc-java.version>1.12.0</protoc-gen-grpc-java.version>
     <reflections.version>0.9.11</reflections.version>
-    <rocksdb.version>5.13.1</rocksdb.version>
+    <rocksdb.version>6.10.2</rocksdb.version>
     <shrinkwrap.version>3.0.1</shrinkwrap.version>
     <slf4j.version>1.7.25</slf4j.version>
     <snakeyaml.version>1.19</snakeyaml.version>
@@ -191,7 +191,7 @@
     <maven-javadoc-plugin.version>3.1.1</maven-javadoc-plugin.version>
     <maven-shade-plugin.version>3.2.0</maven-shade-plugin.version>
     <maven-source-plugin.version>2.2.1</maven-source-plugin.version>
-    <maven-surefire-plugin.version>2.21.0</maven-surefire-plugin.version>
+    <maven-surefire-plugin.version>3.0.0-M5</maven-surefire-plugin.version>
     <nar-maven-plugin.version>3.5.2</nar-maven-plugin.version>
     <os-maven-plugin.version>1.4.1.Final</os-maven-plugin.version>
     <protobuf-maven-plugin.version>0.5.0</protobuf-maven-plugin.version>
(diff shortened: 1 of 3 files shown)


odidev:
@eolivelli , 

> I see this is a major version bump for RocksDB.
> Is the upgrade of RocksDB automatic? I am talking about files on disk. Do you need manual upgrade of data files?
> In case of problems, if you rollback rocksdb to previous version, will the bookie work?

Latest RocksDB is supported on aarch64, and its latest version was not being used in bookkeeper. So manually boomed RocksDB in bookkeeper in this PR.If rollback rocksdb to previous version , various tests are failing on aarch64 with JNI error like:
`
[ERROR] test(org.apache.bookkeeper.bookie.storage.ldb.ConversionTest)  Time elapsed: 2.506 s  <<< ERROR!
com.google.common.util.concurrent.UncheckedExecutionException: Failed to load RocksDB JNI library
        at org.apache.bookkeeper.bookie.storage.ldb.ConversionTest.test(ConversionTest.java:118)
Caused by: java.io.IOException: Failed to load RocksDB JNI library
        at org.apache.bookkeeper.bookie.storage.ldb.ConversionTest.test(ConversionTest.java:118)
Caused by: java.lang.UnsatisfiedLinkError: /tmp/librocksdbjni4423711401334751285.so: /tmp/librocksdbjni4423711401334751285.so: cannot open shared object file: No such file or directory (Possible cause: can't load AMD 64-bit .so on a AARCH64-bit platform)
        at org.apache.bookkeeper.bookie.storage.ldb.ConversionTest.test(ConversionTest.java:118)
`

So latest rocksdb should be used.Current “Integration test” check failure does not seems specific to rocksdb or netty version updates.

eolivelli:
rerun failure tests

eolivelli:
Btw my concern is about a possible case in which you upgrade to BK 4.12 and you hit a problem and you have to rollback to 4.11.
In this case will RocksDB work?

odidev:
> Btw my concern is about a possible case in which you upgrade to BK 4.12 and you hit a problem and you have to rollback to 4.11.
> In this case will RocksDB work?

In this case, Rocksdb should work on AMD, but not on AArch64 platform because 4.11 release is having older version of rocksdb.

odidev:
@eolivelli 
Please share your opinion on the "integration test" check failure which seems failing due to timeout.

eolivelli:
@odidev I have restarted the job.
If the job fails again you should run the tests locally and perform debug.

You can split this patch in updating Netty + Updating RocksDB, this way it will be clearer which is the culprit

eolivelli:
@odidev I am sorry but on github actions we do not have much log about the error.
I suggest you to split the patch into:
- upgrade Netty
- upgrade RocksDB

I am sure that Netty will run smoothly because we are using that version in production

odidev:
Run the tests locally and found that ‘integration test’ failure was due to maven-surefire plugin compatibility with rocksdb version 5.18.4 onwards.

odidev:
Hi @eolivelli ,
Please have a look at this PR.

pom.xml
```diff
@@ -191,7 +191,7 @@
     <maven-javadoc-plugin.version>3.1.1</maven-javadoc-plugin.version>
     <maven-shade-plugin.version>3.2.0</maven-shade-plugin.version>
     <maven-source-plugin.version>2.2.1</maven-source-plugin.version>
-    <maven-surefire-plugin.version>2.21.0</maven-surefire-plugin.version>
+    <maven-surefire-plugin.version>3.0.0-M5</maven-surefire-plugin.version>
```
eolivelli, line 194: Please use 3.0.0-M4
Unfortunately M5 has some known problems about interactive development (basically it doesn't flush the output often  enough)  

Can you also point which problem did you see with Surefire?  And how it is related to RocksDB?

odidev: Hi @eolivelli ,
Using 3.0.0-M4, 'Integration Tests' checks are failing again. It seems that it is not compatible. 
Refer: https://github.com/apache/bookkeeper/pull/2379/checks?check_run_id=949575926
Please share your opinion.

odidev:
Hi @eolivelli ,
As per review suggestion, we tried to use surefire "3.0.0-M4", but that doesn't seem compatible with these optimized netty versions, and got integration test failures. So we had to use version "3.0.0-M5" again.

sijie:
@eolivelli can you review it again?

PR #2381, opened 2020-07-20
The latency of BenchThroughputLatency may be wrong due to Integer overflow when we do a large scale benchmark test 

maoling:
- When` latency.size` is  is more than `20,000,000`(e.g, `25000000`),
`(size * percentile) / 100` will be a negative number due to Integer overflow and the latency will be wrong.
- For example: `size `= 25000000, `percentile `= 99, `sampleSize `= -18199672


eolivelli:
@rdhabalia @jiazhai PTAL

eolivelli:
rerun failure tests

PR #2383, opened 2020-07-21
fix fillReadAheadCache stat bug

holmes07:
These code blocks will not be executed:
dbLedgerStorageStats.getReadAheadBatchCountStats().registerSuccessfulValue(count);
dbLedgerStorageStats.getReadAheadBatchSizeStats().registerSuccessfulValue(size);

### Motivation

fix fillReadAheadCache stat bug

### Changes

Replace return with break to exit the while loop


eolivelli:
@ivankelly @merlimat @jiazhai @rdhabalia can you please take a look ?

PR #2386, opened 2020-07-27
fix issue 1979

### LINKED ISSUE #1979: Broken "metrics" link in documentation
**BUG REPORT**

***Describe the bug***

The [`metrics.md`](https://github.com/apache/bookkeeper/blob/28dee8464764b0edceeb31ce24424f0947b5789b/site/docs/4.7.1/admin/metrics.md) file has a broken link pointing to an unknown site:
> For a full listing of available metrics, see the [Metrics](https://github.com/apache/bookkeeper/blob/28dee8464764b0edceeb31ce24424f0947b5789b/site/docs/reference/metrics) reference doc. <- (broken link)

This problem is also visible in the documentation page: https://bookkeeper.apache.org/docs/4.7.3/reference/metrics

***To Reproduce***

Just click the link posted above, you should see the problem.

***Expected behavior***

A "Metrics" page is expected with all the metrics that Bookkeeper exposes.

***Screenshots***

n/a.

***Additional context***

n/a.

sijia-w:
Fixes #1979 

Fix the links for metrics and the configuration parameters.

### DIFF
--- a/site/docs/4.10.0/admin/metrics.md
+++ b/site/docs/4.10.0/admin/metrics.md
@@ -4,7 +4,7 @@ title: Metric collection
 
 BookKeeper enables metrics collection through a variety of [stats providers](#stats-providers).
 
-> For a full listing of available metrics, see the [Metrics](../../reference/metrics) reference doc.
+> For a full listing of available metrics, see the [Metrics](https://bookkeeper.apache.org/docs/4.9.0/admin/metrics/) reference doc.
 
 ## Stats providers
 
@@ -22,7 +22,7 @@ Provider | Provider class name
 
 ## Enabling stats providers in bookies
 
-There are two stats-related [configuration parameters](../../reference/config#statistics) available for bookies:
+There are two stats-related [configuration parameters](https://bookkeeper.apache.org/docs/4.9.0/reference/config/) available for bookies:
 
 Parameter | Description | Default
 :---------|:------------|:-------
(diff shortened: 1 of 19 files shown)


sijia-w:
@eolivelli  The second link can be easily changed to a relative link. But the file pointed to by the first link is empty. site/docs/4.10.0/reference/metrics.md This is the path. Should I modify this file at the same time?

sijia-w:
@sijie The file pointed to by the first link is empty. site/docs/4.10.0/reference/metrics.md This is the path. Should I modify this file at the same time? If so, what should be the content? Thank you.

Jennifer88huang:
Since v4.5.0, the `metrics.md` file in the `reference` folder has no content. 
In v4.6.0, another `metrics.md` file is added in the `admin` folder, and add a link to the `metrics.md` in the `reference` folder, thus resulting in the link issue.
Here are two solutions:
1. Add metrics reference info in the `metrics.md` file in the `reference` folder. I suggest that we use different file names in one project.
2. Remove the current link.

Jennifer88huang:
BTW, would you add "Fixes" before "#1979", so when a PR is merged, the related issue will be closed automatically.

sijia-w:
@Jennifer88huang  Thanks very much for your feedback. If we adopt option 1, what should be added? Or you prefer option 2?

sijia-w:
@Jennifer88huang Which option do you prefer? Shall we remove the link for metrics?

Jennifer88huang:
> @Jennifer88huang Which option do you prefer? Shall we remove the link for metrics?

Yes, if no one could provide the whole content for metrics reference at the moment, just remove the link; if the content is ready, we can add it again.

sijia-w:
@Jennifer88huang  Got it. Thanks. Will do soon.

Issue #2385, opened 2020-07-24
NullPointerException in DLEventLog.purgePrevious with ZK 3.6.1


eolivelli:
**BUG REPORT**
The error is due to sending an empty "multi" op to ZK, that is broken in 3.6.1 (see https://github.com/apache/zookeeper/pull/1407)

A workaround would be to not issue the "multi" operation in case there is no op to do.
It will mitigate the problem, when the client is using 3.6.1 and cannot rollback to 3.5.x and untill the fix won't be shipped in 3.6.2

***Describe the bug***

```
20-07-24-16-15-41	java.io.IOException: Encountered exception on waiting result
java.io.IOException: Encountered exception on waiting result
	at org.apache.distributedlog.util.Utils.lambda$ioResult$0(Utils.java:701)
	at org.apache.bookkeeper.common.concurrent.FutureUtils.result(FutureUtils.java:77)
	at org.apache.distributedlog.util.Utils.ioResult(Utils.java:687)
	at magnews.gateway.datalog.distributedlog.DLEventLog.purgePrevious(DLEventLog.java:373)
	at magnews.gateway.autotuning.periodic.PeriodicRuleManager$RuleExecutor.run(PeriodicRuleManager.java:525)
	at magnews.gateway.utils.UniqueRunnable.run(UniqueRunnable.java:51)
	at magnews.gateway.utils.UniqueRunnable.run(UniqueRunnable.java:39)
	at magnews.gateway.utils.TriggeredScheduler.run(TriggeredScheduler.java:193)
	at java.base/java.lang.Thread.run(Thread.java:830)
Caused by: java.lang.NullPointerException
	at org.apache.zookeeper.ZooKeeper.multiInternal(ZooKeeper.java:2105)
	at org.apache.zookeeper.ZooKeeper.multi(ZooKeeper.java:2050)
	at org.apache.bookkeeper.zookeeper.ZooKeeperClient.access$1101(ZooKeeperClient.java:70)
	at org.apache.bookkeeper.zookeeper.ZooKeeperClient$3.zkRun(ZooKeeperClient.java:498)
	at org.apache.bookkeeper.zookeeper.ZooKeeperClient$ZkRetryRunnable.run(ZooKeeperClient.java:389)
	at org.apache.bookkeeper.zookeeper.ZooKeeperClient.multi(ZooKeeperClient.java:510)
	at org.apache.distributedlog.zk.ZKTransaction.execute(ZKTransaction.java:67)
[...]

PR #2390, opened 2020-07-28
Issue #2385: NullPointerException in Zookeeper multiple operations execution with 3.6.1

nicoloboschi:
Fix for #2385

Now we avoid to send out multiple operations request to ZooKeeper when there are not operations to execute

### DIFF
--- a/stream/distributedlog/core/src/main/java/org/apache/distributedlog/zk/ZKTransaction.java
+++ b/stream/distributedlog/core/src/main/java/org/apache/distributedlog/zk/ZKTransaction.java
@@ -44,7 +44,7 @@ public class ZKTransaction implements Transaction<Object>, AsyncCallback.MultiCa
         this.zkc = zkc;
         this.ops = Lists.newArrayList();
         this.zkOps = Lists.newArrayList();
-        this.result = new CompletableFuture<Void>();
+        this.result = new CompletableFuture<>();
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


stream/distributedlog/core/src/main/java/org/apache/distributedlog/zk/ZKTransaction.java
```diff
@@ -63,13 +63,15 @@ public void addOp(Op<Object> operation) {
...
-        try {
-            zkc.get().multi(zkOps, this, result);
-        } catch (ZooKeeperClient.ZooKeeperConnectionException e) {
-            result.completeExceptionally(Utils.zkException(e, ""));
-        } catch (InterruptedException e) {
-            Thread.currentThread().interrupt();
-            result.completeExceptionally(Utils.zkException(e, ""));
+        if (!zkOps.isEmpty()) {
+            try {
+                zkc.get().multi(zkOps, this, result);
```
eolivelli, line 68: you are not completing the `CompletableFuture result`
please check 
`public void processResult(int rc, String path, Object ctx, List<OpResult> results) {`
at line 89

eolivelli:
@rdhabalia @jiazhai Please take a look, the change is trivial but it fixed a bad problem downstream

eolivelli:
This change is to be cherry picked to branch-4.11 please

Issue #2396, opened 2020-08-08
BP-41: Separate BookieId from Bookie Network Address


eolivelli:
**BP**
This is the master ticket for tracking BP-41: Separate BookieId from Bookie Network Address

We want to separate the concepts of BookieId from BookieSocketAddress.
Currently (up to 4.11.x) there is a too strict coupling from the ID of a Bookie and its location:
- We are storing BookieSocketAddresses (simply a hostname:port or ip:port pair) into Ledger Metadata
- The Identity of a bookie is bound to the network address of the primary endpoint of the Bookie ('bookie-rpc' endpoint in term of BP-38) 

This fact is very awkward, because:
- you cannot change the network address of a Bookie
- you cannot use a custom Id for the bookie, something that would be more useful in the context of the deployment of the bookie

The proposal will run thru this problems and it will propose to separate the concept of BookieId from BookieSocketAddress.

The implementation will be divided in two separate works:
- client side changes: drop BookieSocketAddress from LedgerMetadata and use BookieId in EnsemblePlacemenetPolicies 
- bookie side changes: allow to configure a store a BookieId that is independent from the current BookieSocketAddress

Proposal PR  #2397

Patches:
- Introduce BookieId on the client #2404
- Implement custom BookieId configuration on the Bookie #2412
- Implement support in the client to follow the change of the network address of the Bookie (local cache invalidation)
- Implement a flag to not use BookieServiceInfo metadata, in order to save calls to ZK when this feature is not needed

sijie:
@eolivelli this is a great initiative!

jiazhai:
Would like to close this issue. since all related PRs are merged

PR #2397, opened 2020-08-10
BP-41 Separate BookieId from Bookie Network Address

eolivelli:
This is the design document of "BP-41 Separate BookieId from Bookie Network Address"

Master Issue: #2396


eolivelli:
@fpj @sijie @ivankelly @jvrao It would be useful to have your opinion as you did for BP-38.

This is a major change in BK APIs and internals the more  are on it the better will be

eolivelli:
> this feature will be helpful.

Thank you @rdhabalia

PR #2401, opened 2020-08-26
Allow to bypass journal for writes

merlimat:
### Motivation

In some deployment scenarios, applications might want to skip the journal, avoiding the double writing of data to disks.

One such examples is a deployment in a cloud environment with writes on locally attached disks, where absolute throughput is more important than strict data durability. In this case a journal is not providing a lot of value, considering that if the VM is terminated, the local disk is gone.

In such deployments, it's better to just rely on application level replication of data instead of the properties of the journal.

### Changes

Added option to bypass the journal at the bookie level.


karanmehta93:
> In such deployments, it's better to just rely on application level replication of data instead of the properties of the journal.

@merlimat Are we betting that the replication factor of data and low MTTR of bookies in public cloud would be good enough that we would never loose _all_ copies of data? Or are you okay with all the data going away in case of catastrophe? What is the standard configuration that you use?

jiazhai (approved):

eolivelli (changes requested):
The idea is good and we already discussed about it with @jvrao and @sijie 

I see a few limitations with this proposal:
- we are going to violate fencing and LAC in case of bookie restart 
- we are not coordinating with checkpoint, lastMark mechanism....(the position in the journal does not advance?)
- the client is not aware of what's going on

I would accept this patch if we make that the bookie is not able to restart, this way as soon as the bookie is up and running every guarantee is safe, when you reboot the bookie it is like you lost the bookie forever.

Otherwise I was thinking with my colleague @nicoloboschi about a more complex solution, with a new WriteFlag.
But we are still working on BP41 and we had not the chance to share our ideas with the community.

Drafting a BP would also be a good step as this will be a significant feature for BK

(We should also add tests that cover the change)

ivankelly:
@eolivelli we have another change that handles the situation on reboot. The problem here isn't so much about losing the replica of an entry. The problem is losing a replica which we have claimed to persist. 

The solution we have is an integrity checker, which on checks on boot (after a non-clean shutdown), whether the ledgers on the bookie have been closed. If they have not, they are marked a in limbo, which means that they will not be able to respond "I don't have this entry" because that could cause data loss during ledger recovery operations. Equally, an in limbo ledger will never accept a write, as we don't know if that ledger has been fenced.

eolivelli:
@ivankelly so we are on the same page. 


Can you share the second part of the story as a PR?
I would feel comfortable in committing both of the parts and ensure that we deliver them in the same release

ivankelly:
@eolivelli there are a bunch of dependent changes to get that change in, such as some work to make Bookie an interface, dependency injection, changes in cookie handling, etc. I do have a backlog item to start merging these, but time as always is my greatest foe.

I won't get around to looking at this until mid September at least.

IMO, there's no rush in getting the change in this PR in. I think we should go a bit further and completely remove the journal in the case that it is disabled, but again, time hasn't permitted.

eolivelli:
Yes I agree
I was also thinking that we should not even start/create the journal.

Let's keep this PR open as a reminder. I am leaving my 'request changes' vote

I minimal design document about this important change would be very awesome.

eolivelli:
@merlimat @ivankelly do you have plans to port all of this precious work to the community repository?

I am very interested in this work and I would like not to duplicate the work and especially diverge on the implementation.

From my side I would like to have a WriteFlag on the client side in order to control this behaviour, because in EmailSuccess.com we are sharing the BK cluster among lots of different components and only one (https://github.com/diennea/blobit) is very interested in this change, because it helps a lot in reducing the overall write amplification.

karanmehta93:
@eolivelli How would `blobit` use it to provide durability guarantees on ledger close (or blob close)? Does the feature ensure that all previous writes have been persisted too?

joefk:
It would be good to have journal writes controlled by the client on a per ledger basis. From Pulsar, then this can be controlled on a topic/application  level (like all the qourum settings that can be set at the client), and would be useful in any environment, even where disks are not ephemeral.

eolivelli:
@joefk 
I have a similar use case, and I would like to see this controlled by a WriteFlag (that is per-ledger and not per-cluster or per-bookie).
In my case the BK cluster (very small, like 3-5 bookies) is shared by several applications (like Database/HerdDB WAL) and for some of them I need strict durability and for other applications (like bulk data storage) I can accept partial data loss.

it looks like that @ivankelly  and @merlimat are going to contribute this fix with a cluster wide configuration.

eolivelli:
@merlimat @ivankelly 
is there any update on your roadmap for the backport of this feature ?

merlimat:
@eolivelli I think it's important to get this PR in, even though the integrity check is not ready. This is a valid mode of running that uses different tradeoffs compared to the default behavior. Nonetheless it's a very useful mode for several use cases.

eolivelli (commented):
@merlimat can we add a unit test (no need to start a bookie)?

merlimat:
What do you mean exactly with "no need to start a bookie" ? And what kind of test?

eolivelli:
I mean that we can mock the components and just verify the sequence of calls internally to the bookie.
We already have lots of tests that start ZK + BK and they are very slow.


Btw add the test the way you prefer

eolivelli:
The test is about verifying that we are not writing to the disk with that configuration option

eolivelli:
@merlimat if you want I can pickup this branch and add a test case if you do not have cycles

hsaputra:
@rdhabalia , @ivankelly - any comments or concerns about this PR?

merlimat:
@eolivelli @hsaputra added test

eolivelli (approved):
Lgtm

PR #2404, opened 2020-08-31
BP-41 BookieId - client side implementation

eolivelli:
### Motivation

This is the client side implementation of BP-41, see the design doc here
https://github.com/apache/bookkeeper/blob/master/site/bps/BP-41-bookieid.md

### Changes

- Introduce BookeId concept
- Separated BookieId from BookieSocketAddress
- in LedgerMetadata we use now BookieId and not BookieSocketAddress
- a BookieId is basically a wrapper around a string
- BookieId is compatible with BookieSocketAddress
- In order to *resolve" a BookieId to a BookieSocketAddress we use a BookieAddressResolver
- the DefaultBookieAddressResolver implementation uses BookieServiceInfo metadata, read from ZooKeeper in order to discover the current network address of the Bookie (see BP-38)

Unfortunately the patch is huge due to the changes in the client API and LedgerMetadata. This patch tries to be the smallest set of changes, see #2396 for the list of the follow up patches that will be contributed.

Notes:
- The new client and the bookie is 100% compatible with previous Bookies and Clients (and integration tests were not touched)
- We are going to introduce, in a follow up patch, a flag to not use the metadata on ZooKeeper for the resolution of the BookieSocketAddress, in order to save resources
- The ability to configure a custom BookieId, different from the network address will come in a separate patch

Master Issue: #2396


bookkeeper-benchmark/src/main/java/org/apache/bookkeeper/benchmark/BenchBookie.java
```diff
@@ -35,6 +35,7 @@
 import org.apache.bookkeeper.client.api.WriteFlag;
 import org.apache.bookkeeper.common.util.OrderedExecutor;
 import org.apache.bookkeeper.conf.ClientConfiguration;
+import org.apache.bookkeeper.net.BookieId;
```
jvrao, line 38: This is good. I am just wondering the real usecase behind this. Do you expect your bookies to change the ip and nodenames ? We had this requirement earlier when we moved to k8s where ip addresses could change, but we worked around by keeping all clients and bookies in the same k8s cluster where the node names are constant.

eolivelli: @jvrao primary use cases are:
- dynamic addresses for Bookies (or simply the ability to change hostname/ip for a bookie)
- support more endpoints and protocols for Bookie (now we are stuck to one single host:port endpoint, and we had to use StartTLS, with BP-38 the Bookie advertises all of the endpoints and thus the client will be able in the future to connect to the preferred one)

see https://github.com/apache/bookkeeper/blob/master/site/bps/BP-41-bookieid.md for more details

jiazhai: I recalled that, user from puslar side reported same requirements, they want to change a hostname/ip for a bookie

jiazhai (approved):

eolivelli:
@Ghatage very good points!

Regarding useShortHostName=true,  we have a few minimal tests (that broke during my first impl). 
With this new behaviour that flag should work because it is about the advertised address of the bookie (that is currently the same as the bookieId, in a follow up patch you will see bookieId configurable)
If you change that flag in an existing bookie and use  the automatic bookieId generation then clients would not be able to read old ledgers because they refer to unadvertised bookieIds.
With BP41 you will be able to change the advertised address and in the future even the protocol advertised by the bookie and leave existing ledger metadata valid.

Regarding 2)....that function and RackAware stuff will still use BookieSocketAddress to map a bookie to a current rack. All tests are passing. I don't have much experience here but im my opinion there won't be any need to change the placement policies. 
You will be able to implement more sophisticated policies and of you use some smart bookieId you won't need to pass from DNS or other script to map a bookie to the region/rack.

Ghatage (approved):
LGTM 👍

PR #2406, opened 2020-09-01
Fix Detected both log4j-over-slf4j.jar AND slf4j-log4j12.jar on the class path, preempting StackOverflowError.

### LINKED ISSUE #2405: Detected both log4j-over-slf4j.jar AND slf4j-log4j12.jar on the class path, preempting StackOverflowError.
I find that bk depends on log4j-over-slf4j and slf4j-log4j12, sometimes this will cause error. 
http://www.slf4j.org/codes.html#log4jDelegationLoop

Technoboy-:
Fix #2405 .

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -251,11 +251,6 @@
         <artifactId>slf4j-log4j12</artifactId>
         <version>${slf4j.version}</version>
       </dependency>
-      <dependency>
-        <groupId>org.slf4j</groupId>
-        <artifactId>log4j-over-slf4j</artifactId>
-        <version>${slf4j.version}</version>
-      </dependency>
 
       <!-- commons dependencies -->
       <dependency>
--- a/microbenchmarks/pom.xml
+++ b/microbenchmarks/pom.xml
@@ -44,7 +44,7 @@
     </dependency>
     <dependency>
       <groupId>org.slf4j</groupId>
-      <artifactId>log4j-over-slf4j</artifactId>
+      <artifactId>slf4j-log4j12</artifactId>
     </dependency>
     <dependency>
       <groupId>org.apache.bookkeeper</groupId>


eolivelli:
Did it happen only on microbenchmarks ?
Interesting

Technoboy-:
> Did it happen only on microbenchmarks ?
> Interesting

Yes.  And if we start bookie locally, it will throw the error.

eolivelli:
I am not sure I follow, the microbenchmarks is not intended to have an impact on the bookie service.
Do you have a stacktrace to share ?

Technoboy-:
> I am not sure I follow, the microbenchmarks is not intended to have an impact on the bookie service.
> Do you have a stacktrace to share ?

From this [wiki](http://www.slf4j.org/codes.html#log4jDelegationLoop),  it suggests not to keep the two jar together.
If we run Bookie locally, two jars are all in the classpath.

eolivelli:
You change does not affect the list of jars on the classpath, because you modified only the 'microbenchmarks' module, that is not a source for the 'bookie' service.
In fact we have an automatic check that checks every jar in the binary tarball and detects if there is a companion LICENSE file.

If your change drops a jar from the classpath of the bookie (and so it is no more present in the .tar.gz file) I expect that check to fail because we did not update the list of LICENSE files.

Can you please verify in the two .tar.gz files created inside bookkeeper-dist/server and bookkeeper-dist/all modules that the jar was present before the patch and it is now no more  present ?

Technoboy-:
> You change does not affect the list of jars on the classpath, because you modified only the 'microbenchmarks' module, that is not a source for the 'bookie' service.

I have removed the dependency of log4j-over-slf4j.jar in the parent pom.

pom.xml
```diff
@@ -251,11 +251,6 @@
         <artifactId>slf4j-log4j12</artifactId>
         <version>${slf4j.version}</version>
       </dependency>
-      <dependency>
```
eolivelli, line 254: this is inside "<dependencyManagement>" section, it it only a declaration of the version, it is not using the jar
see
https://stackoverflow.com/questions/2619598/differences-between-dependencymanagement-and-dependencies-in-maven

Technoboy-: Yes, I know.  But it's really occurs in my local and do not reproduce after retry.
Well, it is also ambiguous to declare the two.

Technoboy-: Should I close this pr ?

Technoboy-:
Should I close this pr ?

PR #2430, opened 2020-10-06
Do not hide Maven log in github actions

eolivelli:
Set log level to INFO.

@fpj noticed that sometimes Maven build fails but the action completes successfully

### DIFF
--- a/.github/workflows/bookie-tests.yml
+++ b/.github/workflows/bookie-tests.yml
@@ -44,13 +44,13 @@ jobs:
           java-version: 1.8
 
       - name: Maven build bookkeeper-server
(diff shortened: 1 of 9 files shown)


eolivelli:
@fpj 
as you can see here there is an error, but we are using "rerunFailingTestsCount" option
https://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html

Maven reports that the test has been re-run and the second time it passed
https://github.com/apache/bookkeeper/pull/2430/checks?check_run_id=1214125885

I would like to merge this patch

sijie (approved):

PR #2431, opened 2020-10-08
Support SNI routing in bookkeeper

sijie:

### Motivation

Netty supports sending SNI header on TLS handshake (https://github.com/netty/netty/issues/3801#issuecomment-104274440).

This PR uses that Netty feature to send SNI header while connecting to proxy software. So we can add proxy servers (e.g. Envoy, Nginx)
in front of bookies.

### Changes

- Add two configurations: `bookieServiceUri` and `bookieConnectionMode`. So the bookkeeper client can configure how it should connect to bookies.
- Send SNI header on TLS handshake

### DIFF
--- a/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/net/ServiceURI.java
+++ b/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/net/ServiceURI.java
@@ -131,6 +131,12 @@ public class ServiceURI {
     public static final ServiceURI DEFAULT_LOCAL_STREAM_STORAGE_SERVICE_URI =
         ServiceURI.create("bk://localhost:4181");
 
+    /**
+     * Service string for bookies.
+     */
+    public static final String SERVICE_BOOKIE = "bookie";
+    public static final int SERVICE_BOOKIE_PORT = 3181;
+
     private static final String SERVICE_SEP = "+";
     private static final String SERVICE_DLOG_SEP = "-";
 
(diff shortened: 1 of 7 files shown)


sijie:
This is a draft to introduce sending SNI header in bookkeeper client. However, I think we need one more chance to make it work. Because in bookkeeper protocol, we use `STARTTLS` for promoting an insecure connection to TLS secured connection. That means we send requests before `STARTTLS`.

rdhabalia (approved):
LGTM..

eolivelli:
Now that we have advanced service discovery (BP-38) and the new BookieId abstraction (BP-41, not yet landed completely) the Bookie could publish a pure-TLS endpoint on discovery metadata (BookieServiceInfo) and then the client could use directly that information

sijie:
@eolivelli Are you working on publishing a pure-TLS endpoint on discovery metadata?

eolivelli:
@sijie  I am not working on TLS, the mechanism of BP-41 is quite generic, so we will be free to support new protocols for the Bookie seamlessly in the future, once the BP-41 is finished (missing a couple of patches currently, but I feel it will be ready for BK 4.12.0)

sijie:
@eolivelli okay will look into publishing a pure-TLS endpoint

eolivelli:
@sijie PTAL here for the discovery mechanisms (BP-38)
https://github.com/apache/bookkeeper/pull/2214/files

here (BP-41 part 1 - already landed to master branch) the client is now able to leverage BP-38
https://github.com/apache/bookkeeper/pull/2404

jiazhai (approved):

eolivelli:
BP-41 has basically landed to master, so this work can move forward

dlg99:
please update TLS tests with test for SNI

PR #2435, opened 2020-10-10
BP-41 Bookie Address changes tracking

eolivelli:
- Allow the client to follow Bookie Address Changes
- If a Bookie changes network address the clients will be able to connect to the new addresess
- In order to leverage the most of this feature the Bookie must configure a fixed bookieId (see BP-41) 
- add flag enableBookieAddressTracking to enable/disable this feature, as it needs a ZK watch for every active bookie in the cluster and if you have a static cluster there is no need to invalidate the local cache and issue more ZK requests
- handle ZK session expiration events, when the ZK client is not receiving watch notifications it could miss bookie address changes, so we have to eagerly flush the whole local cache of bookie addresses

This is part of BP-41
Master ticket #2396


eolivelli:
also @nicoloboschi PTAL

eolivelli:
gently nudging at @sijie @rdhabalia @jiazhai

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ClientConfiguration.java
```diff
@@ -1764,6 +1767,27 @@ public ClientConfiguration setDelayEnsembleChange(boolean enabled) {
...
+    }
+
+    /**
+     * Enable/Disable bookie address changes tracking.
+     *
+     * @param enabled
+     *          flag to enable/disable bookie address changes tracking
+     * @return client configuration.
+     */
+    public ClientConfiguration setEnableBookieAddressTracking(boolean enabled) {
```
Ghatage, line 1786: nit: Value set can be true/false enabling/disabling, not sure if its clear to name variable as `enabled`

Ghatage, line 1786: nit: We can use this to enable/disable, so not sure if naming variable as `enabled` is clear. It could just be `value`

bookkeeper-server/src/main/java/org/apache/bookkeeper/discover/ZKRegistrationClient.java
```diff
@@ -447,4 +461,46 @@ public synchronized void unwatchReadOnlyBookies(RegistrationListener listener) {
...
+                return BookieId.parse(path.substring(slash + 1));
+            } catch (IllegalArgumentException e) {
+                log.warn("Cannot decode bookieId from {}", path, e);
+            }
+        }
+        return null;
+    }
+
+    private class BookieServiceInfoCacheInvalidationWatcher implements Watcher {
+
```
Ghatage, line 477: How frequently can a bookieID change? What factors drive its change?

Ghatage (approved):
Overall LGTM.
Can you please explain the upgrade scenario for this?
As I understand, after upgrading the bits we can still use the older bookie id but if we want, we can start the bookies with BP-41 enabled, and then we have to restart the clients with it enabled too? Is that right?

bookkeeper-server/src/main/java/org/apache/bookkeeper/discover/ZKRegistrationClient.java
eolivelli: I see these known cases:
- a bookie with a dynamic network address
- changing configuration (change DNS name....)
- you change configuration and add new endpoints (this will happen in the future, when for instance @sijie will contribute the pure TLS endpoint)

So basically I think they won't change very ofter

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ClientConfiguration.java
eolivelli: changed

eolivelli:
> Can you please explain the upgrade scenario for this?
> As I understand, after upgrading the bits we can still use the older bookie id but if we want, we can start the bookies with BP-41 enabled, and then we have to restart the clients with it enabled too? Is that right?

If you don't configure the bookieId the bookie will work as it worked in 4.11, so no problem for old clients.
If you configure bookieId in your bookie then only 4.12 clients will be able to use it.

If you want to leverage Bookie Address Change Tracking you have to:
- use bookidId
- use 4.12
- enable it in ClientConfiguration (it is enabled by default, in order to be friendly with bookies with dynamic addresses by default)

We cannot set a bookieId by default in order to be backward compatible

eolivelli:
If you simply upgrade an existing cluster from 4.9/10/11 to 4.12 everything will work well.
You can set enableBookieAddressTracking=false safely as bookies are using hostname:port as bookieId and it is not expected that this address changes.

I am setting  enableBookieAddressTracking=true by default because I think this is the future for BK and also in order to have all tests working with this feature.

The case of enableBookieAddressTracking=false is to save resources on ZK, by reducing the number of watcher.

If we were on ZK 3.6.x (client + server) we could leverage "Persistent Recursive Watches", that is a single watch for an entire tree of znodes, without the need to set it again after it triggers, and this will also mitigate the impact of these new additional watches

eolivelli:
@jiazhai @rdhabalia Please help me merging this patch

nicoloboschi (approved):
LGTM

PR #2436, opened 2020-10-11
Detect Java 11 JRE and not only JDK

eolivelli:
- Use 'lib/modules' in order to detect a modern JDK
- With this change we can detect both JDK and JRE

### DIFF
--- a/bin/common.sh
+++ b/bin/common.sh
@@ -72,7 +72,7 @@ source ${BK_CONFDIR}/bk_cli_env.sh
 
 detect_jdk8() {
 
-  if [ -f "$JAVA_HOME/bin/jshell" ]; then
+  if [ -f "$JAVA_HOME/lib/modules" ]; then
      echo "0"
   else
      echo "1"


eolivelli:
@Ghatage @nicoloboschi  PTAL

nicoloboschi (approved):
+1

bin/common.sh
```diff
@@ -72,7 +72,7 @@ source ${BK_CONFDIR}/bk_cli_env.sh
 
 detect_jdk8() {
 
-  if [ -f "$JAVA_HOME/bin/jshell" ]; then
+  if [ -f "$JAVA_HOME/lib/modules" ]; then
```
Ghatage, line 75: How does this differentiate between JDK8 and JDK11?

I see this for JDK11
```
 ~> ls $JAVA_HOME/lib/modules                                                                                                                                                                                                                                                                       
/Library/Java/JavaVirtualMachines/openjdk-11.0.2.jdk/Contents/Home/lib/modules
```

Ghatage: Note: This is on a mac, is this script designed to be run on all platforms? @eolivelli

eolivelli: I have developed this fix on Linux.

We only have to detect that we are on JDK9+ because of lots of command line options that have been dropped and prevent the JVM to start.
The main difference in JDK9 is the JPMS (Module System) so I think this is a good "test".
In JDK8 there is no "lib/modules" file, it contains all of the java "modules".
In the "JDK" and not JRE version there is also a lot of ".jmods" files, but they are not present in the single JRE version.
I don't want to execute "java -version" because we will have to parse the result and it will be too hard to maintain.

I have tested with AdoptOpenJDK and with Zulu JDK (both JDK and JRE)

jiazhai (approved):

eolivelli:
@Ghatage do I have answered to your concern?
Can we merge this patch ?

Ghatage:
Yes LGTM

PR #2450, opened 2020-10-22
ZookeeperClient should retry instantly on zk session expiry

karanmehta93:

### Motivation

Currently, we wait for `sessionTimeoutMs` before initiating a new zk cnxn. Session expiry is recevied only
when zk quprum is in a healthy state and can acquire a quorum to expire a session.
All other connection loss retries happen at zk client library transparently.

There is not much benefit waiting at BK client level.

### Changes

Updated the values for backoff and maxbackoff to 0 seconds.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/zookeeper/ZooKeeperClient.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/zookeeper/ZooKeeperClient.java
@@ -228,6 +228,9 @@ public class ZooKeeperClient extends ZooKeeper implements Watcher, AutoCloseable
             checkArgument(retryExecThreadCount > 0);
 
             if (null == connectRetryPolicy) {
+                // Session expiry event is received by client only when zk quorum is well established.
+                // All other connection loss retries happen at zk client library transparently.
+                // Hence, we don't need to wait before retrying.
                 connectRetryPolicy =
                         new BoundExponentialBackoffRetryPolicy(0, 0, Integer.MAX_VALUE);
             }
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/zookeeper/ZooKeeperClient.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/zookeeper/ZooKeeperClient.java
@@ -229,7 +229,7 @@ public class ZooKeeperClient extends ZooKeeper implements Watcher, AutoCloseable
 
             if (null == connectRetryPolicy) {
                 connectRetryPolicy =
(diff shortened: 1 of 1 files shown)


Ghatage (commented):
@karanmehta93 Do you want to also consider the `ZooKeeperClient.java` in the distributedlog module as well?

Somehow this feels like it should be configurable vs hardcoding the value though.
Will wait for more comments from others.

bookkeeper-server/src/main/java/org/apache/bookkeeper/zookeeper/ZooKeeperClient.java
```diff
@@ -229,7 +229,7 @@ public ZooKeeperClient build() throws IOException, KeeperException, InterruptedE
 
             if (null == connectRetryPolicy) {
                 connectRetryPolicy =
-                        new BoundExponentialBackoffRetryPolicy(sessionTimeoutMs, sessionTimeoutMs, Integer.MAX_VALUE);
+                        new BoundExponentialBackoffRetryPolicy(0, 0, Integer.MAX_VALUE);
```
eolivelli, line 232: can we add a comment explaining this choice ?
this way in the future it will be easier to not fall into a question like "why are we using an hardcoded 0 value" ?

eolivelli (changes requested):

bookkeeper-server/src/main/java/org/apache/bookkeeper/zookeeper/ZooKeeperClient.java
karanmehta93: I will add it.

karanmehta93:
I am not sure if distributedlog module would want that or not. Based on our current experience with BK code, these hardcoded options would work just fine. Adding a new config parameter will be adding complexity to something not probably won't be tuned much anyways.

eolivelli:
I agree with @karanmehta93 
there is no need for an additional tunable parameter.

Regarding DLog I am not sure.

"Don't fix things that are not broken":
If there is no user that is using that is reporting problems and @karanmehta93  has no experience with that code we could break it instead of fixing it

@diegosalvi is using DLog in EmailSuccess, probably he can have more context

bookkeeper-server/src/main/java/org/apache/bookkeeper/zookeeper/ZooKeeperClient.java
karanmehta93: Done.

eolivelli (approved):
LGTM

we are also seeing a problem on Pulsar (that uses BookKeeper ZooKeeper client) that would benefit from this patch.

cc @jiazhai @sijie @rdhabalia @merlimat

jiazhai (approved):

PR #2453, opened 2020-10-23
BP-41 Cleanup tools logs

eolivelli:
### Motivation

BP-41 introduces the new bookieId concept, and tools have to be improvement in order to provide better user experience.

### Changes

- Fix NPE while shutting down the ZKRegistration client no function impact)
- Clean up logs in general
- Print "bookieId" instead of Bookie Address in "bin/bkctl bookies list" command
- Print bookieId at Bookie startup (using standard scripts)

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/discover/ZKRegistrationClient.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/discover/ZKRegistrationClient.java
@@ -241,7 +241,7 @@ public class ZKRegistrationClient implements RegistrationClient {
         // because it can happen than this method is called inside the main
         // zookeeper client event loop thread
         Versioned<BookieServiceInfo> resultFromCache = bookieServiceInfoCache.get(bookieId);
-        log.info("getBookieServiceInfo {} -> {}", bookieId, resultFromCache);
+        log.debug("getBookieServiceInfo {} -> {}", bookieId, resultFromCache);
         if (resultFromCache != null) {
             return CompletableFuture.completedFuture(resultFromCache);
         } else {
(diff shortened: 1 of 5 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/discover/ZKRegistrationClient.java
```diff
@@ -462,6 +465,9 @@ public synchronized void unwatchReadOnlyBookies(RegistrationListener listener) {
     }
 
     private static BookieId stripBookieIdFromPath(String path) {
+        if (path == null) {
```
eolivelli, line 468: this path can be null for system ZK events, like ConnectionLoss

eolivelli:
@Ghatage PTAL

Ghatage:
Will take a look tomorrow.

jiazhai (approved):

bookkeeper-server/src/main/java/org/apache/bookkeeper/server/Main.java
```diff
@@ -254,23 +254,14 @@ private static ServerConfiguration parseCommandLine(String[] args)
...
-            "Hello, I'm your bookie, listening on port %1$s. Metadata service uri is %2$s."
-                + " Journals are in %3$s. Ledgers are stored in %4$s.",
+            "Hello, I'm your bookie, bookieId is %1$s, listening on port %2$s. Metadata service uri is %3$s."
+                + " Journals are in %4$s. Ledgers are stored in %5$s.",
+            conf.getBookieId() != null ? conf.getBookieId() : "<not-set>",
             conf.getBookiePort(),
             conf.getMetadataServiceUriUnchecked(),
             Arrays.asList(conf.getJournalDirNames()),
-            sb);
+            Arrays.asList(conf.getLedgerDirNames()));
```
Ghatage, line 264: Good job on the clean up here. We should have started to use it this way a long time ago.

bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/helpers/CommandHelpers.java
```diff
@@ -59,14 +58,14 @@ public static String getBookieSocketAddrStringRepresentation(BookieId bookidId,
            }
            realHostname = hostname;
         }
-        return formatBookieSocketAddress(bookieID, ip, networkAddress.getPort(), realHostname);
+        return formatBookieSocketAddress(bookidId, ip, networkAddress.getPort(), realHostname);
```
Ghatage, line 61: nit: typo `bookidId` should be `bookieId`

Ghatage (approved):
LGTM. Minor nit which is ok to be ignored.
We should do more of these clean ups!

bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/helpers/CommandHelpers.java
eolivelli: fixed. @Ghatage PTAL again

Issue #2209, opened 2019-12-04
BP-40 Audit logging for BookieShell


Ghatage:
**BP**

This is the master ticket for tracking BP-40 : Audit logging for BookieShell.

In order to make a software compliant with the security and privacy controls, we need to meet certain audit requirements.
One of them being able to track every human / user intervention in the software.
Zookeeper has something similar going on in [ZOOKEEPER-1260](https://issues.apache.org/jira/browse/ZOOKEEPER-1260).

The idea is that all user and human actions should be logged separately in a different log file which has stricter permissions and all interactions should be logged in detail.

This is useful for admins to later surmise what user interactions happened and can reconstruct the state of the system before and after the user interaction based on the audit logs.

Proposal PR - [Link](https://docs.google.com/document/d/1skocaIYC-9ZUsG7w2VAiGDbUG39ybYivrfvw9o5r2VA/edit?usp=sharing)

PR #2455, opened 2020-10-25
BP-40: Convert all System.*.print statements to LOG.* statements

Ghatage:
### Motivation

In order to have an audit log of all user interactions, all print statements which come from such interactions need to be logged.
We have two user driven ways to mutate BookKeeper:
* BookieShell
* bkctl
Both internally use the same code from `tools/cli`

### Changes

The logging in the code was inconsistent. At times we used `System.out.println` and at times we used `LOG.info`.
As a part of this work item we change these in the following way:

* For every `cli` class, confirm if a Logger object is present, if not create one.
* Convert all `System.out.println()` -> `LOG.info()`
* Convert all `System.err.println()` -> `LOG.error()`

### Ramifications

BookieShell uses `log4j.shell.properties`
bkctl uses `log4j.cli.properties`

Both log4j property files have `CONSOLE appenders`, so the like previous `System.out.println()`, the new `LOG` statements will also be printed to console.
This is mentioned explicitly as certain upstream consumers / scripts utilize BookieShell / bkctl to assess the state of the BK cluster.

Master Issue: #2209

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/commands/autorecovery/ListUnderReplicatedCommand.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/commands/autorecovery/ListUnderReplicatedCommand.java
@@ -144,14 +144,14 @@ public class ListUnderReplicatedCommand extends BookieCommand<ListUnderReplicate
             while (iter.hasNext()) {
                 UnderreplicatedLedger underreplicatedLedger = iter.next();
                 long urLedgerId = underreplicatedLedger.getLedgerId();
-                System.out.println(ledgerIdFormatter.formatLedgerId(urLedgerId));
+                LOG.info(ledgerIdFormatter.formatLedgerId(urLedgerId));
                 long ctime = underreplicatedLedger.getCtime();
                 if (ctime != UnderreplicatedLedger.UNASSIGNED_CTIME) {
-                    System.out.println("\tCtime : " + ctime);
+                    LOG.info("\tCtime : " + ctime);
                 }
                 if (printMissingReplica) {
                     underreplicatedLedger.getReplicaList().forEach((missingReplica) -> {
-                        System.out.println("\tMissingReplica : " + missingReplica);
+                        LOG.info("\tMissingReplica : " + missingReplica);
                     });
                 }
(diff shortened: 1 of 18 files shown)


Ghatage:
@zymap also PTAL since you have done lots of work in this area!

eolivelli:
I am not sure this is good.
I have to think about it a bit more.
How can we distinguish the output for the user from the logs of the application ?

The logger usually is something that writes to a file useful information, it is not part of the UI (the stuff the user sees)

for instance, if I want to ask for a username I am not going to use
log.info("Please tell me the username:")
console.readLine();

but
console.println("Please tell me the username:")
console.readLine();

these tool are interactive, I am not sure the log system is what we want to use

Ghatage:
Hi @eolivelli,

Apologies as I think I wasn't clear. Let me explain.

1. BookieShell / bkctl already use Loggers and print to a log file **as well as** to the console.
For example in [this command](https://github.com/apache/bookkeeper/blob/aff51a53dbe38470c9c5f80954210af56965e73b/bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/commands/bookie/LocalConsistencyCheckCommand.java#L61) like many others already we have `LOG.info()` which prints to the console as well as to the log file.

2. So now we know that when we execute these interactive commands, we print to the console as well as log to a file.
The problem arises when there is inconsistency. 
From my changes you can see that there are places where we have `System.out.println()` instead of `LOG.info()`
So in those cases, we **only print to the console** and don't save that interaction output in the log file.

This change relieves that inconsistency and makes all commands print to console **as well as** log file.
This way, when we implement audit logging, it will save all user interactions.


> How can we distinguish the output for the user from the logs of the application ?

We currently log bookieshell logs differently in a different log file.
So interactions from the user (bookieshell/bkctl) will go into that log file.
The logs of the server are in a separate log file.
Log file name, location and access properties for bkctl are controlled by [log4j.cli.properties](https://github.com/apache/bookkeeper/blob/master/conf/log4j.cli.properties)
and log file name, location and other properties for BookieShell are controlled by [log4j.shell.properties](https://github.com/apache/bookkeeper/blob/master/conf/log4j.shell.properties)

eolivelli (approved):
I see.
thank you @Ghatage 

merging this patch now

Issue #2463, opened 2020-10-28
java.lang.NullPointerException with  metadataServiceUri = zk:///ledgers (Default on Docker image)


eolivelli:
***Describe the bug***
if you configure a bad metadata service uri you can this error

java.lang.NullPointerException
	at org.apache.bookkeeper.meta.zk.ZKMetadataDriverBase.getZKServersFromServiceUri(ZKMetadataDriverBase.java:64)
	at org.apache.bookkeeper.meta.zk.ZKMetadataDriverBase.initialize(ZKMetadataDriverBase.java:186)
	at org.apache.bookkeeper.meta.zk.ZKMetadataBookieDriver.initialize(ZKMetadataBookieDriver.java:60)
	at org.apache.bookkeeper.bookie.Bookie.instantiateMetadataDriver(Bookie.java:1080)
	at org.apache.bookkeeper.bookie.Bookie.<init>(Bookie.java:698)
	at org.apache.bookkeeper.proto.BookieServer.newBookie(BookieServer.java:140)
	at org.apache.bookkeeper.proto.BookieServer.<init>(BookieServer.java:108)
	at org.apache.bookkeeper.server.service.BookieService.<init>(BookieService.java:52)
	at org.apache.bookkeeper.server.Main.buildBookieServer(Main.java:313)
	at org.apache.bookkeeper.server.Main.doMain(Main.java:226)
	at org.apache.bookkeeper.server.Main.main(Main.java:208)
***To Reproduce***

docker run apache/bookkeeper:4.11.1

***Expected behavior***

A clearer error message like 'Invalid metadata service URI"

eolivelli:
@nicoloboschi PTAL

nicoloboschi:
I can take care of it

nicoloboschi:
@eolivelli maybe we should change default zkServer param? I'd like ready-to-run solution, like 'localhost'? I don't really know BK docker container use cases

Otherwise we may add some useful logs when the container starts, explaining where to set zkserver parameter

eolivelli:
The problem is not about docker. The problem us the NPE.
We should report a better error.

No one will ever run the docker image without ZK

PR #2469, opened 2020-10-29
java.lang.NullPointerException with metadataServiceUri = zk:///ledger…

nicoloboschi:
…s (Default on Docker image) #2463

Avoid NPE and throw IllegalArgumentException in case of invalid service metadata URI

Master Issue: #2463

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/zk/ZKMetadataDriverBase.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/zk/ZKMetadataDriverBase.java
@@ -60,7 +60,7 @@ public class ZKMetadataDriverBase implements AutoCloseable {
 
     protected static final String SCHEME = "zk";
 
-    public static String getZKServersFromServiceUri(URI uri) throws IllegalArgumentException {
+    public static String getZKServersFromServiceUri(URI uri) {
         String authority = uri.getAuthority();
         if (authority == null) {
             throw new IllegalArgumentException("Invalid metadata service URI format: " + uri);
@@ -69,7 +69,7 @@ public class ZKMetadataDriverBase implements AutoCloseable {
     }
 
     @SuppressWarnings("deprecation")
-    public static String resolveZkServers(AbstractConfiguration<?> conf) throws IllegalArgumentException {
+    public static String resolveZkServers(AbstractConfiguration<?> conf) {
         String metadataServiceUriStr = conf.getMetadataServiceUriUnchecked();
         if (null == metadataServiceUriStr) {
             return conf.getZkServers();
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/zk/ZKMetadataDriverBase.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/zk/ZKMetadataDriverBase.java
@@ -33,7 +33,6 @@ import lombok.Setter;
 import lombok.SneakyThrows;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.bookkeeper.conf.AbstractConfiguration;
-import org.apache.bookkeeper.conf.UncheckedConfigurationException;
 import org.apache.bookkeeper.meta.AbstractZkLedgerManagerFactory;
 import org.apache.bookkeeper.meta.HierarchicalLedgerManagerFactory;
 import org.apache.bookkeeper.meta.LayoutManager;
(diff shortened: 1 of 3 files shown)


nicoloboschi:
rerun failure checks

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/zk/ZKMetadataDriverBase.java
```diff
@@ -60,8 +60,12 @@
 
     protected static final String SCHEME = "zk";
 
-    public static String getZKServersFromServiceUri(URI uri) {
-        return uri.getAuthority().replace(";", ",");
+    public static String getZKServersFromServiceUri(URI uri) throws IllegalArgumentException {
```
Ghatage, line 63: This is great @nicoloboschi!
Can we also get a testcase for this?
Also `getZKServersFromServiceUri` is used in multiple places - can we also account for those usages with respective `catch()` blocks for `IllegalArugmentException`?
Otherwise we'll end up with an annoying stack trace.

nicoloboschi: Thanks @Ghatage 
I added a test for bookie initialization as I think is the only place where it is worth to handle with a checked exception

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/zk/ZKMetadataDriverBase.java
```diff
@@ -60,12 +60,16 @@
...
+        String authority = uri.getAuthority();
+        if (authority == null) {
+            throw new IllegalArgumentException("Invalid metadata service URI format: " + uri);
+        }
+        return authority.replace(";", ",");
     }
 
     @SuppressWarnings("deprecation")
-    public static String resolveZkServers(AbstractConfiguration<?> conf) {
+    public static String resolveZkServers(AbstractConfiguration<?> conf) throws IllegalArgumentException {
```
eolivelli, line 72: who is calling this method ? "resolveZkServers"
we should catch IllegalArgumentException as well

eolivelli (changes requested):

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/zk/ZKMetadataDriverBase.java
reddycharan, line 72: 'IllegalArgumentException' is RuntimeException (uncheckedexception), it is unusual to declare method throwing RuntimeException.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieInitializationTest.java
```diff
@@ -1576,4 +1576,17 @@ private void bookieConnectAfterCookieDeleteWorker(ServerConfiguration conf, Regi
             LOG.info("As expected Bookie fails to come up without a cookie in metadata store.");
         }
     }
+
+    @Test
+    public void testInvalidServiceMetadataURI() throws Exception {
+        ServerConfiguration conf = TestBKConfiguration.newServerConfiguration();
+        conf.setMetadataServiceUri("zk+flat:///ledgers"); // default value in docker image
```
eolivelli, line 1583: a more sensible example is zk+null and not zk+flat

we could also add tests about other silly cases, like:
- single slash
- no slash at all
- empty string
- null string

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/zk/ZKMetadataDriverBase.java
nicoloboschi: removed

nicoloboschi: I checked the usages of resolveZkServers and I think it is ok to keep IAE unwrapped

nicoloboschi:
@eolivelli @Ghatage please have a look

eolivelli (approved):
LGTM
@Ghatage @reddycharan you left comments, please take a look again

sijie (approved):

nicoloboschi:
@Ghatage if you approve the changes we can do the merge

Ghatage (approved):
LGTM

PR #2470, opened 2020-10-29
Introduce new retry policy for metadata operations

karanmehta93:

### Motivation
Current retry backoff policy is truly exponential, hence retries are spaced out at large intervals. There is maxBackOff time that limits it, but it doesn't have a concept of deadline. Hence, applications with shorter SLA's have issues dealing with ZK failures and recovering from it.

### Changes

The new policy introduces an aggressive retry mechanism with the backoff constants as less exponential. The deadline ensures that we retry at least once at the deadline. The stack trace would indicate the operation at risk.

This doesn't update the product code, which is still in testing. I am splitting PR's for for sake of reviews.


Ghatage:
rerun failure checks

eolivelli (approved):
LGTM

eolivelli:
@Ghatage this patch is not trivial, I guess you "approve" it.
feel free to merge

jiazhai (approved):

PR #2472, opened 2020-10-29
Add Add ForceAuditChecksCmd to reset auditor checks last run time

karanmehta93:

### Motivation

Reset the last run time for auditor checks. Reboot auditor leader immediately to trigger the checks. This is mostly used for testing infra and get the sanity check of a cluster on demand.

### Changes

Added a new shell command and tests

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
@@ -28,6 +28,7 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.attribute.BasicFileAttributes;
 import java.nio.file.attribute.FileTime;
+import java.sql.Timestamp;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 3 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
```diff
@@ -1799,6 +1804,76 @@ public int runCmd(CommandLine cmdLine) throws Exception {
...
+
+            if (checkAllLedgersCheck || placementPolicyCheck  || replicasCheck) {
+                runFunctionWithLedgerManagerFactory(bkConf, mFactory -> {
+                    try {
+                        try (LedgerUnderreplicationManager underreplicationManager =
+                                     mFactory.newLedgerUnderreplicationManager()) {
+                            // Arbitrary value of 21 days chosen since current freq of all checks is less than 21 days
+                            long time = System.currentTimeMillis() - (21 * 24 * 60 * 60 * 1000);
+                            if (checkAllLedgersCheck) {
+                                LOG.info("Resetting CheckAllLedgersCTime to : " + time);
```
eolivelli, line 1851: can we ALSO print `time` formatted for humans ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
```diff
@@ -1799,6 +1804,76 @@ public int runCmd(CommandLine cmdLine) throws Exception {
...
+                                underreplicationManager.setCheckAllLedgersCTime(time);
+                            }
+                            if (placementPolicyCheck) {
+                                LOG.info("Resetting PlacementPolicyCheckCTime to : " + time);
```
eolivelli, line 1855: can we ALSO print `time` formatted for humans ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
```diff
@@ -1799,6 +1804,76 @@ public int runCmd(CommandLine cmdLine) throws Exception {
...
+                                underreplicationManager.setPlacementPolicyCheckCTime(time);
+                            }
+                            if (replicasCheck) {
+                                LOG.info("Resetting ReplicasCheckCTime to : " + time);
```
eolivelli, line 1859: can we ALSO print `time` formatted for humans ?

eolivelli (changes requested):
very good !
thanks for this useful tool.
I left a comment.
Also, I believe that this patch does not add the same command to "bkctl" but only to BookieShell.
BookieShell is the old tool, all of the new commands should also be added to bkctl.
It is not a matter of duplicating code, but only to register the handler also in bkctl runtime, take any other command as example

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
karanmehta93: Updated it.

sijie (approved):

eolivelli (approved):
LGTM

Issue #1533, opened 2018-07-03
Information (Header) of TransientLedgerInfo is not persisted


reddycharan:
**FEATURE REQUEST**

1. Please describe the feature you are requesting.

TransientLedgerInfo class borrows the logic from FileInfo, which is used for holding all the transient states for a given ledger when DbLedgerStorage is used for LedgerStorage. But unlike FileInfo, in TransientLedgerInfo, Header (containing masterKey, stateBits and explicitLac) is not defined and it is not persisted. 

2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have). Are you currently using any workarounds to address this issue?

With https://github.com/apache/bookkeeper/issues/1527 we are going to persist explicitLac, but since TransientLedgerInfo doesn't has Header/persistence logic, for now I'm going to ignore persisting explicitLac logic in TransientLedgerInfo.

reddycharan:
Hey @sijie , as we discussed earlier, created open issue for Header/persistence logic in TransientLedgerInfo

@merlimat FYI

eolivelli:
This means that woth DbLedgerStorage we will not persist ExplicitLAC ?
Are you planning to implement it in a follow up patch? Otherwise we should take care of documenting this important difference in DbLedgerStorage vs SortedLedgerStorage

eolivelli:
I am working on this task.
Hopefully we can deliver this fix for 4.12.0.

PR #2474, opened 2020-10-31
Issue 1533: DbLedgerStorage does not persist ExplicitLAC: Information (Header) of TransientLedgerInfo is not persisted

eolivelli:
### Motivation
DbLedgerStorage does not persist "ExplicitLAC" on RocksDB, so when you restart a Bookie the information is lost.
This work can be considered a follow up of @reddycharan  work at #1527

### Changes

- persist ExplicitLAC on DBLedgerStorage
- save ExplicitLAC on RocksDB (it is an optional field, so this change is backward compatible)
- enable testExplicitLACIsPersisted test even for DBLedgerStorage
- on TransientLedgerInfo rewind the ByteBuf in order to be able to use it again while writing to RocksDB
- use computeIfAbsent in getOrAddLedgerInfo (code clean up, but code looked strange, probably subject to some possible race condition)

Master Issue: #1533

### DIFF
--- a/bookkeeper-proto/src/main/proto/DbLedgerStorageDataFormats.proto
+++ b/bookkeeper-proto/src/main/proto/DbLedgerStorageDataFormats.proto
@@ -27,4 +27,5 @@ message LedgerData {
     required bool exists = 1;
     required bool fenced = 2;
     required bytes masterKey = 3;
+    optional bytes explicitLac = 4;
 }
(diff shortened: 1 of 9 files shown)


eolivelli:
@reddycharan please take a look as this is a follow up of your work
@merlimat please take a look as you are more confident with DbLedgerStorage than me

@jiazhai  this is the needed fix in order to fully leverage ExplicitLAC on the new API and on Pulsar

@sijie I am not sure if we have to change also tryReadLastAddConfirmedAsync and readLastAddConfirmedAndEntryAsync, those are part of the Long-Poll feature, currently long-poll still does not benefit from ExplicitLAC

jiazhai (approved):

PR #2480, opened 2020-11-04
Upgrade Guava to 30.0-jre

eolivelli:
- Upgrade Guava to 30.0.jre (see https://github.com/google/guava/releases)
- adapt code, as Futures.addCallback(3-args) has been removed (see https://github.com/google/guava/commit/dfb00017144bf48c408bc4cfcebadc0a0498dc73)

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/DefaultSpeculativeRequestExecutionPolicy.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/DefaultSpeculativeRequestExecutionPolicy.java
@@ -20,6 +20,7 @@
  */
 package org.apache.bookkeeper.client;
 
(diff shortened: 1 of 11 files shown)


eolivelli:
rerun failure tests

eolivelli:
@nicoloboschi this change will impact every downstream user of DistributedLog, I saw this problem while writing integration tests of Pulsar + EmailSuccess.com, EmailSuccess uses DLog and it does not support latest version of Guava

eolivelli:
I would like to put this change in 4.12.0, this way users will have a way to upgrade to latest Guava.
The method that I am now using is present in older versions of Guava, so it should be a backward compatible change

eolivelli:
@sijie can we include this fix in 4.12.0 ?

sijie (approved):

jiazhai (approved):

PR #2484, opened 2020-11-11
Publish BP-38 BookieEndpointInfo for BookKeeper Standalone mode

eolivelli:
### Motivation

When you run "bin/bookkeeper standalone" the server does not publish BP-38 endpoint info.
This way you cannot use bookieId (BP-41).

### Changes

Fix BookieService in "standalone" package, that is very like to the BookieService in "bookkeeper-server".
The code cannot be reused, even if the name is similar the overall design of the class hierarchy is different.

### DIFF
--- a/conf/standalone.conf
+++ b/conf/standalone.conf
@@ -16,12 +16,15 @@
 # * limitations under the License.
 # */
 
-# Standalone configuration
+# This file is used by 'bin/bookkeeper standalone' command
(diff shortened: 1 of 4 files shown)


eolivelli:
@mino181295 you implemented BP-38, please take a look as well

jiazhai (approved):

eolivelli:
rerun failure checks

PR #2491, opened 2020-11-18
Upgrade to Curator 5.1

eolivelli:

Curator 5.x support ZooKeeper 3.6.x.
This change is needed to able the upgrade of Pulsar to ZooKeeper 3.6.2 

### Changes

Upgrade Curator

### DIFF
--- a/stream/storage/impl/src/main/java/org/apache/bookkeeper/stream/storage/impl/cluster/ZkClusterMetadataStore.java
+++ b/stream/storage/impl/src/main/java/org/apache/bookkeeper/stream/storage/impl/cluster/ZkClusterMetadataStore.java
@@ -156,7 +156,7 @@ public class ZkClusterMetadataStore implements ClusterMetadataStore {
         synchronized (this) {
             if (assignmentDataCache == null) {
                 assignmentDataCache = new org.apache.curator.framework.recipes.cache.NodeCache(client,
-							zkClusterAssignmentPath);
+                                                                                               zkClusterAssignmentPath);
                 try {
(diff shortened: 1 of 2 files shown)


eolivelli:
rerun failure checks

stream/storage/impl/src/main/java/org/apache/bookkeeper/stream/storage/impl/cluster/ZkClusterMetadataStore.java
```diff
@@ -58,7 +58,7 @@
     private final String zkClusterAssignmentPath;
 
     private final Map<Consumer<Void>, NodeCacheListener> assignmentDataConsumers;
-    private NodeCache assignmentDataCache;
+    private org.apache.curator.framework.recipes.cache.NodeCache assignmentDataCache;
```
Ghatage, line 61: Sorry I think I might be missing something. Are there overloaded types of `NodeCache` in the new version?
Why are we giving the fully qualified class name?

eolivelli: NodeCache as been deprecated, there is a new API, but I wanted to keep the patch as small as possible

Ghatage: I see, thanks

Ghatage (approved):
LGTM

PR #2493, opened 2020-11-19
Opportunistic Striping

eolivelli:
### Motivation
If you have a very small cluster, 3 bookies, you must use **2-2-2** as replication parameters in order to guarantee durability and tolerate the failure of at least one bookie.
With this situation you cannot leverage the striping feature of BookKeeper, when **ensembleSize > writeQuorumSize**.

With striping you can distribute your data among all of the bookies, and use better the available disk space.
You can also exploit the ability to write to/read from more disks (bookies) in parallel, achieving better overall performances.

With this change we introduce a new client option "opportunistic striping", that basically tells to the BK client to use as much bookies as configured in the ensembleSize parameter in the best case scenario, and to use less and less bookies, up to the configured writeQuorumSize.

### Changes
- New ClientConfiguration option "opportunisticStriping"
- in case of opportunisticStriping the client is allowed to use less bookies than the requested ensemble size, up to writeQuorumSize
- add test cases

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerCreateOp.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerCreateOp.java
@@ -125,27 +125,49 @@ class LedgerCreateOp {
      * Initiates the operation.
      */
     public void initiate() {
+        int actualEnsembleSize = ensembleSize;
+        List<BookieId> ensemble = null;
+        // select bookies for first ensemble
+        if (bk.getConf().getOpportunisticStriping()) {
+            BKNotEnoughBookiesException lastError = null;
+            // we would like to select ensembleSize bookies, but
+            // we can settle to writeQuorumSize
+            while (actualEnsembleSize >= writeQuorumSize) {
+                try {
+                    ensemble = bk.getBookieWatcher()
+                        .newEnsemble(actualEnsembleSize, writeQuorumSize, ackQuorumSize, customMetadata);
+                    lastError = null;
+                    break;
+                } catch (BKNotEnoughBookiesException e) {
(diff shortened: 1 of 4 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerCreateOp.java
```diff
@@ -125,28 +125,52 @@
...
+            if (lastError != null) {
+                createComplete(lastError.getCode(), null);
+                return;
+            }
+        } else {
+            try {
+                ensemble = bk.getBookieWatcher()
+                        .newEnsemble(actualEnsembleSize, writeQuorumSize, ackQuorumSize, customMetadata);
+            } catch (BKNotEnoughBookiesException e) {
+                LOG.error("Not enough bookies to create ledger with {}/{}/{}",
```
Ghatage, line 157: Very useful error log, can we also have some details as to what the three arguments are?
something like `Not enough bookies to create ledger with EnsembleSize/WriteQuorum/AckQuorum : {}/{}/{}`
So it'll be helpful for users to know which is which.

Ghatage (approved):
LGTM

nicoloboschi (approved):
Great idea, +1

eolivelli (commented):
@Ghatage @nicoloboschi 
I have enhance the logging, now we are writing an "info" message per each failed trail, up to the last one, which is logged as "error" level

eolivelli:
@Ghatage please cherry pick to branch-4.12 and branch-4.11

Ghatage:
> @Ghatage please cherry pick to branch-4.12 and branch-4.11
Will do.

Ghatage:
rerun failure checks

PR #2494, opened 2020-11-22
[WIP][LedgerHandler] Try add read ahead capability to ledgerhandler.

MarvinCai:

### Motivation

Try add read ahead capability to ledgerhandler.

### Changes

Try to implement read ahead logic in ledgerhandler similar to what DLog.
Try to prefetch entry in order from dirstentry(0) and repeat till cache is full.
Try to fulfill read request from cache first and file read request if entry is not in cache and combine the final result.

Master Issue: #apache/pulsar#3804

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
@@ -47,6 +47,7 @@ import java.util.Queue;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
@@ -78,6 +79,7 @@ import org.apache.bookkeeper.common.concurrent.FutureUtils;
 import org.apache.bookkeeper.common.util.MathUtils;
 import org.apache.bookkeeper.net.BookieId;
 import org.apache.bookkeeper.proto.BookieProtocol;
+import org.apache.bookkeeper.proto.BookkeeperInternalCallbacks;
 import org.apache.bookkeeper.proto.checksum.DigestManager;
 import org.apache.bookkeeper.stats.Counter;
 import org.apache.bookkeeper.stats.Gauge;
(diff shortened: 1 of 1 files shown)


MarvinCai:
Currently using a map with expiration as cache, but it can't release entry when evicting expired entries.

MarvinCai:
@eolivelli I'm trying to add read ahead in ledgerhandler, this is like initial implementation, can you help take a look to see if I'm on the right track or if I'm missing something. Thanks.

eolivelli:
@MarvinCai I missed your message, sorry for late replay.

Can you please tell more about your approach ?

It would be also good to start to add tests in order to demonstrate the feature.

MarvinCai:
@eolivelli hi, no worry
I'm trying to add the readahead by using a map of [entryId -> content] as readahead entry cache of size X
Entry will be read into the cache by the doReadAhead method which will read Y entry a time and keep reading till fill the cache or encounter error.
The cached entry will have ttl and will be evicted when expired, a background task should also try to fill the cache at the same rate of entry expiration (instead of delete cached entry every time it's used to fulfill a read request cause I thought it's possible an entry might be read multiple times)
When read request come in we'll first try to search by entryId in cache to fulfill the request, if it missed we'll try to read with ledgerHandler and do concatenation if necessary (in case some of read operation for cached entry failed) 
I'll definitely have test case for it but just want to make sure this sounds like a valid approach

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -173,6 +184,33 @@
     final Counter lacUpdateMissesCounter;
     private final OpStatsLogger clientChannelWriteWaitStats;
 
+    /**
+     * Read ahead related parameters
+     */
+    // Number of read ahead entries allowed in cache.
+    private final int maxNumReadAheadEntries = 0;
```
eolivelli, line 191: can we group all of these fields into a separate datastructure ?
all of these fields are useful only in case of read ahead feature enabled

eolivelli (changes requested):
I am not sure about the API.
How do you enable this feature ?
will is be a ClientConfiguration setting ?

you could add a new ReadFlag.
we already have WriteFlags, in this case you could introduce the concept of ReadFlags and implement a new ReadFlag#READ_AHEAD

MarvinCai:
close this one in flavor of : https://github.com/apache/pulsar/pull/9056

PR #2496, opened 2020-11-24
Start HTTP Server in Bookie Standlone Mode

eolivelli:
### Motivation
- when you launch "bookkeeper standalone" service the system does not take into account the HTTP Server configuration
- the HTTP Server is important because it allows to use the HTTP API https://bookkeeper.apache.org/docs/latest/admin/http/
- the standalone service is very useful when you are working on client applications (like BKVM)

### Changes

- consider HTTP server configuration in 'bin/bookkeeper standalone' command
- add default example configuration
- add VertxHttpServer to dependencies of Standalone Module
- the change is basically a copy and paste of this part of Main.java, but applied to the Main class of the Standalone Starter, which is a separate code: https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/server/Main.java#L327

 ## Verify this change
After this change when you run "bookkeeper standalone" the HTTP API will be enabled
For instance you can check the Bookie API with
http://localhost:8080/api/v1/bookie/info

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/server/service/HttpService.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/server/service/HttpService.java
@@ -47,7 +47,8 @@ public class HttpService extends ServerLifecycleComponent {
 
         HttpServerLoader.loadHttpServer(conf.getServerConf());
         server = HttpServerLoader.get();
-        checkNotNull(server);
+        checkNotNull(server, "httpServerClass is not configured or it could not be started,"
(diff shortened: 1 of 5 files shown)


nicoloboschi (approved):
+1 LGTM

zymap (approved):
Nice patch!

mino181295 (approved):

merlimat (approved):

rdhabalia (approved):

sijie (approved):

PR #2507, opened 2020-12-04
Add Some Clarifications to the AutoRecovery Documentation

michaeljmarshall:

### Motivation

I recently started using Apache Bookkeeper as part of my Apache Pulsar cluster. In learning some of the operational tasks related to managing the bookies and autorecovery, I noticed a few places where the documentation could be improved.

### Changes

The changes are all to the `AutoRecovery.md` file for version 4.12.0. They might apply to earlier versions, but I'm not sure how documentation updates are handled. My first change is to explicitly mention the way to turn off the AutoRecovery daemon on the bookie nodes. This configuration was not obvious to me as a new user of the project, and I accidentally had recovery workers running on all nodes. (Regarding that, I mention in the PR that it is possible to run the process on bookie and autorecovery nodes. Perhaps this isn't an intended use case though? If that is so, it should be mentioned in the documentation.) I also added a clarification to the `disable` auto recovery documentation because it wasn't obvious to me that running the script would stop autorecovery for the whole cluster.

Thanks!

### DIFF
--- a/site/docs/4.12.0/admin/autorecovery.md
+++ b/site/docs/4.12.0/admin/autorecovery.md
@@ -45,10 +45,11 @@ AutoRecovery is a process that:
 * automatically detects when a {% pop bookie %} in your BookKeeper cluster has become unavailable and then
 * rereplicates all the {% pop ledgers %} that were stored on that bookie.
 
-AutoRecovery can be run in two ways:
+AutoRecovery can be run in three ways:
 
 1. On dedicated nodes in your BookKeeper cluster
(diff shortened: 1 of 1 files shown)


eolivelli (approved):
Thank you !

eolivelli:
can you please update the pages in 'latest' directory as well ?

this is the template for the upcoming versions
https://github.com/apache/bookkeeper/tree/master/site/docs/latest

it is fine to update 4.12.0 as well as you did

sijie (approved):

PR #2519, opened 2020-12-10
Allow DNSToSwitchMapping to access BookieAddressResolver

eolivelli:
In order to allow Pulsar to leverage BookieId with custom Bookie-To-Rack Mapping (ZkBookieRackAffinityMapping) we need to inject the BookieAddressResolver into it.

There is not way to do it currently because it is created by the BookKeeper client, and we have to inject the same BookieAddressResolver used by the client itself.

*Changes*
- add new setBookieAddressResolver method (use 'default' methods in order to preserve backward compatibility
- call the new method in every point in which we instantiate a DNSToSwitchMapping and when we receive an instance at construction time

*Compatibility*
- this change is 100% compatible with previous versions


eolivelli:
related Pulsar issue
https://github.com/apache/pulsar/issues/8901

rdhabalia (approved):
LGTM

eolivelli:
@Ghatage can you please commit this patch?

@rdhabalia Already approved

Ghatage:
Yes I actually have this open and reviewing it, will merge if I have no questions

Ghatage (approved):

PR #2537, opened 2021-01-15
Reduce logs in case of down bookies: re-replicator and shell commands

eolivelli:
### Motivation

When a bookie is down you can see many error both on command line tools and during replicator work.

### Changes

Reduce the log level, and save printing useless stacktraces


eolivelli:
github messed up with a rebase. closing this PR

PR #2538, opened 2021-01-15
Fix BookKeeper shell "list bookies" in case of downbookie and reduce noisy log during re-replication in case of down bookie

eolivelli:
### Motivation

When a bookie is down you can see many error both on command line tools and during replicator work.

Also the CLI is not working at all

```

This is the error on "bookkeeper shell listbookies" when a bookie is downb
org.apache.bookkeeper.client.BKException$BKBookieHandleNotAvailableException: Bookie handle is not available
	at org.apache.bookkeeper.discover.ZKRegistrationClient.getBookieServiceInfo(ZKRegistrationClient.java:248) ~[org.apache.bookkeeper-bookkeeper-server-4.12.0.jar:4.12.0]
	at org.apache.bookkeeper.client.DefaultBookieAddressResolver.resolve(DefaultBookieAddressResolver.java:43) [org.apache.bookkeeper-bookkeeper-server-4.12.0.jar:4.12.0]
	at org.apache.bookkeeper.tools.cli.helpers.CommandHelpers.getBookieSocketAddrStringRepresentation(CommandHelpers.java:45) [org.apache.bookkeeper-bookkeeper-server-4.12.0.jar:4.12.0]
	at org.apache.bookkeeper.tools.cli.commands.bookies.ListBookiesCommand.printBookies(ListBookiesCommand.java:124) [org.apache.bookkeeper-bookkeeper-server-4.12.0.jar:4.12.0]
	at org.apache.bookkeeper.tools.cli.commands.bookies.ListBookiesCommand.run(ListBookiesCommand.java:113) [org.apache.bookkeeper-bookkeeper-server-4.12.0.jar:4.12.0]
	at org.apache.bookkeeper.tools.cli.commands.bookies.ListBookiesCommand.run(ListBookiesCommand.java:43) [org.apache.bookkeeper-bookkeeper-server-4.12.0.jar:4.12.0]
	at org.apache.bookkeeper.tools.cli.helpers.DiscoveryCommand.apply(DiscoveryCommand.java:57) [org.apache.bookkeeper-bookkeeper-server-4.12.0.jar:4.12.0]
	at org.apache.bookkeeper.tools.cli.helpers.ClientCommand.apply(ClientCommand.java:60) [org.apache.bookkeeper-bookkeeper-server-4.12.0.jar:4.12.0]
	at org.apache.bookkeeper.bookie.BookieShell$ListBookiesCmd.runCmd(BookieShell.java:1202) [org.apache.bookkeeper-bookkeeper-server-4.12.0.jar:4.12.0]
	at org.apache.bookkeeper.bookie.BookieShell$MyCommand.runCmd(BookieShell.java:236) [org.apache.bookkeeper-bookkeeper-server-4.12.0.jar:4.12.0]
	at org.apache.bookkeeper.bookie.BookieShell.run(BookieShell.java:2235) [org.apache.bookkeeper-bookkeeper-server-4.12.0.jar:4.12.0]
	at org.apache.bookkeeper.bookie.BookieShell.main(BookieShell.java:2326) [org.apache.bookkeeper-bookkeeper-server-4.12.0.jar:4.12.0]

```

### Changes

Reduce the log level, and save printing useless stacktraces:
- CommandsHelpers -> prevent error on the CLI in case of down bookie (this is a blocker for the usage of the CLI)
- PerChannelBookieClient -> any time you try to connect to a down bookie
- DefaultBookieAddressResolver -> duplicate log line, the error already bubbles up on the caller

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/DefaultBookieAddressResolver.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/DefaultBookieAddressResolver.java
@@ -59,13 +59,13 @@ public class DefaultBookieAddressResolver implements BookieAddressResolver {
                 log.debug("Resolving dummy bookie Id {} using legacy bookie resolver", bookieId);
                 return BookieSocketAddress.resolveDummyBookieId(bookieId);
             }
-            log.info("Cannot resolve {}, bookie is unknown", bookieId, ex);
+            log.debug("Cannot resolve {}, bookie is unknown", bookieId, ex.toString());
             throw new BookieIdNotResolvedException(bookieId, ex);
         } catch (Exception ex) {
             if (ex instanceof InterruptedException) {
                 Thread.currentThread().interrupt();
             }
-            log.info("Cannot resolve {} ", bookieId, ex);
+            log.debug("Cannot resolve {} ", bookieId, ex);
             throw new BookieIdNotResolvedException(bookieId, ex);
         }
     }
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/DefaultBookieAddressResolver.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/DefaultBookieAddressResolver.java
@@ -59,7 +59,7 @@ public class DefaultBookieAddressResolver implements BookieAddressResolver {
                 log.debug("Resolving dummy bookie Id {} using legacy bookie resolver", bookieId);
                 return BookieSocketAddress.resolveDummyBookieId(bookieId);
             }
-            log.debug("Cannot resolve {}, bookie is unknown", bookieId, ex.toString());
+            log.info("Cannot resolve {}, bookie is unknown", bookieId, ex.toString());
             throw new BookieIdNotResolvedException(bookieId, ex);
         } catch (Exception ex) {
             if (ex instanceof InterruptedException) {
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/DefaultBookieAddressResolver.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/DefaultBookieAddressResolver.java
@@ -65,7 +65,6 @@ public class DefaultBookieAddressResolver implements BookieAddressResolver {
             if (ex instanceof InterruptedException) {
                 Thread.currentThread().interrupt();
             }
-            log.debug("Cannot resolve {} ", bookieId, ex);
             throw new BookieIdNotResolvedException(bookieId, ex);
         }
     }
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
@@ -542,7 +542,7 @@ public class PerChannelBookieClient extends ChannelInboundHandlerAdapter {
         try {
             addr = bookieAddressResolver.resolve(bookieId);
         } catch (BookieAddressResolver.BookieIdNotResolvedException err) {
-            LOG.error("Cannot connect to {} as endpopint resolution failed", bookieId, err);
+            LOG.error("Cannot connect to {} as endpopint resolution failed (probably bookie is down)", bookieId, err.toString());
             return processBookieNotResolvedError(startTime, err);
         }
 
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
@@ -542,7 +542,8 @@ public class PerChannelBookieClient extends ChannelInboundHandlerAdapter {
         try {
             addr = bookieAddressResolver.resolve(bookieId);
         } catch (BookieAddressResolver.BookieIdNotResolvedException err) {
-            LOG.error("Cannot connect to {} as endpopint resolution failed (probably bookie is down)", bookieId, err.toString());
+            LOG.error("Cannot connect to {} as endpoint resolution failed (probably bookie is down)",
+                    bookieId, err.toString());
             return processBookieNotResolvedError(startTime, err);
         }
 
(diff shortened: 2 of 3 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/DefaultBookieAddressResolver.java
```diff
@@ -59,13 +59,13 @@ public BookieSocketAddress resolve(BookieId bookieId) {
                 log.debug("Resolving dummy bookie Id {} using legacy bookie resolver", bookieId);
                 return BookieSocketAddress.resolveDummyBookieId(bookieId);
             }
-            log.info("Cannot resolve {}, bookie is unknown", bookieId, ex);
+            log.debug("Cannot resolve {}, bookie is unknown", bookieId, ex.toString());
```
fpj, line 62: It is important to know of an address resolution error, I'd not make this `debug`, I'd say it is at least `info` if not `warn`.

Also, removing the stack traces certainly makes sense for the shell, but they might be useful for general debugging.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/DefaultBookieAddressResolver.java
```diff
@@ -59,13 +59,13 @@ public BookieSocketAddress resolve(BookieId bookieId) {
...
             throw new BookieIdNotResolvedException(bookieId, ex);
         } catch (Exception ex) {
             if (ex instanceof InterruptedException) {
                 Thread.currentThread().interrupt();
             }
-            log.info("Cannot resolve {} ", bookieId, ex);
+            log.debug("Cannot resolve {} ", bookieId, ex);
```
fpj, line 68: Same here.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/DefaultBookieAddressResolver.java
eolivelli: that was my thought when I first wrote this code, but actually we are appending the exception to the chain of the thrown BookieIdNotResolvedException, so we are going to report this error twice (here and in the caller).
I can leave it at "info" level, but the stacktrace is redundant this is case

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/DefaultBookieAddressResolver.java
eolivelli: we are appending the exception to the chain of the thrown BookieIdNotResolvedException, so we are going to report this error twice (here and in the caller).

fpj: In this case here, we would be logging messages like the one of line 47 above at debug level, but at the same time we are rethrowing, so it is possible that this log message here is unnecessary altogether.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
```diff
@@ -542,7 +542,8 @@ protected ChannelFuture connect() {
         try {
             addr = bookieAddressResolver.resolve(bookieId);
         } catch (BookieAddressResolver.BookieIdNotResolvedException err) {
-            LOG.error("Cannot connect to {} as endpopint resolution failed", bookieId, err);
+            LOG.error("Cannot connect to {} as endpoint resolution failed (probably bookie is down)",
```
fpj, line 545: I'm a bit confused about the `probably bookie is down` part. Why resolving the address requires the bookie to be up? I'm probably missing something basic.

eolivelli: this "resolution" is about looking up the "current" Bookie network address on ZooKeeper, it is not DNS resolution.

When the bookie is up, it advertises all of its open endpoints (bookie-rpc,http....) on ZK, in the same znode that we are using to detect if the bookie is writable/readonly, it is an ephemeral znode.

When you want to connect to a bookie, you start from a BookieID (an opaque string), you lookup the current network address on ZK (we have a cache of course) and then you are good to go.

but if the Bookie is down the ephemeral znode does not exist and the client does not know the network address to connect to.

Usually when you see this log line the problem is that you cannot connect to the bookie because you do not know the network address, because it is not advertising it, and this usually happens when the bookie is down. 

for reference:
http://bookkeeper.apache.org/bps/BP-41-bookieid/
http://bookkeeper.apache.org/bps/BP-38-bookie-endpoint-discovery/

eolivelli: I can remove it, but I find it quite useful in the day to day support to people looking at this kind of logs

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/DefaultBookieAddressResolver.java
fpj: Quick question: since we are re-throwing, should we simply eliminate this log message altogether and let it be logged upstream?

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
fpj: Could this be a ZK error too, like ZK is down and it can't talk to ZK? If so, then perhaps add it to the message. Otherwise, it is fine to leave it there.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/DefaultBookieAddressResolver.java
eolivelli: Right. I will remove this line at all.
Less noise is better while debugging

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
eolivelli: That exception means that we currently cannot resolve  the is.
So I am keeping the message, according to your answer.
Basically in this API we are throwing that Exception instead of returning null.

Beside note: we are not performing blocking accesses to ZK for this operation, we are always reading from an in-memory cache

eolivelli:
@fpj  I have updated the patch as we said yesterday

fpj (approved):

Issue #2505, opened 2020-12-02
NullPointerException during the stop of a BookKeeper with BookKeeper Stream Storage Server


eolivelli:
**BUG REPORT**

***Describe the bug***

When you stop a Bookie with the StreamStorage server enabled you can see a scary NullPointerException.

We should avoid to throw the NPE, that is usually the symptom of a bug

***To Reproduce***

1. Start a bookie with `extraServerComponents=org.apache.bookkeeper.stream.server.StreamStorageLifecycleComponent`
2. Shutdown the bookie

You can see this error:

```
2020-12-02 11:04:24,962 - ERROR - [component-shutdown-thread:ComponentStarter$ComponentShutdownHook@51] - Failed to close component bookie-server in shutdown hook gracefully, Exiting anyway
java.lang.NullPointerException
	at org.apache.bookkeeper.stream.server.service.RegistrationStateService.doClose(RegistrationStateService.java:107)
	at org.apache.bookkeeper.common.component.AbstractLifecycleComponent.close(AbstractLifecycleComponent.java:123)
	at org.apache.bookkeeper.common.component.LifecycleComponentStack.lambda$close$6(LifecycleComponentStack.java:154)
	at com.google.common.collect.ImmutableList.forEach(ImmutableList.java:405)
	at org.apache.bookkeeper.common.component.LifecycleComponentStack.close(LifecycleComponentStack.java:154)
	at org.apache.bookkeeper.stream.server.StreamStorageLifecycleComponent.doClose(StreamStorageLifecycleComponent.java:59)
	at org.apache.bookkeeper.common.component.AbstractLifecycleComponent.close(AbstractLifecycleComponent.java:123)
	at org.apache.bookkeeper.common.component.LifecycleComponentStack.lambda$close$6(LifecycleComponentStack.java:154)
	at com.google.common.collect.ImmutableList.forEach(ImmutableList.java:405)
	at org.apache.bookkeeper.common.component.LifecycleComponentStack.close(LifecycleComponentStack.java:154)
	at org.apache.bookkeeper.common.component.ComponentStarter$ComponentShutdownHook.run(ComponentStarter.java:47)
	at java.base/java.lang.Thread.run(Thread.java:832)
```

rudy2steiner:
I'll work on it

PR #2542, opened 2021-01-16
Check np for LifecycleComponent close

rudy2steiner:

 Check NP of  BookKeeper Stream Storage Server LifecycleComponents

### Motivation

Fix issue #2505 
 
### Changes

Master Issue: #2505

### DIFF
--- a/stream/server/src/main/java/org/apache/bookkeeper/stream/server/service/RegistrationStateService.java
+++ b/stream/server/src/main/java/org/apache/bookkeeper/stream/server/service/RegistrationStateService.java
@@ -95,6 +95,10 @@ public class RegistrationStateService
 
     @Override
     protected void doStop() {
+        if (null == stateManager) {
+            log.warn("State Manager is null, no need to stop it.");
+            return;
+        }
         stateManager.forceToShuttingDown();
 
         // turn the server to readonly during shutting down process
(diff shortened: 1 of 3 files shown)


eolivelli (approved):
LGTM
@Ghatage  @sijie @rdhabalia @fpj PTAL

this is a good candidate for 4.12.2, we have a few little changes that are helping in improving the user experience, especially during "problems"

stream/server/src/main/java/org/apache/bookkeeper/stream/server/service/RegistrationStateService.java
```diff
@@ -95,6 +95,10 @@ protected void doStart() {
 
     @Override
     protected void doStop() {
+        if (null == stateManager) {
+            log.warn("State Manager is null, no need to stop it.");
```
fpj, line 99: A warn message for me indicates a potential action from the user. In this case, it is just information that the state manager is null, so no action is required, correct? If so, then it is an info message at best, if not debug.

rudy2steiner: yes

eolivelli:
@fpj 's has been addresses, I merged the patch

thank you @rudy2steiner

PR #2543, opened 2021-01-18
Allow to configure Prometheus HTTP Server bind address

Shoothzj:

make prometheus bind addr can cofigure

### Motivation

For security, make prometheus bind addr can config.
For compatibility, make default value is 0.0.0.0

### Changes

make prometheus bind addr can cofigure

### DIFF
--- a/bookkeeper-stats-providers/prometheus-metrics-provider/src/main/java/org/apache/bookkeeper/stats/prometheus/PrometheusMetricsProvider.java
+++ b/bookkeeper-stats-providers/prometheus-metrics-provider/src/main/java/org/apache/bookkeeper/stats/prometheus/PrometheusMetricsProvider.java
@@ -63,7 +63,7 @@ public class PrometheusMetricsProvider implements StatsProvider {
     public static final String PROMETHEUS_STATS_HTTP_ENABLE = "prometheusStatsHttpEnable";
     public static final boolean DEFAULT_PROMETHEUS_STATS_HTTP_ENABLE = true;
 
-    public static final String PROMETHEUS_STATS_HTTP_ADDR = "prometheusStatsHttpAddr";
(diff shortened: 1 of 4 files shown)


bookkeeper-stats-providers/prometheus-metrics-provider/src/main/java/org/apache/bookkeeper/stats/prometheus/PrometheusMetricsProvider.java
```diff
@@ -63,6 +63,9 @@
     public static final String PROMETHEUS_STATS_HTTP_ENABLE = "prometheusStatsHttpEnable";
     public static final boolean DEFAULT_PROMETHEUS_STATS_HTTP_ENABLE = true;
 
+    public static final String PROMETHEUS_STATS_HTTP_ADDR = "prometheusStatsHttpAddr";
```
eolivelli, line 66: can we name this 'prometheusStatsHttpAddress' ?

eolivelli (changes requested):
awesome !
can you please also add a test ?

it would be super good to follow up and publish this value (and the port) to the BookieEndpointInfo,
this way tools (like BKVM) could access this value
see
https://bookkeeper.apache.org/bps/BP-38-bookie-endpoint-discovery/

cc @dianacle

bookkeeper-stats-providers/prometheus-metrics-provider/src/main/java/org/apache/bookkeeper/stats/prometheus/PrometheusMetricsProvider.java
Shoothzj: Fixed

sijie (approved):

eolivelli (approved):
LGTM

PR #2570, opened 2021-02-01
Fix the current site to allow click links in doc

hsaputra:
Since there will be some time before we get site2 up and running, sending
this fix to allow links in doc content ot be clickable.

Currently, if user go to https://bookkeeper.apache.org/docs/latest/ the links
in the doc can not be click since the div overlay of navigator on left is
covering the content body.

The simple proposed fix is to set width of navigator container class to
certain width to not cover the body content.

To verify behavior and proposed fix difference:

1. Go to https://bookkeeper.apache.org/docs/latest/ and try to click the `WAL` link. You will not see the cursor change to hand icon to click the link
2. Now, visit the proposed fix at https://hsaputra.github.io/bookkeeper-staging-site/docs/latest/, then when  you hover on top of `WAL` link you should see the cursor change to hand point and allow you to click the link and navigate to the URL.

### DIFF
--- a/site/README.md
+++ b/site/README.md
@@ -6,8 +6,6 @@ Welcome to the code for the Apache BookKeeper website! Instructions on building 
 
 The site is built using Jekyll and Sass.
 
-> I'll provide more specific info here later.
-
 ## Prerequisities
 
 In order to run the site locally, you need to have the following installed:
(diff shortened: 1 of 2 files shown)


eolivelli (approved):
Cool!
Thanks.
I will merge and build the website and push it manually (currently the automatic procedure is broken see my messages on dev mailing list)

hsaputra:
Sounds, good @eolivelli ! Thanks for quick review.

This is not fancy fix since I am not full time web developer =P
But this should help making BK site more usable until we have next gen project site up and running.

Issue #2588, opened 2021-02-09
Clarify getAllBookies javadoc


michaeljmarshall:
It looks like the javadoc for `getWritableBookies` was copied to `getAllBookies`. I'm sure that `getAllBookies` includes all bookies, not just writable ones. However, I would like to clarify what "all bookies" actually includes. Does it include bookies that have shutdown and be unavailable but are not yet decommissioned? In other words, what triggers a bookie to be removed from the list of all bookies? Is it on shutdown or on decommission?

Here's the javadoc:

https://github.com/apache/bookkeeper/blob/67a02db73d62188fc8a143bd9a37038ae770e90a/bookkeeper-server/src/main/java/org/apache/bookkeeper/discover/RegistrationClient.java#L51-L63

I'm happy to submit a PR to update the javadoc, if someone can clarify the intended meaning for me. Thanks!

PR #2589, opened 2021-02-09
Fix javadoc for getAllBookies in RegistrationClient interface

michaeljmarshall:

Basic javadoc update from what looks to be an accidental copy from another javadoc.

I opened https://github.com/apache/bookkeeper/issues/2588 to clarify what the meaning of "all" is. Either way, I'm sure we just need to update the javadoc as I did in this PR.

If https://github.com/apache/bookkeeper/issues/2588 reveals something interesting, it might be worth updating this javadoc to include that information.


eolivelli (approved):
LGTM

thank you @michaeljmarshall

PR #2587, opened 2021-02-08
Fix hanging tests tests TestStorageServerClientManagerImpl

pkumar-singh:
Descriptions of the changes in this PR:

**Motivation**
Fixing hanging tests TestStorageServerClientManagerImpl and other tests in the directory

Some tests were not running because they were hanging waiting on future while storage container was being fetched.
But inproc grpc server was running on endpoint fake-server while client was connecting on fake-server:4181

**Changes**
Make GRPC inproc server run on fake-server:4181 and client connect on fake-server:4181

PR #2590, opened 2021-02-09
CI - Add streamstorage tests

eolivelli:
### Motivation

Stream Storage tests now are back in good shape again, we must enable verification on CI

### Changes

Add StreamStorage tests on master branch.
Tests do not pass on olde branches and there is no way to pick all of the fixes without breaking changes (like important dependencies upgrades).


eolivelli:
This patch is to be committed after https://github.com/apache/bookkeeper/pull/2587

eolivelli:
I used the wrong base branch

PR #2605, opened 2021-02-17
Fix logger member not correct、some javadoc  problem

Shoothzj:
Fix java doc problem

1. Some log in slf4j, the pattern number and the member log not equal
2. Some javadoc grammar problem

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
@@ -269,9 +269,9 @@ public class RackawareEnsemblePlacementPolicyImpl extends TopologyAwareEnsembleP
                 }
             } catch (RuntimeException re) {
                 if (!conf.getEnforceMinNumRacksPerWriteQuorum()) {
-                    LOG.error("Failed to initialize DNS Resolver {}, used default subnet resolver : {}",
+                    LOG.error("Failed to initialize DNS Resolver {}, used default subnet resolver : {} {}",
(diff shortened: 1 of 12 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
```diff
@@ -269,9 +269,9 @@ public RackawareEnsemblePlacementPolicyImpl initialize(ClientConfiguration conf,
                 }
             } catch (RuntimeException re) {
                 if (!conf.getEnforceMinNumRacksPerWriteQuorum()) {
-                    LOG.error("Failed to initialize DNS Resolver {}, used default subnet resolver : {}",
+                    LOG.error("Failed to initialize DNS Resolver {}, used default subnet resolver : {} {}",
```
dlg99, line 272: i'd just flip the order of `re` and  `re.getMessage()` on the next line and leave an exception without matching `{}` to let logger print the stack. Or remove `re.getMessage` altogether with both `{}` at the end

dlg99 (commented):
LGTM, one nit to fix

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
Shoothzj: I remeber only one exception args can print without matching `{}`, I remove the re.getMessage.

Shoothzj:
@eolivelli PTAL, thanks

merlimat (approved):

PR #2609, opened 2021-02-17
Fix docker image build - add python3

### LINKED ISSUE #2601: There is no DockerHub image for 4.12.1
**BUG REPORT**

***Describe the bug***

There is no DockerHub image for 4.12.1

***To Reproduce***

1. Go to https://hub.docker.com/r/apache/bookkeeper/tags?page=1&ordering=last_updated&name=4.12
2. Observe there is no 4.12.1 image

***Expected behavior***

There is a 4.12.1 image

eolivelli:
### Motivation
Building the docker image does not work anymore, see #2601 

### Changes
Add python3 to the list of available packages

## How to verify
cd docker
docker build --build-arg BK_VERSION=4.12.1 -t testbk .

Fixes #2601

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -33,7 +33,7 @@ ENV JAVA_HOME=/usr/lib/jvm/java-11
 # Download Apache Bookkeeper, untar and clean up
 RUN set -x \
     && adduser "${BK_USER}" \
-    && yum install -y java-11-openjdk-devel wget bash python sudo\
+    && yum install -y java-11-openjdk-devel wget bash python3 sudo\
     && mkdir -pv /opt \
     && cd /opt \
     && wget -q "${DISTRO_URL}" \
@@ -48,6 +48,7 @@ RUN set -x \
     && rm -rf "$DISTRO_NAME.tar.gz" "$DISTRO_NAME.tar.gz.asc" "$DISTRO_NAME.tar.gz.sha512" \
     # install zookeeper shell
     && wget -q https://bootstrap.pypa.io/2.7/get-pip.py \
+    && python --version \
     && python get-pip.py \
     && pip install zk-shell \
     && rm -rf get-pip.py \


eolivelli:
Honestly I don't know why changing to python3 fixes the build, we need Python 2.7 in order to run zookeeper-shell.

Probably it is the pip2.7 installer that now needs python3 ?

docker/Dockerfile
```diff
@@ -48,6 +48,7 @@ RUN set -x \
     && rm -rf "$DISTRO_NAME.tar.gz" "$DISTRO_NAME.tar.gz.asc" "$DISTRO_NAME.tar.gz.sha512" \
     # install zookeeper shell
     && wget -q https://bootstrap.pypa.io/2.7/get-pip.py \
+    && python --version \
```
zzzming, line 51: If python3 is installed, you should update pip 3 too at line 50.

zzzming (changes requested):

docker/Dockerfile
eolivelli: This line returns "2.7"

I tried to move to pip3, but I need pip 2 because we need to install zookeeper-shell and with pip3 it fails

dlg99 (approved):
LGTM

michaeljmarshall:
@eolivelli - as far as while the docker image build broke, I think this stack overflow is relevant:  https://stackoverflow.com/questions/65896334/python-pip-broken-wiith-sys-stderr-writeferror-exc. (You posted a pretty similar stack trace in https://github.com/apache/bookkeeper/issues/2601.)

docker/Dockerfile
michaeljmarshall: I wonder why it fails. I was able to run `pip3 install zk-shell` successfully from inside the latest `python` docker image. Given that python 2.7 is deprecated, I think it might be worth creating an issue to upgrade the version of python shipped in the docker image.

zzzming: LGTM

eolivelli:
@michaeljmarshall thanks for your pointer

michaeljmarshall:
@eolivelli - I think part of the problem is the centos:7 image. It ships with python 2.7 as the `python` version. I think it might make sense to fully upgrade to python 3 and to pip 3, given that python 2.7 reached its end of life. Is there any reason I shouldn't look into a PR for this? Based on a quick prototype, I think the centos:8 image might be better for python 3. Is there any reason to stay on centos:7? Thanks!

Issue #2614, opened 2021-02-20
Existing fencing not enough to prevent data loss


Vanlightly:
I have implemented the BookKeeper replication protocol in TLA+. You can find the specification and a readme here: https://github.com/Vanlightly/bookkeeper-tlaplus

The TLA+ model checker found a data loss scenario related to how fencing is used to protect split-brain and data loss.

**Protocol Defect**

The BookKeeper protocol defect found with this spec is that fencing of LAC reads is not enough to prevent committing an add. The below example involves the loss of a fencing request. However, the same is achievable if one fencing LAC req is delivered after a Recovery Read Req, for example due to some network delay and a connection failure between the requests (I have not verified if that is possible though).

Paste into https://sequencediagram.org/ or use link https://sequencediagram.org/index.html#initialData=A4QwTgLglgxloDsIAIDuBGAUKSt4iTQCZtxo5EUAjLHc-QqkuvS5KgZlNwoJQC8A1pgwBaUQA8aALmQBTAAwiiogHxDZAW3CDkIAM7IoCZGDkwA9gDc5YAJ7K1M5ADM5CGHMw01qIrIAbEBhkAF5kUSwxVSZZRW8VVQw4hTQwKAgId0cY-1d3TwTfPKCQ8MjHSU5ZNw8vBAss5GtbNHQAGg18uuR9AFcYT31DCxN0ZAaAEzlkYFbg3QBHPoswPs0AHQRSo0NIrYAxArlJ2QBvGnamAF925ABRJHS5fVkAHjeaZGlVZDPb9hEb6-M6KAGcYF-a6qVQ5ZxmECTeRKPxOPIIpHxVExDiyDHI7zoYqyAByFgAygMABaPCD2bwcYnIMmUmA0p4ObFdGABCz6GYBE4Ac3mhjkmmAEE5RJxslQ6Sa8U4vnQKTSGSyCBEMuSehgggaqEFkxFyK2yAgFmQMBAAUFYEwDSaLTAbU6glkAEF9YbjSKkfKMjNFLt5C43DBoDYAnZkLz9CgDEYUKgk3S+h4QFkkVRY5o5BBEVmQEA

```
participant w1
participant w2
participant b1
participant b2
participant b3
participant zk
w1--xb1: e0
w2->zk: mark as in recovery
w2->b1: fence
b1->w2: lac = -1
w1->b2: e0
b2->w1: e0 written
w2->b2: fence
b2->w2: lac = -1
[...]

Issue #2612, opened 2021-02-18
Wrong ReadLastAddConfirmed logic that can lead to data loss in client applications


RaulGracia:
**BUG REPORT**

***Describe the bug***

We have found a bug when invoking `ReadLastAddConfirmed` that may lead the Bookkeeper client to return a wrong last entry or no entries at all when querying the Bookies. This behavior goes against of what is stated in the Javadoc, as it does not return the maximum entry id for a ledger within an ensemble: https://github.com/apache/bookkeeper/blob/67a02db73d62188fc8a143bd9a37038ae770e90a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/ReadHandle.java#L116
```
Obtains asynchronously the last confirmed write from a quorum of bookies. This call obtains the the last add confirmed each bookie has received for this ledger and returns the maximum.
```
Applications that rely on this method to return the actual last valid entry in a ledger may wrongly believe that a ledger is empty and discard it (e.g., rollover), thus losing data.

This is a real reproduction of the problem. Lets consider an ensemble with 3 Bookies (`pravega-bk-bookie-0`, `pravega-bk-bookie-1`, and `pravega-bk-bookie-2` in the logs) where `ensembleSize=writeQuorum=ackQuorum=3`. After storing some data in the system we induce a destructive restart in `pravega-bk-bookie-2`, meaning that its data is wiped out (but it's identity is still the same). Then, our application tries to recover its internal state reading the data stored in Bookkeeper, and the problem manifests as follows:

1. Our application tries to read from ledger 5 to recover its internal state after the failure induced in `pravega-bk-bookie-2`:
```
2021-02-17 12:29:47,741 2251502 [Thread-10239] INFO  i.p.s.server.logs.DurableLog - DurableLog[1]: Starting.
```
2. To this end, the first thing it does is to ask the Bookkeeper client for the last entry of ledger 5 to start its internal recovery process with the data stored in Bookkeeper. As you can see, the metadata is correctly fetched from Zookeeper for ledger 5:
```
2021-02-17 12:29:47,750 2251511 [Thread-10234-EventThread] DEBUG o.a.b.client.MetadataUpdateLoop - UpdateLoop(ledgerId=5,loopId=6425f4c2) success
```

3. We observe parallel reads issued to the 3 Bookies in the ensemble:
```
[...]

PR #2616, opened 2021-02-20
Add fencing to recovery reads to avoid data loss issue

Vanlightly:

### Motivation

Adding the fencing flag to recovery reads avoids a data loss scenario as described in [issue 2614](https://github.com/apache/bookkeeper/issues/2614)

### Changes

Added the fencing flag to recovery reads. Refactored some mocking unit test code to introduce fencing and allow two writers to share the same bookie state. Added a new unit to verify the fix. You can recreate the initial problem by removing the fencing flag from recovery reads and running the new unit test.

Master Issue: #2614

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
@@ -568,9 +568,14 @@ class PendingReadOp implements ReadEntryCallback, SafeRunnable {
             lh.throttler.acquire();
         }
 
-        int flags = isRecoveryRead ? BookieProtocol.FLAG_HIGH_PRIORITY : BookieProtocol.FLAG_NONE;
-        clientCtx.getBookieClient().readEntry(to, lh.ledgerId, entry.eId,
-                                              this, new ReadContext(bookieIndex, to, entry), flags);
+        if (isRecoveryRead) {
+            int flags = BookieProtocol.FLAG_HIGH_PRIORITY | BookieProtocol.FLAG_DO_FENCING;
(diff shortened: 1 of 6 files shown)


eolivelli (approved):
LGTM 

@dlg99 please take a look, probably it is worth to pick this fix for 4.13.0

eolivelli:
@Vanlightly PR validation failed. PTAL 


Thank you very much for sharing your findings and also provide a fix

dlg99 (approved):
@Vanlightly it is an awesome catch and the fix lgtm.
One note about the test: often reliance on Thread.sleep() leads to test flakiness. 
Can I ask you to try to remove these and replace with wait for completion of specific events?

dlg99:
Another note: NoSuchLedger/NoSuchEntry have special handling in PendingReadLacOp/RoundRobinDistributionSchedule, see https://github.com/apache/bookkeeper/issues/2612
Is it possible to repro case when at lac N write N+1 is lost?

Vanlightly:
Yes of course. I'll take a look at both suggestions.

Vanlightly:
@dlg99 I've replaced the Thread.sleep with CompletableFutures to ensure the right sequence is followed in a non-flaky way.

Regarding the issue #2612, I've left a comment there. Basically the protocol is not designed to handle arbitrary failures of that kind and data loss can result. In terms of this particular issue, which is not an arbitrary failure scenario, I think the current test is probably enough.

dlg99:
@Vanlightly Thank you! The changes lgtm.

eolivelli:
pushed to master branch and to branch-4.12.

thank you very much @Vanlightly 

I believe you are now unblocked in sending the next PR

Issue #2622, opened 2021-02-24
bookie_journal_JOURNAL_CB_QUEUE_SIZE grows unbounded


michaeljmarshall:
**BUG REPORT**

***Describe the bug***

The `bookie_journal_JOURNAL_CB_QUEUE_SIZE` grows unbounded.

***To Reproduce***

1. I'm running pulsar 2.7.0 deployed with bookkeeper 4.12.0.
2. As entries are added, this metric increases, and does not show any signs of decreasing.
3. I'm seeing this through prometheus, but in my brief investigation, it appears to be fundamentally tied to the implementation.

***Expected behavior***

I expect the size of a queue to go up and down, not to increase forever, especially because I know this queue is bounded. This queue's size should be similar to `bookie_journal_JOURNAL_FORCE_WRITE_QUEUE_SIZE` and `bookie_journal_JOURNAL_SIZE`, given that they're measuring similar things.

***Additional context***

In briefly looking at the code, I can see that we never call `dec` on the `journalCbQueueSize` metric, and I can also see that we call `inc`. As such, the exhibited behavior makes sense. I'm going to try to put together a PR tonight to propose a fix.

Note also that I can see on line 871 and 880 below, we pass in the `journalStats.getJournalQueueSize`

https://github.com/apache/bookkeeper/blob/26120e0436d0076bf7990b5cea1e41a3e287af04/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java#L867-L881

And yet, the parameter for that argument is named `journalCbQueueSize` (as is the variable it gets assigned to), as seen on lines 313 and 324:

https://github.com/apache/bookkeeper/blob/26120e0436d0076bf7990b5cea1e41a3e287af04/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java#L311-L325

michaeljmarshall:
I noticed this issue while digging into https://github.com/apache/bookkeeper/issues/2619.

PR #2623, opened 2021-02-24
Issue 2622: Fix usage of JournalQueueSize and JournalCbQueueSize; JournalCbQueueSize should be bounded

michaeljmarshall:

Fixes https://github.com/apache/bookkeeper/issues/2622.

This PR only affects metrics. Specifically, it improves the usage of `JournalQueueSize` and `JournalCbQueueSize` (including `bookie_journal_JOURNAL_QUEUE_SIZE` and `bookie_journal_JOURNAL_CB_QUEUE_SIZE`).

### Motivation

As mentioned in the issue, `bookie_journal_JOURNAL_CB_QUEUE_SIZE` grows unbounded. This PR looks to fix that, and in doing so, fixes `bookie_journal_JOURNAL_QUEUE_SIZE` as well.

### Changes

The `bookie_journal_JOURNAL_CB_QUEUE_SIZE ` grows unbounded because it is never decremented. Ideally, it would measure the number of call backs outstanding for the Journal. I've updated the `Counter` object that is passed into the `QueueEntry` class so that the `Counter` is the one for the call back queue size. In changing that code, I also needed to update the way the `JournalQueueSize` was handled. I updated the code so that it gets incremented when adding elements to the queue and decremented when elements are removed. The journal queue size should represent the size of the actual `queue` variable in the `Journal` class.

### Note
If my understanding of any of these metrics is off base, please question me. I'm new to this class, and while I read through it, I might have missed something. Thanks!

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
@@ -864,11 +864,11 @@ public class Journal extends BookieCriticalThread implements CheckpointSource {
         entry.retain();
 
         journalStats.getJournalQueueSize().inc();
-        journalStats.getJournalCBQueueSize().inc();
+        journalStats.getJournalCbQueueSize().inc();
         queue.put(QueueEntry.create(
                 entry, ackBeforeSync,  ledgerId, entryId, cb, ctx, MathUtils.nowInNano(),
                 journalStats.getJournalAddEntryStats(),
-                journalStats.getJournalCBQueueSize()));
+                journalStats.getJournalCbQueueSize()));
     }
 
     void forceLedger(long ledgerId, WriteCallback cb, Object ctx) {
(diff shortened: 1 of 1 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -390,7 +390,6 @@ public int process(boolean shouldForceWrite) throws IOException {
                     if (qe != null) {
                         cbThreadPool.execute(qe);
                     }
-                    journalStats.getJournalCbQueueSize().inc();
```
michaeljmarshall, line 393: It's possible that we actually do want to increment the `JournalCbQueueSize` here, if we're looking to measure the number of call backs being run (or waiting to run on the `cbThreadPool`). The running of this `QueueEntry` will trigger the following:

```java
        public void run() {
            if (LOG.isDebugEnabled()) {
                LOG.debug("Acknowledge Ledger: {}, Entry: {}", ledgerId, entryId);
            }
            journalCbQueueSize.dec();
            journalAddEntryStats.registerSuccessfulEvent(MathUtils.elapsedNanos(enqueueTime), TimeUnit.NANOSECONDS);
            cb.writeComplete(0, ledgerId, entryId, null, ctx);
            recycle();
        }
```

The PR currently assumes that we're measuring the number of call backs in the queue until the call back is executed and decrements the metric.

dlg99 (approved):
LGTM

michaeljmarshall:
@sijie and @eolivelli PTAL, thanks!

eolivelli (approved):
+1

@sijie PTAL

merlimat (approved):

PR #2627, opened 2021-02-25
Allow DBLedgerStorage to force GC by disk listener

### LINKED ISSUE #2596: Enable `isForceGCAllowWhenNoSpace` for the DBLedgerStorage
**FEATURE REQUEST**

In some cases, users disabled the entry log compaction task to reduce the I/O load, this will also make the write latency more stable(I/O resource competition). I have checked the source code of the BookKeeper, I think the `isForceGCAllowWhenNoSpace` does not work for the DBLedgerStorage(the default value of Pulsar). Currently, the behavior of the `isForceGCAllowWhenNoSpace` is It is a disk listener to decide whether to start the compaction task while the disk of the Ledger almost full. It should be the same behavior when we call `gc` endpoint manually.

Currently, call `gc` endpoint only works while the entry topic compaction enabled. It's better to support force gc to allow to trigger gc while the entry log compaction disabled.

zymap:
---

Fixes #2596

*Motivation*

The isForceGCAllowWhenNoSpace does not work for the DBLedgerStorage.
Currently, the behavior of the isForceGCAllowWhenNoSpace is It is a
disk listener to decide whether to start the compaction task while
the disk of the Ledger almost full. 

*Modifications*

- Add dir listener for the SingleDirectoryDbledgerStorage

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
@@ -37,6 +37,8 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import java.util.concurrent.atomic.AtomicLong;
+
+import lombok.Getter;
 import org.apache.bookkeeper.bookie.GarbageCollector.GarbageCleaner;
 import org.apache.bookkeeper.bookie.stats.GarbageCollectorStats;
 import org.apache.bookkeeper.conf.ServerConfiguration;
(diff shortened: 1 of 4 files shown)


codelipenghui (approved):
LGTM

ping @sijie Please help review this PR, thanks.

zymap:
ping @sijie

sijie (approved):
@zymap is it possible to add a test case?

zymap:
It's a little difficult to trigger the disk listener. Let me take a deep look to check if there has any way to add a test.
I also found some issues in my PR. Let's hold this PR.

zymap:
It's difficult to test trigger the disk listener. I add a simple test for testing the listener number.

jiazhai (approved):

PR #2630, opened 2021-02-26
[doc] fix the no striping occurs condition to Qw=E

yangl:

### Motivation

the no striping occurs condition should Qw=E, not the Qw=Qa

### Changes

update the condition to Qw=E

### DIFF
--- a/site/docs/4.10.0/development/protocol.md
+++ b/site/docs/4.10.0/development/protocol.md
@@ -64,7 +64,7 @@ Entry | Write quorum
 4 | B1, B2, B3
 5 | B2, B3, B4
 
-There are only **E** distinct write quorums in any ensemble. If **Q<sub>w</sub>** = **Q<sub>a</sub>**, then there is only one, as no striping occurs.
+There are only **E** distinct write quorums in any ensemble. If **Q<sub>w</sub>** = **E**, then there is only one, as no striping occurs.
 
 ### Ack quorums
 
--- a/site/docs/4.11.0/development/protocol.md
+++ b/site/docs/4.11.0/development/protocol.md
@@ -64,7 +64,7 @@ Entry | Write quorum
 4 | B1, B2, B3
 5 | B2, B3, B4
 
-There are only **E** distinct write quorums in any ensemble. If **Q<sub>w</sub>** = **Q<sub>a</sub>**, then there is only one, as no striping occurs.
+There are only **E** distinct write quorums in any ensemble. If **Q<sub>w</sub>** = **E**, then there is only one, as no striping occurs.
 
 ### Ack quorums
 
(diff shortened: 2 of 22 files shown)


site/docs/4.10.0/development/protocol.md
```diff
@@ -64,7 +64,7 @@ Entry | Write quorum
 4 | B1, B2, B3
 5 | B2, B3, B4
 
-There are only **E** distinct write quorums in any ensemble. If **Q<sub>w</sub>** = **Q<sub>a</sub>**, then there is only one, as no striping occurs.
+There are only **E** distinct write quorums in any ensemble. If **Q<sub>w</sub>** = **E**, then there is only one, as no striping occurs.
```
michaeljmarshall, line 67: It seems to me that these two sentences should be combined to indicate that it is an either/or (exclusive) relationship between the first and the second. For example:

> There are only **E** distinct write quorums in any ensemble, unless **Q<sub>w</sub>** = **E**, then there is only one, as no striping occurs.

I'd be interested to know what one of the committers says though.

michaeljmarshall (commented):
Not technically a comment on your change, but in seeing the lines changed, I thought it was worth adding the comment. I'm not authoritative though, so happy for input from others.

site/docs/4.10.0/development/protocol.md
dlg99: @michaeljmarshall To the best of my understanding:
E >= Qw >= Qa. 
There is no striping in case of (E, Qw, Qa) ==  (3, ,3 ,3) or (3, 3, 2). 
There is striping in case of (5, 3, 3)

if Qw > Qa the client will ack the write when Qa writes completed while (Qw-Qa) possibly failed or still in progress; it does not affect striping.

If E > Qw, the write quorums will be spread across larger set of bookies (see RoundRobinDistributionSchedule)
I.e. for E = 3 and Qw = 2, entries will go to
e0: (b0, b1)
e1: (b1, b2)
e2: (b2, b0)
e3: (b0, b1)
etc.

so the unique stripes in this case are (b0, b1), (b1, b2), (b2, b0)

In case of Qw = 3 all writes go to (b0, b1, b2).
IIRC, reads will still "rotate" between bookies to distribute reads more evenly, like e0 will be read from b0 (fallback to b1, b2), e1 from b1 (fallback to b2, b0) etc 

Hope this helps

dlg99 (approved):
LGTM, thank you for fixing this

site/docs/4.10.0/development/protocol.md
michaeljmarshall: @dlg99 - thanks for your comment. I think what I wrote is consistent with your comment, please let me know if you think otherwise.

> There is no striping in case of (E, Qw, Qa) == (3, ,3 ,3) or (3, 3, 2).

In both cases above, E = Qw, and there is only 1 distinct write quorum.

> There is striping in case of (5, 3, 3)

In this case, E > Qw, and there are E distinct write quorums.

My only point is that I think the documentation would be clearer as a single sentence (with an obvious either/or relationship) instead of two.

No worries if we leave it.

PR #2632, opened 2021-03-01
Fix NetworkTopologyImpl#getLeaves returning set with null value in case of non existing scope

mkozioro:

NetworkTopologyImpl#getLeaves(String scope) was returning set with null value, when requested scope did not exist.
I've added null check in NetworkTopologyImpl#doGetLeaves(String scope), used by NetworkTopologyImpl#getLeaves method to not have null node returned as leaf node.

It would be nice to have this merged also with non master branch of bookeeper so maintenance releases could have that fix.

### Motivation
I've found my Apache Pulsar cluster polluted with log:
```
 12:01:46.462 pulsar2-dev pulsar-broker {"logLevel":"ERROR","logThread":"pulsar-io-23-1","logger":"org.apache.bookkeeper.client.RackawareEnsemblePlacementPolicyImpl","message":"found non-BookieNode: null as leaf of defaultrack: /default-rack","stack_trace":null}
```

After digging I've found that, because https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java#L324
was returning HashSet with null node, when no bookkeepers were in /default-rack.

### Changes
- Modified doGetLeaves() method to return empty HashSet instead of HashSet with null element when scope node can't be found.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/net/NetworkTopologyImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/net/NetworkTopologyImpl.java
@@ -754,6 +754,10 @@ public class NetworkTopologyImpl implements NetworkTopology {
     private Set<Node> doGetLeaves(String scope) {
         Node node = getNode(scope);
         Set<Node> leafNodes = new HashSet<Node>();
+        if (node == null) {
+            return leafNodes;
+        }
+
         if (!(node instanceof InnerNode)) {
             leafNodes.add(node);
         } else {
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/net/NetworkTopologyImplTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/net/NetworkTopologyImplTest.java
@@ -1,5 +1,4 @@
-/**
- *
+/*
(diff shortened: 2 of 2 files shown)


eolivelli:
@mkozioro  thanks for contributing your fix
we can pick it to active release branches, are you interested in a particular version? 4.13 ? 

can you please add a unit test that reproduces the problem ?
also the title of the PR is not telling about the impact of the patch "doGetLeaves" is not very representative.
Probably you can change it and refer to RackawareEnsemblePlacementPolicyImpl and/or reducing spammy log

mkozioro:
Hi @eolivelli,

I've added test for failing case.
I'm interested in 4.12 branch. 
Is PR name little better now?

eolivelli:
@mkozioro very good !

Can you also add some positive tests ? some test that does not return an empty set ?

mkozioro:
Hi @eolivelli,

I've added test case for getLeaves() positive case and also for excluded scope.

eolivelli (approved):
LGTM

dlg99 (approved):
lgtm

eolivelli:
Pushed to master, branch-4.13 and branch-4.12

mkozioro:
Thanks.

PR #2633, opened 2021-03-01
add flag to control whether basic metrics expose to prometheus or not

hangc0276:
### Motivation
When using prometheus-metrics-provider lib provided by BookKeeper, it will expose basic metrics such as cpu, jvm, memory metrics etc. However, other systems, eg. Apache Pulsar, may also expose those metrics, which will lead to those metrics register twice, and prometheus metrics provider thread will start failed.

### Changes
1. Add a flag to control whether basic metrics expose to prometheus or not.

### DIFF
--- a/bookkeeper-stats-providers/prometheus-metrics-provider/src/main/java/org/apache/bookkeeper/stats/prometheus/PrometheusMetricsProvider.java
+++ b/bookkeeper-stats-providers/prometheus-metrics-provider/src/main/java/org/apache/bookkeeper/stats/prometheus/PrometheusMetricsProvider.java
@@ -75,10 +75,14 @@ public class PrometheusMetricsProvider implements StatsProvider {
     public static final String PROMETHEUS_BASIC_STATS_ENABLE = "prometheusBasicStatsEnable";
     public static final boolean DEFAULT_PROMETHEUS_BASIC_STATS_ENABLE = true;
 
+    public static final String PROMETHEUS_METRIC_PREFIX = "prometheusMetricPrefix";
+    public static final String DEFAULT_PROMETHEUS_METRIC_PREFIX = "";
+
     final CollectorRegistry registry;
 
     Server server;
     private final CachingStatsProvider cachingStatsProvider;
+    private String metricPrefix = DEFAULT_PROMETHEUS_METRIC_PREFIX;
 
     /*
      * These acts a registry of the metrics defined in this provider
@@ -115,7 +119,8 @@ public class PrometheusMetricsProvider implements StatsProvider {
                 if (statsComponents.length == 0) {
                     return "";
                 } else if (statsComponents[0].isEmpty()) {
-                    completeName = StringUtils.join(statsComponents, '_', 1, statsComponents.length);
+                    completeName = StringUtils.join(statsComponents, '_',
+                        1, statsComponents.length);
                 } else {
                     completeName = StringUtils.join(statsComponents, '_');
                 }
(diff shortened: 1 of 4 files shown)


eolivelli (approved):
LGTM

dlg99 (changes requested):
Please consider configurable metric prefix instead, like "bk." for bookkeeper, "pulsar." for pulsar. Just pass the prefix as a sting, empty by default to keep current behavior.
In this case use a parameter to configure the prefix, and then do something along the lines of  `registerMetrics(Gauge.build(metricPrefix + "jvm_memory_direct_bytes_used", "-").create()`

This way you can monitor both bookie and pulsar, alert on them both etc.
Now your option is to monitor pulsar and ignore bookie if they are co-hosted. Eventually you'll need to troubleshoot bookie and won't have metrics that help.

eolivelli:
@dlg99 
I believe that the standard metrics cannot be renamed
```
            registerMetrics(new StandardExports());
            registerMetrics(new MemoryPoolsExports());
            registerMetrics(new GarbageCollectorExports());
            registerMetrics(new ThreadExports());
```

dlg99:
@eolivelli can the names be changed there https://github.com/apache/bookkeeper/blob/5700641ea509be36a6395f56aae2f60a88fdcaab/bookkeeper-stats-providers/prometheus-metrics-provider/src/main/java/org/apache/bookkeeper/stats/prometheus/PrometheusMetricsProvider.java#L113

otherwise standard metrics can be extended / collect() overridden to provide overridden name https://github.com/prometheus/client_java/blob/7a6b9f1300c5445f477fb0452c5ec7b6f80464a1/simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/GarbageCollectorExports.java#L40

hangc0276:
> @eolivelli can the names be changed there
> 
> https://github.com/apache/bookkeeper/blob/5700641ea509be36a6395f56aae2f60a88fdcaab/bookkeeper-stats-providers/prometheus-metrics-provider/src/main/java/org/apache/bookkeeper/stats/prometheus/PrometheusMetricsProvider.java#L113
> 
> otherwise standard metrics can be extended / collect() overridden to provide overridden name https://github.com/prometheus/client_java/blob/7a6b9f1300c5445f477fb0452c5ec7b6f80464a1/simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/GarbageCollectorExports.java#L40

eolivelli (approved):
LGTM

hangc0276:
> @eolivelli can the names be changed there
> 
> https://github.com/apache/bookkeeper/blob/5700641ea509be36a6395f56aae2f60a88fdcaab/bookkeeper-stats-providers/prometheus-metrics-provider/src/main/java/org/apache/bookkeeper/stats/prometheus/PrometheusMetricsProvider.java#L113
> 
> otherwise standard metrics can be extended / collect() overridden to provide overridden name https://github.com/prometheus/client_java/blob/7a6b9f1300c5445f477fb0452c5ec7b6f80464a1/simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/GarbageCollectorExports.java#L40

@dlg99 Thanks for you feedback. This method can override all the metrics' name when exposing to http port, but the metics name which registered into `CollectorRegistry` can't be override. Thus it will also lead to the same metric name registered twice and the prometheus metrics provider thread start failed. Thus, we should keep the flag to turn off `standard metrics` register and expose.

I also update the code and support adding metric prefix on metric name for standard metrics and pre-defined metrics. For user registered metric, and metric prefix won't be added.

@eolivelli @dlg99  PTAL thanks.

dlg99 (approved):
LGTM. 
please rebase/resolve conflicts.

StevenLuMT:

PR #2635, opened 2021-03-03
Handle corrupted checkpoints

sursingh:
### Motivation

If a checkpoint is corrupted, checkpoint restoration will fail. Currently we will keep
retrying to restore the same checkpoint. The system will never recover from this.
The only way out is to manually remove the checkpoint from the zookeeper

### Changes

When a checkpoint is corrupted, we need to try and restore from previous checkpoint.
This  is done recursively till we hit a checkpoint that works. If none of the checkpoints
work, we will fallback to a `nullcheckpoint` which will initialize rocksdb from scratch
and replay all state from journal.

Instead of just dealing with the latest checkpoint, we now return a list of checkpoints
ordered by age. We then simply walk this list and stop as soon as a checkpoint is
successfully restored. We currently only keep a single checkpoint,  but with this change
we can potentially keep `N` checkpoints to provide more robustness.

### DIFF
--- a/stream/statelib/src/main/java/org/apache/bookkeeper/statelib/impl/rocksdb/checkpoint/CheckpointInfo.java
+++ b/stream/statelib/src/main/java/org/apache/bookkeeper/statelib/impl/rocksdb/checkpoint/CheckpointInfo.java
@@ -1,0 +1,150 @@
+/*
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+package org.apache.bookkeeper.statelib.impl.rocksdb.checkpoint;
+
+import com.google.common.io.MoreFiles;
+import com.google.common.io.RecursiveDeleteOption;
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.UUID;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.bookkeeper.statelib.api.checkpoint.CheckpointStore;
+import org.apache.bookkeeper.statelib.api.exceptions.StateStoreException;
+import org.apache.bookkeeper.stream.proto.kv.store.CheckpointMetadata;
+
+
+/**
+ * CheckpointInfo encapsulated information and operatation for a checkpoint.
+ */
+@Slf4j
+public class CheckpointInfo implements Comparable<CheckpointInfo> {
+
+    final String id;
+    CheckpointMetadata metadata;
+
+    public CheckpointInfo(String id, InputStream is) throws IOException {
+        this.id = id;
+        this.metadata = CheckpointMetadata.parseFrom(is);
+    }
+
+    public CheckpointInfo(String id) {
+        this.id = id;
+    }
+    public String getId() {
+        return id;
+    }
+
+    public CheckpointMetadata getMetadata() {
+        return metadata;
+    }
+
+    @Override
+    public String toString() {
+        return "Checkpoint{ID='" + id + "', createdAt: " + getCreatedAt() + " " + metadata + "}";
+    }
+
+    // default fallback checkpoint. This essentially creates a blank store.
+    public static CheckpointInfo nullCheckpoint() {
+        return new CheckpointInfo(UUID.randomUUID().toString()) {
+            public CheckpointMetadata restore(String dbName,
+                                              File dbPath,
+                                              CheckpointStore store) throws StateStoreException {
+                try {
+                    Files.createDirectories(getCheckpointPath(dbPath));
+                } catch (IOException ioe) {
+                    throw new StateStoreException("Failed to create dir " + dbName, ioe);
+                }
+                return null;
+            }
+            // for proper sorting
+            public Long getCreatedAt() {
+                return Long.valueOf(0);
+            }
+        };
+    }
+
+    public void remove(File dbPath) {
+        try {
+            Path checkpointPath = getCheckpointPath(dbPath);
+            if (checkpointPath.toFile().exists()) {
+                MoreFiles.deleteRecursively(checkpointPath, RecursiveDeleteOption.ALLOW_INSECURE);
+            }
+        } catch (IOException ioe) {
+            log.warn("Failed to remove unused checkpoint {} from {}",
+                id, getCheckpointBaseDir(dbPath), ioe);
+        }
+    }
+
+    public Long getCreatedAt() {
+        return this.metadata.getCreatedAt();
+    }
+
+    public File getCheckpointBaseDir(File dbPath) {
+        return new File(dbPath, "checkpoints");
+    }
+
+    public Path getCheckpointPath(File dbPath) {
+        return Paths.get(getCheckpointBaseDir(dbPath).getAbsolutePath(), id);
+    }
+
+    public File getCurrentDir(File dbPath) {
+        return new File(dbPath, "current");
+    }
+    public Path getCurrentPath(File dbPath) {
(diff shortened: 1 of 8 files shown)


dlg99:
@sursingh I'd like to clarify my understanding of specific scenario and inner working of the stream store:

Store, i.e. RocksdbKVAsyncStore, writes op to a journal (distributed).
periodically it checkpoint/writes local rocks db data to the distributed store (BK)
At some point after checkpoint the store truncates journal.
Last checkpoint + journal is enough to restore consistent dataset / full data

Now if last checkpoint is corrupt we'll fallback to the previous checkpoint.
It will pick up whatever journal available (`getDLSNNotLessThanTxId(checkpointedTxId)`) after the checkpoint
It will not detect if journal data in the range `[checkpointedTxId, firstAvailableInJournalTxId]` is missing.
The store gets data restored form an older checkpoint + journal that starts after the last (corrupt) checkpoint was created.
Now the store has lost some data/has inconsistent data.

Is my understanding correct?

stream/statelib/src/main/java/org/apache/bookkeeper/statelib/impl/rocksdb/checkpoint/CheckpointInfo.java
```diff
@@ -0,0 +1,151 @@
...
+import org.apache.bookkeeper.statelib.api.exceptions.StateStoreException;
+import org.apache.bookkeeper.stream.proto.kv.store.CheckpointMetadata;
+
+
+/**
+ * CheckpointInfo encapsulated information and operatation for a checkpoint.
+ */
+@Slf4j
+public
+class CheckpointInfo implements Comparable<CheckpointInfo> {
```
eolivelli, line 43: nit: format "public"

stream/statelib/src/main/java/org/apache/bookkeeper/statelib/impl/rocksdb/checkpoint/CheckpointInfo.java
```diff
@@ -0,0 +1,151 @@
...
+
+    String id;
```
eolivelli, line 45: can we make these fields "final"

stream/statelib/src/main/java/org/apache/bookkeeper/statelib/impl/rocksdb/checkpoint/RocksCheckpointer.java
```diff
@@ -103,32 +79,38 @@ private static void cleanupLocalCheckpoints(File checkpointsDir, String checkpoi
...
-        List<String> files = checkpointStore.listFiles(remoteCheckpointsPath);
-        CheckpointMetadata latestCheckpoint = null;
-        String latestCheckpointId = null;
+        ArrayList<CheckpointInfo> result = new ArrayList<>();
+        result.add(CheckpointInfo.nullCheckpoint());
+        List<String> files;
+        try {
+            files = store.listFiles(remoteCheckpointsPath);
+        } catch (IOException e) {
+            return result;
```
eolivelli, line 96: can we log this error ?

stream/statelib/src/main/java/org/apache/bookkeeper/statelib/impl/rocksdb/checkpoint/RocksCheckpointer.java
```diff
@@ -103,32 +79,38 @@ private static void cleanupLocalCheckpoints(File checkpointsDir, String checkpoi
         }
     }
 
-    private static Pair<String, CheckpointMetadata> getLatestCheckpoint(
-        String dbPrefix, CheckpointStore checkpointStore) throws IOException {
+    private static CheckpointInfo getLatestCheckpoint(
+        String dbPrefix, CheckpointStore checkpointStore) {
+        List<CheckpointInfo> checkpoints = getCheckpoints(dbPrefix, checkpointStore);
+        return checkpoints.get(0);
```
eolivelli, line 85: can we deal with an empty list ?
maybe it should not happen, but in case it happen we will have a bad runtime exception without a meaningful error message

stream/statelib/src/main/java/org/apache/bookkeeper/statelib/impl/rocksdb/checkpoint/RocksCheckpointer.java
```diff
@@ -103,32 +79,38 @@ private static void cleanupLocalCheckpoints(File checkpointsDir, String checkpoi
...
-                } else if (latestCheckpoint.getCreatedAt() < ckpt.getCreatedAt()) {
-                    latestCheckpointId = checkpointId;
-                    latestCheckpoint = ckpt;
-                }
+            try (InputStream is = store.openInputStream(metadataPath)) {
+                result.add(new CheckpointInfo(checkpointId, is));
             } catch (FileNotFoundException fnfe) {
                 log.error("Metadata is corrupt for the checkpoint {}. Skipping it.", checkpointId);
+            } catch (IOException e) {
+                log.error("IO exception {}. Skipping it.", checkpointId);
```
eolivelli, line 109: can we log the error?

eolivelli (changes requested):
very nice.
I left a few little suggestions

PTAL

sursingh:
@dlg99 Currently there is no explicit truncation of the Journal. There is default setting that drops records older than 72 hours. So if the old checkpoint is older than 72 hours we will be missing the data.

  The way it is currently setup the TxId is stored in the rocks db. We use this to get the last TxId that is available in the checkpoint. After restoring the checkpoint we trigger a replay from the journal starting from the last TxId that we get after restoring the checkpoint. So if we restore from older checkpoint, we will replay from older TxId. This can also detect that there is missing data if the starting TxId is not available in the Journal. I am not sure how that is handled currently by DistributedLog. Ideally that should also throw an exception during journal replay.

stream/statelib/src/main/java/org/apache/bookkeeper/statelib/impl/rocksdb/checkpoint/RocksCheckpointer.java
sursingh: Ideally this will never happen as there should always be a `nullCheckpoint` added to the list. But we add a explicit check to ensure proper error message

stream/statelib/src/main/java/org/apache/bookkeeper/statelib/impl/rocksdb/checkpoint/RocksCheckpointer.java
```diff
@@ -103,32 +79,44 @@ private static void cleanupLocalCheckpoints(File checkpointsDir, String checkpoi
...
-        List<String> files = checkpointStore.listFiles(remoteCheckpointsPath);
-        CheckpointMetadata latestCheckpoint = null;
-        String latestCheckpointId = null;
+        ArrayList<CheckpointInfo> result = new ArrayList<>();
+        result.add(CheckpointInfo.nullCheckpoint());
+        List<String> files;
+        try {
+            files = store.listFiles(remoteCheckpointsPath);
+        } catch (IOException e) {
+            log.warn("No remote checkpoints available. Starting with nullCheckpoint: {}", e);
```
dlg99, line 101: remove `{}` to log exception properly

stream/statelib/src/test/java/org/apache/bookkeeper/statelib/impl/kv/TestRocksdbKVStoreCheckpoint.java
```diff
@@ -0,0 +1,97 @@
...
+import org.junit.rules.TestName;
+
+
+
+
+/**
+ * Test cases for Rocksdb KV Store with checkpoints.
+ */
+@Slf4j
+public class TestRocksdbKVStoreCheckpoint {
```
dlg99, line 40: following up on 

> So if we restore from older checkpoint, we will replay from older TxId. This can also detect that there is missing data if the starting TxId is not available in the Journal. I am not sure how that is handled currently by DistributedLog. Ideally that should also throw an exception during journal replay.

This change needs a test reproing this case to confirm that the change does not introduce a silent data loss.

dlg99 (changes requested):
Overall it looks great.
I bet it safely covers about 99% of the checkpoint corruption cases but I still feel that it needs a test reproing the checkpoint drop after the journal truncation to confirm that the change does not introduce a silent data loss scenario.
I'll be ok with something like "bookie won't start and requires manual interference in that extreme case", we can figure out how to deal with it later.

stream/statelib/src/test/java/org/apache/bookkeeper/statelib/impl/kv/TestRocksdbKVStoreCheckpoint.java
sursingh: DistributedLog starts the replay after the specified transactions. So it skips over the missing records. To address this I have added an explicit verification for starting transaction at the start of journal replay. Also added a test-case to test the same (see testMissingData() above).

dlg99 (approved):
Fantastic work!
Thank you.

sursingh:
@eolivelli I have taken care of your comments. Please take a look.

eolivelli (approved):
LGTM

PR #2641, opened 2021-03-10
Add checksum validation for SST files

sursingh:
### Motivation

Normally the SST file are immutable. A SST file from previous checkpoint can
be reused in subsequent checkpoints. This fact is used to avoid unnecessary
upload of SST files.

However there are scenarios in which just the name comparison doesn't work.
It is possible that the checkpoint process doesn't complete (due to
crash/restart). In such cases the stale SST files are left behind. When the
storage container is restarted, it will be correctly restored from previous
checkpoint. When we do a checkpoint on this new state, a new SST files
are created. Since we only compare the SST file names, we assume that files is
already available in the checkpoint store.

At best the size of the new files will mismatch, and restore will fail. But if
the size of the files match, restore will succeed and we will have invalid
data in the state store.

### Changes

With this change we are adding the checksum for the SST files. The checksum
will be appended to the name when the SST file is uploaded. This will ensure
that the correct files are always uploaded

### DIFF
--- a/stream/statelib/src/main/java/org/apache/bookkeeper/statelib/impl/rocksdb/checkpoint/CheckpointFile.java
+++ b/stream/statelib/src/main/java/org/apache/bookkeeper/statelib/impl/rocksdb/checkpoint/CheckpointFile.java
@@ -1,0 +1,226 @@
+/*
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+package org.apache.bookkeeper.statelib.impl.rocksdb.checkpoint;
+
+import com.google.common.hash.Hashing;
+import com.google.common.io.Files;
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.nio.file.Paths;
+import java.nio.file.StandardCopyOption;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.bookkeeper.statelib.api.checkpoint.CheckpointStore;
+import org.apache.bookkeeper.statelib.impl.rocksdb.RocksUtils;
+import org.apache.bookkeeper.stream.proto.kv.store.CheckpointMetadata;
+import org.apache.bookkeeper.stream.proto.kv.store.FileInfo;
+
+/**
+ * CheckpointFile encapsulates the attributes and operations for a file in checkpoint.
+ */
+@Slf4j
+@lombok.Builder
+@lombok.EqualsAndHashCode
+public class CheckpointFile {
+    private final File file;
+    private final String checksum;
+    private final boolean isSstFile;
+
+    private CheckpointFile(File file, String checksum, boolean isSstFile) {
+        this.file = file;
+        this.checksum = checksum;
+        this.isSstFile = isSstFile;
+    }
+
+    /**
+     * CheckpointFileBuilder for building instances of CheckpointFile objects.
+     */
+    public static class CheckpointFileBuilder {
+        CheckpointFileBuilder file(File file) {
+            this.file = file;
+            this.isSstFile = RocksUtils.isSstFile(this.file);
+            return this;
+        }
+        CheckpointFileBuilder file(File checkpointDir, String filename) {
+            file(new File(checkpointDir, filename));
+            return this;
+        }
+
+        CheckpointFileBuilder computeChecksum() {
+            return checksum(computeChecksum(this.file));
+        }
+
+        private static String computeChecksum(File file) {
+            String ckSum = "invalid-" + System.currentTimeMillis();
+            try {
+                ckSum = Files.asByteSource(file).hash(Hashing.sha256()).toString();
+                return ckSum;
+            } catch (IOException e) {
+                log.error("Failed to get checksum for file {} {}", file.getName(), e.getMessage(), e);
+            }
+            return ckSum;
+        }
+    }
+
+    public File getFile() {
+        return file;
+    }
+
+    public String toString() {
+        return String.format("CheckpointFile: %s", file.getName());
+    }
+
+    public boolean isSstFile() {
+        return isSstFile;
+    }
+
+    public String getName() {
+        return file.getName();
+    }
+
+    public String getNameWithChecksum() {
+        if (checksum != null) {
+            return file.getName() + "_" + checksum;
+        }
+        return getName();
+    }
+
+    public String getRemoteSstPath(String dbPrefix, boolean enableChecksum) {
+        if (enableChecksum) {
+            return RocksUtils.getDestSstPath(dbPrefix, getNameWithChecksum());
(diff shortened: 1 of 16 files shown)


stream/statelib/src/main/java/org/apache/bookkeeper/statelib/impl/rocksdb/checkpoint/CheckpointFile.java
```diff
@@ -0,0 +1,212 @@
...
+ */
+@Slf4j
+public class CheckpointFile {
+    private final File file;
+    private final String checksum;
+    private final boolean isSstFile;
+
+    CheckpointFile(File file) {
+        this.file = file;
+        this.checksum = computeChecksum();
```
dlg99, line 52: what if checksum is `"invalid"` as returned by `computeChecksum()` in case of error?
I.e. it is invalid when saved to the checkpoint and when verified later, will the code pass the checksum check?
"invalid_<timestamp_here>" will avoid that, I bet there is a more elegant solution.

stream/statelib/src/main/java/org/apache/bookkeeper/statelib/impl/rocksdb/checkpoint/RocksdbRestoreTask.java
```diff
@@ -67,65 +67,10 @@ public void restore(String checkpointId, CheckpointMetadata metadata) throws Sta
...
-    }
-
     private void copyFilesFromRemote(String checkpointId,
-                                     File checkpointedDir,
-                                     List<String> remoteFiles) throws IOException {
-        for (String file : remoteFiles) {
-            copyFileFromRemote(checkpointId, checkpointedDir, file);
+                                     List<CheckpointFile> remoteFiles) throws IOException {
+        for (CheckpointFile file : remoteFiles) {
+            file.copyFromRemote(checkpointStore, dbPrefix, checkpointId);
```
dlg99, line 73: I am probably missing something obvious.
I expected checksum validation after file copy from remote, is it happening at some other place (implicitly via equals, as part of some map/set check?)?

dlg99 (changes requested):
My major concern is the backwards compatibility, see the comments

stream/statelib/src/test/java/org/apache/bookkeeper/statelib/impl/rocksdb/checkpoint/RocksCheckpointerTest.java
```diff
@@ -349,6 +347,65 @@ public void testCheckpointOrder() throws Exception {
...
+        String newCheckpoint = testStore.checkpoint("checkpoint-3");
+
+        // Ensure latest checkpoint can be restored.
+        testStore.destroyLocal();
+        testStore.restore();
+        assertEquals("transaction-3", testStore.get("transaction-id"));
+    }
+
+    @Test
+    public void testRestoreOldCheckpointWithoutChecksum() throws Exception {
```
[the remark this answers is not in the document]
eolivelli: +1
yes, please add a compatibility test,
also can you please state in the description how to enable this feature ?

would it be possible to enable this feature by default and deal with old SST files that are missing the checksum ?
otherwise I bet that no one will use this feature

stream/statelib/src/main/java/org/apache/bookkeeper/statelib/impl/rocksdb/checkpoint/RocksdbCheckpointTask.java
```diff
@@ -144,64 +149,38 @@ public String checkpoint(byte[] txid) throws StateStoreException {
...
+                                    boolean enableChecksum) throws IOException {
 
         CheckpointMetadata.Builder metadataBuilder = CheckpointMetadata.newBuilder();
-        for (File file : files) {
-            metadataBuilder.addFiles(file.getName());
+        for (CheckpointFile file : files) {
+            if (enableChecksum) {
+                metadataBuilder.addFileInfos(file.getFileInfo());
+            } else {
+                metadataBuilder.addFiles(file.getName());
```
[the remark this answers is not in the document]
sursingh: Will look into adding an option to retain the files without checksum and add config to control. This can be enabled by default.

stream/statelib/src/test/java/org/apache/bookkeeper/statelib/impl/rocksdb/checkpoint/RocksCheckpointerTest.java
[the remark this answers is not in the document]
sursingh: @eolivelli : This is enabled by default. The change does take care of restoring from old checkpoints without checksum. Andrey's concern is about downgrading from new version to old version that doesn't support checksum. In that case, we will not be able to restore from the checkpoint. Will look into it and see if we can seamlessly support this.

stream/statelib/src/main/java/org/apache/bookkeeper/statelib/impl/rocksdb/checkpoint/CheckpointFile.java
```diff
@@ -0,0 +1,212 @@
...
+        this.isSstFile = RocksUtils.isSstFile(this.file);
+    }
+
+    CheckpointFile(File checkpointDir, String filename) {
+        this.file = new File(checkpointDir, filename);
+        this.checksum = null;
```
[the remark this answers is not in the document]
sursingh: Thanks. Will look into it.

stream/statelib/src/main/java/org/apache/bookkeeper/statelib/impl/rocksdb/checkpoint/RocksdbRestoreTask.java
sursingh: The rocksdb already has validation to verify the contents of the SST files. Here we want to ensure that we use the correct SST file with the checkpoint. Essentially we are giving the SST files a unique name based on their contents.

dlg99: I see, thank you. 

Let's say the name+checksum gets added to both FileInfos and Files.
Will this be good enough for the older bookie to use the checkpoint?

sursingh: No, that will not work. The files need to get renamed to *.sst for the rocksdb restore process to work. Also there are other places where we check if the name ends with ".sst" to determine if this is a `sst` file.

sursingh: Let me see if I can add some clarity.
* Restore code get the list of files from the `metadata` file (which is the first file we restore). All files other than SST files are stored under the checkpoint directory. SST files however are stored under common "ssts" directory. The code uses the extension of the file to determine if the the file is SST or nor. If it is a SST file, it will copy it from common "ssts" folder. If we add checksum to names of file added to `files` field in metadata, this code will fail and will not be able to copy the files from the checkpoint store.
* Now we can work around the above by prefixing the checksum (e.g 00009.sst -> {checksum}_00009.sst). The code will correctly detect this as a SST file and copy it from the common "ssts" folder. These files however would still need to be renamed ({checksum}_00009.sst -> 00009.sst) locally. This is the files that rocksdb will be looking for. Since there is no code in the previous version that can do that, it will not work.

dlg99: would it help if we used relative path there? like "./{checksum}/../00009.sst"?
I am not sure if [normalize()](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Path.html#normalize--) happens automatically or not when the new file created.

sursingh: I don't think that will help here.

sursingh:
@dlg99 : Added support to allow downgrade to older version. There is a config knob that is enabled by default. This will create files with regular names to allow old versions to restore from the checkpoint.

dlg99 (approved):
Looks awesome!
Thank you for going the extra mile with the downgrade compatibility and tests.

eolivelli (approved):
LGTM

great work

sursingh:
rerun failed tests

PR #2642, opened 2021-03-12
fix region/rack aware placement police replace bookie bug

hangc0276:
### Motivation
 When one bookie broken down, the replication worker will start to replicate missing fragment for ledgers which located on lost bookie. If we use region/rack placement policy for bookkeeper, the replication worker's bookie client will select new ensembles according to region/rack placement policy configured in `bk_server.conf`.  The region/rack placement policy will attempt to select bookies which located in the same region/rack with the lost bookie. If the condition couldn't be satisfied, it will trigger random selection from all available bookies.

However, when we attempt to get the lost bookie's region/rack, it will be failed. The location lost bookie's region/rack logic as follow.
1. Get from `knownBookies` map. Due to the bookie has been lost, the `handleBookiesThatLeft` will be triggered, and the bookie's mapping information will be removed. Thus get from `knowBookies` will be failed.
2. Call `createBookieNode` method to resolve region/rack information. It will resolve BookieSocketAddress according to BookieId. However, the bookie has been lost, and the bookie information has been removed from Zookeeper's ZNode, the `bookieAddressResolver.resolve(addr)` will failed and throw `BookieIdNotResolvedException`. In the end, it will use `DEFAULT_REGION_AND_RACK` instead. the The code list as follow.
```Java
protected String resolveNetworkLocation(BookieId addr) {
        try {
            return NetUtils.resolveNetworkLocation(dnsResolver, bookieAddressResolver.resolve(addr));
        } catch (BookieAddressResolver.BookieIdNotResolvedException err) {
            LOG.error("Cannot resolve bookieId {} to a network address, resolving as {}", addr,
                      NetworkTopology.DEFAULT_REGION_AND_RACK, err);
            return NetworkTopology.DEFAULT_REGION_AND_RACK;
        }
    }

public BookieSocketAddress resolve(BookieId bookieId) {
        try {
            BookieServiceInfo info = FutureUtils.result(registrationClient.getBookieServiceInfo(bookieId)).getValue();
            BookieServiceInfo.Endpoint endpoint = info.getEndpoints()
                    .stream().filter(e -> e.getProtocol().equals("bookie-rpc")).findAny().orElse(null);
            if (endpoint == null) {
                throw new Exception("bookie " + bookieId + " does not publish a bookie-rpc endpoint");
            }
            BookieSocketAddress res = new BookieSocketAddress(endpoint.getHost(), endpoint.getPort());
            if (!bookieId.toString().equals(res.toString())) {
                // only print if the information is useful
                log.info("Resolved {} as {}", bookieId, res);
            } else {
                log.debug("Resolved {} as {}", bookieId, res);
            }
            return res;
        } catch (BKException.BKBookieHandleNotAvailableException ex) {
            if (BookieSocketAddress.isDummyBookieIdForHostname(bookieId)) {
                log.debug("Resolving dummy bookie Id {} using legacy bookie resolver", bookieId);
                return BookieSocketAddress.resolveDummyBookieId(bookieId);
            }
            log.info("Cannot resolve {}, bookie is unknown", bookieId, ex.toString());
            throw new BookieIdNotResolvedException(bookieId, ex);
        } catch (Exception ex) {
            if (ex instanceof InterruptedException) {
            }
            throw new BookieIdNotResolvedException(bookieId, ex);
        }
    }
```
3.  Select available bookies from `DEFAULT_REGION_AND_RACK`, and lead to random select from available bookies.

### Changes
1. Add historyBookies map for Region/Rack placement policy, it will add/update `BookieId -> BookieNode` mapping when new bookies joined (trigger `handleBookiesThatJoined` method), and won't be removed when bookie left.
2. When get lost bookie's placement information,  it will get from knowBookies map first. If it failed, then get from historyBookies map. If both failed, it will resolve the bookie placement information.


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
```diff
@@ -456,7 +456,10 @@ public void uninitalize() {
             excludeBookies.addAll(currentEnsemble);
             BookieNode bn = knownBookies.get(bookieToReplace);
             if (null == bn) {
-                bn = createBookieNode(bookieToReplace);
+                bn = historyBookies.get(bookieToReplace);
+                if (null == bn) {
+                    bn = createBookieNode(bookieToReplace);
```
eolivelli, line 461: probably it is better to try to use createBookieNode and fallback to history if it fails.
otherwise you won't be able to follow a change in the network location for the bookie.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RegionAwareEnsemblePlacementPolicy.java
```diff
@@ -446,7 +447,10 @@ public RegionAwareEnsemblePlacementPolicy initialize(ClientConfiguration conf,
 
             BookieNode bookieNodeToReplace = knownBookies.get(bookieToReplace);
             if (null == bookieNodeToReplace) {
-                bookieNodeToReplace = createBookieNode(bookieToReplace);
+                bookieNodeToReplace = historyBookies.get(bookieToReplace);
```
eolivelli, line 450: the same here

eolivelli (changes requested):
thanks for fixing this.

I left a comment.
Also, can you please add a test.
We already have a few tests about Placement policy so you can add your tests next to the others

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
hangc0276: Thanks for your quickly feedback. OK, it's better to fallback to history when resolve bookieId failed, it will update the code.

hangc0276:
> thanks for fixing this.
> 
> I left a comment.
> Also, can you please add a test.
> We already have a few tests about Placement policy so you can add your tests next to the others

@eolivelli ok, I will add the tests soon.

hangc0276:
@eolivelli @sijie @merlimat @codelipenghui  PTAL, thanks.

hangc0276:
rerun failure checks

hangc0276:
rerun failure checks

hangc0276:
rerun failure checks

aahmed-se (approved):

hangc0276:
rerun failure checks

hangc0276:
rerun failure checks

hangc0276:
rerun failure checks

hangc0276:
rerun failure checks

hangc0276:
rerun failure checks

hangc0276:
rerun failure checks

hangc0276:
rerun failure checks

hangc0276:
rerun failure checks

hangc0276:
rerun failure checks

hangc0276:
rerun failure checks

hangc0276:
ping @eolivelli , Please help review this PR, thanks a lot.

hangc0276:
ping @eolivelli , Please help review this PR, thanks a lot.

hangc0276:
rerun failure checks

eolivelli:
@dlg99 @nicoloboschi PTAL

PR #2645, opened 2021-03-15
Defect-Fix Compaction status report into correct buckets.

klwilson227:
The compaction status report is off by 1. 

When compaction is enabled, the 10% log is always empty. Setting the compaction threshold to .2 we were seeing the number of compacted entrylogs that were removed, match the sum of the 20% and 30% buckets. While the 10% bucket remained empty. Increasing the threshold to .3 we would see the sum of the first 4 buckets compacted (10%-40%). Also with the 10% bucket empty. 

Upon inspection of the code, the code is using the .ceil function to calculate the bucket index to be updated. This means that 10% bucket will never be used. However, there could be a rare 100% case in which the code runs into a index out of bounds exception when the entrylog is actually 100% used.  The index calculation should be using floor and not the ceil call using the next index lower. So usage of .00 to .9 ends up in the 0 index instead of the current calculation which puts the .0 - .9  in the 1 index. 

This will leave the 100% index empty most of the time, as rarely will the entry logs be 100% used.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
@@ -454,9 +454,7 @@ public class GarbageCollectorThread extends SafeRunnable {
         int[] entryLogUsageBuckets = new int[numBuckets];
 
         for (EntryLogMetadata meta : logsToCompact) {
-            int bucketIndex = Math.min(
-                    numBuckets - 1,
-                    (int) Math.ceil(meta.getUsage() * numBuckets));
+            int bucketIndex = calculateUsageIndex(numBuckets, meta.getUsage());
             entryLogUsageBuckets[bucketIndex]++;
 
             if (meta.getUsage() >= threshold) {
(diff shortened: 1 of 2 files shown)


eolivelli:
thank you @klwilson227 

can you please add a unit test (no need to start the bookie) ?

klwilson227 (commented):
If there was some sort of start/framework for testing compaction it should be trivial but since no one has taken the time to create the framework I’ve got to learn all of gc/bookkeeper to mock out classes to test the basics first. I do not yet know the gc well enough to be able to efficiently make those changes. Some refactoring for test ability is also in order.

klwilson227:
@eolivelli @jiazhai @sijie  is the integration test flaky as this run was rebased on top of master and had issues loading etc docker container which I believe is not related to change.

klwilson227:
Test cases updated. Please review.

jiazhai:
@klwilson227 Thanks for the help. right. the CI failure is not related. I triggered a re-run

jiazhai (approved):

jiazhai:
@zymap to also take a look

zymap (approved):

codelipenghui (approved):

PR #2646, opened 2021-03-17
Auto-throttle read operations

merlimat:
### Motivation

When bookie is doing reads, it may use a considerable amount of memory if there are many simultaneous readers and especially when the network becomes the bottleneck.

The problem is that the read thread is scheduling the async operation to write on the socket and then it moves on for the next read operation. If the TCP window is full, these write requests on the socket are just piling up in memory, resulting in a spike of direct memory usage, which may exceed the max configured in the JVM.

Instead, if the socket write operation is blocking, we should block the BK read thread in order to auto-throttle and limit the amount of memory. 

This change does not introduce any performance penalty on the read-path. The old behavior will still be available as a config option.

### DIFF
--- a/conf/bk_server.conf
+++ b/conf/bk_server.conf
@@ -137,6 +137,11 @@ extraServerComponents=
 # avoid the executor queue to grow indefinitely
 # maxPendingAddRequestsPerThread=10000
 
+# Use auto-throttling of the read-worker threads. This is done
+# to ensure the bookie is not using unlimited amount of memory
+# to respond to read-requests.
+# readWorkerThreadsThrottle=true
+
 # Option to enable busy-wait settings. Default is false.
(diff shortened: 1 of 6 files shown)


eolivelli (approved):
LGTM

conf/bk_server.conf
```diff
@@ -137,6 +137,11 @@ extraServerComponents=
 # avoid the executor queue to grow indefinitely
 # maxPendingAddRequestsPerThread=10000
 
+# Use auto-throttling of the read-worker threads. This is done
+# to ensure the bookie is not using unlimited amount of memory
+# to respond to read-requests.
+# readWorkerThreadsThrottle=true
```
codelipenghui, line 143:
```suggestion
# readWorkerThreadsThrottleEnabled=true
```

merlimat: Good point. Renamed to "ThrottlingEnabled"

PR #2647, opened 2021-03-17
Limit read-ahead bytes to the size of the read cache

merlimat:
### Motivation

In DbLedgerStorage, the read ahead that feeds the read cache is being configured in terms of number of entries. If the entry size is big and the read cache is small, this could lead to trashing the read cache.

For example, a read-ahead of 1000 entries with 1MB avg size would pre-fetch a 1GB from volume. In the read-ahead there would never be a good read to pre-fetch an amount of data that exceeds the read cache size.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
@@ -134,6 +134,8 @@ public class SingleDirectoryDbLedgerStorage implements CompactableLedgerStorage 
 
     private static final long DEFAULT_MAX_THROTTLE_TIME_MILLIS = TimeUnit.SECONDS.toMillis(10);
 
+    private final long maxReadAheadBytesSize;
+
     public SingleDirectoryDbLedgerStorage(ServerConfiguration conf, LedgerManager ledgerManager,
             LedgerDirsManager ledgerDirsManager, LedgerDirsManager indexDirsManager, StateManager stateManager,
             CheckpointSource checkpointSource, Checkpointer checkpointer, StatsLogger statsLogger,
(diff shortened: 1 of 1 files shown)


eolivelli (approved):
LGTM

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
```diff
@@ -469,7 +474,9 @@ private void fillReadAheadCache(long orginalLedgerId, long firstEntryId, long fi
             int count = 0;
             long size = 0;
 
-            while (count < readAheadCacheBatchSize && currentEntryLogId == firstEntryLogId) {
+            while (count < readAheadCacheBatchSize
+                    && size < readAheadCacheBatchSize
```
codelipenghui, line 478:
```suggestion
                    && size < maxReadAheadBytesSize
```

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
```diff
@@ -155,6 +157,9 @@ public SingleDirectoryDbLedgerStorage(ServerConfiguration conf, LedgerManager le
         readCacheMaxSize = readCacheSize;
         readAheadCacheBatchSize = conf.getInt(READ_AHEAD_CACHE_BATCH_SIZE, DEFAULT_READ_AHEAD_CACHE_BATCH_SIZE);
 
+        // Do not attempt to perform read-ahead more than half the total size of the cache
+        maxReadAheadBytesSize = readCacheMaxSize / 2;
```
codelipenghui, line 161: Do we need to consider the multiple ledger directories?

merlimat: It's already being accounted in that `readCacheMaxSize` size is divided evenly divided by the directories.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
merlimat: Good catch, I messed this up when resolving the conflicts...

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
codelipenghui: ok

PR #2650, opened 2021-03-17
Allow to attach labels to metrics

merlimat:
### Motivation

There are certain metrics reported by bookies or clients that are including the bookie name in the metric name. This is typically an anti-pattern in many metric collections systems. For example, this can create a big number of metrics in Prometheus and it will severely impact the its performance.

In general, metric collection systems support the concept of labels that can be attached to metrics.

### Modifications
 * Added the labels in the BK stats API
 *  Prometheus implementation is using the labels instead of putting that as part of the metric name
 * Other stats providers will keep by default using the label in the metric name for backward compatibility.


eolivelli (approved):
LGTM 

@nicoloboschi please take a look

nicoloboschi (approved):
+1

eolivelli:
@merlimat 
we have a failing test, related to this patch

```
Error:  Tests run: 9, Failures: 3, Errors: 0, Skipped: 0, Time elapsed: 0.545 s <<< FAILURE! - in org.apache.bookkeeper.stats.prometheus.TestPrometheusMetricsProvider
Error:  org.apache.bookkeeper.stats.prometheus.TestPrometheusMetricsProvider.testCache  Time elapsed: 0.117 s  <<< FAILURE!
java.lang.AssertionError: expected same:<org.apache.bookkeeper.stats.prometheus.PrometheusStatsLogger@5f341870> was not:<org.apache.bookkeeper.stats.prometheus.PrometheusStatsLogger@553f17c>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotSame(Assert.java:828)
	at org.junit.Assert.assertSame(Assert.java:771)
	at org.junit.Assert.assertSame(Assert.java:782)
	at org.apache.bookkeeper.stats.prometheus.TestPrometheusMetricsProvider.testCache(TestPrometheusMetricsProvider.java:54)
```

sursingh:
@merlimat : With  this change we are not using caching metrics provider. So caching tests are not longer valid. So the test should be removed

merlimat:
> @merlimat : With this change we are not using caching metrics provider. So caching tests are not longer valid. So the test should be removed

Good point. I'll get this PR back in shape.

Issue #2615, opened 2021-02-20
Ledger recovery can fail due to invalid ensemble creation


Vanlightly:
I have found a sequence of events that could lead to repeated failure of a ledger recovery process that would lead to data unavailability. I have verified this with both my TLA+ specification (https://github.com/Vanlightly/bookkeeper-tlaplus) of the replication protocol and new unit tests.

**BUG REPORT**

When a second writer performs recovery, it can end up trying to create an invalid ensemble which will cause recovery to fail.

The following example involves no timeouts on LAC reads, but an empty ensemble returning -1 for LAC reads. However, the same result would follow if LAC reads timed out, causing the value of -1 to be used as the default.

```
participant w1
participant w2
participant b1
participant b2
participant b3
participant b4
participant b5
participant b6
participant zk

note over w1,zk:Ensembles: 0:{b1,b2}, 1000:{b2,b3}
w1->b1: e1999
w1->b2: e1999
b1->w1: ack e1999
b2->w1: ack e1999
w1--xb1: e2000
w1--xb2: e2000
w1->zk:Add ensemble 2000:{b4,b5}
w2->zk:Start recovery
w2->b4: Read LAC
w2->b5: Read LAC
b4->w2: LAC -1
b5->w2: LAC -1
w2->b1: Read e0
w2->b2: Read e0
b1->w2:e0
b2->w2:e0
w2->b4:e0
w2--xb5:e0
b4->w2:ack e0
w2->zk:Add ensemble 0:{b4,b6}
note over w2,zk:INVALID ENSEMBLE
```

Any ledger recovery is vulnerable to this situation if the ledger has at least 1 existing ensemble, a default of -1 is used for the LAC and an error occurs during the write-back phase causing a new ensemble to be created (from entry 0).

**THE FIX**

The fix is to not use -1 (or 0 in the TLA+ spec) as the minimum but to take the `first entry id` of the current ensemble - 1 as the minimum. This ensures we only try to recover the current ensemble. Previous ensembles, if any, have already been committed and so recovery reads/writes of those ensembles is unnecessary. If a failure occurs during recovery of the current ensemble, then updating that ensemble is a legal operation.

The TLA+ specification uses this new minimum value and so the spec will not currently reach this illegal state.

I will shortly submit a PR with the code fix and new unit tests.

eolivelli:
@Vanlightly thank you for sharing this issue

can you please advertise on dev@bookkeeper about your TLA+ work ?
[...]

PR #2654, opened 2021-03-18
Issue 2615: Fix for invalid ensemble issue during ledger recovery

Vanlightly:
Ensures that only entries of the current ensemble are included in the ledger recovery process, thus avoiding a ledger recovery failure scenario where it tries to append an ensemble with a lower first entry id than the prior ensemble.

This PR includes a small change in the LedgerRecoveryOp that avoids a scenario where ledger recovery tries to create an invalid ensemble thereby failing. This could cause data unavailability for as long as trigger conditions last.

During ledger recovery, only entries of the current ensemble can be included in the read and write back phase. Prior ensembles, if any, are immutable. But it is possible, in a multi-ensemble ledger, for the current ensemble to return an LAC of -1. This then causes the recovery to read entries from prior ensembles and write them back to the current ensemble. This does not cause any data loss, but it is wasteful of both space and time. The main issue is that if an ensemble change occurs when writing back entries, it will try and create a new ensemble with first entry id of 0. This causes an IllegalStateException as there is a check before the CAS metadata op to ensure that the ensemble does not have an entry id lower than an existing ensemble.

If a bookie of the current ensemble were to be down, then the ledger would be unrecoverable until it became available again. 

The solution is that the lowest safe LAC for recovery is: first entry id of current ensemble - 1.

### Changes

Change to LedgerRecoveryOp as described above.
New unit test in LedgerRecoveryTest2.

Master Issue: #2615

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerRecoveryOp.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerRecoveryOp.java
@@ -104,7 +104,24 @@ class LedgerRecoveryOp implements ReadEntryListener, AddCallback {
                     public void readLastConfirmedDataComplete(int rc, RecoveryData data) {
                         if (rc == BKException.Code.OK) {
                             synchronized (lh) {
-                                lh.lastAddPushed = lh.lastAddConfirmed = data.getLastAddConfirmed();
+                                /**
+                                 The lowest an LAC can be for use in recovery is the first entry id
+                                 of the current ensemble - 1.
+                                 All ensembles prior to the current one, if any, are confirmed and
+                                 immutable (so are not part of the recovery process).
+                                 So we take the highest of:
+                                 - the LAC returned by the current bookie ensemble (could be -1)
+                                 - the first entry id of the current ensemble - 1.
+                                 */
+                                Long lastEnsembleEntryId = lh.getVersionedLedgerMetadata()
+                                        .getValue()
+                                        .getAllEnsembles()
+                                        .lastEntry()
+                                        .getKey();
+
+                                lh.lastAddPushed = lh.lastAddConfirmed = Math.max(data.getLastAddConfirmed(),
+                                        (lastEnsembleEntryId - 1));
+
                                 lh.length = data.getLength();
                                 lh.pendingAddsSequenceHead = lh.lastAddConfirmed;
                                 startEntryToRead = endEntryToRead = lh.lastAddConfirmed;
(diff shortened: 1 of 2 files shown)


Vanlightly:
@eolivelli  The testBookieShouldTurnWritableFromReadOnly integration test that is failing is doing so because the BookieWatcherImpl methods processReadOnlyBookiesChanged gets invoked first and then the processWritableBookiesChanged method second. As soon as processReadOnlyBookiesChanged is invoked it unblocks the waitForReadOnlyBookie() in the test which immediately creates a new ledger, before the processWritableBookiesChanged is invoked, allowing the ledger to be created thus failing the test.

Introducing a Thread.sleep before creating the ledger allows both watches to fire first, though it's not a perfect solution.

dlg99 (approved):

Vanlightly:
rerun failure checks

Vanlightly:
rerun failure checks

eolivelli (approved):
LGTM

Vanlightly:
rerun failure checks

eolivelli:
cc @fpj

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerRecoveryOp.java
```diff
@@ -104,7 +104,24 @@ LedgerRecoveryOp setEntryListener(ReadEntryListener entryListener) {
...
+                                 - the LAC returned by the current bookie ensemble (could be -1)
+                                 - the first entry id of the current ensemble - 1.
+                                 */
+                                Long lastEnsembleEntryId = lh.getVersionedLedgerMetadata()
+                                        .getValue()
+                                        .getAllEnsembles()
+                                        .lastEntry()
+                                        .getKey();
+
+                                lh.lastAddPushed = lh.lastAddConfirmed = Math.max(data.getLastAddConfirmed(),
```
fpj, line 122: Thanks for the patch, @Vanlightly . I understand that you are trying to lower bound the value of last add pushed by getting the last entry id from the last ensemble in the ledger history. I'd appreciate if you could clarify a couple of quick things for me:

1- Why is it `lastEnsembleEntryId - 1` and not `lastEnsembleEntryId`? I'm missing why we need the `-1`.
2- The PR description mentions the following:

>  The main issue is that if an ensemble change occurs when writing back entries, it will try and create a new ensemble with first entry id of 0. This causes an IllegalStateException as there is a check before the CAS metadata op to ensure that the ensemble does not have an entry id lower than an existing ensemble.

I'm not sure I understand why the client would create a new ensemble with the first entry id being zero. Shouldn't the first entry be the entry id of the last recorded ensemble + 1? Is it zero because because `readLastConfirmedDataComplete` is returning `-1` we are adding 1 to it?

Vanlightly: The recovery process begins its reads/write-backs at LAC + 1. If the LAC reads were not able to successfully get a value, then the scope of recovery should be the current ensemble only. So to target the current ensemble only, we take its first entry id - 1 as the LAC (as recovery starts at LAC + 1).

The issue #2615 goes into detail about a situation where a client starts the recovery process with LAC=-1 (even though there are multiple ensembles), a write fails during this process causing an ensemble change where the client attempts to append an ensemble with first entry id of 0 (even though ensembles exist with higher first entry ids).

fpj (approved):

hsaputra:
Will merge this PR if no one has more comments by end of days PST. Thanks

PR #2658, opened 2021-03-23
fix always select the same region set bug for RegionAwareEnsemblePlacementPolicy

hangc0276:
### Motivation
When use `RegionAwareEnsemblePlacementPolicy` to select ensemble set, it always select the same region set for all ledger creation.

The reason is that it traverse the availableRegions set in the same order for ensemble selection and select bookies in the same region set, which lead the load not balance between regions.
```java
for (String region: availableRegions) {
        regionsWiseAllocation.put(region, Pair.of(0, 0));
}
...

for (Map.Entry<String, Pair<Integer, Integer>> regionEntry: regionsWiseAllocation.entrySet()) {
    // do selection
}
```

### Changes
1. introduce `regionIndex` field to mark next select index for available region list, avoiding traverse from the same position for each ensemble selection.

I will add the test latter.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RegionAwareEnsemblePlacementPolicy.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RegionAwareEnsemblePlacementPolicy.java
@@ -25,7 +25,6 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import java.util.Random;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RegionAwareEnsemblePlacementPolicy.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RegionAwareEnsemblePlacementPolicy.java
@@ -78,7 +78,7 @@ public class RegionAwareEnsemblePlacementPolicy extends RackawareEnsemblePlaceme
     protected boolean enableValidation = true;
     protected boolean enforceDurabilityInReplace = false;
     protected Feature disableDurabilityFeature;
-    private int regionIndex = new Random().nextInt(Integer.MAX_VALUE);
+    private int lastRegionIndex = 0;
 
     RegionAwareEnsemblePlacementPolicy() {
         super();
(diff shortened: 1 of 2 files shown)


hangc0276:
rerun failure checks

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RegionAwareEnsemblePlacementPolicy.java
```diff
@@ -77,6 +78,7 @@
     protected boolean enableValidation = true;
     protected boolean enforceDurabilityInReplace = false;
     protected Feature disableDurabilityFeature;
+    private int regionIndex = new Random().nextInt(Integer.MAX_VALUE);
```
eolivelli, line 81: what about "lastRegionIndex" ?

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/TestRegionAwareEnsemblePlacementPolicy.java
```diff
@@ -1480,4 +1480,48 @@ public void testNodeWithFailures() throws Exception {
...
+
+        repp.onClusterChanged(addrs, new HashSet<>());
+        try {
+            List<BookieId> ensemble1 = repp.newEnsemble(3,3,2,
+                null, new HashSet<>()).getResult();
+            assertEquals(ensemble1.size(), 3);
+            List<BookieId> ensemble2 = repp.newEnsemble(3, 3, 2,
+                null, new HashSet<>()).getResult();
+            ensemble1.retainAll(ensemble2);
+            assert(ensemble1.size() >= 1);
```
eolivelli, line 1517: if we depend on a Random number, is this test going to be flaky ? probably not, but I would like to double check your thoughts

eolivelli (changes requested):

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RegionAwareEnsemblePlacementPolicy.java
hangc0276: OK, i renamed to `lastRegionIndex`

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/TestRegionAwareEnsemblePlacementPolicy.java
hangc0276: if the `lastRegionIndex` initialized to specific value, all bookkeeper clients will start with the same region, which may lead to hotpot for regions. However those bookkeeper clients create ledger speed won't keep sync,  starting will the same region won't make much difference. I changed the initialized value to 0.

hangc0276:
rerun failure checks

hangc0276:
rerun failure checks

hangc0276:
@eolivelli @merlimat  @sijie Please take a look, thanks.

eolivelli (approved):
LGTM

merlimat (approved):

Issue #2660, opened 2021-03-24
Bookkeeper client throttling logic is based upon entryId instead of ledgerId


karanmehta93:
**BUG REPORT**

***Describe the bug***

Before accepting the write, the client checks if netty channels to all the bookies are available. The logic is currently based on [entryId](https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java#L1343). However, in BookieClientImpl, we select the actual channel based on [ledgerId](https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieClientImpl.java#L329-L332). 

We use 3 channels per bookie in our prod and one of them was down (due to unknown reasons). Hence about 1/3 of the requests were throttled for no reason. The throttling caused the overall app to slow down.

***To Reproduce***

Not very easy to reproduce. Need to check if netty can be mocked to add an integration test.

***Expected behavior***

Throttling should be based on ledgerId, instead of entryId.

***Screenshots***

N/A

***Additional context***

N/A

dlg99:
@karanmehta93 I think you are correct about the ledgerId. 
Please review https://github.com/apache/bookkeeper/pull/2664 for the fix.

I don't think it will help you with underlying problem which you still have to rootcause. 
The worst this problem does it selects wrong channel and detects it as non-writable. With switch to the ledger Id some other ledger will end up blocked on non-writable channel. 
Non-writable PCBC == non-writable netty channel to the bookie. IIRC, this can be related to slow network, non-responsive bookie, bookie being overloaded and disabling autoread on the channel etc.

jvrao:
> The worst this problem does it selects wrong channel and detects it as non-writable.

Yes, but the worst is because we are just throttling and not failing/changing ensemble. Hence latencies wentup without going through our fastfail model. But you right the underlying problem of why the channel is not writable still needs to be figured out.

PR #2664, opened 2021-03-29
fix: Bookkeeper client throttling logic is based upon entryId instead of ledgerId

dlg99:

Fixes: #2660 

### Changes

isWriteSetWritable() to use ledgerId for the client selection

Master Issue: #2660


eolivelli (approved):
makes sense to me.

didn't we have tests that covered this "feature" ?

dlg99:
> didn't we have tests that covered this "feature" ?

Tests use client config with numChannelsPerBookie == 1. :( All tests.
I think this can be tracked/fixed separately.

karanmehta93 (approved):
LGTM, Thanks for taking this up!

ivankelly (approved):
Change looks good, but this stuff really needed to be unit tested. 
Also, commit/PR message needs to contain the rationale for the change.

pkumar-singh:
We can get rid of the template text from PR message.

Issue #2640, opened 2021-03-05
BP-43: Gradle migration


pkumar-singh:
**BP**

This is the master ticket for tracking BP-43 :
This BP intends to migrate existing maven based build and release pipeline to gradle based on. 

Proposal PR - https://github.com/apache/bookkeeper/pull/2639

pkumar-singh:
Addressed all the comments.

PR #2666, opened 2021-03-30
ISSUE-2640: BP-43: Build distributions with gradle

sursingh:
### Motivation

Migrate bookkeeper to gradle.

### Changes

added gradle build support for building distribution packages
  * bookkeeper-dist-bkctl
  * bookkeeper-dist-server
  * bookkeeper-dist-all
  * bookkeeper-dist-src
  
 Also added build files for needed dependencies

Master Issue: #2640

#### Building:
```
./gradlew :bookkeeper-dist-src:distTar :bookkeeper-dist-server:distTar :bookkeeper-dist-bkctl:distTar :bookkeeper-dist-all:distTar
```

The package can be found under corresponding build/distributions
```
$ find bookkeeper-dist -name *.tar.gz -path *distributions*
bookkeeper-dist/bkctl/build/distributions/bkctl-4.14.0-SNAPSHOT.tar.gz
bookkeeper-dist/all/build/distributions/bookkeeper-all-4.14.0-SNAPSHOT.tar.gz
bookkeeper-dist/server/build/distributions/bookkeeper-server-4.14.0-SNAPSHOT.tar.gz
bookkeeper-dist/build/distributions/bookkeeper-4.14.0-SNAPSHOT-src.tar.gz
```


bookkeeper-dist/all/build.gradle
```diff
@@ -0,0 +1,123 @@
...
+        "bouncycastle-1.0.2/LICENSE.html",
+        "jsr-305/LICENSE",
+        "netty-4.1.50.Final/*",
+        "paranamer-2.8/LICENSE.txt",
+        "protobuf-3.14.0/LICENSE",
+        "protobuf-3.12.0/LICENSE",
+        "scala-library-2.11.7/LICENSE.md",
+        "scala-parser-combinators_2.11-1.0.4/LICENSE.md",
+        "scala-reflect-2.11.8/LICENSE.md",
+        "slf4j-1.7.25/LICENSE.txt",
```
merlimat, line 69: Is there any other way to avoid specifying the exact version here, so that we only need to update the version in 1 place?

merlimat (approved):

bookkeeper-dist/all/build.gradle
sursingh: We can add a wildcard for these. But do we want to tie the version of the packaged dependency? That will be a little more involved.

sursingh: Also couldn't see where we are using scala library. So I don't think we should be packaging these licenses. But we should take this up after we move completely to gradle.

eolivelli (approved):
LGTM

bookkeeper-dist/all/build.gradle
sursingh: Another option would be use a https://github.com/jk1/Gradle-License-Report plugin. With this we can automate the license report and also avoid the need to maintain license files.

PR #2668, opened 2021-03-30
Fix logic in Bookkeeper forceAllowCompaction to run only when force i…

klwilson227:
Without the attached fix, when forceAllowCompaction and isForceGCAllowWhenNoSpace are set. The majorCompaction runs with every garbage collection. This results due to the fact that enableMajorCompaction which takes a negative interval into account to disable compaction always returns true since the difference in time will always be larger than -1. The time check must be "and" with the enableMajorCompaciton for the time check to be valid. And the force flag should also only be checked when the isForceGCAllowWhenNoSpace is activated. The attached PR makes the necessary adjustments.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
@@ -356,8 +356,9 @@ public class GarbageCollectorThread extends SafeRunnable {
         }
 
         long curTime = System.currentTimeMillis();
-        if ((isForceMajorCompactionAllow || enableMajorCompaction) && (!suspendMajor)
-            && (force || curTime - lastMajorCompactionTime > majorCompactionInterval)) {
+        if (((isForceMajorCompactionAllow && force)
+                || (enableMajorCompaction && curTime - lastMajorCompactionTime > majorCompactionInterval))
+                && (!suspendMajor)) {
             // enter major compaction
             LOG.info("Enter major compaction, suspendMajor {}", suspendMajor);
             majorCompacting.set(true);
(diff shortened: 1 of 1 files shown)


eolivelli:
thank you very much for this patch

There are test failures, can you please check?

for instance:
```
Error:  org.apache.bookkeeper.bookie.CompactionByEntriesTest.testForceGarbageCollection  Time elapsed: 0.941 s  <<< FAILURE!
java.lang.AssertionError: Minor or major compaction did not trigger even on forcing.
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.apache.bookkeeper.bookie.CompactionTest.lambda$testForceGarbageCollection$0(CompactionTest.java:312)
	at org.apache.bookkeeper.meta.MetadataDrivers.lambda$runFunctionWithLedgerManagerFactory$2(MetadataDrivers.java:389)
	at org.apache.bookkeeper.meta.MetadataDrivers.runFunctionWithMetadataBookieDriver(MetadataDrivers.java:347)
	at org.apache.bookkeeper.meta.MetadataDrivers.runFunctionWithLedgerManagerFactory(MetadataDrivers.java:387)
	at org.apache.bookkeeper.bookie.CompactionTest.testForceGarbageCollection(CompactionTest.java:292)
	at org.apache.bookkeeper.bookie.CompactionTest.testForceGarbageCollection(CompactionTest.java:255)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
```

also, can you please verify if there is a test that covers your change, or add a new test case ?

klwilson227:
I’ll take a look.

klwilson227:
Re-opening new PR, as this does not have the intended update after rebase to squash extra commits.

PR #2670, opened 2021-03-30
Issue 2638: Compaction Limits

klwilson227:
This resolves issue 2638, by allowing a customer to set a limit on the duration of the major and minor compaction runs. This allows the customer to balance the needs for compaction against the need for normal garbage collection. Unbounded execution of compaction can starve the garbage collection process, and may lead to extended periods of high disk IO. By setting a upper bound on the execution duration for the garbage collection, we can ensure a balance is maintained between both GC and Compaction processes. 

Logging has been improved to allow a user to see the complete usage of the buckets, vs. those that were compacted, so that they can decide on any further tuning that may be required. 

Test cases have been added to ensure limits are enforced properly and that defaulting of the values is correctly implemented.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
@@ -71,12 +71,14 @@ public class GarbageCollectorThread extends SafeRunnable {
     boolean enableMinorCompaction = false;
     final double minorCompactionThreshold;
     final long minorCompactionInterval;
+    final long minorCompactionLimit;
     long lastMinorCompactionTime;
 
     boolean isForceMajorCompactionAllow = false;
     boolean enableMajorCompaction = false;
     final double majorCompactionThreshold;
     final long majorCompactionInterval;
+    long majorCompactionLimit;
     long lastMajorCompactionTime;
 
     @Getter
@@ -180,6 +182,9 @@ public class GarbageCollectorThread extends SafeRunnable {
         majorCompactionThreshold = conf.getMajorCompactionThreshold();
         majorCompactionInterval = conf.getMajorCompactionInterval() * SECOND;
         isForceGCAllowWhenNoSpace = conf.getIsForceGCAllowWhenNoSpace();
+        majorCompactionLimit = conf.getMajorCompactionLimitMs();
+        minorCompactionLimit = conf.getMinorCompactionLimitMs();
+
         boolean isForceAllowCompaction = conf.isForceAllowCompaction();
 
         AbstractLogCompactor.LogRemovalListener remover = new AbstractLogCompactor.LogRemovalListener() {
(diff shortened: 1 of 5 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -452,31 +457,48 @@ void doCompactEntryLogs(double threshold) {
 
         final int numBuckets = 10;
         int[] entryLogUsageBuckets = new int[numBuckets];
+        int[] compactedBuckets = new int[numBuckets];
+
+        long start = System.currentTimeMillis();
+        long end = start;
+        long timeDiff = 0;
```
codelipenghui, line 464: Do we need the `end` and the `timeDiff` here? Seems we can only break the following loop if (currentTime-start) >= limit?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -452,31 +457,48 @@ void doCompactEntryLogs(double threshold) {
...
 
-            if (meta.getUsage() >= threshold) {
+            if (timeDiff < limit) {
+                end = System.currentTimeMillis();
+                timeDiff = end - start;
+            }
+            if (meta.getUsage() >= threshold || (limit > 0 && timeDiff > limit) || !running) {
+                // We allow the usage limit calculation to continue so that we get a accurate
+                // report of where the usage was prior to running compaction.
                 continue;
```
codelipenghui, line 479: We should break the loop if exceed the max compaction time? all the remaining logs will be skipped.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
klwilson227: The timediff is held for reporting after loop exit when debugging is enabled.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
klwilson227: The consumer, would not get any indication of how may logs were not compacted if we skip out of loop. This ensures that the exit report of the usage buckets give a complete picture of what was there and what was compacted. The usage gathering from what I saw as extremely fast compared to the compaction. The logs show that this does not take a appreciable amount of time.  However, a consumer whom is tuning the system may need this insight.

conf/bk_server.conf
```diff
@@ -500,6 +500,9 @@ ledgerDirectories=/tmp/bk-data
 # If it is set to less than zero, the minor compaction is disabled.
 # minorCompactionInterval=3600
 
+# Limit in milliseconds to run minor Compaction. Defaults to -1 run indefinitely.
+# minorCompactionLimitMs=-1
```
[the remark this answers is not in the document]
klwilson227: I will make the updates throughout.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
codelipenghui: Ok  I see.

codelipenghui (approved):

sijie (approved):

jiazhai (approved):

PR #2672, opened 2021-03-31
fix 2655 bookie will be down (out of memory), when there are a lots of ledgers under /ledgers/underreplication/ledgers to replication

baomingyu:

### Motivation

In some exception branch, lh is not be unregister from listener,  so I make this change.

### Changes
`if (rc != BKException.Code.OK) {`
`                        if (lh != null && !doRecovery) {`
`                            bk.getClientCtx().getLedgerManager()`
`                                    .unregisterLedgerMetadataListener(ledgerId, lh);`
`                        }`
`                        openComplete(bk.getReturnRc(BKException.Code.ReadException), null);`
`                    }`

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerOpenOp.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerOpenOp.java
@@ -212,6 +212,10 @@ class LedgerOpenOp {
                 public void readLastConfirmedComplete(int rc,
                         long lastConfirmed, Object ctx) {
                     if (rc != BKException.Code.OK) {
+                        if (lh != null && !doRecovery) {
+                            bk.getClientCtx().getLedgerManager()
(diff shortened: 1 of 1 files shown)


eolivelli (changes requested):
thanks for providing your fix.
The fix makes sense to me.

From my point of view it is hard to see how this fix is related to the original problem.
do you have any log, memory dump... or any other hint that points into this direction ?

baomingyu:
> thanks for providing your fix.
> The fix makes sense to me.
> 
> From my point of view it is hard to see how this fix is related to the original problem.
> do you have any log, memory dump... or any other hint that points into this direction ?

This scene is realy hard to see, In my scene , I change  3 3 2 to 2 2 2  , and replicaiton will mark this scene .

my dump file in mat
image

StevenLuMT:

shoothzj:
@baomingyu Thanks for your contribution, could you please update this PR?

PR #2677, opened 2021-04-02
[Bug Fix] Fix Concurrency issue with volatile fields

aahmed-se:
### Motivation

The ConcurrentMap are doing increment and decrement operations on volatile primitive fields. This is not threadsafe.

### Changes

* Use AtomicIntegerFieldUpdater to replace increment and decrement operations.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/util/collections/ConcurrentLongHashMap.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/util/collections/ConcurrentLongHashMap.java
@@ -27,6 +27,7 @@ import com.google.common.collect.Lists;
 
 import java.util.Arrays;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 import java.util.concurrent.locks.StampedLock;
 import java.util.function.LongFunction;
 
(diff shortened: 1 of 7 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/util/collections/ConcurrentLongHashSet.java
```diff
@@ -339,15 +344,15 @@ public void forEach(ConsumerLong processor) {
                 }
 
                 // Go through all the buckets for this section
-                for (int bucket = 0; bucket < table.length; bucket++) {
-                    long storedItem = table[bucket];
+                for (long l : table) {
```
eolivelli, line 347: Thi s is creating an Iterator, can we save this allocation? 
Also the JVM is very good in dealing with counted loops

eolivelli (changes requested):
Very good.
I left one comment.
Do we have the same code in Pulsar?

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/collections/ConcurrentLongHashMap.java
```diff
@@ -306,12 +311,12 @@ V put(long key, V value, int keyHash, boolean onlyIfAbsent, LongFunction<V> valu
                     if (storedKey == key) {
                         if (storedValue == EmptyValue) {
                             values[bucket] = value != null ? value : valueProvider.apply(key);
-                            ++size;
+                            sizeUpdater.incrementAndGet(this);
```
pkumar-singh, line 314: Why there is thread safety concern here while lock is acquired at line 298.  `long stamp = writeLock();`

merlimat (changes requested):
The current code is thread safe, the increments of counters are done while holding the write lock

aahmed-se:
ok will close is pr then.

PR #2678, opened 2021-04-05
ISSUE-2640: BP-43 create shaded jar for bookkeeper-server and distrib…

pkumar-singh:
…utedlog-core

### Motivation
**Gradle migration** 
With maven shaded Jar was created for bookkeeper-sever as well as distributedlog-core.
Though shaded jar was not being used by anybody.
There has been some difference in layout of shaded jar produced by maven and gradles.

Maven produced shaded jar just shades bookkeeper classes and leaves transitive dependency unshaded and as part of dependency in POM file. This behavior is not supported in gradle shadow plugin. Therefore gradle produced shaded jar shades not only bookkeer-server but also its transitive and direct dependencies and leaving nothing in pom file as dependency. 

### Changes
Create gradle build for shading bookkeeper-sever and distributedlog-core

Master Issue: #2640

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -20,3 +20,5 @@
 #
 
 protobufPluginVersion=0.8.15
+shadowPluginVersion=6.1.0
+licenseGradlePluginVersion=0.15.0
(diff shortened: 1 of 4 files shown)


hsaputra (approved):
LGTM +1
This should safe to merge in.

merlimat (approved):

Issue #2669, opened 2021-03-30
Update Netty to 4.1.60.final 


eolivelli:
***Describe the bug***

Upgrade netty to resolve security issues in the current version.

https://nvd.nist.gov/vuln/detail/CVE-2021-21295
https://nvd.nist.gov/vuln/detail/CVE-2021-21290

We upgraded Netty in Pulsar to latest version see
https://github.com/apache/pulsar/pull/10073

ruanwenjun:
Maybe I can do this upgrade task.

PR #2679, opened 2021-04-06
ISSUE-2669:Upgrade Netty to 4.1.60.final

ruanwenjun:
### Motivation

see #2669 

### Changes

upgrade the netty version to 4.1.60.final

Master Issue: #2669


pom.xml
```diff
@@ -866,7 +866,6 @@
           <source>${javac.target}</source>
           <target>${javac.target}</target>
           <compilerArgs>
-            <compilerArg>-Werror</compilerArg>
```
dlg99, line 869: why do you need this change?  This is a build-wide setting.

If netty upgrade resulted in some deprecation warnings you need to fix them

dlg99 (changes requested):
Please undo the werror change, fix anything that blocks the build after that, and update the license files.

pom.xml
ruanwenjun: Yeal, I try to restore this configuration and update the dependency license files. And now I get a problem about `TestTLS`.

dlg99:
Netty is upgraded to 4.1.63 in https://github.com/apache/bookkeeper/pull/2696 ; I'll close this one

PR #2686, opened 2021-04-12
Rocksdb tombstones

mauricebarnum:
### Motivation

After deleting many ledgers, seeking to the end of the RocksDB metadata can take a long time and trigger timeouts upstream. Address this by improving the seek logic as well as compacting out tombstones in situations where we've just deleted many entries.  This affects the entry location index and the ledger metadata index.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/LedgerMetadataIndex.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/LedgerMetadataIndex.java
@@ -228,13 +228,16 @@ public class LedgerMetadataIndex implements Closeable {
 
     public void removeDeletedLedgers() throws IOException {
         LongWrapper key = LongWrapper.get();
+        final byte[] startKey = new byte[key.array.length];
 
         int deletedLedgers = 0;
         while (!pendingDeletedLedgers.isEmpty()) {
             long ledgerId = pendingDeletedLedgers.poll();
             key.set(ledgerId);
             ledgersDb.delete(key.array);
-            deletedLedgers++;
+            if (deletedLedgers++ == 0) {
+                System.arraycopy(key.array, 0, startKey, 0, startKey.length);
+            }
         }
 
         if (log.isDebugEnabled()) {
@@ -242,6 +245,9 @@ public class LedgerMetadataIndex implements Closeable {
         }
 
         ledgersDb.sync();
+        if (deletedLedgers != 0) {
+            ledgersDb.compact(startKey, key.array);
+        }
         key.recycle();
     }
 
--- a/bookkeeper-dist/src/main/resources/LICENSE-all.bin.txt
+++ b/bookkeeper-dist/src/main/resources/LICENSE-all.bin.txt
@@ -253,7 +253,7 @@ Apache Software License, Version 2.
 - lib/org.eclipse.jetty-jetty-server-9.4.33.v20201020.jar [22]
 - lib/org.eclipse.jetty-jetty-servlet-9.4.33.v20201020.jar [22]
 - lib/org.eclipse.jetty-jetty-util-9.4.33.v20201020.jar [22]
-- lib/org.rocksdb-rocksdbjni-6.10.2.jar [23]
+- lib/org.rocksdb-rocksdbjni-6.16.4.jar [23]
 - lib/com.beust-jcommander-1.78.jar [24]
 - lib/com.yahoo.datasketches-memory-0.8.3.jar [25]
 - lib/com.yahoo.datasketches-sketches-core-0.8.3.jar [25]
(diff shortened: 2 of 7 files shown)


pkumar-singh (approved):

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/LedgerMetadataIndex.java
```diff
@@ -228,20 +228,28 @@ public void flush() throws IOException {
...
+            System.arraycopy(key.array, 0, endKey, 0, endKey.length);
         }
 
         if (log.isDebugEnabled()) {
             log.debug("Persisting deletes of ledgers {}", deletedLedgers);
         }
 
         ledgersDb.sync();
+        if (deletedLedgers != 0) {
+            ledgersDb.compact(startKey, endKey);
```
dlg99, line 251: as I understood, this method runs on each checkpoint.
How expensive is this compaction in terms of time?
IIRC checkpointing happens on entry log rotation/gets into the addEntry latency, how much this increases p99.9 under load?

I am not familiar with RocksDB nuances. Looking at https://github.com/facebook/rocksdb/wiki/Manual-Compaction

> In case of universal and FIFO compaction styles, the begin and end arguments are ignored and all files are compacted. Also, files in each level are compacted and left in the same level. For leveled compaction style, all files containing keys in the given range are compacted to the last level containing files. 

Sounds like this is potentially expensive operation creating additional IO.

Would it be better run on a separate thread at lower frequency, better yet as a part of bookie's garbage collection?

dlg99 (changes requested):
Couple of comments.

dlg99:
regarding the failing checks, you may want to rebase on the current master to see if https://github.com/apache/bookkeeper/pull/2688 helps.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/LedgerMetadataIndex.java
```diff
@@ -228,20 +228,28 @@ public void flush() throws IOException {
...
         int deletedLedgers = 0;
         while (!pendingDeletedLedgers.isEmpty()) {
             long ledgerId = pendingDeletedLedgers.poll();
             key.set(ledgerId);
             ledgersDb.delete(key.array);
-            deletedLedgers++;
+            if (deletedLedgers++ == 0) {
+                System.arraycopy(key.array, 0, startKey, 0, startKey.length);
+            }
+            System.arraycopy(key.array, 0, endKey, 0, endKey.length);
```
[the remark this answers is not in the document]
mauricebarnum: good catch. better: only startKey needs to be copied. endKey can be used directly.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/LedgerMetadataIndex.java
mauricebarnum: The two methods I modified are only called by the sync thread, so already in the background.

As far as additional IO: potentially yes.  Each compaction will need to flush the memtable and then read each SST that contains a key in the range. Each key is likely present in at most two SSTs: the delete tombstone and the original put. All of the affected SSTs are then rewritten.  This is background work with regards to the get/put/delete operations but will stall the sync thread.  In the case where writes to ledgers are coming in fast, most of the keys should be adjacent or nearly so.

dlg99 (approved):
lgtm

hsaputra:
Will merge this EOD PST if no more comments. Thanks

hangc0276:
I have noticed that index deletion is sometimes taking around 60 seconds which cause the CPU to spike 100%
```
[2022-02-28T07:25:42.531Z] INFO db-storage-cleanup-10-1 EntryLocationIndex:191 Deleting indexes for ledgers: [3385184, 3385239, 3385159, 3385142, 3385124, 3385193, 3384879, 3385165, 3385916]
[2022-02-28T07:26:34.089Z] INFO db-storage-cleanup-10-1 EntryLocationIndex:266 Deleted indexes for 201065 entries from 9 ledgers in 51.557 seconds
[2022-02-28T07:40:42.534Z] INFO db-storage-cleanup-10-1 EntryLocationIndex:191 Deleting indexes for ledgers: [3385379, 3385367, 3385718, 3385365, 3385412, 3385167, 3385357, 3386141]
[2022-02-28T07:41:47.867Z] INFO db-storage-cleanup-10-1 EntryLocationIndex:266 Deleted indexes for 134590 entries from 8 ledgers in 65.332 seconds
```
@mauricebarnum @dlg99 @eolivelli   Do you have any ideas?

hangc0276:
RocksDB compaction is a heavy operation and the checkpoint will be triggered in high frequency, which cause `db-storage-cleanup` thread always into high load, and make the cpu keeps 100%.

IMO, we should use another thread to do RocksDB compaction and makes the compaction triggered by other conditions instead of checkpoint.

mauricebarnum:
The motivation to call `compactRange` was to quickly drop all of the rocksdb entries in a range when deleting a bunch of ledgers in GC so that seeking wouldn't run into all of the tombstones: the keys will be grouped together and much of the work should simply be deleting SSTs in the range.   A smarter approach for "delete a bunch of ledgers" to cause RocksDB to schedule a background compaction "soon".  When I made this change, stalling GC shouldn't didn't seem to be too bad.  Forcing synchronous compaction during checkpointing was a mistake.

hangc0276:
For this issue, IMO we'd better divide into two steps:
1. remove the compactRange logic in during checkpoint
2. Figure out smarter solution for deleted entry compaction in RocksDB

Do you have any ideas? @mauricebarnum

PR #2687, opened 2021-04-14
ISSUE-2640: BP-43: Fix test cases using gradle build

sursingh:
### Motivation
Migrate bookkeeper build to gradle. Some test cases were failing while running test with gradle

### Changes
* Fix dependencies
* jassTmpDir was created under `target` folder. This fails while running tests using gradle. Instead we should use `createTempDirectory` to create temp folder

Master Issue: #2640


sursingh:
rerun failure checks

PR #2689, opened 2021-04-14
ISSUE-2640: BP-43: Fix test cases that fail when running with gradle

sursingh:
### Motivation

Migrate bookkeeper build to gradle. Some test cases were failing while running test with gradle

### Changes

* Fix dependencies to exclude some transitive dependencies.
* jassTmpDir was created under target folder. This fails while running tests using gradle. Instead we should use createTempDirectory to create temp folder


hsaputra (approved):
LGTM +1

merlimat (approved):

PR #2708, opened 2021-05-07
BookKeeper-Client config to write ledger metadata with configured version

rdhabalia:
### Motivation

We need a way to use old bookie which doesn't support metadata-version-3 and upgrade bookkeeper-client with latest version. Right now, bk-4.12 writes ledger metadata with version-3 which will not be supported by bookie and replicator and replicator fails with below error
```
2021-05-07.00:43:53.115 [main-EventThread] ERROR
                o.a.b.meta.AbstractZkLedgerManager   - Could not parse ledger metadata for ledger: 123456
java.io.IOException: Metadata version not compatible. Expected between 0 and 2, but got 3
        at org.apache.bookkeeper.client.LedgerMetadata.parseConfig(LedgerMetadata.java:465)
        at org.apache.bookkeeper.meta.AbstractZkLedgerManager$3.processResult(AbstractZkLedgerManager.java:414)
        at org.apache.bookkeeper.zookeeper.ZooKeeperClient$19$1.processResult(ZooKeeperClient.java:994)
        at org.apache.zookeeper.ClientCnxn$EventThread.processEvent(ClientCnxn.java:572)
        at org.apache.zookeeper.ClientCnxn$EventThread.run(ClientCnxn.java:508)

```
So, until one upgrades the bookie version, allow bk-client to use the appropriate metadata version.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadataBuilder.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadataBuilder.java
@@ -20,6 +20,7 @@ package org.apache.bookkeeper.client;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkState;
 import static org.apache.bookkeeper.meta.LedgerMetadataSerDe.CURRENT_METADATA_FORMAT_VERSION;
+import static org.apache.bookkeeper.meta.LedgerMetadataSerDe.METADATA_FORMAT_VERSION_1;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableMap;
@@ -108,6 +109,9 @@ public class LedgerMetadataBuilder {
     }
 
     public LedgerMetadataBuilder withMetadataFormatVersion(int version) {
+        if (version < METADATA_FORMAT_VERSION_1 || version > CURRENT_METADATA_FORMAT_VERSION) {
+            return this;
+        }
         this.metadataFormatVersion = version;
         return this;
     }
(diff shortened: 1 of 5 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
```diff
@@ -440,6 +441,25 @@ public String getLedgerManagerFactoryClassName() {
         return getString(LEDGER_MANAGER_FACTORY_CLASS);
     }
 
+    /**
+     * Set Ledger metadata format version.
+     *
+     * @param metadataFormatVersion
+     *          Ledger metadata format version. pass -1 to use default version
+     */
+    public void SetLedgerMetadataFormatVersion(int metadataFormatVersion) {
```
merlimat, line 450:
```suggestion
    public void setLedgerMetadataFormatVersion(int metadataFormatVersion) {
```

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerMetadataBuilder.java
```diff
@@ -108,6 +109,9 @@ public LedgerMetadataBuilder withId(long ledgerId) {
     }
 
     public LedgerMetadataBuilder withMetadataFormatVersion(int version) {
+		if (version < METADATA_FORMAT_VERSION_1 && version > CURRENT_METADATA_FORMAT_VERSION) {
```
merlimat, line 112: indenting seems to be in tab and not spaces (also above)

merlimat (approved):

sijie (approved):

PR #2710, opened 2021-05-07
Impose a memory limit on the bookie journal

merlimat:
### Motivation

Currently, the journal has a configurable max queue size, with default to 10,000 entries. This can be used as a proxy measure to control the max amount of memory that has to be retained from the entries that are in the journal queue.

This is not very flexible and there are few potential problems that arise: 
 1. If using multiple journals, each one of them gets 10,000 entries
 2. entries can be of very different sizes, from few bytes to several MBs, so it can be difficult to estimate the memory usage
 3. if entries are big, it can cause the bookie to go OOM when the queue gets filled up

Instead (or additionally) of number of entries, we should use the total amount of memory as the metric for enabling the backpressure.

### Changes

Added `MemoryLimitController` class, already used in Pulsar (https://github.com/apache/pulsar/pull/8965) and use it to block the thread adding to the journal queue when memory is exhausted.

Defaulting to use 5% of max direct memory (eg: 100MB out of 2GB)

### DIFF
--- a/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/MemoryLimitController.java
+++ b/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/MemoryLimitController.java
@@ -1,0 +1,84 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.bookkeeper.common.util;
+
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.ReentrantLock;
+
+public class MemoryLimitController {
+
+    private final long memoryLimit;
+    private final AtomicLong currentUsage = new AtomicLong();
+    private final ReentrantLock mutex = new ReentrantLock(false);
+    private final Condition condition = mutex.newCondition();
+
+    public MemoryLimitController(long memoryLimitBytes) {
+        this.memoryLimit = memoryLimitBytes;
+    }
+
+    public boolean tryReserveMemory(long size) {
+        while (true) {
+            long current = currentUsage.get();
+            long newUsage = current + size;
+
+            // We allow one request to go over the limit, to make the notification
+            // path simpler and more efficient
+            if (current > memoryLimit && memoryLimit > 0) {
+                return false;
+            }
+
+            if (currentUsage.compareAndSet(current, newUsage)) {
+                return true;
+            }
+        }
+    }
+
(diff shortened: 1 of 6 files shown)


codelipenghui (approved):

jiazhai (approved):

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieJournalMaxMemoryTest.java
```diff
@@ -0,0 +1,109 @@
...
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+import static org.powermock.api.mockito.PowerMockito.whenNew;
+
+/**
+ * Test the bookie journal PageCache flush interval.
```
eolivelli, line 59: nit: fix javadoc

eolivelli (changes requested):
is it possible to disable this feature in case there is some problem ?

like configuring -1 as memory limit

merlimat:
> like configuring -1 as memory limit in order to disable it

As documented in the `bk_server.conf` setting to 0, disables the limit

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/MemoryLimitController.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    public boolean tryReserveMemory(long size) {
+        while (true) {
+            long current = currentUsage.get();
+            long newUsage = current + size;
+
+            // We allow one request to go over the limit, to make the notification
+            // path simpler and more efficient
+            if (current > memoryLimit && memoryLimit > 0) {
```
eolivelli, line 43: so here if we set memoryLimit = 0 we are not entering this branch

isn't it simpler to move 
```
if (memoryLimit <=0) {
    return true;
}
```
at the beginning of this method and a similar check in the beginning of `releaseMemory ` ?

merlimat: The reason is mostly to leave open the possibility to track how much memory is being used, even if a limit is not being enforced.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieJournalMaxMemoryTest.java
merlimat: Ups, fixed it

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/MemoryLimitController.java
merlimat: ... which raises the good point that we should have such metric exported :).. I'll quickly add it here

merlimat: Added metrics

merlimat:
@eolivelli PTAL again

merlimat:
@dlg99 I marked this on purpose for 4.14

conf/bk_server.conf
```diff
@@ -358,6 +358,11 @@ journalDirectories=/tmp/bk-txn
 # Set the size of the journal queue.
 # journalQueueSize=10000
 
+# Set the max amount of memory that can be used by the journal.
+# If empty, this will be set to use 5% of available direct memory
+# Setting it to 0, it will disable the max memory control for the journal.
+# journalMaxMemorySizeMb=
```
hsaputra, line 364: So looks like this will be additional limit to t he max Journal Q size?

merlimat: Correct. that would still be in place

eolivelli (approved):
Lgtm 
@dlg99 please merge before packaging 4.14

PR #2714, opened 2021-05-13
Include gradle files into the source artifact for releases, exclude site2/**

dlg99:
### Motivation

gradle files weren't included into the source artifact of the release
site2/ was included.

### Changes

Added gradle into the include pattern.
Added site2 into the exclude patterns.


eolivelli (approved):
LGTM merging

PR #2717, opened 2021-05-22
Turn Bookie into an interface

pkumar-singh:
### Motivation

Bookie was previously a concrete class that was used and abused all
over the place, especially in tests. A classic example of the God
object antipattern. The extensive use in tests, resulted in test cases
which spin up many instances of the whole system, which is very heavy
and very slow, especially when trying to unit tests a particular
feature.

This change is the first step to resolving this situation. Bookie is
now an interface, implemented by BookieImpl. Subsequent changes will
break out parts of the interface, cleanup calls and add dependency
injection.


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -16,154 +15,80 @@
...
-
-    final SyncThread syncThread;
-    final LedgerManagerFactory ledgerManagerFactory;
-    final LedgerManager ledgerManager;
-    final LedgerStorage ledgerStorage;
-    final List<Journal> journals;
-
-    final HandleFactory handles;
-    final boolean entryLogPerLedgerEnabled;
+import com.google.common.util.concurrent.SettableFuture;
```
eolivelli, line 25: thinking about the future...
it would be also great to get rid of SettableFuture

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
```diff
@@ -16,154 +15,80 @@
...
-    private final BookieStats bookieStats;
+    /**
+     * Fences a ledger. From this point on, clients will be unable to
+     * write to this ledger. Only recoveryAddEntry will be
+     * able to add entries to the ledger.
+     * This method is idempotent. Once a ledger is fenced, it can
+     * never be unfenced. Fencing a fenced ledger has no effect.
+     */
+    // TODO: replace settable future with completablefuture
+    SettableFuture<Boolean> fenceLedger(long ledgerId, byte[] masterKey)
```
eolivelli, line 70: yes, that exactly what I meant.

is these TODO to be applied in this patch ?
otherwise I would prefer to not leave TODOs in the code
or at least to create github issues in order to track the work and have references

eolivelli (commented):
very good.

I know that this is a "draft", but I left a couple of comments

thanks for this contribution

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
pkumar-singh: Yeah done.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java
pkumar-singh: Done.

eolivelli (approved):
Looks great

+1

hsaputra (approved):
+1
LGTM

hsaputra:
Will merge this by EOD if no more comment or concern. Thank you

merlimat (approved):

dlg99 (approved):
+1

PR #2718, opened 2021-05-24
fix prometheus metric provider bug and add test to cover label scope …

hangc0276:
### Motivation
After add label for prometheus metric by https://github.com/apache/bookkeeper/pull/2650, it will cause prometheus metric format check failed when no label specified for a statsLogger. The metric list as follow.
```
replication_bookkeeper_client_bookkeeper_client_bookie_watcher_NEW_ENSEMBLE_TIME{success="false",quantile="0.9999", } NaN
```

### Modification
1. add label empty check for `PrometheusTextFormatUtil`
2. add label scope check test cover
3. add prometheus metric regex pattern check in test case

### DIFF
--- a/bookkeeper-stats-providers/prometheus-metrics-provider/src/main/java/org/apache/bookkeeper/stats/prometheus/PrometheusTextFormatUtil.java
+++ b/bookkeeper-stats-providers/prometheus-metrics-provider/src/main/java/org/apache/bookkeeper/stats/prometheus/PrometheusTextFormatUtil.java
@@ -139,25 +139,34 @@ public class PrometheusTextFormatUtil {
         w.append(name)
                 .append("{success=\"").append(success.toString())
                 .append("\",quantile=\"").append(Double.toString(quantile))
-                .append("\", ");
-        writeLabelsNoBraces(w, opStat.getLabels());
+                .append("\"");
+        if (!opStat.getLabels().isEmpty()) {
+            w.append(", ");
+            writeLabelsNoBraces(w, opStat.getLabels());
+        }
         w.append("} ")
                 .append(Double.toString(opStat.getQuantileValue(success, quantile))).append('\n');
     }
 
     private static void writeCount(Writer w, DataSketchesOpStatsLogger opStat, String name, Boolean success)
             throws IOException {
-        w.append(name).append("_count{success=\"").append(success.toString()).append("\", ");
-        writeLabelsNoBraces(w, opStat.getLabels());
-        w.append("\"} ")
+        w.append(name).append("_count{success=\"").append(success.toString()).append("\"");
+        if (!opStat.getLabels().isEmpty()) {
+            w.append(", ");
+            writeLabelsNoBraces(w, opStat.getLabels());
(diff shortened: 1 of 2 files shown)


codelipenghui (approved):

addisonj:
Lgtm

jiazhai (approved):

eolivelli (approved):
Lgtm

merlimat (approved):

bookkeeper-stats-providers/prometheus-metrics-provider/src/main/java/org/apache/bookkeeper/stats/prometheus/PrometheusTextFormatUtil.java
```diff
@@ -139,25 +139,34 @@ private static void writeQuantile(Writer w, DataSketchesOpStatsLogger opStat, St
         w.append(name)
                 .append("{success=\"").append(success.toString())
                 .append("\",quantile=\"").append(Double.toString(quantile))
-                .append("\", ");
-        writeLabelsNoBraces(w, opStat.getLabels());
+                .append("\"");
+        if (!opStat.getLabels().isEmpty()) {
```
dlg99, line 143: is it guaranteed that opStat.getLabels() is never null?

dlg99 (approved):
Overall LGTM.
One question regarding need to nullcheck opStat.getLabels()

bookkeeper-stats-providers/prometheus-metrics-provider/src/main/java/org/apache/bookkeeper/stats/prometheus/PrometheusTextFormatUtil.java
merlimat: Yes, it will return an empty map when there are no labels.

hangc0276:
```Java
// PrometheusMetricsProvider.java
public StatsLogger getStatsLogger(String scope) {
        return new PrometheusStatsLogger(PrometheusMetricsProvider.this, scope, Collections.emptyMap());
    }

// PrometheusStatsLogger.java
PrometheusStatsLogger(PrometheusMetricsProvider provider, String scope, Map<String, String> labels) {
        this.provider = provider;
        this.scope = scope;
        this.labels = labels;
    }

@Override
public OpStatsLogger getOpStatsLogger(String name) {
    return provider.opStats.computeIfAbsent(scopeContext(name), x -> new DataSketchesOpStatsLogger(labels));
}
```
Yes, it will use `Collections.emptyMap()` to init labels when there are no labels.

hangc0276:
Whether we need to include this PR in 4.14.0 release before Pulsar 2.8.0 release?

eolivelli:
Unfortunately 4.14.0 is already completed.

Probably we must cut a 4.14.1 release

@dlg99 ?

PR #2723, opened 2021-05-25
Refactor book keeper cluster test case

pkumar-singh:
### Motivation
BookKeeperClusterTestCase has historically exposed its members to all
subclasses, which would then manipulate them in many ways. There was
an array of objects for configurations, bookieServers, autorecovery,
which implicit linking between the objects based on maps and indices.
Individual subclasses manipulated these arrays.

This makes it hard to add any dependency injection on the objects
managed by BookKeeperClusterTestCase as the objects. To add DI, we
need each object to have a bunch of other objects associated with
it. For example, for each Bookie, we need to create the
Journal. Maintaining these in separate arrays will lead to fragile
tests.

This change encapsulates all the testing objects in a per bookie
object, and only allows manipulation through methods. This will allow
us to group the objects needed for DI clearly.

Disable testFollowBookieAddressChangeTrckingDisabled

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
@@ -504,12 +504,6 @@ public class PerChannelBookieClient extends ChannelInboundHandlerAdapter {
         };
     }
 
-    public static String buildStatsLoggerScopeName(BookieId addr) {
-        StringBuilder nameBuilder = new StringBuilder();
-        nameBuilder.append(addr.toString().replace('.', '_').replace('-', '_').replace(":", "_"));
-        return nameBuilder.toString();
-    }
-
     private void completeOperation(GenericCallback<PerChannelBookieClient> op, int rc) {
         closeLock.readLock().lock();
         try {
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
@@ -504,6 +504,12 @@ public class PerChannelBookieClient extends ChannelInboundHandlerAdapter {
         };
     }
 
+    public static String buildStatsLoggerScopeName(BookieId addr) {
+        StringBuilder nameBuilder = new StringBuilder();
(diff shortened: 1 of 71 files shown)


hsaputra (changes requested):
Looks like some compilation errors happening

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
```diff
@@ -504,6 +504,12 @@ public boolean isSecure() {
         };
     }
 
+    public static String buildStatsLoggerScopeName(BookieId addr) {
+        StringBuilder nameBuilder = new StringBuilder();
+        nameBuilder.append(addr.toString().replace('.', '_').replace('-', '_').replace(":", "_"));
```
hsaputra, line 509: Probably do not need StringBuilder for this.
Just need to return `addr.toString().replace('.', '_').replace('-', '_').replace(":", "_")`

hsaputra (approved):
LGTM +1 , with just small comment.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
```diff
@@ -504,6 +504,12 @@ public boolean isSecure() {
         };
     }
 
+    public static String buildStatsLoggerScopeName(BookieId addr) {
```
merlimat, line 507: The name of this method can be misleading. We already removed the bookie address from the metric names (instead it's added as a label). Sine this appears to be only by tests, we should just move the methods to tests folder.

pkumar-singh: Addressed.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
pkumar-singh: Yes. Addressed.

pkumar-singh:
@merlimat Addressed your comments.

merlimat (approved):

hsaputra:
Thanks for the reviews - will merge this PR by EOD PST if no more comments or concerns

PR #2735, opened 2021-06-10
[Security] Exclude grpc-okhttp dependency

### LINKED ISSUE #2733: [Security] com.squareup.okhttp-okhttp-2.7.4.jar is old and vulnerable
**BUG REPORT**

In library vulnerability scanning, Bookkeeper gets flagged because of com.squareup.okhttp-okhttp-2.7.4.jar . This happens in Sonatype IQ scanning. 

This library version contains several problems:
* it is more than 5 years old
* CVE-2021-0341
* sonatype-2018-0035

This issue is most likely not a real security issue, but a problem that Bookkeeper gets flagged as vulnerable because of the use of okhttp 2.7.4 .

***Expected behavior***

com.squareup.okhttp:okhttp dependency should be upgraded. Newest com.squareup.okhttp:okhttp version is 2.7.5 , however since that is no longer maintained, it would be better to upgrade to okhttp 3.14.9 .

lhotari:
Fixes #2733

### Motivation

The okhttp dependency version 2.7.4 is old and vulnerable. This dependency isn't needed and it causes Bookkeeper to be flagged for security vulnerabilities.

### Changes

- exclude grpc-okhttp dependency which pulls in okhttp 2.7.4 
- update license files

### DIFF
--- a/bookkeeper-dist/src/main/resources/LICENSE-all.bin.txt
+++ b/bookkeeper-dist/src/main/resources/LICENSE-all.bin.txt
@@ -263,8 +263,6 @@ Apache Software License, Version 2.
 - lib/io.opencensus-opencensus-api-0.24.0.jar [30]
 - lib/io.opencensus-opencensus-contrib-http-util-0.24.0.jar [30]
 - lib/io.opencensus-opencensus-proto-0.2.0.jar [30]
-- lib/com.squareup.okhttp-okhttp-2.7.4.jar [31]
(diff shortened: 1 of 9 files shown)


eolivelli (approved):
LGTM

lhotari:
@sijie @merlimat Please review. This is part of the security fixes that were discussed in the Pulsar Community meeting.

sijie (approved):

PR #2739, opened 2021-06-14
[statestore] Cleanup local storage for statestore storage container

sursingh:
### Motivation

While restoring a storage container, we fetch the checkpoint from the
checkpoint store. Currently this checkpoint will never get cleaned up. Every
time we restore the storage container on pod, a new checkpoint will get added.
Over period of time the disk usage keeps going up and eventually we have to
manually delete these stale checkpoints.

### Changes

With this change, we will cleanup the local storage for a storage container
whenever we close the KVStore. This will ensure that stale checkpoints are not
left behind. It is possible that POD may restart before the cleanup can be
done. To avoid these, we will also ensure that local storage for the storage
container is cleaned up before we restore the storage container.

### DIFF
--- a/stream/statelib/src/test/java/org/apache/bookkeeper/statelib/impl/kv/TestStateStore.java
+++ b/stream/statelib/src/test/java/org/apache/bookkeeper/statelib/impl/kv/TestStateStore.java
@@ -53,8 +53,8 @@ import org.rocksdb.Checkpoint;
 public class TestStateStore {
 
     private final String dbName;
-    private final boolean removeLocal;
-    private final boolean removeRemote;
+    private boolean removeLocal;
+    private boolean removeRemote;
 
     private File localDir;
     private File localCheckpointsDir;
@@ -69,6 +69,7 @@ public class TestStateStore {
     private boolean checkpointChecksumEnable;
     private boolean checkpointChecksumCompatible;
     private boolean enableNonChecksumCompatibility;
+    private boolean localStorageCleanup;
 
     public TestStateStore(String dbName,
                           File localDir,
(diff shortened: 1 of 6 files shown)


sursingh:
rerun failure checks

sursingh:
rerun failure checks

stream/statelib/src/test/java/org/apache/bookkeeper/statelib/impl/kv/TestStateStore.java
```diff
@@ -121,6 +123,18 @@ public void enableCheckpoints(boolean enable) {
         }
     }
 
+    public void setRemoveLocal(boolean enable) {
+        removeLocal = enable;
```
hsaputra, line 127: Having this removeLocal and removeRemote to be mutable will not change the state of internal instance of `RocksdbCheckpointTask` which are passed these variables during the init function call, right?

How is the removeLocal and removeRemote variables used beside being passed to the constructor of `RocksdbCheckpointTask` ?

hsaputra (commented):
Did a pass and added question about the update on the test class.

stream/statelib/src/test/java/org/apache/bookkeeper/statelib/impl/kv/TestStateStore.java
sursingh: These variables are only used to initialize the `RocksdbCheckpointTask`. These are mutable to allow use to re-initialize the statestore instance with different options while testing. Also this is only used in test cases.

hsaputra: Ah, yes the `init()` is also called from `restore()` function

hsaputra (approved):
LGTM +1

hsaputra:
Will merge this EOD if no more comments or questions

PR #2631, opened 2021-02-27
Make Bookkeeper FIPS compliant by default

Ghatage:
Descriptions of the changes in this PR:
Mailing list thread: [link](http://mail-archives.apache.org/mod_mbox/bookkeeper-dev/202102.mbox/browser)

### Motivation

FIPS is 'Federal Information Processing Standard'. Basically it's a set of guidelines for security functions such as encryption/decryption/RNG etc. Applications running in FIPS mode are said to be more secure as they adhere to more stringent standards.

Java's security framework is extensible via the JCE (Java crypto extension) allows us to use libraries which implement these functions. In general we use BouncyCastle as it has an active, supportive community and also maintains the fips versions of their libraries.

Pulsar currently has FIPS support but it's not on by default, I had a chat with @jiazhai about it and he mentioned that there is no specific reason as to why we run without it.

This is an attempt to get Apache BookKeeper FIPS compliant by default. If all tests pass and there are no performance regressions, we can merge this in.

### Changes

- Added dependency of FIPS approved BouncyCastle
- Excluded the use of non FIPS BouncyCastle libraries.
- Added flag to start up script which will make it fips compliant.

PR #2740, opened 2021-06-16
Fix Bouncy Castle fips incompatible issue

jiazhai:

### Motivation

More details are provided in [Pulsar # 10937](https://github.com/apache/pulsar/issues/10937).

In #2631, the default BouncyCastle was changed from non-fips into fips version. But the default version of BouncyCastle in Pulsar is the [non-fips](https://github.com/apache/pulsar/blob/v2.8.0/pulsar-client/pom.xml#L56) one(aimed to make it compatible with the old version of Pulsar). 

Bouncy Castle provides both FIPS and non-FIPS versions, but in a JVM, it can not include both of the 2 versions(non-Fips and Fips), and we have to exclude the current version before including the other. This makes the backward compatible a little hard, and that's why Pulsar has to involve an individual module for [Bouncy Castle](https://pulsar.apache.org/docs/en/security-bouncy-castle).

And if we want to start BookKeeper with TLS enabled through Pulsar's binary, it will meet the following error:
```
Exception in thread "main" java.lang.NoClassDefFoundError: org/bouncycastle/jcajce/provider/BouncyCastleFipsProvider
	at java.base/java.lang.Class.forName0(Native Method)
	at java.base/java.lang.Class.forName(Class.java:315)
	at org.apache.bookkeeper.common.util.ReflectionUtils.forName(ReflectionUtils.java:49)
	at org.apache.bookkeeper.tls.SecurityProviderFactoryFactory.getSecurityProviderFactory(SecurityProviderFactoryFactory.java:39)
	at org.apache.bookkeeper.proto.BookieServer.<init>(BookieServer.java:129)
	at org.apache.bookkeeper.server.service.BookieService.<init>(BookieService.java:52)
	at org.apache.bookkeeper.server.Main.buildBookieServer(Main.java:304)
	at org.apache.bookkeeper.server.Main.doMain(Main.java:226)
	at org.apache.bookkeeper.server.Main.main(Main.java:208)
Caused by: java.lang.ClassNotFoundException: org.bouncycastle.jcajce.provider.BouncyCastleFipsProvider
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:581)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:522)
	... 9 more
```

This fix is to use the reflection to get the loaded bc version to avoid the hard-coded bc version.

### Changes

Use the reflection to get the loaded bc version to avoid the hard-coded bc version
Add backward compatible test for bc-non-fips version

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/tls/TLSContextFactory.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/tls/TLSContextFactory.java
@@ -53,8 +53,6 @@ import org.apache.bookkeeper.conf.AbstractConfiguration;
 import org.apache.bookkeeper.conf.ClientConfiguration;
 import org.apache.bookkeeper.conf.ServerConfiguration;
 import org.apache.commons.io.FileUtils;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 /**
  * A factory to manage TLS contexts.
@@ -144,7 +142,6 @@ public class TLSContextFactory implements SecurityHandlerFactory {
         }
     }
 
-    private static final Logger LOG = LoggerFactory.getLogger(TLSContextFactory.class);
     private static final String TLSCONTEXT_HANDLER_NAME = "tls";
     private String[] protocols;
     private String[] ciphers;
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/tls/TLSContextFactory.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/tls/TLSContextFactory.java
@@ -37,6 +37,8 @@ import java.security.KeyStore;
 import java.security.KeyStoreException;
 import java.security.NoSuchAlgorithmException;
 import java.security.NoSuchProviderException;
+import java.security.Provider;
+import java.security.Security;
 import java.security.UnrecoverableKeyException;
 import java.security.cert.CertificateException;
 import java.security.spec.InvalidKeySpecException;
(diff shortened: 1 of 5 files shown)


jiazhai:
@zymap  @codelipenghui FYI.

eolivelli:
@Ghatage PTAL

eolivelli:
@jiazhai TLS tests are failing, PTAL

sijie:
@jiazhai overall looks good to me. Can you check why TLS tests are failing?

bookkeeper-server/src/main/java/org/apache/bookkeeper/tls/TLSContextFactory.java
```diff
@@ -58,9 +60,66 @@
...
+     *  1. try get from classpath.
+     *  2. try get from Nar.
+     */
+    public static Provider getProvider() {
+        boolean isProviderInstalled =
+            Security.getProvider(BC) != null || Security.getProvider(BC_FIPS) != null;
+
+        if (isProviderInstalled) {
+            Provider provider = Security.getProvider(BC) != null
+                ? Security.getProvider(BC)
```
Ghatage, line 83: This will never be true because since we don't add the BC provider in code, and expect it to be loaded via classpath, it should be present in the classpath.

Currently, we add the dependencies of the project via a set_module to the classpath at runtime.
[Right now only fips based bouncy castle is present in the dependencies](https://github.com/apache/bookkeeper/blob/90aa086f6870ded621ff9d8a69fc79b56a6f8493/pom.xml#L304). 

Furthermore I see [this flag to be set by default in the bookeeper start script](https://github.com/apache/bookkeeper/blame/90aa086f6870ded621ff9d8a69fc79b56a6f8493/bin/bookkeeper#L149) which is going to run BK in fips mode.

Ghatage (changes requested):
Thanks for doing this @jiazhai, I understand this is a WIP, comments added below.

Thanks for the tag @eolivelli!

bookkeeper-server/src/main/java/org/apache/bookkeeper/tls/TLSContextFactory.java
```diff
@@ -58,9 +60,66 @@
...
+
+    // Security.getProvider("BC") / Security.getProvider("BCFIPS").
+    // also used to get Factories. e.g. CertificateFactory.getInstance("X.509", "BCFIPS")
+    public static final String BC_FIPS = "BCFIPS";
+    public static final String BC = "BC";
+
+    /**
+     * Get Bouncy Castle provider, and call Security.addProvider(provider) if success.
+     *  1. try get from classpath.
+     *  2. try get from Nar.
```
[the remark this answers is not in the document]
jiazhai: This was a typo. Thanks for point it out. we should not use nar because the signature issue.

bookkeeper-server/src/main/java/org/apache/bookkeeper/tls/TLSContextFactory.java
```diff
@@ -58,9 +60,66 @@
...
+    }
+
+    /**
+     * Get Bouncy Castle provider from classpath, and call Security.addProvider.
+     * Throw Exception if failed.
+     */
+    public static Provider getBCProviderFromClassPath() throws Exception {
+        Class clazz;
+        try {
+            // prefer non FIPS, for backward compatibility concern.
```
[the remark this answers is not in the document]
jiazhai: Thanks, @Ghatage for the review.  Since BK already use FIPS version, will change this part of logic to use FIPS firstly. Regarding the BC issue, you could get more info from the original issue [Pulsar # 10937](https://github.com/apache/pulsar/issues/10937)

Ghatage:
Thanks for the context @jiazhai. That helps a lot. And also thanks for doing this!. 
Wouldn't an easy fix be to allow imports of BK's bouncy castle which were disallowed in @eolivelli's [changes here](https://github.com/eolivelli/pulsar/commit/2de612791f0334fe469e8e0a58e291358af77cdb)?
I guess I may be over looking the reasons why they were made in the first place.
Your changes look good otherwise 👍 
Will you be using maven or gradle for your builds?

Side note:
@pkumar-singh Could you please help us understand how `build.gradle` is pulling in the dependencies here?
[In these lines](https://github.com/apache/bookkeeper/blob/a445728a2ab1f6b6805c6904126fb039d71d8faf/dependencies.gradle#L27) I see bouncycastle 1.56 and bc-fips 1.0.2, both deps being pulled in. AFAIK, it's recommended to have only one or to avoid classpath conflicts and runtime errors based on which one gets loaded.
I don't see the same in the base level maven `pom.xml`.

jiazhai:
Thanks for the comments @Ghatage . 
From my view, @eolivelli's [changes](https://github.com/eolivelli/pulsar/commit/2de612791f0334fe469e8e0a58e291358af77cdb) are reasonable, because the bc versions(fips/non-fips) are in conflict, we could only include one of them into the package. Usually it uses include/exclude to solve the issue.
It would help a lot if we could find a way to support both of the versions. 

currently, the changes and tests are included in this PR. And by this change, we could have a choice in Pulsar.

> Thanks for the context @jiazhai. That helps a lot. And also thanks for doing this!.
> Wouldn't an easy fix be to allow imports of BK's bouncy castle which were disallowed in @eolivelli's [changes here](https://github.com/eolivelli/pulsar/commit/2de612791f0334fe469e8e0a58e291358af77cdb)?
> I guess I may be over looking the reasons why they were made in the first place.
> Your changes look good otherwise 👍
> Will you be using maven or gradle for your builds?
> 
> Side note:
> @pkumar-singh Could you please help us understand how `build.gradle` is pulling in the dependencies here?
> [In these lines](https://github.com/apache/bookkeeper/blob/a445728a2ab1f6b6805c6904126fb039d71d8faf/dependencies.gradle#L27) I see bouncycastle 1.56 and bc-fips 1.0.2, both deps being pulled in. AFAIK, it's recommended to have only one or to avoid classpath conflicts and runtime errors based on which one gets loaded.
> I don't see the same in the base level maven `pom.xml`.

Ghatage (approved):
LGTM +1 @jiazhai 
PTAL at the tests, other than that the changes are good and thanks for doing this!

bookkeeper-server/src/main/java/org/apache/bookkeeper/tls/TLSContextFactory.java
jiazhai: already changed the logic

jiazhai:
@zymap Would you please also help take a look?

bookkeeper-server/src/main/java/org/apache/bookkeeper/tls/TLSContextFactory.java
```diff
@@ -130,7 +191,7 @@ private KeyManagerFactory initKeyManagerFactory(String keyStoreType, String keyS
         KeyManagerFactory kmf = null;
 
         if (Strings.isNullOrEmpty(keyStoreLocation)) {
-            LOG.error("Key store location cannot be empty when Mutual Authentication is enabled!");
+            log.error("Key store location cannot be empty when Mutual Authentication is enabled!");
```
zymap, line 194: If we use the slf4j, looks like we can remove line 147?

zymap (commented):
LGTM

jiazhai:
@sijie to take a look

jiazhai:
cherry-picked into branch-4.14

eolivelli:
see 69360e1169a57d23bdc8fbfd76e5b300dac4d9cb

diegosalvi:
+1 to have this released ^_^ I failed to cannot upgrade to 4.14.x due to non fips libs already in my classpath and widely used

jiazhai:
@diegosalvi Would you please also help verify this change in your environment?

diegosalvi:
@jiazhai I can check with this commit next week :)

eolivelli:
@diegosalvi we are going to cut a release soon

@zymap already started a discussion on dev@bookkeeper.apache.org

eolivelli:
Please take time to vote on the release candidate when it will be prepared

PR #2743, opened 2021-06-18
Print usage in autorecovery command if needed

fantapsody:
### Motivation

I was trying to read the code to see how auto-recovery service works, just noticed the method `printUsage` was never called and the program doesn't print the usage even when '-h' is specified.

### Changes

Print the usage of auto-recovery if '-h' is specified or additional unexpected arguments are provided when running the auto-recovery command.


eolivelli (approved):
Thanks

Lgtm

sijie (approved):

PR #2745, opened 2021-06-24
[Server] Refactor some code to improve performance.

mattisonchao:
### Motivation

Refactor some code to improve performance.

### Changes

- Break the loop when the conditions are met.
- Refactor string connection to SpringBuilder append.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/tls/BookieAuthZFactory.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/tls/BookieAuthZFactory.java
@@ -95,6 +95,7 @@ public class BookieAuthZFactory implements BookieAuthProvider.Factory {
                         for (String allowedRole : allowedRoles) {
                             if (certRole[0].equals(allowedRole)) {
                                 authorized = true;
+                                break;
                             }
                         }
                         if (authorized) {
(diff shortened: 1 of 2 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/tls/BookieAuthZFactory.java
```diff
@@ -95,6 +95,7 @@ public void onProtocolUpgrade() {
                         for (String allowedRole : allowedRoles) {
                             if (certRole[0].equals(allowedRole)) {
                                 authorized = true;
+                                break;
                             }
                         }
                         if (authorized) {
```
dlg99, line 101: why do you the `break` here?
it bypasses the rest of the code that is executed depending on the value of `authorized`

dlg99 (changes requested):
Getting that `break` in the code does not make sense to me.
The rest of the changes are ok, though I don't think it is executed frequently enough to make any difference performance-wise.

bookkeeper-server/src/main/java/org/apache/bookkeeper/tls/BookieAuthZFactory.java
mattisonchao: When we first judge that `authorized` is true, we should break out of the loop to avoid the next useless loop. :)

dlg99: gosh, mentally misaligned closing `}`, makes sense.

dlg99 (approved):

Ghatage (approved):
LGTM +1

PR #2754, opened 2021-07-15
Remove the `request` in the stream python client dependency

zymap:
---

Master Issue: #2752

*Motivation*

As discussed at length in https://issues.apache.org/jira/browse/LEGAL-572
we found out that the chardet library used by requests library was a
mandatory dependency to requests and since it has LGPL licence, we
should not release any Apache Software with it.


zymap:
Hi @eolivelli, Looks like the `required` lib is not necessary for us. We can remove it by this PR

eolivelli (approved):
LGTM

zymap:
@eolivelli Should we include this PR into the 4.14.2 release?

PR #2757, opened 2021-07-28
Change log level from error to warn in getReadLacResponse

frankxieke:
### Motivation

Two errors do not have an impact on functionality, so change log level from error to warn

### Changes

change log level from error to warn in getReadLacResponse

do not need documentation.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadLacProcessorV3.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadLacProcessorV3.java
@@ -74,7 +74,7 @@ class ReadLacProcessorV3 extends PacketProcessorBaseV3 implements Runnable {
             }
         } catch (Bookie.NoLedgerException e) {
             status = StatusCode.ENOLEDGER;
-            logger.error("No ledger found while performing readLac from ledger: {}", ledgerId, e);
+            logger.warn("No ledger found while performing readLac from ledger: {}", ledgerId, e);
         } catch (IOException e) {
             status = StatusCode.EIO;
             logger.error("IOException while performing readLac from ledger: {}", ledgerId);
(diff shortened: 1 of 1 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadLacProcessorV3.java
```diff
@@ -74,10 +74,10 @@ private ReadLacResponse getReadLacResponse() {
             }
         } catch (Bookie.NoLedgerException e) {
             status = StatusCode.ENOLEDGER;
-            logger.error("No ledger found while performing readLac from ledger: {}", ledgerId, e);
+            logger.warn("No ledger found while performing readLac from ledger: {}", ledgerId, e);
         } catch (IOException e) {
             status = StatusCode.EIO;
-            logger.error("IOException while performing readLac from ledger: {}", ledgerId);
+            logger.warn("IOException while performing readLac from ledger: {}", ledgerId);
```
eolivelli, line 80: an IO error is a signal that something is going wrong.
I would keep it at "error" level

eolivelli (commented):
thanks for sharing your patch.

I left one comment PTAL

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadLacProcessorV3.java
frankxieke: yes, it should be error here,  i change it back to error.

There is another IOException error log, i think it maybe "warn" level, help to review. thanks

frankxieke: the message is like this
`1)      03:32:11.130 [BookieReadThreadPool-OrderedExecutor-0-0] ERROR org.apache.bookkeeper.proto.ReadLacProcessorV3 - IOException while trying to read last entry: 2
org.apache.bookkeeper.bookie.Bookie$NoEntryException: Entry -1 not found in 2
`

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadLacProcessorV3.java
```diff
@@ -89,10 +89,10 @@ private ReadLacResponse getReadLacResponse() {
             }
         } catch (Bookie.NoLedgerException e) {
             status = StatusCode.ENOLEDGER;
-            logger.error("No ledger found while trying to read last entry: {}", ledgerId, e);
+            logger.warn("No ledger found while trying to read last entry: {}", ledgerId, e);
         } catch (IOException e) {
             status = StatusCode.EIO;
-            logger.error("IOException while trying to read last entry: {}", ledgerId, e);
+            logger.warn("IOException while trying to read last entry: {}", ledgerId, e);
```
codelipenghui, line 95: We should only change the logger level to WARN for `NoEntryException` not all  IOException. The purpose of this PR should be if users provided a wrong ledger ID or an entry ID, the log level should be WARN.

codelipenghui (changes requested):

codelipenghui (approved):

eolivelli (approved):
LGTM

zymap (approved):

PR #2767, opened 2021-08-19
skip replicating ledger when bookkeeper server is online

### LINKED ISSUE #2766: Ledgers remaining underreplicated even when bookies come back from the dead
### Describe
When 2 or more bookies in the cluster are down, the auto-recovery gets triggered, but as these come back online, the rereplication worker should Ideally skip the rereplication of the ledgers that are marked as underreplicated. But the ledgers are rereplicated instead

### To Reproduce

Create a cluster with enough bookies for this test (>5)
Start publishing the messages to create a sizeable batch of ledgers
Before killing the pods check the number of undereplicated ledges
Kill 2 or more bookie pods to trigger auto-recovery process
Check the list of underreplicated ledgers now
Keep monitoring the undereplicated ledgers while the bookie come back online

### Expected behavior
Ideally as and when the bookies come up the ledgers marked as underreplicated would be read by the rereplicaton worker and from the metadata the worked should skip these as the bookies are available.

gaozhangmin:
Fixes #2766 

### Motivation

When 2 or more bookies in the cluster are down, the auto-recovery gets triggered, but as these come back online, the rereplication worker should Ideally skip the rereplication of the ledgers that are marked as underreplicated. But the ledgers are rereplicated instead
### Changes
Add `skipReplcateLedgersOnStartup` config to control whether to skip replicate ledger on startup.
skip replicated ledger when bookkeeper server is online.


gaozhangmin:
@eolivelli  
Maybe the idea remove ledger from underreplicated  when  bookie starts up has some problem, What important is, replicate legder puts so much heavy loads on cluster. I want to reduce the affect.


suppose there are 5 copies and 4 bookies are lost, when one lost bookie is online again,  I don't simply mark the ledger replicated.  When the replicateWork gets the failed bookies, it will  ignore the one which online again.  So, this will reduce replicating one copy of data.

PR #2769, opened 2021-08-21
Support specifying bookie http port as a command argument

fantapsody:
### Motivation

I was trying to start multiple bookies locally and found it's a bit inconvenient to specify different http ports for different bookies.

### Changes

Add a command-line argument `httpport` to the bookie command to support specifying bookie http port from the command line.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/server/Main.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/server/Main.java
@@ -76,7 +76,7 @@ public class Main {
         BK_OPTS.addOption("z", "zkserver", true, "Zookeeper Server");
         BK_OPTS.addOption("m", "zkledgerpath", true, "Zookeeper ledgers root path");
         BK_OPTS.addOption("p", "bookieport", true, "bookie port exported");
-        BK_OPTS.addOption("httpport", true, "bookie http port exported");
+        BK_OPTS.addOption("hp", "httpport", true, "bookie http port exported");
         BK_OPTS.addOption("j", "journal", true, "bookie journal directory");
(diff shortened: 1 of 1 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/server/Main.java
```diff
@@ -76,6 +76,7 @@
         BK_OPTS.addOption("z", "zkserver", true, "Zookeeper Server");
         BK_OPTS.addOption("m", "zkledgerpath", true, "Zookeeper ledgers root path");
         BK_OPTS.addOption("p", "bookieport", true, "bookie port exported");
+        BK_OPTS.addOption("httpport", true, "bookie http port exported");
```
nicoloboschi, line 79: "hp" as short-hand ?

nicoloboschi (changes requested):
LGTM, I'd like to add short-hand option name as the other options

bookkeeper-server/src/main/java/org/apache/bookkeeper/server/Main.java
eolivelli: makes sense

fantapsody: Thanks for the suggestion, just added it.

eolivelli:
@nicoloboschi PTAL

eolivelli (approved):
LGTM

nicoloboschi (approved):
+1

addisonj:
@sijie can you merge?

hangc0276 (approved):
LGTM

zymap (approved):

PR #2775, opened 2021-08-24
[STREAMSTORAGE] support an optional time to live (TTL) on a per table basis

mauricebarnum:
=== MOTIVIATION

Bookkeeper's state store is often used, especially through Pulsar
functions, to store processing state in a streaming application which
require persistent short term storage.  By allowing clients to set a
TTL, this data will get cleaned up by the system automatically.

=== CHANGES

Specify the TTL via stream.StreamConfiguration.ttlSeconds, default is an
infinite TTL.

Note that changing the TTL on existing stream will have no effect until
the stream is next loaded by a storage container.

Update RocksDB to 6.22.1.1 to address failures when using TtlDb.

### Motivation

### Changes

> ---
>
> ---
> quickly and easily:
>
>
>
> Otherwise:
> 
> 
> ---


stream/storage/impl/src/main/java/org/apache/bookkeeper/stream/storage/impl/service/RangeStoreServiceImpl.java
```diff
@@ -128,10 +140,31 @@
...
+        this.streamProperties = new ConcurrentHashMap<Long, StreamProperties>();
+    }
+
+    private CompletableFuture<StreamProperties> getStreamProperties(long streamId) {
+        final StreamProperties p = streamProperties.get(streamId);
+        if (p != null) {
+            return FutureUtils.value(p);
+        }
+
+        return clientManager.getStreamProperties(streamId)
```
pkumar-singh, line 157: Probably you can do this.
```
GetStreamRequest streamRequest = GetStreamRequest.newBuilder().setStreamId(streamId).build();
getStream(streamRequest)
```
Or you can expose `RootRangeStoreImpl:getStreamProps(long streamId)`
The one you are using is.
Making a GRPC call, which will eventually land here and execute same code.
Moreover, if you choose to use clientManager, you do not have to cache stream properties, because clientManager already does that, though incorrectly. I think ConcurrentHashmap does not protect you against race-condition unless you use putIfAbsent() API.

mauricebarnum: RootRange reads from local storage. The getStreamProps service call fails in the base impl. I tried that first. :)

I'll delete the cache. CHM would protect against corrupting the internals bit potentially overwrite one reference with another equivalent one.  Since I didn't avoid concurrent RPC due to the unlikelihood once we get to the put it's hard to care.

pkumar-singh:
LGTM.

stream/statelib/src/main/java/org/apache/bookkeeper/statelib/impl/kv/RocksdbKVStore.java
```diff
@@ -367,11 +376,35 @@ protected void openRocksdb(StateStoreSpec spec) throws StateStoreException {
             }
 
             List<ColumnFamilyHandle> cfHandles = Lists.newArrayListWithExpectedSize(2);
-            RocksDB db = RocksDB.open(
-                options,
-                dbDir.getAbsolutePath(),
-                Lists.newArrayList(metaDesc, dataDesc),
-                cfHandles);
+
+            RocksDB db = TtlDB.open(
```
eolivelli, line 380: if ttlSeconds is not set, can we simply open the Database in the old way ?

every existing BK cluster does not have this option
IIUC with this change we are going to open/close/open the DB
we can go with `RocksDB.open` without even trying to do `TtlDB.open`

eolivelli (changes requested):
Overall LGTM

I left one comment PTAL

stream/statelib/src/main/java/org/apache/bookkeeper/statelib/impl/kv/RocksdbKVStore.java
mauricebarnum: Makes sense. I was hoping to eventually remove the conditional logic, but with no practical way to upgrade existing DBs…

I’ll simplify the logic and change the error message to indicate what is wrong.  This will also enable a BK rollback if the option hasn’t been used yet to create any new tables with a ttl.

I need to think about a way to fail opening of a ttl db with the old code. Using different column family names should suffice.

mauricebarnum: Updated to check that an existing DB is being opened with a TTL settings compatible with a previous open.  This provides a better error indication upon a mismatch.  This also avoids re-playing a writeahead log against the wrong implementation; this isn't a current concern but a subtle bug if someone changes the code to use a local WAL.

hsaputra: FYI @eolivelli

eolivelli (approved):
Great work

hsaputra:
rerun failure checks

hsaputra (approved):
LGTM +1

hsaputra:
Will merge this EOD if no more comments or concerns

Issue #2784, opened 2021-09-03
Can not get real ip when use RackawareEnsemblePlacementPolicy


casuallc:
**BUG REPORT**

***Describe the bug***

- set ensemblePlacementPolicy=org.apache.bookkeeper.client.RackawareEnsemblePlacementPolicy
- set reppDnsResolverClass=myclass(implements DNSToSwitchMapping)

In some server, I get 127.0.0.1 when the method DNSToSwitchMapping.resolve called.
But the real ip (129.x.x.x) is expected.

***Screenshots***

![image](https://user-images.githubusercontent.com/9473606/131961337-404f8fb3-fadf-4fdf-9a45-372e6ba05459.png)

***Additional context***

PR #2785, opened 2021-09-03
[ISSUE 2784] get local host from config file

casuallc:

#2784


dlg99 (changes requested):
Please add unit tests and add new config into the default conf file with description to make the feature discoverable.

I think this might be useful for hosts with multiple network interfaces

casuallc:
> Please add unit tests and add new config into the default conf file with description to make the feature discoverable.
> 
> I think this might be useful for hosts with multiple network interfaces

@dlg99  I do not think there need a new config, we just use the existed config `advertisedAddress`. 

If we use `InetAddress.getLocalHost()` instead of use config `advertisedAddress`, there is problem.

I add more log, take a look please. #2784

StevenLuMT:

PR #2788, opened 2021-09-08
fix-npe-when-pulsar-ZkBookieRackAffinityMapping-getBookieAddressResolver

gaozhangmin:
`BookieAddressResolver` should be set before  `((Configurable) dnsResolver).setConf(conf);`  

It will throw npe. when pulsar `ZkBookieRackAffinityMapping` invoke getBookieAddressResolver`


gaozhangmin:
@hangc0276  PTAL

eolivelli (approved):
+1

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
```diff
@@ -256,10 +256,14 @@ public RackawareEnsemblePlacementPolicyImpl initialize(ClientConfiguration conf,
         DNSToSwitchMapping dnsResolver;
         if (optionalDnsResolver.isPresent()) {
             dnsResolver = optionalDnsResolver.get();
+            dnsResolver.setBookieAddressResolver(bookieAddressResolver);
         } else {
             String dnsResolverName = conf.getString(REPP_DNS_RESOLVER_CLASS, ScriptBasedMapping.class.getName());
             try {
                 dnsResolver = ReflectionUtils.newInstance(dnsResolverName, DNSToSwitchMapping.class);
+                if (dnsResolver != null) {
```
lhotari, line 264: is there some way how dnsResolver could be null here?

eolivelli: good catch @lhotari  !
it is always non null here

nicoloboschi (approved):
LGTM

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
gaozhangmin: remove null checker

gaozhangmin:
pulsar also has a related issue  https://github.com/apache/pulsar/pull/11947
@eolivelli  @lhotari  @nicoloboschi  PTAL

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
```diff
@@ -282,9 +284,6 @@ public RackawareEnsemblePlacementPolicyImpl initialize(ClientConfiguration conf,
                 }
             }
         }
-        if (dnsResolver != null) {
-            dnsResolver.setBookieAddressResolver(bookieAddressResolver);
```
lhotari, line 286: This logic is missing when dnsResolver gets set on line 276

lhotari (changes requested):

lhotari (approved):
LGTM

lhotari:
@gaozhangmin Can you add the stacktrace containing the NPE to the description? That could help find this PR if someone searches by stacktrace elements?

gaozhangmin:
> @gaozhangmin Can you add the stacktrace containing the NPE to the description? That could help find this PR if someone searches by stacktrace elements.

there no stacktrace in log ,  only java.lang.nullpointexception

lhotari:
> there no stacktrace in log , only java.lang.nullpointexception

The log line with that could also help locate this issue when someone with the same problem searches for it.  Please share.

gaozhangmin:
> > there no stacktrace in log , only java.lang.nullpointexception
> 
> The log line with that could also help locate this issue when someone with the same problem searches for it. Please share.

added

lhotari:
Thanks for providing the log message @gaozhangmin . It looks like the logging is broken since the provided sample log message didn't include the stacktrace.  I'd propose to fix it in the same PR since it's closely related to the change that is being made.

gaozhangmin:
> Thanks for providing the log message @gaozhangmin . It looks like the logging is broken since the provided sample log message didn't include the stacktrace. I'd propose to fix it in the same PR since it's closely related to the change that is being made.
> 
> This looks invalid:
> `LOG.error("Failed to initialize DNS Resolver {}, used default subnet resolver : {} {}", dnsResolverName, re, re.getMessage());`
> 
> I wonder why it's not simply
> `LOG.error("Failed to initialize DNS Resolver {}, using default subnet resolver.", dnsResolverName, re);` ?

changed

lhotari (approved):
LGTM

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
```diff
@@ -256,10 +256,12 @@ public RackawareEnsemblePlacementPolicyImpl initialize(ClientConfiguration conf,
         DNSToSwitchMapping dnsResolver;
         if (optionalDnsResolver.isPresent()) {
             dnsResolver = optionalDnsResolver.get();
+            dnsResolver.setBookieAddressResolver(bookieAddressResolver);
```
dlg99, line 259: I'd assume that provided `optionalDnsResolver` should be ready to use / have the `bookieAddressResolver` set, as it happens there:
https://github.com/apache/bookkeeper/blob/97818f5123999396e66f5246420d3c7e3d25f53d/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeper.java#L500-L507

Have you traced why we happen to have the optional dns resolver provided here yet not properly configured?
I think the right place to fix the issue is where the dns resolver is created and passed to the dnsResolver.
Shall we move bookieAddressResolver into  DNSToSwitchMapping constructor (or Builder?) and make it mandatory (preconditions check)?

dlg99 (changes requested):
Can I ask you to:

- configure dnsResolver where it is created/before passing it to downstream components instead of setting config to the instance passed here (this can have side-effects, I don't remember if we have to worry about thread safety in that setter etc.) See my comment on the code.
Overall I'd assume that component should get a usable parameter value instead of a half-baked one, so patching the configuration there smells like a hack instead of a fix.
In case that's not possible for some reason I'd love to understand why (to be fair, i didn't look at the code that deep to remember all nuances).

- add unit-tests

-  check other policies (region-aware/zone-aware) to make sure these problems don't exist there.

i.e. in `ZoneawareEnsemblePlacementPolicyImpl` I see
```
        DNSToSwitchMapping actualDNSResolver;
        if (optionalDnsResolver.isPresent()) {
            actualDNSResolver = optionalDnsResolver.get();
        } else {
```
I assume that we'll be fine if we pass properly configured resolver there. 
But if there is an issue that prevents us from doing so, both policies will have similar problems.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
gaozhangmin: Agree

gaozhangmin: Because here it missed condition that dnsResolver is null, so bookieAddressResolver would not be set

dlg99: @gaozhangmin `RackawareEnsemblePlacementPolicyImpl` gets Optional (stack from `initializeEnsemblePlacementPolicy` down to the policy will do `Optional.ofNullable(dnsResolver)` to pass it.) 
`Optional.isPresent()` in `RackawareEnsemblePlacementPolicyImpl` would be false in case of null (which is handled).

Where is the dnsResolver created and set, how did it end up partially configured?

gaozhangmin: if dnsResolver is null, `RackawareEnsemblePlacementPolicyImpl.java ` will create a Configurable resolver. here, bookieAddressResolver should be set. or ,it will throw NPE

I checked the code,  `ZoneawareEnsemblePlacementPolicyImpl` also has the same problem, i will fix it.

gaozhangmin: Configure dnsResolver where it is created/before passing it to downstream components.
This may be a better way fixing the issue.  But I am not sure it will be accepted, cause too many code changes,

eolivelli:
Canbyou add a test cases that shows the problem?
This way it will be clear what we are fixing and that you fixed it.
I would prefer to refactor interfaces or pluggable components in order to not introduce API changes, in that case it would be harder to upgrade BK

gaozhangmin:
> I would prefer to refactor interfaces or pluggable components in order to not introduce API changes, in that case it would be harder to upgrade BK

This need pulsar `org.apache.pulsar.zookeeper.ZkBookieRackAffinityMapping` class, I do not know how to test this.

eolivelli:
Sorry I meant:

**I would NOT prefer** to refactor interfaces or pluggable components in order to not introduce API changes, in that case it would be harder to upgrade BK

So this change in this current form works for me.
hat said, adding a simple test case that reproduces the NPE will be better, otherwise we could break ZkBookieRackAffinityMapping again

gaozhangmin:
> Sorry I meant:
> 
> **I would NOT prefer** to refactor interfaces or pluggable components in order to not introduce API changes, in that case it would be harder to upgrade BK
> 
> So this change in this current form works for me.
> That said, adding a simple test case that reproduces the NPE will be better, otherwise we could break ZkBookieRackAffinityMapping again
> 
> If you have the error on ZkBookieRackAffinityMapping you could try to use the bookieAddressResolver in the way that ZkBookieRackAffinityMapping does, but in a dummy implementation

Unit test added, PTAL

eolivelli (commented):
I believe we are on our way.

Is it possible to add some test cases, or enhance some existing test cases, in order to cover the changes you made ?

gaozhangmin:
> I believe we are on our way.
> 
> Is it possible to add some test cases, or enhance some existing test cases, in order to cover the changes you made ?

I think the unit test `TestRackawareEnsemblePlacementPolicy` already covered this change.

like `repp.initialize(conf, Optional.<DNSToSwitchMapping>empty(), timer,
                DISABLE_ALL, NullStatsLogger.INSTANCE, BookieSocketAddress.LEGACY_BOOKIEID_RESOLVER);`

the dnsResolver is null, will reach the condition `!optionalDnsResolver.isPresent() `

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
dlg99: @gaozhangmin let me reiterate my point.
dns resolver is either null or not in this constructor https://github.com/apache/bookkeeper/blob/c7236adc3cb659e65ae5ce53b7156569d7f50ebd/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeper.java#L410 and it is used to initialize placement policy there. In case it is null the policy will create and initialize (correctly, with address resolver) a new one. If it is not null, the constructor will set the resolver.

The change below, where the policy creates the new dnsResolver, now sets the address resolver is absolutely fine and a fix I am agreeing with.
What I do not understand is why do set the address resolver in case above:
```
        if (optionalDnsResolver.isPresent()) {
            dnsResolver = optionalDnsResolver.get();

  // Why is the next line needed? the policy have to have address resolver set already
  // 
            dnsResolver.setBookieAddressResolver(bookieAddressResolver);
```

The BK client constructor should not be used directly (outside of unit tests) and the builder should be used.
For example:
- here the dns resolver is passed to the builder: https://github.com/apache/bookkeeper/blob/c7236adc3cb659e65ae5ce53b7156569d7f50ebd/stream/distributedlog/core/src/main/java/org/apache/distributedlog/BookKeeperClient.java#L110-L125 and I'd expect it to get configured with the address resolver.
- here the dns resolver is not passed to the builder https://github.com/apache/bookkeeper/blob/9f06a4ce82f61a9dc4b086650a301734332f4246/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java#L270 and I expect placement policy to create and iniialize a new one.

So, either the part is I am confused about is not actually needed or I am missing specific scenario, but in that case it needs to be addressed differently.

gaozhangmin: Remove duplicate setBookieAddressResolver when optionalDnsResolver isPresent

dlg99 (approved):
Looks awesome!
Thank you for going extra mile with the tests.

gaozhangmin:
any problem with this pr?

eolivelli (approved):
Lgtm

eolivelli:
We will merge as soon as CI passes.

gaozhangmin:
The failed tests seems coming from timeouts. it's not related with this change.

gaozhangmin:
rerun failure tests

gaozhangmin:
@eolivelli  Please approve running workflows

gaozhangmin:
@dlg99 Please help approve runn workflows

gaozhangmin:
run failed checks

gaozhangmin:
rerun failure tests

gaozhangmin:
@eolivelli  the failed tests come from timeouts, please rerun failed tests

PR #2790, opened 2021-09-10
only update topology when bookie rack changed

gaozhangmin:
It is unnecessary to update  topology, removing and adding the same bookieNode, if the rack of bookie stay unchanged.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/TopologyAwareEnsemblePlacementPolicy.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/TopologyAwareEnsemblePlacementPolicy.java
@@ -749,10 +749,12 @@ abstract class TopologyAwareEnsemblePlacementPolicy implements
                     if (node != null) {
                         // refresh the rack info if its a known bookie
                         BookieNode newNode = createBookieNode(bookieAddress);
-                        topology.remove(node);
-                        topology.add(newNode);
-                        knownBookies.put(bookieAddress, newNode);
-                        historyBookies.put(bookieAddress, newNode);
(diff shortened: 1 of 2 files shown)


eolivelli (commented):
Can you explain the benefits of this patch?
Do you have some use case or problem?

Adding a test case will help in understanding the problem better

gaozhangmin:
> Can you explain the benefits of this patch?
When using pulsar bookie affinity group, if We update the bookie region info,  this logic will update all the bookie info in the topology, including the unchanged bookie.  we should only update the bookies which rack has been changed.
I will enhanced the existing unit test.

gaozhangmin:
> Can you explain the benefits of this patch?
> Do you have some use case or problem?
> 
> Adding a test case will help in understanding the problem better

When using pulsar bookie affinity group, if We update the bookie region info,  this logic will update all the bookie info in the topology, including the unchanged bookie.  we should only update the bookies which rack has been changed.
I will enhanced the existing unit test.

dlg99 (approved):
LGTM

gaozhangmin:
@eolivelli   Could you please take a look at this pr.

gaozhangmin:
@eolivelli  PTAL again, It has been for a lone time

gaozhangmin:
@zymap  PTAL

dlg99 (changes requested):
Please add tests

gaozhangmin:
@dlg99  Test added. PTAL.

nicoloboschi (approved):
LGTM 
@dlg99 can you review again? 

I'd like to cherry-pick it to 4.14.5

eolivelli (approved):
LGTM

dlg99 (approved):
LGTM

PR #2794, opened 2021-09-15
Memory leak problem when ledger replication

gaozhangmin:
### Motivation

production environment, memory leak always happened, and there were ledger cannot be replicated successfully.

image

This cause by when `openLedgerNoRecovery` with `BKNotEnoughBookiesException`,  the LedgerHandler won't  closed properly, caused memory leak

https://github.com/apache/bookkeeper/blob/c7236adc3cb659e65ae5ce53b7156569d7f50ebd/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/ReplicationWorker.java#L364-L424

### Changes

close LedgerHandler when openComplete with exception

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerOpenOp.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerOpenOp.java
@@ -196,8 +196,28 @@ class LedgerOpenOp {
                     if (rc == BKException.Code.OK) {
                         openComplete(BKException.Code.OK, lh);
                     } else if (rc == BKException.Code.UnauthorizedAccessException) {
+                        try {
+                            if (lh != null) {
+                                lh.close();
+                            }
+                        } catch (InterruptedException e) {
+                            Thread.currentThread().interrupt();
+                            LOG.info("InterruptedException while closing ledger {}", ledgerId, e);
+                        } catch (BKException e) {
+                            LOG.warn("BKException while closing ledger {} ", ledgerId, e);
+                        }
                         openComplete(BKException.Code.UnauthorizedAccessException, null);
                     } else {
+                        try {
+                            if (lh != null) {
+                                lh.close();
+                            }
+                        } catch (InterruptedException e) {
+                            Thread.currentThread().interrupt();
+                            LOG.info("InterruptedException while closing ledger {}", ledgerId, e);
+                        } catch (BKException e) {
+                            LOG.warn("BKException while closing ledger {} ", ledgerId, e);
+                        }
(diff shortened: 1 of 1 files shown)


gaozhangmin:
@eolivelli PTAL

gaozhangmin:
@hangc0276

eolivelli:
@nicoloboschi @RaulGracia @Ghatage  PTAL  if you have time please

@gaozhangmin which memory is leaked ? direct memory ?

gaozhangmin:
> @nicoloboschi @RaulGracia @Ghatage PTAL if you have time please
> 
> @gaozhangmin which memory is leaked ? direct memory ?

heap memory

gaozhangmin:
> @nicoloboschi @RaulGracia @Ghatage PTAL if you have time please
> 
> @gaozhangmin which memory is leaked ? direct memory ?

image

gaozhangmin:
image

gaozhangmin:
I've tested this pr in my test cluster, it can solve the heap memory leak problem.

nicoloboschi (approved):
+1 nice catch

I'd like to point out that the dead objects are kept alive by [AbstractZkLedgerManager](https://github.com/apache/bookkeeper/blob/c7236adc3cb659e65ae5ce53b7156569d7f50ebd/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractZkLedgerManager.java#L86)

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerOpenOp.java
```diff
@@ -196,8 +196,28 @@ public void safeOperationComplete(int rc, Void result) {
                     if (rc == BKException.Code.OK) {
                         openComplete(BKException.Code.OK, lh);
                     } else if (rc == BKException.Code.UnauthorizedAccessException) {
+                        try {
```
dlg99, line 199: i'd refactor this into something like `closeLedger(LedgerHandle lh)`

dlg99 (approved):
LGTM, with one nit

gaozhangmin:
@dlg99  PTAL

dlg99:
@gaozhangmin looks good! Thank you.

RaulGracia (commented):
LGTM, would it make sense to add some specific test to validate the scenario in which memory was getting leaked (i.e., ledger handles are being closed in the situation in which they were left open before)?

gaozhangmin:
> LGTM, would it make sense to add some specific test to validate the scenario in which memory was getting leaked (i.e., ledger handles are being closed in the situation in which they were left open before)?

I tried, But it's hard to verify the `Ledgerhandler`  closed。 there is no 'isClosed' method.

Issue #2795, opened 2021-09-16
Bookkeeper upgrade using Bookie ID may fail due to cookie mismatch


RaulGracia:
**BUG REPORT**

***Describe the bug***

We have tested upgrading Bookkeeper (in Kubernetes) from 4.11 to 4.14. Concretely, in the new version of Bookkeeper, we wanted to make use of the [Bookie ID functionality](https://github.com/apache/bookkeeper/issues/2396) to assign arbitrary identifiers to Bookies, different to the network-related info.

During the upgrade process, we found a cookie mismatch issue as the following one:

```
2021-09-14 18:10:44,499 - ERROR - [main:Main@228] - Failed to build bookie server
org.apache.bookkeeper.bookie.BookieException$InvalidCookieException: Cookie [4
bookieHost: "bookkeeper-bookie-4-10107"
journalDir: "/bk/journal/j0,/bk/journal/j1,/bk/journal/j2,/bk/journal/j3"
ledgerDirs: "4\t/bk/ledgers/l0\t/bk/ledgers/l1\t/bk/ledgers/l2\t/bk/ledgers/l3"
instanceId: "8666dd7f-38ad-4815-8b72-1e181b89a0ab"
] is not matching with [4
bookieHost: "bookkeeper-bookie-4.bookkeeper-bookie-headless.default.svc.cluster.local:3181"
journalDir: "/bk/journal/j0,/bk/journal/j1,/bk/journal/j2,/bk/journal/j3"
ledgerDirs: "4\t/bk/ledgers/l0\t/bk/ledgers/l1\t/bk/ledgers/l2\t/bk/ledgers/l3"
instanceId: "8666dd7f-38ad-4815-8b72-1e181b89a0ab"
]
        at org.apache.bookkeeper.bookie.Cookie.verifyInternal(Cookie.java:136)
        at org.apache.bookkeeper.bookie.Cookie.verify(Cookie.java:147)
        at org.apache.bookkeeper.bookie.Bookie.readAndVerifyCookieFromRegistrationManager(Bookie.java:374)
        at org.apache.bookkeeper.bookie.Bookie.checkEnvironmentWithStorageExpansion(Bookie.java:446)
        at org.apache.bookkeeper.bookie.Bookie.checkEnvironment(Bookie.java:273)
        at org.apache.bookkeeper.bookie.Bookie.<init>(Bookie.java:731)
        at org.apache.bookkeeper.proto.BookieServer.newBookie(BookieServer.java:152)
[...]

PR #2796, opened 2021-09-16
Issue 2795: Bookkeeper upgrade using Bookie ID may fail due to cookie mismatch

RaulGracia:
### Motivation

If a Bookie uses the new Bookie ID scheme explicitly (set via configuration), it should prevail over default network information to identify the Bookie and lookup for cookies in Zookeeper. Otherwise, the Bookie may fetch legacy cookies related to the same hostname but are invalid because the id differs. Essentially, this PR tries to choose between a configuration-based Bookie ID or a default network info ID when looking for cookies in Zookeeper (considering both is problematic in some cases).

### Changes

The change is simple: when collecting the Bookie IDs that will be used to fetch cookies from Zookeeper (`BookieImpl.possibleBookieIds()`), the code now checks if there is a Bookie ID passed by configuration. If so, we just used that one for lookup cookies in Zookeeper. Otherwise, we keep the existing behavior of using network information to fetch cookies from Zookeeper.

Master Issue: #2795

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
@@ -286,26 +286,27 @@ public class BookieImpl extends BookieCriticalThread implements Bookie {
         // we are checking all possibilities here, so we don't need to fail if we can only get
         // loopback address. it will fail anyway when the bookie attempts to listen on loopback address.
         try {
-            // ip address
-            addresses.add(getBookieAddress(
-                new ServerConfiguration(conf)
-                    .setUseHostNameAsBookieID(false)
-                    .setAdvertisedAddress(null)
-                    .setAllowLoopback(true)
-            ).toBookieId());
-            // host name
-            addresses.add(getBookieAddress(
-                new ServerConfiguration(conf)
-                    .setUseHostNameAsBookieID(true)
-                    .setAdvertisedAddress(null)
-                    .setAllowLoopback(true)
-            ).toBookieId());
-            // advertised address
-            if (null != conf.getAdvertisedAddress()) {
-                addresses.add(getBookieAddress(conf).toBookieId());
-            }
-            if (null != conf.getBookieId()) {
+	     if (null != conf.getBookieId()) {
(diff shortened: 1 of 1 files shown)


eolivelli:
This change makes sense to me.

The upgrade story for enabling BookieID is to compute the same BookieID as it is written in the bookie, like you did in Pravega, is this correct ?

We should write this somewhere in the docs (not in the scope of this PR)

In order to finalise the patch we have to add a unit test, probably a unit test about this method that you changed is enough.
A test that shows how to switch to BookieID will be also very good

RaulGracia:
@eolivelli added a couple of tests to check the behavior of using Bookie ID in the configuration.

eolivelli (approved):
LGTM

nicoloboschi (approved):
LGTM

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieInitializationTest.java
```diff
@@ -1614,4 +1614,34 @@ private void testInvalidServiceMetadataURICase(String uri) throws Exception {
...
+        server.shutdown();
+
+        // If BookieID is set, it takes precedence over network info. Because of that, the new Bookie start
+        // should fail with an InvalidCookieException, as now the custom BookieID takes precedence.
+        String customBookieId = "customId";
+        conf.setBookieId(customBookieId);
+        server = new MockBookieServer(conf);
+        server.start();
+        assertEquals(customBookieId, server.getBookieId().toString());
+        server.shutdown();
```
dlg99, line 1645: The test expected exception, comment says to expect exception after the start.
Do you need assert and shutdown after the code that's supposed to throw?

eolivelli: very good point @dlg99 
we should verify the Exception in a `catch` block and call "shutdown" in a `finally` block in order to perform any cleanup that is needed (and do not fall into leaks)

RaulGracia: Thanks for the feedback. Now the test catches the expected exception and shuts down the mock server at the end of the test.

PR #2801, opened 2021-09-21
Issue 2787: TestPerChannelBookieClient.testEpollChannelTcpUserTimeout fails

### LINKED ISSUE #2787: TestPerChannelBookieClient.testEpollChannelTcpUserTimeout fails
**BUG REPORT**
After we committed https://github.com/apache/bookkeeper/commit/c8d4086697f85c9093f4da2c907a13e17c198914 I see this test failing very often on DataStax CI, that run the tests against current BookKeeper master, on JDK8.

***Describe the bug***
This is the error

[ERROR] Failures: 
[ERROR] org.apache.bookkeeper.proto.TestPerChannelBookieClient.testEpollChannelTcpUserTimeout
[ERROR]   Run 1: TestPerChannelBookieClient.testEpollChannelTcpUserTimeout:337 expected:<1236> but was:<1234>
[ERROR]   Run 2: TestPerChannelBookieClient.testEpollChannelTcpUserTimeout:337 expected:<1236> but was:<1234>
[ERROR]   Run 3: TestPerChannelBookieClient.testEpollChannelTcpUserTimeout:337 expected:<1236> but was:<1234>

This is the line
https://github.com/apache/bookkeeper/commit/c8d4086697f85c9093f4da2c907a13e17c198914#diff-271106bdb4f954bd92d6de6e38f2e9dfa542754e611001050350a72bd7a0b9acR338

eolivelli:
- Fixes #2787
- On some Linux systems the value 1234 is rounded to 1236, this patch simply uses 1236 as value in order to prevent failures

### DIFF
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/proto/TestPerChannelBookieClient.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/proto/TestPerChannelBookieClient.java
@@ -314,7 +314,7 @@ public class TestPerChannelBookieClient extends BookKeeperClusterTestCase {
         EventLoopGroup eventLoopGroup = new EpollEventLoopGroup();
         OrderedExecutor executor = getOrderedSafeExecutor();
         ClientConfiguration conf = new ClientConfiguration();
-        int tcpUserTimeout = 1234;
+        int tcpUserTimeout = 1236; // this value may be rounded on some Linux implementations
         BookieId addr = getBookie(0);
 
         // Pass to the PerChannelBookieClient object the client configuration with TCP user timeout.


eolivelli:
@RaulGracia PTAL

bookkeeper-server/src/test/java/org/apache/bookkeeper/proto/TestPerChannelBookieClient.java
```diff
@@ -314,7 +314,7 @@ public void testEpollChannelTcpUserTimeout() throws Exception {
         EventLoopGroup eventLoopGroup = new EpollEventLoopGroup();
         OrderedExecutor executor = getOrderedSafeExecutor();
         ClientConfiguration conf = new ClientConfiguration();
-        int tcpUserTimeout = 1234;
+        int tcpUserTimeout = 1236; // this value may be rounded on some Linux implementations
```
fpj, line 317: Out of curiosity, is it documented anywhere? How did you find that this rounding is going, @eolivelli ?

eolivelli: Some references

https://datatracker.ietf.org/doc/html/rfc5482

I didn't find more

fpj: Sure, that's the RFC, but where did you see that some linux implementations round it?

eolivelli: I have this setup:

> mvn -v
> Maven home: /usr/share/maven
> Java version: 1.8.0_292, vendor: Private Build, runtime: /usr/lib/jvm/java-8-openjdk-amd64/jre
> Default locale: en, platform encoding: UTF-8
> OS name: "linux", version: "4.15.0-64-generic", arch: "amd64", family: "unix"

>  uname -a
> Linux jenkins-pulsar-basic-runner-324 4.15.0-64-generic #73-Ubuntu SMP Thu Sep 12 13:16:13 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

The problem always reproduces on our Jenkins slaves at DataStax, basically every test run starts a new VM, installs Ubuntu and Java and then builds bookkeeper from the master ASF branch.
it is 100% reproducible on those machines.

I had not time to investigate, because I should learn more about the machines.
We should check that kernel version + Java version....

RaulGracia: @eolivelli do you know if with this change the test is passing consistently in your environment?

dlg99: @eolivelli this smells like an endianness problem somewhere rather than a rounding problem. See https://github.com/jnr/jnr-unixsocket/pull/55 for example. Did it start at some BK version (dependencies updates etc) or it was the first run on these VMs?

dlg99: Another suspect is that [TCP_USER_TIMEOUT](https://man7.org/linux/man-pages/man7/tcp.7.html) (tcpUserTimeout feeds into it) is `unsigned int` but netty's interface to jni passes it as it is jint: https://github.com/netty/netty/blob/23405e2000427e9cad104913e7071d8d08e91a3c/transport-native-epoll/src/main/c/netty_epoll_linuxsocket.c#L144-L146

i.e. for multicast netty casts parameter to `u_int`: https://github.com/netty/netty/blob/23405e2000427e9cad104913e7071d8d08e91a3c/transport-native-epoll/src/main/c/netty_epoll_linuxsocket.c#L78-L81

eolivelli: The test was born a few weeks ago.
It passes on GH CI and it fails on Datastax CI.

Let me paste the results of the tests on this branch

eolivelli:
The tests passed on this branch on DataStax CI

```
[INFO] Running org.apache.bookkeeper.proto.TestPerChannelBookieClient
[INFO] Tests run: 5, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 8.099 s - in org.apache.bookkeeper.proto.TestPerChannelBookieClient
```

bookkeeper-server/src/test/java/org/apache/bookkeeper/proto/TestPerChannelBookieClient.java
eolivelli: @dlg99 is it possible that 1234 is corrupted to 1236 due an endianness problem ?
but 1236 is not changed to another value

RaulGracia (approved):
The change makes no difference to the test logic and if it solves the corner case reported, I'm fine with it.

bookkeeper-server/src/test/java/org/apache/bookkeeper/proto/TestPerChannelBookieClient.java
dlg99: TBH I do not know.
I experimented with endianness conversion, neither 1234 nor 1236 seem problematic or radically different, even if I add casting to unsigned int.
OTOH, TCP_USER_TIMEOUT is defined as a 15bit unsigned in the RFC but I don't think 15 vs 16bit matters for either value. 

but extra 2ms timeout making the test pass makes even less sense

dlg99 (approved):
I'd prefer to understand the rootcause but if it unblocks right now it is ok to go.
Please add an issue to investigate/follow up later

bookkeeper-server/src/test/java/org/apache/bookkeeper/proto/TestPerChannelBookieClient.java
eolivelli: The other cure for this problem is to revert the change at all.

@RaulGracia said that in the test environments this feature is a good enhancement and fixes a problem.

So probably we can change the test to set a value that is likely to be used in production.
If Netty/Linux or other parts om the stack do not preserve every value it is not a BK problem.
We can open a ticket on Netty, I am pretty sure that we cannot so anything in BK.

I want to fix this test that is consistently failing and it is blocking some CI pipeline.

PR #2802, opened 2021-09-22
Add auditor get ledger throttle to avoid auto recovery zk session expire

hangc0276:
### Motivation
When we start BookKeeper auditor, it will start checking all ledgers task with given interval(default 7days) compared to last check timestamp,  and the check ledger operation will process all the activeLedgers shown in the follow code.
https://github.com/apache/bookkeeper/blob/c7236adc3cb659e65ae5ce53b7156569d7f50ebd/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractZkLedgerManager.java#L571-L573

 In the logic of process ledger, it will call `readLedgerMetadata` , which will call getData using zkClient and parse ZNode data in callback, which is time consuming.
https://github.com/apache/bookkeeper/blob/c7236adc3cb659e65ae5ce53b7156569d7f50ebd/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractZkLedgerManager.java#L433-L474

When use use NettySocketChannel for zk client, and the activeLedger number grows more than 1500+, it will send all ledgers'getData  request to zkServer at a time (before starting receive data from zookeeper server), and call `serDe.parseConfig` to parse ZNode data in callback.  However, the parse Znode is time consuming, which will block zk client send heartbeat to zk server, and will cause zk session expire.

When zk session expire, the BookKeeper auto recovery process will shutdown.

### Modification
1. Add throttle semaphore for Auditor to open ledgers, which will call getData using Zookeeper client.
2. Use  individual thread to call `processor.process(ledger, mcb);` for all ledgers instead of Zookeeper client callback thread.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
@@ -204,7 +204,10 @@ public class ServerConfiguration extends AbstractConfiguration<ServerConfigurati
     protected static final String UNDERREPLICATED_LEDGER_RECOVERY_GRACE_PERIOD =
             "underreplicatedLedgerRecoveryGracePeriod";
     protected static final String AUDITOR_REPLICAS_CHECK_INTERVAL = "auditorReplicasCheckInterval";
-    protected static final String AUDITOR_GET_LEDGER_SEMAPHORE = "auditorGetLedgerSemaphore";
+    protected static final String AUDITOR_MAX_NUMBER_OF_CONCURRENT_OPEN_LEDGER_OPERATIONS =
+        "auditorMaxNumberOfConcurrentOpenLedgerOperations";
+    protected static final String AUDITOR_ACQUIRE_CONCURRENT_OPEN_LEDGER_OPERATIONS_TIMEOUT_SEC =
+        "auditorAcquireConcurrentOpenLedgerOperationsTimeOutSec";
 
     // Worker Thread parameters.
     protected static final String NUM_ADD_WORKER_THREADS = "numAddWorkerThreads";
@@ -2530,16 +2533,33 @@ public class ServerConfiguration extends AbstractConfiguration<ServerConfigurati
      *
      * @return The semaphore value. By default it is 500.
      */
-    public int getAuditorGetLedgerSemaphore() {
-        return getInt(AUDITOR_GET_LEDGER_SEMAPHORE, 500);
+    public int getAuditorMaxNumberOfConcurrentOpenLedgerOperations() {
+        return getInt(AUDITOR_MAX_NUMBER_OF_CONCURRENT_OPEN_LEDGER_OPERATIONS, 500);
     }
 
     /**
      * Set the semaphore limit value for getting ledger from zookeeper in auto recovery.
      * @param semaphore
      */
-    public void setAuditorGetLedgerSemaphore(int semaphore) {
-        setProperty(AUDITOR_GET_LEDGER_SEMAPHORE, semaphore);
+    public void setAuditorMaxNumberOfConcurrentOpenLedgerOperations(int semaphore) {
+        setProperty(AUDITOR_MAX_NUMBER_OF_CONCURRENT_OPEN_LEDGER_OPERATIONS, semaphore);
+    }
+
+    /**
+     * Get the acquire concurrent open ledger operations timeout.
+     *
+     * @return The timeout values. By default it is 30s
+     */
+    public int getAuditorAcquireConcurrentOpenLedgerOperationsTimeoutSec() {
+        return getInt(AUDITOR_ACQUIRE_CONCURRENT_OPEN_LEDGER_OPERATIONS_TIMEOUT_SEC, 120);
+    }
+
+    /**
+     * Set the acquire concurrent open ledger operations timeout.
+     * @param timeoutMs
+     */
+    public void setAuditorAcquireConcurrentOpenLedgerOperationsTimeoutSec(int timeoutMs) {
+        setProperty(AUDITOR_ACQUIRE_CONCURRENT_OPEN_LEDGER_OPERATIONS_TIMEOUT_SEC, timeoutMs);
     }
 
 
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
@@ -206,8 +206,8 @@ public class ServerConfiguration extends AbstractConfiguration<ServerConfigurati
     protected static final String AUDITOR_REPLICAS_CHECK_INTERVAL = "auditorReplicasCheckInterval";
     protected static final String AUDITOR_MAX_NUMBER_OF_CONCURRENT_OPEN_LEDGER_OPERATIONS =
         "auditorMaxNumberOfConcurrentOpenLedgerOperations";
-    protected static final String AUDITOR_ACQUIRE_CONCURRENT_OPEN_LEDGER_OPERATIONS_TIMEOUT_SEC =
-        "auditorAcquireConcurrentOpenLedgerOperationsTimeOutSec";
+    protected static final String AUDITOR_ACQUIRE_CONCURRENT_OPEN_LEDGER_OPERATIONS_TIMEOUT_MSEC =
+        "auditorAcquireConcurrentOpenLedgerOperationsTimeOutMSec";
 
     // Worker Thread parameters.
     protected static final String NUM_ADD_WORKER_THREADS = "numAddWorkerThreads";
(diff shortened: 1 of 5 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
```diff
@@ -204,6 +204,7 @@
     protected static final String UNDERREPLICATED_LEDGER_RECOVERY_GRACE_PERIOD =
             "underreplicatedLedgerRecoveryGracePeriod";
     protected static final String AUDITOR_REPLICAS_CHECK_INTERVAL = "auditorReplicasCheckInterval";
+    protected static final String AUDITOR_GET_LEDGER_SEMAPHORE = "auditorGetLedgerSemaphore";
```
eolivelli, line 207: what about `auditorMaxNumberOfConcurrentOpenLedgerOperations` ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
```diff
@@ -165,6 +165,7 @@
     private final AtomicInteger numLedgersFoundHavingLessThanWQReplicasOfAnEntry;
     private final long underreplicatedLedgerRecoveryGracePeriod;
     private final int zkOpTimeoutMs;
+    private final Semaphore semaphore;
```
eolivelli, line 168: what about `openLedgerNoRecoverySemaphore` ?

bookkeeper-server/src/test/java/org/apache/bookkeeper/replication/AuditorPeriodicCheckTest.java
```diff
@@ -365,6 +366,36 @@ public void run() {
...
+    @Test
+    public void testGetLedgerFromZookeeperThrottled() throws Exception {
+        for (AuditorElector e : auditorElectors.values()) {
+            e.shutdown();
+        }
+
+        final int numberLedgers = 100;
+        // write ledgers into bookkeeper cluster
+        for (int i = 0; i < numberLedgers; ++i) {
+            LedgerHandle lh = bkc.createLedger(3, 3, DigestType.CRC32, "passwd".getBytes());
```
eolivelli, line 378: nit: you can use try-with-resources here

bookkeeper-server/src/test/java/org/apache/bookkeeper/replication/AuditorPeriodicCheckTest.java
```diff
@@ -365,6 +366,36 @@ public void run() {
...
+            lh.close();
+        }
+
+        // create auditor and call `checkAllLedgers`
+        ServerConfiguration configuration = confByIndex(0);
+        configuration.setAuditorGetLedgerSemaphore(5);
+        try (final Auditor auditor = new Auditor(
+            BookieImpl.getBookieId(configuration).toString(),
+            configuration, NullStatsLogger.INSTANCE)) {
+            auditor.checkAllLedgers();
```
eolivelli, line 391: is there any way to verify that the operation completed without errors or at least that we called asyncOpenLedgerNoRecovery the expected number of times ?
(with PowerMock should be easy)

eolivelli (changes requested):
Overall LGTM
I left some comments.

We talked offline about this problem and I believe this is a good way to fix the problem

bookkeeper-server/src/test/java/org/apache/bookkeeper/replication/AuditorPeriodicCheckTest.java
nicoloboschi: +1 for this

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
```diff
@@ -347,6 +348,7 @@ public Auditor(final String bookieIdentifier,
         this.numLedgersFoundHavingLessThanAQReplicasOfAnEntry = new AtomicInteger(0);
         this.numLedgersHavingLessThanWQReplicasOfAnEntryGuageValue = new AtomicInteger(0);
         this.numLedgersFoundHavingLessThanWQReplicasOfAnEntry = new AtomicInteger(0);
+        this.semaphore = new Semaphore(conf.getAuditorGetLedgerSemaphore());
```
nicoloboschi, line 351: maybe it is better to validate this value (must be greater than zero) and throw an exception with a useful error message (and indicate the invalid config prop name)

nicoloboschi (changes requested):
good work, I left minor comments

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
```diff
@@ -1230,7 +1232,16 @@ void checkAllLedgers() throws BKException, IOException, InterruptedException, Ke
                     return;
                 }
 
+                try {
+                    semaphore.acquire();
```
dlg99, line 1236: I'd consider using tryAcquire() with some reasonable timeout and either just treating the timeout similarly to `InterruptedException` (return) or retry in the loop. Plus add logging on the timeout.
Otherwise one will have to deal with thread dumps to understand why auditor is not progressing when it is stuck on that semaphore in case of some obscure bug in zk client or server.

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
```diff
@@ -1230,7 +1232,16 @@ void checkAllLedgers() throws BKException, IOException, InterruptedException, Ke
...
+                } catch (InterruptedException e) {
+                    LOG.error("Unable to acquire open ledger operation semaphore ", e);
```
dlg99, line 1238: add `Thread.currentThread().interrupt();`

dlg99 (changes requested):
Overall looks good.
I added couple of suggestions + please address comments from other reviewers.

hangc0276:
I have updated the code, Please help review this PR, thanks a lot. @eolivelli @dlg99 @nicoloboschi

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
```diff
@@ -2524,6 +2528,41 @@ public long getAuditorPeriodicReplicasCheckInterval() {
...
+     * @param semaphore
+     */
+    public void setAuditorMaxNumberOfConcurrentOpenLedgerOperations(int semaphore) {
+        setProperty(AUDITOR_MAX_NUMBER_OF_CONCURRENT_OPEN_LEDGER_OPERATIONS, semaphore);
+    }
+
+    /**
+     * Get the acquire concurrent open ledger operations timeout.
+     *
+     * @return The timeout values. By default it is 30s
```
nicoloboschi, line 2551: 30 or 120? i think the value should be in ms rather than seconds

conf/bk_server.conf
```diff
@@ -875,6 +875,13 @@ zkEnableSecurity=false
 # the provided digest type provided at `digestType` and the provided passwd provided at `passwd`.
 # enableDigestTypeAutodetection=true
 
+# Semaphore limit of getting ledger from zookeeper. Used to throttle the zookeeper client request operation
+# sending to Zookeeper server. Default value is 500
+# auditorMaxNumberOfConcurrentOpenLedgerOperations=500
+
+# Wait time out of acquiring semaphore of concurrent open ledger operations. Default value is 120s.
```
nicoloboschi, line 882: timeout?

nicoloboschi (changes requested):
I left comments on the new property for timeout

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
hangc0276: I set the default timeout to 120s.

conf/bk_server.conf
hangc0276: Ok, i update the doc

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
eolivelli: @hangc0276 
`Thread.currentThread().interrupt()` is to be called only in case of `InterruptedException`
we need to separate the two catch clauses

eolivelli (approved):
overall LGTM
I left one minor comment

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
hangc0276: @eolivelli Ok, i just return when tryAcquire failed.

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
```diff
@@ -1230,7 +1266,23 @@ void checkAllLedgers() throws BKException, IOException, InterruptedException, Ke
                     return;
                 }
 
+                try {
+                    if (!openLedgerNoRecoverySemaphore.tryAcquire(openLedgerNoRecoverySemaphoreWaitTimeoutMSec,
```
dlg99, line 1270: I think openLedgerNoRecoverySemaphoreWaitTimeoutMSec can be safely set to zk client timeout, or 2x of that timeout

dlg99 (approved):
LGTM.
Thank you.

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
hangc0276: The default zk client timeout is 10s, if set to 2 * zkTimeout, it will be 20s. 
However, in the replicas check process, it use `REPLICAS_CHECK_TIMEOUT_IN_SECS=120` as maxConcurrentSemaphore tryAcquire timeout.
So i doubt whether put `openLedgerNoRecoverySemaphoreWaitTimeoutMSec` in to `bk_server.conf` or just hard code to `2 * zkTimeout` and the default value set to 120s or not. 
@dlg99  Would you please give me some ideas?

dlg99: @hangc0276 ouch. let's just leave it as it is.

codelipenghui (approved):

codelipenghui:
@sijie @merlimat Please help review this PR

codelipenghui:
Can we create a new minor release with this fix? It's a critical issue(the recovery service continues restart) if users have many ledgers.

eolivelli:
@codelipenghui yes we can.

Can you please ask for it in dev@bookkeeper.apache.org?

We are also going to release 4.15.0

PR #2803, opened 2021-09-23
Use correct line separator instead of \n

yuruguo:
### Motivation
Using `\n` can only pass on Linux and macOS but not on Windows.

### Changes
use `System.lineSeparator()` instead of `\n`

### DIFF
--- a/bookkeeper-http/http-server/src/main/java/org/apache/bookkeeper/http/service/HeartbeatService.java
+++ b/bookkeeper-http/http-server/src/main/java/org/apache/bookkeeper/http/service/HeartbeatService.java
@@ -27,7 +27,7 @@ import org.apache.bookkeeper.http.HttpServer;
  */
 public class HeartbeatService implements HttpEndpointService {
 
-    public static final String HEARTBEAT = "OK\n";
+    public static final String HEARTBEAT = "OK" + System.lineSeparator();
 
     @Override
     public HttpServiceResponse handle(HttpServiceRequest request) {
--- a/bookkeeper-http/http-server/src/main/java/org/apache/bookkeeper/http/service/NullHttpService.java
+++ b/bookkeeper-http/http-server/src/main/java/org/apache/bookkeeper/http/service/NullHttpService.java
@@ -26,7 +26,7 @@ import org.apache.bookkeeper.http.HttpServer;
  * HttpEndpointService that return fixed content.
  */
 public class NullHttpService implements HttpEndpointService {
-    public static final String CONTENT = "NullHttpService\n";
+    public static final String CONTENT = "NullHttpService" + System.lineSeparator();
 
     @Override
     public HttpServiceResponse handle(HttpServiceRequest request) {
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
@@ -563,7 +563,7 @@ public class BookieImpl extends BookieCriticalThread implements Bookie {
             && !conf.getAllowLoopback()) {
             throw new UnknownHostException("Trying to listen on loopback address, "
                     + addr + " but this is forbidden by default "
-                    + "(see ServerConfiguration#getAllowLoopback()).\n"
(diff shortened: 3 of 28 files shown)


yuruguo:
@merlimat @eolivelli @dlg99 @lhotari PTAL

dlg99 (approved):
LGTM but:
- I do not run BK on windows
- AFAIK majority of the contributors don't use windows so it's easy to sneak `\n` later. You might want to try some checkstyle rule or some other code check to prevent these from happening in the future. 

Just out of the curiosity, are you running BK on windows in production or this is strictly for the development?

yuruguo:
rerun failure checks

yuruguo:
return failure checks

yuruguo:
rerun failure checks

yuruguo:
> LGTM but:
> 
> * I do not run BK on windows
> * AFAIK majority of the contributors don't use windows so it's easy to sneak `\n` later. You might want to try some checkstyle rule or some other code check to prevent these from happening in the future.
> 
> Just out of the curiosity, are you running BK on windows in production or this is strictly for the development?

Thanks for you review.
In fact, I don’t deploy it on windows. The reason for making this change is to maintain code compatibility (I found that pulsar can be deployed on windows, and corresponding BK components are also possible, although people rarely use it like this. )

aahmed-se (approved):

bookkeeper-http/http-server/src/main/java/org/apache/bookkeeper/http/service/HeartbeatService.java
```diff
@@ -27,7 +27,7 @@
  */
 public class HeartbeatService implements HttpEndpointService {
 
-    public static final String HEARTBEAT = "OK\n";
+    public static final String HEARTBEAT = "OK" + System.lineSeparator();
```
eolivelli, line 30: This is not good, because here we are in the HTTP API, the result cannot depend on the OS that is running the bookie

bookkeeper-http/http-server/src/main/java/org/apache/bookkeeper/http/service/NullHttpService.java
```diff
@@ -26,7 +26,7 @@
  * HttpEndpointService that return fixed content.
  */
 public class NullHttpService implements HttpEndpointService {
-    public static final String CONTENT = "NullHttpService\n";
+    public static final String CONTENT = "NullHttpService" + System.lineSeparator();
```
eolivelli, line 29: the same here

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
```diff
@@ -563,7 +563,7 @@ public static BookieSocketAddress getBookieAddress(ServerConfiguration conf)
             && !conf.getAllowLoopback()) {
             throw new UnknownHostException("Trying to listen on loopback address, "
                     + addr + " but this is forbidden by default "
-                    + "(see ServerConfiguration#getAllowLoopback()).\n"
+                    + "(see ServerConfiguration#getAllowLoopback())." + System.lineSeparator()
```
eolivelli, line 566: why?
This is an Exception message, probably it is better to remove the '\n' character

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Cookie.java
```diff
@@ -164,17 +164,17 @@ public String toString() {
             builder.setInstanceId(instanceId);
         }
         StringBuilder b = new StringBuilder();
-        b.append(CURRENT_COOKIE_LAYOUT_VERSION).append("\n");
+        b.append(CURRENT_COOKIE_LAYOUT_VERSION).append(System.lineSeparator());
```
eolivelli, line 167: This is a breaking change in the Cookie

eolivelli (changes requested):
I am sorry but I don't understand this patch.
It looks like a blind replace of "\n" with the system line separator.

We should verify each case separately and we cannot introduce breaking changes.

If we really want to go this way we have to discuss more on dev@ and split the patch in smaller pieces

PR #2805, opened 2021-09-23
replication stat num-under-replicated-ledgers changed as with the pro…

gaozhangmin:
Motivation
Now ReplicationStats numUnderReplicatedLedger registers when `publishSuspectedLedgersAsync`, but its value doesn't decrease as with the ledger replicated successfully, We cannot know the progress of replication from the stat.

Changes
registers a notifyUnderReplicationLedgerChanged when auditor starts. numUnderReplicatedLedger value will decrease when the ledger path under replicate deleted.


gaozhangmin:
@eolivelli  PTAL

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/ZkLedgerUnderreplicationManager.java
```diff
@@ -854,6 +855,28 @@ public int getLostBookieRecoveryDelay() throws UnavailableException {
         }
     }
 
+    @Override
+    public void notifyUnderReplicationLedgerChanged(GenericCallback<Void> cb) throws UnavailableException {
+        LOG.debug("notifyUnderReplicationLedgerChanged()");
+        Watcher w = new Watcher() {
+            @Override
+            public void process(WatchedEvent e) {
+                if (e.getPath().contains("urL") && e.getType() == Event.EventType.NodeDeleted) {
```
nicoloboschi, line 864: checking the event type is cheaper than the contains, I suggest to invert the conditions order

I think it is better to use `getLedgerId(String path)` method instead of the String contains

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
```diff
@@ -691,6 +691,14 @@ public void start() {
                 submitShutdownTask();
             }
 
+            try {
+                this.ledgerUnderreplicationManager.notifyUnderReplicationLedgerChanged(new UnderReplicatedLedgersChangedCb());
+            }catch (UnavailableException ue) {
```
nicoloboschi, line 696: missing a space before the 'catch'

nicoloboschi (changes requested):
LGTM, minor comments here

gaozhangmin:
@nicoloboschi  PTAL

nicoloboschi (approved):
+1

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
```diff
@@ -998,6 +1006,13 @@ public void run() {
         }), initialDelay, interval, TimeUnit.SECONDS);
     }
 
+    private class UnderReplicatedLedgersChangedCb implements GenericCallback<Void> {
+        @Override
+        public void operationComplete(int rc, Void result) {
+            numUnderReplicatedLedger.registerSuccessfulValue(-1);
```
dlg99, line 1012: Please double check me, my understanding is:
numUnderReplicatedLedger is an OpStatsLogger. This works for registerSuccessfulValue only with Prometheus, afaict Codahale simply treats it as a duration in millis (similar to the registerSuccessfulEvent but with time init set in millis) and adds into a histogram.
But that's a side problem.

If your intent is to "know the progress of replication from the stat" then you are getting into a situation when the metric can increase and decrease during the reporting interval and you won't see the progress.
I'd suggest to have a gauge with number of currently under replicated ledgers (numUnderreplicatedLedgersElapsedRecoveryGracePeriod is doing this right now) and a counter incremented for successfully replicated ledgers. Note that ledger can get out of under replicated state by either being re-replicated or if a missing bookie returns back into the cluster.

dlg99 (changes requested):
I am not convinced that it solves the problem as you outlined it (see my comment on the review) + the PR needs a unit test.

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
gaozhangmin: Zookeeper Watcher added, the metric will be updated  When under replicated ledger path  deleted.
I

gaozhangmin: I will try to add a new gauge with number of currently under replicated ledgers.

gaozhangmin: I added three new metrics.  `UNDER_REPLICATED_LEDGERS_COUNT UNDER_REPLICATED_LEDGERS_SIZE REPLICATED_LEDGERS_COUNT` 
previous metrics are named `NUM_UNDER_REPLICATED_LEDGERS UNDER_REPLICATED_LEDGERS_TOTAL_SIZE`
I prefer to delete the previous OpStatsLogger metrics. so only `REPLICATED_LEDGERS_COUNT` will be added.

gaozhangmin: @dlg99

dlg99 (approved):

dlg99:
@gaozhangmin 
```
[INFO] There is 1 error reported by Checkstyle 6.19 with buildtools/src/main/resources/bookkeeper/checkstyle.xml ruleset.
Error:  src/main/java/org/apache/bookkeeper/replication/Auditor.java:[744] (sizes) LineLength: Line is longer than 120 characters (found 126).
```

fyi: https://bookkeeper.apache.org/community/coding_guide/ and https://bookkeeper.apache.org/community/contributing/#optional-ide-setup for checkstyle configuration

gaozhangmin:
@eolivelli  PTAL again, It has been for a lone time

gaozhangmin:
@zymap  PTAL

gaozhangmin:
rerun failure checks

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/ZkLedgerUnderreplicationManager.java
```diff
@@ -866,6 +867,28 @@ public int getLostBookieRecoveryDelay() throws UnavailableException {
...
+        Watcher w = new Watcher() {
+            @Override
+            public void process(WatchedEvent e) {
+                if (e.getType() == Event.EventType.NodeDeleted && idExtractionPattern.matcher(e.getPath()).find()) {
+                    cb.operationComplete(0, null);
+                }
+            }
+        };
+        try {
+            zkc.addWatch(urLedgerPath, w, AddWatchMode.PERSISTENT_RECURSIVE);
```
hangc0276, line 882: It will register a lot of watcher on Zookeeper

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
```diff
@@ -1014,6 +1051,16 @@ public void run() {
         }), initialDelay, interval, TimeUnit.SECONDS);
     }
 
+    private class UnderReplicatedLedgersChangedCb implements GenericCallback<Void> {
+        @Override
+        public void operationComplete(int rc, Void result) {
+            Iterator<UnderreplicatedLedger> underreplicatedLedgersInfo = ledgerUnderreplicationManager
+                    .listLedgersToRereplicate(null);
+            underReplicatedLedgersGuageValue.set(Iterators.size(underreplicatedLedgersInfo));
```
hangc0276, line 1059: The `Iterators.size` will iterate the whole Znodes in an under-replicated path which brings heavy pressure on Zookeeper and lead to dead lock.

hangc0276 (commented):
-1 for this change

hangc0276:
This PR introduced a deadlock described: https://github.com/apache/pulsar/pull/21010

PR #2822, opened 2021-10-13
Correct the comment of caching index & filter block with additional options to improve performance

Myasuka:

### Motivation

The previous comment of `tableOptions.setCacheIndexAndFilterBlocks(true)` is `Options best suited for HDDs`. If the comment thought that we could cache index&filter blocks when using HDD for better performance, however, this is not correct as this option actually helps on limit the memory usage of index & filter blocks. More internal details could refer to https://github.com/facebook/rocksdb/wiki/Block-Cache#caching-index-filter-and-compression-dictionary-blocks

Apart from correcting the comment, this PR also add additional options to help improve performance.

### Changes

Correct the comment of caching index & filter block with additional options to improve performance.


StevenLuMT:

PR #2832, opened 2021-10-15
Fix semaphore leak when EntryMemTable#addEntry accepts the same entries

BewareMyPower:
### Motivation

If `EntryMemTable#addEntry` accepts the same entries (assuming size is `N`), which means they have the same ledger id and entry id, the semaphore will be acquired multiple times (assuming `M` times) but only one entry will be added to `kvmap`. When the `EntryMemTable` is flushed, only `N` permits will be released, while `M * N` permits are acquired.  Then the semaphore leak happens.

### Changes

In `EntryMemTable#addEntry`, release the acquired permits if `internalAdd` returns 0. Because it means the entry size is zero or the entry has already existed in `kvmap`. `testAddSameEntries` is added to verify when the same entries are added, `flush` can still release all acquired permits.


BewareMyPower:
Here's a block thread in production. We can see `addEntry` stuck at `skipListSemaphore.acquireUninterruptibly(len)`, which means the semaphore is exhausted. 

image

PTAL. @eolivelli @dlg99 @nicoloboschi @hangc0276 @codelipenghui

PR #2835, opened 2021-10-18
BP-44 USE metrics

Vanlightly:
BP-44: USE metrics. A proposal for improving BookKeeper metrics so that operators can employ the USE method for diagnosing performance issues.


dlg99:
Overall looks like a great addition.
Can it benefit from MDC? We already pass MDC in the thread pools, adding tags to MDC and using the tags when reporting metrics should be simple.

eolivelli:
@RaulGracia @fpj would you mind to take a look as well?

PR #1755, opened 2018-10-23
 Configure Netty allocator in bookie and client 

merlimat:
### Motivation

This is based on #1754. Adding the code to configure and use the allocator wrapper in bookie and client.

(I'll rebase once the first PR is merged)

PR #2836, opened 2021-10-19
Use OutOfMemoryPolicy when the direct memory is insufficient when reading the entry in ReadCache

wenbingshen:
### Motivation

Original PR: https://github.com/apache/bookkeeper/pull/1755,
It should be that this PR forgot to modify the memory application method.

When the direct memory is insufficient, it does not fall back to the jvm memory, and the bookie hangs directly.

image

image

### Changes

Use `OutOfMemoryPolicy` when the direct memory is insufficient when reading the entry in `ReadCache`.


dlg99:
closed/reopened to re-trigger merge/CI

dlg99 (approved):

PR #2858, opened 2021-10-26
Fix checkstyle on "microbenchmarks"

nicoloboschi:
### Motivation

Currently we are forced to skip `microbenchmarks` checkstyle task because it fails. 
In fact, the "pr-validation" check skips that task.

### Changes

- fixed checkstyle
- removed the skip in the "pr-validation" check


eolivelli (approved):
LGTM
thanks

PR #2860, opened 2021-10-27
Return too many requests error when the write cache flush timeout occurs

pradeepbn:

### Motivation
When the write cache flush timeout occurs in SingleDirectoryDbLedgerStorage either because the disk is slow or if there are too many requests and the disk is busy, instead of returning EIO status to the client, ETOOMANYREQUESTS should be returned which is more appropriate for this situation.

### Changes
- Whenever there is a resource constraint by the bookkeeper internal components OperationRejectedException is thrown. Now this exception send status=ETOOMANYREQUESTS
- Changed the Javadoc of OperationRejectedException to reflect the usage of OperationRejectedException appropriately


dlg99 (approved):
Looks good

ivankelly (approved):

Issue #2834, opened 2021-10-18
BP-44: USE metrics


Vanlightly:
**BP**

This is the master ticket for tracking BP-44 : USE metrics

Updating metrics to allow operators to employ the USE method for diagnosing bottlenecks.

Proposal PR - #2835

shoothzj:
closed by #2835

PR #2861, opened 2021-10-27
BP-44: OrderedExecutor metrics

Vanlightly:
### Motivation

See BP-44 for full motivation. 

### Changes

This PR focuses on the OrderedExecutor and OrderedScheduler.
- Removed double nesting of TimedCallable/TimedRunnable from OrderedExecutor. These were being added by both the decorators and explicitly in the overridden methods causing double recording of metrics.
- Moved thread ordinal numbers in metric names to a label, in line with modern practices.
- Metrics `task_execution` and `task_queued` changed to be thread-scoped, rather than being an aggregate across all threads. This is configurable, disabled by default. Only enabled in the BookieRequestProcessor. Not in the client or any tests.

Master Issue: #2834


eolivelli (approved):
LGTM

PR #2865, opened 2021-10-28
BP-44: Journal metrics enhancement

Vanlightly:
### Motivation

See BP-44 for motivation.

### Changes

The following metrics changes added to provider greater
resolution to existing metrics:
- all journal metrics now include the journalIndex as a label
- journal memory max and used metrics moved into JournalStats
  and scoped to the journal instance (rather than being and
  aggregated metric).

Following changes to allow for time utilization/saturation
calculations:
- journal, force write and callback threads report time spent
- journal queue max size added as a gauge
- number of journal directories added as a gauge

Master Issue: #2834


eolivelli (approved):
LGTM

PR #2866, opened 2021-10-28
BP-44: Rejected requests counter metrics

Vanlightly:
### Motivation

See BP-44 for motivation.

### Changes

Addition of metrics to explicitly record rejected read and write requests. Requests are rejected because either the write thread the request was routed to had a full task queue, or because the DbLedgerStorage write cache is full and the timeout reached.

Having metrics specifically for rejected requests is vitally important for diagnosing when a bookie is under stress.

Additionally, the `addEntryStats` and `readEntryStats` OpStatsLoggers are changed to be thread-scoped as they execute on the read/write thread pools which allows for more detailed drill down on read/write op execution by thread.

Master Issue: #2834


eolivelli (approved):
lgtm

PR #2870, opened 2021-10-29
Add skip unrecoverable ledger option for bookkeeper shell recover command

hangc0276:
### Motivation
When we use `bin/bookkeeper shell recover bookieId` command to recover specific bookie's ledgers, the recover process will exit when occurs recover ledger failed.

In our production bookkeeper cluster, we found some ledgers in Open state and has no entry.  When we call `bin/bookkeeper shell recover bookieId` command, it will traverse all the ledgers level by level. In the end, for each ledger, it will call the following code to process recover.
```Java
Processor<Long> ledgerProcessor = new Processor<Long>() {
            @Override
            public void process(Long ledgerId, AsyncCallback.VoidCallback iterCallback) {
                recoverLedger(bookiesSrc, ledgerId, dryrun, skipOpenLedgers, skipUnrecoverableLedgers, iterCallback);
            }
};
```

In the `recoverLedger` method, it will call `asyncOpenLedgerNoRecovery` to open ledger and get LAC if the ledger in `OPEN` state. For the `getLAC` request, if the request ledger has no entry, it will return entry = -1 and return ERROR for this `getLAC` request.
https://github.com/apache/bookkeeper/blob/98ddf8149592572eebcfaf6bdd4916f295ffd9d7/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperAdmin.java#L756-L769

And for the `asyncOpenLedgerNoRecovery` callback, it will return error for this process. It will stop the recover process of the following ledgers.

In the end, the recover command runs failed, and the following ledger can't be recovered.

### Changes
We should expose a flag for user to determine whether to move forward to recover the following ledgers when some ledgers recover failed.

So, I provide the parameter `sku` to handle this case.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperAdmin.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperAdmin.java
@@ -835,13 +835,20 @@ public class BookKeeperAdmin implements AutoCloseable {
                         @Override
                         public void openComplete(int newrc, final LedgerHandle newlh, Object newctx) {
                             if (newrc != BKException.Code.OK) {
-                                LOG.error("BK error close ledger: {}", lId, BKException.create(newrc));
-                                finalLedgerIterCb.processResult(newrc, null, null);
+                                if (skipUnrecoverableLedgers) {
+                                    LOG.warn("BK error opening ledger: {}, skip recover it.",
+                                        lId, BKException.create(newrc));
+                                    finalLedgerIterCb.processResult(BKException.Code.OK, null, null);
+                                } else {
+                                    LOG.error("BK error close ledger: {}", lId, BKException.create(newrc));
+                                    finalLedgerIterCb.processResult(newrc, null, null);
+                                }
                                 return;
                             }
                             bkc.mainWorkerPool.submit(() -> {
                                 // do recovery
-                                recoverLedger(bookiesSrc, lId, dryrun, skipOpenLedgers, finalLedgerIterCb);
+                                recoverLedger(bookiesSrc, lId, dryrun, skipOpenLedgers,
+                                    skipUnrecoverableLedgers, finalLedgerIterCb);
                             });
                         }
                     }, null);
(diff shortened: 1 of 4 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperAdmin.java
```diff
@@ -798,7 +835,7 @@ public void openComplete(int rc, final LedgerHandle lh, Object ctx) {
                         @Override
                         public void openComplete(int newrc, final LedgerHandle newlh, Object newctx) {
                             if (newrc != BKException.Code.OK) {
-                                LOG.error("BK error close ledger: " + lId, BKException.create(newrc));
+                                LOG.error("BK error close ledger: {}", lId, BKException.create(newrc));
                                 finalLedgerIterCb.processResult(newrc, null, null);
```
zymap, line 839: Do we need to complete the `finalLedgerIterCb` normally? In the recover ledger method, I saw multiple places still failed the `finalLedgerIterCb` and it will break the recover process. Should we handle them?

hangc0276: Thanks for you reply @zymap. You are right,  I update the code to handle them, Please help take a look, thanks.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperAdmin.java
```diff
@@ -798,13 +835,20 @@ public void openComplete(int rc, final LedgerHandle lh, Object ctx) {
                         @Override
                         public void openComplete(int newrc, final LedgerHandle newlh, Object newctx) {
                             if (newrc != BKException.Code.OK) {
-                                LOG.error("BK error close ledger: " + lId, BKException.create(newrc));
-                                finalLedgerIterCb.processResult(newrc, null, null);
+                                if (skipUnrecoverableLedgers) {
```
codelipenghui, line 838: Do we need to check the rc is `NoSuchEntryException` ?

zymap: I think we don't need to check `NoSuchEntryException`. We are just opening a ledger not read data and it looks like there is no way to throw that exception?

hangc0276: For asyncOpenLedger operation, the rc only has `BKException.Code.ReadException` or `Exception.Code.OK`
```java
lh.asyncReadLastConfirmed(new ReadLastConfirmedCallback() {
                @Override
                public void readLastConfirmedComplete(int rc,
                        long lastConfirmed, Object ctx) {
                        closeLedgerHandle();
                        openComplete(bk.getReturnRc(BKException.Code.ReadException), null);
                    } else {
                        lh.lastAddConfirmed = lh.lastAddPushed = lastConfirmed;
                        openComplete(BKException.Code.OK, lh);
                    }
                }
            }, null);
```
So we doesn't need to check `NoSuchEntryException`.

hangc0276:
rerun failure checks

zymap (approved):
LGTM

dlg99 (approved):

PR #2871, opened 2021-10-29
add bookie expand or shrink directories operation guide

hangc0276:
### Motivation

Currently, BookKeeper website doesn't have any operation guide for shrink or expand journal/ledger directories.

### Changes
Add the operation guide for shrink or expand journal/ledger directories.

### DIFF
--- a/site/docs/latest/admin/diskoperation.md
+++ b/site/docs/latest/admin/diskoperation.md
@@ -1,0 +1,74 @@
+---
+title: Journal/Ledger directories expand or shrink
+---
+
+If you want to expand or shrink journal/ledger directories, you can follow the following operation guide.
+
+**STEP1**. Get bookie instanceId 
+
+```bash
+$bin/bookkeeper shell whatisinstanceid
+``` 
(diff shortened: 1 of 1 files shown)


zymap:
Here has a mail about this: http://mail-archives.apache.org/mod_mbox/bookkeeper-dev/202111.mbox/browser

Vanlightly:
> Here has a mail about this: http://mail-archives.apache.org/mod_mbox/bookkeeper-dev/202111.mbox/browser

Considering that data loss is an outcome of this operation, I think that should be clearly communicated here. When this data loss is addressed in a future patch we can modify the docs.

dlg99:
maybe it is better to recommend to NOT use the shrink/expand procedure at all and instead decomm the bookie/add it back with shrunk/expanded storage.

StevenLuMT:
hi @hangc0276 
I would like to know how the current PR is going.

StevenLuMT:

jiazhai (approved):

Issue #2872, opened 2021-10-29
BookieJournalTest fails on JDK11 


eolivelli:
**BUG REPORT**

***Describe the bug***

BookieJournalTest fails with the error below, on ubuntu on JDK11.

***To Reproduce***
Run the test on JDK11 with a virtual machine for which there is no mapping for the "default host"

***Expected behavior***

No error.

***Screenshots***
```
java.lang.IllegalAccessError: class javax.naming.spi.NamingManager (in unnamed module @0x5edf2821) cannot access class jdk.internal.loader.ClassLoaderValue (in module java.base) because module java.base does not export jdk.internal.loader to unnamed module @0x5edf2821
        at javax.naming.spi.NamingManager.<clinit>(NamingManager.java:86)
        at javax.naming.InitialContext.getURLOrDefaultInitCtx(InitialContext.java:332)
        at javax.naming.directory.InitialDirContext.getURLOrDefaultInitDirCtx(InitialDirContext.java:106)
        at javax.naming.directory.InitialDirContext.getAttributes(InitialDirContext.java:142)
        at org.apache.bookkeeper.net.DNS.reverseDns(DNS.java:83)
        at org.apache.bookkeeper.net.DNS.getHosts(DNS.java:239)
        at org.apache.bookkeeper.net.DNS.getDefaultHost(DNS.java:338)
        at org.apache.bookkeeper.net.DNS.getDefaultHost(DNS.java:354)
        at org.apache.bookkeeper.bookie.Bookie.getBookieAddress(Bookie.java:583)
        at org.apache.bookkeeper.proto.SimpleBookieServiceInfoProvider.<init>(SimpleBookieServiceInfoProvider.java:39)
        at org.apache.bookkeeper.bookie.Bookie.<init>(Bookie.java:648)
        at org.apache.bookkeeper.bookie.BookieJournalTest.createBookieAndReadJournal(BookieJournalTest.java:658)
        at org.apache.bookkeeper.bookie.BookieJournalTest.testV4Journal(BookieJournalTest.java:399)
```

Shoothzj:
closed by #2787

PR #2874, opened 2021-11-03
Allow run bookkeeper-server tests with jdk11

zymap:
---

Master Issue: #2872

*Motivation*

Allow run bookkeeper-server tests with jdk11 using gradle

*Modifications*

Add options `java.base/jdk.internal.loader=ALL-UNNAMED` for jvm

### DIFF
--- a/bookkeeper-server/build.gradle
+++ b/bookkeeper-server/build.gradle
@@ -99,4 +99,7 @@ test.doFirst {
         .resolvedArtifacts.find { it.name == 'junit-foundation' }
     jvmArgs("-Djunit.timeout.test=600000", "-Djunit.max.retry=3",
             "-Djava.net.preferIPv4Stack=true", "-Dio.netty.leakDetection.level=paranoid")
+    if (JavaVersion.current() > JavaVersion.VERSION_1_8) {
+        jvmArgs("--add-opens", "java.base/jdk.internal.loader=ALL-UNNAMED")
+    }
 }


bookkeeper-server/build.gradle
```diff
@@ -99,4 +99,7 @@ test.doFirst {
         .resolvedArtifacts.find { it.name == 'junit-foundation' }
     jvmArgs("-Djunit.timeout.test=600000", "-Djunit.max.retry=3",
             "-Djava.net.preferIPv4Stack=true", "-Dio.netty.leakDetection.level=paranoid")
+    if (JavaVersion.current() > JavaVersion.VERSION_1_8) {
+        jvmArgs("--add-opens", "java.base/jdk.internal.loader=ALL-UNNAMED")
```
pkumar-singh, line 103: What problem does this PR solve? Or in other words what this jvmArgs does and lacking this will result in what problem?

pkumar-singh: Another thins to note is. For maven `mvn clean package -B -nsu -DskipBookKeeperServerTests -Dorg.slf4j.simpleLogger.defaultLogLevel=INFO ` therefore for the gradle we are not running bookkeeper-server test on Java 11 as of now.

merlimat: I think we should make sure the tests are passing even without that flag, otherwise we might end up in a scenario where the flag is needed for running the bookie or bk client.

eolivelli: I agree with @merlimat we should find another way of doing this.

@pkumar-singh this is my error https://github.com/apache/bookkeeper/issues/2872 that this PR is trying to solve

zymap: @merlimat @eolivelli I think it only affects the tests. I found a similar fix in pulsar https://github.com/apache/pulsar/pull/9893

PR #2876, opened 2021-11-03
There should only be a single BookieImpl constructor

Vanlightly:
### Motivation

To have DI on the bookie, it's better to have a single constructor
that takes all the injected implementations. With a single
constructor, there's only one place modify in production code when we
want to inject something, rather than having the bookie need to know
how to create a "default" version (which often breaks encapsulation).

If we need convenience constructors for tests, they should live in the
tests.

### Changes

The principle changes are:
- remove the BookieImpl constructor which takes only a ServerConfiguration and chooses default values for the other ijectable dependencies.
- add a TestBookieImpl for use in tests which has a constructor that matches the one removed from BookieImpl.

This is one of many preliminary PRs with refactorings required for the "BP-46 Running without journal" work.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
@@ -30,7 +30,6 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Lists;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufAllocator;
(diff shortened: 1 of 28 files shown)


merlimat (approved):
LGTM. An alternative approach could also be to add builders so that it becomes easier to add more arguments without changing code in multiple places.

nicoloboschi (approved):

dlg99 (approved):

eolivelli (approved):
LGTM

eolivelli:
PTAL to failed tests on CI

Vanlightly:
rerun failure checks

Vanlightly:
The failing test has been fixed.

PR #2880, opened 2021-11-05
BP-44: Removed bookie port number from OrderedExecutor name

Vanlightly:
### Motivation

Including the port number only complicates things for
admins who rely on OrderedExecutor metrics for alerts
and dashboards. Removing the port number reduces,
the complexity for deployments which use various
non-standard bookie ports.

### Changes

Removed port number from high priority and long poll thread pool names.

Master Issue: #2834


eolivelli (approved):
LGTM

PR #2883, opened 2021-11-05
fix: Stream Storage tests are flaky; statelib test times out

dlg99:

Fixed tests
### Motivation

Stream Storage tests are flaky; statelib test times out

### Changes

MockExecutor managed to get into the situation when it submitted the task to the same thread it was running on.
Not sure why it didn't happen consistently, locally I can only repro this in the debugger and only when I step over some code.


eolivelli (approved):
Good catch.
Probably with gradle we are executing the tests in a slightly different way?

Lgtm

PR #2887, opened 2021-11-10
issue #2879 : let bookie quit if journal thread exit

### LINKED ISSUE #2879: Journal thread exit while bookie is still on
**BUG REPORT**
Bookie does not exist when the on journal thread exists.
***Describe the bug***
I have a bookie which has three directories, after running for some days, the bookie is not healthy when test by 
```shell
bin/bookkeeper shell bookiesanity 
```
And the jstack show almost all `bookie-io` thread are waiting for journal adding an entry to the queue.

```
"bookie-io-1-20" #44 prio=5 os_prio=0 tid=0x00007f4741f02800 nid=0x57d8 waiting on condition [0x00007f46f7074000]
   java.lang.Thread.State: WAITING (parking)
	at sun.misc.Unsafe.park(Native Method)
	- parking to wait for  <0x00000005778a3c88> (a java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject)
	at java.util.concurrent.locks.LockSupport.park(LockSupport.java:175)
	at java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.await(AbstractQueuedSynchronizer.java:2039)
	at java.util.concurrent.ArrayBlockingQueue.put(ArrayBlockingQueue.java:353)
	at org.apache.bookkeeper.bookie.Journal.logAddEntry(Journal.java:863)
	at org.apache.bookkeeper.bookie.Journal.logAddEntry(Journal.java:852)
	at org.apache.bookkeeper.bookie.Journal.logAddEntry(Journal.java:842)
	at org.apache.bookkeeper.bookie.Bookie.addEntryInternal(Bookie.java:1265)
```
And also, there are only two threads for the journal left which means one journal thread exists.
So, all `addEntry` requests to this journal will be blocked after the queue is full, and eventually, all `bookie-io` threads will be blocked.

***Expected behavior***

Bookie should suicide at the point the journal thread exit .

***Additional context***

aloyszhang:
fix #2879 
This pull request let bookie quit when there's journal thread exit

### Motivation

As described in #2879, now if a bookie has multi journal directories means it has multi journal thread. Once a journal thread exits, the bookie will be unhealthy due to the block of all bookie-io threads, and then the bookie will not work but progress is still alive.
This pull request tries to fix this problem.

### Changes

check the journal thread alive in a fixed interval, let bookie quit once there's a journal thread exit

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
@@ -879,7 +879,6 @@ public class BookieImpl extends BookieCriticalThread implements Bookie {
                 for (Journal journal : journals) {
                     journal.shutdown();
                 }
-                this.join();
 
                 // Shutdown the EntryLogger which has the GarbageCollector Thread running
                 ledgerStorage.shutdown();
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
@@ -51,6 +51,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import org.apache.bookkeeper.bookie.BookieException.DiskPartitionDuplicationException;
@@ -425,11 +426,13 @@ public class BookieImpl extends BookieCriticalThread implements Bookie {
             }
         }
 
+        JournalAliveListener journalAliveListener =
+                () -> BookieImpl.this.triggerBookieShutdown(ExitCode.BOOKIE_EXCEPTION);
         // instantiate the journals
         journals = Lists.newArrayList();
         for (int i = 0; i < journalDirectories.size(); i++) {
             journals.add(new Journal(i, journalDirectories.get(i),
-                    conf, ledgerDirsManager, statsLogger.scope(JOURNAL_SCOPE), allocator));
+                    conf, ledgerDirsManager, statsLogger.scope(JOURNAL_SCOPE), allocator, journalAliveListener));
         }
 
         this.entryLogPerLedgerEnabled = conf.isEntryLogPerLedgerEnabled();
@@ -849,12 +852,13 @@ public class BookieImpl extends BookieCriticalThread implements Bookie {
     public int shutdown() {
         return shutdown(ExitCode.OK);
     }
-
     // internal shutdown method to let shutdown bookie gracefully
     // when encountering exception
-    synchronized int shutdown(int exitCode) {
+    ReentrantLock lock = new ReentrantLock(true);
+    int shutdown(int exitCode) {
+        lock.lock();
         try {
-            if (isRunning()) { // avoid shutdown twice
+            if (isRunning()) {
                 // the exitCode only set when first shutdown usually due to exception found
                 LOG.info("Shutting down Bookie-{} with exitCode {}",
                          conf.getBookiePort(), exitCode);
@@ -892,6 +896,7 @@ public class BookieImpl extends BookieCriticalThread implements Bookie {
             LOG.error("Got Exception while trying to shutdown Bookie", e);
             throw e;
         } finally {
+            lock.unlock();
             // setting running to false here, so watch thread
             // in bookie server know it only after bookie shut down
             stateManager.close();
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
@@ -701,6 +701,8 @@ public class Journal extends BookieCriticalThread implements CheckpointSource {
     // Expose Stats
     private final JournalStats journalStats;
 
+    private JournalAliveListener journalAliveListener;
+
     public Journal(int journalIndex, File journalDirectory, ServerConfiguration conf,
             LedgerDirsManager ledgerDirsManager) {
         this(journalIndex, journalDirectory, conf, ledgerDirsManager, NullStatsLogger.INSTANCE,
@@ -780,6 +782,13 @@ public class Journal extends BookieCriticalThread implements CheckpointSource {
         // Expose Stats
         this.journalStats = new JournalStats(journalStatsLogger, journalMaxMemory,
                 () -> memoryLimitController.currentUsage());
+    }
+
+    public Journal(int journalIndex, File journalDirectory, ServerConfiguration conf,
+                   LedgerDirsManager ledgerDirsManager, StatsLogger statsLogger,
+                   ByteBufAllocator allocator, JournalAliveListener journalAliveListener) {
+        this(journalIndex, journalDirectory, conf, ledgerDirsManager, statsLogger, allocator);
+        this.journalAliveListener = journalAliveListener;
     }
 
     JournalStats getJournalStats() {
@@ -1253,6 +1262,9 @@ public class Journal extends BookieCriticalThread implements CheckpointSource {
             // close will flush the file system cache making any previous
             // cached writes durable so this is fine as well.
             IOUtils.close(LOG, bc);
+            if (journalAliveListener != null) {
+                journalAliveListener.onJournalExit();
+            }
         }
         LOG.info("Journal exited loop!");
     }
--- a/pom.xml
+++ b/pom.xml
@@ -172,6 +172,7 @@
     <spotbugs-annotations.version>4.6.0</spotbugs-annotations.version>
     <javax-annotations-api.version>1.3.2</javax-annotations-api.version>
     <testcontainers.version>1.15.1</testcontainers.version>
+    <awaitility.version>4.0.2</awaitility.version>
     <vertx.version>3.9.8</vertx.version>
     <zookeeper.version>3.8.0</zookeeper.version>
     <snappy.version>1.1.7.7</snappy.version>
@@ -766,6 +767,11 @@
         <groupId>org.jsoup</groupId>
         <artifactId>jsoup</artifactId>
         <version>${jsoup.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.awaitility</groupId>
+        <artifactId>awaitility</artifactId>
+        <version>${awaitility.version}</version>
       </dependency>
 
       <!-- benchmark dependencies -->
--- a/pom.xml
+++ b/pom.xml
@@ -172,7 +172,6 @@
     <spotbugs-annotations.version>4.6.0</spotbugs-annotations.version>
     <javax-annotations-api.version>1.3.2</javax-annotations-api.version>
     <testcontainers.version>1.15.1</testcontainers.version>
-    <awaitility.version>4.0.2</awaitility.version>
     <vertx.version>3.9.8</vertx.version>
     <zookeeper.version>3.8.0</zookeeper.version>
     <snappy.version>1.1.7.7</snappy.version>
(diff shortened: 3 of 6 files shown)


Vanlightly:
We don't need a new configuration for this. This kind of thing should be a hidden implementation detail, there is no benefit to allowing users to configure an interval.

Another approach would be something like passing a single CountdownLatch to each journal that can be waited on from the bookie thread.

Also note that any dependency changes need to be added to Gradle or won't compile.

aloyszhang:
@Vanlightly 
CountdownLatch sounds good. Thanks for your suggestions.

aloyszhang:
I'm not familiar with gradle and I'll check the dependency problem first.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
```diff
@@ -714,12 +717,14 @@ public BookieImpl(ServerConfiguration conf, StatsLogger statsLogger,
...
-
+        this.journalExistCountDown = new CountDownLatch(1);
         // instantiate the journals
         journals = Lists.newArrayList();
         for (int i = 0; i < journalDirectories.size(); i++) {
-            journals.add(new Journal(i, journalDirectories.get(i),
-                    conf, ledgerDirsManager, statsLogger.scope(JOURNAL_SCOPE), allocator));
+            Journal journal = new Journal(i, journalDirectories.get(i),
+                    conf, ledgerDirsManager, statsLogger.scope(JOURNAL_SCOPE), allocator);
+            journal.setThreadAliveCounterDown(journalExistCountDown);
```
Vanlightly, line 726: Better for dependencies to be injected through the constructor. Tests use one constructor, which doesn't need the latch, and the BookieImpl uses the other. So you won't need to update lots of tests.

aloyszhang: Actually, if we decide whether a bookie thread should exit by the CountDownLatch,  many behaviours in the unit test are broken even if `Journal` has two constructors. 
For lots of tests, the `Journal` does not have the CountDownLatch,  shutdown the `BookieImpl` or stop the `BookieServer` will be blocked.

Vanlightly (changes requested):
The JournalAliveListener approach looks good. Should it be named JournalDeathListener though? We do need to check that concurrent invocations to shutdown() either directly due to BookieServer shutdown or a journal exiting is handled. There is already handling for multiple invocations of triggerBookieShutdown, but not from that and a BookierServer shutdown. BookieServer.shutdown() should block on an already in-progress BookieImpl.shutdown() to avoid the Java process exiting too soon.

dependencies.gradle
```diff
@@ -84,6 +84,7 @@ depVersions = [
     vertx: "3.9.8",
     yahooDatasketches: "0.8.3",
     zookeeper: "3.6.2",
+    awaitility: "4.0.2",
```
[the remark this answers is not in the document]
aloyszhang: fixed

dependencies.gradle
```diff
@@ -174,6 +175,7 @@ depLibs = [
     metricsJvm: "io.dropwizard.metrics:metrics-jvm:${depVersions.dropwizard}",
     metricsGraphite: "io.dropwizard.metrics:metrics-graphite:${depVersions.dropwizard}",
     mockito: "org.mockito:mockito-core:${depVersions.mockito}",
+    awaitility: "org.awaitility:awaitility:${depVersions.awaitility}",
```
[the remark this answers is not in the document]
aloyszhang: fixed

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -702,6 +704,67 @@ public Journal(int journalIndex, File journalDirectory, ServerConfiguration conf
         this.journalStats = new JournalStats(statsLogger);
     }
 
+    public Journal(int journalIndex, File journalDirectory, ServerConfiguration conf,
```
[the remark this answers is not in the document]
aloyszhang: fixed

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
```diff
@@ -1179,12 +1196,12 @@ public void run() {
     public int shutdown() {
         return shutdown(ExitCode.OK);
     }
-
+    private final AtomicBoolean shuttingDown = new AtomicBoolean(false);
```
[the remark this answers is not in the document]
aloyszhang: fixed

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
```diff
@@ -715,13 +716,23 @@ public BookieImpl(ServerConfiguration conf, StatsLogger statsLogger,
             }
         }
 
-        // instantiate the journals
         journals = Lists.newArrayList();
-        for (int i = 0; i < journalDirectories.size(); i++) {
-            journals.add(new Journal(i, journalDirectories.get(i),
-                    conf, ledgerDirsManager, statsLogger.scope(JOURNAL_SCOPE), allocator));
+        if (listenJournalAlive) {
```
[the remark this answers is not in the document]
aloyszhang: fixed

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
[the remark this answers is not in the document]
aloyszhang: After checking again, we can't use the `shutdownTriggered`. Since if `triggerBookieShutdown` invokes `shutdown`, `shutdownTriggered` will use twice, and will block `shutdown`

Vanlightly:
@aloyszhang I mentioned in a previous comment that we need to ensure that if a shutdown is trigger by an OS signal, but a shutdown is in progress, then it needs to wait for that shutdown to complete, else the process may exit before the shutdown is complete. I didn't see that the BookieImpl.shutdown() method is synchronized but it is, and so should do that blocking for us.

Being synchronized also means that we shouldn't need that AtomicBoolean `shuttingDown` as if only one thread can execute it at a time then the `isRunning()` check is enough. However, when I tested it I found that two threads were able to enter the method concurrently! This seems due to the `this.join()` as that is the moment when another thread is permitted to enter the method. So synchronized isn't doing the job for us. The AtomicBoolean you added also isn't enough because we need an OS signal triggered shutdown to be blocked until an in-progress shutdown is completed.

So that leads us to using an explicit lock on the shutdown method to replace `synchronized` and the AtomicBoolean you added. With those changes, we should have everything covered.

Note that I have done a test of a crashing journal followed by an OS sigint and the bookie does terminate before the shutdown was completed, so this is something that does need addressing.

aloyszhang:
@Vanlightly 

> This seems due to the this.join() as that is the moment when another thread is permitted to enter the method. So synchronized isn't doing the job for us. 

Yes, the original `synchronized ` does not work due to the join().

> Note that I have done a test of a crashing journal followed by an OS signal and the bookie does terminate before the shutdown was completed, so this is something that does need addressing.

Do you mean that there is a common shutdown in progress which invoked by the stop command `bookkeeper-deamon stop bookie`, at the same time, the journal thread crashed due to os failure, and the bookie server will shutdown without waiting for the common shutdown to complete.

Vanlightly:
@aloyszhang Exactly, if the `triggerBookieShutdown` happens first, which can happen due to a number of reasons, and then while it is in progress an operator initiates a controlled shutdown then the bookie can terminate before the BookieImpl shutdown is complete.

aloyszhang:
[PR Validation](https://pipelines.actions.githubusercontent.com/EMeCU0s3nV8mEzpqEGqbbDjtOUKCLJV4IPXg8mh4qtgdpzCGW6/_apis/pipelines/1/runs/15672/signedlogcontent/2?urlExpires=2021-11-18T18%3A24%3A00.7029144Z&urlSigningMethod=HMACV1&urlSignature=avrDweCvI23lkG87qvsLU2tMkCI1yAvG197gadU2atY%3D) failed.  But from the log, I can't find any useful things. Can you help with this?

Vanlightly:
@aloyszhang run this to get the report: `./gradlew bookkeeper-server:spotbugsMain`

aloyszhang:
@Vanlightly Validation issue has been resolved, PTAL, thanks!

Vanlightly:
@aloyszhang An exception thrown during the shutdown would leave the CountdownLatch at 1, blocking the second caller. I would replace the AtomicBoolean and CountdownLatch with a simple lock, that is acquired before `if (isRunning())` and that would do everything needed.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
```diff
@@ -1184,56 +1184,50 @@ public int shutdown() {
     }
     // internal shutdown method to let shutdown bookie gracefully
     // when encountering exception
-    AtomicBoolean shuttingDown = new AtomicBoolean(false);
-    private final CountDownLatch countDownLatch = new CountDownLatch(1);
+    ReentrantLock lock = new ReentrantLock(true);
     int shutdown(int exitCode) {
         try {
+            lock.lock();
```
Vanlightly, line 1190: The lock should be placed before the try.

aloyszhang: fixed

aloyszhang:
@Vanlightly PTAL

Vanlightly (approved):
LGTM
@eolivelli @nicoloboschi PTAL

aloyszhang:
CC @eolivelli PTAL, thanks!

eolivelli (approved):
LGTM

can you please resolve the conflicts ?
the patch is ready to be merged

aloyszhang:
after rebase the master branch, there're some test failed, I'll resolve this soon.

dlg99:
@aloyszhang Please rebase/resolve merge conflicts

aloyszhang:
Sorry for late, I'll do this soon

Shoothzj (approved):

hangc0276:
ping @aloyszhang, Would you please rebase the master?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
```diff
@@ -849,12 +852,13 @@ public void run() {
     public int shutdown() {
         return shutdown(ExitCode.OK);
     }
-
     // internal shutdown method to let shutdown bookie gracefully
     // when encountering exception
-    synchronized int shutdown(int exitCode) {
+    ReentrantLock lock = new ReentrantLock(true);
+    int shutdown(int exitCode) {
```
hangc0276, line 858: Why use `ReentrantLock` instead of `synchronized` ?

pom.xml
```diff
@@ -172,6 +172,7 @@
     <spotbugs-annotations.version>4.6.0</spotbugs-annotations.version>
     <javax-annotations-api.version>1.3.2</javax-annotations-api.version>
     <testcontainers.version>1.15.1</testcontainers.version>
+    <awaitility.version>4.0.2</awaitility.version>
```
hangc0276, line 175: duplicated import

pom.xml
```diff
@@ -767,6 +768,11 @@
         <artifactId>jsoup</artifactId>
         <version>${jsoup.version}</version>
       </dependency>
+      <dependency>
```
hangc0276, line 771: duplicated import

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
aloyszhang: > I think the bad case is that if a journal crash first and already triggers BookieImpl.triggerBookieShutdown(), then we shutdown bookie by the command bookkeeper-deamon stop bookie, the shutdown progress by the command may not execute completely. 

This is the reason.

zymap:
@aloyszhang Could you rebase the master and address the comment?

StevenLuMT (approved):
nice

StevenLuMT (changes requested):
check failed,please rebase from the newest master branch

image

hangc0276 (approved):
Good job!

PR #2888, opened 2021-11-10
Reorder the sequence of the bookkeeper server shutdown so that there are no read/ write ops while shutting down the bookie

pradeepbn:

### Motivation

After the readcache has been shutdown, all the buffers will be released in memory. In the meantime, if read and write ops are performed on the read cache, then it results in the segmentation fault on invalid memory. So, this PR will shutdown the request processor that does readops and writeops before the bookie shutdown sequence.

### Changes

- Stop all the requestProcessor threads before bookie shutdown sequence to avoid read/write ops during bookie shutdown
- Force shutdown the executors so that it is ensured to stop all executor threads before bookie shutdown.

### How to reproduce
- Apply this diff https://gist.github.com/pradeepbn/a033cae7171f9e6da9a7f92737f843b9 on the master branch to reproduce the issue
- Use this (https://gist.github.com/pradeepbn/65ef387a164953fb3f90057860b12da1) client to pump traffic and initiate reads
- Wait until the console shows `Waiting for reads to be done` and kill the bookie with ctrl+c

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
@@ -272,6 +272,7 @@ public class BookieRequestProcessor implements RequestProcessor {
         }
         shutdownExecutor(highPriorityThreadPool);
         requestTimer.stop();
+        LOG.info("Closed RequestProcessor");
     }
 
     private OrderedExecutor createExecutor(
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
@@ -264,6 +264,7 @@ public class BookieRequestProcessor implements RequestProcessor {
 
     @Override
     public void close() {
+        LOG.info("Closing RequestProcessor");
         shutdownExecutor(writeThreadPool);
         shutdownExecutor(readThreadPool);
         if (serverCfg.getNumLongPollWorkerThreads() > 0 || readThreadPool == null) {
@@ -294,7 +295,8 @@ public class BookieRequestProcessor implements RequestProcessor {
 
     private void shutdownExecutor(OrderedExecutor service) {
         if (null != service) {
-            service.forceShutdown(1000, TimeUnit.SECONDS);
+            service.shutdown();
+            service.forceShutdown(10, TimeUnit.SECONDS);
         }
     }
 
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
@@ -294,7 +294,7 @@ public class BookieRequestProcessor implements RequestProcessor {
 
     private void shutdownExecutor(OrderedExecutor service) {
         if (null != service) {
-            service.shutdown();
+            service.forceShutdown(1000, TimeUnit.SECONDS);
         }
     }
 
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PacketProcessorBase.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PacketProcessorBase.java
@@ -66,7 +66,9 @@ abstract class PacketProcessorBase<T extends Request> extends SafeRunnable {
     }
 
     protected void sendResponse(int rc, Object response, OpStatsLogger statsLogger) {
-        channel.writeAndFlush(response, channel.voidPromise());
+        if (channel.isActive()) {
+            channel.writeAndFlush(response, channel.voidPromise());
+        }
         if (BookieProtocol.EOK == rc) {
             statsLogger.registerSuccessfulEvent(MathUtils.elapsedNanos(enqueueNanos), TimeUnit.NANOSECONDS);
         } else {
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PacketProcessorBase.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PacketProcessorBase.java
@@ -68,6 +68,8 @@ abstract class PacketProcessorBase<T extends Request> extends SafeRunnable {
     protected void sendResponse(int rc, Object response, OpStatsLogger statsLogger) {
         if (channel.isActive()) {
             channel.writeAndFlush(response, channel.voidPromise());
+        } else {
+            LOGGER.info("Netty channel is inactive, hence bypassing netty channel writeAndFlush during sendResponse");
         }
         if (BookieProtocol.EOK == rc) {
             statsLogger.registerSuccessfulEvent(MathUtils.elapsedNanos(enqueueNanos), TimeUnit.NANOSECONDS);
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PacketProcessorBase.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PacketProcessorBase.java
@@ -69,7 +69,8 @@ abstract class PacketProcessorBase<T extends Request> extends SafeRunnable {
         if (channel.isActive()) {
             channel.writeAndFlush(response, channel.voidPromise());
         } else {
-            LOGGER.info("Netty channel is inactive, hence bypassing netty channel writeAndFlush during sendResponse");
+            LOGGER.debug("Netty channel {} is inactive, "
+                    + "hence bypassing netty channel writeAndFlush during sendResponse", channel);
         }
         if (BookieProtocol.EOK == rc) {
             statsLogger.registerSuccessfulEvent(MathUtils.elapsedNanos(enqueueNanos), TimeUnit.NANOSECONDS);
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PacketProcessorBaseV3.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PacketProcessorBaseV3.java
@@ -87,25 +87,26 @@ public abstract class PacketProcessorBaseV3 extends SafeRunnable {
                 requestProcessor.invalidateBlacklist(channel);
             }
         }
-
-        channel.writeAndFlush(response).addListener(new ChannelFutureListener() {
-            @Override
-            public void operationComplete(ChannelFuture future) throws Exception {
-                long writeElapsedNanos = MathUtils.elapsedNanos(writeNanos);
-                if (!future.isSuccess()) {
-                    requestProcessor.getRequestStats().getChannelWriteStats()
-                        .registerFailedEvent(writeElapsedNanos, TimeUnit.NANOSECONDS);
-                } else {
-                    requestProcessor.getRequestStats().getChannelWriteStats()
-                        .registerSuccessfulEvent(writeElapsedNanos, TimeUnit.NANOSECONDS);
+        if (channel.isActive()) {
+            channel.writeAndFlush(response).addListener(new ChannelFutureListener() {
+                @Override
+                public void operationComplete(ChannelFuture future) throws Exception {
+                    long writeElapsedNanos = MathUtils.elapsedNanos(writeNanos);
+                    if (!future.isSuccess()) {
+                        requestProcessor.getRequestStats().getChannelWriteStats()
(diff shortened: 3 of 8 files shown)


Vanlightly:
The order of the shutdown looks good. One significant problem is that the shutdown of the OrderExecutors was not done quite right, causing the shutdown to block for a significant time period (1000 seconds x number of total threads).

The forceShutdown method should only be called after having called shutdown() as what forceShutdown does is simply wait for the ExecutorService to terminate, for those 1000 seconds calling shutdownNow() if the timeout is reached. Because shutdown() wasn't called first the ExecutorService is still running normally. So you need to ensure that `service.shutdown()` is called first.

I would also say 1000 seconds is quite high, this is the upper bound for each thread, called serially. If we call shutdown() on all executors first, then we could time box the total time we're willing to wait for all executors to be terminated.

docker/scripts/entrypoint.sh
```diff
@@ -41,7 +41,7 @@ function run_command() {
         chmod -R +x ${BINDIR}
         chmod -R +x ${SCRIPTS_DIR}
         echo "This is root, will use user $BK_USER to run command '$@'"
-        sudo -s -E -u "$BK_USER" /bin/bash "$@"
+        exec sudo -s -E -u "$BK_USER" /bin/bash "$@"
```
Vanlightly, line 44: This is already covered by #2857

pradeepbn: reverted the change

pradeepbn:
> The order of the shutdown looks good. One significant problem is that the shutdown of the OrderExecutors was not done quite right, causing the shutdown to block for a significant time period (1000 seconds x number of total threads).
> 
> The forceShutdown method should only be called after having called shutdown() as what forceShutdown does is simply wait for the ExecutorService to terminate, for those 1000 seconds calling shutdownNow() if the timeout is reached. Because shutdown() wasn't called first the ExecutorService is still running normally. So you need to ensure that `service.shutdown()` is called first.
> 
> I would also say 1000 seconds is quite high, this is the upper bound for each thread, called serially. If we call shutdown() on all executors first, then we could time box the total time we're willing to wait for all executors to be terminated.

@Vanlightly as discussed bumped it down to 10 seconds. By the time the first thread in the sequence is flushed, we expect other threads to have flushed as well. So, the max upper bound will be 10 + 5s.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
```diff
@@ -264,6 +264,7 @@ int maxReadsInProgressCount() {
 
     @Override
     public void close() {
+        LOG.info("Closing RequestProcessor");
```
hsaputra, line 267: Any additional info with this log to help observability or traceability? 
I am not sure logging this will help troubleshooting anything.

pradeepbn: That's a good idea @hsaputra I think we should holistically look at the improving observability across the entire shutdown sequence. Thoughts? @hsaputra @Vanlightly

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PacketProcessorBaseV3.java
```diff
@@ -87,25 +87,26 @@ protected void sendResponse(StatusCode code, Object response, OpStatsLogger stat
...
-                        .registerFailedEvent(writeElapsedNanos, TimeUnit.NANOSECONDS);
-                } else {
-                    requestProcessor.getRequestStats().getChannelWriteStats()
-                        .registerSuccessfulEvent(writeElapsedNanos, TimeUnit.NANOSECONDS);
-                }
-                if (StatusCode.EOK == code) {
-                    statsLogger.registerSuccessfulEvent(MathUtils.elapsedNanos(enqueueNanos), TimeUnit.NANOSECONDS);
-                } else {
-                    statsLogger.registerFailedEvent(MathUtils.elapsedNanos(enqueueNanos), TimeUnit.NANOSECONDS);
+        if (channel.isActive()) {
```
Vanlightly, line 90: This could also be added to PacketProcessorBase.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
Vanlightly: Actually, when I tested this myself, I added a log statement so I could see when it started and completed shutdown of the request processor. Given that it could block for a little bit.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PacketProcessorBaseV3.java
pradeepbn: Done.

Vanlightly (approved):
LGTM

Vanlightly:
@eolivelli could you review?

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PacketProcessorBaseV3.java
eolivelli, line 90: what about adding a "ELSE" branch with a log message ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PacketProcessorBase.java
```diff
@@ -66,7 +66,9 @@ protected boolean isVersionCompatible() {
     }
 
     protected void sendResponse(int rc, Object response, OpStatsLogger statsLogger) {
-        channel.writeAndFlush(response, channel.voidPromise());
+        if (channel.isActive()) {
```
eolivelli, line 69: what about adding a "ELSE" branch with a log message ?

Vanlightly: Currently what happens when the BookieServer is shutdown is that for each in-progress op, when the response is sent it will fail because the channel is closed, and that will get logged as a metric:

```
if (!future.isSuccess()) {
    requestProcessor.getRequestStats().getChannelWriteStats()
                        .registerFailedEvent(writeElapsedNanos, TimeUnit.NANOSECONDS);
} else {
    requestProcessor.getRequestStats().getChannelWriteStats()
                        .registerSuccessfulEvent(writeElapsedNanos, TimeUnit.NANOSECONDS);
 }
```

So either we add an ELSE branch with the same registerFailedEvent call, or we don't use an IF at all and allow it to use the existing logic flow.

@pradeepbn what was the reason for avoiding `channel.writeAndFlush`?

Vanlightly: I'd say this change is not strictly related to shutdown reordering, so I think removing these channel checks can be removed from this PR.

pradeepbn: `bookie1_1  | 02:15:08,926 ERROR Failed to submit a listener notification task. Event loop shut down?
bookie1_1  | java.util.concurrent.RejectedExecutionException: event executor terminated
bookie1_1  |    at io.netty.util.concurrent.SingleThreadEventExecutor.reject(SingleThreadEventExecutor.java:923) ~[netty-common-4.1.68.Final.jar:4.1.68.Final]
bookie1_1  |    at io.netty.util.concurrent.SingleThreadEventExecutor.offerTask(SingleThreadEventExecutor.java:350) ~[netty-common-4.1.68.Final.jar:4.1.68.Final]
bookie1_1  |    at io.netty.util.concurrent.SingleThreadEventExecutor.addTask(SingleThreadEventExecutor.java:343) ~[netty-common-4.1.68.Final.jar:4.1.68.Final]
bookie1_1  |    at io.netty.util.concurrent.SingleThreadEventExecutor.execute(SingleThreadEventExecutor.java:825) ~[netty-common-4.1.68.Final.jar:4.1.68.Final]
bookie1_1  |    at io.netty.util.concurrent.SingleThreadEventExecutor.execute(SingleThreadEventExecutor.java:815) ~[netty-common-4.1.68.Final.jar:4.1.68.Final]
bookie1_1  |    at io.netty.util.concurrent.DefaultPromise.safeExecute(DefaultPromise.java:841) [netty-common-4.1.68.Final.jar:4.1.68.Final]
bookie1_1  |    at io.netty.util.concurrent.DefaultPromise.notifyListeners(DefaultPromise.java:499) [netty-common-4.1.68.Final.jar:4.1.68.Final]
bookie1_1  |    at io.netty.util.concurrent.DefaultPromise.addListener(DefaultPromise.java:184) [netty-common-4.1.68.Final.jar:4.1.68.Final]
bookie1_1  |    at io.netty.channel.DefaultChannelPromise.addListener(DefaultChannelPromise.java:95) [netty-transport-4.1.68.Final.jar:4.1.68.Final]
bookie1_1  |    at io.netty.channel.DefaultChannelPromise.addListener(DefaultChannelPromise.java:30) [netty-transport-4.1.68.Final.jar:4.1.68.Final]
bookie1_1  |    at org.apache.bookkeeper.proto.PacketProcessorBaseV3.sendResponse(PacketProcessorBaseV3.java:91) [bookkeeper-server.jar:?]
bookie1_1  |    at org.apache.bookkeeper.proto.WriteEntryProcessorV3.sendResponse(WriteEntryProcessorV3.java:180) [bookkeeper-server.jar:?]
bookie1_1  |    at org.apache.bookkeeper.proto.WriteEntryProcessorV3$1.writeComplete(WriteEntryProcessorV3.java:108) [bookkeeper-server.jar:?]
bookie1_1  |    at org.apache.bookkeeper.bookie.Journal$QueueEntry.run(Journal.java:335) [bookkeeper-server.jar:?]
bookie1_1  |    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128) [?:?]
bookie1_1  |    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628) [?:?]
bookie1_1  |    at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30) [netty-common-4.1.68.Final.jar:4.1.68.Final]
bookie1_1  |    at java.lang.Thread.run(Thread.java:829) [?:?]`

If we do not check for isActive(), we will get this exception at the time of shutdown. This happens because of nettyserver shutdown before the request processor.

Vanlightly: This is a controlled exception within Netty but it will pollute the log during shutdown. So I think that's its worth keeping the IF statement. What do you think @eolivelli?

eolivelli: ok, let's keep the original "if"

we could add an "else" branch with a DEBUG log that says that we are skipping the write, this would help in debugging tests failures probably one day

pradeepbn: Added the else statement with the logs. CC: @eolivelli @Vanlightly

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PacketProcessorBaseV3.java
pradeepbn: Done

pradeepbn:
rerun failure tests

pradeepbn:
rerun failure checks

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PacketProcessorBase.java
```diff
@@ -66,7 +66,11 @@ protected boolean isVersionCompatible() {
...
+            channel.writeAndFlush(response, channel.voidPromise());
+        } else {
+            LOGGER.info("Netty channel is inactive, hence bypassing netty channel writeAndFlush during sendResponse");
```
eolivelli, line 72: please report the channel(), it will bring useful information about the client that won't receive the response.
please use "debug" and not "info"

pradeepbn: Changed it. As an example, it prints like this: `Netty channel [id: 0x8f9a1f37, L:/192.168.160.3:3181 ! R:/192.168.160.1:61752] is inactive, hence bypassing netty channel writeAndFlush during sendResponse`

eolivelli (approved):
LGTM

zymap:
cherry-picked this to branch-4.14 to resolve the conflict

Issue #2806, opened 2021-09-24
OOM as 1 million ledgers per entry log


suiyuzeng:
**BUG REPORT**

***Describe the bug***
There is about 1M ledgers in per entry log. After running for a while, OOM will appear. And there is still enough memory.
There are two OOM positions, as follows:
Position 1：
2021-09-21 02:22:08,323 [SyncThread-7-1] ERROR org.apache.bookkeeper.bookie.SyncThread - Exception in SyncThread
java.lang.OutOfMemoryError: Java heap space
    at org.apache.bookkeeper.bookie.storage.ldb.WriteCache.forEach(WriteCache.java:222) ~[org.apache.bookkeeper-bookkeeper-server-4.14.1.jar:4.14.1]

Position 2：
2021-09-21 02:24:14,987 [SyncThread-7-1] ERROR org.apache.bookkeeper.bookie.SyncThread - Exception in SyncThread
java.lang.OutOfMemoryError: Java heap space
    at org.apache.bookkeeper.util.collections.ConcurrentLongLongHashMap$Section.rehash(ConcurrentLongLongHashMap.java:673) ~[org.apache.bookkeeper-bookkeeper-server-4.14.1.jar:4.14.1]
    at org.apache.bookkeeper.util.collections.ConcurrentLongLongHashMap$Section.addAndGet(ConcurrentLongLongHashMap.java:456) ~[org.apache.bookkeeper-bookkeeper-server-4.14.1.jar:4.14.1]
    at org.apache.bookkeeper.util.collections.ConcurrentLongLongHashMap.addAndGet(ConcurrentLongLongHashMap.java:186) ~[org.apache.bookkeeper-bookkeeper-server-4.14.1.jar:4.14.1]
    at org.apache.bookkeeper.bookie.EntryLogMetadata.addLedgerSize(EntryLogMetadata.java:47) ~[org.apache.bookkeeper-bookkeeper-server-4.14.1.jar:4.14.1]

main GC log:
2021-09-21T02:22:08.437+0800: 105453.194: [GC pause (G1 Humongous Allocation) 
2021-09-21T02:22:08.449+0800: 105453.206: [Full GC (Allocation Failure)  10G->10G(20G), 1.9652874 secs]
[Eden: 0.0B(992.0M)->0.0B(1024.0M) Survivors: 32.0M->0.0B Heap: 10.3G(20.0G)->10.3G(20.0G)], [Metaspace: 35551K->35539K(1081344K)]
[...]

PR #2891, opened 2021-11-12
ISSUE-2806: OOM as 1 million ledgers per entry log

suiyuzeng:
### Motivation

fix issue [2806](https://github.com/apache/bookkeeper/issues/2806) OOM as 1 million ledgers per entry log

### Changes
1.Add two config entryLogLedgerMapConcurrency and flushEntrySortBufferInitSize, to avoid allocating a large contiguous memory.
2.Reduce memory usage of EntryLogMetadata of entry logs in garbage collector. If the retention time of all the entry is the same, we do not need to extract the entry logs in retention time to load meta to memory. In org.apache.bookkeeper.bookie.GarbageCollectorThread#extractMetaFromEntryLogs, if the entry log in retention time, do not extract the entry log.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogMetadata.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogMetadata.java
@@ -35,10 +35,13 @@ public class EntryLogMetadata {
     private final ConcurrentLongLongHashMap ledgersMap;
 
     public EntryLogMetadata(long logId) {
+        this(logId, 1);
+    }
+    public EntryLogMetadata(long logId, int concurrencyLevel) {
         this.entryLogId = logId;
 
         totalSize = remainingSize = 0;
-        ledgersMap = new ConcurrentLongLongHashMap(256, 1);
+        ledgersMap = new ConcurrentLongLongHashMap(256, concurrencyLevel);
     }
 
     public void addLedgerSize(long ledgerId, long size) {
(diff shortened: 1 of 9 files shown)


Vanlightly:
Regarding garbage collection, retention is not determined by BookKeeper but whoever controls the ledger metadata (Pulsar for example). I don't think using a configured time period is the way to address this. The issue is that we try to store all metadata in memory, but given that each entry log has a ledger map, we can perform a scan process where we incrementally perform GC without all metadata in memory.

suiyuzeng:
> Regarding garbage collection, retention is not determined by BookKeeper but whoever controls the ledger metadata (Pulsar for example). I don't think using a configured time period is the way to address this. The issue is that we try to store all metadata in memory, but given that each entry log has a ledger map, we can perform a scan process where we incrementally perform GC without all metadata in memory.

Thanks for you suggestion.
I meet the issue in puslar for iot. There are millions topics which has the retention time. It works well in this scene. But it may be not common way to the problem like this. A scan process is better for this. I will fix the issue in this way.

suiyuzeng:
If the count of ledger is not veray large but the entry log count is very large, the memory of the ledger map will not be very large and it will lead to io read, especially for extractEntryLogMetadataByScanning. The current way is better for this. How about reserving two way and adding a config to choose?
@Vanlightly

Vanlightly:
@suiyuzeng it could be a way forward. I think depending on the result we may be happy with just using the scan method. So at first make it configurable which will allow us to evaluate both methods under similar conditions.

suiyuzeng:
ok

suiyuzeng:
> @suiyuzeng it could be a way forward. I think depending on the result we may be happy with just using the scan method. So at first make it configurable which will allow us to evaluate both methods under similar conditions.

@Vanlightly  hi, I find #1949 merged recently fix the issue when i develop the scan way. It store the meta into rocksdb. This way reduce the memory and just read the meta from the entrylog for one times. Do we need the scan way any more?

StevenLuMT:
if #1949  fix the issue, please know it @hangc0276 @Vanlightly 
@suiyuzeng please close your pr, thanks

PR #2892, opened 2021-11-12
Reduce flakiness of TestBKDistributedLogManager

dlg99:
### Motivation

TestBKDistributedLogManager is flaky. 
Sometimes it passes on the CI because of the retries, sometimes it fails.
It passes all tests locally (or in debugger/IDE) if the cases run one by one.
Full suite often fails, typically with random cases failing. 

AFAICT, frequently tests fail because of 
```
[DL-io-0] ERROR org.apache.bookkeeper.common.allocator.impl.ByteBufAllocatorImpl - Unable to allocate memory
java.lang.OutOfMemoryError: Direct buffer memory
```
As result bookies fail, and then test cases fail with something like "BKTransmitException: Failed to write to bookkeeper" 

I don't see anything from netty's leak detection.

It does NOT fix https://github.com/apache/bookkeeper/issues/2898 but it helped to get https://github.com/apache/bookkeeper/issues/2898 reproed without hitting OOME/OODME in the process

### Changes

Reduced number of bookies used for the test.
Closed various resources left open in tests.

### DIFF
--- a/stream/distributedlog/core/src/test/java/org/apache/distributedlog/TestBKDistributedLogManager.java
+++ b/stream/distributedlog/core/src/test/java/org/apache/distributedlog/TestBKDistributedLogManager.java
@@ -37,6 +37,8 @@ import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
+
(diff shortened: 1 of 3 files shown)


pkumar-singh (approved):
LGTM.

nicoloboschi (approved):

Vanlightly (approved):
LGTM

PR #2895, opened 2021-11-14
fix Uneven distribution of journal directory bug

1559924775:


Vanlightly:
Changing the routing of ledgers to journal and ledger directories without any kind of data migration process is unsafe. At the least a large proportion of closed ledgers will become unreadable. Worse, open ledgers will be partly written to one directory and partly to another. Any ledger recovery that occurs can end up truncating ledgers.

Just as with storage expansion, we cannot change routing to directories without a controlled data migration process.

If we really do have an issue of a non-uniform distribution of ledger ids, then I think we should fix that instead. Changing ledger id routing is the more dangerous option that requires a migration.

eolivelli (changes requested):
I can't understand this patch.
Why are you deleting one file?
Also you are changing a function that is meant to compute the mod operation.

I suggest to introduce a specific function to assign the journal and not to change the behaviour of other functions.

1559924775:
In our application scenario, we need to provide a single machine with 200m / s capability, so we need to configure the journal to two disks

I think there is no problem with uneven distribution of ledgerids. Ledgerids themselves are just an ID and have no other role. Routing strategies should not consider whether ledgerids are unevenly distributed. They should exclude the influence of the upper layer and provide a general strategy that can route evenly.

1559924775:
Sorry, I just rarely use GitHub

PR #2901, opened 2021-11-17
Construct and inject bookie components

Vanlightly:
This PR is a set of refactoring changes that are required before the "running without the journal" changes can be submitted. I apologise for the size of this PR but due to the refactorings involved, many classes and many tests were impacted. Functionality is basically the same, the changes are around when and where components are constructed and their injection. 

### Motivation

Simplify and centralize the construction of bookie components outside of the BookieImpl. By constructing all dependencies and injecting them into the places required such as BookieImpl, we simplify this process and avoid circular dependencies. Many of these dependencies are added to the lifecycle component stack which cleanly handles their lifetimes.

Also we extract some functionality from BookieImpl into interface/impl, such as cookie validation.

### Changes

The metadata driver is a factory that creates the correct metadata accessors based on the configuration of the bookie. This is used by multiple components, so should be constructed before these components and passed in as a parameter.

Previously it was constructed by the bookie, and other components called into the bookie to get the instance, which broken encapsulation of the bookie and led to a bunch of ugly mocking.

This change also removes a circular dependency in the construction of the RegistrationManager. Previously it was requiring a listener on construction, but the listener needed the bookie instance, so it all went through a supplier. Now the listener is not needed at construction, but rather it is registered then the bookie starts.

This change also adds an injectable interface for cookie validation. The legacy implementation of this interface is then injected into the BookieImpl on creation. The legacy implementation is code moved out of the BookieImpl class itself.

Injecting ledger storage is good for encapsulation, as it stops the ledgerstorage depending on bookie behaviour. To that end, this change highlights a couple of circular dependencies. Notably, between the ledger storage and each of the statemanager, checkpointer and checkpointState. I've removed each of these from LedgerStorage#initialize() so that it is clearer that these are circular, and so that they can be fixed one by one.

Many the components are now shutdown as lifecycle components, rather than by the BookieImpl. The new AutoCloseableLifecycleComponent is provided to allow classes to be added as lifecycle components and their shutdown handled for them. BookieImpl is still responsible for shutting down the LedgerStorage. I would have changed this, but it would have meant a lot of changes, as there are good few ledger storage implementations.

Because components get constructed outside of the BookieImpl, some changes to the tests were required such as introducing the LocalBookie class that would perform these duties in tests.


bookkeeper-common-allocator/src/main/java/org/apache/bookkeeper/common/allocator/ByteBufAllocatorBuilder.java
```diff
@@ -39,7 +39,7 @@ static ByteBufAllocatorBuilder create() {
     /**
      * Finalize the configured {@link ByteBufAllocator}.
      */
-    ByteBufAllocator build();
+    ByteBufAllocatorWithOomHandler build();
```
eolivelli, line 42: is this used by Pulsar ?
Probably it will cause some NoSuchMethodError for users who try to upgrade BookKeeper without upgrading Pulsar (I know some projects that create a new Pulsar package and select different BookKeeper versions)

is it necessary to change this ?

eolivelli: can we move this ByteBufAllocator change to a separate PR ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLoggerAllocator.java
```diff
@@ -185,7 +185,7 @@ private synchronized BufferedLogChannel allocateNewLog(File dirForNextEntryLog,
 
 
     private synchronized void closePreAllocateLog() {
-        if (preallocatedLogId != -1) {
+        if (preallocation != null) {
```
eolivelli, line 188: this change looks unrelated

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperAdmin.java
```diff
@@ -1231,7 +1232,8 @@ public static boolean format(ServerConfiguration conf,
             boolean isInteractive, boolean force) throws Exception {
         return runFunctionWithMetadataBookieDriver(conf, driver -> {
             try {
-                boolean ledgerRootExists = driver.getRegistrationManager().prepareFormat();
+                RegistrationManager regManager = driver.createRegistrationManager();
```
eolivelli, line 1235: do we have to close this ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/MetadataDrivers.java
```diff
@@ -369,7 +369,8 @@ public static MetadataBookieDriver getBookieDriver(URI uri) {
     public static <T> T runFunctionWithRegistrationManager(ServerConfiguration conf,
                                                            Function<RegistrationManager, T> function)
             throws MetadataException, ExecutionException {
-        return runFunctionWithMetadataBookieDriver(conf, driver -> function.apply(driver.getRegistrationManager()));
+        return runFunctionWithMetadataBookieDriver(
+                conf, driver -> function.apply(driver.createRegistrationManager()));
```
eolivelli, line 373: will the instance returned by  createRegistrationManager be closed ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/PortManager.java
```diff
@@ -54,21 +52,14 @@ public static void initPort(int initPort) {
     public static synchronized int nextFreePort() {
         int exceptionCount = 0;
         while (true) {
-            int port = nextPort++;
```
eolivelli, line 57: should we drop "nextPort" ?

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieInitializationTest.java
```diff
@@ -513,7 +511,9 @@ public void testBookieRegistrationBookieServiceInfo() throws Exception {
                 + ":" + conf.getBookiePort());
         String bkRegPath = ZKMetadataDriverBase.resolveZkLedgersRootPath(conf) + "/" + AVAILABLE_NODE + "/" + bookieId;
 
-        driver.initialize(conf, () -> {}, NullStatsLogger.INSTANCE);
+        MetadataBookieDriver metadataDriver = BookieResources.createMetadataDriver(
```
eolivelli, line 514: do we need to close this ?

eolivelli (changes requested):
Great work

I left some feedback, PTAL

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLoggerAllocator.java
Vanlightly: Yeah, that was causing a test to fail so I fixed it.

bookkeeper-common-allocator/src/main/java/org/apache/bookkeeper/common/allocator/ByteBufAllocatorBuilder.java
Vanlightly: I don't know what you are referring to regarding Pulsar. 

But regarding a separate PR, this is related to this construction/injection PR as we allow the OOM handling to be set in a different place to the construction of the allocator. Basically it gets constructed, then later the BookieService sets the OOM handler so that it integrates with the AbstractLifecycleComponent uncaughtExceptionHandler which will trigger the bookie to shutdown gracefully on an OOM.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperAdmin.java
Vanlightly: Yes good catch. I have done an audit of all calls to createRegistrationManager() and ensured that the lifetime of the registration manager is correctly handled.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/MetadataDrivers.java
Vanlightly: Fixed.

bookkeeper-common-allocator/src/main/java/org/apache/bookkeeper/common/allocator/ByteBufAllocatorBuilder.java
eolivelli: ok, let's keep this.

for Pulsar, IIRC we recently started to use BK ByteBufAllocator to deal with OutOfMemory cases for Direct Memory
and we are using this library

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieInitializationTest.java
Vanlightly: Fixed.

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/PortManager.java
Vanlightly: Yes I think so. I've added a commit to remove it and all the initialPort arguments that would end up setting it.

Vanlightly:
@eolivelli The StandaloneTest failure seems to be a problem unrelated to this patch. For one the docker image name is wrong but when corrected, two tests still fail. The table service commands run by these two tests (createNamespace and createTable) do not terminate. I get the same if I run these commands on a docker container myself.

I presume the StandaloneTest was not being executed by Maven?

Vanlightly:
@eolivelli @merlimat @ivankelly all tests are passing now. PTAL.

ivankelly (approved):
lgtm

eolivelli (approved):
LGTM

eolivelli:
Merged

Issue #2770, opened 2021-08-23
In class SortedLedgerStorage's method ledgerExists, why don't we simplely return false


Shoothzj:
Class SortedLedgerStorage, method ledgerExists
```java
@Override
    public boolean ledgerExists(long ledgerId) throws IOException {
        // Done this way because checking the skip list is an O(logN) operation compared to
        // the O(1) for the ledgerCache.
        if (!interleavedLedgerStorage.ledgerExists(ledgerId)) {
            EntryKeyValue kv = memTable.getLastEntry(ledgerId);
            if (null == kv) {
                return interleavedLedgerStorage.ledgerExists(ledgerId);
            }
        }
        return true;
    }
```
I want to know why calculate ledgerExists() two times? Thanks

PR #2903, opened 2021-11-20
ISSUE-2770: Code refactor calling ledgerExists two times.

jadireddi:
### Changes
SortedLedgerStorage's method `ledgerExists`, is being called two times to check for the ledger exists or not. Refactored to call only once.

Master Issue: https://github.com/apache/bookkeeper/issues/2770

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SortedLedgerStorage.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SortedLedgerStorage.java
@@ -144,9 +144,7 @@ public class SortedLedgerStorage
         // the O(1) for the ledgerCache.
         if (!interleavedLedgerStorage.ledgerExists(ledgerId)) {
             EntryKeyValue kv = memTable.getLastEntry(ledgerId);
-            if (null == kv) {
-                return interleavedLedgerStorage.ledgerExists(ledgerId);
-            }
+            return null != kv;
         }
         return true;
     }


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SortedLedgerStorage.java
```diff
@@ -144,9 +144,7 @@ public boolean ledgerExists(long ledgerId) throws IOException {
         // the O(1) for the ledgerCache.
         if (!interleavedLedgerStorage.ledgerExists(ledgerId)) {
             EntryKeyValue kv = memTable.getLastEntry(ledgerId);
-            if (null == kv) {
-                return interleavedLedgerStorage.ledgerExists(ledgerId);
```
eolivelli, line 148: I have seen this patch before.
Maybe there is already another open PR

My question is...
Why this code was written the way it is?
In order to deal with some possible race condition ?

@merlimat do you know?

eolivelli (approved):
Lgtm

I would like to double check with @merlimat and @dlg99

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SortedLedgerStorage.java
hangc0276: Is there any case that line 145 returns `false`, but line 148 returns `true`?

zymap, line 148: As the comment said, checking the skip list is an O(logN) operation compared to the O(1) for the ledgerCache.

Which means the `memTable.getLastEntry(ledgerId)` is time consuming operation. After check that, the ledgerCache may be filled so we check it again.

zymap:
Remove it to the next release because I think it needs more people to confirm.

StevenLuMT:

Issue #2896, opened 2021-11-14
Adjust checkpoint timing adaptively according to throughput


1559924775:
**FEATURE REQUEST**

1. Please describe the feature you are requesting.

When using the singleentrylog mode, depending on the throughput, for example, under the throughput of 200m / s, The default value of the configuration parameter logsizelimit is 1.2g. At this time, due to the large throughput, the generation of checkpoint tasks is fast, but the execution time of checkpoint tasks is long. The speed of the two is inconsistent. A large number of tasks accumulate in the checkpoint thread, resulting in the checkpoint bits recorded in the lastmark file greatly lagging behind the real-time data. Restarting bookie will be abnormally slow. Unnecessary checkpoint tasks can be eliminated according to the execution time of checkpoint tasks. The next effective checkpoint task is generated only after one checkpoint task is executed. The specific way is to clear the accumulated tasks in the checkpoint thread pool every time a newly generated checkpoint task is added.

2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have).
   must-have

3. Provide any additional detail on your proposed use case for this feature.

PR #2911, opened 2021-11-26
Adjust checkpoint timing adaptively according to throughput

1559924775:

### Motivation

When using the singleentrylog mode, depending on the throughput, for example, under the throughput of 200m / s, The default value of the configuration parameter logsizelimit is 1.2g. At this time, due to the large throughput, the generation of checkpoint tasks is fast, but the execution time of checkpoint tasks is long. The speed of the two is inconsistent. A large number of tasks accumulate in the checkpoint thread, resulting in the checkpoint bits recorded in the lastmark file greatly lagging behind the real-time data. Restarting bookie will be abnormally slow. Unnecessary checkpoint tasks can be eliminated according to the execution time of checkpoint tasks. The next effective checkpoint task is generated only after one checkpoint task is executed. 

### Changes
The specific way is to clear the accumulated tasks in the checkpoint thread pool every time a newly generated checkpoint task is added. In this way, no matter how large the throughput is, the time to generate the checkpoint task and the time to execute the checkpoint task can match.

Master Issue: #2896

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SyncThread.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SyncThread.java
@@ -83,6 +83,20 @@ class SyncThread implements Checkpointer {
     }
 
     protected void doCheckpoint(Checkpoint checkpoint) {
+	// Clear stacked checkpoint tasks
+        try {
+            Field filed = executor.getClass().getDeclaredField("e");
+            filed.setAccessible(true);
+            ScheduledThreadPoolExecutor scheduledThreadPoolExecutor = (ScheduledThreadPoolExecutor) filed.get(executor);
+            BlockingQueue<Runnable> workQueue = scheduledThreadPoolExecutor.getQueue();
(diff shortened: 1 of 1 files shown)


Vanlightly:
@1559924775 thanks for spotting this. However I think the proposed solution isn't quite right. I think there should be a cleaner solution without resorting to reflection.

For example, one solution might be:
1. In doCheckpoint, don't submit it to the executor. Run it synchronously within the runnable.
2. Change the `scheduleAtFixedRate` to `schedule`. Inside doCheckpoint, submit the work again using `schedule`, calculating how long the delay should be based on how long it took the checkpoint to run.

There is also `scheduledWithFixedDelay` which would be simpler, though would always put the same delay between executions, even if a checkpoint takes a very long time.

StevenLuMT (changes requested):
can you add a testcase to cover this updating? @1559924775

StevenLuMT:
Inactive authors, can be removed from 4.16.0 first,I think @hangc0276 @zymap

StevenLuMT:

PR #2913, opened 2021-11-26
Make sure the LedgerHandle close callback can be completed when encounter exception

codelipenghui:
Related to https://github.com/apache/pulsar/pull/12993

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
@@ -573,8 +573,9 @@ public class LedgerHandle implements WriteHandle {
                 // running under any bk locks.
                 try {
                     errorOutPendingAdds(rc, pendingAdds);
-                } catch (Exception e) {
+                } catch (Throwable e) {
                     closePromise.completeExceptionally(e);
+                    return;
                 }
 
                 if (prevHandleState != HandleState.CLOSED) {
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
@@ -571,7 +571,11 @@ public class LedgerHandle implements WriteHandle {
 
                 // error out all pending adds during closing, the callbacks shouldn't be
                 // running under any bk locks.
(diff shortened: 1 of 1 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -571,7 +571,11 @@ public void safeRun() {
 
                 // error out all pending adds during closing, the callbacks shouldn't be
                 // running under any bk locks.
-                errorOutPendingAdds(rc, pendingAdds);
+                try {
+                    errorOutPendingAdds(rc, pendingAdds);
+                } catch (Exception e) {
+                    closePromise.completeExceptionally(e);
```
zymap, line 577: If there has an exception, do we need to do the `tearDownWriteHandleState()` and ` MetadataUpdateLoop` operations?

hangc0276: Yes, it still needs the following operations. Catch this exception just to ensure the closePromise can complete.

hangc0276 (approved):

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
eolivelli: we shouldn't "completeExceptionally" or "complete" the `closePromise` twice

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -571,7 +571,11 @@ public void safeRun() {
 
                 // error out all pending adds during closing, the callbacks shouldn't be
                 // running under any bk locks.
-                errorOutPendingAdds(rc, pendingAdds);
+                try {
+                    errorOutPendingAdds(rc, pendingAdds);
+                } catch (Exception e) {
```
eolivelli, line 576: what about catching "Throwable" ?

eolivelli (changes requested):

codelipenghui:
@eolivelli Fixed, please check again.

StevenLuMT (approved):
lgtm

dlg99 (approved):

eolivelli (approved):
Lgtm

PR #2931, opened 2021-12-07
Add ensemble relocation command which adheres to placement policy

equanz:
### Motivation

Currently, we can't recover the ensemble which isn't adhering to placement policy directly.
The `recover` command or autorecovery process is one of the alternative approaches. Unfortunately, however, we can't.

For example, suppose
* the bookie cluster has two racks( `/region/rack1` , `/region/rack2` )
* target ledger has the ensemble ( `[bookie1(/region/rack1), bookie2(/region/rack1)]` )
* the replicator uses RackawareEnsemblePlacementPolicy

When `bookie1` or `bookie2` goes down, then run autorecovery process.
If the bookie cluster has remained bookies which place to `/region/rack1`, the ledger recovers to the `/region/rack1` bookie again.
https://github.com/apache/bookkeeper/blob/release-4.14.4/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/ReplicationWorker.java#L225-L246
https://github.com/apache/bookkeeper/blob/release-4.14.4/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperAdmin.java#L1055-L1103
https://github.com/apache/bookkeeper/blob/release-4.14.4/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java#L474-L481

Therefore, we can't recover the ensemble which isn't adhering to placement policy directly.
The `recover` command is similar to this behavior.

I want to add a relocation feature to adhere placement policy.

### Changes

Add relocation feature to shell command.
So, we can use it synchronously instead of executing asynchronously like the autorecovery process.

In this specification, optimize the new ensemble arrangement to minify the number of bookies replaced.
Therefore, we should implement the new method that returns the new ensemble by policies.
First, I introduce this feature to RackawareEnsemblePlacementPolicy in this PR.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicy.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicy.java
@@ -237,6 +237,28 @@ public class RackawareEnsemblePlacementPolicy extends RackawareEnsemblePlacement
     }
 
     @Override
+    public PlacementResult<List<BookieId>> replaceToAdherePlacementPolicy(
+            int ensembleSize,
+            int writeQuorumSize,
+            int ackQuorumSize,
+            Set<BookieId> excludeBookies,
+            List<BookieId> currentEnsemble) {
+        final PlacementResult<List<BookieId>> placementResult =
+                super.replaceToAdherePlacementPolicy(ensembleSize, writeQuorumSize, ackQuorumSize,
+                        excludeBookies, currentEnsemble);
+        if (placementResult.isAdheringToPolicy() != PlacementPolicyAdherence.FAIL) {
+            return placementResult;
+        } else {
+            if (slave == null) {
+                return placementResult;
+            } else {
+                return slave.replaceToAdherePlacementPolicy(ensembleSize, writeQuorumSize, ackQuorumSize,
+                        excludeBookies, currentEnsemble);
+            }
+        }
+    }
+
+    @Override
     public void handleBookiesThatLeft(Set<BookieId> leftBookies) {
         super.handleBookiesThatLeft(leftBookies);
         if (null != slave) {
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
@@ -1131,6 +1131,13 @@ public class RackawareEnsemblePlacementPolicyImpl extends TopologyAwareEnsembleP
                     prevNode = replaceToAdherePlacementPolicyInternal(
                             curRack, excludeNodes, ensemble, ensemble,
                             provisionalEnsembleNodes, i, ensembleSize, minNumRacksPerWriteQuorumForThisEnsemble);
+                    // got a good candidate
+                    if (ensemble.addNode(prevNode)) {
+                        // add the candidate to exclude set
+                        excludeNodes.add(prevNode);
+                    } else {
+                        throw new BKNotEnoughBookiesException();
+                    }
                     // replace to newer node
                     provisionalEnsembleNodes.set(i, prevNode);
                 } catch (BKNotEnoughBookiesException e) {
@@ -1159,10 +1166,6 @@ public class RackawareEnsemblePlacementPolicyImpl extends TopologyAwareEnsembleP
         final BookieNode currentNode = provisionalEnsembleNodes.get(ensembleIndex);
         // if the current bookie could be applied to the ensemble, apply it to minify the number of bookies replaced
         if (!excludeBookies.contains(currentNode) && predicate.apply(currentNode, ensemble)) {
-            if (ensemble.addNode(currentNode)) {
-                // add the candidate to exclude set
-                excludeBookies.add(currentNode);
-            }
             return currentNode;
         }
 
@@ -1234,11 +1237,6 @@ public class RackawareEnsemblePlacementPolicyImpl extends TopologyAwareEnsembleP
                     continue;
                 }
                 BookieNode bn = (BookieNode) n;
-                // got a good candidate
-                if (ensemble.addNode(bn)) {
-                    // add the candidate to exclude set
-                    excludeBookies.add(bn);
-                }
                 return bn;
             }
         }
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
@@ -46,6 +46,8 @@ import java.util.Set;
 import java.util.concurrent.TimeUnit;
 
 import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import org.apache.bookkeeper.client.BKException.BKNotEnoughBookiesException;
 import org.apache.bookkeeper.client.BookieInfoReader.BookieInfo;
 import org.apache.bookkeeper.client.WeightedRandomSelection.WeightedObject;
@@ -69,6 +71,7 @@ import org.apache.bookkeeper.stats.OpStatsLogger;
 import org.apache.bookkeeper.stats.StatsLogger;
 import org.apache.bookkeeper.stats.annotations.StatsDoc;
 import org.apache.commons.collections4.CollectionUtils;
+import org.apache.commons.lang3.tuple.Pair;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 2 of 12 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
```diff
@@ -1071,4 +1074,175 @@ public boolean areAckedBookiesAdheringToPlacementPolicy(Set<BookieId> ackedBooki
...
+    private BookieNode replaceToAdherePlacementPolicyInternal(
+            String netPath, Set<Node> excludeBookies, Predicate<BookieNode> predicate,
+            Ensemble<BookieNode> ensemble, List<BookieNode> provisionalEnsembleNodes, int ensembleIndex,
+            int ensembleSize, int minNumRacksPerWriteQuorumForThisEnsemble) throws BKNotEnoughBookiesException {
+        final BookieNode currentNode = provisionalEnsembleNodes.get(ensembleIndex);
+        // if the current bookie could be applied to the ensemble, apply it to minify the number of bookies replaced
+        if (!excludeBookies.contains(currentNode) && predicate.apply(currentNode, ensemble)) {
+            if (ensemble.addNode(currentNode)) {
+                // add the candidate to exclude set
+                excludeBookies.add(currentNode);
```
eolivelli, line 1164: we should not modify the `excludeBookies` Set passed as argument, please create a copy in the beginning of the method
otherwise this method will have unwanted side effects

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/CorrectEnsemblePlacementCmdTest.java
```diff
@@ -0,0 +1,235 @@
...
+        writableBookies.add(addr3.toBookieId());
+        writableBookies.add(addr4.toBookieId());
+        writableBookies.add(addr5.toBookieId());
+        writableBookies.add(addr6.toBookieId());
+        writableBookies.add(addr7.toBookieId());
+        writableBookies.add(addr8.toBookieId());
+        writableBookies.add(addr9.toBookieId());
+
+        // add bookie node to resolver
+        StaticDNSResolver.reset();
```
eolivelli, line 121: we should do this also in a "After" method, in order to ensure that we do not pollute the test environment

eolivelli (changes requested):
This is a great work!

I did a first pass, it is a big patch.
So we'll need more eyes

@Ghatage @dlg99 @merlimat @ivankelly

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
equanz: modify the `excludeBookies` to notify exclude bookie to called method like below.
https://github.com/apache/bookkeeper/blob/release-4.14.4/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java#L671-L675

However, as you said, it might have unwanted side effects. Therefore, I'll modify to put to excludeBookies at called method.

```
diff --git a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
index d9167c321..fa3e59cd6 100644
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
@@ -1131,6 +1131,13 @@ public class RackawareEnsemblePlacementPolicyImpl extends TopologyAwareEnsembleP
                     prevNode = replaceToAdherePlacementPolicyInternal(
                             curRack, excludeNodes, ensemble, ensemble,
                             provisionalEnsembleNodes, i, ensembleSize, minNumRacksPerWriteQuorumForThisEnsemble);
+                    // got a good candidate
+                    if (ensemble.addNode(prevNode)) {
+                        // add the candidate to exclude set
+                        excludeNodes.add(prevNode);
+                    } else {
+                        throw new BKNotEnoughBookiesException();
+                    }
                     // replace to newer node
                     provisionalEnsembleNodes.set(i, prevNode);
                 } catch (BKNotEnoughBookiesException e) {
@@ -1159,10 +1166,6 @@ public class RackawareEnsemblePlacementPolicyImpl extends TopologyAwareEnsembleP
         final BookieNode currentNode = provisionalEnsembleNodes.get(ensembleIndex);
         // if the current bookie could be applied to the ensemble, apply it to minify the number of bookies replaced
         if (!excludeBookies.contains(currentNode) && predicate.apply(currentNode, ensemble)) {
-            if (ensemble.addNode(currentNode)) {
-                // add the candidate to exclude set
-                excludeBookies.add(currentNode);
-            }
             return currentNode;
         }

@@ -1234,11 +1237,6 @@ public class RackawareEnsemblePlacementPolicyImpl extends TopologyAwareEnsembleP
                     continue;
                 }
                 BookieNode bn = (BookieNode) n;
-                // got a good candidate
-                if (ensemble.addNode(bn)) {
-                    // add the candidate to exclude set
-                    excludeBookies.add(bn);
-                }
                 return bn;
             }
         }
```

equanz:
rerun failure checks

equanz:
I've created the issue about the failing `OWASP Dependency Check`.
https://github.com/apache/bookkeeper/issues/3134

equanz:
rerun failure checks

equanz:
rerun failure checks

equanz:
rerun failure checks

equanz:
rerun failure checks

equanz:
@eolivelli Addressed your first comments. PTAL.
(Also rebase to current master to follow https://github.com/apache/bookkeeper/pull/3140 and https://github.com/apache/bookkeeper/pull/3149 .)

bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/commands/bookies/CorrectEnsemblePlacementCommand.java
```diff
@@ -0,0 +1,321 @@
...
+
+        @Cleanup
+        final MetadataBookieDriver metadataDriver = instantiateMetadataDriver(conf);
+        @Cleanup
+        final LedgerManagerFactory lmf = metadataDriver.getLedgerManagerFactory();
+        @Cleanup
+        final LedgerUnderreplicationManager lum = lmf.newLedgerUnderreplicationManager();
+
+        final List<Long> targetLedgers =
+                flags.ledgerIds.stream().parallel().distinct().filter(ledgerId -> {
```
eolivelli, line 156: What's the reason for 'parallel'?
This loop is doing metadata operations, you can kill the zookeeper if you don't throttle properly the requests.

bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/commands/bookies/CorrectEnsemblePlacementCommand.java
```diff
@@ -0,0 +1,321 @@
...
+        if (targetLedgers.isEmpty()) {
+            LOG.info("None of ledgers are relocated.");
+            return true;
+        }
+
+        final NavigableSet<Long> unAcquirableLedgers = new TreeSet<>();
+        final NavigableSet<Pair<Long, Long>> failedTargets = new ConcurrentSkipListSet<>();
+        final CountDownLatch latch = new CountDownLatch(targetLedgers.size());
+        for (long ledgerId : targetLedgers) {
+            if (!flags.dryRun) {
```
eolivelli, line 175: What about logging the ledger id here?

bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/commands/bookies/CorrectEnsemblePlacementCommand.java
```diff
@@ -0,0 +1,321 @@
...
+                try {
+                    lum.acquireUnderreplicatedLedger(ledgerId);
+                } catch (ReplicationException e) {
+                    LOG.warn("Failed to acquire ledger to under replicated {}.", ledgerId);
+                    unAcquirableLedgers.add(ledgerId);
+                    latch.countDown();
+                    continue;
+                }
+            }
+            admin.asyncOpenLedger(ledgerId, (rc, lh, ctx) -> {
```
eolivelli, line 185: We should move all this block into some BookKeeperAdmin method.
It will be easier to write tests and also to use this feature programmatically

eolivelli (changes requested):

bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/commands/bookies/CorrectEnsemblePlacementCommand.java
equanz: > What's the reason for 'parallel'?

It was no strong reason.

> This loop is doing metadata operations, you can kill the zookeeper if you don't throttle properly the requests.

I missed that point. I'll remove it.

equanz:
rerun failure checks

equanz:
rerun failure checks

equanz:
@eolivelli Addressed your comments. PTAL.

equanz:
@eolivelli

dlg99:
I have rather generic comments.

> If the bookie cluster has remained bookies which place to /region/rack1, the ledger recovers to the /region/rack1 bookie again.

As I understand, there is enforceMinNumRacksPerWriteQuorum option to prevent that (and enforceMinNumFaultDomainsForWrite, enforceStrictZoneawarePlacement for other placement policies)

As I understand, Auditor's placementPolicyCheck just detects the problem. Maybe it makes sense to make Auditor (optionally) put the ledgers with bad placement for re-replication and make AutoRecovery handle that?
In this case the CLI command is already existing triggeraudit.

Another note is that the test only covers rack-aware policy. What happens in case of the region-aware or zone-aware policies?

dlg99:
Also: https://github.com/apache/bookkeeper/pull/3359

equanz:
> Also: https://github.com/apache/bookkeeper/pull/3359

Thank you for sharing. I'll check it first.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicy.java
```diff
@@ -236,6 +236,28 @@ public BookieNode selectFromNetworkLocation(
         }
     }
 
+    @Override
+    public PlacementResult<List<BookieId>> replaceToAdherePlacementPolicy(
```
horizonzy, line 240: In some case, the replace policy will replace more bookie. There are two racks.(rack1, rack2). 
rack1: bookie1,bookie2,bookie3,bookie4
rack2: bookie5,bookie6,bookie7,bookie8
`E:6 WQ:2 AQ:2`. The ensemble is (5,6,1,7,2,8). (rack2,rack2,rack1,rack2,rack1,rack2), only replace the first  bookie5 to bookie3, it adhere the placement policy. But now the implements didn't replace the firstly, it will replace from second element(bookie6), it will replace more bookie.

equanz: You're right. We can't "minimize" replacement bookies in some cases.

horizonzy (commented):
I notice that the implements didn’t handle the bookie which already shutdown. The shutdown bookie will be as the defaultRack, can we replace it?

equanz:
@dlg99 https://github.com/apache/bookkeeper/pull/2931#issuecomment-1190559164

Sorry for the late reply. I've checked https://github.com/apache/bookkeeper/pull/3359#pullrequestreview-1061219674 .
The approach (using autorecovery) sounds better. If it was merged and still has the above issue, I'll rebase to follow https://github.com/apache/bookkeeper/pull/3359 interfaces.

> As I understand, there is enforceMinNumRacksPerWriteQuorum option to prevent that (and enforceMinNumFaultDomainsForWrite, enforceStrictZoneawarePlacement for other placement policies)

In my understanding, if we use `enforceMinNumRacksPerWriteQuorum` , we can't recover quorum temporally for redundancy when a rack goes down. I want to avoid it.

> As I understand, Auditor's placementPolicyCheck just detects the problem. Maybe it makes sense to make Auditor (optionally) put the ledgers with bad placement for re-replication and make AutoRecovery handle that?
In this case the CLI command is already existing triggeraudit.

I think so too.
I thought the next step is asynchronous relocation, like the auto-recovery feature (mentioned in [here](https://lists.apache.org/thread/vqwmff7rmsn54fxhcojnm2zqvz3s80lj)). However, it was implemented in https://github.com/apache/bookkeeper/pull/3359 .

> Another note is that the test only covers rack-aware policy. What happens in the case of the region-aware or zone-aware policies?

When using non-Rackaware, it throws an Exception. I'll add the test.
https://github.com/equanz/bookkeeper/blob/62970b3122fa1f8266763f0d3f6115a066fcabe9/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/EnsemblePlacementPolicy.java#L460-L467

horizonzy:
I want to use the EnsemblePlacementPolicy#replaceToAdherePlacementPolicy in #3359, it's more graceful, and I will enhance it in some case.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicy.java
equanz: > Maybe we can list the kinds of replace result, and pick the "minimize" result.

Yeah.
I think, alternatively, if the minimization is too costly for calculation, reduce as much as possible in a realistic amount of time.

equanz:
@horizonzy 

> I notice that the implements didn’t handle the bookie which already shutdown. The shutdown bookie will be as the defaultRack, can we replace it?

Currently, we can't do it. This feature expects all of the bookies were already recovered to a normal state.
However, your point is correct.

> I want to use the EnsemblePlacementPolicy#replaceToAdherePlacementPolicy in https://github.com/apache/bookkeeper/pull/3359, it's more graceful, and I will enhance it in some cases.

It's okay.
Just confirm, do you mean "bring this feature to https://github.com/apache/bookkeeper/pull/3359 and improve it"?

horizonzy:
Improve it both #3359 and this pr. I think the improvement is common.

equanz:
>> I notice that the implements didn’t handle the bookie which already shutdown. The shutdown bookie will be as the defaultRack, can we replace it?
> 
> Currently, maybe, we can't do it. This feature expects all of the bookies were already recovered to a normal state.
> However, your point is correct.

I've checked the current code and found handing. If the `enforceMinNumRacksPerWriteQuorum` is true then handle in below.
https://github.com/equanz/bookkeeper/blob/62970b3122fa1f8266763f0d3f6115a066fcabe9/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java#L1089-L1090

However, it is not expected of me. I'll modify to add to excludeNodes regardless of `enforceMinNumRacksPerWriteQuorum` value.

Moreover, `EnsemblePlacementPolicy#isEnsembleAdheringToPlacementPolicy` has a similar process like below.
https://github.com/equanz/bookkeeper/blob/62970b3122fa1f8266763f0d3f6115a066fcabe9/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java#L1043

horizonzy:
>> I notice that the implements didn’t handle the bookie which already shutdown. The shutdown bookie will be as the defaultRack, can we replace it?

> If the default rack name is /default-rack, then we can't add it.
https://github.com/equanz/bookkeeper/blob/62970b3122fa1f8266763f0d3f6115a066fcabe9/bookkeeper-server/src/main/java/org/apache/bookkeeper/net/NetworkTopologyImpl.java#L416-L420

I means the `defaultRack` is not literal, I means the `defaultRack` is `/default-region/default-rack`. 😂

horizonzy:
> However, it is not expected of me. I'll modify to add to excludeNodes regardless of enforceMinNumRacksPerWriteQuorum value.

Yes, If we already knows which bookie  shutdown, I can add it to excludeNodes to ignore it.

equanz:
In my understanding, currently, defaultRack of Rackaware is `/default-rack`. So we should consider about it if use it.
https://github.com/equanz/bookkeeper/blob/62970b3122fa1f8266763f0d3f6115a066fcabe9/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java#L138
https://github.com/equanz/bookkeeper/blob/62970b3122fa1f8266763f0d3f6115a066fcabe9/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java#L234-L248

>> However, it is not expected of me. I'll modify to add to excludeNodes regardless of enforceMinNumRacksPerWriteQuorum value.
> 
> Yes, If we already knows which bookie shutdown, I can add it to excludeNodes to ignore it.

Could you tell me more about shutdown? I think resolver doesn't return `/default-region/default-rack` just call `TopologyAwareEnsemblePlacementPolicy#handleBookiesThatJoined`.

horizonzy:
> In my understanding, currently, defaultRack of Rackaware is `/default-rack`. So we should consider about it if use it.

Yes, you are right.  If the test case, we changed the default rack to `/defualt-region/default-rack`.
```
repp.withDefaultRack(NetworkTopology.DEFAULT_REGION_AND_RACK);
```

If we didn't change it, the default rack is `/default-rack`.

StevenLuMT:

StevenLuMT:

equanz:
Part of the code are ported to https://github.com/apache/bookkeeper/pull/3359 and it was merged.
So, I'll close this PR.

PR #2934, opened 2021-12-10
Upgrade to Grpc 1.42.1

merlimat:
### Motivation

`BKRegistrationNameResolverProvider` is relying on an interface from older Grpc version that has changed since then. 

This causes conflicts when trying to use BK with a newer version of Ggrpc, which can be the case when other components are requesting a newer Grpc version.


zymap (approved):

hangc0276 (approved):

nicoloboschi (approved):

eolivelli (approved):
LGTM

Issue #2930, opened 2021-12-06
Enable backward-compatibility tests


nicoloboschi:
Currently, building BookKeeper with Maven doesn't trigger the tests under `tests/backward-compat`. (neither with `-PintegrationTests` and `-DintegrationTests` 
With Gradle these tests run in the test phase but they fail due to timeout while connecting to ZK container.

We need to
1) make sure the tests are run during build (both Maven and Gradle)
2) fix all the tests

I will start a discussion on dev@ for moving to TestContainers and remove the Arquillian Cube framework from the codebase.

This will lead to:
1. Add test-containers setup; starts ZK and _n_ bookie containers (upgrading ZK container to the current used version by BK)
2. Remove Arquillian Cube dependency and setup files (arquillian.xml)
3. Put together all these tests in a single module to increase readability - code reusability - decrease verbosity

hsaputra:
FYI @Vanlightly , @pkumar-singh , @ivankelly

pkumar-singh:
@nicoloboschi Very good idea. Giving some pointers. We are already using gradle to run integration tests with test container. Kindly see If you can lift something from there and reuse.

PR #2937, opened 2021-12-13
Gradle - Enable backward-compatibility tests

nicoloboschi:
### Motivation

Integration tests under `backward-compat` do not work neither with Gradle and Maven. 
For Maven, they are not triggered at all
For Gradle, they run and fail because in Gradle the docker images used inside the tests are not built at all. 

### Changes

* Build docker images under `tests/docker-images`
* Updated Dockerfile files to work with Gradle 
* Fix some tests failing due to BookieId migration
* Fix `tests/backward-compat/upgrade/src/test/groovy/org/apache/bookkeeper/tests/backwardcompat/TestCompatUpgrade.groovy` test which needs tests running in a specific order
* Fix `MavenClassLoader`removing the need to have the current SNAPSHOT version jars in the local repository. Now it uses the distribution tar.gz
* Upgraded ZK in these tests to 3.6.2
* Add these tests in GH actions
* Split a couple tests in different classes in order to not reuse the same containers environment; this helps reduce flakyness

Some important notes:
* I haven't found the date where tests were not being run anymore but I think is more than 1 year ago, at least since BP-41
* I kept Arquillian Cube as docker container runner since there's no real need to migrate to testcontainers
* **These tests only work on Linux** because they rely on `Host` docker network which is only available on Linux 
* I haven't fixed Maven execution since we are already using Gradle in CI  

Master Issue: #2930


tests/docker-images/statestore-image/image_builder.sh
```diff
@@ -18,22 +18,34 @@
 # * limitations under the License.
 # */
 
+IMAGE_NAME=apachebookkeeper/bookkeeper-current:latest
```
pkumar-singh, line 21: One problem I see with this optimization is that, when debugging or locally running test, I would not know for sometime why changes I made is not taking affect, until I see that image_builder is not building new image anymore.

tests/docker-images/statestore-image/image_builder.sh
```diff
@@ -18,22 +18,34 @@
...
+if [[ "$(docker images -q $IMAGE_NAME 2> /dev/null)" != "" ]]; then
+  echo "reusing local image: $IMAGE_NAME"
+  exit 0
```
pkumar-singh, line 24: One problem I see with this optimization is that, when debugging or locally running test, I would not know for sometime why changes I made is not taking affect, until I see that image_builder is not building new image anymore.

tests/backward-compat/current-server-old-clients/src/test/groovy/org/apache/bookkeeper/tests/backwardcompat/AbstractClientFencingTest.groovy
```diff
@@ -0,0 +1,97 @@
...
+            ledger0.close()
+
+
+            def ledger1 = fencingBK.openLedger(ledger0.getId(), fencingCL.digestType("CRC32"), PASSWD)
+
+            // cannot write any more
+            try {
+                ledger0.addEntry("shouldn't work".getBytes())
+                Assert.fail("Shouldn't have been able to add any more")
+            } catch (Exception e) {
```
pkumar-singh, line 71: Why not just catch BKLedgerClosedException

.github/workflows/backward-compat-tests.yml
```diff
@@ -0,0 +1,58 @@
+#
```
pkumar-singh, line 1: Kindly also see to it that these tests are not running in the catch all tests.

pkumar-singh:
LGTM. Provided couple of observations.

tests/backward-compat/current-server-old-clients/src/test/groovy/org/apache/bookkeeper/tests/backwardcompat/AbstractClientFencingTest.groovy
nicoloboschi: because we don't have it in the classpath at compile time

tests/docker-images/statestore-image/image_builder.sh
nicoloboschi: I added a env property to force the rebuilding of the images. By default is false since it's uncommon you want to refresh those images

.github/workflows/backward-compat-tests.yml
nicoloboschi: good observation. I checked it and they run only in this job, thanks

nicoloboschi:
rerun failure checks

nicoloboschi:
rerun failure checks

eolivelli (approved):
looks great !

thank you 

@pkumar-singh can we commit ? you wrote LGTM but you didn't "approve" the patch

pkumar-singh (approved):

pkumar-singh:
@eolivelli Approved for what ever it's worth. Please feel free to commit.

eolivelli:
Prashant, IIUC you are a committer now, so your Approval is binding to commit the patch.
I wonder why you still haven't write permissions to the repository.

eolivelli:
Great work @nicoloboschi 
committed to master branch !

PR #2939, opened 2021-12-14
Fix check compatibility workflows failing due to regex

Vanlightly:
### Motivation

The java 8 and 11 compatibility workflows are failing because the regex in TestPrometheusFormatter.java does not work correctly, in different ways, for Java 8 and 11 on Ubuntu.

I can reproduce this on Ubuntu 18.04.

### Changes

Changed the regex used for parsing the metrics lines so that it produces the right matches.


nicoloboschi (approved):
LGTM

Vanlightly:
Looks like #2934 fixed this just now. Closing.

Issue #2941, opened 2021-12-15
[BUG] Bookie will infinite waiting for shutdown due to throttler limit


wenbingshen:
**BUG REPORT**

***Describe the bug***

Our compaction current limiting configuration is as follows, the traffic configuration is very small.
The compact limit configuration is so small because we encountered compact which caused the disk io to be too high.

**bookkeeper.conf**
`isThrottleByBytes=true`
`compactionRateByEntries=1000`
`compactionRateByBytes=150000`
![image](https://user-images.githubusercontent.com/35599757/146189635-da561e67-0173-4e64-a668-3d4282245076.png)

Our bookkeeper cluster is always running stably, but when we need to restart bookie, it took us half an hour to stop a bookie node.
You can observe our log as follows. After log investigation, we found that the same major compaction task in a certain GC thread of Bookie has been running for two days and has not been completed. The remove operation between each entry log is separated by 20 to 30 minutes.
![image](https://user-images.githubusercontent.com/35599757/146188911-c9bbe546-b3bc-4a98-9cf1-282c5a8b0d5d.png)

When we run the command line tool to stop bookie, `component-shutdown-thread` will start to work. But the `component-shutdown-thread` is always blocked in the shutdown logic of the GC thread.
![image](https://user-images.githubusercontent.com/35599757/146190762-afd73921-ad8f-4661-8369-7b24a6150ab6.png)

The `GarbageCollectorThread` thread has been blocked in `RateLimiter.acquire`
![image](https://user-images.githubusercontent.com/35599757/146191195-dee9d53c-b864-4ad1-9698-1d26538f2cd9.png)

We stopped bookie at 14:10, but bookie didn't really stop until 14:42. The last stop time depends on the last time of all GC threads to recover from the current limit.

**`GarbageCollectorThread-15-1`**
Stopped at 14:42:48
![image](https://user-images.githubusercontent.com/35599757/146191864-bac9fdb3-2e55-4c3d-b4bc-f3f8f3f5bd33.png)

**`GarbageCollectorThread-21-1`**
Stopped at 14:38:15
![image](https://user-images.githubusercontent.com/35599757/146192021-02c7f3d3-d77c-4a1a-ac84-4da7bd9ac566.png)

**`GarbageCollectorThread-24-1`**
Stopped at 14:22:04
![image](https://user-images.githubusercontent.com/35599757/146192130-64b96fb0-96ef-45bd-8529-cd7e12c7cdb4.png)

[...]

PR #2942, opened 2021-12-15
Fix the infinite waiting for shutdown due to throttler limit

wenbingshen:
### Motivation

If the compactor is limited, the shutdown priority should be higher than waiting for RateLimiter.acquire.

### Changes

According to @hangc0276 suggestion, when processing the shutdown logic of `GarbageCollectorThread`, we should check the status of the `newScanner.process` method. If the status is false, throw an `IOException` and stop compact immediately.

Master Issue: #2941


wenbingshen:
@eolivelli @merlimat @codelipenghui Can you take a look at this pr. Thanks.

hangc0276:
@wenbingshen Thanks for you enormous contribution.
The root cause of preventing GarbageCollectorThread shutdown is that the `compacting` flag only control one compact entry phase. 
https://github.com/apache/bookkeeper/blob/7385d90189b2aa9ae11418a7c1c0fa60e5d0ea25/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java#L558-L579

So, it will wait for one entry log file compact complete before GarbageCollectorThread shutdown. If we throttle the compaction in low throughput, it will take multi minutes to compaction 2GB entry log file.

IMO, we'd better introduce a new compact flag for `AbstractLogCompactor` and check the status on `newScanner.process` method, if the status is false, just throw an IOException, and stop the compact immediately. It can work for both `EntryLogCompactor` and `TransactionalEntryLogCompactor`.

wenbingshen:
@hangc0276 Gotcha! Thanks for your review. I will implement it soon. :)

wenbingshen:
rerun failure checks

wenbingshen:
rerun failure checks

wenbingshen:
rerun failure checks

wenbingshen:
rerun failure checks

wenbingshen:
@hangc0276 PTAL, thanks. :)

dlg99 (approved):
LGTM

wenbingshen:
ping @eolivelli PTAL, Thanks.

wenbingshen:
@eolivelli PTAL Thanks

wenbingshen:
@pkumar-singh It's been a long time. Can you take a look at this pr? Thanks.

wenbingshen:
@eolivelli It's been over 3 months since this PR was created. Can you help take a look at this PR? Thanks very much!

hangc0276:
@wenbingshen  Would you please rebase the master? thanks.

hangc0276 (approved):
Good job!

hangc0276:
https://github.com/apache/bookkeeper/runs/7519602108?check_suite_focus=true
The CI failed, Please take a look, thanks.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -653,6 +653,8 @@ public synchronized void shutdown() throws InterruptedException {
         }
         LOG.info("Shutting down GarbageCollectorThread");
 
+        throttler.cancelledAcquire();
+        compactor.throttler.cancelledAcquire();
```
wenbingshen, line 657: #2963 introduces another limiter in the GC thread. When the current limit is set very small, it will also cause the GC thread to hang for a long time, so the current limit interrupt logic Move to Throttler.

In this way, where necessary, we can use the `Throttler.cancelledAcquire` method to interrupt the long-term thread sleep caused by current limiting.

wenbingshen:
rerun failure checks

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/AbstractLogCompactor.java
```diff
@@ -66,6 +70,8 @@ public void cleanUpAndRecover() {}
     public static class Throttler {
         private final RateLimiter rateLimiter;
         private final boolean isThrottleByBytes;
+        private final AtomicBoolean cancelled = new AtomicBoolean(false);
+        private final AtomicBoolean throttlerInterrupted = new AtomicBoolean(false);
```
hangc0276, line 74: This one just used for test?

wenbingshen: Thanks for your review.I think `throttlerInterrupted` can be removed. Because we can use `org.apache.bookkeeper.bookie.GarbageCollectorThread#running` in the test to determine whether the GC thread is closed, because if the `throttler` is not interrupted, the `compact` will not end, and the GC thread will always be stuck in `shutdown`.
```java
public synchronized void shutdown() throws InterruptedException {
        if (!this.running) {
            return;
        }
        LOG.info("Shutting down GarbageCollectorThread");

        throttler.cancelledAcquire();
        compactor.throttler.cancelledAcquire();
        while (!compacting.compareAndSet(false, true)) {
            // Wait till the thread stops compacting
            Thread.sleep(100);
        }

        this.running = false;
        // Interrupt GC executor thread
        gcExecutor.shutdownNow();
        try {
            entryLogMetaMap.close();
        } catch (Exception e) {
            LOG.warn("Failed to close entryLog metadata-map", e);
        }
    }
```

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/AbstractLogCompactor.java
```diff
@@ -74,8 +78,32 @@ public static class Throttler {
...
+            long timeout = 100;
+            long start = System.currentTimeMillis();
+            while (!tryAcquire(permits, timeout, TimeUnit.MILLISECONDS)) {
+                if (cancelled.get()) {
+                    throw new IOException("Failed to get permits takes "
+                            + (System.currentTimeMillis() - start)
+                            + " ms may be compactor has been shutting down");
+                }
+                try {
+                    TimeUnit.NANOSECONDS.sleep(timeout);
```
hangc0276, line 98: Why use nano second here?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/AbstractLogCompactor.java
```diff
@@ -74,8 +78,32 @@ public static class Throttler {
...
+            return rateLimiter.tryAcquire(this.isThrottleByBytes ? permits : 1, timeout, unit);
+        }
+
+        // GC thread will check the status for the rate limiter
+        // If the compactor is being stopped by other threads,
+        // and the GC thread is still limited, the compact task will be stopped.
+        public void acquire(int permits) throws IOException {
+            long timeout = 100;
+            long start = System.currentTimeMillis();
+            while (!tryAcquire(permits, timeout, TimeUnit.MILLISECONDS)) {
```
StevenLuMT, line 91: MILLISECONDS in tryAcquire
but TimeUnit.NANOSECONDS.sleep(timeout)?
mistake??? @wenbingshen

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/AbstractLogCompactor.java
wenbingshen: Sorry, I wrote the wrong unit.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/AbstractLogCompactor.java
wenbingshen: yes, it's a mistake. Addressed.

PR #2944, opened 2021-12-16
Fix wrong ledger id parse radix for index relocation file in IndexPersistenceMgr

kezhuw:

### Motivation
1. `IndexPersistenceMgr` fails to start after crashing index file movement.

### Changes
1. Add test to assert index file relocation
2. Fix wrong ledger id parse radix for index relocation file in IndexPersistenceMgr

Master Issue: None

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/IndexPersistenceMgr.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/IndexPersistenceMgr.java
@@ -274,8 +274,9 @@ public class IndexPersistenceMgr {
                                 // name is the HexString representation of the
                                 // ledgerId.
                                 String ledgerIdInHex = index.getName().replace(RLOC, "").replace(IDX, "");
+                                long ledgerId = Long.parseLong(ledgerIdInHex, 16);
                                 if (index.getName().endsWith(RLOC)) {
-                                    if (findIndexFile(Long.parseLong(ledgerIdInHex)) != null) {
+                                    if (findIndexFile(ledgerId) != null) {
                                         if (!index.delete()) {
                                             LOG.warn("Deleting the rloc file " + index + " failed");
                                         }
@@ -288,7 +289,7 @@ public class IndexPersistenceMgr {
                                         }
                                     }
                                 }
-                                activeLedgers.put(Long.parseLong(ledgerIdInHex, 16), true);
+                                activeLedgers.put(ledgerId, true);
                             }
                         }
                     }
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/IndexPersistenceMgrTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/IndexPersistenceMgrTest.java
@@ -22,10 +22,13 @@ package org.apache.bookkeeper.bookie;
 
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.spy;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufUtil;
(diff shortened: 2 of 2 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileInfo.java
```diff
@@ -549,6 +549,7 @@ public synchronized void moveToNewLocation(File newFile, long size) throws IOExc
         }
         fc = new RandomAccessFile(newFile, mode).getChannel();
         lf = newFile;
+        deleted = false;
```
nicoloboschi, line 552: it'd be better to have a method `delete(boolean markAsDeleted)` and passing false in this method

in your new code we still have few lines where the `deleted` flag is equals to `true` and it can lead in undefined behaviour

nicoloboschi (changes requested):

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileInfo.java
kezhuw: Seems that all access methods of `deleted` and `fc` are `synchronized`. So I think it is safe ?

Actually, I could keep `deleted` untouched by using `lf.delete()` directly to delete file instead of `FileInfo.delete` which is used as spying point to mimic crashing behavior in tests. But, this way I have to introduce an extra spying point to intercept. Should I go this way and make spying point more explicitly ?

dlg99 (approved):

kezhuw:
@dlg99 I have fixed CI failures. The CI workflow has passed in my local repository. Could you please trigger a CI workflow run again for this pr ? Thanks.

dlg99:
Close to reopen (and trigger rebase/CI run)

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileInfo.java
dlg99: @nicoloboschi are you ok with this?

dlg99: @nicoloboschi ping

nicoloboschi (approved):

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileInfo.java
kezhuw: > `moveToNewLocation` is not expected the be in deleted state at all during the rename


It is only correct in ideal situation. Could this operation happen cross filesystems or even devices ? I think this method make no assumption about this. From my side, `delete` in `moveToNewLocation` is more like a conservative operation than solely `Files.delete()`(which throw exception on deletion failure). After failure of `FileInfo.delete`, `FileInfo` is unusable as `FileInfo.fc` is closed and `FileInfo.lf` is not guaranteed to exist.

I have no preference from approaches:
* Current: `FileInfo.delete()` and throws exception on failure.
* Alternative: `FileInfo.lf.delete()` and mark `deleted` and throws exception on failure.

I think they are identical.

kezhuw:
rerun failure checks

kezhuw:
@dlg99 @nicoloboschi Could you please take another look ? I have introduced `FileInfo.deleteFileDirectly` as spying point to avoid touching `deleted` in success path.

kezhuw:
@dlg99 @nicoloboschi @eolivelli @merlimat Could you please take a look ? It has been a while since approval and my last  change.

hangc0276 (commented):
@kezhuw  Thanks for your contribution.
From the code change, does it fixed by https://github.com/apache/bookkeeper/pull/3421

kezhuw:
@hangc0276  #3421 solved motivation#1, but not motivation#2. I guess I have to rebase to #3421.

kezhuw:
@hangc0276 I have rebased this pr to origin/master and updated pr description. The context for previous reply has lost but you could get it from `edited` history in pr description form.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileInfo.java
kezhuw: I have reverted related changes since #3421 fixed this.

StevenLuMT (approved):
nice

StevenLuMT:
I think it can be merged,have a look @hangc0276 @dlg99 @eolivelli

StevenLuMT:
rerun failure checks

StevenLuMT:
please rebase form master's new code

hangc0276 (approved):

Issue #2974, opened 2022-01-05
When the journal directory is even, the journal data is unevenly distributed


1559924775:
**BUG REPORT**

***Describe the bug***

When the journal directory is configured to be even, the ledger allocated to each journal will be uneven. The root cause is that almost all the ledger IDs generated by the LedgerIdGenerator class are even. The number of ZK temporary order nodes is determined by the cversion of the parent node. Whenever a child node is added or deleted, it will increase by 1. The LedgerIdGenerator class uses ZK's temporary order node to generate the ledgerid, and deletes the node immediately after generation. This way of use causes almost all cversions to be even, resulting in all ledgerids to be even.

***To Reproduce***

1. Configure in bk_server.conf: jouranlDirectorie=/tmp/bk-txn1,/tmp/bk-txn2
2. production data to bookie

***Expected behavior***

Almost all ledgerids are assigned to directory 1

***Screenshots***

zookeeper:  PrepRequestProcessor::pRequest2TxnCreate:
![image2021-11-8_16-1-33](https://user-images.githubusercontent.com/35036009/141671319-418a2c51-0c2b-4f43-849f-c9fb340f114f.png)

<img width="702" alt="journal" src="https://user-images.githubusercontent.com/35036009/141671860-bd2eaa8b-9b01-41d3-a159-688ceea08554.png">

<img width="503" alt="mathutil" src="https://user-images.githubusercontent.com/35036009/141671870-fca81d57-9d97-49ed-8669-b77cdb3176c5.png">

bugfix:
<img width="942" alt="mathutilnew" src="https://user-images.githubusercontent.com/35036009/141671914-469286b1-b07d-423f-9fef-664286bc2122.png">

Vanlightly:
Changing the routing without some kind of migration will result in effective data loss. Plus an md5 hash is relatively expensive.

The issue here is that ledger ids are all even. I am seeing this myself which results in only even numbered read threads from being routed to. This problem of even ledger ids affects the selection of journals, ledger storage, read threads and write threads. It is a real problem.

@eolivelli This is a pretty serious problem as we're throwing away a bunch of scaling mechanisms.

dlg99:
@Vanlightly @eolivelli I updated OrderedExecutor to better distribute data across the threads, added tests reproducing the issue. please take a look: 

https://github.com/apache/bookkeeper/pull/3023

This does not address the fact that generated ledger ids are skewed towards even numbers.
This does not fix routing of the journals/data dirs.

[...]

PR #2975, opened 2022-01-05
Changing the routing of ledgers to journal directories

1559924775:
issue : #2974

**BUG REPORT**

***Describe the bug***

When the journal directory is configured to be even, the ledger allocated to each journal will be uneven. The root cause is that almost all the ledger IDs generated by the LedgerIdGenerator class are even. The number of ZK temporary order nodes is determined by the cversion of the parent node. Whenever a child node is added or deleted, it will increase by 1. The LedgerIdGenerator class uses ZK's temporary order node to generate the ledgerid, and deletes the node immediately after generation. This way of use causes almost all cversions to be even, resulting in all ledgerids to be even.

***To Reproduce***

1. Configure in bk_server.conf: jouranlDirectorie=/tmp/bk-txn1,/tmp/bk-txn2
2. production data to bookie

***Expected behavior***

Almost all ledgerids are assigned to directory 1

***Screenshots***

ledgerIdGenerator class :  ZKledgerIdGenerator

journal

mathutil

Changing the routing of ledgers to journal directories can fix this bug
mathutilnew

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
@@ -36,6 +36,7 @@ import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.FilenameFilter;
 import java.io.IOException;
+import java.math.BigInteger;
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.URI;
@@ -43,6 +44,7 @@ import java.net.UnknownHostException;
 import java.nio.ByteBuffer;
 import java.nio.file.FileStore;
 import java.nio.file.Files;
+import java.security.MessageDigest;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
@@ -1259,7 +1261,17 @@ public class BookieImpl extends BookieCriticalThread implements Bookie {
     }
 
     private Journal getJournal(long ledgerId) {
-        return journals.get(MathUtils.signSafeMod(ledgerId, journals.size()));
+        String dividendString = Double.toString(ledgerId);
+        byte[] secretBytes = null;
+        int index = 0;
+        try {
+            secretBytes = MessageDigest.getInstance("md5").digest(dividendString.getBytes());
+            BigInteger md5Code = new BigInteger(1, secretBytes);
+            index = md5Code.mod(new BigInteger(Integer.toString(journals.size()))).intValue();
+        } catch (Exception e) {
+            LOG.error("get journal directory failed exception:{}", e.getMessage());
+        }
+        return journals.get(index);
     }
 
     /**


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
```diff
@@ -1259,7 +1261,17 @@ LedgerDescriptor getLedgerForEntry(ByteBuf entry, final byte[] masterKey)
     }
 
     private Journal getJournal(long ledgerId) {
-        return journals.get(MathUtils.signSafeMod(ledgerId, journals.size()));
+        String dividendString = Double.toString(ledgerId);
+        byte[] secretBytes = null;
+        int index = 0;
+        try {
+            secretBytes = MessageDigest.getInstance("md5").digest(dividendString.getBytes());
+            BigInteger md5Code = new BigInteger(1, secretBytes);
```
eolivelli, line 1269: It looks to me that computing this value for every write to the journal is quite expensive.

Can we find something less resource consuming?

dlg99: +1 for the performance concern.

Would something like
```
final int PRIME = 31;
return journals.get(MathUtils.signSafeMod(ledgerId, PRIME * journals.size()));
```
work well enough?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
```diff
@@ -1259,7 +1261,17 @@ LedgerDescriptor getLedgerForEntry(ByteBuf entry, final byte[] masterKey)
     }
 
     private Journal getJournal(long ledgerId) {
-        return journals.get(MathUtils.signSafeMod(ledgerId, journals.size()));
+        String dividendString = Double.toString(ledgerId);
```
pkumar-singh, line 1264: Probably makes more sense to use Long.toString(ledgerId); since ledgerId is long?

dlg99:
@1559924775 I hope you are still interested in fixing this issue.
My PR https://github.com/apache/bookkeeper/pull/3023 fixes similar problem with routing of requests to threads, I think the routing logic change can be reused for the journals and it is faster than md5/murmur/other hashing. 

As a follow up to the discussion https://github.com/apache/bookkeeper/issues/2974#issuecomment-1028054818 the change requires some additional thinking to prevent data loss during upgrade/downgrade.
I think backwards compat tests can help with this, but you will need to add tests for upgrade from old version with 2 journal/2 data disks to the new version with changed routing. The test will need to repro the case when journal replay is needed etc.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
1559924775: > @1559924775 I hope you are still interested in fixing this issue. My PR #3023 fixes similar problem with routing of requests to threads, I think the routing logic change can be reused for the journals and it is faster than md5/murmur/other hashing.
> 
> As a follow up to the discussion [#2974 (comment)](https://github.com/apache/bookkeeper/issues/2974#issuecomment-1028054818) the change requires some additional thinking to prevent data loss during upgrade/downgrade. I think backwards compat tests can help with this, but you will need to add tests for upgrade from old version with 2 journal/2 data disks to the new version with changed routing. The test will need to repro the case when journal replay is needed etc.

Thanks for your idea. I'll consider it.

StevenLuMT:
rerun failure checks

StevenLuMT (changes requested):
The author of this pr has not been updated for a long time, 
should it be removed from 4.16.0 first? @hangc0276  @eolivelli 
 or should I propose a new pr to solve this problem?

StevenLuMT:

PR #3006, opened 2022-01-21
improve rocksDB

StevenLuMT:

### Motivation

1. some old parameters in rocksDB is not configurable
2. the rocksDB write/read has no rate limiter

### Changes

1.rocks all old parameter change to be configurable
2.add RateLimiter feature,default closed


StevenLuMT:
rerun failure checks

StevenLuMT:
@eolivelli @pkumar-singh @zymap
If you have time, please help me review it, thank you

mauricebarnum:
have you considered adding support to use the native RocksDB configuration file support? https://github.com/facebook/rocksdb/blob/main/java/src/main/java/org/rocksdb/OptionsUtil.java

StevenLuMT:
> have you considered adding support to use the native RocksDB configuration file support? https://github.com/facebook/rocksdb/blob/main/java/src/main/java/org/rocksdb/OptionsUtil.java

I think it's a good idea, finished this pr, I will research that direction

zymap:
I would suggest using the RocksDB config file directly. That will make us don't need to introduce extra configuration changes in the bookkeeper repo.

PR #3010, opened 2022-01-25
Forced the same version of guava (plus upgraded it); fixed deprecations etc.

dlg99:
### Motivation

Older versions of guava (w/CVEs) used in some subprojects

### Changes

Forced the same version of guava; fixed deprecation problems (murmur3_32) and compilation problems (checkArgument).
checkArgument cannot be statically imported because there are now overrides of it; checkstyle was not very cooperative so I had to remove the import altogether. 
Then updated the guava version to match one in Pulsar.


nicoloboschi (approved):

eolivelli (approved):
LGTM

eolivelli:
please fix the conflicts

dlg99:
@eolivelli I rebased it, I believe we'll get conflicts in either this or netty update PR, depends on which one is merged first.

PR #3017, opened 2022-01-27
[build] do not run spotbugsTest by default

nicoloboschi:
### Motivation

`spotbugsTest` fails on a few submodules.
Run spotbugs against test files is useless and slow down development.

We can just skip it by default without having to specify `-x spotbugsTest`

### Changes

Skip `spotbugsTest` by default

### DIFF
--- a/.github/workflows/pr-validation.yml
+++ b/.github/workflows/pr-validation.yml
@@ -47,6 +47,6 @@ jobs:
         with:
           java-version: 1.8
       - name: Validate pull request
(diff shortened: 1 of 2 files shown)


eolivelli (changes requested):
We should run spotbugs by default.
This way when you build the project after making some changes you are sure that you are still  over the quality bar that we set for the project.

Spotbugs must always pass.

nicoloboschi:
I'm removing it only for the test classes

eolivelli (approved):
thanks
My Gradle fu is so weak :-) 

LGTM

PR #3021, opened 2022-01-27
CI workflow to check dependencies with OWASP

dlg99:
### Motivation

Detect CVEs early and make them visible to the committers

### Changes

Added GH workflow that will fail if new CVE with level > 7 detected. The plan is to not treat it as blocking merge.
Currently some projects that will not pass the check until other pending PRs are merged.

### DIFF
--- a/.github/workflows/owasp-dep-check.yml
+++ b/.github/workflows/owasp-dep-check.yml
@@ -47,8 +47,8 @@ jobs:
         with:
           java-version: 1.8
 
-      - name: run "clean install verify" to trigger dependency check
-        run: ./gradlew clean build -x microbenchmarks:checkstyleMain -x spotbugsTest -x signDistTar -x test dependencyCheckAggregate
+      - name: run "clean build dependencyCheckAggregate" to trigger dependency check
+        run: ./gradlew clean build -x signDistTar -x test dependencyCheckAggregate
 
       - name: Upload report
         uses: actions/upload-artifact@v2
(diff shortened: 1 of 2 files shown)


.github/workflows/owasp-dep-check.yml
```diff
@@ -0,0 +1,59 @@
...
+
+      - name: Tune Runner VM
+        uses: ./.github/actions/tune-runner-vm
+
+      - name: Set up JDK 1.8
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+
+      - name: run "clean install verify" to trigger dependency check
```
nicoloboschi, line 50: Gradle?

.github/workflows/owasp-dep-check.yml
```diff
@@ -0,0 +1,59 @@
...
+        run: ./gradlew clean build -x microbenchmarks:checkstyleMain -x spotbugsTest -x signDistTar -x test dependencyCheckAggregate
```
nicoloboschi, line 51: -x sptbugsTest and -x microbenchmarks:checkstyleMain are not necessary

nicoloboschi (changes requested):

.github/workflows/owasp-dep-check.yml
dlg99: fixed

.github/workflows/owasp-dep-check.yml
dlg99: fixed

PR #3023, opened 2022-02-04
Issue 2974: better thread selection for the Ordered Executor

dlg99:
### Motivation

uneven routing of requests to threads caused by uneven distribution of ledger Ids can leave some threads underutilized

### Changes

Experimented with a few approaches and picked the simplest/fastest one that produced good results.
Locally tested with keys (ledger ids) up to 100mil across up to 512 threads.

This does not address the fact that generated ledger ids are skewed towards even numbers.
This does not fix routing of the journals/data dirs. 

Master Issue: #2974


eolivelli (approved):
This is a great idea !
small change, big benefit !

@Vanlightly would you have time to take a look ?

eolivelli:
@Vanlightly FYI

PR #3030, opened 2022-02-09
Fix performance issues to avoid unnecessary loops.

mattisonchao:
### Motivation

When we get the right ``authorized`` state, we need to return immediately, do not need to do an unnecessary loop.

### Changes

- Add ``break``, when get the right ``authorized`` state.


mattisonchao:
@merlimat @eolivelli @dlg99 @nicoloboschi  PTAL :)

eolivelli (approved):
LGTM

but this is not a big deal, as "allowedRoles" is usually a very small list, like 1 entry

mattisonchao:
Oh, thanks for the explanation.  @eolivelli 

But I think I can start small to get involved in community contributions. What would you think about it?

mattisonchao:
Could I use some commands to rerun failed test? 
I think it's a flaky test, my change can not affect this.

mattisonchao:
rerun failure checks

dlg99 (approved):

nicoloboschi (approved):

merlimat (approved):

Issue #3031, opened 2022-02-09
[FLAKY TEST] stream:storage:impl:test / MVCCStoreFactoryImplTest flaked on CI


dlg99:
**BUG REPORT**

MVCCStoreFactoryImplTest flaked on CI run

```
./gradlew stream:storage:impl:test
...
  java.lang.AssertionError: expected same:<com.google.common.util.concurrent.MoreExecutors$ScheduledListeningDecorator@7f4701b1[java.util.concurrent.ScheduledThreadPoolExecutor@adcfad9[Running, pool size = 1, active threads = 0, queued tasks = 0, completed tasks = 1]]> was not:<com.google.common.util.concurrent.MoreExecutors$ScheduledListeningDecorator@fb49fdf[java.util.concurrent.ScheduledThreadPoolExecutor@3a2e9f5b[Running, pool size = 1, active threads = 0, queued tasks = 1, completed tasks = 6]]>
      at org.apache.bookkeeper.stream.storage.impl.store.MVCCStoreFactoryImplTest.testOpenStore(MVCCStoreFactoryImplTest.java:149)
```

***Expected behavior***

stable passing test

shoothzj:
closed by #3032

PR #3032, opened 2022-02-09
[ISSUE 3031] fixed test; chooseThread() uses orderingKey as a param, not a thread index

dlg99:
### Motivation

Test flaked on CI

### Changes

Fixed test

Master Issue: #3031


eolivelli (approved):
Amazing catch!

Issue #3038, opened 2022-02-10
[FLAKY TEST] org.apache.bookkeeper.bookie.CompactionByBytesTest flaked on CI with "AssertionError: Found entry log file ([0,1,2].log that should have not been compacted"


dlg99:
**BUG REPORT**

***To Reproduce***

run `./gradlew bookkeeper-server:test --tests="org.apache.bookkeeper.bookie.CompactionByBytesTest"` or `./gradlew bookkeeper-server:test --tests="org.apache.bookkeeper.bookie.CompactionByBytesTest.testMinorCompactionWithMaxTimeMillis"`
test flakes

```
org.apache.bookkeeper.bookie.CompactionByBytesTest

  Test testMinorCompactionWithMaxTimeMillis FAILED (19.8s)

  java.lang.AssertionError: Found entry log file ([0,1,2].log that should have not been compacted in ledgerDirectory: /var/folders/12/xc2pp74n2mg79jtbp_f3329m0000gn/T/bookie4335487681326103853test
```

***Expected behavior***

test passes reliably

PR #3039, opened 2022-02-11
[ISSUE 3038] Fixed flaky CompactionTest.testMinorCompactionWithMaxTimeMillis

dlg99:

### Motivation

Flaky  test

### Changes

AFAICT the test flaked because sometimes provided 15ms timeout for the compaction was not enough.
Possibly related to addition of `baseConf.setGcEntryLogMetadataCacheEnabled(true);` 
I split the test into two:
- one with much higher timeout to make sure compaction can finish
- one with a very short timeout to make sure at least one entry log won't get compacted

Master Issue: #3038


eolivelli (approved):
LGTM

thanks

PR #3042, opened 2022-02-11
[ci] Dump stacktrace when a job is cancelled

nicoloboschi:
### Motivation

Sometimes CI jobs fail due to timeout. It would be useful understand what the latest test was doing before being interrupted. 

### Changes

* Added a new script for dumping stacktrace.
* Added in all the jobs the step in case of `cancelled()` is true.


dlg99 (approved):

nicoloboschi:
rerun failure checks

nicoloboschi:
rerun failure checks

nicoloboschi:
rerun failure checks

PR #3046, opened 2022-02-14
Avoiding call fileChannelProvider init multiple times

hangc0276:
### Motivation
In current JournalChannel implementation, it create fileChannelProvider instance every time, which is unnecessary.
https://github.com/apache/bookkeeper/blob/4debbbf84db09f80ffeb4154e8e6feed134be0c0/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/JournalChannel.java#L162

### Changes
Create fileChannelProvider in Journal creation, avoiding call fileChannelProvider init multiple times.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
@@ -618,8 +618,8 @@ public class Journal extends BookieCriticalThread implements CheckpointSource {
     final File journalDirectory;
     final ServerConfiguration conf;
     final ForceWriteThread forceWriteThread;
+    final FileChannelProvider fileChannelProvider;
 
-    private FileChannelProvider fileChannelProvider;
     // Time after which we will stop grouping and issue the flush
     private final long maxGroupWaitInNanos;
     // Threshold after which we flush any buffered journal entries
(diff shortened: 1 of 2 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -618,6 +618,8 @@ static void writePaddingBytes(JournalChannel jc, ByteBuf paddingBuffer, int jour
     final File journalDirectory;
     final ServerConfiguration conf;
     final ForceWriteThread forceWriteThread;
+
+    private FileChannelProvider fileChannelProvider;
```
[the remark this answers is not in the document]
hangc0276: We have set the fileChannelProvider twice in try-catch, so it can't be final

hangc0276:
rerun failure checks

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -731,6 +733,14 @@ public Journal(int journalIndex, File journalDirectory, ServerConfiguration conf
             LOG.debug("Last Log Mark : {}", lastLogMark.getCurMark());
         }
 
+        try {
+            this.fileChannelProvider = FileChannelProvider.newProvider(conf.getJournalChannelProvider());
+        } catch (IOException e) {
+            LOG.warn("Failed to initiate {}, use default fileSystem provider instead.",
+                conf.getJournalChannelProvider());
+            this.fileChannelProvider = new DefaultFileChannelProvider();
```
eolivelli, line 741: We should fail the bookie, otherwise the system will not work as configured by the administrator

eolivelli (changes requested):
Overall looks good  but I left one comment

zymap (approved):

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
hangc0276: done

hangc0276:
rerun failure checks

hangc0276:
rerun failure checks

hangc0276:
ping @eolivelli  @nicoloboschi , Please help review this PR, thanks a lot.

eolivelli (approved):

Issue #3048, opened 2022-02-14
Flaky-test: "Compatibility Check Java11" times out frequently


dlg99:
"Compatibility Check Java11" workflow times out frequently on CI

dlg99:
Duplicate of https://github.com/apache/bookkeeper/issues/3044

PR #3050, opened 2022-02-14
[Issue 3048] Increase timeout for Java11 workflow

dlg99:

### Motivation

Java11 compat workflow timesout frequently but passes on rerun.

### Changes

Increased timeout

Master Issue: #3048


dlg99:
It's probably not the best idea.
When the tests succeed they succeed in 15min.

PR #3060, opened 2022-02-17
Catch onBookieRackChange exception

hangc0276:
### Motivation
When we update the bookie rack info, it will use all the bookie list to update rack topology. However If one bookie update failed and throw exception out, it will throw the exception out and the remains bookie info won't be updated into the rack topology, which will affect the ledger ensemble selection.

### Changes

Catch the bookie topology update exception to ensure the remaining bookies' info can be updated into the rack topology.


hangc0276:
rerun failure checks

hangc0276:
rerun failure checks

hangc0276:
rerun failure checks

hangc0276:
rerun failure checks

hangc0276:
rerun failure checks

zymap (commented):
Overall looks good. Could you please add a unit test?

zymap (approved):

PR #3062, opened 2022-02-17
Support isolate read write thread pool

### LINKED ISSUE #3003: Split read and write orderExecutor
**BUG REPORT**

***Describe the bug***

Now the read and write of BookkeeperClient share an orderExecutor. When cold data is read, it will take a long time. At this time, the read request will be blocked, resulting in the write timeout.

***Expected behavior***
Thread pools for reading and writing should be separated

315157973:
fix #3003
### Motivation
Now , when the Ledger is in the `Writable` state, the read and write of BookkeeperClient share the same orderExecutor. 
But, we can not guarantee that the data is not read from disk. 
In order to avoid reading from affecting writing, we need to add a configuration item to allow read and write requests to use different thread pools

### Changes
Add a param to support isolate read write thread pool.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ClientInternalConf.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ClientInternalConf.java
@@ -49,6 +49,7 @@ class ClientInternalConf {
     final boolean enableBookieFailureTracking;
     final boolean useV2WireProtocol;
     final boolean enforceMinNumFaultDomainsForWrite;
+    final boolean isolateReadWriteThreadPool;
 
     static ClientInternalConf defaultValues() {
         return fromConfig(new ClientConfiguration());
@@ -83,6 +84,7 @@ class ClientInternalConf {
         this.useV2WireProtocol = conf.getUseV2WireProtocol();
         this.enableStickyReads = conf.isStickyReadsEnabled();
         this.enforceMinNumFaultDomainsForWrite = conf.getEnforceMinNumFaultDomainsForWrite();
+        this.isolateReadWriteThreadPool = conf.getIsolateReadWriteThreadPool();
 
         if (conf.getFirstSpeculativeReadTimeout() > 0) {
             this.readSpeculativeRequestPolicy =
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
@@ -516,6 +516,10 @@ public class LedgerHandle implements WriteHandle {
         return !getLedgerMetadata().isClosed() && handleState == HandleState.OPEN;
     }
 
+    boolean isolateReadWriteThreadPool() {
+        return clientCtx.getConf().isolateReadWriteThreadPool;
+    }
+
     void asyncCloseInternal(final CloseCallback cb, final Object ctx, final int rc) {
         try {
             doAsyncCloseInternal(cb, ctx, rc);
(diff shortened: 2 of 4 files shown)


315157973:
rerun failure checks

dlg99 (changes requested):
1. read request doesn't wait for the completion on that threadpool. 

> When cold data is read, it will take a long time. At this time, the read request will be blocked, resulting in the write timeout.

Can I see a test reproing this?
i suspect the problem is somewhere else.

2. The ordering there is required for a reason. 

The test in this PR simply does not cover anything related to actual functionality. I suggest trying to enable the feature by default and running all tests to see what breaks, it's hard to come up with something more specific off the top of my head. 

3. Do you have apples-to-apples comparison of performance, same load, same bookies etc, the feature off/on?

315157973:
> 1. read request doesn't wait for the completion on that threadpool.
> 
> > When cold data is read, it will take a long time. At this time, the read request will be blocked, resulting in the write timeout.
> 
> Can I see a test reproing this? i suspect the problem is somewhere else.
> 
> 2. The ordering there is required for a reason.
> 
> The test in this PR simply does not cover anything related to actual functionality. I suggest trying to enable the feature by default and running all tests to see what breaks, it's hard to come up with something more specific off the top of my head.
> 
> 3. Do you have apples-to-apples comparison of performance, same load, same bookies etc, the feature off/on?

OK, I will  do a performance test and add more monitoring information

lhotari:
btw. It's worth considering thread safety issues and preventing them. In Pulsar, https://github.com/apache/pulsar/issues/14436 seems to be caused by a thread safety issue in Bookkeeper client. It might be caused by #1792 changes which are in the same area as this PR.

StevenLuMT:

lordcheng10:
> 1. 读取请求不会等待该线程池完成。
> 
> > 读取冷数据时，需要很长时间。这时候读请求会被阻塞，导致写超时。
> 
> 我可以看到一个测试重现这个吗？ 我怀疑问题出在其他地方。
> 
> 2. 那里的订购是有原因的。
> 
> 此 PR 中的测试根本不涵盖与实际功能相关的任何内容。我建议尝试默认启用该功能并运行所有测试以查看有什么问题，很难想出更具体的东西。
> 
> 3. 您是否对性能、相同负载、相同 bookies 等、功能关闭/打开进行了同类比较？



> 1. read request doesn't wait for the completion on that threadpool.
> 
> > When cold data is read, it will take a long time. At this time, the read request will be blocked, resulting in the write timeout.
> 
> Can I see a test reproing this? i suspect the problem is somewhere else.
> 
> 2. The ordering there is required for a reason.
> 
> The test in this PR simply does not cover anything related to actual functionality. I suggest trying to enable the feature by default and running all tests to see what breaks, it's hard to come up with something more specific off the top of my head.
> 
> 3. Do you have apples-to-apples comparison of performance, same load, same bookies etc, the feature off/on?

For ledgers in the Writable state, reading and writing share the same thread pool, and there is no way to isolate reading and writing;
IMO, it is useful to isolate reads and writes to different threads of execution; @dlg99

lordcheng10:
> 1. read request doesn't wait for the completion on that threadpool.
> 
> > When cold data is read, it will take a long time. At this time, the read request will be blocked, resulting in the write timeout.
> 
> Can I see a test reproing this? i suspect the problem is somewhere else.
> 
> 2. The ordering there is required for a reason.
> 
> The test in this PR simply does not cover anything related to actual functionality. I suggest trying to enable the feature by default and running all tests to see what breaks, it's hard to come up with something more specific off the top of my head.
> 
> 3. Do you have apples-to-apples comparison of performance, same load, same bookies etc, the feature off/on?

For ledgers in the Writable state, reading and writing share the same thread pool, and there is no way to isolate reading and writing;
IMO, it is useful to isolate reads and writes to different threads of execution

PR #3067, opened 2022-02-19
Recover only those entries which as supposed to be on dead bookie

pkumar-singh:
**Motivation**

When WQ < E there would be case when some entries of a fragment will be on Bookie B1 and not on B2 while some will on B2 but not B1. Currently If B1 is dead and being recovered, the process will not only recover entries which were on B1 but also those entries which were not on B1 but were on B2.

**Change**
From the fragment recover only those entries which were supposed to be on dead bookie

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerFragmentReplicator.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerFragmentReplicator.java
@@ -133,6 +133,16 @@ public class LedgerFragmentReplicator {
     private static final Logger LOG = LoggerFactory
             .getLogger(LedgerFragmentReplicator.class);
 
+    private boolean isEntryExpectedOnDeadBookie(final LedgerHandle lh, Long eId, Set<Integer> deadBookieIndex) {
+        DistributionSchedule.WriteSet writeSet = lh.getWriteSetForReadOperation(eId);
+        for (Integer deadIndex : deadBookieIndex) {
+            if (!writeSet.contains(deadIndex)) {
+                return false;
+            }
+        }
+        return true;
+    }
+
     private void replicateFragmentInternal(final LedgerHandle lh,
             final LedgerFragment lf,
             final AsyncCallback.VoidCallback ledgerFragmentMcb,
@@ -173,7 +183,9 @@ public class LedgerFragmentReplicator {
         List<Long> entriesToReplicate = new LinkedList<Long>();
         long lastStoredEntryId = lf.getLastStoredEntryId();
         for (long i = lf.getFirstStoredEntryId(); i <= lastStoredEntryId; i++) {
-            entriesToReplicate.add(i);
+            if (isEntryExpectedOnDeadBookie(lh, i, lf.getBookiesIndexes())) {
+                entriesToReplicate.add(i);
+            }
         }
         /*
          * Now asynchronously replicate all of the entries for the ledger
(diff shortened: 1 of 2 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerFragmentReplicator.java
```diff
@@ -133,6 +133,16 @@ public LedgerFragmentReplicator(BookKeeper bkc, ClientConfiguration conf) {
     private static final Logger LOG = LoggerFactory
             .getLogger(LedgerFragmentReplicator.class);
 
+    private boolean isEntryExpectedOnDeadBookie(final LedgerHandle lh, Long eId, Set<Integer> deadBookieIndex) {
```
mauricebarnum, line 136: looks like this would work and might be the shortest path from where you are to where you want to get.

it's somewhat unfortunate you can't just configure the LedgerHandle to avoid the dead bookies when creating the write set.  there's almost support in there for doing that, but it isn't too flexible and ultimately optional at both the LedgerHandle and DistributionSchedule implementation levels.  see how bookie health info is almost useful.

zymap:
Could you please add a test to cover this?

pkumar-singh:
> Could you please add a test to cover this?

Done.

nicoloboschi (approved):

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerFragmentReplicator.java
eolivelli, line 136: why the name is "deadBookieIndex" ? you are passing "lf.getBookiesIndexes()"

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerFragmentReplicator.java
```diff
@@ -133,6 +133,16 @@ public LedgerFragmentReplicator(BookKeeper bkc, ClientConfiguration conf) {
...
+        DistributionSchedule.WriteSet writeSet = lh.getWriteSetForReadOperation(eId);
```
eolivelli, line 137: we should call `writeSet .recycle()` in a finally block in order to reduce allocations

eolivelli (changes requested):

hangc0276:
ping @pkumar-singh, Would you please take a look at the above comments?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerFragmentReplicator.java
StevenLuMT: The same advice,
you needs to recycle writeSet at the end,code like 

```
try{
     DistributionSchedule.WriteSet writeSet = lh.getWriteSetForReadOperation(eId);
     for (Integer deadIndex : deadBookieIndex) {
            if (!writeSet.contains(deadIndex)) {
                return false;
            }
      }
      writeSet.recycle();
}
```

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerFragmentReplicator.java
```diff
@@ -133,6 +133,16 @@ public LedgerFragmentReplicator(BookKeeper bkc, ClientConfiguration conf) {
...
+        for (Integer deadIndex : deadBookieIndex) {
+            if (!writeSet.contains(deadIndex)) {
+                return false;
+            }
+        }
```
StevenLuMT, line 142: The same advice as @eolivelli 
@pkumar-singh you needs to recycle writeSet at the end,code like 

```
DistributionSchedule.WriteSet writeSet=null;
try{
     writeSet = lh.getWriteSetForReadOperation(eId);
     for (Integer deadIndex : deadBookieIndex) {
            if (!writeSet.contains(deadIndex)) {
                return false;
            }
      }
} finally {
   if (writeSet!=null){
    writeSet.recycle();
   }
}

```

StevenLuMT:

PR #3077, opened 2022-02-24
[tests] Replace powermockito usages with mockito-inline - 'tools' submodule

nicoloboschi:
### Motivation

PowerMockito static mocks don't work well with the latest JDK class encapsulation mechanisms. The main problem is that PowerMockito, in order to mock a single static method of a class, try to setAccessible a lot of other apparently related classes. This means, starting from JDK17, you need to add `--add-opens` to the Java runtime options in order to avoid runtime errors.

PowerMockito development isn't much active and they don't seem to be ready for the new JDK releases. 
The best replacement is to use [mockito-inline](https://javadoc.io/doc/org.mockito/mockito-core/latest/org/mockito/Mockito.html#48) (additional part of `mockito-core`), which uses a different algorithm in order to mock static methods - a lot less invasive.  

### Changes

* Mockito upgrade
* Replaced PowerMockito#mockStatic usages with MockedStatic
* Added a general purpose helper class with a less verbose API for mocking static class and constructions, based on mockito-inline API
* Modified only the 'tools' module


nicoloboschi:
rerun failure checks

dlg99 (approved):
Excellent work!
Thank you.

dlg99:
image

But seriously, I looked through it, lots of similar/boilerplate changes and the rest makes sense. 
Thank you for taking on this task.

nicoloboschi:
@eolivelli @merlimat PTAL

eolivelli (approved):
Great work.
I am surprised that we can do everything with only one MockedConstruction per class per test.
BTW if it works I am supportive of this patch

@merlimat @Ghatage @dlg99 PTAL

PR #3084, opened 2022-03-07
Upgrade ZooKeeper dependency to 3.8.0 - gradle

eolivelli:
### Motivation

We are using an old version of ZooKeeper, 3.6.x (and ZooKeeper 3.5.x was declared EOL a few weeks ago).
We should upgrade to the latest version.

### Changes

Upgraded ZooKeeper dependency.


nicoloboschi (changes requested):
we should also update the backward compatibilty tests
https://github.com/apache/bookkeeper/blob/74bbeec7a61ecf948188485f9f5e0f6a2a210567/tests/integration-tests-topologies/src/main/resources/cube-definitions/3-node-all-version-unstarted.yaml#L25

dlg99:
License check failed, OWASP detects new CVEs in zk 3.8 (are these real or misdetections?) + test failures.
These have to be addressed.

Shoothzj (approved):

eolivelli:
@dlg99 

regarding OWASP we can follow this discussion
https://github.com/apache/pulsar/pull/14630

basically there are false positives about "Jetty" that the OWASP checker reports in zookeeper jars

in ZooKeeper project we added these exclusions
https://github.com/apache/zookeeper/commit/3004c909b78b3056985c8e39925e14bde3baa430


Regarding the LICENSE check...my Gradle Fu is very weak and I am not able to tell Gradle to fully exclude LogBack and fix the package.

Any suggestion is really appreciated

dlg99:
You can suppress false positives at https://github.com/apache/bookkeeper/blob/master/src/owasp-dependency-check-suppressions.xml

As for the logback, try something like
```
configurations {
  runtime.exclude group: "......", module: "....."
}
```
or 
```
configurations {
  all.exclude group: "......", module: "....."
}
```
Search for "configurations" in the `build.gradle`

Maybe @lhotari  knows a better way

eolivelli:
@dlg99 thanks for your advice but I cannot make it work

I hope that some Gradle expert could help
@pkumar-singh @lhotari @nicoloboschi

Issue #3098, opened 2022-03-12
support some thread pool associated with CPU cores


leizhiyuan:
**FEATURE REQUEST**

1. Please describe the feature you are requesting.

```
numAddWorkerThreads   

numHighPriorityWorkerThreads

numJournalCallbackThreads

numReadWorkerThreads
```

this thread pool is hard code, we use 16/32/64core cpu, it was small for performance.

3. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have).
   Are you currently using any workarounds to address this issue?

4. Provide any additional detail on your proposed use case for this feature.

shoothzj:
I think as the #3099 describes. It's a breaking change. We can take a consider in next major version.

PR #3099, opened 2022-03-12
feat: use Runtime.getRuntime().availableProcessors() as thread pool size

leizhiyuan:

### Motivation

### Changes

Master Issue: #3098 

> ---
>
> ---
> quickly and easily:
>
>
>
> Otherwise:
> 
> 
> ---


eolivelli (changes requested):
Thanks for your contribution 

Can you explain more about the motivation for this change?

leizhiyuan:
> Thanks for your contribution
> 
> Can you explain more about the motivation for this change?

Thanks for your reply , now, we have different bk clusters, if all clusters have different cpu cores, we need to provider some different configs ,but it can be set by CPU cores automatically

hangc0276 (changes requested):
It will change the default behavior. If the node has 128 cores, by default, it will create a lot of threads

liangyuanpeng (commented):
Maybe we can make them configurable.

StevenLuMT:

PR #3100, opened 2022-03-12
Fix bkperf message rate limit to 2GB/s

hangc0276:
### Motivation
Current bkperf message rate limit use `int` to store the rate value, however, max rate will be limit by `MAX_INTEGER` value to 2GB. We'd better use long to support higher message rate limiter.

### Changes

Use `long` instead of `int` to store the message rate limit value.


hangc0276:
rerun failure checks

eolivelli (approved):
Lgtm

eolivelli:
@dlg99 fyi

Issue #3104, opened 2022-03-14
Recycled LedgerEntryImpl instances are corrupted due to invalid recycling in BK client


lhotari:
**BUG REPORT**

***Describe the bug***

See https://github.com/apache/pulsar/issues/14436#issuecomment-1064758710 for a detailed analysis by @congbobo184 about the symptoms of the thread safety issue.

***To Reproduce***

Follow instructions in https://github.com/apache/pulsar/issues/14436

***Expected behavior***

BK client should handle LedgerEntryImpl instances correctly so that it would prevent the issue where entries are used in 2 threads concurrently, perhaps due to "double recycling".

eolivelli:
Makes sense to me.
I am fine with reverting #3105

data corruption is a bad problem, "being faster" is only an enhancement.

we should revert that change on master branch and on other active branches and possibly cut a new release

eolivelli:
PR for the revert:
https://github.com/apache/bookkeeper/pull/3106

great work @lhotari @congbo1984

lhotari:
This will be fixed by #3110

lhotari:
Fixed by #3110

Issue #3105, opened 2022-03-14
Improve performance of OrderedExecutor by switching to a more performant BlockingQueue implementation


lhotari:
**FEATURE REQUEST**

[OrderedExecutor uses `LinkedBlockingQueue`](https://github.com/apache/bookkeeper/blob/3db4de9da1447e2a5135ac9330d4347b9d220a0d/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/OrderedExecutor.java#L307-L308) (when `enableBusyWait`=false). This is not optimal. 

It looks like @merlimat introduced [GrowableArrayBlockingQueue](https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/util/collections/GrowableArrayBlockingQueue.java)  in #153 with the intention to later replace `LinkedBlockingQueue` usage with `GrowableArrayBlockingQueue`. However the swtich hasn't happened.
> In multiple places, (eg: journal, ordered executor, etc..), we are using LinkedBlockingQueue instances to pass objects between threads.

See #153

### Additional context

The proposal of this issue is to make this change since it would be an alternative mitigation for the performance issue #1791 . The changes in #1792 will most likely need to be reverted because the change introduced thread safety issues such as #3104. By improving the queue implementation, there would be better performance without breaking thread safety.

lhotari:
Pulsar contains GrowableArrayBlockingQueue too, https://github.com/apache/pulsar/blob/master/pulsar-common/src/main/java/org/apache/pulsar/common/util/collections/GrowableArrayBlockingQueue.java . I picked one fix to the BK version of GrowableArrayBlockingQueue when comparing the classes.

PR #3106, opened 2022-03-14
Revert "Issue #1791: Read Submission should bypass OSE Threads"

eolivelli:
This reverts commit 6b99ff73278d4e655509f61ea44284e54716f5a8.

Revert #1792 "Issue #1791: Read Submission should bypass OSE Threads"

### Motivation
See https://github.com/apache/bookkeeper/issues/3104

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
@@ -887,14 +887,7 @@ public class LedgerHandle implements WriteHandle {
                 ws.recycle();
             }
 
-            if (isHandleWritable()) {
-                // Ledger handle in read/write mode: submit to OSE for ordered execution.
-                clientCtx.getMainWorkerPool().executeOrdered(ledgerId, op);
-            } else {
-                // Read-only ledger handle: bypass OSE and execute read directly in client thread.
-                // This avoids a context-switch to OSE thread and thus reduces latency.
-                op.run();
-            }
+            clientCtx.getMainWorkerPool().executeOrdered(ledgerId, op);
         } else {
             op.future().completeExceptionally(BKException.create(ClientClosedException));
         }


eolivelli:
@diegosalvi this change is also important for HerdDB and DistributedLog

lhotari:
There's #3105 as a possible solution for mitigating the performance issue which was the original motivation for the PR which is reverted by this PR.

lhotari (approved):

diegosalvi:
@eolivelli interesting, thank you. We'll test disabling netty recycler as stated in https://github.com/apache/pulsar/issues/14436 for now

nicoloboschi (approved):

dlg99 (changes requested):
@congbobo184 @lhotari As I understand, there is no confirmation that reverting this PR actually fixes the problem reported in Pulsar.

If this problem is related to calling netty's recycle from another thread (which netty had and fixed) I'd love to see something that actually proves the theory so netty can fix it. 

I don't see anything in PendingReadOp that requires initiate() and readEntryComplete() to run on the same thread for the read-only ledger. I can be wrong, please help me understand what I am missing. readEntryComplete() still runs on the OSE.

What you see in the log mentioned in [the comment](https://github.com/apache/pulsar/issues/14436#issuecomment-1064758710)  might happen if:
- request took longer than speculative retry time
- speculative read got submitted
- both original request and the speculative retry succeeded

It is possible that handling of this case somehow corrupts recyclable entry though i don't see yet where it could happen (callbacks should be handled on the same OSE thread etc)

I need more proof to confirm this actually fixes the problem and not just reverts some PR that touches code you are suspecting.

jvrao:
> 

+1 on @dlg99 's comment. I could not understand where exactly the problem and how it could get addressed by making create and recycle run in the same thread. You could always have outstanding read responses after erroring out. But even if that happens, what exactly the error/corruption that is visible to the client?

congbobo184:
https://github.com/apache/bookkeeper/pull/3110 will fix it

lhotari:
> You could always have outstanding read responses after erroring out. But even if that happens, what exactly the error/corruption that is visible to the client?

@jvrao The issue description is #3104 and there are references to a Pulsar issue with a lot of troubleshooting comments. @congbobo184 has also summarized the behavior in the PR #3110 description.

eolivelli:
Closing for now

hangc0276 (changes requested):
The root cause of https://github.com/apache/bookkeeper/issues/3104 is not caused by the thread safety problem. And the isolation thread pool for cold and hot data reading will greatly reduce thread switching overhead and make the throughput more stable.

This issue has been fixed by https://github.com/apache/bookkeeper/pull/3110

PR #3108, opened 2022-03-14
Issue #3105: Optimize OrderedExecutor performance by using GrowableArrayBlockingQueue

### LINKED ISSUE #3105: Improve performance of OrderedExecutor by switching to a more performant BlockingQueue implementation

lhotari:
Fixes #3105

### Motivation

- See #3105 for details. OrderedExecutor uses LinkedBlockingQueue which isn't the most performant BlockingQueue implementation. For example, it causes object allocations each time new items are added to the queue. More details in #153 where GrowableArrayBlockingQueue was added.

### Changes

- Use GrowableArrayBlockingQueue introduced in #153

### DIFF
--- a/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/collections/GrowableArrayBlockingQueue.java
+++ b/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/collections/GrowableArrayBlockingQueue.java
@@ -18,7 +18,7 @@
  * under the License.
  *
  */
-package org.apache.bookkeeper.util.collections;
+package org.apache.bookkeeper.common.collections;
 
 import java.util.AbstractQueue;
 import java.util.Collection;
@@ -29,7 +29,6 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 import java.util.concurrent.locks.Condition;
 import java.util.concurrent.locks.ReentrantLock;
-
 import org.apache.bookkeeper.util.MathUtils;
 
 
--- a/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/OrderedExecutor.java
+++ b/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/OrderedExecutor.java
@@ -46,6 +46,7 @@ import java.util.stream.Collectors;
 import lombok.extern.slf4j.Slf4j;
 
 import org.apache.bookkeeper.common.collections.BlockingMpscQueue;
+import org.apache.bookkeeper.common.collections.GrowableArrayBlockingQueue;
 import org.apache.bookkeeper.common.util.affinity.CpuAffinity;
 import org.apache.bookkeeper.stats.Gauge;
 import org.apache.bookkeeper.stats.NullStatsLogger;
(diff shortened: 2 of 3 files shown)


bookkeeper-common/src/main/java/org/apache/bookkeeper/common/collections/GrowableArrayBlockingQueue.java
```diff
@@ -83,6 +82,7 @@ public T poll() {
         try {
             if (SIZE_UPDATER.get(this) > 0) {
                 T item = data[headIndex.value];
+                data[headIndex.value] = null;
```
eolivelli, line 85: is there a test that cover this change ?

eolivelli (changes requested):
it looks like you are also fixing a bug in GrowableArrayBlockingQueue

is it possible to cover the fix with a test case please ?

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/collections/GrowableArrayBlockingQueue.java
lhotari: Pulsar contains GrowableArrayBlockingQueue too, https://github.com/apache/pulsar/blob/master/pulsar-common/src/main/java/org/apache/pulsar/common/util/collections/GrowableArrayBlockingQueue.java . I picked this fix to the BK version of GrowableArrayBlockingQueue when comparing the classes. 

I didn't see an explicit test.

lhotari: I'll look more into it.

lhotari: https://github.com/apache/pulsar/pull/1292/files#diff-013ecf9234e6aaa381510e81420c91664149af6d7f16866f1a36b5b3eb71aadcR87 is where the fix was made. It seems that there's not an explicit test for it.

lhotari:
> it looks like you are also fixing a bug in GrowableArrayBlockingQueue
> 
> is it possible to cover the fix with a test case please ?

Pulsar contains GrowableArrayBlockingQueue too, https://github.com/apache/pulsar/blob/master/pulsar-common/src/main/java/org/apache/pulsar/common/util/collections/GrowableArrayBlockingQueue.java . I picked this fix to the BK version of GrowableArrayBlockingQueue when comparing the classes. 
The Pulsar version contains also a few other features such as `remove` methods, `forEach` and `toList`. I didn't copy those over.

https://github.com/apache/pulsar/pull/1292/files#diff-013ecf9234e6aaa381510e81420c91664149af6d7f16866f1a36b5b3eb71aadcR87 is where the fix was made. It seems that there's not an explicit test for the issue that was fixed there.

I guess one way to test for it would be it inspect the internal state?

dlg99 (approved):
LGTM

PR #3109, opened 2022-03-14
Fix getMetadataServiceUri

gaozhangmin:
### Motivation
when config metadataStoreUri like `zk1:2181,zk2:2181/test`, `getMetadataServiceUri ` will only return  `zk1:2181`.  it's incorrect.

### Changes
Use `getList ` instead of `getString`

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
@@ -268,7 +268,11 @@ public abstract class AbstractConfiguration<T extends AbstractConfiguration>
      * @throws ConfigurationException if the metadata service uri is invalid.
      */
     public String getMetadataServiceUri() throws ConfigurationException {
-        String serviceUri = getString(METADATA_SERVICE_URI);
+        List servers = getList(METADATA_SERVICE_URI, null);
+        if (null == servers || 0 == servers.size()) {
+            return null;
+        }
+        String serviceUri = StringUtils.join(servers, ",");
         if (StringUtils.isBlank(serviceUri)) {
             // no service uri is defined, fallback to old settings
             String ledgerManagerType;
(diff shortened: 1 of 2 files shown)


eolivelli (changes requested):
this looks like a breaking change.

can you please add tests ?

eolivelli:
I believe you should use the semi colon to separate zk servers:
`zk1:2181;zk2:2181/test`

gaozhangmin:
> I believe you should use the semi colon to separate zk servers: `zk1:2181;zk2:2181/test`

`zk1:2181;zk2:2181/test` is a  incorrect connection string,  I tried. @eolivelli

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
```diff
@@ -268,7 +268,11 @@ public String getMetadataServiceUriUnchecked() throws UncheckedConfigurationExce
      * @throws ConfigurationException if the metadata service uri is invalid.
      */
     public String getMetadataServiceUri() throws ConfigurationException {
-        String serviceUri = getString(METADATA_SERVICE_URI);
+        List servers = getList(METADATA_SERVICE_URI, null);
```
eolivelli, line 271: This is not correct, METADATA_SERVICE_URI is a URI, it is not a list of servers

for instance it should start with zk://

bookkeeper-server/src/test/java/org/apache/bookkeeper/conf/AbstractConfigurationTest.java
```diff
@@ -56,6 +56,12 @@ public void setup() {
         this.conf.setZkLedgersRootPath("/path/to/ledgers");
     }
 
+    @Test
+    public void testSetGetServiceUri() throws Exception {
+        this.conf.setMetadataServiceUri("zk1:2181,zk2:2181/test");
```
eolivelli, line 61: this is not a valid URI

it should be something like `zk://zk1:2181,zk2:2181/test` or `zk+null://zk1:2181,zk2:2181/test`

eolivelli (changes requested):

bookkeeper-server/src/test/java/org/apache/bookkeeper/conf/AbstractConfigurationTest.java
gaozhangmin: https://github.com/apache/pulsar/blob/d02bd7378acd1f202d8bcf50aa6999749e61b20b/pulsar-broker/src/main/java/org/apache/pulsar/PulsarClusterMetadataSetup.java#L246-L248

PR #3110, opened 2022-03-15
PendingReadOp: Fix ledgerEntryImpl reuse problem

congbobo184:
In bookeeper client the ledgerEntryImpl wrongly recycle cause data parsing errors.

`org.apache.bookkeeper.client.PendingReadOp` have more `LedgerEntryRequest`

for example in PendingReadOp(1) has two LedgerEntryRequest (entryId = 1, entryId = 2)
another PendingReadOp(2) has two LedgerEntryRequest (entryId = 3, entryId = 4)

1. When `LedgerEntryRequest (entryId = 1)` read fail will close all `LedgerEntryRequest (entryId = 1, entryId = 2)` in `PendingReadOp(1)`.
2. `LedgerEntryImpl` in `LedgerEntryRequest (entryId = 2)` will be recycle.
3. `LedgerEntryRequest(entryId = 3`) will reuse this `LedgerEntryImpl` in another `PendingReadOp(2)` because the `LedgerEntryImpl` in `LedgerEntryRequest (entryId = 2)` has been recycle.
4. Another `LedgerEntryRequest(entryId = 3)` readComplete will set the ByteBuf to this `LedgerEntryImpl`
5. `LedgerEntryRequest(entryId = 2)` readComplete also use this `LedgerEntryImpl` and set the ByteBuf to this `LedgerEntryImpl`.
6. Now the `LedgerEntryImpl` the entryId is 3 but the ByteBuf is entryId = 2. 

### Motivation

fix this problem

### Changes

when `LedgerEntryRequest(entryId = 2)` reponse come can't change the `LedgerEntryImpl` ByteBuf

when LedgerEntryRequest change the complete to true, the response will not change the ByteBuf. because `LedgerEntryRequest(entryId = 2)` readComplete will check complete

the same Ledger read will in the same thread.


congbobo184:
rerun failure checks

lhotari:
Thanks @congbobo184 for making this PR.

I'm just wondering that `close` should get called once as a result of the guard by `org.apache.bookkeeper.client.PendingReadOp`'s `AtomicBoolean complete` field:

https://github.com/apache/bookkeeper/blob/10825c93bccd9e170df75ab9f2cf85cb6043f649/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java#L619-L622

Why isn't that sufficient?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
```diff
@@ -563,6 +565,10 @@ public long getLastAddConfirmed() {
         }
     }
 
+    public static ReadContext createReadContext(int bookieIndex, BookieId to, LedgerEntryRequest entry) {
```
codelipenghui, line 568: Don't need to be a public method? Looks only used for tests.

codelipenghui (commented):
Thanks for the great work.

The change looks good to me.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
hangc0276: Could we move this method into test?

hangc0276 (approved):
Great work! Thanks @congbobo184 
Overall looks good to me.

hangc0276:
> Thanks @congbobo184 for making this PR.
> 
> I'm just wondering that `close` should get called once as a result of the guard by `org.apache.bookkeeper.client.PendingReadOp`'s `AtomicBoolean complete` field:
> 
> https://github.com/apache/bookkeeper/blob/10825c93bccd9e170df75ab9f2cf85cb6043f649/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java#L619-L622
> 
> Why isn't that sufficient?

@lhotari  It changed the  request's `complete` flag from `false` to `true`. And when the previous request's response came back, whose request has already been closed, it will be filtered by this complete flag and prevent tampering other entry's data.

lhotari:
> > Thanks @congbobo184 for making this PR.
> > I'm just wondering that `close` should get called once as a result of the guard by `org.apache.bookkeeper.client.PendingReadOp`'s `AtomicBoolean complete` field:
> > https://github.com/apache/bookkeeper/blob/10825c93bccd9e170df75ab9f2cf85cb6043f649/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java#L619-L622
> > 
> > Why isn't that sufficient?
> 
> @lhotari It changed the request's `complete` flag from `false` to `true`. And when the previous request's response came back, whose request has already been closed, it will be filtered by this complete flag and prevent tampering other entry's data.

@hangc0276 That doesn't answer my question.

I meant a different thing.

It seems that LedgerEntryRequest.close can only get called via this line

https://github.com/apache/bookkeeper/blob/017307bc67431a7616861ad09927c4e3327633d1/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java#L644

or this line

https://github.com/apache/bookkeeper/blob/7087fda4e91e9fe8974f9d36e01e275a08bf38f1/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/impl/LedgerEntriesImpl.java#L59

both of those calls should only originate from the call chains starting from PendingReadOp#submitCallback:

https://github.com/apache/bookkeeper/blob/017307bc67431a7616861ad09927c4e3327633d1/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java#L611-L650

Why isn't the `if (!complete.compareAndSet(false, true)) {` sufficient in PendingReadOp#submitCallback?

hangc0276:
> Why isn't the `if (!complete.compareAndSet(false, true)) {` sufficient in PendingReadOp#submitCallback?

@lhotari  The `if (!complete.compareAndSet(false, true)) {` in PendingReadOp#submitCallback just change the complete flag in PendingReadOp instance. 

But the 
```
if (complete.compareAndSet(false, true)) {
                entryImpl.close();
            }
```
Change the `complete` flag in LedgerEntryRequest. They are different instances.

lhotari:
> Change the `complete` flag in LedgerEntryRequest. They are different instances.

yes, I have noticed that there's 2 separate `complete` fields: `org.apache.bookkeeper.client.PendingReadOp#complete` and `org.apache.bookkeeper.client.PendingReadOp.LedgerEntryRequest#complete`.

https://github.com/apache/bookkeeper/blob/017307bc67431a7616861ad09927c4e3327633d1/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java#L58-L84

That doesn't change my original question:  LedgerEntryRequest.close seems to only get called via call chains originating from PendingReadOp.submitCallback, why isn't the PendingReadOp's complete field preventing the duplicate calls to LedgerEntryRequest.close ?

congbobo184:
rerun failure checks

congbobo184:
rerun failure checks

hangc0276:
> close

@lhotari  
For example
1. there is a read request for ledger 1, entryId request range is [1, 100].  When entryId [1, 10] read requests has been sent out and entryId 1 's response come back. it will close all the following ReadEntryRequests, and `LedgerEntryImpl` object will be recycled. 

2. There is another read request for ledger 2, entryId request range is [1, 2]. It will get `LedgerEntryImpl` object, which may be the ref is the previous one. For example, ReadRequest 2:1 used the same `LedgerEntryImpl` object with ReadRequest 1:2. Then send entryId 1 request to the bookkeeper server

3.  Read request 2:1 response came back, and run into `readEntryComplete` callback. It will call `entry.complete(rctx.bookieIndex, rctx.to, buffer)` to put the buffer into entryImpl's EntryBuf. And then put it into seq list to wait for complete

4. Read request 1:2 response came back. It will run into `readEntryComplete` callback. It will call `entry.complete(rctx.bookieIndex, rctx.to, buffer)` to put the buffer into entryImpl's EntryBuf.  However, the entry has been hold by read request 2:1. **This operation will put entry (1:2) bytebuf into entry (2:1) bytebuf**.  when run into `submitCallback(BKException.Code.OK);` logic, it will be filtered by the complete flag of PendingReadOP.

5. When ledger 2 's read request [1, 2] complete and all the response came back, it will return the data to client.

6. However, the entry (2:1) 's data has been tampered by step 4. This is the root cause.

eolivelli:
@hangc0276 
in your last explanation your are not citing that there is a read error.
My understanding is that this problem may happen only when there is a read error, in RendingReadOP#submitCallback

see: 
https://github.com/apache/bookkeeper/blob/017307bc67431a7616861ad09927c4e3327633d1/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java#L644

in the happy path it is the client application that triggers the close of the LedgerEntryImpl

is this correct ?

hangc0276:
> @hangc0276 in your last explanation your are not citing that there is a read error. My understanding is that this problem may happen only when there is a read error, in RendingReadOP#submitCallback
> 
> see:
> 
> https://github.com/apache/bookkeeper/blob/017307bc67431a7616861ad09927c4e3327633d1/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java#L644
> 
> in the happy path it is the client application that triggers the close of the LedgerEntryImpl
> 
> is this correct ?

@eolivelli  Yes, that's right. I have updated the comments and added the cite.

eolivelli:
thank you @hangc0276 for your clarification.

eolivelli (dismissed):
LGTM

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
```diff
@@ -110,6 +110,7 @@
 
         @Override
         public void close() {
+            complete.set(true);
```
lhotari, line 113: wouldn't it be safer to use `if (complete.compareAndSet(false, true)) {` to shield against races?

lhotari (changes requested):

lhotari:
Thanks for the explanation @hangc0276 . I now understood how the race condition happens in `org.apache.bookkeeper.client.PendingReadOp#readEntryComplete` method and it answers my question.

To prevent the problem I think that the current code needs revisiting. Why did @congbobo184 make this change: 7a1504ff8ee6140b4f63a9a6d0c1cd3485cf5887 ?
I would assume that the close method would have to be as Congbo originally wrote
```java
        public void close() {
            if (complete.compareAndSet(false, true)) {
                entryImpl.close();
            }
        }
```

instead of the current version
```java
        public void close() {
            complete.set(true);
            entryImpl.close();
        }
```

lhotari:
I now see the reason why Congbo chose to implement it using
```java
        public void close() {
            complete.set(true);
            entryImpl.close();
        }
```

That seems to be since `entryImpl.close()` should be called when `close` is called regardless of whether the complete has been set to true.
The reason to set complete to true is to prevent using `entryImpl` after close has been called.
This feels like using `complete` field isn't the correct solution for this.

nicoloboschi (commented):
Good analysis. 
I think, as said by @lhotari, we should use the compareAndSet. If we don't think it could happen to call `close()` twice, we can raise an exception if the comparison fails - it probably means there's another race uncovered

dlg99 (changes requested):
1. `ListenerBasedPendingReadOp` is similar but does not close all entries in sequence on `submitCallback()`. I think we need to critically look at both. I didn't look at other ops.

2. Simply calling "close()" does not recycle entry's writeSet (memory leak). It's expected that the entry either fails or completes before close. This can be done either on close if the entry is not completed yet or, probably more explicit as

```
diff --git a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
index b48761ec6..18577a3de 100644
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
@@ -20,6 +20,7 @@
  */
 
+import com.google.common.base.Preconditions;
 import com.google.common.collect.Lists;
 import com.google.common.util.concurrent.ListenableFuture;
@@ -110,6 +111,7 @@ class PendingReadOp implements ReadEntryCallback, SafeRunnable {
 
         @Override
         public void close() {
+            Preconditions.checkArgument(isComplete(), "Closing non-completed entry");
             entryImpl.close();
         }
 
@@ -166,9 +168,15 @@ class PendingReadOp implements ReadEntryCallback, SafeRunnable {
          * @return true if we managed to fail the entry; otherwise return false if it already failed or completed.
          */
         boolean fail(int rc) {
+            return fail(rc, true);
+        }
+
+        boolean fail(int rc, boolean executeCallback) {
             if (complete.compareAndSet(false, true)) {
                 this.rc = rc;
-                submitCallback(rc);
+                if (executeCallback) {
+                    submitCallback(rc);
+                }
                 writeSet.recycle();
                 return true;
             } else {
@@ -641,7 +649,12 @@ class PendingReadOp implements ReadEntryCallback, SafeRunnable {
                     BKException.getMessage(code), firstUnread, firstRc);
             clientCtx.getClientStats().getReadOpLogger().registerFailedEvent(latencyNanos, TimeUnit.NANOSECONDS);
             // release the entries
-            seq.forEach(LedgerEntryRequest::close);
+            seq.forEach(entry -> {
+                // avoid calling the callback recursively
+                // fail() is re-entrant, won't do anything to already completed entries
+                entry.fail(code, false);
+                entry.close();
+            });
             future.completeExceptionally(BKException.create(code));
         } else {
             clientCtx.getClientStats().getReadOpLogger().registerSuccessfulEvent(latencyNanos, TimeUnit.NANOSECONDS);
```

eolivelli:
Just to clarify: I don't think that not recycling a object is a memory leak, as the jvm will discard it, but we are losing the chanceto save a allocation.

BTW I second @dlg99 's comments

congbobo184:
https://github.com/apache/bookkeeper/blob/e99cd4915b56ae3e9891ef9a83eb5e8f089505d7/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java#L146

complete will be change in this way and then close will not close ledgerEntryImpl if you use compareAndSet
@lhotari @nicoloboschi

congbobo184:
@dlg99 hi, 
1. although `ListenerBasedPendingReadOp` don't close all ledgerEntryRequset, we only change complete, it will not change any logical for `ListenerBasedPendingReadOp`, we can fix `ListenerBasedPendingReadOp` in next pr, not in this PR
2. it will not memory leak when don't not recycle entry's writeSet. now in this PR we only fix LedgerEntryImpl.close.

eolivelli:
I agree that we can follow up with a separate patch to ListenerBasedPendingReadOp

lhotari:
> https://github.com/apache/bookkeeper/blob/e99cd4915b56ae3e9891ef9a83eb5e8f089505d7/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java#L146
> 
> complete will be changed in this way and then close will not close ledgerEntryImpl if you use compareAndSet
> 
> https://github.com/apache/bookkeeper/blob/e99cd4915b56ae3e9891ef9a83eb5e8f089505d7/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java#L169
> 
> if read fail, in this way we already change complete, so close will do anything. @lhotari @nicoloboschi

Yes I understand that, I posted a follow up comment https://github.com/apache/bookkeeper/pull/3110#issuecomment-1068136249 . Would you mind adding a code comment that explains the intent to future readers of the code?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
lhotari: addressed by https://github.com/apache/bookkeeper/pull/3110#issuecomment-1068857090

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/TestParallelRead.java
```diff
@@ -238,4 +247,59 @@ public void testParallelReadFailureWithFailedBookies() throws Exception {
...
+        assertTrue(first.complete.get());
+
+        assertEquals(-1, second.entryImpl.getEntryId());
+        assertEquals(-1, second.entryImpl.getLedgerId());
+        assertEquals(-1, second.entryImpl.getLength());
+        assertNull(second.entryImpl.getEntryBuffer());
+        assertTrue(second.complete.get());
+
+        // Mock ledgerEntryImpl reuse
+        Method method = PendingReadOp.class.getDeclaredMethod("createReadContext",
```
nicoloboschi, line 289: what about making `ReadContext` class visibility to default and instantiate the object in the test ?

nicoloboschi (changes requested):
I re-read the scenario and I now understand why the compare-and-set would not work.
Good work!

lhotari (approved):

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/TestParallelRead.java
congbobo184: This PR is best not to change the ReadContext class visibility, keep the original level

congbobo184:
rerun failure checks

eolivelli:
@congbobo184 what about applying @dlg99 's patch to your branch ?

congbobo184:
@dlg99 I have add writeSet recycle logical, please review again, thanks

congbobo184:
rerun failure checks

congbobo184:
rerun failure checks

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
eolivelli: Now that we have the `writeSet.recycle();` block above we could go back to compareAndSet()

What about this ?

```
if (complete.compareAndSet(false, true)) {
   writeSet.recycle();
}
```

This way:
- we do writeSet.recycle(); only if complete is "false"
- eventually complete will be "true" in any case

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
```diff
@@ -580,18 +590,23 @@ void sendReadTo(int bookieIndex, BookieId to, LedgerEntryRequest entry) throws I
...
     @Override
     public void readEntryComplete(int rc, long ledgerId, final long entryId, final ByteBuf buffer, Object ctx) {
+
         final ReadContext rctx = (ReadContext) ctx;
         final LedgerEntryRequest entry = rctx.entry;
 
         if (rc != BKException.Code.OK) {
-            entry.logErrorAndReattemptRead(rctx.bookieIndex, rctx.to, "Error: " + BKException.getMessage(rc), rc);
+            // if entry complete = true, don't need logErrorAndReattemptRead
+            if (!entry.complete.get()) {
```
eolivelli, line 599: we should add a test covering this change: calling readEntryComplete with a non-OK error code twice

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
```diff
@@ -580,18 +590,23 @@ void sendReadTo(int bookieIndex, BookieId to, LedgerEntryRequest entry) throws I
...
+                entry.logErrorAndReattemptRead(rctx.bookieIndex, rctx.to, "Error: " + BKException.getMessage(rc), rc);
+            }
             return;
         }
 
         heardFromHosts.add(rctx.to);
         heardFromHostsBitSet.set(rctx.bookieIndex, true);
 
         buffer.retain();
+        // if entry has completed don't handle twice
```
eolivelli, line 609: do we have a test for this case ?
can you add this case to your new tests please ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
congbobo184:
```
            if (!complete.get()) {
                writeSet.recycle();
            }
            complete.set(true);
```
I think that would also work

congbobo184:
rerun failure checks

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
dlg99: @congbobo184 with this approach we risk calling recycle twice if close() is called from different threads. 
At least we need to implement @eolivelli 's suggestion. I'd set rc to something as well, eg:

```
if (complete.compareAndSet(false, true)) {
   rc = BKException.Code.UnexpectedConditionException;
   writeSet.recycle();
}
```

congbobo184:
rerun failure checks

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
eolivelli: @congbobo184 this is the last comment, then we are good to go (and also cut a release)

congbobo184:
rerun failure checks

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
```diff
@@ -110,6 +110,14 @@
...
         @Override
         public void close() {
+            // this request has succeeded before, can't recycle writeSet again
+            if (complete.compareAndSet(false, true)) {
+                rc = BKException.Code.UnexpectedConditionException;
+                writeSet.recycle();
+            }
+            // To be able to close this when complete = true, can't use compareAndSet.
+            // Because readComplete will make complete = true then close will not close LedgerEntryImpl
+            complete.set(true);
```
dlg99, line 120: this is not needed anymore

eolivelli (approved):
LGTM

dlg99 (approved):
Looks great!
Thank you.

congbobo184:
rerun failure checks

Issue #3112, opened 2022-03-16
Can't start bookie by bookkeeper-daemon.sh, because the last pid occupied by other progress’s thread


Nicklee007:
**BUG REPORT**

***Describe the bug***
After bookie progress is killed or occur some non-normal exit，can't start bookie by bin/bookkeeper-daemon.sh, because the last pid occupied by other progress’s thread.

***To Reproduce***

1. Go to directly kill the bookie;
2. Change the bin/bookkeeper-bookie.pid  to other progress’s one thread pid
3. bin/bookkeeper-daemon.sh start bookie
4. See error 'bookie running as process $PREVIOUS_PID.  Stop it first'

***Expected behavior***
we failed start broker when the pid in bin/bookkeeper-bookie.pid occupied by other progress, but we can start when the pid just occupied by the thread in other progress’s.

***Screenshots***
start broker by bin/pulsar-daemon script failed, some error 'bookie running as process 30210. Stop it first'
<img width="791" alt="image" src="https://user-images.githubusercontent.com/84127069/158438849-3e660a81-0904-4aca-adcf-176f7a2875db.png">

<img width="683" alt="image" src="https://user-images.githubusercontent.com/84127069/158438945-f5b0f861-8ab6-43fd-9539-ab082286cd0e.png">

but we don't find the progress which pid is 30210
<img width="602" alt="image" src="https://user-images.githubusercontent.com/84127069/158439075-421adf31-a9e3-4a14-9226-99026bc41586.png">

<img width="928" alt="image" src="https://user-images.githubusercontent.com/84127069/158439192-72887222-f929-4cad-891a-523b3e8bbb77.png">

The pid 30210 is occupied by the pid 30154 progress's one thread.
<img width="941" alt="image" src="https://user-images.githubusercontent.com/84127069/158439416-2a07a433-089b-4608-aff5-dfbb23aa09da.png">

PR #3113, opened 2022-03-16
Fix the pid occupied check when use bookkeeper-daemon.sh start or stop

Nicklee007:
Master Issue: #3112

### Motivation

Fix the failed pid occupied check. we'll fail when use bookkeeper-daemon.sh start or stop bookie, after the last time we exit the bookie direct kill or the bookie occurred non-normal exit, then the bin/bookkeeper-bookie.pid are retained, and the pid in the file is occupied by the thread in other progress.

### Changes
Change the pid occupied check from 'kill -0 $pid' to 'ps -p $pid'. The both will return true when the pid is occupied by one progress, but 'kill -0 $pid' will return true and the 'ps -p $pid' will return false when the pid is occupied by one progress's sub thread.

StackOverflow discussion:
https://stackoverflow.com/questions/30958964/why-do-kill-0-pid-echo-and-ps-ppid-echo-sometimes-differ

### DIFF
--- a/bin/bookkeeper-daemon.sh
+++ b/bin/bookkeeper-daemon.sh
@@ -112,7 +112,7 @@ start()
 {
   if [ -f $pid_file ]; then
     PREVIOUS_PID=$(cat $pid_file)
-    if kill -0 $PREVIOUS_PID > /dev/null 2>&1; then
+    if ps -p $PREVIOUS_PID > /dev/null 2>&1; then
       echo $command running as process $PREVIOUS_PID.  Stop it first.
       exit 1
     fi
@@ -125,7 +125,7 @@ start()
   echo $! > $pid_file
   sleep 1; head $out
   sleep 2;
-  if ! kill -0 $! > /dev/null ; then
+  if ! ps -p $! > /dev/null ; then
     exit 1
   fi
 }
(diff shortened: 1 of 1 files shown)


eolivelli:
on which platforms did you test this ?
Did you test this on Linux and MacOS ?

eolivelli:
Ci is not running for some reason
@nicoloboschi do you have any ideas ?

nicoloboschi:
perhaps: https://www.githubstatus.com/incidents/fpk08rxnqjz2

can you try to close and reopen the PR ?

Nicklee007:
> on which platforms did you test this ? Did you test this on Linux and MacOS ?

@eolivelli I test the case in  on Linux and MacOS. On Linux the  'kill -0 $pid' and 'ps -p $pid'  performance diff as the discussion (https://stackoverflow.com/questions/30958964/why-do-kill-0-pid-echo-and-ps-ppid-echo-sometimes-differ);On MacOs, the both exhibit the same result;

StevenLuMT (changes requested):
it look very nice 
but I have a question:
how to verify the correctness of this shell change? there is no relevant testcase for the shell, right? @Nicklee007 @eolivelli

StevenLuMT:

shoothzj:
@Nicklee007 Could you please rebase the code? I will push on this forward. Thanks :)

shoothzj (approved):

StevenLuMT (approved):
it look very nice
but I have a question:
how to verify the correctness of this shell change? there is no relevant testcase for the shell, right?

shoothzj:
@StevenLuMT Currently we don't have shell test suite, and it's difficult to built. So I think manual testing is enough

PR #3133, opened 2022-03-22
[feature] [distributedlog]  Add a api for delete empty zookeeper node

### LINKED ISSUE #3132: [feature] [distributedlog] The api for delete empty zookeeper node
**FEATURE REQUEST**

````
// Def a log
DistributedLogManager logManager = namespace.openLog(“default/example/v0.1”);
// Write 
DLOutputStream.openWriterAsync( logManager )
.thenCompose(dlOutputStream -> dlOutputStream.writeAsync(inputStream)).get();

// Delte it 
namespace.getNamespaceDriver()
    .getLogMetadataStore()
    .getLogLocation(“default/example/v0.1”)
    .thenCompose(uri -> uri.map(value -> namespace.getNamespaceDriver()
                                                                                     .getLogStreamMetadataStore(NamespaceDriver.Role.WRITER)
                                                                                     .deleteLog(value, path)
    );
````

After deleted, there has an dirty zk-node:  default/example

I think there should has a safe way to remove nodes, just like this: 

````
/**
 * Delete the metadata of a log, this method will not delete any child node,
 * When {@param streamName} is not empty, delete nothing.
 *
 * @param uri the location to store the metadata of the log
 * @param streamName the name of the log stream
 * @return When delete finish return {@link Boolean#TRUE};
 *         when {@param streamName} not exists return {@link Boolean#TRUE}
 *         when delete nothing return {@link Boolean#FALSE}
 */
CompletableFuture<Boolean> deleteLogUnRecursive(URI uri, final String streamName);
````

### PR DESCRIPTION
**Pull request by poorbarcode**

Fixes #3132 

Master Issue: #3132 

### Motivation

Provide another API for delete log path.

Why not use `CompletableFuture<Void> deleteLog(URI uri, String streamName)` ? 
Because this method will delete node recursively.  If there's a case like this: 

````
// Thead-1
create log "default/example/v0.1"

// Thdead-2
delete log "default"

// Now thead-1's create-operate be removed
````

### Changes

Add a new API like this : 
` CompletableFuture<Boolean> deleteLogUnRecursive(URI uri, final String streamName)`

### DIFF
--- a/stream/distributedlog/core/src/main/java/org/apache/distributedlog/impl/metadata/ZKLogStreamMetadataStore.java
+++ b/stream/distributedlog/core/src/main/java/org/apache/distributedlog/impl/metadata/ZKLogStreamMetadataStore.java
@@ -680,6 +680,40 @@ public class ZKLogStreamMetadataStore implements LogStreamMetadataStore {
         return promise;
     }
 
+    @Override
+    public CompletableFuture<Boolean> deleteLogUnRecursive(URI uri, final String logName) {
+        final CompletableFuture<Boolean> promise = new CompletableFuture<Boolean>();
+        try {
+            String streamPath = LogMetadata.getLogStreamPath(uri, logName);
+            zooKeeperClient.get().delete(streamPath, -1, new AsyncCallback.VoidCallback() {
+                @Override
+                public void processResult(int rc, String path, Object ctx) {
+                    if (KeeperException.Code.OK.intValue() == rc){
+                        FutureUtils.complete(promise, true);
+                    } else if (KeeperException.Code.NONODE.intValue() == rc){
+                        FutureUtils.complete(promise, true);
+                    } else if (KeeperException.Code.NOTEMPTY.intValue() == rc ) {
+                        FutureUtils.complete(promise, false);
+                    }else {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: stream/distributedlog/core/src/main/java/org/apache/distributedlog/impl/metadata/ZKLogStreamMetadataStore.java [hunk 1]
```diff
@@ -680,6 +680,40 @@ public void processResult(int rc, String path, Object ctx) {
         return promise;
     }
 
+    @Override
+    public CompletableFuture<Boolean> deleteLogUnRecursive(URI uri, final String logName) {
```
zymap, line 684: the distributed log will create a lot of metadata paths under the log, you can check the getLog method. Looks like the deleteLogUnRecursive will fail because we only delete the log path?

poorbarcode: > the distributed log will create a lot of metadata paths under the log, you can check the getLog method. Looks like the deleteLogUnRecursive will fail because we only delete the log path?

ok

poorbarcode:
Can't just do this, need more detailed planning

Issue #3129, opened 2022-03-21
Gradle Build: bad file bookkeeper-4.15.0-SNAPSHOT-src-bin.tar.gz instead of bookkeeper-4.15.0-SNAPSHOT-src.tar.gz


eolivelli:
When I build BookKeeper from current master I see that the bookkeeper-dist project generates a file `bookkeeper-4.15.0-SNAPSHOT-src-bin.tar.gz` I would expect `bookkeeper-4.15.0-SNAPSHOT-src.tar.gz`

shoothzj:
gradle is removed now, close the issue.

PR #3135, opened 2022-03-24
Fix bad file bookkeeper-4.15.0-SNAPSHOT-src-bin.tar.gz instead of bookkeeper-4.15.0-SNAPSHOT-src.tar.gz

mattisonchao:
### Motivation

See #3129 

### Changes

Master Issue:#3129 

### Modification

- Remove ``archiveClassifier = "bin"``


build.gradle
```diff
@@ -230,7 +230,6 @@ allprojects {
                     dependsOn(':bookkeeper-common:jar')
                     dependsOn(':bookkeeper-common-allocator:jar')
                 }
-                archiveClassifier = "bin"
```
nicoloboschi, line 233: the "bin" classifier is needed for the bin tarball
it's the "src" tarball that doesn't need it

nicoloboschi (changes requested):

PR #3137, opened 2022-03-24
fix duplicate typeline for prometheus type

### LINKED ISSUE #3136: second TYPE line for metric name when use metricbeat prometheus module
***Describe the bug***
second TYPE line for metric name when use metricbeat prometheus module

Shoothzj:
Fixes #3136 

### Motivation

fix duplicate typeline for prometheus type

### Changes

remember the prometheus type, record in set, incase print again

Master Issue: #3136

### DIFF
--- a/bookkeeper-stats-providers/prometheus-metrics-provider/src/main/java/org/apache/bookkeeper/stats/prometheus/PrometheusMetricsProvider.java
+++ b/bookkeeper-stats-providers/prometheus-metrics-provider/src/main/java/org/apache/bookkeeper/stats/prometheus/PrometheusMetricsProvider.java
@@ -172,12 +172,13 @@ public class PrometheusMetricsProvider implements StatsProvider {
 
     @Override
     public void writeAllMetrics(Writer writer) throws IOException {
-        PrometheusTextFormatUtil.writeMetricsCollectedByPrometheusClient(writer, registry);
+        PrometheusTextFormat prometheusTextFormat = new PrometheusTextFormat();
+        PrometheusTextFormat.writeMetricsCollectedByPrometheusClient(writer, registry);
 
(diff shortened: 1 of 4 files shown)


Shoothzj:
@eolivelli @dlg99 PTAL

dlg99 (approved):

bookkeeper-stats-providers/prometheus-metrics-provider/src/main/java/org/apache/bookkeeper/stats/prometheus/PrometheusMetricsProvider.java
```diff
@@ -172,6 +172,7 @@ public StatsLogger getStatsLogger(String scope) {
 
     @Override
     public void writeAllMetrics(Writer writer) throws IOException {
+        PrometheusTextFormatUtil.reset();
```
eolivelli, line 175: This is a sign of something going wrong.

What happens with two concurrent calls to this method?

eolivelli (changes requested):

bookkeeper-stats-providers/prometheus-metrics-provider/src/main/java/org/apache/bookkeeper/stats/prometheus/PrometheusMetricsProvider.java
Shoothzj: @eolivelli you are right, we don't know what will happen. Let me push a fix.

Shoothzj:
@eolivelli PTAL again thanks, I refractor `PrometheusTextFormatUtil` to `PrometheusTextFormat` to make it can serve two concurrent calls.

For an alternate approach, I can add HashSet params to every method in `PrometheusTextFormatUtil`. I think that the former is better.

eolivelli (approved):
Lgtm

PR #3145, opened 2022-03-27
Upgrade ZooKeeper to 3.8.0

eolivelli:
### Motivation

We are using an old version of ZooKeeper, 3.6.x (and ZooKeeper 3.5.x was declared EOL a few weeks ago).
We should upgrade to the latest version.

### Changes

- Upgraded ZooKeeper dependency.
- Exclude LogBack
- Add junit5 jupiter-api required by org.apache.zookeeper.test.ClientBase

### Notes
In PR #3084 I tried to update ZooKeeper in Gradle files, but without success.
This patch focuses only on Maven

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -29,7 +29,7 @@ depVersions = [
     commonsCli: "1.2",
     commonsCodec: "1.6",
     commonsCollections4: "4.1",
-    commonsCompress: "1.19",
+    commonsCompress: "1.21",
     commonsConfiguration: "1.10",
     commonsIO: "2.7",
     commonsLang2: "2.6",
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -86,7 +86,7 @@ depVersions = [
     testcontainers: "1.16.3",
     vertx: "3.9.8",
     yahooDatasketches: "0.8.3",
-    zookeeper: "3.6.2",
+    zookeeper: "3.8.0"
 ]
 
 depLibs = [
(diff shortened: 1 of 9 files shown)


nicoloboschi (changes requested):
Can you add owasp suppressions?

eolivelli:
> Can you add owasp suppressions?

sure, I was just working on that

dlg99 (approved):

dependencies.gradle
```diff
@@ -29,7 +29,7 @@ depVersions = [
     commonsCli: "1.2",
     commonsCodec: "1.6",
     commonsCollections4: "4.1",
-    commonsCompress: "1.21",
+    commonsCompress: "1.19",
```
nicoloboschi, line 32: why this one? 
1.19 is vulnerable
we have 1.21 in Maven pom

nicoloboschi (changes requested):
good job! I left one comment about commons-compress downgrade

eolivelli:
I have temporary cherry-picked these patches to unlock CI
- https://github.com/apache/bookkeeper/pull/3148
- https://github.com/apache/bookkeeper/pull/3149

it it better to at least commit #3149, then I will rebase this patch

eolivelli:
the failed test passes locally on my laptop
```
 org.apache.bookkeeper.test.BookieZKExpireTest.testBookieServerZKExpireBehaviour
Error:    Run 1: BookieZKExpireTest.testBookieServerZKExpireBehaviour:103 Bookie should not shutdown on losing zk session
Error:    Run 2: BookieZKExpireTest.testBookieServerZKExpireBehaviour:103 Bookie should not shutdown on losing zk session
```

dependencies.gradle
eolivelli: good catch. I picked up a old version of the file from master branch

PR #3146, opened 2022-03-27
Recycle LedgerEntriesImpl after call back.

horizonzy:

Recycle LedgerEntriesImpl after call back.


eolivelli (changes requested):
we cannot do this.
The user is responsible for calling `close`.
We cannot assume that the object won't be used after the (successful) execution of the callback

PR #3148, opened 2022-03-28
Add Log4j2 configuration in order to see logs in tests

eolivelli:


nicoloboschi (changes requested):
How about other modules?

dlg99:
can we do something like adding `-Dlog4j.configuration=file://....` into MAVEN_OPTS in gh workflows? https://github.com/apache/bookkeeper/blob/e4a2b54240d608763e8f9acd6ffb91740dfd3f10/.github/workflows/client-tests.yml#L34

eolivelli:
> How about other modules?
We previously had log4j configuration in bookkeeper-server, this is the same thing basically.

Usually people work on bookkeeper-server.
I am not sure it is good to put files everywhere if no one needs them

eolivelli:
> can we do something like adding `-Dlog4j.configuration=file://....` into MAVEN_OPTS in gh workflows?
> 
> https://github.com/apache/bookkeeper/blob/e4a2b54240d608763e8f9acd6ffb91740dfd3f10/.github/workflows/client-tests.yml#L34

@dlg99 I am adding this file in order to see the logs in the IDE and not on CI.
Probably there is a smarted way to put the file only once.

In Pulsar we have a log4j2.xml in buildtools.
maybe we can use the same approach

dlg99:
@eolivelli There are more problems than just tests:

1. shell scripts that use log4j.properties (I only checked bin/ directory):

```
$ ggrep -rI log4j.properties                                                                                                                                             [11:55:13]
common.sh:DEFAULT_LOG_CONF=${BK_CONFDIR}/log4j.properties
dlog:DEFAULT_LOG_CONF=${BK_HOME}/conf/log4j.properties
common_gradle.sh:DEFAULT_LOG_CONF=${BK_CONFDIR}/log4j.properties
```

2. tests

there is metadata-drivers/etcd/src/test/resources/log4j2.xml - I think we can just copy it everywhere we see log4j.properties in test resources: 
```
$ find . | grep log4j.properties | grep -v /build/ | grep -v /target/ | grep  /resources | grep /test/  

./tools/ledger/src/test/resources/log4j.properties
./bookkeeper-benchmark/src/test/resources/log4j.properties
./stream/clients/java/kv/src/test/resources/log4j.properties
./stream/clients/java/all/src/test/resources/log4j.properties
./stream/clients/java/base/src/test/resources/log4j.properties
./stream/statelib/src/test/resources/log4j.properties
./stream/proto/src/test/resources/log4j.properties
./stream/distributedlog/core/src/test/resources/log4j.properties
./stream/distributedlog/protocol/src/test/resources/log4j.properties
./stream/server/src/test/resources/log4j.properties
./stream/storage/impl/src/test/resources/log4j.properties
./bookkeeper-common/src/test/resources/log4j.properties
./tests/integration/cluster/src/test/resources/log4j.properties
./tests/integration/standalone/src/test/resources/log4j.properties
./circe-checksum/src/test/resources/log4j.properties
./bookkeeper-server/src/test/resources/log4j.properties
./metadata-drivers/etcd/src/test/resources/log4j.properties
```

3. Default log4j config that can be used in prod (bookkeeper-server/src/main/resources/log4j.properties)

these need review, possibly same approach works:
```
 $ find . | grep log4j.properties | grep -v /build/ | grep -v /target/ | grep  /resources | grep -v /test/  
 
./buildtools/src/main/resources/log4j.properties
./tests/integration-tests-utils/src/main/resources/log4j.properties
./bookkeeper-server/src/main/resources/log4j.properties
```

4. Prod configs (or their examples)

These require extra work:
```
$ find . | grep log4j.properties | grep -v /build/ | grep -v /target/ | grep -v /resources

./bookkeeper-benchmark/conf/log4j.properties
./stream/conf/log4j.properties
./tests/docker-images/statestore-image/conf/log4j.properties
./conf/log4j.properties
```

This makes it a release blocker for 4.15

5. Docs (haven't checked)

Shoothzj:
@eolivelli I think this pr can be closed by https://github.com/apache/bookkeeper/pull/3208

PR #3158, opened 2022-03-30
support ipv6 address validation

wangjialing218:
### Motivation

current bookkeeper URI does not support ipv6 address validate.

### Changes

Support ipv6 address validate. mainly copy codes from https://github.com/apache/pulsar/pull/8120#issue-707796479


bookkeeper-common/src/main/java/org/apache/bookkeeper/common/net/ServiceURI.java
```diff
@@ -145,6 +146,39 @@ public class ServiceURI {
...
+                        .map(host -> validateHostName(serviceURI.getServiceName(), host))
+                        .collect(Collectors.toList());
+                return new ServiceURI(
+                        serviceURI.getServiceName(),
+                        serviceURI.getServiceInfos(),
+                        serviceURI.getServiceUser(),
+                        multiHosts.toArray(new String[multiHosts.size()]),
+                        serviceURI.getServicePath(),
+                        serviceURI.getUri());
+            }
```
hangc0276, line 179: Do we need to deal with hosts.size() == 1 ?

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/net/ServiceURI.java
```diff
@@ -215,21 +249,21 @@ public static ServiceURI create(URI uri) {
     }
 
     private static String validateHostName(String serviceName, String hostname) {
-        String[] parts = hostname.split(":");
-        if (parts.length >= 3) {
+        URI uri = null;
+        try {
+            uri = URI.create("dummyscheme://" + hostname);
```
StevenLuMT, line 254: `uri = URI.create("dummyscheme://" + hostname);`
I think testcase doesn't cover this

StevenLuMT (changes requested):
updating is not very versatile
I think you can add a new class named ServiceIpv6URI

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/net/ServiceURI.java
wangjialing218: No, `URI.create(uriStr)` could handle this case, `testIPv6Address()` cover this test case

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/net/ServiceURI.java
wangjialing218: In fact any one of testcase in ServiceURITest could cover this,  `ServiceURI#create` call `ServiceURI#validateHostName` to validate the format of host name.

bookkeeper-common/src/test/java/org/apache/bookkeeper/common/net/ServiceURITest.java
```diff
@@ -206,6 +206,30 @@ public void testServiceInfoPlus() {
             "/path/to/namespace");
     }
 
+    @Test
```
zymap, line 209: I think it's better to test the IPV6 with all the current test cases.

wangjialing218: Done

StevenLuMT:

hangc0276:
@horizonzy @zymap Please help take a look, thanks.

hangc0276:
@wangjialing218 Would you please help rebase the master? thanks.

jiazhai:
Thanks for the help. Do we need some documents for this?

horizonzy (commented):
In org.apache.bookkeeper.net.DNS#reverseDns, it can't handle ipv6, also need to fix it.

Shoothzj (approved):

wangjialing218:
> 



> In org.apache.bookkeeper.net.DNS#reverseDns, it can't handle ipv6, also need to fix it.

@horizonzy `reverseDns` only execute when `advertisedAddress` is not set. In this case, the ipv6 address of the host will be ignored. Let's keep this behave unchanged, bookie server will advertise ipv4 address when `advertisedAddress` is not set.

To enable advertise ipv6 address for bookie server, we should set `advertisedAddress` with ipv6 address of the host.

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/net/ServiceURI.java
```diff
@@ -145,6 +146,39 @@ public class ServiceURI {
     public static ServiceURI create(String uriStr) {
         checkNotNull(uriStr, "service uri string is null");
 
+        if (uriStr.contains("[") && uriStr.contains("]")) {
```
horizonzy, line 149: Which case does the code cover?

horizonzy (commented):
Which case does the code cover?

zymap:
Close this since it is open for a long time without any updates. Feel free to reopen it if you want to continue

PR #3160, opened 2022-03-30
[build] Fix various spotbugs warnings 

nicoloboschi:
### Motivation
If you run spotbugs with JDK8 it fail with many errors.

Spotbugs with jdk11 is disabled in maven and it explains why it doesn't fail in CI
In gradle we set the reportLevel to HIGH so we haven't seen the Medium level for a while

In branch-4.14 there's still a job with a spotbugs enabled and it fails

### Changes
* Fix spotbugs warnings
   * Mostly suppressions due to https://github.com/spotbugs/spotbugs/issues/756
   * RxJava usage - call dispose of `Disposable`
   * Better synchronization on `LedgerFragmentReplicator`
* Restore spotbugs check with jdk8 and jdk11 on CI 

We'll need to cherry-pick to branch-4.14 to fix the compilation

PR #3167, opened 2022-03-31
[branch-4.14] Replace Log4J with Reload4J

nicoloboschi:
### Motivation

Log4J is dead and it has a lot of vulnerabilities. We already switched to Log4J2 on master branch.
The easiest way to replace it is to use [reload4j](https://reload4j.qos.ch/)

### Changes
* Update org.slf4j:log4j-over-slf4j to 1.7.36 which automatically replace the jar with reload4j replacement

From the official [website](https://www.slf4j.org/news.html)
>2022-01-25 - Release of SLF4J 1.7.35
• In this release, the "slf4j-log4j12" artifact automatically instructs Maven to use the "slf4j-reload4j" artifact instead. As you might have guessed, the "slf4j-reload4j" binding delegates log processing to the reload4j logging framework.


nicoloboschi:
rerun failure checks

pom.xml
```diff
@@ -160,7 +160,7 @@
     <reflections.version>0.9.11</reflections.version>
     <rocksdb.version>6.29.4.1</rocksdb.version>
     <shrinkwrap.version>3.0.1</shrinkwrap.version>
-    <slf4j.version>1.7.32</slf4j.version>
+    <slf4j.version>1.7.36</slf4j.version>
```
dlg99, line 163: that got me confused for a sec, why there is no new dependency added.
In case someone has the same question, this is because of 
> The SLF4J project has released version 1.7.35 containing the slf4j-reload4j module which supports reload4j. Moreover, the "slf4j-log4j12" artifact contains a Maven relocation directive for the "slf4j-reload4j" artifact.

https://reload4j.qos.ch/

dlg99 (approved):

dlg99:
there are couple of failed CI jobs. One feels like a flake, I restarted it. Another one might need rebase on top of the spotbugs fixes:

```
[INFO] Error size is 0
[INFO] Total bugs: 1
Error:  Exception is caught when Exception is not thrown in org.apache.bookkeeper.proto.checksum.DirectMemoryCRC32Digest.<static initializer for DirectMemoryCRC32Digest>() [org.apache.bookkeeper.proto.checksum.DirectMemoryCRC32Digest] At DirectMemoryCRC32Digest.java:[line 86] REC_CATCH_EXCEPTION
[INFO] 


To see bug detail using the Spotbugs GUI, use the following command "mvn spotbugs:gui"
```

pom.xml
nicoloboschi: it was in the pull description 😢

nicoloboschi:
spotbugs won't pass until we merge the fix https://github.com/apache/bookkeeper/pull/3160
also the lib changes should not affect spotbugs anyway

eolivelli (approved):
LGTM

but we can do this change only after a discussion on dev@

This is because:
- it is a non trivial library upgrade
- reload4j is a fork of Log4j1 but it is not exactly the same thing)

I believe it is enough to drop a message to dev@ and if no one objects we can move forward and merge this patch

eolivelli:
we should fix spotbugs errors on JDK8, CI is not passing (this deserves a separate patch)

nicoloboschi:
>we should fix spotbugs errors on JDK8, CI is not passing (this deserves a separate patch)

PR for spotbugs on branch-4.14: https://github.com/apache/bookkeeper/pull/3170

PR #3172, opened 2022-04-01
[netty] remove no longer used properties io.netty.recycler.linkCapacity and io.netty.recycler.maxCapacity.default

nicoloboschi:
### Motivation

Those two options doesn't exist anymore in Netty.

See https://github.com/netty/netty/commit/9fe97103 which renamed the property io.netty.recycler.maxCapacity.default to io.netty.recycler.maxCapacity in 2014 for netty-4.1.0

The io.netty.recycler.linkCapacity system property was removed in Netty 4.1.71.Final
by commit https://github.com/netty/netty/commit/40196a63

### Changes
* Removed those options without replacement in order to not change the current behavior


nicoloboschi:
rerun failure checks

Shoothzj (approved):

merlimat (approved):

PR #3173, opened 2022-04-02
Expose Netty threads num config to bk_server.conf

hangc0276:
### Motivation
https://github.com/apache/bookkeeper/pull/1612 this PR make the Netty io threads number configurable, but doesn't expose to the `bk_server.conf`. This parameter will be usually changed in performance tuning especially deploy this bookie on bare metal which have many cpus. For operation engineer who is unfamiliar with the source code will be hard to do performance tuning.

### Changes
Expose this parameter configuration into `bk_server.conf` and website. And add more explanation about it.


hangc0276:
rerun failure checks

nicoloboschi (approved):

Issue #2943, opened 2021-12-15
BP-44: Support for writing ledger entries with O_DIRECT


mauricebarnum:
**BP**

> Follow the instructions at http://bookkeeper.apache.org/community/bookkeeper_proposals/ to create a proposal.

This is the master ticket for tracking BP-44 :

Add support for writing ledger entries, bypassing the file system's buffering.  On supported systems (currently Linux and MacOS), files are opened with O_DIRECT and page-aligned buffers are maintained by the BookKeeper application.  Access to the operating system's IO API is via a thin JNI binding.

Proposal PR - https://github.com/apache/bookkeeper/pull/2932

Vanlightly:
This will need to be BP-47. BP-44 - 46 are not yet merged but do exist.

hangc0276:
@mauricebarnum Would you please give more details about the motivation and design for this proposal?

hangc0276:
### Motivation
#### Ledger read/write logic
When the BookKeeper server receives a write entry request, it will write the entry into the memory table, which is a bookie-level cache. After the memory table is full, it will sort and trigger a flush into the operating system PageCache. The operating system PageCache will buffer that data again. When the PageCache flush is triggered, the data will be flushed to disk.

When the BookKeeper server receives a read entry request, it will check whether the memory table and read cache contains the target entry. If both caches are missed, it will query rocksDB to get the entry's location in the entry log file, and then read the target entry from the entry log file. After reading the entry, the read cache will pre-read more entries from the entry log files to ensure the following read keeps high cache hit rate. From the operating system perspective, when reading a log file from a specific position, it will check whether the target data has been cached in PageCache. If PageCache hits, it will be directly returned from PageCache. Otherwise, it will read the block data from the log file and cache the block data into PageCache. The block may contain multiple entries located near the target read position.


#### Drawbacks
For ledger writes, it will limit the write throughput for the following reasons.
1. The memtable and OS PageCache will double buffer entry data, which will consume more memory. 
2. The flush mechanism of PageCache is controlled by the kernel and it's hard to tune by application, which is very important for IO intensive applications. 
3. The number of kernel sync threads is limited by the number of disks, which is not conducive to RAID composed of multiple disks.

For ledger reads, it will also limit the read throughput for the following reasons.
[...]

PR #3192, opened 2022-04-08
Fix compaction throttle imprecise

hangc0276:
### Motivation
When do compaction for an entry log file, it will scan each entry header metadata from beginning to the end of the file. When an entry doesn't deleted, it will read the entry data and write into a new entry log file. The throttle only works on the entry which doesn't deleted.
https://github.com/apache/bookkeeper/blob/b0030ab5d9132e5c5cd5092b8e6dab79e9fbf16c/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java#L1012-L1051

For those deleted entries, the throttle strategy doesn't applied. However, we use `BufferedChannel` to read the entry log file which will prefetch data from file when the read buffer missed.
https://github.com/apache/bookkeeper/blob/b0030ab5d9132e5c5cd5092b8e6dab79e9fbf16c/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedReadChannel.java#L91-L97

For an entry log file, more than 90% of entries has been deleted, the compactor will scan those entry's header metadata one by one. When reading one entry's metadata, it will missed in BufferedChannel read buffer, it will trigger prefetch from disk. For the following entries, the header metadata reading will also missed in BufferedChannel read buffer, and will continue to prefetch from disk without throttle. It will lead to ledger disk IO util runs high. 

Moreover, for each prefetch operation from disk, it will also trigger OS PageCache prefetch. For the compaction model, the OS PageCache prefetch will lead to PageCache pollution，and maybe also affect the journal sync latency. For this one, we can use the Direct IO to reduce the PageCache effect. https://github.com/apache/bookkeeper/issues/2943

### Changes
Add throttle on entry header metadata check stage.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedReadChannel.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedReadChannel.java
@@ -73,14 +73,15 @@ public class BufferedReadChannel extends BufferedChannelBase  {
      * from the file channel, return the prefetched bytes for throttle.
      * @param dest
      * @param pos
-     * @return Pair<Integer, Integer>
+     * @return
      * @throws IOException
      */
     public Pair<Integer, Integer> readWithPrefetchedBytes(ByteBuf dest, long pos) throws IOException {
         return readWithPrefetchedBytes(dest, pos, dest.writableBytes());
     }
 
-    public synchronized Pair<Integer, Integer> readWithPrefetchedBytes(ByteBuf dest, long pos, int length) throws IOException {
+    public synchronized Pair<Integer, Integer> readWithPrefetchedBytes(ByteBuf dest, long pos, int length)
+        throws IOException {
         invocationCount++;
         long currentPosition = pos;
         long eof = validateAndGetFileChannel().size();
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedReadChannel.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedReadChannel.java
@@ -63,25 +63,11 @@ public class BufferedReadChannel extends BufferedChannelBase  {
      *         -1 if the given position is greater than or equal to the file's current size.
      * @throws IOException if I/O error occurs
      */
-    public int read(ByteBuf dest, long pos) throws IOException {
+    public Pair<Integer, Integer> read(ByteBuf dest, long pos) throws IOException {
         return read(dest, pos, dest.writableBytes());
     }
 
-    /**
-     * Read as many bytes into dest as dest.capacity() starting at position pos in the
-     * FileChannel. This function can read from the buffer or the file channel. If it fetched
-     * from the file channel, return the prefetched bytes for throttle.
-     * @param dest
-     * @param pos
-     * @return
-     * @throws IOException
-     */
-    public Pair<Integer, Integer> readWithPrefetchedBytes(ByteBuf dest, long pos) throws IOException {
-        return readWithPrefetchedBytes(dest, pos, dest.writableBytes());
-    }
-
-    public synchronized Pair<Integer, Integer> readWithPrefetchedBytes(ByteBuf dest, long pos, int length)
-        throws IOException {
+    public synchronized Pair<Integer, Integer> read(ByteBuf dest, long pos, int length) throws IOException {
         invocationCount++;
         long currentPosition = pos;
         long eof = validateAndGetFileChannel().size();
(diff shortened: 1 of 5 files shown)


hangc0276:
rerun failure checks

hangc0276:
rerun failure checks

hangc0276:
rerun failure checks

hangc0276:
rerun failure checks

hangc0276:
rerun failure checks

zymap (commented):
Overall looks good. Do we need to add a test to make sure the throttler work as expected?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -1014,10 +1041,18 @@ public void scanEntryLog(long entryLogId, EntryLogScanner scanner) throws IOExce
...
+
+                Pair<Integer, Integer> readBytesPair =
+                    readFromLogChannelWithPrefetchBytes(entryLogId, bc, headerBuffer, pos);
+                if (readBytesPair.getLeft() != headerBuffer.capacity()) {
                     LOG.warn("Short read for entry size from entrylog {}", entryLogId);
                     return;
                 }
+
+                if (throttler != null) {
+                    throttler.acquire(readBytesPair.getRight());
```
zymap, line 1053: When do we release it?

hangc0276: The release operation controlled by `RateLimiter` library in guava

zymap (approved):

zymap:
@hangc0276 Looks like the compaction tests failed. Could you please take a look?

hangc0276:
> @hangc0276 Looks like the compaction tests failed. Could you please take a look?

Ok, I will take a look today.

hangc0276:
rerun failure checks

eolivelli (changes requested):
Good work

We definitely need tests about these new methods

hangc0276:
> Good work
> 
> We definitely need tests about these new methods

@eolivelli  I removed the duplicated new methods, and return the prefetched bytes when calling `BufferedReadChannel.read()`. So we do not need extra unit tests.

hangc0276:
rerun failure checks

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedReadChannel.java
```diff
@@ -95,9 +98,10 @@ public synchronized int read(ByteBuf dest, long pos, int length) throws IOExcept
                     throw new IOException("Reading from filechannel returned a non-positive value. Short read.");
                 }
                 readBuffer.writerIndex(readBytes);
+                bytesFetchedFromFile += readBytes;
             }
         }
-        return (int) (currentPosition - pos);
+        return new ImmutablePair<> ((int) (currentPosition - pos), bytesFetchedFromFile);
```
dlg99, line 104: This is a very frequently called code, can we avoid creating a new object?
I think the throttler can be safely passed down the call chain into this method.

dlg99 (changes requested):
Couple of things:
1. one note regarding new object creation (increased load on Java GC)
2. +1 to @eolivelli 's ask for the test. This is not done for the sake of refactoring. As you said, for the deleted entries the throttle strategy isn't applied. Let's add a test that shows that, fails without the fix, and passes with the fix.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedReadChannel.java
hangc0276: Good catch. I'm considering how to avoiding creating new object. 
The `BufferedReadChannel.read` method is synchronized, we should use throttle in the read method to block it when throttled, otherwise, it will block other threading calling.  I still have no good idea about it.

StevenLuMT: @hangc0276 @dlg99 
I think you can write like this simple code :
`return Pair.of((int) (currentPosition - pos), bytesFetchedFromFile); `
instead of 
`return new ImmutablePair<> ((int) (currentPosition - pos), bytesFetchedFromFile); `

but I think it's ok, frequent creation of ImmutablePair has no affect 
image

Shoothzj (approved):

StevenLuMT:

hangc0276:
The compaction introduced performance issue has been addressed by https://github.com/apache/bookkeeper/pull/3959, close this PR.

Issue #3193, opened 2022-04-09
BP-50: Add journal reuse feature to support Intel Pmem disk


hangc0276:
**BP**

> Follow the instructions at http://bookkeeper.apache.org/community/bookkeeper_proposals/ to create a proposal.

This is the master ticket for tracking BP-50 :

### Motivation
This PR https://github.com/apache/bookkeeper/pull/2742 has been introduced the `FileChannelProvider` interface to support different file provider. The default is `DefaultFileChannelProvider` using standard fileSystem to store journal file.

For [Intel Optane Persistent Memory](https://www.intel.com/content/www/us/en/architecture-and-technology/optane-dc-persistent-memory.html) disk (Pmem), it can use the Pmem disk matched library [PMDK](https://pmem.io/pmdk/) to provide higher throughput and lower latency than using standard fileSystem.

For Pmem disk based on PMDK library, it will pre-write the whole file first, and then write data into the file. **That means it will write the Pmem disk twice for each file stored on the Pmem disk**.  We have tested on the Pmem disk, and the result is as expected.

There are two options to fix this issue. 
 - Re-compile the linux kernel to turn off pre-write feature.
 - Provide a file pool to reuse the files stored on the Pmem disk on application side.

Consider the read/write model of BookKeeper journal disk, we recommend to add the reuse feature for journal to support Intel Pmem disk to achieve high throughput.

### Proposal

We provide a flag `journalReuseFiles` in conf/bk_server.conf to control whether turn the reuse journal file feature. The flag default is `false`. The journal file pool size is controlled by `journalMaxBackups`. When we need a log file for writing, we first check if there are enough log files in the log file pool, if not, we will create a new one, otherwise, we will choose an old log file to overwrite.

There is a problem for the relationship between journal file name and the logId. Current implementation is keep the journal file name sync with logId. The logId will be stored in journal mark file through checkpoint.

When we enable journal file reuse feature, we should separate the journal file name with the logId.  The mapping strategy is `logFileName = Long.toHexString(logId % journalMaxBackups)`

When we replay the journal file on bookie startup, it will get the start journal log file according to markPosition stored in journalLastMark, and replay the following journal files. The risk is that it will replay some old journal files, but doesn't affect the data correctness.

Proposal PR - https://github.com/apache/bookkeeper/pull/3194

PR #3194, opened 2022-04-09
BP-50: Add reuse journal file feature to support Intel PMem disk

hangc0276:
### Motivation
The implementation of #3193

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/JournalChannel.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/JournalChannel.java
@@ -47,8 +47,8 @@ class JournalChannel implements Closeable {
     final BookieFileChannel channel;
     final int fd;
     final FileChannel fc;
-    final BufferedChannel bc;
     final int formatVersion;
+    BufferedChannel bc;
     long nextPrealloc = 0;
 
     final byte[] magicWord = "BKLG".getBytes(UTF_8);
@@ -177,8 +177,8 @@ class JournalChannel implements Closeable {
         }
 
         LOG.info("Opening journal {}", fn);
-        if (!channel.fileExists(fn) || reuseFile) { // new file, write version
-            if (!reuseFile && !fn.createNewFile()) {
+        if (!channel.fileExists(fn)) { // create new journal file to write, write version
+            if (!fn.createNewFile()) {
                 LOG.error("Journal file {}, that shouldn't exist, already exists. "
                           + " is there another bookie process running?", fn);
                 throw new IOException("File " + fn
@@ -186,21 +186,12 @@ class JournalChannel implements Closeable {
             }
             fc = channel.getFileChannel();
             formatVersion = formatVersionToWrite;
-
-            int headerSize = (V4 == formatVersion) ? VERSION_HEADER_SIZE : HEADER_SIZE;
-            ByteBuffer bb = ByteBuffer.allocate(headerSize);
-            ZeroBuffer.put(bb);
-            bb.clear();
-            bb.put(magicWord);
-            bb.putInt(formatVersion);
-            bb.clear();
-            fc.write(bb);
-
-            bc = bcBuilder.create(fc, writeBufferSize);
-            forceWrite(true);
-            nextPrealloc = this.preAllocSize;
-            fc.write(zeros, nextPrealloc - journalAlignSize);
-        } else {  // open an existing file
+            writeHeader(bcBuilder, writeBufferSize);
+        } else if (reuseFile) { // Open an existing journal to write, it needs fileChannelProvider support reuse file.
+            fc = channel.getFileChannel();
+            formatVersion = formatVersionToWrite;
+            writeHeader(bcBuilder, writeBufferSize);
+        } else {  // open an existing file to read.
             fc = channel.getFileChannel();
             bc = null; // readonly
 
@@ -255,6 +246,23 @@ class JournalChannel implements Closeable {
         } else {
             this.fd = -1;
         }
+    }
+
+    private void writeHeader(Journal.BufferedChannelBuilder bcBuilder,
+                             int writeBufferSize) throws IOException {
+        int headerSize = (V4 == formatVersion) ? VERSION_HEADER_SIZE : HEADER_SIZE;
+        ByteBuffer bb = ByteBuffer.allocate(headerSize);
+        ZeroBuffer.put(bb);
+        bb.clear();
+        bb.put(magicWord);
+        bb.putInt(formatVersion);
+        bb.clear();
+        fc.write(bb);
+
+        bc = bcBuilder.create(fc, writeBufferSize);
+        forceWrite(true);
+        nextPrealloc = this.preAllocSize;
+        fc.write(zeros, nextPrealloc - journalAlignSize);
     }
 
     public static void renameJournalFile(File source, File target) throws IOException {
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/JournalChannel.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/JournalChannel.java
@@ -258,7 +258,7 @@ class JournalChannel implements Closeable {
     }
 
     public static void renameJournalFile(File source, File target) throws IOException {
-        if (!source.renameTo(target)) {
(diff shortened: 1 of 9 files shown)


dlg99 (changes requested):
Please add tests, at least to:
* make sure journals aren't reused when feature is disabled.
* cover recovery from journal/after crash with reused journal. is it possible to use wrong journal file in that case? or end up with corrupted one? What happens with last log mark and checkpointing? e.g. see BookieJournalTest. 
* Bookie's readJournal()/replay() scan journals sequentially because Journal.listJournalIds assumes they are. With this option the assumption changes to circular with unclear stop point.

hangc0276:
> when the bookie is restarted with replaying the journal, how to handle the corner cases below: 1)bookie crash with reused journal **enabled**, but the restart bookie with the flag **disabled**; 2)bookie crash with reused journal **disabled**, but the restart bookie with the flag **enabled**.

@ArvinDevel Thanks for your review.

Currently, we reuse the journal files by renaming the file name. 
For example, the journal file pool size is 5, and current journal files are: [101, 102, 103, 104, 105], and the current mark deletion position is (103, 2000). We need a new journal file named 106, and if we enabled the reuse journal file, we will do the following check.
1) The journal files list is: [101, 102, 103, 104, 105], and the mark deletion position point to file: 103, it is safe to reuse file: 101
2) We pick up file: 101 and rename it to file: 106, after rename, the journal files list is: [102, 103, 104, 105, 106]

For the first case.
>  1)bookie crash with reused journal **enabled**, but the restart bookie with the flag **disabled**

Bookie crash with reused journal enabled, the last written journal file is 106, and the last mark delete position is (103, 2000). When the bookie restarts with the flag disabled, it will replay from file: 103 to file 106.

> 2) bookie crash with reused journal **disabled**, but the restart bookie with the flag **enabled**

Bookie crash with reuse journal disabled, the last written journal file is 106, and the last mark delete position is (103, 2000). When the bookie restarts with the flag enabled, it will replay from file 103 to file 106 and for new file 107 to be written, it does the above check, renames file 101 to file 107, and open it for writing.

hangc0276:
> Please add tests, at least to:
> 
> * make sure journals aren't reused when feature is disabled.
> * cover recovery from journal/after crash with reused journal. is it possible to use wrong journal file in that case? or end up with corrupted one? What happens with last log mark and checkpointing? e.g. see BookieJournalTest.
> * Bookie's readJournal()/replay() scan journals sequentially because Journal.listJournalIds assumes they are. With this option the assumption changes to circular with unclear stop point.

@dlg99 Thanks for your review.
Please refer to: https://github.com/apache/bookkeeper/pull/3194#issuecomment-1189912416

For the test, we need a special FileChannel which records the current written position when rewriting an existing journal file. I'm thinking about how to add the test.

> * cover recovery from journal/after crash with reused journal. is it possible to use wrong journal file in that case? or end up with corrupted one? What happens with last log mark and checkpointing? e.g. see BookieJournalTest.

We rename the old journal file to the new one and keep the journal file name increasing as the current behavior. It can avoid complex cases such as using the wrong journal file and so on.

> * Bookie's readJournal()/replay() scan journals sequentially because Journal.listJournalIds assumes they are. With this option the assumption changes to circular with unclear stop point.

We rename the old journal file to the new one, we can file the stop point with the largest file name on the journal replay stage.

bookkeeper-server/src/test/java/org/apache/bookkeeper/test/BookieJournalRollingWithReuseJournalTest.java
```diff
@@ -0,0 +1,83 @@
...
+        // Write enough ledger entries so that we roll over journals
+        LedgerHandle[] lhs = writeLedgerEntries(10, 1024, 1024);
+        long[] ledgerIds = new long[lhs.length];
+        for (int i = 0; i < lhs.length; i++) {
+            ledgerIds[i] = lhs[i].getId();
+            lhs[i].close();
+        }
+
+        // Sleep for a while to ensure data are flushed
+        Thread.sleep(2000);
```
eolivelli, line 62: is there a better way ? like using Awayatility ?

eolivelli (approved):
LGTM

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/JournalChannel.java
```diff
@@ -177,7 +190,21 @@ private JournalChannel(File journalDirectory, long logId,
...
+        if (fRemoveFromPageCache) {
+            this.fd = NativeIO.getSysFileDescriptor(channel.getFD());
+        } else {
+            this.fd = -1;
+        }
+    }
 
+    public void writeHeader() throws IOException {
+        try {
+            bc = bcBuilder.create(fc, writeBufferSize);
```
ArvinDevel, line 207: the `bc` is already created at line 193, right? if so, I think this line is redundant.

ArvinDevel (changes requested):
@hangc0276  thanks for your contribution at first. I left one small comment, please check it. From the change, I can see that the main changes is that extract `writeHeader`  and allow  to writeHeader to existing file if it exists and `skipHeader` when scanning the journal file to replace current `fileExists` trigger, correct me if I'm wrong.

merlimat (approved):

hangc0276:
> @hangc0276 thanks for your contribution at first. I left one small comment, please check it. From the change, I can see that the main changes is that extract `writeHeader` and allow to writeHeader to existing file if it exists and `skipHeader` when scanning the journal file to replace current `fileExists` trigger, correct me if I'm wrong.

@ArvinDevel  Thanks for your review. Yes, you are right. The original logic is a little complex and it has bugs. I rewrite the logic, Please help take a look again. Thanks a lot.

dlg99 (approved):
Looks good!

zymap (approved):

ArvinDevel (changes requested):
@hangc0276 thanks for your clarification, please add the failure cases(which are not solved by previous version)  in the test to ensure the refactored code works as expected.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/JournalChannel.java
```diff
@@ -153,23 +153,32 @@ private JournalChannel(File journalDirectory, long logId,
...
 
         if (formatVersionToWrite < V4) {
             throw new IOException("Invalid journal format to write : version = " + formatVersionToWrite);
         }
 
         LOG.info("Opening journal {}", fn);
-        if (!channel.fileExists(fn)) { // new file, write version
-            if (!fn.createNewFile()) {
+        if (!channel.fileExists(fn) || reuseFile) { // new file, write version
+            if (!reuseFile && !fn.createNewFile()) {
```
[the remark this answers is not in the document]
hangc0276: @ArvinDevel  Thanks for your review.

>when the reuseFile is true, we have done renameJournalFile(toReplaceFile, fn), so we expect the channel.fileExists(fn), if not then throw exception, is this alert what you want to ensure?

No, there are two code paths to run into line 180. 
1) For normal create a new journal file, the `channel.fileExists(fn)` is `false` and `reuseFile` flag is `false`. It will call `createNewFile` and then write the header into the new journal file.
2) For open an existed journal file to write, it will rename the journal file first. The `channel.fileExists(fn)` is `true` and `reuseFile` is true. It will run into line 181, due to `reuseFile` being `true`, so it won't call `fn.createNewFile` to create a new journal file. In the end, it will write the new header into the journal file to override existing data in the file.

For the test, we need a special FileChannel which records the current written position when rewriting an existing journal file. It is a little hard to run into a reuseFile case.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/JournalChannel.java
```diff
@@ -248,6 +257,13 @@ private JournalChannel(File journalDirectory, long logId,
         }
     }
 
+    public static void renameJournalFile(File source, File target) throws IOException {
+        if (!source.renameTo(target)) {
```
StevenLuMT, line 261: I  think you should check source's value like:
`if (source == null) {
            throw new NullPointerException();
        }`

StevenLuMT:
rerun failure checks

ArvinDevel (approved):
It's pretty clear now. Thanks for your excellent work. @hangc0276 
LGTM!

PR #3196, opened 2022-04-10
Simplify throttle code in metadata scanning

hangc0276:
### Motivation
The throttler has been checked whether throttle by bytes in `acquire` method, we doesn't need double check.

### Changes

Remove the double check of whether throttle by bytes.


Shoothzj (approved):

hangc0276:
rerun failure checks

zymap (approved):

PR #3202, opened 2022-04-13
remove log4j1 from bk test code

Shoothzj:
### Motivation
this is a part work of remove log4j1 from bookeeper, first remove it from our test code, unit test and integrate test

### Changes
- remove log4j1 scope test dependency
- use log4j2.xml instread of log4j.properties


Shoothzj:
@eolivelli @dlg99 @nicoloboschi @merlimat @pkumar-singh PTAL

nicoloboschi (changes requested):
I think a cleaner solution would be to have a module with the log configuration file and make all the other modules from it, what do you think @Shoothzj

Shoothzj:
> I think a cleaner solution would be to have a module with the log configuration file and make all the other modules from it, what do you think @Shoothzj 

you mean only test configuration? what'f modules need characteristic configuration? to override it?

nicoloboschi:
yes, I mean the log configuration file. we can have that module with the `log4j2.xml` file. Then if a module needs particular configuration (very unlikely) we can add a `log4j2-test.xml` file which has precedence over log4j2.xml one. 
(see the docs [here](https://logging.apache.org/log4j/2.x/manual/configuration.html))
>Log4j will inspect the "log4j2.configurationFile" system property and, if set, will attempt to load the configuration using the ConfigurationFactory that matches the file extension. Note that this is not restricted to a location on the local file system and may contain a URL.
If no system property is set the properties ConfigurationFactory will look for log4j2-test.properties in the classpath.
If no such file is found the YAML ConfigurationFactory will look for log4j2-test.yaml or log4j2-test.yml in the classpath.
If no such file is found the JSON ConfigurationFactory will look for log4j2-test.json or log4j2-test.jsn in the classpath.
If no such file is found the XML ConfigurationFactory will look for log4j2-test.xml in the classpath.
If a test file cannot be located the properties ConfigurationFactory will look for log4j2.properties on the classpath.
If a properties file cannot be located the YAML ConfigurationFactory will look for log4j2.yaml or log4j2.yml on the classpath.
If a YAML file cannot be located the JSON ConfigurationFactory will look for log4j2.json or log4j2.jsn on the classpath.
If a JSON file cannot be located the XML ConfigurationFactory will try to locate log4j2.xml on the classpath.

PR #3204, opened 2022-04-13
Fix master broken ci due to zookeeper CVE fp

Shoothzj:
### Changes
- bump the dependency check version, I think it don't worth to open another pr.
- add cve suppression


Shoothzj:
rerun failure checks

eolivelli (approved):
LGTM

PR #3207, opened 2022-04-15
Fix testCompactionPersistence do not cleanup created directory

hangc0276:
### Motivation
In `CompactionTest.testCompactionPersistence()`, it specified a directory as rocksdb storage path, but does't clean up it when the test finished.
https://github.com/apache/bookkeeper/blob/e937ae9a6838fd533c3973d6ae6c32d98b6619bd/bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/CompactionTest.java#L1011-L1012

### Changes
Use tmpDir to manage the created directory and clean up the resources when test finished.


hangc0276:
rerun failure checks

eolivelli (approved):
Great catch

PR #3212, opened 2022-04-18
Added flag to control whether to transition to read-only mode when any disks full

### LINKED ISSUE #3141: auto read only mode support multi disks
**FEATURE REQUEST**

when diskUsageThreshold is full, the bookie will convert to read-only 

```


readOnlyModeEnabled | If all ledger directories configured are full, then support only read requests for clients. If “readOnlyModeEnabled=true” then on all ledger disks full, bookie will be converted to read-only mode and serve only read requests. Otherwise the bookie will be shutdown. By default this will be disabled.
-- | --



diskUsageThreshold | For each ledger dir, maximum disk space which can be used. Default is 0.95f. i.e. 95% of disk can be used at most after which nothing will be written to that partition. If all ledger dir partions are full, then bookie will turn to readonly mode if ‘readOnlyModeEnabled=true’ is set, else it will shutdown. Valid values should be in between 0 and 1 (exclusive).
-- | --


```

now, if we have one disk, ledger dirs  are l1,l2,l3,l4 it runs ok

but if we have two disks, one for l1,l2, the other for l3,l4, because of the difference of every entry, after run a long time, 

the usage of disk1 will be 95%, but the other maybe 80%, then `DiskChecker` will think `If all ledger dir partions are full`  this is not true, they will write more date to disk, then disk 1 will be 100%, and goes down.

2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have).
   Are you currently using any workarounds to address this issue?

should-have

3. Provide any additional detail on your proposed use case for this feature.

maybe the usage of any of the disk be 0.95, we can translate to read-only, or we can translate disk1 to read only, but disk2 can be written

hangc0276:
### Motivation
Fix #3141 

If we configure multi ledger disks, the bookie will transition to read-only mode when all disks are reached max usage threshold. However, if just one ledger disk reached max usage threshold, the bookie will continue to receive write requests, which will lead to the ledger disk out of space.

### Changes
Add a flag to control whether to transition to read-only mode when any disks reaches max usage threshold, and when all disks recovered to less than max usage threshold, the bookie will transition to read-write mode.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDirsMonitor.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDirsMonitor.java
@@ -68,9 +68,9 @@ class LedgerDirsMonitor {
 
     private void check(final LedgerDirsManager ldm) {
         final ConcurrentMap<File, Float> diskUsages = ldm.getDiskUsages();
-        boolean anyDiskFulled = false;
+        boolean someDiskFulled = false;
         boolean highPriorityWritesAllowed = true;
-        boolean anyDiskRecovered = false;
+        boolean someDiskRecovered = false;
 
         try {
             List<File> writableDirs = ldm.getWritableLedgerDirs();
@@ -103,7 +103,7 @@ class LedgerDirsMonitor {
                     });
                     // Notify disk full to all listeners
                     ldm.addToFilledDirs(dir);
-                    anyDiskFulled = true;
+                    someDiskFulled = true;
                 }
             }
             // Let's get NoWritableLedgerDirException without waiting for the next iteration
@@ -150,7 +150,7 @@ class LedgerDirsMonitor {
                     if (makeWritable) {
                         ldm.addToWritableDirs(dir, true);
                     }
-                    anyDiskRecovered = true;
+                    someDiskRecovered = true;
                 } catch (DiskErrorException e) {
                     // Notify disk failure to all the listeners
                     for (LedgerDirsListener listener : ldm.getListeners()) {
@@ -162,7 +162,7 @@ class LedgerDirsMonitor {
                     if (makeWritable) {
                         ldm.addToWritableDirs(dir, false);
                     }
-                    anyDiskRecovered = true;
+                    someDiskRecovered = true;
                 } catch (DiskOutOfSpaceException e) {
                     // the full-filled dir is still full-filled
                     diskUsages.put(dir, e.getUsage());
@@ -176,14 +176,14 @@ class LedgerDirsMonitor {
         }
 
         if (conf.isReadOnlyModeOnAnyDiskFullEnabled()) {
-            if (anyDiskFulled && !ldm.getFullFilledLedgerDirs().isEmpty()) {
+            if (someDiskFulled && !ldm.getFullFilledLedgerDirs().isEmpty()) {
                 // notify any disk full.
                 for (LedgerDirsListener listener : ldm.getListeners()) {
                     listener.anyDiskFull(highPriorityWritesAllowed);
                 }
             }
 
-            if (anyDiskRecovered && ldm.getFullFilledLedgerDirs().isEmpty()) {
+            if (someDiskRecovered && ldm.getFullFilledLedgerDirs().isEmpty()) {
                 // notify all disk recovered.
                 for (LedgerDirsListener listener : ldm.getListeners()) {
                     listener.allDisksWritable();
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDirsMonitor.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDirsMonitor.java
@@ -113,7 +113,6 @@ class LedgerDirsMonitor {
             ldm.getWritableLedgerDirs();
         } catch (NoWritableLedgerDirException e) {
             LOG.warn("LedgerDirsMonitor check process: All ledger directories are non writable");
-            anyDiskFulled = true;
             try {
                 // disk check can be frequent, so disable 'loggingNoWritable' to avoid log flooding.
                 ldm.getDirsAboveUsableThresholdSize(minUsableSizeForHighPriorityWrites, false);
@@ -163,6 +162,7 @@ class LedgerDirsMonitor {
                     if (makeWritable) {
                         ldm.addToWritableDirs(dir, false);
                     }
+                    anyDiskRecovered = true;
                 } catch (DiskOutOfSpaceException e) {
                     // the full-filled dir is still full-filled
                     diskUsages.put(dir, e.getUsage());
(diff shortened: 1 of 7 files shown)


hangc0276:
rerun failure checks

hangc0276:
ping @eolivelli @dlg99 @Vanlightly @nicoloboschi  Please help take a look, thanks a lot.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDirsMonitor.java
```diff
@@ -68,6 +68,10 @@ public LedgerDirsMonitor(final ServerConfiguration conf,
 
     private void check(final LedgerDirsManager ldm) {
         final ConcurrentMap<File, Float> diskUsages = ldm.getDiskUsages();
+        boolean anyDiskFulled = false;
```
eolivelli, line 71: What about 'SomeDiskFull'

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDirsMonitor.java
```diff
@@ -68,6 +68,10 @@ public LedgerDirsMonitor(final ServerConfiguration conf,
...
+        boolean highPriorityWritesAllowed = true;
+        boolean anyDiskRecovered = false;
```
eolivelli, line 73: someDiskRecovered sounds better

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
```diff
@@ -2395,6 +2396,27 @@ public boolean isReadOnlyModeEnabled() {
...
+        return this;
+    }
+
+    /**
+     * Get whether read-only mode is enable when any disk is full. The default is false.
+     *
+     * @return boolean
+     */
+    public boolean isReadOnlyModeOnAnyDiskFullEnabled() {
+        return getBoolean(READ_ONLY_MODE_ON_ANY_DISK_FULL_ENABLED, false);
```
eolivelli, line 2417: In this case it is better to set this to true.
It is a safe feature and if we don't enable it by default nobody will leverage it

I wonder if we need a flag at all.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/LedgerDirsManagerTest.java
```diff
@@ -240,6 +240,76 @@ private void testLedgerDirsMonitorDuringTransition(boolean highPriorityWritesAll
...
+        usageMap.put(curDir1, 0.1f);
+        usageMap.put(curDir2, 0.1f);
+        mockDiskChecker.setUsageMap(usageMap);
+        ledgerMonitor.init();
+        final MockLedgerDirsListener mockLedgerDirsListener = new MockLedgerDirsListener();
+        dirsManager.addLedgerDirsListener(mockLedgerDirsListener);
+        ledgerMonitor.start();
+
+        Thread.sleep((diskCheckInterval * 2) + 100);
+        assertFalse(mockLedgerDirsListener.readOnly);
```
eolivelli, line 284: Can we use awaitility here?

eolivelli (changes requested):
Very good

I left some feedback

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDirsMonitor.java
hangc0276: make sense

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
hangc0276: If we enable is feature, it changed the current behavior. I wonder if we make the default false first, and send an email to dev mail list to discuss it. If approved, I use another PR to change the default value.

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/LedgerDirsManagerTest.java
hangc0276: In BookKeeper code, there are a lot of `Thread.sleep`. I will use another PR to introduce awaitility to make all tests get ride of `Thread.sleep`

hangc0276:
rerun failure checks

dlg99:
I'll look at the change a little bit later, but the first thing I noticed is that the new config option is not added to/not documented at the `conf/bk_server.conf` and not added to the docs.

hangc0276:
> I'll look at the change a little bit later, but the first thing I noticed is that the new config option is not added to/not documented at the `conf/bk_server.conf` and not added to the docs.

IMO, we'd better to keep `conf/bk_server.conf` simple and just include import config. For advanced users who want to change chose config item, they can get the description from `ServerConfiguration.java`.  Just like Apache Pulsar, we put all the config item into the `conf/broker.conf`, leading to nearly 1k lines of the configuration file. For common users, they will be confused.

dlg99:
Overall: LGTM.

I still insist on updating config file with description of the feature + the docs.

@hangc0276 I hear your point about the length of the file but there are other things to consider:
* On many occasions doc were out of sync and config works a s a more up-to-date version of the doc
* config is easy to diff against an existing prod config to check if there are any new features one wants to try, how the current one is different form defaults etc. Diffing code is a bit more painful, especially considering that here is ServerConfiguration, AbstractConfiguration etc
* new user is unlikely to start digging into the code to look for the config options
* currently community agreement, as I remember, is to add new options and their description into the config file. You can start a discussion on ML to challenge this but we'll need to come up with a better solution for options discoverability than forcing users to dig into the code.

hangc0276:
> Overall: LGTM.
> 
> I still insist on updating config file with description of the feature + the docs.
> 
> @hangc0276 I hear your point about the length of the file but there are other things to consider:
> 
> * On many occasions doc were out of sync and config works a s a more up-to-date version of the doc
> * config is easy to diff against an existing prod config to check if there are any new features one wants to try, how the current one is different form defaults etc. Diffing code is a bit more painful, especially considering that here is ServerConfiguration, AbstractConfiguration etc
> * new user is unlikely to start digging into the code to look for the config options
> * currently community agreement, as I remember, is to add new options and their description into the config file. You can start a discussion on ML to challenge this but we'll need to come up with a better solution for options discoverability than forcing users to dig into the code.

@dlg99  Thanks for your clarification. I will add config item in `conf/bookkeeper.conf` and add doc into website. And then I will start a discussion on mail list about the configuration issue that I concerned.

hangc0276:
> Overall: LGTM.
> 
> I still insist on updating config file with description of the feature + the docs.
> 
> @hangc0276 I hear your point about the length of the file but there are other things to consider:
> 
> * On many occasions doc were out of sync and config works a s a more up-to-date version of the doc
> * config is easy to diff against an existing prod config to check if there are any new features one wants to try, how the current one is different form defaults etc. Diffing code is a bit more painful, especially considering that here is ServerConfiguration, AbstractConfiguration etc
> * new user is unlikely to start digging into the code to look for the config options
> * currently community agreement, as I remember, is to add new options and their description into the config file. You can start a discussion on ML to challenge this but we'll need to come up with a better solution for options discoverability than forcing users to dig into the code.

@dlg99  I have update the doc, Please help take a look, thanks a lot. I wonder if we can include this Pr into 4.15.0 release.

hangc0276:
rerun failure checks

hangc0276:
rerun failure checks

zymap (approved):

merlimat (approved):

leizhiyuan:
thanks for your work

PR #3216, opened 2022-04-19
Bump jackson databind version to 2.13.2.2

Shoothzj:
### Changes
- jackson bom will release a version when everytime its components releases, so we don't need to define variables like `jackson-data.version`
- bump jackson databind version to 2.13.2.2


Shoothzj:
rerun failure checks

eolivelli (approved):
LGTM

PR #3234, opened 2022-04-21
Enable Netty and BookKeeper IO optimizations on jdk17

nicoloboschi:
### Motivation

Starting from JDK17 the reflection is not allowed in the jdk internals modules. Until JDK16, it will print a warning like this:
```
Error:  WARNING: An illegal reflective access operation has occurred
Error:  WARNING: Illegal reflective access by org.apache.bookkeeper.proto.checksum.DirectMemoryCRC32Digest (file:/home/runner/work/bookkeeper/bookkeeper/bookkeeper-server/target/classes/) to method java.util.zip.CRC32.updateByteBuffer(int,long,int,int)
Error:  WARNING: Please consider reporting this to the maintainers of org.apache.bookkeeper.proto.checksum.DirectMemoryCRC32Digest
Error:  WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
Error:  WARNING: All illegal access operations will be denied in a future release
```

With JDK17, BookKeeper and Netty work well even if this optimization are not allowed. In order to continue to use them, we need to open the required modules

For Netty, we can follow this issue: https://github.com/netty/netty/issues/12265
For BookKeeper, I identified DirectMemoryCRC32Digest and NativeIO classes

### Changes

* Add the open modules options to open the required module (only if jdk9+ because jdk8 doesn't know the `--add-opens` instruction

### DIFF
--- a/bin/common.sh
+++ b/bin/common.sh
@@ -284,14 +284,13 @@ build_cli_jvm_opts() {
 }
 
 build_netty_opts() {
-  NETTY_OPTS="-Dio.netty.leakDetectionLevel=${NETTY_LEAK_DETECTION_LEVEL}"
-  echo "-Dio.netty.leakDetectionLevel=${NETTY_LEAK_DETECTION_LEVEL}"
-
+  NETTY_OPTS="-Dio.netty.leakDetectionLevel=${NETTY_LEAK_DETECTION_LEVEL} -Dio.netty.tryReflectionSetAccessible=true"
   # --add-opens does not exist on jdk8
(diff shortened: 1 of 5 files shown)


nicoloboschi:
rerun failure checks

Shoothzj (approved):

dlg99 (approved):

Shoothzj:
image
image
I suggest that we'd better keep the script consistent.

merlimat (approved):

lordcheng10 (approved):
LGTM

Issue #3220, opened 2022-04-19
Autorecovery does not process underreplicated empty ledgers


dlg99:
**BUG REPORT**

BK 4.14.3

bookie1 is decommissioned.

`bin/bookkeeper shell listunderreplicated -printmissingreplica`

``` 
2022-04-19T17:17:14,704 [main] INFO  org.apache.bookkeeper.tools.cli.commands.autorecovery.ListUnderReplicatedCommand - 8691131535
2022-04-19T17:17:14,704 [main] INFO  org.apache.bookkeeper.tools.cli.commands.autorecovery.ListUnderReplicatedCommand - 	Ctime : 1650388634626
2022-04-19T17:17:14,705 [main] INFO  org.apache.bookkeeper.tools.cli.commands.autorecovery.ListUnderReplicatedCommand - 	MissingReplica : bookie1:3181
```

```
org.apache.bookkeeper.tools.cli.commands.client.LedgerMetaDataCommand - LedgerMetadata{formatVersion=2, ensembleSize=3, writeQuorumSize=3, ackQuorumSize=2, state=CLOSED, length=0, lastEntryId=-1, digestType=CRC32, password=base64:, ensembles={0=[bookie3:3181, bookie2:3181, bookie1:3181]}, customMetadata={}}
```

Note `state=CLOSED, length=0, lastEntryId=-1`

Autorecovery does not delete it, does not update metadata.

***Additional context***

From prod.

PR #3239, opened 2022-04-22
ISSUE 3220: Autorecovery does not process underreplicated empty ledgers

dlg99:

### Motivation

Currently decomm malfunctions on an empty ledger:

Autorecovery ends up removing such ledger from the list of underreplicated ledgers but does not update the ensemble leaving the bookie being decommed there and Auditor will add these ledger back on the next run. 

In greater details:

Currently the ReplicationWorker/LedgerChecker end up skipping fragment where ES > WQ and failed bookie is not in the writeset (or a ledger is empty). 

E.g. with ES = 3, WQ = 2
Ensemble: (bk1, bk2, bk3)
bk3 is dead.
if ledger is empty (and e.g. closed) it is added to the list of underreplicated ledgers but autorecovery skips it (no failed fragments, dead bookie remains in the metadata) and simply removes it from UR ledgers (until next Auditor's run adds it back).

This becomes a problem when a few bookies with large number of empty ledgers get decommed. 
Every run of the Auditor ends up in increase of znodes (added UR ledgers) and a lot of busy work for the autorecovery (overall increased load on zookeeper). 

### Changes

Added test for the scenario I encountered.
Tests expanded with post-autorecovery validation of ledger ensembles to make sure the decomemd bookie no longer used there.

Ledger checker now does not skip fragments with firstEntryId == -1 but checks if the bookie is actually available.

This changes current behavior. 
The fragments for an empty ledger will end up "rereplicated" (metadata updated).

Master Issue: #3220

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerChecker.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerChecker.java
@@ -40,7 +40,6 @@ import org.apache.bookkeeper.proto.BookkeeperInternalCallbacks.ReadEntryCallback
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-
 /**
  * A utility class to check the complete ledger and finds the UnderReplicated fragments if any.
  *
@@ -238,7 +237,13 @@ public class LedgerChecker {
             if (lastStored != LedgerHandle.INVALID_ENTRY_ID) {
                 throw new InvalidFragmentException();
             }
-            cb.operationComplete(BKException.Code.OK, fragment);
+
+            if (bookieWatcher.isBookieUnavailable(fragment.getAddress(bookieIndex))) {
+                // fragment is on this bookie, but already know it's unavailable, so skip the call
+                cb.operationComplete(BKException.Code.BookieHandleNotAvailableException, fragment);
+            } else {
+                cb.operationComplete(BKException.Code.OK, fragment);
+            }
         } else if (bookieWatcher.isBookieUnavailable(fragment.getAddress(bookieIndex))) {
             // fragment is on this bookie, but already know it's unavailable, so skip the call
             cb.operationComplete(BKException.Code.BookieHandleNotAvailableException, fragment);
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieDecommissionTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieDecommissionTest.java
@@ -19,8 +19,11 @@
 package org.apache.bookkeeper.client;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.fail;
 import java.util.Iterator;
+import java.util.LinkedList;
+import java.util.List;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.bookkeeper.bookie.BookieImpl;
 import org.apache.bookkeeper.client.BKException.BKIllegalOpException;
@@ -29,6 +32,7 @@ import org.apache.bookkeeper.common.testing.annotations.FlakyTest;
 import org.apache.bookkeeper.conf.ServerConfiguration;
 import org.apache.bookkeeper.meta.UnderreplicatedLedger;
 import org.apache.bookkeeper.meta.ZkLedgerUnderreplicationManager;
+import org.apache.bookkeeper.net.BookieId;
 import org.apache.bookkeeper.test.BookKeeperClusterTestCase;
 import org.junit.Test;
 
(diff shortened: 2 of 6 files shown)


bookkeeper-server/src/test/java/org/apache/bookkeeper/replication/BookieAutoRecoveryTest.java
```diff
@@ -383,6 +383,7 @@ public void testEmptyLedgerLosesQuorumEventually() throws Exception {
         LOG.info("Killing last bookie, {}, in ensemble {}", replicaToKill,
                  lh.getLedgerMetadata().getAllEnsembles().get(0L));
         killBookie(replicaToKill);
+        startNewBookie();
```
eolivelli, line 386: why are we changing this existing test ?

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieDecommissionTest.java
```diff
@@ -130,11 +144,16 @@ public void testDecommissionForLedgersWithMultipleSegmentsAndNotWriteClosed() th
             lh4.addEntry(j, "data".getBytes());
         }
 
+        // avoiding autorecovery fencing the ledger
+        servers.forEach(srv -> srv.stopAutoRecovery());
```
eolivelli, line 148: I am not sure that "stopAutoRecovery" waits for autoRecovery to be totally stopped, maybe there is still some task running?

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieDecommissionTest.java
```diff
@@ -88,7 +98,7 @@ public void testDecommissionBookie() throws Exception {
          */
         bkAdmin.decommissionBookie(BookieImpl.getBookieId(killedBookieConf));
         bkAdmin.triggerAudit();
-        Thread.sleep(500);
+        Thread.sleep(5000);
```
eolivelli, line 101: out of the scope of this PR, but in the future it would be better to wait for a specific condition, in order to reduce flakyness

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieDecommissionTest.java
```diff
@@ -44,19 +49,23 @@ public class BookieDecommissionTest extends BookKeeperClusterTestCase {
 
     public BookieDecommissionTest() {
         super(NUM_OF_BOOKIES, 480);
-        baseConf.setOpenLedgerRereplicationGracePeriod(String.valueOf(30000));
+        baseConf.setOpenLedgerRereplicationGracePeriod(String.valueOf(1000));
         setAutoRecoveryEnabled(true);
     }
 
     @FlakyTest("https://github.com/apache/bookkeeper/issues/502")
+    @Test
```
eolivelli, line 57: so basically we were not running this test anymore.

great to see this running again

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerChecker.java
```diff
@@ -238,7 +237,13 @@ private void verifyLedgerFragment(LedgerFragment fragment,
             if (lastStored != LedgerHandle.INVALID_ENTRY_ID) {
                 throw new InvalidFragmentException();
```
eolivelli, line 238: what about taking this chance of changing the code to adding a meaningful message to all the InvalidFragmentException thrown in this method ?

eolivelli (commented):
the fix (in LedgerChecker) makes sense to me.

I left some comments.
I am happy to see that we are re-enabling a test that was lost because it was too flaky

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieDecommissionTest.java
dlg99: it does wait.
I reduced openLedgerRereplicationGracePeriod, thus autorecovery may fence the ledger and now the test has to stop/start it to avoid flakiness in this case (add to the ledger to force ensemble change)

bookkeeper-server/src/test/java/org/apache/bookkeeper/replication/BookieAutoRecoveryTest.java
dlg99: The behavior has changed now.
basically autorecovery used to ignore failed bookie for empty ledgers if it is not in the writeset for entryId 0 when WQ < ES.
e.g. ensemble = (bl1, bk2, bk3), WQ = 2, ledger is empty.
bk3 is down, it is not in the writeset for entryId=0.

Auditor would add it to underreplicated, autorecovery would skip it and simply remove from underreplicated, go back to the start of the sentence.
This is nice until you end up with an empty ledger like that. or a few thousands of them.

eolivelli (approved):
Lgtm

Awesome work

nicoloboschi (approved):
LGTM

PR #3248, opened 2022-04-26
Support set local region for RegionAwareEnsemblePlacementPolicy

wangjialing218:
### Motivation

When RegionAwareEnsemblePlacementPolicy is enabled, write set of one entry contains bookie nodes in different region.
When bookkeeper client read entries, the policy provide a read sequence to read from bookies in the same region first.  see  
https://github.com/apache/bookkeeper/blob/4debbbf84db09f80ffeb4154e8e6feed134be0c0/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java#L834

It depend on `myRegion` which indicate the local region of bookkeeper client (mostly pulsar broker).
Local region  is not a configurable value, and it's generated by local host address. see https://github.com/apache/bookkeeper/blob/4debbbf84db09f80ffeb4154e8e6feed134be0c0/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java#L208

It's okey when there is a bookie in the same host of broker.
But in most case, brokers were deployed in other nodes (or pod in kubernetes), and bookkeeper client will fail to get local region value.

### Changes

Add configuration of local region for bookkeeper client.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RegionAwareEnsemblePlacementPolicy.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RegionAwareEnsemblePlacementPolicy.java
@@ -63,6 +63,7 @@ public class RegionAwareEnsemblePlacementPolicy extends RackawareEnsemblePlaceme
     public static final String REPP_DISABLE_DURABILITY_ENFORCEMENT_FEATURE = "reppDisableDurabilityEnforcementFeature";
     public static final String REPP_ENABLE_VALIDATION = "reppEnableValidation";
     public static final String REGION_AWARE_ANOMALOUS_ENSEMBLE = "region_aware_anomalous_ensemble";
+    public static final String REPP_LOCAL_REGION = "reppLocalRegion";
     static final int MINIMUM_REGIONS_FOR_DURABILITY_DEFAULT = 2;
     static final int REGIONID_DISTANCE_FROM_LEAVES = 2;
     static final String UNKNOWN_REGION = "UnknownRegion";
@@ -79,7 +80,7 @@ public class RegionAwareEnsemblePlacementPolicy extends RackawareEnsemblePlaceme
     protected Feature disableDurabilityFeature;
     private int lastRegionIndex = 0;
 
-    RegionAwareEnsemblePlacementPolicy() {
+    protected RegionAwareEnsemblePlacementPolicy() {
         super();
         perRegionPlacement = new HashMap<String, TopologyAwareEnsemblePlacementPolicy>();
         address2Region = new ConcurrentHashMap<BookieId, String>();
@@ -173,6 +174,10 @@ public class RegionAwareEnsemblePlacementPolicy extends RackawareEnsemblePlaceme
         super.initialize(conf, optionalDnsResolver, timer, featureProvider, statsLogger, bookieAddressResolver)
                 .withDefaultRack(NetworkTopology.DEFAULT_REGION_AND_RACK);
         myRegion = getLocalRegion(localNode);
+        String localRegion = conf.getString(REPP_LOCAL_REGION, null);
+        if (null != localRegion) {
+            myRegion = localRegion;
+        }
         enableValidation = conf.getBoolean(REPP_ENABLE_VALIDATION, true);
         // We have to statically provide regions we want the writes to go through and how many regions
         // are required for durability. This decision cannot be driven by the active bookies as the
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/TestRegionAwareEnsemblePlacementPolicy.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/TestRegionAwareEnsemblePlacementPolicy.java
@@ -1208,13 +1208,13 @@ public class TestRegionAwareEnsemblePlacementPolicy extends TestCase {
 
     @Test
     public void testBasicReorderReadSequenceWithLocalRegion() throws Exception {
-        prepareNetworkTopologyForReorderTests(myRegion);
+        prepareNetworkTopologyForReorderTests("region2");
         basicReorderReadSequenceWithLocalRegionTest("region2", false);
     }
 
     @Test
     public void testBasicReorderReadLACSequenceWithLocalRegion() throws Exception {
-        prepareNetworkTopologyForReorderTests(myRegion);
+        prepareNetworkTopologyForReorderTests("region2");
         basicReorderReadSequenceWithLocalRegionTest("region2", true);
     }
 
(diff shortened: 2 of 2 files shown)


bookkeeper-server/src/test/java/org/apache/bookkeeper/client/TestRegionAwareEnsemblePlacementPolicy.java
```diff
@@ -1208,16 +1208,31 @@ private void prepareNetworkTopologyForReorderTests(String myRegion) throws Excep
 
     @Test
     public void testBasicReorderReadSequenceWithLocalRegion() throws Exception {
+        prepareNetworkTopologyForReorderTests("region2");
         basicReorderReadSequenceWithLocalRegionTest("region2", false);
     }
 
     @Test
     public void testBasicReorderReadLACSequenceWithLocalRegion() throws Exception {
+        prepareNetworkTopologyForReorderTests("region2");
```
hangc0276, line 1217: We should both test read and write.

hangc0276 (changes requested):
For your motivation, If we read all the entries from the configured region, it may cause hot bookie, and increase the quarantine rate of the hot bookie.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/TestRegionAwareEnsemblePlacementPolicy.java
wangjialing218: > For your motivation, If we read all the entries from the configured region, it may cause hot bookie, and increase the quarantine rate of the hot bookie.

As current policy, read sequence is as follows https://github.com/apache/bookkeeper/blob/4debbbf84db09f80ffeb4154e8e6feed134be0c0/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java#L810
This motivation just add a configured to set the local region for client, bookies in the configured local region will be orderd as `1. available (local) bookies`, when these bookies become unavailable or slow, they will be orderd as `3. R* remaining (local) bookies` or `6. slow bookies`.

> We should both test read and write.

This motivation just add a configured related to the read order of bookies in a writeSet. IMO we do not need test write.

StevenLuMT:
rerun failure checks

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RegionAwareEnsemblePlacementPolicy.java
```diff
@@ -173,6 +174,10 @@ public RegionAwareEnsemblePlacementPolicy initialize(ClientConfiguration conf,
         super.initialize(conf, optionalDnsResolver, timer, featureProvider, statsLogger, bookieAddressResolver)
                 .withDefaultRack(NetworkTopology.DEFAULT_REGION_AND_RACK);
         myRegion = getLocalRegion(localNode);
+        String localRegion = conf.getString(REPP_LOCAL_REGION, null);
```
StevenLuMT, line 177: simple code 
`String localRegion = conf.getString(REPP_LOCAL_REGION);`
instead of
`String localRegion = conf.getString(REPP_LOCAL_REGION, null);`

image

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RegionAwareEnsemblePlacementPolicy.java
```diff
@@ -79,7 +80,7 @@ public class RegionAwareEnsemblePlacementPolicy extends RackawareEnsemblePlaceme
     protected Feature disableDurabilityFeature;
     private int lastRegionIndex = 0;
 
-    RegionAwareEnsemblePlacementPolicy() {
+    protected RegionAwareEnsemblePlacementPolicy() {
```
StevenLuMT, line 83: why not change it to public?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RegionAwareEnsemblePlacementPolicy.java
wangjialing218: I'm not sure whether  `throwExceptionOnMissing` could be set true in some condition.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RegionAwareEnsemblePlacementPolicy.java
wangjialing218: Change to protected so that we could extends `RegionAwareEnsemblePlacementPolicy` and implement some addtion policy in broker side in future.
IMO no need change to public.

StevenLuMT:

dlg99:
What alternative solutions have you looked at?

> It's okey when there is a bookie in the same host of broker.

AFACT, the dummy bookie node is created to make use of the resolver API that takes BookieNode as a parameter. 
After that it goes through the same resolution mechanism as any bookie; so if you use script-based mapping you can update it to resolve the client's address to region/rack etc, it does not really matter if the client is co-hosted with a bookie.

wangjialing218:
> What alternative solutions have you looked at?

In synchronous geo-replication solution, suppose we have 4 bookies in two regions, bk1 and bk2 with rack setting `/region1/rack`, bk3 and bk4 with rack setting `/region2/rack`。
A broker in region1 need to read ledger data and the writeSet is {bk1,bk3}, it's better to let broker read from bk1 which is in same region to avoid across region tracffic. If the topic unloaded to another broker in region2, the broker could read ledger data from bk3 instead.

> AFACT, the dummy bookie node is created to make use of the resolver API that takes BookieNode as a parameter. After that it goes through the same resolution mechanism as any bookie; so if you use script-based mapping you can update it to resolve the client's address to region/rack etc, it does not really matter if the client is co-hosted with a bookie.

It's okey to resolve the broker's address to region/rack to achieve, we need to configure the resolver and it's not an easy way if brokers deployed in kubernetes.
This PR provide a  bookkeeper client configuration `reppLocalRegion` for broker, so we could add region-aware setting for each broker according to where the broker is placement by command or configuration.

hangc0276:
>It's okey to resolve the broker's address to region/rack to achieve, we need to configure the resolver and it's not an easy way if brokers deployed in kubernetes.

@wangjialing218 It is easy to configure by `bin/pulsar-admin bookies set-bookie-rack` command

If we introduce a configuration `reppLocalRegion` for the broker, and the broker is scheduled to another region by k8s, we still need to change the `reppLocalRegion` in conf/broker.conf. It will be hard to maintain.

PR #3269, opened 2022-05-11
Fix dead link and add CI check

CalvinKirs:

### Motivation
I found some dead links while reading the README, to avoid this, I added a CI check.

### Changes
- Fix the dead links
- Add the markdown-link-check check


nicoloboschi (changes requested):
cool! 
you need to exclude the new file from the license check

CalvinKirs:
> cool! you need to exclude the new file from the license check

Thanks for your reminder, I will exclude all JSON files, it's not necessary to check the license.

.github/workflows/dead-link-checker.yaml
```diff
@@ -0,0 +1,36 @@
...
+  pull_request:
+
+concurrency:
+  group: dlc-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  CheckDeadLinks:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
```
dlg99, line 29: We only need to run this if select dirs have changes (e.g. `site3/`) and can skip for majority of the PRs

CalvinKirs: Hi, I've thought of doing that, but if someone adds MD files to another dir, that may result in frequent maintenance of this directory (most times. people probably forget to do this).

It's a simple CI, so I'd rather run it for every PR, WDYT?

CalvinKirs: now,the check is only triggered when the MD file is modified

CalvinKirs: I didn't notice before that there is such a way to determine the file type 🤣

CalvinKirs:
Good to go, PTAL, thx @dlg99 @nicoloboschi 
ps: I ignored some dead link checks because I'm not sure what the correct one should be, maybe we can create an issue to see if other familiar people are interested in doing it.
Another CI fail has nothing to do with this, maybe we need to rerun

.github/workflows/dead-link-checker.yaml
```diff
@@ -0,0 +1,42 @@
...
+    branches:
+      - master
+      - branch-*
+
+concurrency:
+  group: dlc-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  CheckDeadLinks:
```
nicoloboschi, line 33: please use kebab-case (check-dead-links)

.github/workflows/dead-link-checker.yaml
```diff
@@ -0,0 +1,42 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Dead Link Checker
+
+on:
+  push:
+    paths:
```
nicoloboschi, line 21: this should be included also in the pull_request config?

nicoloboschi (changes requested):
really cool!

.github/workflows/dead-link-checker.yaml
CalvinKirs: Thx, done

.github/workflows/dead-link-checker.yaml
CalvinKirs: thanks you for pointing it out

nicoloboschi (approved):

nicoloboschi:
rerun failure checks

hangc0276 (approved):

PR #3291, opened 2022-05-25
[website] Make download page fully ASF compliant

### LINKED ISSUE #3290: New website download page is not compliant with the ASF policy
The download page has three ways it is breaking ASF Release Policies.
- Links to current releases should not be to [archive.apache.org](http://archive.apache.org/) - the preferred form is [www.apache.org/dyn/closer.lua/](http://www.apache.org/dyn/closer.lua/)
- There is no link to a KEYS file on [downloads.apache.org](http://downloads.apache.org/) with a discussion about verification by checking the asc detached signature with the release signing key.

Reference: https://infra.apache.org/release-download-pages.html

nicoloboschi:
Fix #3290

### Changes

* Fixed the SHA link, we use SHA512
* Fixed source and binary links to use closer.lua instead of archive.apache.org
* Increased the visibility of the checksum validation procedure

Master Issue: #3290


eolivelli (approved):
LGTM

PR #3301, opened 2022-05-31
[ci] [build] add macos build check

Shoothzj:
### Motivation
add macos build check to avoid broken change on macos x86


Shoothzj:
rerun failure checks

eolivelli (approved):

Issue #3306, opened 2022-06-03
Bookie can't start after rebooting due the cookie mistmatch


OlegKashtanov:
**BUG REPORT**

***Describe the bug***
Bookie image version: **apache/bookkeeper:4.15.0**
The issue was found using Bookkeeper in Kubernetes. After restarting the Bookie pod, it will always be in the **CrashLoopBackOff** state due the cookies mismatch:
```
09:15:54.921 [main] DEBUG org.apache.bookkeeper.zookeeper.ZooWorker - Stopped executing getData (/pravega/pravega/bookkeeper/ledgers/cookies/bookkeeper-bookie-0.bookkeeper-bookie-headless.default.svc.cluster.local:3181, watcher = null) after 1 attempts.
09:15:54.921 [main] DEBUG org.apache.bookkeeper.zookeeper.ZooWorker - Stopped executing getData (/pravega/pravega/bookkeeper/ledgers/cookies/bookkeeper-bookie-0.bookkeeper-bookie-headless.default.svc.cluster.local:3181, watcher = false) after 1 attempts.
09:15:54.978 [main] ERROR org.apache.bookkeeper.bookie.LegacyCookieValidation - Cookie for this bookie is not stored in metadata store. Bookie failing to come up
09:15:54.979 [main] ERROR org.apache.bookkeeper.server.Main - Failed to build bookie server
org.apache.bookkeeper.bookie.BookieException$InvalidCookieException:
        at org.apache.bookkeeper.bookie.LegacyCookieValidation.checkCookies(LegacyCookieValidation.java:122)
        at org.apache.bookkeeper.server.Main.buildBookieServer(Main.java:422)
        at org.apache.bookkeeper.server.Main.doMain(Main.java:272)
        at org.apache.bookkeeper.server.Main.main(Main.java:255)

```
At the initial start Bookie uses the znode with **Bookie ID** for creating the cookie: 
```
09:10:36.416 [main] DEBUG org.apache.bookkeeper.zookeeper.ZooWorker - Execute create (/pravega/pravega/bookkeeper/ledgers/cookies/bookkeeper-bookie-0-12842, acl = [31,s{'world,'anyone}], mode = PERSISTENT) at 0 retry attempt.
```
After  the restart it tries to find the cookie using the znode with **Bookie hostname**, but it's the wrong znode:
```
[...]

PR #3308, opened 2022-06-03
Issue 3306: Bookie can't start after rebooting due the cookie mistmatch

RaulGracia:
### Motivation
Fixes a bug when validating the Cookie that may fail to consider a custom Bookie ID when configured.

### Changes

Added custom Bookie ID when executing `possibleBookieIds()` in `LegacyCookieValidation`. Note that this was already fixed few months back, concretely in this PR: https://github.com/apache/bookkeeper/pull/2796
But it seems that the regression bringing back the old behavior was introduced in PR #2901.

This change may need to be cherry-picked to `branch-4.15`.

Master Issue: #3306

Signed-off-by: Raúl Gracia <raul.gracia@emc.com>


eolivelli (approved):
good catch

nicoloboschi (approved):
LGTM

PR #3314, opened 2022-06-05
Remove gradle from workflow

Shoothzj:
### Motivation
Remove gradle from the workflow check. 
The whole goal is remove gradle from the project. See https://lists.apache.org/thread/k3pdlpbvkzt061338fkxtppq3z70nk03

### changes
fix the maven groovy tests
See https://maven.apache.org/surefire/maven-surefire-plugin/examples/spock.html


.github/workflows/backward-compat-tests.yml
```diff
@@ -49,10 +49,17 @@ jobs:
         with:
           java-version: 1.8
       - name: Build
-        run: ./gradlew stream:server:build -x test ${GRADLE_ARGS}
+        run: mvn -B -nsu -am -pl :stream-storage-server test
```
eolivelli, line 52: this is not the same thing.
in Gradle "-x test" means to skip tests and not to run tests

pom.xml
```diff
@@ -148,7 +148,7 @@
     <!-- required by zookeeper test utilities imported from ZooKeeper -->
     <junit5.version>5.8.2</junit5.version>
     <libthrift.version>0.14.2</libthrift.version>
-    <lombok.version>1.18.22</lombok.version>
+    <lombok.version>1.18.24</lombok.version>
```
eolivelli, line 151: please upgrade dependencies in a separate PR

eolivelli (changes requested):

nicoloboschi (changes requested):
@Shoothzj So we decided to drop gradle? it's better to clear it on the ML before starting remove Gradle stuff

pom.xml
Shoothzj: @eolivelli reverted

.github/workflows/backward-compat-tests.yml
Shoothzj: @eolivelli PTAL again

Shoothzj:
@eolivelli @nicoloboschi PTAL again, thanks

.github/workflows/backward-compat-tests.yml
```diff
@@ -49,10 +49,17 @@ jobs:
         with:
           java-version: 1.8
       - name: Build
-        run: ./gradlew stream:server:build -x test ${GRADLE_ARGS}
+        run: mvn -B -nsu -am -pl :stream-storage-server clean install
       - name: Test current server with old clients
-        run: ./gradlew :tests:backward-compat:current-server-old-clients:test ${GRADLE_ARGS} || (tail -n +1 tests/backward-compat/current-server-old-clients/build/reports/tests/test/classes/* && tail -n +1 tests/backward-compat/current-server-old-clients/build/container-logs/**/* && exit 1)
+        run: mvn -B -nsu -am -pl :backward-compat-current-server-old-clients test
```
nicoloboschi, line 54: -am is not needed if the maven phase is "test" 
In this case will run the tests for every dependant module ?

Shoothzj: @nicoloboschi `-am` only compile the module, it will not run tests. And it's needed, otherwise the test module can't compile success.

Shoothzj:
@eolivelli @nicoloboschi PTAL again, thanks for your review

eolivelli (approved):
Lgtm as far as CI passes

PR #3321, opened 2022-06-07
Fix the PendingAddOp is not recycled when LedgerHandler closed

zymap:
---

### Motivation

When adding an entry into a bookie, the entry data lifecycle
is handled by the bookie client. The data buffer will be
released after receiving a corresponding response from the
bookie server. So the user doesn't care about the entry
buffer releasing.
But when the ledgerHandler is closed, the PendingAddOp is not
recycled which leads to the data buffer never being released.
We should release that after the callback is executed.


eolivelli (approved):
The change makes sense to me.
I am not sure we buy much advantage by recycling the Op object in this corner case.

I am afraid of possible side-effects that we would find only in the future.

btw +1 from me, good catch !

merlimat (approved):

lordcheng10 (approved):
nice catch.
I have encountered the problem of too many PendingAddOp objects causing GC in pulsar, this may solve my problem.

PR #3323, opened 2022-06-09
 Fix the V2 AddRequest object leak issue

zymap:
---

**Motivation**

If the request is a V2 add request, we retained the data's
reference when creating the AddRequest object. To avoid the
object leak, we need to release the reference if we met
any errors before sending it.


nicoloboschi (approved):

hangc0276 (approved):

zymap:
cherry-picked to branch-4.14 and branch-4.15

PR #3329, opened 2022-06-12
Fix data lost when configured multiple ledger directories

hangc0276:
### Motivation
We found one place where the bookie may lose data even though we turn on fsync for the journal.
Condition:
- One journal disk, and turn on fsync for the journal
- Configure two ledger disks, ledger1, and ledger2

Assume we write 100MB data into one bookie, 70MB data written into ledger1's write cache, and 30 MB data written into ledger2's write cache. Ledger1's write cache is full and triggers flush. In flushing the write cache, it will trigger a checkpoint to mark the journal’s lastMark position (100MB’s offset) and write the lastMark position into both ledger1 and ledger2's lastMark file.

At this time, this bookie shutdown without flush write cache, such as shutdown by `kill -9` command, and ledger2's write cache (30MB) doesn’t flush into ledger disk. But ledger2's lastMark position which persisted into lastMark file has been updated to 100MB’s offset.

When the bookie starts up, the journal reply position will be `min(ledger1's lastMark, ledger2's lastMark)`, and it will be 100MB’s offset. The ledger2's 30MB data won’t reply and that data will be lost.

Discussion thread:
https://lists.apache.org/thread/zz5vvv2yd80vqy22fv8wg5s2lqtkrzh9

### Solutions
The root cause of this bug is that EntryLogger1 triggers a checkpoint when its write cache is full, updating both EntryLogger1  and EntryLogger2's `lastMark` position. However, EntryLogger2's data may still be in WriteCache, which may lead to data loss when the bookie shutdown will `kill -9`

There are two solutions for this bug.
#### Update `lastMark` position individually.
- When EntryLogger1 triggers the checkpoint, we only update EntryLogger1's `lastMark` position instead of updating EntryLogger2's `lastMark` position at the same time.
- When SyncThread triggers the checkpoint, we update all the EntryLoggers' `lastMark` positions.
- When determining whether a journal file can be deleted, we should get the smallest `lastMark` position among all the writeable EntryLoggers, and delete the journal files which less than the smallest `lastMark` position.
- When replying to the journal on bookie startups, we need to get the smallest `lastMark` position, and reply to the journal files with this position, otherwise, we will lose data.

However, there is one case being hard to handle in replying to the journal stage.
When one ledger disk transfers from ReadOnly mode to Writeable mode, the `lastMark` position is an old value. Using the old position to reply to the journal files will lead to a target journal file not found exception.

#### Only update `lastMark` position in SyncThread
There are two places that can trigger a checkpoint.
- The scheduled tasks in SyncThread.doCheckpoint
- The ledgerDir write-cache full and then flush

The second way is the root cause of data loss if the ledger is configured with multiple directories.
We can turn off the second way's update `lastMark` position operation and only make SyncThread update the `lastMark` position in a checkpoint when the ledger is configured with multiple directories.

This is the simplest way to fix this bug, but it has two drawbacks.
- The `lastMark` position updates depend on SyncThread doing checkpoint intervals. In Pulsar, the default interval is 60s. It means the journal file expires with at least 60s 
- The bookie startup replying journal depend on the `lastMark` position. It means the journal will reply to at least 60s journal data before the start-up is complete. It may lead to the bookie start-up speed slowing down.

IMO, compared to data loss, the above two drawbacks can be acceptable.

### Changes
I choose the second solution to fix this bug.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
@@ -857,11 +857,12 @@ public class BookieImpl extends BookieCriticalThread implements Bookie {
 
                 stateManager.forceToShuttingDown();
 
-                // Shutdown Sync thread
-                syncThread.shutdown();
                 // turn bookie to read only during shutting down process
                 LOG.info("Turning bookie to read only during shut down");
                 stateManager.forceToReadOnly();
+
+                // Shutdown Sync thread
+                syncThread.shutdown();
 
                 // Shutdown journals
                 for (Journal journal : journals) {
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
@@ -857,6 +857,8 @@ public class BookieImpl extends BookieCriticalThread implements Bookie {
 
                 stateManager.forceToShuttingDown();
 
+                // Shutdown Sync thread
+                syncThread.shutdown();
                 // turn bookie to read only during shutting down process
                 LOG.info("Turning bookie to read only during shut down");
                 stateManager.forceToReadOnly();
@@ -868,9 +870,6 @@ public class BookieImpl extends BookieCriticalThread implements Bookie {
 
                 // Shutdown the EntryLogger which has the GarbageCollector Thread running
                 ledgerStorage.shutdown();
-
-                // Shutdown Sync thread
-                syncThread.shutdown();
 
                 //Shutdown disk checker
                 dirsMonitor.shutdown();
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
@@ -861,9 +861,6 @@ public class BookieImpl extends BookieCriticalThread implements Bookie {
                 LOG.info("Turning bookie to read only during shut down");
                 stateManager.forceToReadOnly();
 
-                // Shutdown Sync thread
-                syncThread.shutdown();
-
                 // Shutdown journals
                 for (Journal journal : journals) {
                     journal.shutdown();
@@ -871,6 +868,9 @@ public class BookieImpl extends BookieCriticalThread implements Bookie {
 
                 // Shutdown the EntryLogger which has the GarbageCollector Thread running
                 ledgerStorage.shutdown();
+
+                // Shutdown Sync thread
+                syncThread.shutdown();
 
                 //Shutdown disk checker
                 dirsMonitor.shutdown();
@@ -1284,5 +1284,10 @@ public class BookieImpl extends BookieCriticalThread implements Bookie {
                 bookieStats.getReadEntryStats().registerFailedEvent(elapsedNanos, TimeUnit.NANOSECONDS);
             }
         }
+    }
+
+    @VisibleForTesting
+    public List<Journal> getJournals() {
+        return this.journals;
     }
 }
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
@@ -226,7 +226,7 @@ public class Journal extends BookieCriticalThread implements CheckpointSource {
          * The last mark should first be max journal log id,
          * and then max log position in max journal log.
          */
-        void readLog() {
+        public void readLog() {
             byte[] buff = new byte[16];
             ByteBuffer bb = ByteBuffer.wrap(buff);
             LogMark mark = new LogMark();
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
@@ -143,6 +143,7 @@ public class SingleDirectoryDbLedgerStorage implements CompactableLedgerStorage 
     private final long maxReadAheadBytesSize;
 
     private final Counter flushExecutorTime;
+    private final boolean singleLedgerDirs;
 
     public SingleDirectoryDbLedgerStorage(ServerConfiguration conf, LedgerManager ledgerManager,
                                           LedgerDirsManager ledgerDirsManager, LedgerDirsManager indexDirsManager,
@@ -172,6 +173,7 @@ public class SingleDirectoryDbLedgerStorage implements CompactableLedgerStorage 
         this.writeCacheMaxSize = writeCacheSize;
         this.writeCache = new WriteCache(allocator, writeCacheMaxSize / 2);
         this.writeCacheBeingFlushed = new WriteCache(allocator, writeCacheMaxSize / 2);
+        this.singleLedgerDirs = conf.getLedgerDirs().length == 1;
 
         readCacheMaxSize = readCacheSize;
         this.readAheadCacheBatchSize = readAheadCacheBatchSize;
@@ -905,7 +907,9 @@ public class SingleDirectoryDbLedgerStorage implements CompactableLedgerStorage 
     public void flush() throws IOException {
         Checkpoint cp = checkpointSource.newCheckpoint();
         checkpoint(cp);
-        checkpointSource.checkpointComplete(cp, true);
+        if (singleLedgerDirs) {
+            checkpointSource.checkpointComplete(cp, true);
+        }
     }
 
     @Override
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorageTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorageTest.java
@@ -31,16 +31,21 @@ import io.netty.buffer.ByteBufUtil;
 import io.netty.buffer.Unpooled;
 import io.netty.util.ReferenceCountUtil;
 import java.io.File;
+import java.io.FileInputStream;
 import java.io.IOException;
+import java.nio.ByteBuffer;
 import java.util.List;
 import org.apache.bookkeeper.bookie.Bookie;
 import org.apache.bookkeeper.bookie.Bookie.NoEntryException;
 import org.apache.bookkeeper.bookie.BookieException;
 import org.apache.bookkeeper.bookie.BookieImpl;
+import org.apache.bookkeeper.bookie.CheckpointSource;
+import org.apache.bookkeeper.bookie.CheckpointSourceList;
 import org.apache.bookkeeper.bookie.DefaultEntryLogger;
 import org.apache.bookkeeper.bookie.EntryLocation;
 import org.apache.bookkeeper.bookie.LedgerDirsManager;
 import org.apache.bookkeeper.bookie.LedgerStorage;
+import org.apache.bookkeeper.bookie.LogMark;
 import org.apache.bookkeeper.bookie.TestBookieImpl;
 import org.apache.bookkeeper.bookie.storage.EntryLogger;
 import org.apache.bookkeeper.conf.ServerConfiguration;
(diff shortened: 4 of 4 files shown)


gaozhangmin:
@hangc0276 I found when syncThread do flush, there exists duplicate `checkpointComplete` invoke.

SyncThread flush:
```
private void flush() {
        Checkpoint checkpoint = checkpointSource.newCheckpoint();
        try {
            ledgerStorage.flush();
        } catch (NoWritableLedgerDirException e) {
            log.error("No writeable ledger directories", e);
            dirsListener.allDisksFull(true);
            return;
        } catch (IOException e) {
            log.error("Exception flushing ledgers", e);
            return;
        }

        if (disableCheckpoint) {
            return;
        }

        log.info("Flush ledger storage at checkpoint {}.", checkpoint);
        try {
            checkpointSource.checkpointComplete(null, checkpoint, false);
        } catch (IOException e) {
            log.error("Exception marking checkpoint as complete", e);
            dirsListener.allDisksFull(true);
        }
    }
```

```
@Override
    public void flush() throws IOException {
        Checkpoint cp = checkpointSource.newCheckpoint();
        checkpoint(cp);
        checkpointSource.checkpointComplete(ledgerDir, cp, true);
    }
```

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
```diff
@@ -369,8 +369,9 @@ public Checkpoint newCheckpoint() {
                 }
 
                 @Override
-                public void checkpointComplete(Checkpoint checkpoint, boolean compact)
-                        throws IOException {
+                public void checkpointComplete(Checkpoint checkpoint,
+                                               boolean compact,
+                                               LedgerDirsManager ledgerDirsManager1) throws IOException {
```
eolivelli, line 374: nit: `ledgerDirsManager`

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
```diff
@@ -603,7 +604,7 @@ private void replay(Journal journal, JournalScanner scanner) throws IOException
             journalId >= markedLog.getLogFileId());
         // last log mark may be missed due to no sync up before
         // validate filtered log ids only when we have markedLogId
-        if (markedLog.getLogFileId() > 0) {
+        if (markedLog.getLogFileId() > 0 && markedLog.getLogFileId() != Long.MAX_VALUE) {
```
eolivelli, line 607: I wonder if it would be better to have a constant for Long.MAX_VALUE

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -660,7 +660,7 @@ static void writePaddingBytes(JournalChannel jc, ByteBuf paddingBuffer, int jour
     // Should data be fsynced on disk before triggering the callback
     private final boolean syncData;
 
-    private final LastLogMark lastLogMark = new LastLogMark(0, 0);
+    private final LastLogMark lastLogMark = new LastLogMark(Long.MAX_VALUE, Long.MAX_VALUE);
```
eolivelli, line 663: we should make a public constant here, otherwise magic numbers are easily forgotten.

also, I may understand why we are changing from 0 to MAX_VALUE, but...what is the impact ?

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieAccessor.java
```diff
@@ -35,6 +35,6 @@ public static void forceFlush(BookieImpl b) throws IOException {
         CheckpointSourceList source = new CheckpointSourceList(b.journals);
         Checkpoint cp = source.newCheckpoint();
         b.ledgerStorage.flush();
-        source.checkpointComplete(cp, true);
+        source.checkpointComplete(cp, true, null);
```
eolivelli, line 38: so `null` means "all"...
we should document this in the javadocs

tools/perf/src/main/java/org/apache/bookkeeper/tools/perf/journal/JournalWriter.java
```diff
@@ -272,7 +272,7 @@ void execute(Journal[] journals) throws Exception {
             for (Journal journal : journals) {
                 Checkpoint cp = journal.newCheckpoint();
                 try {
-                    journal.checkpointComplete(cp, true);
+                    journal.checkpointComplete(cp, true, null);
```
eolivelli, line 275: what about adding a comment here ?
`null` means that the checkpoint is not started by a specific LedgersDirManager

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/CheckpointSource.java
```diff
@@ -89,7 +89,9 @@ public String toString() {
      * @param compact
      *          Flag to compact old checkpoints.
      */
-    void checkpointComplete(Checkpoint checkpoint, boolean compact) throws IOException;
+    void checkpointComplete(Checkpoint checkpoint,
+                            boolean compact,
+                            LedgerDirsManager ledgerDirsManager) throws IOException;
```
eolivelli, line 94: we should add javadocs here and explain why we have this `ledgerDirsManager` and when it may be null

eolivelli (changes requested):
I think that you are on your way.
good work !

I left some minor feedback.
Waiting for tests and for more reviews.

This change should be picked in all active branches.
it is a serious bug

nicoloboschi (changes requested):
it's better to add proper test case to avoid future regressions

lordcheng10:
> 2.When bookie startup, read the minimal lastMark instead of the maximal lastMark as current last mark.
@hangc0276 How is this logic implemented? Do you need to modify the following logic?:
image

lordcheng10:
> @hangc0276 I found when syncThread do flush, there exists duplicate `checkpointComplete` invoke.
> 
> SyncThread flush:
> 
> ```
> private void flush() {
>         Checkpoint checkpoint = checkpointSource.newCheckpoint();
>         try {
>             ledgerStorage.flush();
>         } catch (NoWritableLedgerDirException e) {
>             log.error("No writeable ledger directories", e);
>             dirsListener.allDisksFull(true);
>             return;
>         } catch (IOException e) {
>             log.error("Exception flushing ledgers", e);
>             return;
>         }
> 
>         if (disableCheckpoint) {
>             return;
>         }
> 
>         log.info("Flush ledger storage at checkpoint {}.", checkpoint);
>         try {
>             checkpointSource.checkpointComplete(null, checkpoint, false);
>         } catch (IOException e) {
>             log.error("Exception marking checkpoint as complete", e);
>             dirsListener.allDisksFull(true);
>         }
>     }
> ```
> 
> ```
> @Override
>     public void flush() throws IOException {
>         Checkpoint cp = checkpointSource.newCheckpoint();
>         checkpoint(cp);
>         checkpointSource.checkpointComplete(ledgerDir, cp, true);
>     }
> ```

I also think the checkpoint is duplicated here

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
gaozhangmin: Change should also be made at `Journal.readLog`, which curMark should be set to the min mark from ledgers dir.

gaozhangmin (changes requested):

gaozhangmin:
> > 2.When bookie startup, read the minimal lastMark instead of the maximal lastMark as current last mark.
> > @hangc0276 How is this logic implemented? Do you need to modify the following logic?:
> 
> image

+1, changes should also be made at `Journal.readLog`

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
hangc0276: When creating a Journal instance, it will initiate `lastLogMark` by reading each ledger directory's `lastMark` file and get the minimum position as the replay start point. So we should init lastLogMark with the MAX_VALUE.

In fact, the original logic of init `lastLogMark` with 0, and get the maximum position of all the ledger directory's `lastMark` file to init the lastLogMark. IMO, it will lose data.

hangc0276:
> it's better to add proper test case to avoid future regressions

@nicoloboschi I have added the test to cover this change, please help take a look, thanks.

tools/perf/src/main/java/org/apache/bookkeeper/tools/perf/journal/JournalWriter.java
hangc0276: I have added the javadoc on the `checkpointComplete` method interface.

hangc0276:
@merlimat  @eolivelli @nicoloboschi @gaozhangmin @lordcheng10 I have updated the code and added the test to cover this change, Please help take a look, thanks.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
hangc0276: Maybe @merlimat has more context about init `lastLogMark` to 0

hangc0276:
rerun failure checks

hangc0276:
rerun failure checks

hangc0276:
rerun failure checks

gaozhangmin:
> > @hangc0276 I found when syncThread do flush, there exists duplicate `checkpointComplete` invoke.
> > SyncThread flush:
> > ```
> > private void flush() {
> >         Checkpoint checkpoint = checkpointSource.newCheckpoint();
> >         try {
> >             ledgerStorage.flush();
> >         } catch (NoWritableLedgerDirException e) {
> >             log.error("No writeable ledger directories", e);
> >             dirsListener.allDisksFull(true);
> >             return;
> >         } catch (IOException e) {
> >             log.error("Exception flushing ledgers", e);
> >             return;
> >         }
> > 
> >         if (disableCheckpoint) {
> >             return;
> >         }
> > 
> >         log.info("Flush ledger storage at checkpoint {}.", checkpoint);
> >         try {
> >             checkpointSource.checkpointComplete(null, checkpoint, false);
> >         } catch (IOException e) {
> >             log.error("Exception marking checkpoint as complete", e);
> >             dirsListener.allDisksFull(true);
> >         }
> >     }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```
> > @Override
> >     public void flush() throws IOException {
> >         Checkpoint cp = checkpointSource.newCheckpoint();
> >         checkpoint(cp);
> >         checkpointSource.checkpointComplete(ledgerDir, cp, true);
> >     }
> > ```
> 
> I also think the checkpoint is duplicated here

@hangc0276  This problem is not going to be resolved  here, right?

hangc0276:
> > > @hangc0276 I found when syncThread do flush, there exists duplicate `checkpointComplete` invoke.
> > > SyncThread flush:
> > > ```
> > > private void flush() {
> > >         Checkpoint checkpoint = checkpointSource.newCheckpoint();
> > >         try {
> > >             ledgerStorage.flush();
> > >         } catch (NoWritableLedgerDirException e) {
> > >             log.error("No writeable ledger directories", e);
> > >             dirsListener.allDisksFull(true);
> > >             return;
> > >         } catch (IOException e) {
> > >             log.error("Exception flushing ledgers", e);
> > >             return;
> > >         }
> > > 
> > >         if (disableCheckpoint) {
> > >             return;
> > >         }
> > > 
> > >         log.info("Flush ledger storage at checkpoint {}.", checkpoint);
> > >         try {
> > >             checkpointSource.checkpointComplete(null, checkpoint, false);
> > >         } catch (IOException e) {
> > >             log.error("Exception marking checkpoint as complete", e);
> > >             dirsListener.allDisksFull(true);
> > >         }
> > >     }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > ```
> > > @Override
> > >     public void flush() throws IOException {
> > >         Checkpoint cp = checkpointSource.newCheckpoint();
> > >         checkpoint(cp);
> > >         checkpointSource.checkpointComplete(ledgerDir, cp, true);
> > >     }
> > > ```
> > 
> > 
> > I also think the checkpoint is duplicated here
> 
> @hangc0276 This problem is not going to be resolved here, right?

@gaozhangmin  Yes，we can use another PR to solve it.

hangc0276:
rerun failure checks

gaozhangmin (approved):

gaozhangmin:
> > > > @hangc0276 I found when syncThread do flush, there exists duplicate `checkpointComplete` invoke.
> > > > SyncThread flush:
> > > > ```
> > > > private void flush() {
> > > >         Checkpoint checkpoint = checkpointSource.newCheckpoint();
> > > >         try {
> > > >             ledgerStorage.flush();
> > > >         } catch (NoWritableLedgerDirException e) {
> > > >             log.error("No writeable ledger directories", e);
> > > >             dirsListener.allDisksFull(true);
> > > >             return;
> > > >         } catch (IOException e) {
> > > >             log.error("Exception flushing ledgers", e);
> > > >             return;
> > > >         }
> > > > 
> > > >         if (disableCheckpoint) {
> > > >             return;
> > > >         }
> > > > 
> > > >         log.info("Flush ledger storage at checkpoint {}.", checkpoint);
> > > >         try {
> > > >             checkpointSource.checkpointComplete(null, checkpoint, false);
> > > >         } catch (IOException e) {
> > > >             log.error("Exception marking checkpoint as complete", e);
> > > >             dirsListener.allDisksFull(true);
> > > >         }
> > > >     }
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > ```
> > > > @Override
> > > >     public void flush() throws IOException {
> > > >         Checkpoint cp = checkpointSource.newCheckpoint();
> > > >         checkpoint(cp);
> > > >         checkpointSource.checkpointComplete(ledgerDir, cp, true);
> > > >     }
> > > > ```
> > > 
> > > 
> > > I also think the checkpoint is duplicated here
> > 
> > 
> > @hangc0276 This problem is not going to be resolved here, right?
> 
> @gaozhangmin Yes，we can use another PR to solve it.

I submit pr #3353 to solve this issue. @hangc0276  PTAL

wenbingshen (commented):
Since we are now replaying the journal from the smallest `LogMark`, can we get the journal log mark position of the current entry to the `JournalScanner` and compare it with the checkpoint position on the ledger disk where the entry to be restored is located, and only restore the entry whose `logMark` position is greater than the `checkpoint`?  So as to avoid repeatedly writing the data that has been flushed to the disk.

https://github.com/apache/bookkeeper/blob/90d55017da68111ffc578b611de998269f701f82/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java#L906
```java
                 if (!isPaddingRecord) {
                    scanner.process(journalVersion, offset, recBuff, journalId, recLog.fc.position());
                }
```

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -245,7 +246,8 @@ void readLog() {
                     }
                     bb.clear();
                     mark.readLogMark(bb);
-                    if (curMark.compare(mark) < 0) {
+                    // get the minimum mark position from all the ledger directories to ensure no data loss.
+                    if (curMark.compare(mark) > 0) {
```
wenbingshen, line 250: Since we are now replaying the journal from the smallest LogMark, can we get the journal log mark position of the current entry to the JournalScanner and compare it with the checkpoint position on the ledger disk where the entry to be restored is located, and only restore the entry whose logMark position is greater than the checkpoint? So as to avoid repeatedly writing the data that has been flushed to the disk.

hangc0276: It will introduce complex logic for this comparison.
1. If the ledger directory expands or shrinks, the map logic of the ledgerId to the ledger directory (`logMark`) also changed. It may be located on the wrong `logMark` file, and will lead to skipping the unflushed entries.
2. There are many kinds of storage implementation, such as dbLedgerStorage, SortedLedgerStorage, and InterleavedLedgerStorage, we should get the ledgerId related storage instance to check the `logMark` position for each storage implementation. This operation will introduce complex logic.
3. For the comparison, we can only save the write ledger throughput. We also need to read the data from the journal log file out.

Based on the above reason, I prefer to replay all entries in the journal log file based on the min `logMark` position.

StevenLuMT:
rerun failure checks

hangc0276:
This PR changed the default `logMark` behaviour, it lead to some unit test failure, I'm working on it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
```diff
@@ -829,7 +831,7 @@ private void swapWriteCache() {
     public void flush() throws IOException {
         Checkpoint cp = checkpointSource.newCheckpoint();
         checkpoint(cp);
-        checkpointSource.checkpointComplete(cp, true);
+        checkpointSource.checkpointComplete(cp, true, ledgerDirsManager);
```
StevenLuMT, line 834: @hangc0276 
SingleDirectoryDbLedgerStorage's ledgerDirsManager can't cover all directories, only currently managed directory,I think you need all ledger' dirs 
image

so the structure of LedgerDirsManager is very confusing, I am also optimizing this, see for details in mail talking with @eolivelli 
https://lists.apache.org/thread/njw4yhdgv9db64notq5o3hc0k8bbpl8v

StevenLuMT:

aloyszhang:
There are two places that can trigger `checkpoint`.
1. the scheduled tasks in `SyncThread.doCheckpoint` 
2. the ledgerDir write-cache full and then flush
The second way is the root cause of data loss.

If removing the `checkpointSource.checkpointComplete` logic in `flush`, then at this point we will not delete journal files.

The scheduled task `SyncThread.doCheckpoint` will invoke `checkpointSource.checkpointComplete` and it's safe here to delete journal files since we have already flushed all write-caches for all ledger directories.

WDYT @hangc0276

hangc0276:
> There are two places that can trigger `checkpoint`.
> 
> 1. the scheduled tasks in `SyncThread.doCheckpoint`
> 2. the ledgerDir write-cache full and then flush
>    The second way is the root cause of data loss.
> 
> If removing the `checkpointSource.checkpointComplete` logic in `flush`, then at this point we will not delete journal files.
> 
> The scheduled task `SyncThread.doCheckpoint` will invoke `checkpointSource.checkpointComplete` and it's safe here to delete journal files since we have already flushed all write-caches for all ledger directories.
> 
> WDYT @hangc0276
`
@aloyszhang Thanks for your suggestion. Yes, making the `SyncThread.doCheckpoint` the only endpoint is the simplest way to solve this problem. I updated the PR description, please help take a look, thanks.

aloyszhang (approved):
LGTM

wenbingshen (approved):

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
```diff
@@ -872,6 +869,9 @@ int shutdown(int exitCode) {
                 // Shutdown the EntryLogger which has the GarbageCollector Thread running
                 ledgerStorage.shutdown();
 
+                // Shutdown Sync thread
+                syncThread.shutdown();
```
aloyszhang, line 873: Here may through exceptions when shutdown the `syncThread` which will call `checkpoint`  of ledgerStoreage, since we have already shut down the ledgeerStorage before

hangc0276: Yes, you are right. I updated the code, please help take a look, thanks.

aloyszhang (approved):

hangc0276:
@eolivelli @merlimat @dlg99 @zymap I updated the code, and need your eyes for this PR, thanks.

zymap (approved):

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
```diff
@@ -905,7 +907,9 @@ private void swapWriteCache() {
     public void flush() throws IOException {
         Checkpoint cp = checkpointSource.newCheckpoint();
         checkpoint(cp);
-        checkpointSource.checkpointComplete(cp, true);
+        if (singleLedgerDirs) {
```
eolivelli, line 910: Please add a small comment with a quick description of the motivation for this condition

eolivelli (approved):
+1

PR #3333, opened 2022-06-15
journals init with capacity

AnonHxy:

### Motivation

`journals` should be initiated with capacity

### Changes

Initiating`journals`  with capacity

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
@@ -426,7 +426,7 @@ public class BookieImpl extends BookieCriticalThread implements Bookie {
         }
 
         // instantiate the journals
-        journals = Lists.newArrayList();
+        journals = Lists.newArrayListWithCapacity(journalDirectories.size());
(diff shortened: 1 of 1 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
```diff
@@ -426,7 +426,7 @@ public BookieImpl(ServerConfiguration conf,
         }
 
         // instantiate the journals
-        journals = Lists.newArrayList();
+        journals = Lists.newArrayListWithCapacity(journalDirectories.size());
```
Shoothzj, line 429: I think `new ArrayList<>(size)` is better. Guava also recommend jdk8+ to use `new ArrayList<>()`

Shoothzj (approved):

eolivelli (changes requested):
hello,
there is no actual gain with this change.
it is better to keep the code simple if not needed.

we are not in a hot path and pre-sizing a list with 1-2 elements is not a big win

AnonHxy:
OK, thanks for your suggestion. Now I close this PR @eolivelli

PR #3335, opened 2022-06-15
Pre break loop when self create layoutZNode succeed.

horizonzy:
Pre break loop when self create layoutZNode succeed.


nicoloboschi (commented):
It would be clearer to add the break just after the node creation, line 190

hangc0276:
@horizonzy  Would you please rebase the master?

StevenLuMT (changes requested):
I think this change is not necessary

horizonzy:
> I think this change is not necessary

It can reduce once to invoke `zkc.exists(layoutZNode, false)`

hangc0276 (approved):

Issue #3338, opened 2022-06-16
[improve] decommissionBookie always waiting too long time after ledgers be replicated completed


Nicklee007:
**improve**

As the Decommissioning bookie case， always change the bookie status to readonly firstly, and then wait some data expired, but always it has some ledgers (about 100+ -- 300+) legacy not be cleaned and the leaved ledgers only has little data , when we running `bin/bookkeeper shell decommissionbookie -bookieid  ` to decommission the bookie ,  we always pending   about 10 min and have not any log print, but we could find the znode /ledgers/underreplication/ledgers cleaned only few seconds and then the ledgers be  rereplicate completed。

***To Reproduce***

1. Go to change the bookie status to readonly;
2. waiting the most ledger expired;
3. stop bookie and run `bin/bookkeeper shell decommissionbookie -bookieid  ` 
4. See will wait long time about 10 min, even the ledgers which have few data is replicated completed, and after `Count of Ledgers which need to be rereplicated:` the log printed, the 10 min have not any other printed.

***Expected behavior***

The waiting time  not too long and tell us what happened.

dlg99:
I think the wait is related to https://github.com/apache/bookkeeper/blob/eadbdd4b6bfeef9924a3ff2c59fc3718cf3dc06b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperAdmin.java#L1623

You can make this time configurable via the decommission command flag / add logging.

Nicklee007:
> You can make this time configurable via the decommission command flag / add logging.
@dlg99 
Yep, the wait time related to the config of `maxSleepTimeInBetweenChecks`.  If we add command  parameter to change the `maxSleepTimeInBetweenChecks`, the users maybe could not forecast how long will it take to rereplicate completed. It's a risk if user set a small `maxSleepTimeInBetweenChecks` and then the auditor is running to do some time-consuming operation like `checkAllLedger` could not audit bookie immediately，will cause  check `areEntriesOfLedgerStoredInTheBookie` through zk too frequently and affect the zk server performance.

The PR #3339 will judgment if the` /ledgers/underreplication/ledgers` and `/ledgers/underreplication/locks` is empty to help us check if the  rereplicate is completed, and backoff when the  auditor is running as CheckAllLedgers or other time-consuming operation.
could you help me check the PR, Thx.

PR #3339, opened 2022-06-16
[improve] decommissionBookie waiting for ledgers to be replicated back off policy

Nicklee007:
Master Issue: #3338

### Motivation
As the Decommissioning bookie case， always change the bookie status to readonly firstly, and then wait some data expired, but always it has some ledgers (about 100+ -- 300+) legacy not be cleaned and the leaved ledgers only has little data , when we running `bin/bookkeeper shell decommissionbookie -bookieid  ` to decommission the bookie ,  we always pending on `waitForLedgersToBeReplicated()` about 10 min and have not any log print, but we could find the znode /ledgers/underreplication/ledgers cleaned only few seconds and then the ledgers be  rereplicate completed, we find  the sleep time is defined as the `Min(ledgers.size() * sleepTimePerLedger(10s) , maxSleepTimeInBetweenChecks(10 min))`, in the way , the both time always wait too long and has not any print will let user confused。
 

### Changes
In the bookie has many data to rereplicate case , we need the backoff policy to protect zk server, so 
1. To look the `/ledgers/underreplication/ledgers ` and ` /ledgers/underreplication/locks` every 10 sec, help us check if the ledgers replicate completed.
2. To avoid the auditor is running as CheckAllLedgers or other time-consuming operation when we trigger the audit bookie, then `/ledgers/underreplication/ledgers ` and ` /ledgers/underreplication/locks` will empty to long time , we need a back off policy to avoid frequent request zk to `validateBookieIsNotPartOfEnsemble`


Nicklee007:
@eolivelli @dlg99 Could you help to check the logic is correct?

Nicklee007:
rerun failure checks

Nicklee007:
rerun failure checks

StevenLuMT (changes requested):
Can you add some testcase to cover these functions?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperAdmin.java
```diff
@@ -1622,13 +1622,33 @@ private void waitForLedgersToBeReplicated(Collection<Long> ledgers, BookieId thi
...
+        int rereplicateBackoffMs = 10 * 1000; // 10 secs
+        int backOffCount = 0;
         Predicate<Long> validateBookieIsNotPartOfEnsemble = ledgerId -> !areEntriesOfLedgerStoredInTheBookie(ledgerId,
                 thisBookieAddress, ledgerManager);
         while (!ledgers.isEmpty()) {
             LOG.info("Count of Ledgers which need to be rereplicated: {}", ledgers.size());
             int sleepTimeForThisCheck = (long) ledgers.size() * sleepTimePerLedger > maxSleepTimeInBetweenChecks
                     ? maxSleepTimeInBetweenChecks : ledgers.size() * sleepTimePerLedger;
-            Thread.sleep(sleepTimeForThisCheck);
+            long nextValidateTime = System.currentTimeMillis() + sleepTimeForThisCheck;
```
hangc0276, line 1633: Could you use a backoff policy to simplify the code?  You can refer to: https://github.com/apache/pulsar/blob/master/pulsar-client/src/main/java/org/apache/pulsar/client/impl/Backoff.java

StevenLuMT:

PR #3342, opened 2022-06-17
Bookie cache optimization

StevenLuMT:

### Motivation

When flushing, trigger flush to exchange writeCache and flushBeingCache
flushBeingCache will be cleaned up immediately after flushed
this logic causes the cache capacity to fluctuate greatly

### Changes

This optimization mainly solves this point and realizes the logic:
1. When flushBeingCache is flushed to disk, it is not deleted immediately, but dumped to lastFlushCache
2. After the dump is completed, clear the flushBeingCache again
Eventually: the cache has a fixed cache duration

### Test conclusion
if open this bookie cache optimization,with same qps/bps,it can make writecache increase cache data by 80%
`isWriteCacheFixedLengthEnabled=true  #  open this bookie cache optimization`

### Test Data
The following is the environmental information and test data:
1. Old Config:
     jvm important config:
            -XX:MaxDirectMemorySize=32g
     bookie important config:
             entryLogPerLedgerEnabled=true
             flushInterval=18000

2. Optimized Config:
     jvm important config:
            -XX:MaxDirectMemorySize=32g
     bookie important config:
             isWriteCacheFixedLengthEnabled=true  #  open this bookie cache optimization
             entryLogPerLedgerEnabled=true
             flushInterval=18000

#### Test Data
write cache size :  use metric bookie_write_cache_size
max cache data compare: old config is 3386866184, optimized config is 6210369204, it can make writecache increase cache data by 80%
image

Bps compare :  use metric rate(bookie_BOOKIE_ADD_ENTRY_BYTES_sum)
image

qps compare : use metric  rate(bookkeeper_server_ADD_ENTRY_REQUEST_count)
image

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
@@ -668,7 +668,7 @@ public class SingleDirectoryDbLedgerStorage implements CompactableLedgerStorage 
                     long entryId = entry.readLong();
                     entry.resetReaderIndex();
                     if (log.isDebugEnabled()) {
-                        log.debug("Found last entry for ledger {} in write cache being flushing: {}", ledgerId, entryId);
+                        log.debug("Found last entry for ledger {} in write cache flushing: {}", ledgerId, entryId);
                     }
                 }
 
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
@@ -700,10 +700,10 @@ public class SingleDirectoryDbLedgerStorage implements CompactableLedgerStorage 
                 if (entry != null) {
                     if (log.isDebugEnabled()) {
                         entry.readLong(); // ledgedId
-                        long entryId = entry.readLong();
+                        long eId = entry.readLong();
                         entry.resetReaderIndex();
                         if (log.isDebugEnabled()) {
-                            log.debug("Found last entry for ledger {} in write cache being flushed: {}", ledgerId, entryId);
+                            log.debug("Found last entry for ledger {} in write cache being flushed: {}", ledgerId, eId);
                         }
                     }
                     dbLedgerStorageStats.getWriteCacheHitCounter().inc();
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
@@ -148,6 +148,8 @@ public class SingleDirectoryDbLedgerStorage implements CompactableLedgerStorage 
 
     private final Counter flushExecutorTime;
 
+    private boolean isWriteCacheFixedLengthEnabled = false;
+
     public SingleDirectoryDbLedgerStorage(ServerConfiguration conf, LedgerManager ledgerManager,
             LedgerDirsManager ledgerDirsManager, LedgerDirsManager indexDirsManager, StatsLogger statsLogger,
             ByteBufAllocator allocator, ScheduledExecutorService gcExecutor, long writeCacheSize, long readCacheSize,
@@ -155,16 +157,24 @@ public class SingleDirectoryDbLedgerStorage implements CompactableLedgerStorage 
         checkArgument(ledgerDirsManager.getAllLedgerDirs().size() == 1,
                 "Db implementation only allows for one storage dir");
 
+
         String baseDir = ledgerDirsManager.getAllLedgerDirs().get(0).toString();
         log.info("Creating single directory db ledger storage on {}", baseDir);
 
         StatsLogger ledgerDirStatsLogger = statsLogger.scopeLabel("ledgerDir",
                 ledgerDirsManager.getAllLedgerDirs().get(0).getPath());
 
+        this.isWriteCacheFixedLengthEnabled = conf.isWriteCacheFixedLengthEnabled();
         this.writeCacheMaxSize = writeCacheSize;
-        this.writeCache = new WriteCache(allocator, writeCacheMaxSize / 2);
-        this.writeCacheBeingFlushed = new WriteCache(allocator, writeCacheMaxSize / 2);
-        this.writeCacheLastFlushed = new WriteCache(allocator, writeCacheMaxSize / 2);
+        if (this.isWriteCacheFixedLengthEnabled) {
+            this.writeCache = new WriteCache(allocator, writeCacheMaxSize / 3);
+            this.writeCacheBeingFlushed = new WriteCache(allocator, writeCacheMaxSize / 3);
+            this.writeCacheLastFlushed = new WriteCache(allocator, writeCacheMaxSize / 3);
+        } else {
+            this.writeCache = new WriteCache(allocator, writeCacheMaxSize / 2);
+            this.writeCacheBeingFlushed = new WriteCache(allocator, writeCacheMaxSize / 2);
+            this.writeCacheLastFlushed = null;
+        }
         readCacheMaxSize = readCacheSize;
         this.readAheadCacheBatchSize = readAheadCacheBatchSize;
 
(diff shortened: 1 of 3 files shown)


StevenLuMT:
@dlg99 @eolivelli @pkumar-singh @zymap @hangc0276 @lordcheng10
If you have time, please help me review it, thank you.

StevenLuMT:
rerun failure checks

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
```diff
@@ -161,7 +164,7 @@ public SingleDirectoryDbLedgerStorage(ServerConfiguration conf, LedgerManager le
         this.writeCacheMaxSize = writeCacheSize;
         this.writeCache = new WriteCache(allocator, writeCacheMaxSize / 2);
         this.writeCacheBeingFlushed = new WriteCache(allocator, writeCacheMaxSize / 2);
-
+        this.writeCacheLastFlushed = new WriteCache(allocator, writeCacheMaxSize / 2);
```
merlimat, line 167: This will make us use 150% of the configured memory for the write cache.

StevenLuMT: @merlimat 
writeCacheBeingFlushe and writeCacheLastFlushed almost  only one of the two exists at the same time
so I think this just 100% memory for the write cache.
image

StevenLuMT:
rerun failure checks

eolivelli:
the change makes sense to me,  do you have some numbers ?

hangc0276 (changes requested):
The motivation is to increase the write cache hit rate, right? The cost is that it will use more than 50% write cache memory. I suggestion to tune the cache hit rate on the Pulsar broker side instead of the bookie side. 

Otherwise, the write cache holds all the recent data, and the hit rate maybe not be higher than the read cache. I prefer to tune the broker's cache and the bookie's read cache.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
merlimat: The "flush" to disk is not instantaneous. In fact, it can take a very long time and during that time all 3 segments will be filled up during that time.

StevenLuMT:
rerun failure checks

StevenLuMT:
rerun failure checks

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
StevenLuMT: > 

@merlimat 
thanks for your suggestion，I have update the init code
image

StevenLuMT:
rerun failure checks

StevenLuMT:
> the change makes sense to me, do you have some numbers ?

yes, I have added some test data in Descriptions of the pr @eolivelli

StevenLuMT:
> the change makes sense to me, do you have some numbers ?

yes, I have added some test data in Descriptions of the pr @eolivelli ,help me review it ,thanks

StevenLuMT:
rerun failure checks

HQebupt (approved):
Good work.

eolivelli (dismissed):
+1 very nice

StevenLuMT:
rerun failure checks

StevenLuMT:
rerun failure checks

StevenLuMT:
rerun failure checks

StevenLuMT:
rerun failure checks

StevenLuMT:
rerun failure checks

StevenLuMT:
rerun failure checks

StevenLuMT:
rerun failure checks

StevenLuMT:
rerun failure checks

StevenLuMT:
rerun failure checks

StevenLuMT:
rerun failure checks

StevenLuMT:
rerun failure checks

StevenLuMT:
rerun failure checks

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
```diff
@@ -191,8 +204,10 @@ public SingleDirectoryDbLedgerStorage(ServerConfiguration conf, LedgerManager le
 
         dbLedgerStorageStats = new DbLedgerStorageStats(
                 ledgerDirStatsLogger,
-            () -> writeCache.size() + writeCacheBeingFlushed.size(),
-            () -> writeCache.count() + writeCacheBeingFlushed.count(),
+            () -> this.isWriteCacheFixedLengthEnabled ? writeCache.size() + writeCacheBeingFlushed.size()
+                    + writeCacheLastFlushed.size() : writeCache.size() + writeCacheBeingFlushed.size(),
+            () -> this.isWriteCacheFixedLengthEnabled ? writeCache.count() + writeCacheBeingFlushed.count()
+                    + writeCacheLastFlushed.count() : writeCache.count() + writeCacheBeingFlushed.count(),
```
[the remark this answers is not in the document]
StevenLuMT: good suggestion

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
```diff
@@ -793,6 +843,13 @@ public void checkpoint(Checkpoint checkpoint) throws IOException {
         }
     }
 
+    private void cloneFlushedCache2LastFlushCache(WriteCache writeCacheBeingFlushed, WriteCache writeCacheLastFlushed) {
+        writeCacheLastFlushed.clear();
+        writeCacheBeingFlushed.forEach((ledgerId, entryId, data) -> {
```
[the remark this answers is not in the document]
StevenLuMT: > this looks like an expensive operation. Is there any reason to not swap two caches like in `swapWriteCache()`?

I think this way can make your data more safe

StevenLuMT:
rerun failure checks

dlg99 (approved):

StevenLuMT:
> The motivation is to increase the write cache hit rate, right? The cost is that it will use more than 50% write cache memory. I suggestion to tune the cache hit rate on the Pulsar broker side instead of the bookie side.
> 
> Otherwise, the write cache holds all the recent data, and the hit rate maybe not be higher than the read cache. I prefer to tune the broker's cache and the bookie's read cache.

I add a switch ,if user can enable open this feature @hangc0276

merlimat:
@StevenLuMT Since you have now reduce the sizes to 1/3 each, what will happen is that the writes are blocked when we reach 66% of memory instead of 100%. 

I don't really understand why is it a problem for the write cache to fluctuate in size.
The write cache job is to make sure we can decouple write path from the flushing. The read cache job is to provide high hit-rate.

StevenLuMT:
> Since you have now reduce the sizes to 1/3 each, what will happen is that the writes are blocked when we reach 66% of memory instead of 100%.
> 
> I don't really understand why is it a problem for the write cache to fluctuate in size. The write cache job is to make sure we can decouple write path from the flushing. The read cache job is to provide high hit-rate.

I think this is an option when more real-time data needs to be persisted in writecache,instead of triggering read disk to load into readcache @merlimat

hangc0276:
> > Since you have now reduce the sizes to 1/3 each, what will happen is that the writes are blocked when we reach 66% of memory instead of 100%.
> > I don't really understand why is it a problem for the write cache to fluctuate in size. The write cache job is to make sure we can decouple write path from the flushing. The read cache job is to provide high hit-rate.
> 
> I think this is an option when more real-time data needs to be persisted in writecache, instead of triggering read disk to load into readcache @merlimat

@StevenLuMT I agree with Matteo. I do not really understand the motivation of this PR. I have the following concerns.
- From your test data, you compared the cached data size, it certainly will be more than the old one. But the size of cached data is not our goal, but the bookie cache hit rate. We should leverage the gain and the cost. The gain is the hit rate, and the cost is the extra memory used.
- What we are concerned about is whether this change is valuable, not just adding an option to turn it on or off.
- We use 1/3 memory cache to cache the last flushed data. Those data contain all the newly written data, it may not improve the write cache hit rate a lot due to those cached data without purpose. This is why we evict the flushed data from OS PageCache once it is written into the journal disk. If we really need to cache the last flushed data into memory, I prefer to use OS PageCache.
- I'm not sure whether you use BookKeeper based on Pulsar. If you use Pulsar, I prefer to tunning Pulsar broker cache instead of BookKeeper write cache. The Pulsar broker cache and BookKeeper read cache are driven by reading. It will have higher hit rate.

StevenLuMT:
> @StevenLuMT Do we still need this cache optimization?

Close it for now, I'll bring it up later when I have better ideas @hangc0276

PR #3350, opened 2022-06-21
Fix: NPE in RackawareEnsemblePlacementPolicyImpl logged by AutoRecovery

dlg99:
### Motivation

AR keeps on logging NPE during rereplication:

```
[ReplicationWorker] WARN  org.apache.bookkeeper.client.RackawareEnsemblePlacementPolicyImpl - Received exception while trying to get network location of bookie: ...
java.lang.NullPointerException: null 
```

The NPE seems to be easily avoidable, no need to throw exception and log it just to ignore.

### Changes

Added check for bookie being present in the knownBookies map before accessing it.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
@@ -1026,7 +1026,11 @@ public class RackawareEnsemblePlacementPolicyImpl extends TopologyAwareEnsembleP
             for (int j = 0; j < writeQuorumSize; j++) {
                 bookie = ensembleList.get((i + j) % ensembleSize);
                 try {
-                    racksInQuorum.add(knownBookies.get(bookie).getNetworkLocation());
+                    if (knownBookies.containsKey(bookie)) {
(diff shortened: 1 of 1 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
```diff
@@ -1026,7 +1026,11 @@ public PlacementPolicyAdherence isEnsembleAdheringToPlacementPolicy(List<BookieI
             for (int j = 0; j < writeQuorumSize; j++) {
                 bookie = ensembleList.get((i + j) % ensembleSize);
                 try {
-                    racksInQuorum.add(knownBookies.get(bookie).getNetworkLocation());
+                    if (knownBookies.containsKey(bookie)) {
+                        racksInQuorum.add(knownBookies.get(bookie).getNetworkLocation());
```
merlimat, line 1030: nit: we could check on the map one single time: 

```java
String kb = knownBookies.get(bookie);
if (kb != null) {
   ///... 
}
```

merlimat (approved):

hangc0276 (approved):

horizonzy (approved):

PR #3353, opened 2022-06-23
Fix checkpointComplete invoked twice in SyncThread

gaozhangmin:
### Motivation

SyncThread flush would invoke  `checkpointSource.checkpointComplete`  twice. it's unnecessary.  
### Changes

Change `org.apache.bookkeeper.bookie.LedgerStorage#flush` signature . add `doCheckpointComplete` parameter.

default value is true,  only set to false in `org.apache.bookkeeper.bookie.SyncThread#flush`

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerStorage.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerStorage.java
@@ -180,8 +180,9 @@ public interface LedgerStorage {
      * Flushes all data in the storage. Once this is called,
      * add data written to the LedgerStorage up until this point
      * has been persisted to perminant storage
+     * @param doCheckpointComplete
      */
-    void flush() throws IOException;
+    void flush(boolean doCheckpointComplete) throws IOException;
 
     /**
      * Ask the ledger storage to sync data until the given <i>checkpoint</i>.
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SyncThread.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SyncThread.java
@@ -133,7 +133,7 @@ class SyncThread implements Checkpointer {
     private void flush() {
         Checkpoint checkpoint = checkpointSource.newCheckpoint();
         try {
-            ledgerStorage.flush();
+            ledgerStorage.flush(false);
         } catch (NoWritableLedgerDirException e) {
             log.error("No writeable ledger directories", e);
             dirsListener.allDisksFull(true);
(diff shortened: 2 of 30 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
```diff
@@ -820,10 +820,12 @@ private void swapWriteCache() {
     }
 
     @Override
-    public void flush() throws IOException {
+    public void flush(boolean doCheckpointComplete) throws IOException {
         Checkpoint cp = checkpointSource.newCheckpoint();
         checkpoint(cp);
-        checkpointSource.checkpointComplete(cp, true);
+        if (doCheckpointComplete) {
```
hangc0276, line 826: It is a little strange that we flush the WriteCache data into OS PageCache without writing the checkpoint `lastMark` into ledger directory.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/SyncThread.java
```diff
@@ -133,7 +133,7 @@ public Future requestFlush() {
     private void flush() {
         Checkpoint checkpoint = checkpointSource.newCheckpoint();
```
hangc0276, line 134: We create a checkpoint at line 134, and write the checkpoint lastMark position into ledger directory at line 152. There is a gap that the actually flushed lastMark position is less than write cache flushed position.

hangc0276 (commented):
The only impact is that the DbLedgerStorage will write check `lastMark` twice in the checkpoint.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
gaozhangmin: In SyncThread,  We can disableCheckpoint, just return without executing `checkpointSource.checkpointComplete(checkpoint, false);` I don't think this is strange.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerStorage.java
```diff
@@ -180,8 +180,9 @@ void cancelWaitForLastAddConfirmedUpdate(long ledgerId,
      * Flushes all data in the storage. Once this is called,
      * add data written to the LedgerStorage up until this point
      * has been persisted to perminant storage
+     * @param doCheckpointComplete
      */
-    void flush() throws IOException;
+    void flush(boolean doCheckpointComplete) throws IOException;
```
StevenLuMT, line 185: How do we charge what this value(doCheckpointComplete) is passed in?

StevenLuMT:

PR #3354, opened 2022-06-23
Resolve network location if the bookie is down but history holds it

horizonzy:
### Motivation

When resolve a bookie network location, only get from knownBookies. If the bookie shutdown, we can't resolve the network location. In fact, the history bookies maybe hold it.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
@@ -1026,7 +1026,14 @@ public class RackawareEnsemblePlacementPolicyImpl extends TopologyAwareEnsembleP
             for (int j = 0; j < writeQuorumSize; j++) {
                 bookie = ensembleList.get((i + j) % ensembleSize);
                 try {
-                    racksInQuorum.add(knownBookies.get(bookie).getNetworkLocation());
+                    BookieNode bookieNode = knownBookies.get(bookie);
+                    if (bookieNode == null) {
+                        bookieNode = historyBookies.get(bookie);
+                    }
+                    if (bookieNode == null) {
+                        continue;
+                    }
+                    racksInQuorum.add(bookieNode.getNetworkLocation());
                 } catch (Exception e) {
                     /*
                      * any issue/exception in analyzing whether ensemble is


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
```diff
@@ -1026,7 +1026,14 @@ public PlacementPolicyAdherence isEnsembleAdheringToPlacementPolicy(List<BookieI
             for (int j = 0; j < writeQuorumSize; j++) {
                 bookie = ensembleList.get((i + j) % ensembleSize);
                 try {
-                    racksInQuorum.add(knownBookies.get(bookie).getNetworkLocation());
+                    BookieNode bookieNode = knownBookies.get(bookie);
+                    if (bookieNode == null) {
+                        bookieNode = historyBookies.get(bookie);
```
dlg99, line 1031: now the behavior depends on whether the bookie ran for awhile and hs history or freshly restarted.

AFAICT this should affect Auditor which does isEnsembleAdheringToPlacementPolicy but I guess our test coverage for the rackaware policy + bookie removal/Auditor runs is not good enough.

i think the only case where the historyBookies is to protect from DNS resolver errors, this changes expectations.

dlg99 (changes requested):
I think this will break Auditor's behavior and, unfortunately, we don't have enough tests for that

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
horizonzy: If the knownBookies didn't hold bookie, it will thown NPE. Can't resolve it.

horizonzy: And the logic is a little bit strange, if we want to resolve bookie, the bookie must exist in the newest writeable bookies or history bookies. If there is a bookie exist in old ledger's ensembles, the bookie maybe not exist in  the newest writeable bookies or history bookies, so I can't resolve the network location. It's unacceptable.

horizonzy: I report it. #3355

horizonzy: Maybe I got it, if the bookie didn't active anymore, we didn't care the network location it is.

dlg99: Fix for NPE: https://github.com/apache/bookkeeper/pull/3350

Issue #3362, opened 2022-06-25
Race condition problme in AbstractZkLedgerManager


horizonzy:
**BUG REPORT**

There are two methods in AbstractZkLedgerManager.
1. **registerLedgerMetadataListener** 
https://github.com/apache/bookkeeper/blob/677ccec3eb84f5be1b3556537871e14eb5e8359c/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractZkLedgerManager.java#L389-L409

2. **unregisterLedgerMetadataListener**
https://github.com/apache/bookkeeper/blob/677ccec3eb84f5be1b3556537871e14eb5e8359c/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractZkLedgerManager.java#L412-L426

Thread 1: doing unregisterLedgerMetadataListener, it unregister ledger 1.
It get listenerSet and remove watcher, the listenerSet is empty, then remove key ledger 1.

Thread 2: doing registerLedgerMetadataListener, it get same listenerSet, at the same time, thread 1 not remove ledger 1 from listenerSet yet. Thread 2 will add listener to this listenerSet which will be remove from listeners.

Final Result: 
AbstractZkLedgerManager#listeners is empty. In fact, it should hold thread 2 listener.

The pr come later.

PR #3365, opened 2022-06-26
Fix race condition.

horizonzy:
### Motivation
fixes-#3362

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractZkLedgerManager.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractZkLedgerManager.java
@@ -402,6 +402,7 @@ public abstract class AbstractZkLedgerManager implements LedgerManager, Watcher 
                 }
             }
             synchronized (listenerSet) {
+                listenerSet = listeners.computeIfAbsent(ledgerId, k -> new HashSet<>());
                 listenerSet.add(listener);
             }
             new ReadLedgerMetadataTask(ledgerId).run();


hangc0276 (approved):
Nice catch!

Please add more details about how the race condition happened.

horizonzy:
> Nice catch!
> 
> Please add more details about how the race condition happened.

see #3362, update the detail info

zymap (commented):
Overall look good to me. Could you please add a unit test to cover this?

horizonzy:
> Overall look good to me. Could you please add a unit test to cover this?

fine.

Shoothzj (approved):

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractZkLedgerManager.java
```diff
@@ -402,6 +402,7 @@ public void registerLedgerMetadataListener(long ledgerId, LedgerMetadataListener
                 }
             }
             synchronized (listenerSet) {
+                listenerSet = listeners.computeIfAbsent(ledgerId, k -> new HashSet<>());
```
dlg99, line 405: with this, do we still need the lines above:
```
Set<LedgerMetadataListener> listenerSet = listeners.get(ledgerId);
            if (listenerSet == null) {
                Set<LedgerMetadataListener> newListenerSet = new HashSet<LedgerMetadataListener>();
                Set<LedgerMetadataListener> oldListenerSet = listeners.putIfAbsent(ledgerId, newListenerSet);
                if (null != oldListenerSet) {
                    listenerSet = oldListenerSet;
                } else {
                    listenerSet = newListenerSet;
                }
            }
```

dlg99, line 405: I think this should be before the synchronized block, not inside of it

dlg99 (changes requested):
I left couple of comments to address.

As for the original issue, is this a theoretical find or from a production case?
Most of the ledger-specific mutations in BK happen in the ordered executor where the ledgerid is used as a key, so the mutations should happen on the same thread sequentially.

horizonzy:
> I left couple of comments to address.
> 
> As for the original issue, is this a theoretical find or from a production case? Most of the ledger-specific mutations in BK happen in the ordered executor where the ledgerid is used as a key, so the mutations should happen on the same thread sequentially.



> I left couple of comments to address.
> 
> As for the original issue, is this a theoretical find or from a production case? Most of the ledger-specific mutations in BK happen in the ordered executor where the ledgerid is used as a key, so the mutations should happen on the same thread sequentially.

This is not real case, just for analyze. We can't ensure the user open ledger and close ledger in same thread, so if you want to process register and unregister serially, should handle it in AbstractZkLedgerManageer, not outside by invoke user.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractZkLedgerManager.java
horizonzy: > I think this should be before the synchronized block, not inside of it

Maybe not, if outside, the race condition problem still exist.

horizonzy: This logic can be simply.

dlg99: with this
```
            Set<LedgerMetadataListener> listenerSet = listeners.computeIfAbsent(ledgerId, k -> new HashSet<>());
            synchronized (listenerSet) {
                listenerSet = listeners.computeIfAbsent(ledgerId, k -> new HashSet<>());
                listenerSet.add(listener);
            }
```
it looks like the codes assume that between two calls of `listeners.computeIfAbsent` it may need to create a new instance (listeners can be modified/removed concurrently.) 

so you are getting (or creating) one instance to synchronize on, meanwhile, listeners change, now synchronize locks on one instance of listenerSet while modifying another.

so either the second call to computeIfAbsent is not needed or the locking needs to be done differently

horizonzy: If we don't want second `computeIfAbsent` and still lock by ledgerId, we should define other object which is immutable. likes Map<String, Object> key: ledgerId, value: empty object. But it will introduce more object.
The simplest way is make register and unregister synchronize, but it will reduce concurrency.

In summary, I think second `computeIfAbsent` is the relatively good way.

StevenLuMT (approved):
lgtm

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractZkLedgerManager.java
dlg99: Let's use lock object (`private final Object lock = new Object();` is good enough); and replace other instance of `synchronized (listenerSet)` with `synchronized (lock) ...`
Otherwise I think it will just make the problem harder to reproduce

horizonzy: It will reduce too many concurrency. If we listen too many zk nodes, when received event, also synchronized, it maybe make program slow.
see https://github.com/apache/bookkeeper/blob/ded05aa2218f1e98c650070e9d0152213c4d09fe/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractZkLedgerManager.java#L125.

StevenLuMT:

horizonzy:
rerun failure checks

PR #3369, opened 2022-06-27
If ensembleList is empty, return PlacementPolicyAdherence.FAIL.

horizonzy:
If the ensembleList is empty, should not return PlacementPolicyAdherence.MEETS_STRICT.
Return PlacementPolicyAdherence.FAIL to remind user.


dlg99:
I think empty or null ensemble deserves throwing IllegalStateException

horizonzy:
> I think empty or null ensemble deserves throwing IllegalStateException

This may have some risks. When I use it to check in new feature, the ensemble is empry, but it return MEET_STRICT. It may mislead users.

StevenLuMT (approved):
lgtm

StevenLuMT:
> I think empty or null ensemble deserves throwing IllegalStateException

I think return PlacementPolicyAdherence.FAIL is right, this just mark creating policy result @dlg99

horizonzy:
I think it again, throw IllegalStateException maybe better. If the ensemble list is empty, throw exception will be better to notify user.

horizonzy:
> I think it again, throw IllegalStateException maybe better. If the ensemble list is empty, throw exception will be better to notify user.

But if we throw an illegalStateException, I'm worried about it causing compatibility problems.

dlg99 (approved):

zymap (approved):

PR #3382, opened 2022-07-01
Include BK error code in default case for BKException#getMessage

michaeljmarshall:
### Motivation

When a client gets an exception code from the server that is unexpected, it's hard to debug because the logs hide the error code. It'd be helpful to see the error code in those cases.

### Changes

* Make `Unexpected condition` its own case with the error code in the `BKException#getMessage` method.
* Add new default called `Unknown code: <code>`

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/BKException.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/BKException.java
@@ -173,8 +173,10 @@ public class BKException extends Exception {
             return "Failed to serialize metadata";
         case Code.DataUnknownException:
             return "Ledger in limbo";
+        case Code.UnexpectedConditionException:
+            return "Unexpected condition";
         default:
-            return "Unexpected condition: " + code;
+            return "Unknown code: " + code;
         }
     }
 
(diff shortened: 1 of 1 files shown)


hangc0276 (approved):

michaeljmarshall:
I updated the PR because `Unknown Condition` is actually a known error code.

zymap:
@michaeljmarshall Could you please take a look at the failure tests?

michaeljmarshall:
After inspecting the test failures, I think the right solution is to change the downstream usage. https://github.com/apache/bookkeeper/pull/868 introduced better log errors, and it looks like it'd be better to use that feature than the `getMessage` method. @hangc0276 @dlg99 - do you agree or do you think it makes sense to fix tests and merge this PR?

StevenLuMT:
rerun failure checks

StevenLuMT (approved):
lgtm

StevenLuMT:
rerun failure checks

StevenLuMT:
rerun failure checks

StevenLuMT:
rerun failure checks

StevenLuMT:
rerun failure checks

StevenLuMT:
hi @michaeljmarshall please rebase for branch master,all checks will be passed

hangc0276:
> After inspecting the test failures, I think the right solution is to change the downstream usage. #868 introduced better log errors, and it looks like it'd be better to use that feature than the `getMessage` method. @hangc0276 @dlg99 - do you agree or do you think it makes sense to fix tests and merge this PR?

@michaeljmarshall Thanks for your work. I support the log feature introduced by #868

michaeljmarshall:
@hangc0276 - thank you for your review. In August, I plan to open a PR for Pulsar to utilize the method added in #868.

PR #3384, opened 2022-07-03
Switch back ordered executor to LinkedBlockingQueue

merlimat:
### Motivation

In #3108, we have switched to use GrowableArrayBlockingQueue for the OrderedExecutor.

I've been doing some testing recently, with respect with throughput and latencies when there are many small requests being passed on. This was on newer machines and using Java 17, so my conclusion is that many things have changed since when I did add that class back in the day :). 

I've tested with `LinkedBlockingQueue`, `GrowableArrayBlockingQueue`, JCTools based queue and Cognizant disruptor based queue, in the context of the OrderedExecutor, running in a Pulsar broker and BK bookie.

I've actually lost the document with all the results :/, though the summary was that: 
 1. `LinkedBlockingQueue` had the best throughput and latency. Behavior was very stable
 2. `GrowableArrayBlockingQueue` had decent throughput (slightly lower than LBQ) and worse latency, a bit jumpy. 
 3. JCTools and Cognizant disruptor based blocking MP-SC queues all had much worse performance overall, and causing lot of variance in the throughput and latency.
 4. I tried to perform some improvements on `GrowableArrayBlockingQueue` aimed at reducing contention and avoid locking in few cases to assume a single consumer, though I wasn't able to achieve any better result there.

Result: I think we should stick the default LBQ.

cc/ @lhotari

### DIFF
--- a/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/OrderedExecutor.java
+++ b/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/OrderedExecutor.java
@@ -36,6 +36,7 @@ import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
+import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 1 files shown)


hangc0276 (approved):

dlg99 (approved):

dlg99:
@merlimat it looks like OrderedExecutor is the only place where the GrowableArrayBlockingQueue was used. We can safely delete GrowableArrayBlockingQueue and related tests, as a follow up PR.

eolivelli (approved):

PR #3385, opened 2022-07-06
Remove microbenchmark module

Shoothzj:
### Motivation
`microbenchmark` module depends on `jmh-core` which is license by GPLv2. 
For more informations, See mail list https://lists.apache.org/thread/sp8jf5rxvkzdvl4vy8chcqwkdmppwd6t

### Changes
- remove microbenchmark module


michaeljmarshall (approved):

kezhenxu94:
Hi, from my understanding, if you didn't include the related binary jars in your binary distribution package and you didn't copy their source codes into your source package, it's OK to have these libs as build time dependencies according to
 https://www.apache.org/legal/resolved.html#prohibited

> Apache projects may not distribute Category X licensed components, in source or binary form; in ASF source code or in convenience binaries. As with the previous question on platforms, you can rely on the component if its license terms do not affect the Apache product's licensing. For example, using a GPL'ed tool during the build is okay, but including GPL'ed source code is not.

dave2wave:
Calvin's analysis is correct. As long as the GPL code is used as tooling and is not included in the source the use of GPL is fine.

dlg99 (changes requested):
No need to remove, jmh is used as tooling / to run the tests and should not be packaged with the releases

PR #3388, opened 2022-07-06
Add static declaration for useFallocate in DirectWriter

wenbingshen:
### Motivation

The work of the `BP-47` has been very meaningful. I am reading and learning it.

As the title, `useFallocate` in `DirectWriter` has no practical meaning, because the judgment of SystemUtils.IS_OS_LINUX always happens.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/directentrylogger/DirectWriter.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/directentrylogger/DirectWriter.java
@@ -77,7 +77,7 @@ class DirectWriter implements LogWriter {
 
         if (useFallocate) {
             if (!SystemUtils.IS_OS_LINUX) {
-                useFallocate = false;
+                disableUseFallocate();
                 slog.warn(Events.FALLOCATE_NOT_AVAILABLE);
             } else {
                 try {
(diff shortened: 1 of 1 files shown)


wenbingshen:
@hangc0276 @dlg99 @eolivelli Can you help take a look? Thanks.

wenbingshen:
Or `useFallocate` should be defined with `static`.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/directentrylogger/DirectWriter.java
```diff
@@ -49,7 +49,6 @@ class DirectWriter implements LogWriter {
     final List<Future<?>> outstandingWrites = new ArrayList<Future<?>>();
     Buffer nativeBuffer;
     long offset;
-    private volatile boolean useFallocate = true;
```
hangc0276, line 52: It's better to use `static`

hangc0276 (changes requested):

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/directentrylogger/DirectWriter.java
wenbingshen: It makes sense to me. I have addressed your comment. PTAL.

wenbingshen:
PTAL. @zymap Thanks.

dlg99 (approved):

PR #3407, opened 2022-07-16
Move exceptions out of `Bookie` interface.

mattisonchao:
### Motivation

Follow the to-do list as below, and move the exception `NoLedgerException` and `NoEntryException` out of the `Bookie` interface.

> Move this exceptions somewhere else

### Changes

- move the exception `NoLedgerException` and `NoEntryException` out of the `Bookie` interface.


mattisonchao:
rerun failure checks

mattisonchao:
@nicoloboschi 
I just want to remove the to-do in the codes. If it's not a good idea, I can close it.

PR #3416, opened 2022-07-21
Fix configuration cannot be leave unset in bk_server.conf

gaozhangmin:
### Motivation
We cannot leave  `dbStorage_readAheadCacheBatchSize` `dbStorage_maxThrottleTimeMs` unset like  `dbStorage_writeCacheMaxSizeMb`

### Changes
If configuration is not set in bk_server.conf, default value will be used.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
@@ -174,7 +174,8 @@ public class DbLedgerStorage implements LedgerStorage {
 
         long perDirectoryWriteCacheSize = writeCacheMaxSize / numberOfDirs;
         long perDirectoryReadCacheSize = readCacheMaxSize / numberOfDirs;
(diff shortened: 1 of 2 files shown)


StevenLuMT (changes requested):
no testcase for the new function(getIntVariableOrDefault)?

dlg99:
`ServerConfiguration` already has
```
public long getLong(String key, long defaultValue) 
```
and similarly for Int, etc.

why do we need these helper methods (and, therefore, unit tests for them)?

hangc0276:
> `ServerConfiguration` already has
> 
> ```
> public long getLong(String key, long defaultValue) 
> ```
> 
> and similarly for Int, etc.
> 
> why do we need these helper methods (and, therefore, unit tests for them)?

@gaozhangmin Would you please take a look at this comment?

gaozhangmin:
@hangc0276  @dlg99 I will close this pr, Comparing with `dbStorage_writeCacheMaxSizeMb `,it's hard to set a  default value in bk_server.conf for it (it use a proportion of direct memory). So it need to be allowed unset.  
but dbStorage_readAheadCacheBatchSize and dbStorage_maxThrottleTimeMs we should set a default value in bk_server.conf for then.

PR #3417, opened 2022-07-21
Optimize concurrent collection's shrink logic

wenbingshen:
### Motivation

Optimize concurrent collection's shrink and clear logic

### Changes
1. Reduce the repeated `Arrays.fill` in the clear process
2. When `capacity` is already equal to `initCapacity`,`rehash` should not be executed
3. Reduce the `rehash` logic in the `clear` process
4. Shrinking must at least ensure `initCapacity`, so as to avoid frequent shrinking and expansion near `initCapacity`, frequent shrinking and expansion, additionally opened `arrays` will consume more memory and affect GC.

If this PR is accepted, I will optimize the same `concurrent collection's shrink and clear logic ` defined in pulsar.

Related to #3061 and #3074

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/util/collections/ConcurrentLongHashMap.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/util/collections/ConcurrentLongHashMap.java
@@ -507,7 +507,11 @@ public class ConcurrentLongHashMap<V> {
             } finally {
                 if (autoShrink && size < resizeThresholdBelow) {
                     try {
-                        int newCapacity = alignToPowerOfTwo((int) (capacity / shrinkFactor));
+                        // Shrinking must at least ensure initCapacity,
+                        // so as to avoid frequent shrinking and expansion near initCapacity,
+                        // frequent shrinking and expansion,
+                        // additionally opened arrays will consume more memory and affect GC
+                        int newCapacity = Math.max(alignToPowerOfTwo((int) (capacity / shrinkFactor)), initCapacity);
                         int newResizeThresholdUp = (int) (newCapacity * mapFillFactor);
                         if (newCapacity < capacity && newResizeThresholdUp > size) {
                             // shrink the hashmap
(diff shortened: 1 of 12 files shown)


wenbingshen:
Ping @eolivelli @lordcheng10 @hangc0276 PTAL, Thanks.

wenbingshen:
rerun failure checks

wenbingshen:
ping @eolivelli PTAL.

wenbingshen:
ping @dlg99 PTAL.

wenbingshen:
ping @Shoothzj PTAL.

Shoothzj (approved):

dlg99 (approved):

StevenLuMT (approved):
nice

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/collections/ConcurrentLongHashMap.java
```diff
@@ -507,7 +507,11 @@ private V remove(long key, Object value, int keyHash) {
             } finally {
                 if (autoShrink && size < resizeThresholdBelow) {
                     try {
-                        int newCapacity = alignToPowerOfTwo((int) (capacity / shrinkFactor));
+                        // Shrinking must at least ensure initCapacity,
+                        // so as to avoid frequent shrinking and expansion near initCapacity,
+                        // frequent shrinking and expansion,
+                        // additionally opened arrays will consume more memory and affect GC
+                        int newCapacity = Math.max(alignToPowerOfTwo((int) (capacity / shrinkFactor)), initCapacity);
```
zymap, line 514: Why not just skip shrink when the size is smaller than the alignToPowerOfTwo(initCapacity)? That would save a lot of computing of `alignToPowerOfTwo((int) (capacity / shrinkFactor))`

wenbingshen: Assuming that `initCapacity` is 256,`mapFillFactor, mapIdleFactor, expandFactor, shrinkFactor ` keep the default values:

| initCapacity | capacity | mapFillFactor | mapIdleFactor | resizeThresholdUp | resizeThresholdBelow | expandFactor | shrinkFactor |
| -----------  | -------- | ------------- | ------------- | ----------------- | -------------------- | ------------ | ------------ |
|    256       |    256   |     0.66f     |     0.15f     |         168       |           38         |      2       |        2     |
|    256       |    512   |     0.66f     |     0.15f     |         337       |           76         |      2       |        2     |
|    256       |    1024  |     0.66f     |     0.15f     |         675       |           153        |      2       |        2     |
|    256       |    2048  |     0.66f     |     0.15f     |         1351      |           307        |      2       |        2     |

When the `capacity` is expanded to 1024, the `size` needs to be less than 153 to shrink, and size=153 is less than initCapacity=256, causing at least 1024-153 of space wasted. This waste is worse when initCapacity is larger.

Therefore, we cannot use `size` to judge that it will not shrink when it is smaller than `initCapacity`.

@zymap PTAL

wenbingshen: This calculation will only happen if `size < resizeThresholdBelow`, it is worthwhile to use the occasional calculation to relieve the free space that is not being used.

zymap: Thanks for your detailed explanation!

zymap (approved):

PR #3415, opened 2022-07-21
Issue 3206: Flaky-test: BookieZKExpireTest.testBookieServerZKExpireBehaviour

dlg99:
### Motivation

Flaky test

### Changes

AFAICT from [my repro](https://github.com/apache/bookkeeper/issues/3206#issuecomment-1190618336), the test flakes depending whether the ZK client gets the SessionExpired error or not.
Reworked the test to reflect the behavior, added some logging.

Master Issue: #3206

Issue #3206, opened 2022-04-13
Flaky-test: BookieZKExpireTest.testBookieServerZKExpireBehaviour


hangc0276:
When running CI, the test occasionally with the following exception
```
Error:  org.apache.bookkeeper.test.BookieZKExpireTest.testBookieServerZKExpireBehaviour  Time elapsed: 16.286 s  <<< FAILURE!
java.lang.AssertionError: Bookie should not shutdown on losing zk session
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.apache.bookkeeper.test.BookieZKExpireTest.testBookieServerZKExpireBehaviour(BookieZKExpireTest.java:103)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:298)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:292)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.lang.Thread.run(Thread.java:829)
```

dlg99:
Fixed

PR #3418, opened 2022-07-23
fix Flaky-test: BookieZKExpireTest.testBookieServerZKSessionExpireBehaviour

wenbingshen:
### Motivation

fix Flaky-test: `BookieZKExpireTest.testBookieServerZKSessionExpireBehaviour` by disable retry session expired.

issue #3206  was fixed by PR #3415, but it still introduced another flaky test: `BookieZKExpireTest.testBookieServerZKSessionExpireBehaviour`

According to the investigation, the reason is: 
`ZookeeperClient` still creates a new `zookeeper` instance when the old zookeeper client session time out.

Due to the asynchronous execution of two threads executing bookie temporary node re-registration and zk instance re-creation, the test program sometimes succeeds and sometimes fails.

1. When the temporary node re-registration is performed before the zk re-instantiation, the temporary node creation will use the old zk instance, which will cause a session timeout error, the bookie service will be shutdown, and the test will be successful;
 
2. When the zk re-instantiation precedes the re-registration of the temporary node, the temporary node creation will use the new re-instantiated zk instance, then the temporary node will be successfully created, the bookie service is running normally, and the test fails.

```java
        try {
            connectExecutor.submit(clientCreator);
        } catch (RejectedExecutionException ree) {
            if (!closed.get()) {
                logger.error("ZooKeeper reconnect task is rejected : ", ree);
            }
        } catch (Exception t) {
            logger.error("Failed to submit zookeeper reconnect task due to runtime exception : ", t);
        }
```

### Changes
Add a `retryExpired` flag to indicate whether to run the zk instance and retry to create a new instance after the session times out.
Set this flag to false for `ZKMetadataBookieDriver`;
Other ZookeeperClient and normal ZookeeperClient applications will generate the default value true or set to true, which is consistent with the original behavior.

### Test the behavior of this PR:
**Before this PR:**
Executed the test 10 times, all failed.
image

**After this PR:**
Executed the test 10 times, all successful.
image

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/zk/ZKMetadataDriverBase.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/zk/ZKMetadataDriverBase.java
@@ -165,7 +165,8 @@ public class ZKMetadataDriverBase implements AutoCloseable {
     protected void initialize(AbstractConfiguration<?> conf,
                               StatsLogger statsLogger,
                               RetryPolicy zkRetryPolicy,
-                              Optional<Object> optionalCtx) throws MetadataException {
+                              Optional<Object> optionalCtx,
+                              boolean zkRetryExpired) throws MetadataException {
         this.conf = conf;
         this.acls = ZkUtils.getACLs(conf);
 
@@ -213,6 +214,7 @@ public class ZKMetadataDriverBase implements AutoCloseable {
                     .operationRetryPolicy(zkRetryPolicy)
                     .requestRateLimit(conf.getZkRequestRateLimit())
                     .statsLogger(statsLogger)
+                    .retryExpired(zkRetryExpired)
                     .build();
 
                 if (null == zk.exists(bookieReadonlyRegistrationPath, false)) {
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/zookeeper/ZooKeeperClient.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/zookeeper/ZooKeeperClient.java
@@ -77,6 +77,7 @@ public class ZooKeeperClient extends ZooKeeper implements Watcher, AutoCloseable
     private final String connectString;
     private final int sessionTimeoutMs;
     private final boolean allowReadOnlyMode;
+    private final boolean retryExpired;
 
     // state for the zookeeper client
     private final AtomicReference<ZooKeeper> zk = new AtomicReference<ZooKeeper>();
@@ -173,6 +174,7 @@ public class ZooKeeperClient extends ZooKeeper implements Watcher, AutoCloseable
         int retryExecThreadCount = DEFAULT_RETRY_EXECUTOR_THREAD_COUNT;
         double requestRateLimit = 0;
         boolean allowReadOnlyMode = false;
+        boolean retryExpired = true;
 
         private Builder() {}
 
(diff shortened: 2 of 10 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/zookeeper/ZooKeeperClient.java
```diff
@@ -356,6 +366,13 @@ private void onExpired() {
 
         logger.info("ZooKeeper session {} is expired from {}.",
                 Long.toHexString(getSessionId()), connectString);
+
+        if (!retryExpired) {
+            logger.warn("ZooKeeper session expired retries have been disabled");
+            return;
+        }
+
```
wenbingshen, line 374: Recreating new zookeeper instances is prohibited here.

bookkeeper-server/src/test/java/org/apache/bookkeeper/test/BookieZKExpireTest.java
```diff
@@ -177,13 +179,15 @@ conf, new TestBookieImpl(conf),
             log.info("Resuming threads");
             sendthread.resume();
 
-            // allow watcher thread to run
-            Thread.sleep(3000);
+            // allow client.waitForConnection() timeout
+            Thread.sleep(10000);
             assertFalse("Bookie should shutdown on losing zk session", server.isBookieRunning());
             assertFalse("Bookie Server should shutdown on losing zk session", server.isRunning());
```
wenbingshen, line 185: The increase from 3 seconds to 10 seconds is because creating a bookie temporary node needs to wait for the old zk client to establish a connection with the server. There is a timeout period of conf.getZkTimeout()=6 seconds. Once the timeout period expires, the bookie service will Closed due to session timeout, the extra 4 seconds is to consolidate the stability of the test and give the bookie service a little extra time to perform the shutdown.

wenbingshen:
ping @dlg99 @hangc0276 @Shoothzj @eolivelli PTAL.

HQebupt:
@StevenLuMT PTAL

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/zk/ZKMetadataDriverBase.java
```diff
@@ -165,7 +165,8 @@ public String getScheme() {
     protected void initialize(AbstractConfiguration<?> conf,
                               StatsLogger statsLogger,
                               RetryPolicy zkRetryPolicy,
-                              Optional<Object> optionalCtx) throws MetadataException {
+                              Optional<Object> optionalCtx,
+                              boolean zkRetryExpired) throws MetadataException {
```
StevenLuMT, line 169: this check(zkRetryExpired) is difficult to understand, what scene is set to true and what scene is set to false

StevenLuMT (changes requested):
Is there a better way to fix it?

dlg99:
can you predictably repro the problem? Maybe my fix was not good enough :( but I only could repro it under docker with limited number of CPUs and after the fix it did not fail after running in the loop 150 times.
BookieStateManager's registration listener will shutdown the bookie if it cannot reconnect immediately so I guess there is still some timing issue between it and the test killing zk session / letting it recover.

StevenLuMT:

poorbarcode:
@dlg99 

> can you predictably repro the problem? Maybe my fix was not good enough :( but I only could repro it under docker with limited number of CPUs and after the fix it did not fail after running in the loop 150 times.
BookieStateManager's registration listener will shutdown the bookie if it cannot reconnect immediately so I guess there is still some timing issue between it and the test killing zk session / letting it recover.

When I run the test in my local, it is very flaky

hangc0276:
> @dlg99
> 
> > can you predictably repro the problem? Maybe my fix was not good enough :( but I only could repro it under docker with limited number of CPUs and after the fix it did not fail after running in the loop 150 times.
> > BookieStateManager's registration listener will shutdown the bookie if it cannot reconnect immediately so I guess there is still some timing issue between it and the test killing zk session / letting it recover.
> 
> When I run the test in my local, it is very flaky

@poorbarcode There is a race condition describe in the motivation.

Issue #2665, opened 2021-03-29
ReadOnlyBookieTest is flaky


dlg99:
**BUG REPORT**

***Describe the bug***

ReadOnlyBookieTest is flaky on current master branch

***To Reproduce***

mvn clean install -DskipTests
cd bookkeeper-server/
run a few times:
 mvn test -Dtest=org.apache.bookkeeper.test.ReadOnlyBookieTest

```
[INFO] Running org.apache.bookkeeper.test.ReadOnlyBookieTest
[ERROR] Tests run: 6, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 31.14 s <<< FAILURE! - in org.apache.bookkeeper.test.ReadOnlyBookieTest
[ERROR] org.apache.bookkeeper.test.ReadOnlyBookieTest.testBookieContinueWritingIfMultipleLedgersPresent  Time elapsed: 4.151 s  <<< ERROR!
org.apache.bookkeeper.client.BKException$BKNotEnoughBookiesException: Not enough non-faulty bookies available
	at org.apache.bookkeeper.client.SyncCallbackUtils.finish(SyncCallbackUtils.java:83)
	at org.apache.bookkeeper.client.SyncCallbackUtils$SyncAddCallback.addComplete(SyncCallbackUtils.java:251)
	at org.apache.bookkeeper.client.AsyncCallback$AddCallback.addCompleteWithLatency(AsyncCallback.java:92)
	at org.apache.bookkeeper.client.PendingAddOp.submitCallback(PendingAddOp.java:431)
	at org.apache.bookkeeper.client.LedgerHandle.errorOutPendingAdds(LedgerHandle.java:1799)
	at org.apache.bookkeeper.client.LedgerHandle$5.safeRun(LedgerHandle.java:574)
	at org.apache.bookkeeper.common.util.SafeRunnable.run(SafeRunnable.java:36)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	at java.base/java.lang.Thread.run(Thread.java:834)

[INFO]
[INFO] Results:
[INFO]
[WARNING] Flakes:
[...]

PR #3421, opened 2022-07-24
fix Flaky-test: testBookieContinueWritingIfMultipleLedgersPresent

wenbingshen:
### Motivation

Fixes issue #2665 

When the disk is full, `InterleavedLedgerStorage` creates a new entryLog and index file, copies and moves the old index (newFile) `FileInfo` to the new directory, and finally deletes the old index file.

The old file of `FileInfo` was `deleted`, but it was copied and moved a newFile to the new directory. Therefore, the delete status flag should be restored. If it is not set to false,`checkOpen` will throw `FileInfoDeletedException` when judging the fence status of `ledgerHandle` when adding an entry, causing the writing to fail. When the client changes the `ensemble`, it throws an exception because the test has only two bookie nodes and cannot satisfy the two writeSet: `Not enough non-faulty bookies available`.
```java
    public synchronized boolean isFenced() throws IOException {
        checkOpen(false);
        return (stateBits & STATE_FENCED_BIT) == STATE_FENCED_BIT;
    }

    private synchronized void checkOpen(boolean create, boolean openBeforeClose)
            throws IOException {
        if (deleted) {
            throw new FileInfoDeletedException();
        }
      .....
    }
``` 

### Changes
When `FileInfo` moves to a new directory by `moveToNewLocation`,delete the old index file and restore the `delete` flag, because the new index file is ready.

Set the disk check interval to the `Integer.MAX_VALUE`, which is stable 
`assertEquals("writable dirs should have one dir", 1, ledgerDirsManager
                .getWritableLedgerDirs().size());`
Because once the test thrashes, the disk check thread will restore the full disk to a writable state.
```java
newConf.setDiskCheckInterval(Integer.MAX_VALUE);
```

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileInfo.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileInfo.java
@@ -549,6 +549,7 @@ class FileInfo extends Watchable<LastAddConfirmedUpdateNotification> {
         }
         fc = new RandomAccessFile(newFile, mode).getChannel();
         lf = newFile;
+        deleted = false;
     }
 
     public synchronized byte[] getMasterKey() throws IOException {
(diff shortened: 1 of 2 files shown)


wenbingshen:
ping @hangc0276 @dlg99 @Shoothzj @eolivelli PTAL.

Shoothzj (approved):

wenbingshen:
rerun failure checks

dlg99 (approved):

hangc0276 (approved):

Issue #3422, opened 2022-07-25
Flaky-test: org.apache.bookkeeper.test.ReadOnlyBookieTest#testBookieShouldTurnWritableFromReadOnly


wenbingshen:
Got this on CI: https://github.com/apache/bookkeeper/runs/7487903809?check_suite_focus=true
```
Warning:  Flakes: 
Warning:  org.apache.bookkeeper.test.ReadOnlyBookieTest.testBookieShouldTurnWritableFromReadOnly
Error:    Run 1: ReadOnlyBookieTest.testBookieShouldTurnWritableFromReadOnly:141 Should fail to create a ledger since there isn't enough bookies alive.
```

PR #3423, opened 2022-07-25
fix Flaky-test: testBookieShouldTurnWritableFromReadOnly

wenbingshen:
### Motivation
Fixes issue #3422 

1.`waitForReadOnlyBookie` adds another `listener` thread to observe the node status of bookie, which may be out of sync with the triggering of node changes(`onClusterChanged`) in `EnsemblePlacementPolicy`.This `sequence` leads to flaky test. 

2.Expose bookie nodes in EnsemblePlacementPolicy to client.

3.Just change from watching zk to thread sleep.

We currently don't have any additional requirements for getting bookie nodes on the client side, so it's better to just use thread sleep here. 

### Changes
Change from `waitForReadOnlyBookie` and `waitForWritableBookie` to thread `sleep`.
3 seconds is enough for most.

### DIFF
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/test/ReadOnlyBookieTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/test/ReadOnlyBookieTest.java
@@ -25,7 +25,6 @@ import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 import java.io.File;
-import java.time.Duration;
 import java.util.Enumeration;
 import java.util.concurrent.TimeUnit;
 import org.apache.bookkeeper.bookie.Bookie;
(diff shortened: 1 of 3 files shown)


wenbingshen:
ping @hangc0276 @Shoothzj @dlg99 @eolivelli @nicoloboschi PTAL.

wenbingshen:
rerun failure checks

bookkeeper-server/src/test/java/org/apache/bookkeeper/test/ReadOnlyBookieTest.java
```diff
@@ -128,8 +129,10 @@ public void testBookieShouldTurnWritableFromReadOnly() throws Exception {
             // Expected
         }
 
-        bkc.waitForReadOnlyBookie(BookieImpl.getBookieId(confByIndex(1)))
-            .get(30, TimeUnit.SECONDS);
+        // waitForReadOnlyBookie adds another listener thread to observe the node status of bookie,
+        // which may be out of sync with the triggering of node changes in EnsemblePlacementPolicy.
+        // This sequence leads to flaky test. So change from watching zk to thread sleep.
+        Thread.sleep(3000);
```
hangc0276, line 135: Would you please use `Awayatility` instead of sleep ?

wenbingshen: sure. I have addressed your comments. PTAL.

wenbingshen:
reopen trigger CI again

wenbingshen:
rerun failure checks

hangc0276 (approved):

PR #3431, opened 2022-07-28
Fix  cookie compatible problem. 

### LINKED ISSUE #3430: Cookie compatible problem.
**BUG REPORT**

In #3372, we introduce indexDirs in cookie. When the bookie start, it will check the indexDirs, there is a problem.

If the user already register cookie to RM before this pr, the cookie didn't contains indexDirs, cause the before version didn't contains property `indexDirs`, although user config `indexDirs` in the ServerConfiguration.

After this pr, the bookie restart, it will config indexDirs at local, when it fetch remote cookie from rm, the response cookie didn't contains indexDirs, so the check indexDirs will fail.

This is a breaking compatible problem.

https://github.com/apache/bookkeeper/blob/442e3bbad384fba9b09f8ea774e28562a8d5c6d7/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Cookie.java#L139-L153

At line 144, `null == c.indexDirs` is true.

horizonzy:
### Motivation
Fixes #3430, make cookie validation is compatible with old version.

### DIFF
--- a/bookkeeper-proto/src/main/proto/DataFormats.proto
+++ b/bookkeeper-proto/src/main/proto/DataFormats.proto
@@ -83,6 +83,7 @@ message CookieFormat {
     required string ledgerDirs = 3;
     optional string instanceId = 4;
     optional string indexDirs = 5;
+    optional bool isNew = 6;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


horizonzy:
@hangc0276 pls check it.

StevenLuMT:
I have communicated with @horizonzy 
there is another way to fix it,I am writing the code and will submit a pr later

horizonzy:
use #3433

PR #3444, opened 2022-08-10
Add latency stats for entry location index lookup so that possible RocksDB bottleneck can be detected

lhotari:
### Motivation

- this metric will help to detect when the bottleneck is in the entry location index lookup operations
  and RocksDB tuning is needed

### Changes

Add new operation latency metrics `lookup-entry-location` which will be provided at the same level of metrics as the existing `entries-count` metric.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/EntryLocationIndex.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/EntryLocationIndex.java
@@ -32,6 +32,7 @@ import org.apache.bookkeeper.bookie.storage.ldb.KeyValueStorage.Batch;
 import org.apache.bookkeeper.bookie.storage.ldb.KeyValueStorageFactory.DbConfigType;
 import org.apache.bookkeeper.conf.ServerConfiguration;
 import org.apache.bookkeeper.stats.StatsLogger;
+import org.apache.bookkeeper.util.MathUtils;
 import org.apache.bookkeeper.util.collections.ConcurrentLongHashSet;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -73,6 +74,8 @@ public class EntryLocationIndex implements Closeable {
         LongPairWrapper key = LongPairWrapper.get(ledgerId, entryId);
         LongWrapper value = LongWrapper.get();
 
+        long startTimeNanos = MathUtils.nowInNano();
+        boolean operationSuccess = false;
         try {
             if (locationsDb.get(key.array, value.array) < 0) {
                 if (log.isDebugEnabled()) {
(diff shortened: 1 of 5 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/EntryLocationIndex.java
```diff
@@ -73,18 +74,28 @@ public long getLocation(long ledgerId, long entryId) throws IOException {
...
                 }
                 return 0;
             }
-
+            operationSuccess = true;
             return value.getValue();
         } finally {
             key.recycle();
             value.recycle();
+            long eventLatencyMillis = MathUtils.elapsedNanos(startTimeNanos);
```
michaeljmarshall, line 91: Nit
```suggestion
            long eventLatencyNanos = MathUtils.elapsedNanos(startTimeNanos);
```

michaeljmarshall (changes requested):

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/EntryLocationIndex.java
StevenLuMT, line 91: I think no need declare a tmp eventLatency,just write the code like :
```
if (operationSuccess) {
                stats.getLookupEntryLocationStats()
                        .registerSuccessfulEvent(MathUtils.elapsedNanos(startTimeNanos), TimeUnit.NANOSECONDS);
            } else {
                stats.getLookupEntryLocationStats()
                        .registerFailedEvent(MathUtils.elapsedNanos(startTimeNanos), TimeUnit.NANOSECONDS);
            }
```

lhotari: Thanks for catching that. I'll inline it as suggested by @StevenLuMT .

lhotari: thanks for the suggestion. done.

lhotari: @michaeljmarshall  I now noticed where the name `eventLatencyMillis` came from. This was from the OpStatsLogger interface methods where the parameter was named `eventLatencyMillis`. When I extracted the variable with IntelliJ extract variable refactoring, it suggested `eventLatencyMillis` as the variable name. I'm fixing that inconsistency as part of this PR.

michaeljmarshall (approved):

zymap (approved):

StevenLuMT (approved):
LGTM

hangc0276 (commented):
We already have the index lookup stats `read-locations-index-time`, does it has any case this metric can't cover?
https://github.com/apache/bookkeeper/blob/de5d45fb823dc558f15bcb9d958b9c0a5144043c/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorageStats.java#L48

lhotari:
> We already have the index lookup stats `read-locations-index-time`, does it has any case this metric can't cover?
> 
> https://github.com/apache/bookkeeper/blob/de5d45fb823dc558f15bcb9d958b9c0a5144043c/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorageStats.java#L48

thanks for pointing that out @hangc0276 . 

The problem with the `read-locations-index-time` is that it's not a latency metric. it's a counter metric that records the total time spent for `getLocationIndex` calls. 
I also noticed that the metric is not used for the getLocation call in entryExists method: https://github.com/apache/bookkeeper/blob/4ee8c0411878689417c39b7027899ed7ce815e7e/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java#L344-L348 .

eolivelli (approved):
LGTM

PR #2432, opened 2020-10-08
ISSUE 2377 : Relax grpcio requirement for python client to support python3.8 on mac os

ca-simone-chiorazzo:
Descriptions of the changes in this PR:

Require a grpcio version that is minor of 1.28. This allow to install the 1.27.2 that has a pre-compiled wheel for python 3.8 on mac os.
Add tests for python 3.8

### Motivation

grpcio 1.25.0 does not allow the installation using python 3.8 on mac os x. This is due to the fact that the pre-compiled wheels have not been published before 1.26.0

### Changes

Updated the setup.py to relax a bit the requirement for grpcio
Add python3.8 as tested version
Updated the setup.py to install the latest pymmh3 version. **Note**: master tests are broken, I've updated the library and removed the 'signed' parameter that does not exists not anymore.

Master Issue: #<master-issue-number>

cc @eolivelli 

> ---
> In order to uphold a high standard for quality for code contributions, Apache BookKeeper runs various precommit
> checks for pull requests. A pull request can only be merged when it passes precommit checks.
>
> ---
> Be sure to do all of the following to help us incorporate your contribution
> quickly and easily:
>
> If this PR is a BookKeeper Proposal (BP):
>
> - [ ] Make sure the PR title is formatted like:
>     `<BP-#>: Description of bookkeeper proposal`
>     `e.g. BP-1: 64 bits ledger is support`
> - [ ] Attach the master issue link in the description of this PR.
> - [ ] Attach the google doc link if the BP is written in Google Doc.
>
> Otherwise:
> 
> - [ ] Make sure the PR title is formatted like:
>     `<Issue #>: Description of pull request`
>     `e.g. Issue 123: Description ...`
> - [ ] Make sure tests pass via `mvn clean apache-rat:check install spotbugs:check`.
> - [ ] Replace `<Issue #>` in the title with the actual Issue number.
> 
> ---

PR #3449, opened 2022-08-18
Bump grpcio dep version of Python client for supporting ARM

tisonkun:
Reference:

* https://github.com/apache/bookkeeper/pull/2432
* https://github.com/apache/pulsar/issues/7477
* https://github.com/grpc/grpc/pull/25418
* https://github.com/apache/pulsar/discussions/17166#discussioncomment-3424343

In the previous PR https://github.com/apache/bookkeeper/pull/2432 we "release" the version range instead of requiring a minimal version. It seems incorrect and if we'd like to support M1, it seems at least 1.37 is required. Send this patch for testing.

### DIFF
--- a/stream/clients/python/setup.py
+++ b/stream/clients/python/setup.py
@@ -31,7 +31,7 @@ dependencies = [
     'six>=1.10.0',
     'pytz',
     'futures>=3.2.0;python_version<"3.2"',
-    'grpcio<1.28,>=1.8.2',
+    'grpcio>=1.37',
     'pymmh3>=0.0.5'
 ]
 extras = {


stream/clients/python/setup.py
```diff
@@ -31,7 +31,7 @@
     'six>=1.10.0',
     'pytz',
     'futures>=3.2.0;python_version<"3.2"',
-    'grpcio<1.28,>=1.8.2',
+    'grpcio>=1.37',
```
eolivelli, line 34: My understanding is that we have to put a cap here  otherwise  auto upgrades may break the client.
IIRC it happened in the past.

I I am mot sure that we have CI that is covering this change.

tisonkun: Thank you! I threw the change here to see if it breaks anything. To prepare a solid patch I need more time to understand the mechanism and see how to verify the change.

tisonkun:
Closed for now.

merlimat:
@tisonkun I think we should re-open this, and move to newer Grpc. Otherwise the Pulsar Python client is still stuck with unsupported Grpc version

merlimat:
My bad, this was already addressed in #3875

PR #3471, opened 2022-09-05
add log's dependency for bookkeeper-dist-server

StevenLuMT:

### Motivation

like pulsar's distribution/server
add log's dependency in bookkeeper-dist/server 

### Changes

1.  add log's dependency in bookkeeper-dist/server/pom.xml
2. then auto add log's dependency to bookkeeper-server-*.*.*-SNAPSHOT-bin.tar.gz

### DIFF
--- a/bookkeeper-dist/server/pom.xml
+++ b/bookkeeper-dist/server/pom.xml
@@ -71,12 +71,6 @@
 
     <dependency>
       <groupId>org.apache.logging.log4j</groupId>
-      <artifactId>log4j-1.2-api</artifactId>
-      <version>${log4j.version}</version>
-    </dependency>
-
-    <dependency>
-      <groupId>org.apache.logging.log4j</groupId>
       <artifactId>log4j-api</artifactId>
       <version>${log4j.version}</version>
     </dependency>
--- a/bookkeeper-dist/server/pom.xml
+++ b/bookkeeper-dist/server/pom.xml
@@ -31,7 +31,80 @@
   <packaging>jar</packaging>
   <name>Apache BookKeeper :: Dist (Server)</name>
 
+  <properties>
+    <disruptor.version>3.4.0</disruptor.version>
+  </properties>
+
   <dependencies>
+    <dependency>
+      <groupId>org.yaml</groupId>
+      <artifactId>snakeyaml</artifactId>
+    </dependency>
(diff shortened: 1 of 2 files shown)


StevenLuMT:
rerun failure checks

bookkeeper-dist/server/pom.xml
```diff
@@ -31,7 +31,80 @@
   <packaging>jar</packaging>
   <name>Apache BookKeeper :: Dist (Server)</name>
 
+  <properties>
+    <disruptor.version>3.4.0</disruptor.version>
```
eolivelli, line 35: this should be here

also, why do you want to add this dependency ?

eolivelli (changes requested):

bookkeeper-dist/server/pom.xml
StevenLuMT: @eolivelli 
log4j-core need disruptor to write log, so import disruptor like pulsar
image

Shoothzj:
I am confused too. why we need this dependency ? We can print log well for now.
Do you want to support Async log on bookkeeper?

cc @eolivelli

StevenLuMT:
> Do you want to support Async log on bookkeeper?

yes, this pr's motivation is to support async log  @Shoothzj

Shoothzj (changes requested):
Why  we need add `log4j-1.2-api`  dependency?

StevenLuMT:
rerun failure checks

StevenLuMT:
> Why we need add `log4j-1.2-api` dependency?

 @Shoothzj 
Legacy Log4j 1.2 API , it's the dependency in org.apache.logging.log4j.log4j-bom

image

Shoothzj:
@StevenLuMT of course it's defined in the bom, but it's not the reason we put it in our dependency! I didn't see any usage of log4j1 in bookkeeper.

StevenLuMT:
> @StevenLuMT of course it's defined in the bom, but it's not the reason we put it in our dependency! I didn't see any usage of log4j1 in bookkeeper.

good catch, log4j-1.2-api is no use in project, 
I misunderstood,thanks very much @Shoothzj

shoothzj:
closed by #4269 thanks for your initiation contribution.

PR #3480, opened 2022-09-09
Fix the deadlock when only using io thread to handle request

zymap:
---

*Motivation*

If the user doesn't configure the ReadWorker thread pool, the request will process with io thread. We cannot call await() from an IO thread, if the socket buffer is full, that blocking call would cause a deadlock.

*Modification*

- only wait for the promise when the thread is not io thread

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PacketProcessorBase.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PacketProcessorBase.java
@@ -18,8 +18,8 @@
 package org.apache.bookkeeper.proto;
 
 import io.netty.channel.Channel;
-import java.util.concurrent.TimeUnit;
 import io.netty.channel.ChannelFuture;
+import java.util.concurrent.TimeUnit;
 import org.apache.bookkeeper.proto.BookieProtocol.Request;
 import org.apache.bookkeeper.stats.OpStatsLogger;
 import org.apache.bookkeeper.util.MathUtils;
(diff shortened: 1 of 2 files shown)


merlimat (approved):

hangc0276:
Nice catch!

Please fix the check style.
```
Error:  /home/runner/work/bookkeeper/bookkeeper/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PacketProcessorBase.java:22:1: Import io.netty.channel.ChannelFuture appears after other imports that it should precede [ImportOrder]
Error:  /home/runner/work/bookkeeper/bookkeeper/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieRecoveryUseIOThreadTest.java:26:1: Extra separation in import group before 'java.io.IOException' [ImportOrder]
Error:  /home/runner/work/bookkeeper/bookkeeper/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieRecoveryUseIOThreadTest.java:26:1: Import java.io.IOException appears after other imports that it should precede [ImportOrder]
Audit done.
[INFO] There are 3 errors reported by Checkstyle 9.3 with buildtools/src/main/resources/bookkeeper/checkstyle.xml ruleset.
Error:  src/main/java/org/apache/bookkeeper/proto/PacketProcessorBase.java:[22,1] (imports) ImportOrder: Import io.netty.channel.ChannelFuture appears after other imports that it should precede
Error:  src/test/java/org/apache/bookkeeper/client/BookieRecoveryUseIOThreadTest.java:[26,1] (imports) ImportOrder: Extra separation in import group before 'java.io.IOException'
Error:  src/test/java/org/apache/bookkeeper/client/BookieRecoveryUseIOThreadTest.java:[26,1] (imports) ImportOrder: Import java.io.IOException appears after other imports that it should precede
```

hangc0276 (approved):

eolivelli (approved):
LGTM

Issue #2945, opened 2021-12-16
Bookkeeper "READ_ENTRY (Inprogress)" keeps non zero


dragonls:
**BUG REPORT**

***Describe the bug***

I'm using Pulsar 2.7.2, the Bookkeeper version is 4.12.0.

The grafana panel shows that "READ_ENTRY (Inprogress)", the metrics of "bookkeeper_server_READ_ENTRY_IN_PROGRESS" , is not 0.
![image](https://user-images.githubusercontent.com/2565118/146320423-cc9aac11-8b3d-4c11-9ed7-bfd510f24a8f.png)

While the indicator is rising abnormally, the read latency also rises abnormally.
![image](https://user-images.githubusercontent.com/2565118/146320636-e5373cd9-4986-4bd2-a38e-94684994db22.png)

***To Reproduce***
No certain way to reproduce.

***Expected behavior***

"READ_ENTRY (Inprogress)" is always 0.

***Additional context***

This case is happened before, and I had to restart the bookkeeper to keep it 0.

dragonls:
The disk IO seems to be OK, nothing special.
![image](https://user-images.githubusercontent.com/2565118/146321844-9ce21b6b-3181-4269-a473-1bb2ad0f76c3.png)

dragonls:
The log shows that lots of `No ledger found while trying to read last entry: xxx` first:
```
22:25:19.742 [db-storage-14-1] INFO  org.apache.bookkeeper.bookie.Journal - garbage collected journal 17ce633929b.txn
22:25:20.292 [bookie-io-1-9] INFO  org.apache.bookkeeper.proto.AuthHandler - Authentication success on server side
22:25:20.292 [bookie-io-1-9] INFO  org.apache.bookkeeper.proto.BookieRequestHandler - Channel connected  [id: 0x793e94c8, L:/9.146.225.131:3181 - R:/9.146.225.179:40152]
22:25:20.440 [BookieReadThreadPool-OrderedExecutor-0-0] ERROR org.apache.bookkeeper.proto.ReadLacProcessorV3 - No ledger found while performing readLac from ledger: 930007
org.apache.bookkeeper.bookie.Bookie$NoLedgerException: Ledger 930007 not found
        at org.apache.bookkeeper.bookie.LedgerDescriptor.createReadOnly(LedgerDescriptor.java:52) ~[org.apache.bookkeeper-bookkeeper-server-4.12.0.jar:4.12.0]
[...]

PR #3482, opened 2022-09-09
Add missed call to onReadRequestFinish() when read request rejected

michaeljmarshall:
Fixes: https://github.com/apache/bookkeeper/issues/2945

### Motivation

When a read request is rejected due to the limit on the number of reads that can be enqueued, we need to decrement the metric for `bookkeeper_server_READ_ENTRY_IN_PROGRESS` and release the `readsSemaphore`, if it is not null.

### Changes

* Call `onReadRequestFinish()` for failures in the V2 and V3 request processing logic.

### Observation
It looks like the V2 responses for this kind of failure are not throttled, even when `throttleReadResponses` is true. We could call `read.sendReadReqResponse(` to enable this throttling. I didn't change it here because there might be a reason we're not throttling these small responses.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
@@ -539,6 +539,7 @@ public class BookieRequestProcessor implements RequestProcessor {
                     .setReadResponse(readResponse);
                 BookkeeperProtocol.Response resp = response.build();
                 read.sendResponse(readResponse.getStatus(), resp, requestStats.getReadRequestStats());
+                onReadRequestFinish();
             }
         }
     }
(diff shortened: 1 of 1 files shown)


michaeljmarshall:
@dlg99 - PTAL. I don't see a good place to add a test for this change. Do you have any recommendations? It looks like we're not testing this metric for any scenario right now.

eolivelli (approved):
Lgtm


Good catch!

michaeljmarshall:
rerun failure checks

StevenLuMT (approved):
Great work

Shoothzj (approved):

michaeljmarshall:
@Shoothzj @eolivelli - this bug was observed in 4.12. Are you able to cherry pick this (or label it to be cherry picked) for the active release branches?

Shoothzj:
@michaeljmarshall I have done cherry pick this to `branch-4.15`, and labeled it. :)

michaeljmarshall:
Thanks @Shoothzj!

PR #3483, opened 2022-09-11
Migrate Dockerfile from centos to ubuntu

Shoothzj:
### Motivation
centos is not maintained anymore, switch to ubuntu

### Changes
- migrate from centos to ubuntu
- use python3 instead of python2
- Bump bookkeeper version from 4.12.1 to latest 4.15.1

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -17,10 +17,12 @@
 # under the License.
 #
 
-FROM centos:7
+FROM ubuntu:22.10
 MAINTAINER Apache BookKeeper <dev@bookkeeper.apache.org>
 
-ARG BK_VERSION=4.12.1
+ARG TARGETARCH
+
+ARG BK_VERSION=4.15.1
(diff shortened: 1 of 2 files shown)


Shoothzj:
@merlimat @nicoloboschi @eolivelli @hangc0276 PTAL

merlimat (approved):

eolivelli (approved):
really good !

+1

zymap:
I think we need to make the docker file changes apply to other branches.

lhotari:
Putting compilers in a production docker image isn't usually recommended. There shouldn't be any need to add gcc and g++.

See https://github.com/apache/pulsar/pull/14093 . When installing python3 and pip3, the way to prevent installing compilers is to use `apt-get -y install --no-install-recommends python3 python3-pip` for the installation. 

Blog post which explains the need to use `--no-install-recommends` when creating docker images: https://ubuntu.com/blog/we-reduced-our-docker-images-by-60-with-no-install-recommends

PR #3484, opened 2022-09-11
[unit tests] optimize SlowBookieTest

Shoothzj:
### Changes
- fix typo
- rename variable `state` to `writable`
- close the resource LedgerHandle


eolivelli (approved):
Lgtm

This is only a minor code clean up.
I don't think this is a 'optimization' in the sense that the test will run better.

BTW we reduce the tech detto.
Thanks

PR #3486, opened 2022-09-14
Check if channel closed before processing read request

michaeljmarshall:
### Motivation

When a client disconnects from a bookie server, there is a chance that there are pending read requests in the `BookieRequestProcessor`'s queue. Since these are read operations cannot succeed, there is no reason to attempt them.

The primary motivation is to prevent unnecessary reads to the disk as well as decrease the memory utilization associated with unnecessary reads. Additionally, reads are currently performed as blocking calls on the `BookieRequestProcessor` processing thread pool, so failing fast can increase the processing speed of the thread pool. See:

https://github.com/apache/bookkeeper/blob/a65e888b25eb966921ff39032e04f367a553d634/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Bookie.java#L57-L58

Note: we cannot add this logic for writes because those can be "recovered" when the client reconnects.

### Changes

* Only process a read request if the channel is open.

### Alternative Designs

If we are concerned about the cost of reading the volatile variable from the channel, we could possibly speed up the logic by only running the check if the operation has been enqueued for over some threshold, like 100 millis.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadEntryProcessor.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadEntryProcessor.java
@@ -61,6 +61,12 @@ class ReadEntryProcessor extends PacketProcessorBase<ReadRequest> {
         if (LOG.isDebugEnabled()) {
             LOG.debug("Received new read request: {}", request);
         }
+        if (!channel.isOpen()) {
+            if (LOG.isDebugEnabled()) {
+                LOG.debug("Dropping read request for closed channel: {}", channel);
+            }
+            return;
+        }
         int errorCode = BookieProtocol.EOK;
         long startTimeNanos = MathUtils.nowInNano();
         ByteBuf data = null;
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadEntryProcessor.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadEntryProcessor.java
@@ -65,6 +65,7 @@ class ReadEntryProcessor extends PacketProcessorBase<ReadRequest> {
             if (LOG.isDebugEnabled()) {
                 LOG.debug("Dropping read request for closed channel: {}", channel);
             }
+            requestProcessor.onReadRequestFinish();
             return;
         }
         int errorCode = BookieProtocol.EOK;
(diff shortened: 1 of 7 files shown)


lhotari (approved):

michaeljmarshall:
rerun failure checks

michaeljmarshall:
rerun failure checks

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadEntryProcessorV3.java
```diff
@@ -249,6 +249,12 @@ protected ReadResponse getReadResponse() {
     public void safeRun() {
         requestProcessor.getRequestStats().getReadEntrySchedulingDelayStats().registerSuccessfulEvent(
             MathUtils.elapsedNanos(enqueueNanos), TimeUnit.NANOSECONDS);
+        if (!channel.isOpen()) {
+            if (LOG.isDebugEnabled()) {
+                LOG.debug("Dropping read request for closed channel: {}", channel);
+            }
+            return;
```
dlg99, line 256: should it call `requestProcessor.onReadRequestFinish();` before return?

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadEntryProcessor.java
```diff
@@ -61,6 +61,12 @@ protected void processPacket() {
         if (LOG.isDebugEnabled()) {
             LOG.debug("Received new read request: {}", request);
         }
+        if (!channel.isOpen()) {
+            if (LOG.isDebugEnabled()) {
+                LOG.debug("Dropping read request for closed channel: {}", channel);
+            }
+            return;
```
dlg99, line 68: should it call `requestProcessor.onReadRequestFinish();` before return?

dlg99 (changes requested):
I think `requestProcessor.onReadRequestFinish();`is missing

StevenLuMT:
rerun failure checks

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadEntryProcessorV3.java
michaeljmarshall: Yes, great catch! Thank you. I definitely should have caught that since I just fixed the same bug last week.

michaeljmarshall:
rerun failure checks

Shoothzj (approved):

michaeljmarshall:
rerun failure checks

Issue #3488, opened 2022-09-14
A new simple API to create embeddable server instances


diegosalvi:
**FEATURE REQUEST**

1. Please describe the feature you are requesting.
From 4.15 Bookie and BookieServer creation is changed. Is enough common for project to start an embedded version of BK using such classes. Instead of rewrite every time boilerplate code the full server instantiation can be extracted from Main class creating a public API to create BookieService from configurations and/or provided component instances.

2. Indicate the importance of this issue to you (blocker, must-have, should-have, nice-to-have).
Nice-to-have

3. Are you currently using any workarounds to address this issue?
To migrate to 4.15 I had to rewrite classes to create an embedded BK instance looking into "internal" classes

4. Provide any additional detail on your proposed use case for this feature.
Create an embedded BK instance could be as easy ad invoke a builder. This can be really useful on both embedding BK in production project and in tests environments

PR #3489, opened 2022-09-14
Issue #3488 Create an EmbeddedSever extracting code from Main class and adding a builder pattern

diegosalvi:

### Motivation
Is enough common for project to start an embedded version of BK using such classes. Instead of rewrite every time boilerplate code the full server instantiation can be extracted from Main class creating a public API to create BookieService from configurations and/or provided component instances.

### Changes
I've took code from Main class and refactored into a builder as EmbeddedServer class.
Just providing the configuration to the builder produces the same server that Main class was generating.
Added multiple builder properties to permit to inject custom implementations for various components. "Injected" component lifecycle is responsability of the implementor (ie.: if you provide specific instances you need to remember to start and/or stop such instances accordingly)

Master Issue: #3488

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/server/EmbeddedServer.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/server/EmbeddedServer.java
@@ -196,6 +196,8 @@ public class EmbeddedServer {
 
         private BookieConfiguration conf;
 
+        boolean addExternalResourcesToLifecycle = true;
+
         private StatsProvider statsProvider;
 
         private MetadataBookieDriver metadataDriver;
@@ -215,6 +217,11 @@ public class EmbeddedServer {
             checkNotNull(conf, "bookieConfiguration cannot be null");
 
             this.conf = conf;
+        }
+
+        public Builder addExternalResourcesToLifecycle(boolean addExternalResourcesToLifecycle) {
+            this.addExternalResourcesToLifecycle = addExternalResourcesToLifecycle;
+            return this;
         }
 
         public Builder statsProvider(StatsProvider statsProvider) {
@@ -306,17 +313,23 @@ public class EmbeddedServer {
                     metadataDriver = BookieResources.createMetadataDriver(conf.getServerConf(), rootStatsLogger);
                     serverBuilder.addComponent(new AutoCloseableLifecycleComponent("metadataDriver", metadataDriver));
                 }
+            } else if (addExternalResourcesToLifecycle) {
+                serverBuilder.addComponent(new AutoCloseableLifecycleComponent("metadataDriver", metadataDriver));
             }
 
             if (registrationManager == null) {
                 registrationManager = metadataDriver.createRegistrationManager();
                 serverBuilder.addComponent(
                         new AutoCloseableLifecycleComponent("registrationManager", registrationManager));
+            } else if (addExternalResourcesToLifecycle) {
+                serverBuilder.addComponent(new AutoCloseableLifecycleComponent("registrationManager", registrationManager));
             }
 
             // 3. Build ledger manager
             if (ledgerManagerFactory == null) {
                 ledgerManagerFactory = metadataDriver.getLedgerManagerFactory();
+                serverBuilder.addComponent(new AutoCloseableLifecycleComponent("lmFactory", ledgerManagerFactory));
+            } else if (addExternalResourcesToLifecycle) {
                 serverBuilder.addComponent(new AutoCloseableLifecycleComponent("lmFactory", ledgerManagerFactory));
             }
             LedgerManager ledgerManager = ledgerManagerFactory.newLedgerManager();
@@ -352,10 +365,9 @@ public class EmbeddedServer {
 
             if (uncleanShutdownDetection == null) {
                 uncleanShutdownDetection = new UncleanShutdownDetectionImpl(ledgerDirsManager);
-                if (uncleanShutdownDetection.lastShutdownWasUnclean()) {
-                    log.info("Unclean shutdown detected."
-                            + " The bookie did not register a graceful shutdown prior to this boot.");
-                }
+            }
+            if (uncleanShutdownDetection.lastShutdownWasUnclean()) {
+                log.info("Unclean shutdown detected. The bookie did not register a graceful shutdown prior to this boot.");
             }
 
             // bookie takes ownership of storage, so shuts it down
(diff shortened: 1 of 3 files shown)


bookkeeper-server/src/test/java/org/apache/bookkeeper/server/TestEmbeddedServer.java
```diff
@@ -1,60 +1,67 @@
 /*
- * Licensed to the Apache Software Foundation (ASF) under one
- * or more contributor license agreements.  See the NOTICE file
- * distributed with this work for additional information
- * regarding copyright ownership.  The ASF licenses this file
- * to you under the Apache License, Version 2.0 (the
- * "License"); you may not use this file except in compliance
- * with the License.  You may obtain a copy of the License at
+ * Copyright 2022 The Apache Software Foundation.
```
eolivelli, line 2: The license seems wrong

bookkeeper-server/src/main/java/org/apache/bookkeeper/server/EmbeddedServer.java
```diff
@@ -0,0 +1,630 @@
+/*
+ * Copyright 2022 The Apache Software Foundation.
```
eolivelli, line 2: This license seems wrong, please copy/paste from other files

eolivelli (changes requested):
Great work

I generally support this patch.

Please start a discussion on the dev@ mailing list.
I think that we need a BP for this, as it is a new public API.

bookkeeper-server/src/test/java/org/apache/bookkeeper/server/TestEmbeddedServer.java
diegosalvi: Changed with a copy took from other files

bookkeeper-server/src/main/java/org/apache/bookkeeper/server/EmbeddedServer.java
diegosalvi: Changed with a copy took from other files

bookkeeper-server/src/main/java/org/apache/bookkeeper/server/EmbeddedServer.java
```diff
@@ -0,0 +1,636 @@
...
+import org.apache.commons.lang3.StringUtils;
+
+/**
+ * An embedded server is a server that run bookie and serving rpc requests.
+ *
+ * <p>
+ * It is a rewritten server using {@link org.apache.bookkeeper.common.component.LifecycleComponent}, replacing the
+ * legacy server {@link org.apache.bookkeeper.proto.BookieServer}.
+ */
+public class EmbeddedServer {
```
RaulGracia, line 98: If I understand correctly, this is embedding the Bookie server and resources, right? So it can be used both in tests or when running a real Bookie. I think that I'm in favor of this approach, mainly because it facilitates testing to other projects using Bookkeeper. For instance, in this PR I had to do some gymnastics to instantiate test Bookies when upgrading to Bookkeeper 4.15 in Pravega: https://github.com/pravega/pravega/pull/6676

My understanding is that, with this change, just instantiating a new `EmbeddedServer` (maybe the name of the class could be discussed) with the right configuration, the external code should not care about the resources of the Bookie itself, right? If this is the case, I think that it is an improvement in usability.

diegosalvi:
Righ, it is exactly the case.

Il Mer 14 Set 2022, 16:20 Raúl Gracia ***@***.***> ha scritto:

bookkeeper-server/src/main/java/org/apache/bookkeeper/server/EmbeddedServer.java
diegosalvi: Right this is exactly the case. I needed it both for tests and embedding a bookie server

StevenLuMT (approved):
great work,
looking forward to the launch of this feature

StevenLuMT:
rerun failure checks

eolivelli (approved):
Lgtm as far as the vote passes

StevenLuMT:
rerun failure checks

Shoothzj (approved):

dlg99 (approved):
LGTM.
There are couple of failing tests on the CI, on the first glance seem to be related to the change. Please take a look.

RaulGracia (approved):

StevenLuMT:
> LGTM. There are couple of failing tests on the CI, on the first glance seem to be related to the change. Please take a look.

@diegosalvi have a look at failing tests on the CI

hangc0276 (approved):
Great job!

hangc0276:
rerun failure checks

bookkeeper-server/src/main/java/org/apache/bookkeeper/server/EmbeddedServer.java
```diff
@@ -0,0 +1,648 @@
...
+    public static final Builder builder(BookieConfiguration conf) {
+        return new Builder(conf);
+    }
+
+    @Slf4j
+    public static class Builder {
+
+        private BookieConfiguration conf;
+
+        boolean addExternalResourcesToLifecycle = true;
```
eolivelli, line 199: What does this mean?
If a component is passed from outside we shouldn't handle it.
Why are you adding this feature?
Also, if you really need this it must be turned off by default

eolivelli (changes requested):

bookkeeper-server/src/main/java/org/apache/bookkeeper/server/EmbeddedServer.java
diegosalvi: Sorry is a WIP fix but I had to see running it on CI

diegosalvi: Set to default false in the last commit

diegosalvi:
Finally checks all green

bookkeeper-server/src/main/java/org/apache/bookkeeper/server/EmbeddedServer.java
eolivelli: if this is not required to make CI pass I prefer to drop the flag at all.
Otherwise we add a feature that won't be used by anyone but we will have to support it forever.
Currently in the case of Pravega, Pulsar and HerdDB that I know, there is no need to let the Bookie take control of externally provided components

diegosalvi: Removed addExternalResourcesToLifecycle

Issue #3250, opened 2022-04-28
Should we shutdown the bookie or retry when registration failure


Shoothzj:
about the flaky test https://github.com/apache/bookkeeper/issues/3206

I have figured out why testBookieServerZKExpireBehaviour falky.

It's due to Registration task, bookie will retry to registration when zk session expires. If the registration fail, it will cause the bookie shutdown ExitCode.ZK_REG_FAIL.

@dlg99 @eolivelli @merlimat PTAL

dlg99:
de-facto doc:
https://github.com/apache/bookkeeper/blob/6b03b012f731536af56cc63ba29511f09994d11b/conf/bk_server.conf#L804-L809

BK will exit on SessionExpiredException from zk client; it is hard to recover from it (have to re-create the zk client + redo all initialization steps of all subcomponents) so better to exit and let k8s/supervisord restart the service.

PR #3495, opened 2022-09-20
Add RegistrationSessionExpiredPolicy for session expired event

wenbingshen:
### Motivation
The status of the Bookie service, including `writable` and `read-only`, needs to be registered with the `Registration` service.
At present, most of the time we use `zookeeper` as the `Registration` service, where `BookieStateManager` is mainly responsible for managing these state switches, including the  `re-registration` logic after processing the `session expired event`.
`BookieStateManager` uses `ZKRegistrationManager` to register the state, where `zk` in `ZKRegistrationManager` is a subclass of `Zookeeper` to implement `ZooKeeperClient`. When `session expired` occurs,`ZooKeeperClient` will re-instantiate a new `Zookeeper` for `zk`.
If the new `zk` is successfully constructed, the `BookieStateManager` will re-register the bookie state with zookeeper server.

Here we have a problem, from the existing unit tests and the actual show, we seem to want the bookie to be closed when the session expired occurs, because we can't prove whether the bookie can be in the correct working state just by reinitializing a new zk, refer to comments on the following issue: https://github.com/apache/bookkeeper/issues/3250#issuecomment-1190996890

Like pulsar, we can provide session expired poilicy to choose whether it should continue to reconnect.

### Changes
Add `RegistrationSessionExpiredPolicy` and server-side configuration items: `registrationSessionExpiredPolicy`
Contains two optional values:  `reconnect, shutdown`

To stay consistent with the existing logic, this PR will set the policy to `reconnect`.

If we want to solve the problem mentioned in #3418 , we can modify the policy to shutdown here at #3418.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStateManager.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStateManager.java
@@ -119,6 +119,12 @@ public class BookieStateManager implements StateManager {
         this.rm = rm;
         if (this.rm != null) {
             rm.addRegistrationListener(() -> {
+                LOG.info("Received registration expired event");
+                if (conf.getRegistrationSessionExpiredPolicy() == RegistrationSessionExpiredPolicy.shutdown) {
+                    LOG.warn("The session with registration service was lost. Shutting down.");
+                    shutdownHandler.shutdown(ExitCode.BOOKIE_EXCEPTION);
+                    return;
+                }
                 log.info("Trying to re-register the bookie");
                 forceToUnregistered();
                 // schedule a re-register operation
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/RegistrationSessionExpiredPolicy.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/RegistrationSessionExpiredPolicy.java
@@ -1,0 +1,30 @@
+/**
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+
+package org.apache.bookkeeper.bookie;
+
+/**
+ * Enum with different policies do apply when bookies are loosing session with registration service.
+ */
+public enum RegistrationSessionExpiredPolicy {
+    reconnect,
+    shutdown,
+}
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
@@ -28,6 +28,7 @@ import java.util.concurrent.TimeUnit;
 import org.apache.bookkeeper.bookie.FileChannelProvider;
 import org.apache.bookkeeper.bookie.InterleavedLedgerStorage;
 import org.apache.bookkeeper.bookie.LedgerStorage;
+import org.apache.bookkeeper.bookie.RegistrationSessionExpiredPolicy;
 import org.apache.bookkeeper.bookie.SortedLedgerStorage;
 import org.apache.bookkeeper.bookie.storage.ldb.DbLedgerStorage;
 import org.apache.bookkeeper.common.conf.ConfigDef;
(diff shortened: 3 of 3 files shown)


wenbingshen:
ping @hangc0276 @dlg99 @Shoothzj @eolivelli PTAL. Thanks.

wenbingshen:
rerun failure checks

StevenLuMT (changes requested):
@wenbingshen good idea, but I have 2 questions:
1. what scenarios are these two enumeration values of registrationSessionExpiredPolicy suitable for?
2. please set two testcases for different enumerations to cover different logic

wenbingshen:
@StevenLuMT Thanks for your review.

> 1. what scenarios are these two enumeration values of registrationSessionExpiredPolicy suitable for?

This PR tends to introduce two policies `reconnect, shutdown` and keeps the same logic as before `reconnect`.
It is found from the current unit testing and actual use, we should tend to use the `shutdown` policy.
But I'm leaning towards this PR to keep the default `reconnect` policy and instead change to `shutdown` policy in #3418 to fix falky-test.

Because Bookie is difficult to recover from Session Expired, as described in the comment: https://github.com/apache/bookkeeper/issues/3250#issuecomment-1190996890

So the reconnect policy is still a policy to be implemented and verified. `reconnect` was introduced to keep the same logic as before. We can use another PR to implement the complete policy of  `Bookie Reconnect Recovery` from Session Expired.

Here I divide it into three steps:
1. This PR introduces two policy options
2. Change the default policy to `shutdown` to solve flaky-test #3418
3. Implement complete bookie recovery logic for the `reconnect` policy


> 2. please set two testcases for different enumerations to cover different logic

I can introduce a test for `shutdown` Policy, and `reconnect` can reuse the previous `testBookieServerZKSessionExpireBehaviour` test, because `reconnect` is still an unstable recovery policy that can only be verified until the third point mentioned above is implemented: 3. Implement complete bookie recovery logic for the `reconnect` policy.

Shoothzj:
Basically agree, we also have another way of removing the `ReconnectPolicy`, because it's difficult to implement and no one would want to implement it. We can discuss it in maillist. :)

wenbingshen:
> Basically agree, we also have another way of removing the `ReconnectPolicy`, because it's difficult to implement and no one would want to implement it. We can discuss it in maillist. :)

wenbingshen:
> Basically agree, we also have another way of removing the `ReconnectPolicy`, because it's difficult to implement and no one would want to implement it. We can discuss it in maillist. :)

@Shoothzj Thanks for your review. I'll send a discussion email later.

dlg99 (approved):
LGTM but please add it the new config parameter to conf/bk_server.conf and document it there

dlg99:
Also needs tests.

dlg99:
and you are correct, after session expired one has to re-create ZK client ("once a ZooKeeper object is closed or receives a fatal event (SESSION_EXPIRED and AUTH_FAILED), the ZooKeeper object becomes invalid." see https://zookeeper.apache.org/doc/r3.8.0/zookeeperProgrammers.html ) which is doable but needs to be bubbled up the chain of injected dependencies.

wenbingshen:
@dlg99 Thanks for your review. :)
> LGTM but please add it the new config parameter to conf/bk_server.conf and document it there
> Also needs tests.
I will add it to the new config and doc later.
Also I will add test. Thanks. :)

bookkeeper-server/src/test/java/org/apache/bookkeeper/test/BookieZKExpireTest.java
```diff
@@ -67,7 +68,7 @@ public void testBookieServerZKRequestTimeoutBehaviour() throws Exception {
             Thread[] threads = new Thread[threadCount * 2];
             threadCount = Thread.enumerate(threads);
             for (int i = 0; i < threadCount; i++) {
-                if (threads[i].getName().indexOf("SendThread") != -1) {
+                if (threads[i].getName().contains("SendThread")) {
```
eolivelli, line 71: nit: this change doesn't seem relevant

eolivelli (approved):
LGTM

@dlg99 PTAL

bookkeeper-server/src/test/java/org/apache/bookkeeper/test/BookieZKExpireTest.java
wenbingshen: @eolivelli Thanks for your review. I have reverted this minor change.

hangc0276 (changes requested):
IMO, we should keep the session expired policy as `reconnect` because Bookie is a heavy service with the state, and shutdown it when the session expired, will cause the BookKeeper cluster unstable. 

In Pulsar, we have changed the default session expired policy from `shutdown` to `reconnect` to make the Pulsar service more stable.

Currently, Bookie has issues with session expired, and we should enhance the reconnect policy instead of introducing a shutdown policy.

wenbingshen:
> IMO, we should keep the session expired policy as `reconnect` because Bookie is a heavy service with the state, and shutdown it when the session expired, will cause the BookKeeper cluster unstable.

Yes. I think we can introduce these two polices and keep `reconnect` as the default policy. When users encounter the instability of the bookie service due to session expiration, they can adjust to the `shutdown` policy, such as @dlg99's comment at https://github.com/ apache/bookkeeper/issues/3250#issuecomment-1190996890.

> In Pulsar, we have changed the default session expired policy from `shutdown` to `reconnect` to make the Pulsar service more stable.

Yes, I believe this makes sense as long as the reconnect policy matures.

> Currently, Bookie has issues with session expired, and we should enhance the reconnect policy instead of introducing a shutdown policy.

We can keep the reconnect policy as the default, and provide a shutdown policy option until the `reconnect` is perfected and stabilized.

@hangc0276 PTAL.

PR #3496, opened 2022-09-20
Simplified No network topology script is found default log stack output

wenbingshen:
### Motivation
The default `dnsResolver` of `RackawareEnsemblePlacementPolicyImpl` is initialized as `ScriptBasedMapping`. Since the script is missing by default, detailed stack information will be printed, but the code will fall back to `DefaultResolver` by default, so only simple exception information needs to be printed.
image

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
@@ -270,8 +270,8 @@ public class RackawareEnsemblePlacementPolicyImpl extends TopologyAwareEnsembleP
                 }
             } catch (RuntimeException re) {
                 if (!conf.getEnforceMinNumRacksPerWriteQuorum()) {
-                    LOG.warn("Failed to initialize DNS Resolver {}, used default subnet resolver ",
-                            dnsResolverName, re);
+                    LOG.warn("Failed to initialize DNS Resolver {}, used default subnet resolver because {}",
+                            dnsResolverName, re.getMessage());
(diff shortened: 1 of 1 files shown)


wenbingshen:
rerun failure checks

StevenLuMT (approved):
LGTM

Shoothzj (approved):

dlg99 (approved):

zymap (commented):
Actually, I am not sure it is right to remove the detailed stack. The stack will give more information when you have issues with the other DNSToSwitchMapping implementations. Removing it will make the log clear but it will be harder to know what happened exactly.

wenbingshen:
> Actually, I am not sure it is right to remove the detailed stack. The stack will give more information when you have issues with the other DNSToSwitchMapping implementations. Removing it will make the log clear but it will be harder to know what happened exactly.

@zymap Thanks for your review. :)

I looked at some current implementations of DNSToSwitchMapping, only throws RuntimeException in the validateConf method of RawScriptBasedMapping and ScriptBasedMapping, by default enforceMinNumRacksPerWriteQuorum is false, even if RuntimeException is thrown, bookie starts and works normally, even if we print detailed stack info, The user may also not perceive the problem, and in most cases, the user does not care about the stack information, they can tolerate falling back to the DefaultResolver, the stack information here is not complicated, I think just printing the exception information is enough.

For user-defined DNSToSwitchMapping implementations, such exception throwing rules must also be followed.

Users can turn on enforceMinNumRacksPerWriteQuorum to view detailed stack information.

As you mentioned, if there is a need to display a prompt to the user, perhaps the log level should be adjusted to ERROR here, and the WARN log is usually easily ignored.

hangc0276 (approved):

wenbingshen:
@zymap @StevenLuMT Can this PR be merged? Any more questions, please let me know. Thanks.

PR #3501, opened 2022-09-23
unified latency metric unit

StevenLuMT:

### Motivation

The latency of the OpStatsLogger.registerSuccessfulEvent calculation is to convert the time to milliseconds
image
but Counter.add nothing to do

so when using Counter for latency statistics, the time unit and OpStatsLogger are not unified, which is easy to be misleading.
then we unified latency metric unit 

### Changes

1.  change name : Counter.add -->  Counter.addCount
2. add new method Counter.addLatency to count the time and convert the time to milliseconds

then how to use counter correctly:
1. when using Counter for latency metric, call Counter.addLatency
2. when using Counter for count metric, call Counter.addCount

### DIFF
--- a/bookkeeper-common/src/test/java/org/apache/bookkeeper/test/TestStatsProvider.java
+++ b/bookkeeper-common/src/test/java/org/apache/bookkeeper/test/TestStatsProvider.java
@@ -60,8 +60,14 @@ public class TestStatsProvider implements StatsProvider {
         }
 
         @Override
-        public void add(long delta) {
+        public void addCount(long delta) {
             updateMax(val.addAndGet(delta));
+        }
+
+        @Override
+        public void addLatency(long eventLatency, TimeUnit unit) {
+            long valueMillis = unit.toMicros(eventLatency) / 1000;
+            updateMax(val.addAndGet(valueMillis));
         }
 
         @Override
--- a/bookkeeper-common/src/test/java/org/apache/bookkeeper/test/TestStatsProvider.java
+++ b/bookkeeper-common/src/test/java/org/apache/bookkeeper/test/TestStatsProvider.java
@@ -66,7 +66,7 @@ public class TestStatsProvider implements StatsProvider {
 
         @Override
         public void addLatency(long eventLatency, TimeUnit unit) {
-            long valueMillis = unit.toMicros(eventLatency) / 1000;
+            long valueMillis = unit.toMillis(eventLatency);
             updateMax(val.addAndGet(valueMillis));
         }
 
--- a/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/stats/BroadCastStatsLogger.java
+++ b/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/stats/BroadCastStatsLogger.java
@@ -127,9 +127,16 @@ public class BroadCastStatsLogger {
                 }
 
                 @Override
-                public void add(long l) {
-                    firstCounter.add(l);
-                    secondCounter.add(l);
+                public void addCount(long l) {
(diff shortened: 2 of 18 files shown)


eolivelli (approved):
Lgtm

But we can't cherry pick this into released branches

Shoothzj:
will this break prometheus metric result?

StevenLuMT:
> will this break prometheus metric result?

this change don't break prometheus metric result:
This just optimize the statistics,but in the end this logic still use Counter.add to update the data
@Shoothzj 
image

Shoothzj (approved):

wenbingshen (approved):
nice found.

dlg99 (approved):

bookkeeper-common/src/test/java/org/apache/bookkeeper/test/TestStatsProvider.java
```diff
@@ -60,10 +60,16 @@ public void dec() {
...
 
         @Override
-        public void add(long delta) {
+        public void addCount(long delta) {
             updateMax(val.addAndGet(delta));
         }
 
+        @Override
+        public void addLatency(long eventLatency, TimeUnit unit) {
+            long valueMillis = unit.toMicros(eventLatency) / 1000;
```
zymap, line 69: Just have a question. Why not use unit.toMills?

StevenLuMT: > Just have a question. Why not use unit.toMills?

it's the same, just use the logic with OpStatsLogger.registerSuccessfulEvent @zymap 
image

Shoothzj: I think we can both use unit.toMills

zymap (approved):

bookkeeper-common/src/test/java/org/apache/bookkeeper/test/TestStatsProvider.java
StevenLuMT: > I think we can both use unit.toMills

goog idea, I will update it

StevenLuMT:
@Shoothzj  have a look the new update for unit.toMillis

bookkeeper-common/src/test/java/org/apache/bookkeeper/test/TestStatsProvider.java
StevenLuMT: @Shoothzj have a look the new update for unit.toMillis

StevenLuMT:
@Shoothzj have a look the new update for unit.toMillis

eolivelli:
@Shoothzj this is a PR related to a BP and we did not run the VOTE.
I will send a message on dev@

We should follow the community process

Shoothzj:
@eolivelli Sorry, My fault, I didn't realize it. Should we revert the PR now or wait for the vote result?

StevenLuMT:
> @eolivelli Sorry, My fault, I didn't realize it. Should we revert the PR now or wait for the vote result?

I have started the VOTE, see for details https://lists.apache.org/thread/4kxbdwjmls56lfmfhrf853dc525dmsy9
@Shoothzj @eolivelli @dlg99 @zymap

StevenLuMT:
> @eolivelli Sorry, My fault, I didn't realize it. Should we revert the PR now or wait for the vote result?

@Shoothzj don't revert ,see the detail for mail discussion:  
image

https://lists.apache.org/thread/zyj3z3tt4bsx928mc0ogfkv63w9t47bc

PR #3505, opened 2022-09-26
Rename success with writableResult and update final writableResult about wait writeSet

wenbingshen:
### Changes
update `success` about wait writeSet for Writable result.


wenbingshen:
ping @dlg99 @hangc0276 @Shoothzj @eolivelli @StevenLuMT PTAL. Thanks.

Shoothzj (approved):

zymap (changes requested):
Great! Maybe we need to add a unit test?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -1262,7 +1262,7 @@ protected boolean waitForWritable(DistributionSchedule.WriteSet writeSet,
             final int maxBackoff = 4;
             final long deadline = startTime + TimeUnit.MILLISECONDS.toNanos(durationMs);
 
-            while (!isWriteSetWritable(writeSet, allowedNonWritableCount)) {
+            while (!(success = isWriteSetWritable(writeSet, allowedNonWritableCount))) {
```
StevenLuMT, line 1265: I think success can be named writableResult,like the code:
`while (!(writableResult = isWriteSetWritable(writeSet, allowedNonWritableCount))) {`

wenbingshen:
> Great! Maybe we need to add a unit test?

@zymap I have add a unit test. PTAL. Thanks.

eolivelli (approved):
Are you only renaming a variable?

I think that adding a test is the maim content of this patch. Maybe we should change the title

wenbingshen:
> Are you only renaming a variable?

No, i updated the title.

> I think that adding a test is the maim content of this patch. Maybe we should change the title

The code before this PR is like this:
```java
while (!isWriteSetWritable(writeSet, allowedNonWritableCount)) {
           ...     
}
```

The code after this PR is this:
```java
while (!(writableResult = isWriteSetWritable(writeSet, allowedNonWritableCount))) {
           ...     
}
```
writableResult will not update the last result, but this result needs to be used later, we should update it.

StevenLuMT (approved):
LGTM

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/SlowBookieTest.java
```diff
@@ -390,4 +390,32 @@ public void operationComplete(int rc, Set<LedgerFragment> fragments) {
...
+
+            RoundRobinDistributionSchedule schedule = new RoundRobinDistributionSchedule(1, 1, 1);
+            DistributionSchedule.WriteSet writeSet = schedule.getWriteSet(entryId);
+
+            int slowBookieIndex = writeSet.get(ThreadLocalRandom.current().nextInt(writeSet.size()));
+            List<BookieId> curEns = lh.getCurrentEnsemble();
+
+            // disable channel writable
+            setTargetChannelState(bkc, curEns.get(slowBookieIndex), 0, false);
+            boolean isWriteable = lh.waitForWritable(writeSet, 0, 1000);
```
zymap, line 412: This is always true/false if you set the state before you call this method. We need a test to make sure after the backoff retries,  it returns the correct state after the `waitForWritable`. 
So the test case will be like this:
1. change the writable to false.
2. call the `waitForWriable`, it will retry with backoff.
3. at the same time, change the writable to true.
4. we should get the true from the step2 call.

wenbingshen: ok, i'll fix it later.

wenbingshen: @zymap I have addressed your comments. PTAL. Thanks.

hangc0276 (approved):

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/SlowBookieTest.java
```diff
@@ -390,4 +390,43 @@ public void operationComplete(int rc, Set<LedgerFragment> fragments) {
...
+            RoundRobinDistributionSchedule schedule = new RoundRobinDistributionSchedule(1, 1, 1);
+            DistributionSchedule.WriteSet writeSet = schedule.getWriteSet(entryId);
+
+            int slowBookieIndex = writeSet.get(ThreadLocalRandom.current().nextInt(writeSet.size()));
+            List<BookieId> curEns = lh.getCurrentEnsemble();
+
+            // disable channel writable
+            setTargetChannelState(bkc, curEns.get(slowBookieIndex), 0, false);
+
+            final CountDownLatch cdl = new CountDownLatch(1);
```
zymap, line 413:
```suggestion
           
```

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/SlowBookieTest.java
```diff
@@ -390,4 +390,43 @@ public void operationComplete(int rc, Set<LedgerFragment> fragments) {
...
+                isWriteable.set(lh.waitForWritable(writeSet, 0, timeout));
+                cdl.countDown();
+            }).start();
+
+            // enable channel writable
+            setTargetChannelState(bkc, curEns.get(slowBookieIndex), 0, true);
+
+            cdl.await(timeout, TimeUnit.MILLISECONDS);
+
+            assertTrue("We should check b0 is writeable", isWriteable.get());
```
zymap, line 428:
```suggestion
           new Thread(() -> {
                isWriteable.set(lh.waitForWritable(writeSet, 0, timeout));
            }).start();
            TimeUnit.MILLISECONDS.sleep(5000);
            assertFalse(isWriteable.get());

            setTargetChannelState(bkc, curEns.get(slowBookieIndex), 0, true);
            Awaitility.await().untilAsserted(() -> assertTrue(isWriteable.get()));
```

zymap (commented):
Sorry for another minor comment. I just want to make sure the test can not be passed without this change. 
I copied this test and run it locally, and it passed. I left a comment about how to fix it.  PTAL

wenbingshen:
> Sorry for another minor comment. I just want to make sure the test can not be passed without this change.
I copied this test and run it locally, and it passed. I left a comment about how to fix it. PTAL. :)

@zymap Good. Thanks for your comments. I have committed it. PTAL.

zymap (approved):

wenbingshen:
@zymap All tests passed. Can this pr be merged? Thanks.

PR #3509, opened 2022-09-30
Add new api resumeCompaction and suspendCompaction

gaozhangmin:
### Motivation
Compaction would bring high disk io util, It would better allowing to control suspending and resuming compaction through api

### Changes
Add two apis to control resuming and suspending compaction.
`/api/v1/bookie/gc/suspend_compaction`
`/api/v1/bookie/gc/resume_compaction`

### DIFF
--- a/bookkeeper-http/http-server/src/main/java/org/apache/bookkeeper/http/HttpRouter.java
+++ b/bookkeeper-http/http-server/src/main/java/org/apache/bookkeeper/http/HttpRouter.java
@@ -47,6 +47,8 @@ public abstract class HttpRouter<Handler> {
     public static final String EXPAND_STORAGE               = "/api/v1/bookie/expand_storage";
     public static final String GC                           = "/api/v1/bookie/gc";
     public static final String GC_DETAILS                   = "/api/v1/bookie/gc_details";
+    public static final String SUSPEND_GC_COMPACTION        = "/api/v1/bookie/gc/suspend_compaction";
+    public static final String RESUME_GC_COMPACTION         = "/api/v1/bookie/gc/resume_compaction";
     public static final String BOOKIE_STATE                 = "/api/v1/bookie/state";
     public static final String BOOKIE_STATE_READONLY        = "/api/v1/bookie/state/readonly";
     public static final String BOOKIE_IS_READY              = "/api/v1/bookie/is_ready";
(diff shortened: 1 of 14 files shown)


StevenLuMT:
ci fail,have a look the error info
image

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/InterleavedLedgerStorage.java
```diff
@@ -268,6 +268,30 @@ public boolean isInForceGC() {
...
+
+    public void resumeMinorGC() {
+        gcThread.resumeMinorGC();
+    }
+
+    public void resumeMajorGC() {
+        gcThread.resumeMajorGC();
+    }
+
+    public boolean isMajorGcSuspend() {
```
dlg99, line 287: isMajorGcSuspended() / isMinorGcSuspended() etc.

bookkeeper-http/http-server/src/main/java/org/apache/bookkeeper/http/HttpRouter.java
```diff
@@ -51,6 +51,8 @@
     public static final String BOOKIE_STATE_READONLY        = "/api/v1/bookie/state/readonly";
     public static final String BOOKIE_IS_READY              = "/api/v1/bookie/is_ready";
     public static final String BOOKIE_INFO                  = "/api/v1/bookie/info";
+    public static final String SUSPEND_COMPACTION           = "/api/v1/bookie/compaction/suspend";
```
dlg99, line 54: we are using "/gc" (just above) and "/compaction" to control different features of the same functionality. We need to make it less confusing and unify around the same naming.
Maybe deprecate "/api/v1/bookie/gc" / "/api/v1/bookie/gc_details" and do "/api/v1/bookie/compaction/force_major|force_minor" / "/api/v1/bookie/compaction/details" instead?

dlg99 (changes requested):
Great job.
We need to standardize naming for the API to not have two names for the same functionality.
Also needs documentation, see https://bookkeeper.apache.org/docs/admin/http

bookkeeper-http/http-server/src/main/java/org/apache/bookkeeper/http/HttpRouter.java
gaozhangmin: Bookkeeper GC contains two parts, 1. GC ledgers and entry log. 2. compaction entry log files. 
I don't think  we should combine the two apis together.
`/gc` is to trigger 1 and 2, while the new api is to control part 2.

eolivelli (approved):
LGTM

gaozhangmin:
@dlg99  PTAL again, thx

gaozhangmin:
api docs added

gaozhangmin:
rerun failure checks

bookkeeper-http/http-server/src/main/java/org/apache/bookkeeper/http/HttpRouter.java
hangc0276: +1 for @dlg99. I prefer to use `gc` instead of `compaction` endpoint. It won't introduce break changes.

bookkeeper-http/http-server/src/main/java/org/apache/bookkeeper/http/HttpRouter.java
```diff
@@ -51,6 +51,8 @@
...
+    public static final String RESUME_COMPACTION            = "/api/v1/bookie/compaction/resume";
```
hangc0276, line 55: I prefer to use `gc` endpoint instead of `compaction`, It will introduce more confusion.

hangc0276 (changes requested):

gaozhangmin:
@hangc0276  PTAL again @dlg99

hangc0276 (approved):
LGTM

dlg99 (approved):
LGTM, thank you!

PR #3515, opened 2022-10-03
disable direct IO test on windows

Shoothzj:
### Motivation
- disable direct IO test on windows, it's not supported on windows


eolivelli (approved):
LGTM

maybe we could investigate how to run CI on Windows, at least a subset of tests (especially for the client side)

Shoothzj:
@eolivelli Great idea. I will try

PR #3475, opened 2022-09-06
Add 4.15.1 to versions file

zymap:
---

*Motivation*

Add 4.15.1 to version file to update the website

PR #3520, opened 2022-10-08
Generate 4.15.2 docs

zymap:
### Motivation

Generate 4.15.2 docs.


StevenLuMT:
@zymap I have a question:
Why not do this:  keep 4.15.1, add 4.15.2  ?

zymap:
@StevenLuMT The patch release only keeps the latest version docs. See here https://github.com/apache/bookkeeper/pull/3475#issuecomment-1237940759

hangc0276:
I think we'd better use 4.15.x instead of a specific version number.  Pulsar (https://pulsar.apache.org/docs/) and Kafka (https://kafka.apache.org/documentation/) documentation also use major version + ".x" to represent the current major latest version. I do not object to this PR, and I will send out an email to discuss it.

hangc0276 (approved):

PR #3522, opened 2022-10-09
Change masterKeyCache to dynamic size

StevenLuMT:

### Motivation

org.apache.bookkeeper.bookie.BookieImpl#masterKeyCache has a problem, when bookie has started, the data is only added but not removed,
then change masterKeyCache to dynamic size,duration to prevent memory from rising because the number of ledger grow too fast

for the masterKeyCache does not need to be very large, even if the masterKeyCache cannot be hit, 
image
  1) case use DbLedgerStorage as storage:
        there is a memory cache of org.apache.bookkeeper.bookie.storage.ldb.LedgerMetadataIndex#ledgers as the bottom line.
image

   2) case use SortedLedgerStorage as storage:
        there is a memory cache of org.apache.bookkeeper.bookie.IndexPersistenceMgr#fileInfoBackingCache as the bottom line.
image

When the following scenarios occur, this problem will become more obvious:
    1)  when pulsar carries tens of millions of topics
    2)  the length of masterKey is 20 
    3)  the number of ledger increases by 50 per second
image
the memory of masterKeyCache will increase by  about 100MB every day

### Changes

1.  change masterKeyCache from fix size map(ConcurrentLongHashMap) to dynamic size(60 miniutes guava cache)


eolivelli:
What happens when the key is removed from the cache? You are citing only the RocksDB ledger storage.
And if you are using other storage engines?

StevenLuMT:
> What happens when the key is removed from the cache? You are citing only the RocksDB ledger storage. And if you are using other storage engines?

@eolivelli good question, I have updated in the Conversation,
have a look, thanks

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
```diff
@@ -935,9 +936,10 @@ private void addEntryInternal(LedgerDescriptor handle, ByteBuf entry,
 
         // journal `addEntry` should happen after the entry is added to ledger storage.
         // otherwise the journal entry can potentially be rolled before the ledger is created in ledger storage.
-        if (masterKeyCache.get(ledgerId) == null) {
+        if (masterKeyCache.getIfPresent(ledgerId) == null) {
             // Force the load into masterKey cache
-            byte[] oldValue = masterKeyCache.putIfAbsent(ledgerId, masterKey);
+            byte[] oldValue = masterKeyCache.getIfPresent(ledgerId);
+            masterKeyCache.put(ledgerId, masterKey);
             if (oldValue == null) {
```
gaozhangmin, line 943: If the cache is expired, but ledger is still not rolled, what will happen with below code will executed multiple times
```
getJournal(ledgerId).logAddEntry(bb, false /* ackBeforeSync */, new NopWriteCallback(), null);
```

StevenLuMT: @gaozhangmin it doesn't matter:
the masterKey of the same ledgerID is the same. This is only to prevent multiple writes. If it occurs multiple times, there is no risk. It will only cover the information when the journal does relay.

wenbingshen: Good point. I have a question. Is there any problem if the `masterKeyCahce` cache is evicted during `readJournal`? I don't see the role of `masterKeyCache` in this process.

gaozhangmin: We should avoid duplicately writing masterKey to journal file,  even duplicate writing has no problem.

StevenLuMT: > Good point. I have a question. Is there any problem if the `masterKeyCahce` cache is evicted during `readJournal`? I don't see the role of `masterKeyCache` in this process.

@wenbingshen 
The main role of masterKeyCache:
1. When addEntry, synchronize a piece of METAENTRY_ID_LEDGER_KEY data, you need masterKeyCache to write only one piece of this data for each ledgerID
2. When the bookie is restarted and the journal is doing data replay, for METAENTRY_ID_LEDGER_KEY, the following operations are done: restore masterKeyCache, restore HandleFactory

wenbingshen (approved):

StevenLuMT:
@eolivelli @zymap @Shoothzj @dlg99 @nicoloboschi help me review it,thanks

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
nicoloboschi: >We should avoid duplicately writing masterKey to journal file, even it has no problem.
I agree with @gaozhangmin .

Also since this touches the write path I suggest you to make this new cache configurable and disabled by default.

nicoloboschi (changes requested):

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
eolivelli: if we disable it by default nobody will use it, it will fall into the basket of "experimental features"

eolivelli (approved):
LGTM

I have re-read again the patch and now I am convinced that this will help.
The trade-off is that if a ledger is not written for more than 60 minutes we will write again the masterKey to the journal once.

In BK usually ledgers are opened-written-closed and a idle time of 60 minutes is really unlikely to happen. And in case it happens in means that the ledger has very low traffic and the overhead of writing the masterKey is not a big deal

Shoothzj (approved):

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
```diff
@@ -112,8 +113,8 @@ public class BookieImpl extends BookieCriticalThread implements Bookie {
 
     private int exitCode = ExitCode.OK;
 
-    private final ConcurrentLongHashMap<byte[]> masterKeyCache =
-            ConcurrentLongHashMap.<byte[]>newBuilder().build();
+    private final Cache<Long, byte[]> masterKeyCache = CacheBuilder.newBuilder().
+            expireAfterAccess(60, TimeUnit.MINUTES).build();
```
merlimat, line 117: This will also mean that we have to acquire a write lock each time to update the last access time. And The cache implementation has to do the extra work for the eviction.

merlimat (commented):
My question is if would rather make more sense to optimize for the fact that the masterkey is  empty in the overwhelmingly major part of the cases.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
StevenLuMT: > 

Is there any better implementation to replace guava cache? @merlimat

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
StevenLuMT: > if we disable it by default nobody will use it, it will fall into the basket of "experimental features"

have a look @nicoloboschi @gaozhangmin 
I have update the code, when check masterKey exists in cache before write
```
// Force the load into masterKey cache
byte[] oldValue = masterKeyCache.asMap().putIfAbsent(ledgerId, masterKey);
```

StevenLuMT:
have a look @nicoloboschi @gaozhangmin 
I have update the code, when check masterKey exists in cache before write
```
// Force the load into masterKey cache
byte[] oldValue = masterKeyCache.asMap().putIfAbsent(ledgerId, masterKey);
```

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
StevenLuMT: > if we disable it by default nobody will use it, it will fall into the basket of "experimental features"

@nicoloboschi 
yes,I agree @eolivelli ,  We should use the new logic directly, and the old logic is a bug, which is easy to cause oom

gaozhangmin (approved):

nicoloboschi (approved):

hangc0276:
> My question is if would rather make more sense to optimize for the fact that the masterkey is empty in the overwhelmingly major part of the cases.

+1 In Pulsar, each ledger's default password is `"".getBytes(StandardCharsets.UTF_8)` and most clusters are not set ledger passwords, and the `masterKey` is generated from ledger password, which means most of the ledgers' masterKey are default value. We can optimize this case.

StevenLuMT:
> > My question is if would rather make more sense to optimize for the fact that the masterkey is empty in the overwhelmingly major part of the cases.
> 
> +1 In Pulsar, each ledger's default password is `"".getBytes(StandardCharsets.UTF_8)` and most clusters are not set ledger passwords, and the `masterKey` is generated from ledger password, which means most of the ledgers' masterKey are default value. We can optimize this case.
> 
> Changing the cache from `ConcurrentLongHashMap ` to Guava cache will introduce extra cost no matter in `putIfAbsent` or eviction. This cache get and put is a high-frequency operation in Bookie, and I'm not sure about the impact on the addEntry performance.
> 
> Maybe we can try to optimize the default `masterKey` case first to see the achivments.


@hangc0276 
Do you understand the purpose of this PR change?
The current replacement of ConcurrentLongHashMap is not because there is any problem with ConcurrentLongHashMap, but because the current masterKeyCache has not been cleaned up


How to optimize? Please give some specific advice? All I can think of is to regularly clean up the masterKeyCache logic. Now that guava cache has done this, there is no need to do this implementation.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
dlg99: there is Caffeine Cache which claims better perf but I don't know if it behaves any better in this case. 
You can as well switch to expireAfterWrite, make it configurable and let users set it. 
Pulsar has configuration for time based ledger rollover, just have to set it is lower than the expireAfterWrite value in bk.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
```diff
@@ -112,8 +113,9 @@ public class BookieImpl extends BookieCriticalThread implements Bookie {
 
     private int exitCode = ExitCode.OK;
 
-    private final ConcurrentLongHashMap<byte[]> masterKeyCache =
-            ConcurrentLongHashMap.<byte[]>newBuilder().build();
+    // key: masterKey, values: set of ledgerId
+    private final static ConcurrentOpenHashMap<byte[], ConcurrentLongHashSet> masterKeyCache
```
dlg99, line 117: this flips cache from ledgerId -> key to key -> set of ledgerids.
While it reduces memory usage in cases of the same key, it does not help if ledgers are stored (not gc'd) for too long, there are too many ledgers etc. It will slow down rate at which cache eats up the memory. 
It also makes get/remove operations for the cache slow.

What's worse is that outside of Pulsar we don't know how people use the keys, what if some user sets a unique key for each ledger? the problem just got worse for them.

dlg99 (changes requested):
The second approach flips cache from ledgerId -> key to key -> set of ledgerids.
While it reduces memory usage in cases of the same key, it does not help if ledgers are stored (not gc'd) for too long, there are too many ledgers etc. It will slow down rate at which cache eats up the memory.
It also makes get/remove operations for the cache slow.

What's worse is that outside of Pulsar we don't know how people use the keys, what if some user sets a unique key for each ledger? the problem just got worse for them.

I'd love to hear at least from some people that using BK without Pulsar if this an acceptable change. @RaulGracia @jvrao

StevenLuMT:
There are currently 3 schemes for masterKeyCache:
1. Create a 1 hour guava cache
    Applicable scenarios:       simple to implement;
     Disadvantages:       the granularity of the lock is large, and the data structure will be locked when cleaning

2. Use the structure ConcurrentLongHashMap, the key is ledgerId, and the value is masterKey;
     Applicable scenarios:      do not set masterKey or masterKey settings are mostly the same;
     Disadvantage:       there will be room to enlarge if the same masterKey is set for ledgerId

3. Use the structure ConcurrentOpenHashMap, the key is masterKey, the value is ledgerId,
Applicable scenarios:      do not set masterKey or masterKey settings are mostly the same
Disadvantages:      the masterKey settings are different, which increases the time-consuming of get/remove

or else I send a email talking ablout it ?
help me have a look at it @eolivelli @dlg99 @hangc0276 @merlimat

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
StevenLuMT: @dlg99 
use this structure ConcurrentOpenHashMap, the key is masterKey, the value is the set of ledgerIds,
the masterKey settings are different, 
1. which increases the time-consuming of remove, but the remove operation is timed, not so frequent
2. when we add the new data ,we can use `containsMasterKeyCache` to spead up weather to `addMasterKeyCache`

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
ben-manes: fyi, in both caches a read does not require any locking on the read path

eolivelli:
I am currently fine with the current approach.
I am not sure that there will be a visible performance impact.

We should do optimisations like a special path for the empty string in follow up works (if really needed).
Regarding the choice of the implementation of the cache...I am not sure that would make a big difference.

Maybe we can do with this patch and do more testing and get numbers: it won't be easy as the hot spots in writing to the journal are not for sure about accessing this cache

HQebupt (approved):
LGTM

StevenLuMT:
rerun failure checks

dlg99 (approved):
Current implementation LGTM

StevenLuMT:
rerun failure checks

nicoloboschi (changes requested):
LGTM, can you fix the typo?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
```diff
@@ -112,7 +112,7 @@ public class BookieImpl extends BookieCriticalThread implements Bookie {
 
     private int exitCode = ExitCode.OK;
 
-    private final ConcurrentLongHashMap<byte[]> masterKeyCache =
+    private static final ConcurrentLongHashMap<byte[]> MASRER_KEY_CACHE =
```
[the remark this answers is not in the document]
StevenLuMT: Thanks

hangc0276, line 115: Do we need to enable the auto-shrink feature of ConcurrentLongHashMap?

StevenLuMT:
rerun failure checks

lordcheng10 (approved):
LGTM

nicoloboschi (approved):

StevenLuMT:
rerun failure checks

hangc0276 (approved):

bookkeeper-server/src/test/java/org/apache/bookkeeper/util/collections/ConcurrentLongHashMapTest.java
```diff
@@ -117,6 +119,41 @@ public void testClear() {
         assertTrue(map.capacity() == 4);
     }
 
+    private void addSpecifyIncrement(ConcurrentLongHashMap<byte[]> mkc, int start, int end) {
+        for (int i = start; i <= end; i++) {
+            assertNull(mkc.put(i, ("comment:" + i).getBytes(UTF_8)));
+            System.out.println("i=" + i + ",mkc.size=" + mkc.size() + ",mkc.capacity=" + mkc.capacity());
```
eolivelli, line 125: nit: use logger

bookkeeper-server/src/test/java/org/apache/bookkeeper/util/collections/ConcurrentLongHashMapTest.java
```diff
@@ -117,6 +119,41 @@ public void testClear() {
...
+        }
+    }
+
+    private void removeSpecifyIncrement(ConcurrentLongHashMap<byte[]> mkc, int start, int end) {
+        for (int i = end; i >= start; i--) {
+            assertNotNull(mkc.remove(i));
+            System.out.println("i=" + i + ",mkc.size=" + mkc.size() + ",mkc.capacity=" + mkc.capacity());
```
eolivelli, line 132: nit: use logger

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -181,6 +181,7 @@ public GarbageCollectorThread(ServerConfiguration conf,
                 }
                 gcStats.getDeletedLedgerCounter().inc();
                 ledgerStorage.deleteLedger(ledgerId);
+                BookieImpl.removeLedgerIdFromMasterKeyCache(ledgerId);
```
eolivelli, line 184: have to at least move this to a Consumer<Long> and decuple GarbageCollectorThread from BookieImpl

also, having a static variables means that if you run multiple bookies in a test case (here in the BK repository) the values of the cache a messed up. Probably it is not a big deal but it is a big code smell

eolivelli (changes requested):

bookkeeper-server/src/test/java/org/apache/bookkeeper/util/collections/ConcurrentLongHashMapTest.java
StevenLuMT: done

bookkeeper-server/src/test/java/org/apache/bookkeeper/util/collections/ConcurrentLongHashMapTest.java
StevenLuMT: done

StevenLuMT:
rerun failure checks

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
StevenLuMT: Good suggestion, I have updated

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerStorage.java
```diff
@@ -60,6 +61,8 @@ void initialize(ServerConfiguration conf,
     void setStateManager(StateManager stateManager);
     void setCheckpointSource(CheckpointSource checkpointSource);
     void setCheckpointer(Checkpointer checkpointer);
+    void setMasterKeyCache(ConcurrentLongHashMap<byte[]> masterKeyCache);
```
eolivelli, line 64: instead of exposing the internal implementation detail (ConcurrentLongHashMap) I would expose here something like:

```
interface LedgerStorageNotificationListener {
   void ledgerRemovedFromStorage(long ledgerId);
}
```

and here
`void addStorageStorageNotificationListener(LedgerStorageNotificationListener listener)`


and we don't need getters

eolivelli (changes requested):
very good,
one last comment, than we are good to go

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
```diff
@@ -477,10 +478,17 @@ public void start() {
...
+            public void ledgerRemovedFromStorage(long ledgerId) {
+                masterKeyCache.remove(ledgerId);
+            }
+        };
+
         ledgerStorage.setStateManager(stateManager);
         ledgerStorage.setCheckpointSource(checkpointSource);
         ledgerStorage.setCheckpointer(syncThread);
-
+        ledgerStorage.addStorageStorageNotificationListener(storageNotificationListener);
```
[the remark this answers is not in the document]
StevenLuMT: uodated

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerStorage.java
StevenLuMT: finished

HQebupt (approved):

eolivelli (approved):

PR #3056, opened 2022-02-16
change rocksdb init: use OptionsUtil

StevenLuMT:
Descriptions of the changes in this PR:

### Motivation

1. some old parameters in rocksDB is not configurable
2. for all the tuning of rocksdb in the future, there is no need to update the code or introduce configuration to bookie

### Changes

1)   rocks all old parameter change to be configurable
2)  use OptionsUtil to init all params for rocksdb

the old pr #3006  has some rebase error,open a new pr

PR #3523, opened 2022-10-12
Make the rocksDB configuration compatible with previous versions

zymap:
---
### Motivation

In PR #3056, we introduced using rocksDB configuration file to configure it. But it is not compatible with the previous versions of bookkeeper. We used to use the bookie configuration file to configure it. It would make the existing configuration doesn't work. Which causes some performance issues for the user's existing environment.

This PR adds the old configuration back. The new way only works if the rocksDB configuration file is existing.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageRocksDB.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageRocksDB.java
@@ -20,7 +20,13 @@
  */
 package org.apache.bookkeeper.bookie.storage.ldb;
 
+
 import static com.google.common.base.Preconditions.checkState;
+//CHECKSTYLE.OFF: IllegalImport
+//CHECKSTYLE.OFF: ImportOrder
+import static io.netty.util.internal.PlatformDependent.maxDirectMemory;
+//CHECKSTYLE.ON: IllegalImport
+//CHECKSTYLE.ON: ImportOrder
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import java.io.IOException;
@@ -32,27 +38,26 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map.Entry;
 import java.util.concurrent.TimeUnit;
-
-import io.netty.util.internal.PlatformDependent;
 import org.apache.bookkeeper.bookie.storage.ldb.KeyValueStorageFactory.DbConfigType;
 import org.apache.bookkeeper.conf.ServerConfiguration;
 import org.rocksdb.BlockBasedTableConfig;
 import org.rocksdb.BloomFilter;
 import org.rocksdb.Cache;
 import org.rocksdb.ChecksumType;
+import org.rocksdb.ColumnFamilyDescriptor;
+import org.rocksdb.ColumnFamilyHandle;
 import org.rocksdb.CompressionType;
+import org.rocksdb.DBOptions;
+import org.rocksdb.Env;
 import org.rocksdb.InfoLogLevel;
 import org.rocksdb.LRUCache;
 import org.rocksdb.Options;
-import org.rocksdb.ColumnFamilyDescriptor;
-import org.rocksdb.ColumnFamilyHandle;
-import org.rocksdb.DBOptions;
-import org.rocksdb.Env;
 import org.rocksdb.OptionsUtil;
 import org.rocksdb.ReadOptions;
 import org.rocksdb.RocksDB;
 import org.rocksdb.RocksDBException;
 import org.rocksdb.RocksIterator;
+import org.rocksdb.RocksObject;
 import org.rocksdb.Slice;
 import org.rocksdb.WriteBatch;
 import org.rocksdb.WriteOptions;
@@ -68,6 +73,8 @@ public class KeyValueStorageRocksDB implements KeyValueStorage {
             new KeyValueStorageRocksDB(defaultBasePath, subPath, dbConfigType, conf);
 
     private final RocksDB db;
+    private RocksObject options;
+    private List<ColumnFamilyDescriptor> columnFamilyDescriptors;
 
     private final WriteOptions optionSync;
     private final WriteOptions optionDontSync;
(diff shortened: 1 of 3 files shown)


zymap:
I am trying to add a test to make sure it uses the right configuration.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageRocksDB.java
```diff
@@ -87,18 +111,33 @@ public KeyValueStorageRocksDB(String basePath, String subPath, DbConfigType dbCo
         this.emptyBatch = new WriteBatch();
 
         String dbFilePath = "";
+        if (dbConfigType == DbConfigType.EntryLocation) {
+            dbFilePath = conf.getEntryLocationRocksdbConf();
+        } else if (dbConfigType == DbConfigType.LedgerMetadata) {
+            dbFilePath = conf.getLedgerMetadataRocksdbConf();
+        } else {
+            dbFilePath = conf.getDefaultRocksDBConf();
+        }
```
eolivelli, line 120: I would add a log line
`log.info("Searching for a RockDB configuration file in {}", dbFilePath)`

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageRocksDB.java
```diff
@@ -87,18 +111,33 @@ public KeyValueStorageRocksDB(String basePath, String subPath, DbConfigType dbCo
...
+        if (Paths.get(dbFilePath).toFile().exists()) {
+            db = initializeRocksDBWithConfFile(basePath, subPath, dbConfigType, conf, readOnly, dbFilePath);
```
eolivelli, line 122: here we log that we have found the file

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageRocksDB.java
```diff
@@ -87,18 +111,33 @@ public KeyValueStorageRocksDB(String basePath, String subPath, DbConfigType dbCo
...
+        } else {
+            db = initializeRocksDBWithBookieConf(basePath, subPath, dbConfigType, conf, readOnly);
```
eolivelli, line 124: here we log that we have't found the file and that we will read the configuration from the main bookkeeper configuration

zymap:
@eolivelli Added tests and logs. PTAL

StevenLuMT (changes requested):
I have a question why It would make the existing configuration doesn't work?
 which config key  causes some performance issues for the user's existing environment?

zymap:
We are using bk_server.conf to configure the rocksDB before, [this change](https://github.com/apache/bookkeeper/pull/3056) will make the existing configuration in bk_server.conf invalid. 
Please see the discussion in the mailing list: https://lists.apache.org/thread/drh4p5prxbcs8gszhxnd1xsv0g48vvbt

StevenLuMT:
> @StevenLuMT We were using bk_server.conf to configure the rocksDB before, [this change](https://github.com/apache/bookkeeper/pull/3056) will make the existing configuration in bk_server.conf invalid. Please see the discussion in the mailing list: https://lists.apache.org/thread/drh4p5prxbcs8gszhxnd1xsv0g48vvbt

have a look this mail reply, I think the implementation of the compatible scheme can have more @zymap @eolivelli 
image

merlimat (approved):

zymap:
@eolivelli @StevenLuMT please take another look when you have time. Thanks

eolivelli (approved):
LGTM

nicoloboschi (approved):
Great work and great catch!

StevenLuMT (approved):
I think the implementation of the compatible scheme can have more
https://lists.apache.org/thread/xxtx9nsz07rywfbgw6rp15ljh6sw5jyw
we can merge this first, and I will mention a pr later to make this thing better

zymap:
Thank you, guys!

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageRocksDB.java
```diff
@@ -109,26 +158,129 @@ public KeyValueStorageRocksDB(String basePath, String subPath, DbConfigType dbCo
...
+        }
     }
 
     @Override
     public void close() throws IOException {
         db.close();
+        if (cache != null) {
+            cache.close();
+        }
+        if (options != null) {
```
hangc0276, line 281: Why do we need this `options` field?

hangc0276:
In BookKeeper and Pulsar's conf folder, it already has default RocksDB configuration files. In this Pr, the loading order is `RocksDB Configuration file` > `conf/bk_server.conf`, which means it will still load the RocksDB configuration file by default and will cause a compatible issue in the BookKeeper version upgrade.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageRocksDB.java
```diff
@@ -20,27 +20,44 @@
  */
 package org.apache.bookkeeper.bookie.storage.ldb;
 
+
 import static com.google.common.base.Preconditions.checkState;
+//CHECKSTYLE.OFF: IllegalImport
```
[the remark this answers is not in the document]
zymap: The checkstyle doesn't allow we use this imports `io.netty.util.internal.PlatformDependent.maxDirectMemory`. We disabled it before https://github.com/apache/bookkeeper/pull/3056/files#diff-cbb728e9ff8fae16673eb48455d2eebc18d12ed8fec6a93266bb9363527dad97L25

zymap:
@hangc0276 We rename the configuration file in this PR https://github.com/apache/bookkeeper/pull/3540

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageRocksDB.java
zymap: It's just for testing. I need to get configurations from the option to test if it is load the right configuration.

Issue #3527, opened 2022-10-13
Direct memory leak of V2 protocol ReadResponse


horizonzy:
**BUG REPORT**
Recently, in our customer product, we found that the bookie direct memory increases slowly over time, finally it throws netty allocate direct memory failed exception, then bookie restart.

**Dump info**
So we dump the heap and found that there are some unreachable `ReadResponse`, and the data property `ByteBuf` refCnt is 2, and the  `ByteBuf` is also unreachable, so there is no chance to release the `ByteBuf`, it will be deleted after JVM GC.
<img width="1563" alt="image" src="https://user-images.githubusercontent.com/22524871/195585954-608d8da5-de32-480e-91a5-d5b535eac6de.png">

See the picture, we can think the memory leak already happen. 

**Log info**
In the bookie log, we found there are many disconnect info like:
```
"Oct 12, 2022 @ 07:47:28.673","","2022-10-12T07:47:28,672+0000 [bookie-io-1-6] INFO  org.apache.bookkeeper.proto.BookieRequestHandler - Channels disconnected: [id: 0x111df957, L:/xx.xx.xx.xx:3181 ! R:/xx.xx.xx.xx:44396]"
```

**The Root Case**
The memory leak occurred in an edge scene.
In the normal case, we get the data from BookieImpl and wrap it in ReadResponse, and then write the [ReadResponse](https://github.com/apache/bookkeeper/blob/62155369d199f36720938e913541b223dab8b047/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtocol.java#L448) to the netty channel, there is our netty handler to handle the ReadResponse. 
* [ResponseEncoder](https://github.com/apache/bookkeeper/blob/62155369d199f36720938e913541b223dab8b047/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtoEncoding.java#L458) It transfer the ReadResponse to ByteBufList, see [encode](https://github.com/apache/bookkeeper/blob/62155369d199f36720938e913541b223dab8b047/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtoEncoding.java#L251)

[...]

PR #3528, opened 2022-10-13
Fix memory leak when reading entry but the connection disconnected.

horizonzy:
Fixes-#3527

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtocol.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtocol.java
@@ -433,7 +433,8 @@ public interface BookieProtocol {
                                  opCode, ledgerId, entryId, errorCode);
         }
 
-        void release0() {
+        boolean release() {
+            return true;
         }
 
         void recycle() {
@@ -461,11 +462,6 @@ public interface BookieProtocol {
 
         ByteBuf getData() {
             return data;
-        }
-
-        @Override
-        public void release0() {
-            this.release();
         }
 
         @Override
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtocol.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtocol.java
@@ -25,6 +25,7 @@ import io.netty.buffer.Unpooled;
 import io.netty.util.Recycler;
 import io.netty.util.Recycler.Handle;
 import io.netty.util.ReferenceCountUtil;
+import io.netty.util.ReferenceCounted;
 import org.apache.bookkeeper.proto.BookkeeperProtocol.AuthMessage;
 import org.apache.bookkeeper.util.ByteBufList;
 
@@ -432,10 +433,7 @@ public interface BookieProtocol {
                                  opCode, ledgerId, entryId, errorCode);
         }
 
-        void retain() {
-        }
-
-        void release() {
+        void release0() {
         }
 
         void recycle() {
@@ -445,7 +443,7 @@ public interface BookieProtocol {
     /**
      * A request that reads data.
      */
-    class ReadResponse extends Response {
+    class ReadResponse extends Response implements ReferenceCounted {
         final ByteBuf data;
 
         ReadResponse(byte protocolVersion, int errorCode, long ledgerId, long entryId) {
@@ -464,15 +462,45 @@ public interface BookieProtocol {
         ByteBuf getData() {
             return data;
         }
-
+        
         @Override
-        public void retain() {
-            data.retain();
+        public void release0() {
+            this.release();
         }
-
+    
         @Override
-        public void release() {
-            data.release();
+        public int refCnt() {
+            return data.refCnt();
+        }
+    
+        @Override
+        public ReferenceCounted retain() {
+            return data.retain();
+        }
+    
+        @Override
+        public ReferenceCounted retain(int increment) {
+            return data.retain(increment);
+        }
+    
+        @Override
+        public ReferenceCounted touch() {
+            return data.touch();
+        }
+    
+        @Override
+        public ReferenceCounted touch(Object hint) {
+            return data.touch(hint);
+        }
+    
+        @Override
+        public boolean release() {
+            return data.release();
+        }
+    
+        @Override
+        public boolean release(int decrement) {
+            return data.release(decrement);
         }
     }
 
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtocol.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtocol.java
@@ -462,42 +462,42 @@ public interface BookieProtocol {
         ByteBuf getData() {
             return data;
         }
-        
+
         @Override
         public void release0() {
             this.release();
         }
-    
+
         @Override
         public int refCnt() {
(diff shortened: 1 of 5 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PacketProcessorBase.java
```diff
@@ -145,12 +152,12 @@ protected void sendResponseAndWait(int rc, Object response, OpStatsLogger statsL
         try {
             ChannelFuture future = channel.writeAndFlush(response);
             if (!channel.eventLoop().inEventLoop()) {
-                future.await();
```
eolivelli, line 148: what's the difference in this case ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtocol.java
```diff
@@ -432,10 +433,7 @@ public String toString() {
                                  opCode, ledgerId, entryId, errorCode);
         }
 
-        void retain() {
-        }
-
-        void release() {
```
eolivelli, line 438: what about keeping `boolean release()`
IIUC your problem is about implementing `ReferenceCounted`

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtocol.java
```diff
@@ -445,7 +443,7 @@ void recycle() {
     /**
      * A request that reads data.
      */
-    class ReadResponse extends Response {
+    class ReadResponse extends Response implements ReferenceCounted {
```
eolivelli, line 446: why do we need to implement ReferenceCounted ?
can you add a comment in the JavaDoc ?

eolivelli (changes requested):

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtocol.java
hangc0276: It has conflict with the release() interface in `ReferenceCounted`

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtocol.java
hangc0276: https://github.com/apache/bookkeeper/issues/3527 will give more context.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PacketProcessorBase.java
hangc0276: `await()` won't throw the exception out. We'd better catch the exception and log it in debug mode to help debug issues.

hangc0276:
@horizonzy Great Catch!
There are some tests that failed, please help take a look.
https://github.com/apache/bookkeeper/actions/runs/3242629439/jobs/5316169295

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PacketProcessorBase.java
```diff
@@ -119,12 +121,17 @@ protected void sendResponse(int rc, Object response, OpStatsLogger statsLogger)
         }
 
         if (channel.isActive()) {
-            channel.writeAndFlush(response, channel.voidPromise());
+            ChannelPromise promise = channel.newPromise().addListener(future -> {
+                if (!future.isSuccess()) {
+                    logger.debug("Netty channel write exception. ", future.cause());
```
zymap, line 126: Why not log it as an error?

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtocol.java
zymap: Yeah. I think we can keep `boolean release()`. Because `ReferenceCounted` also has the same method.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PacketProcessorBase.java
horizonzy: When this error occurs, it means that the client has already close. Log error is unnecessary, the other log will show the connection already close.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PacketProcessorBase.java
```diff
@@ -145,12 +152,12 @@ protected void sendResponseAndWait(int rc, Object response, OpStatsLogger statsL
         try {
             ChannelFuture future = channel.writeAndFlush(response);
             if (!channel.eventLoop().inEventLoop()) {
-                future.await();
+                future.get();
             }
-        } catch (InterruptedException e) {
+        } catch (ExecutionException | InterruptedException e) {
+            logger.debug("Netty channel write exception. ", e);
```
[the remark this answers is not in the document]
horizonzy: Same reason in above.

gaozhangmin (approved):

merlimat (approved):

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtocol.java
horizonzy: 👍

hangc0276 (approved):

michaeljmarshall (approved):
LGTM

Shoothzj (approved):

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PacketProcessorBase.java
StevenLuMT: I think this need to print the error
`logger.error("Netty channel write exception. ", future.cause());`

hangc0276:
rerun failure checks

wolfstudy:
ping @eolivelli PTAL again, thanks

dlg99 (approved):
Nicely done.

zymap (approved):
LGTM

StevenLuMT (approved):
LGTM

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PacketProcessorBase.java
```diff
@@ -119,12 +121,17 @@ protected void sendResponse(int rc, Object response, OpStatsLogger statsLogger)
         }
 
         if (channel.isActive()) {
-            channel.writeAndFlush(response, channel.voidPromise());
+            ChannelPromise promise = channel.newPromise().addListener(future -> {
+                if (!future.isSuccess()) {
+                    logger.debug("Netty channel write exception. ", future.cause());
+                }
+            });
```
lhotari, line 128: This will cause unnecessary object creation. If the only purpose is to log exceptions, `channel.voidPromise()` already takes care of that.

horizonzy: Thanks for reminder.

nicoloboschi: It has been fixed here: https://github.com/apache/bookkeeper/pull/3733

PR #3543, opened 2022-10-17
Fix max upper bound for last bucket of timer stats

Ghatage:

### Motivation

Codahale metrics are used for dashboarding with Grafana and we must rely on these values to be accurate in order to measure the stability and performance of the service.
For Codahale's fast timer, when a value falls in the last bucket, the max is reported as the upper bound of the last bucket (Long.MAX_VALUE) which can lead to erroneous 99 percentile values.

### Changes

We change the way we get the max value recorded if the bucket we're dealing with is the last bucket (for 99th percentile).
If we are in the last bucket, we return the actual max, otherwise we can return `getBucketBound()` which should return the accurate calculated value.
Also added a test.


hangc0276:
@Ghatage Thanks for your contribution, would you please help resolve the conflicts? Thanks a lot.

Ghatage:
Hi @hangc0276 somehow I missed the comment, sorry for that. Fixing the conflicts and getting this PR ready to go.

Ghatage:
@hangc0276 can you PTAL?

shoothzj (approved):

Ghatage:
@eolivelli @nicoloboschi @merlimat PTAL

PR #3544, opened 2022-10-17
Use BlockingQueue.drainTo() in Journal thread

merlimat:
### Motivation

Use `BlockingQueue.drainTo()` in the Journal thread to minimize the contention between this thread and the threads passing entries to the journal. 

The drain method in the `ArrayBlockingQueue` method will use the mutex only once and take out all the items in one shot.

The logic in the journal dequeuing is already quite complicated. This change aims at not changing the whole logic, rather doing the minimal intervention.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
@@ -1083,19 +1083,7 @@ public class Journal extends BookieCriticalThread implements CheckpointSource {
                             .registerSuccessfulEvent(MathUtils.elapsedNanos(dequeueStartTime), TimeUnit.NANOSECONDS);
                     }
 
-                    if (localQueueEntries.isEmpty()) {
-                        queue.drainTo(localQueueEntries);
-                    }
-
-                    if (!localQueueEntries.isEmpty()) {
-                        journalTime.addLatency(MathUtils.elapsedNanos(busyStartTime), TimeUnit.NANOSECONDS);
-                        qe = localQueueEntries.removeFirst();
-                        dequeueStartTime = MathUtils.nowInNano();
-                        busyStartTime = dequeueStartTime;
-                        journalStats.getJournalQueueSize().dec();
-                        journalStats.getJournalQueueStats()
-                                .registerSuccessfulEvent(MathUtils.elapsedNanos(qe.enqueueTime), TimeUnit.NANOSECONDS);
-                    } else if (numEntriesToFlush == 0) {
+                    if (numEntriesToFlush == 0 && localQueueEntries.isEmpty()) {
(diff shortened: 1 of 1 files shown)


zymap (approved):

zymap:
rerun the failure tests

```
Error:  Failures: 
Error:  org.apache.bookkeeper.bookie.BookieJournalForceTest.testAckBeforeSyncWithJournalBufferedEntriesThreshold
Error:    Run 1: BookieJournalForceTest.testAckBeforeSyncWithJournalBufferedEntriesThreshold:251
Error:    Run 2: BookieJournalForceTest.testAckBeforeSyncWithJournalBufferedEntriesThreshold:251
Error:    Run 3: BookieJournalForceTest.testAckBeforeSyncWithJournalBufferedEntriesThreshold:251
[INFO] 
Warning:  Flakes: 
Warning:  org.apache.bookkeeper.bookie.StateManagerTest.testReadOnlyBookieTransitions
Error:    Run 1: StateManagerTest.testReadOnlyBookieTransitions:173->Object.wait:-2 » TestTimedOut
[INFO]   Run 2: PASS
[INFO] 
Warning:  org.apache.bookkeeper.bookie.UpgradeTest.testUpgradeCurrent
Error:    Run 1: UpgradeTest.testUpgradeCurrent:228->testUpgradeProceedure:179->Object.wait:-2 » TestTimedOut
[INFO]   Run 2: PASS
```

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -1080,7 +1083,19 @@ journalFormatVersionToWrite, getBufferedChannelBuilder(),
                             .registerSuccessfulEvent(MathUtils.elapsedNanos(dequeueStartTime), TimeUnit.NANOSECONDS);
                     }
 
-                    if (numEntriesToFlush == 0) {
+                    if (localQueueEntries.isEmpty()) {
+                        queue.drainTo(localQueueEntries);
+                    }
+
+                    if (!localQueueEntries.isEmpty()) {
```
hangc0276, line 1090: If new entries keep put into the `queue`, the `localQueueEntries` won't be empty, it won't go into line#1107 branch, and the flush won't be triggered.

merlimat: I'm not sure I got the scenario here. I think every time the loop goes back to the top on the `while (true)` and we should always check first the `localQueueEntries`  before attempting the `poll()`

merlimat: Very good points. I actually think we should do it both before the `poll()` and the `take()`. 

@zymap @hangc0276  PTAL again

PR #3545, opened 2022-10-17
Use BlockingQueue.drainTo() in JournalForceWrite thread

merlimat:
### Motivation

Instead of dequeuing all the `ForceWriteRequest` one by one, use `drainTo()` to reduce contention with the main journal thread.


zymap (approved):

hangc0276 (approved):
Great job!

PR #3546, opened 2022-10-17
Use SingleThreadExecutor for OrderedExecutor and drainTo() tasks into local array

merlimat:
### Motivation

`OrderedExecutor` is used extensively throughout the BK codebase and its performance is critical when there are many small entries.

In the past we have experimented with different `BlockingQueue` backends inside the JDK default `ThreadPoolExecutor` implementation. 

The main problem with `ThreadPoolExecutor` is that is not using the `drainTo()` to avoid contention, because it's not making any assumptions on the queue type. If we assume an array based queue (like `ArrayBlockingQueue`), then the `drainTo()` makes it much more efficient.

A single executor with 16 thread submitting tasks can now process 46M tasks/s


zymap:
Error:  Failures: 
Error:  org.apache.bookkeeper.common.util.TestSingleThreadExecutor.testShutdownNow
Error:    Run 1: TestSingleThreadExecutor.testShutdownNow:183 expected:<2> but was:<0>
Error:    Run 2: TestSingleThreadExecutor.testShutdownNow:183 expected:<2> but was:<0>
Error:    Run 3: TestSingleThreadExecutor.testShutdownNow:183 expected:<2> but was:<0>

There are some failure tests related to this change.

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/OrderedExecutor.java
```diff
@@ -434,43 +430,46 @@ protected OrderedExecutor(String baseName, int numThreads, ThreadFactory threadF
...
-                public Number getDefaultValue() {
-                    return 0;
-                }
-
-                @Override
-                public Number getSample() {
-                    return thread.getTaskCount();
-                }
-            });
+            if (thread instanceof ThreadPoolExecutor) {
```
hangc0276, line 433: The thread created by `createSingleThreadExecutor` is not a subclass of `ThreadPoolExecutor`, which means we will lose the following thread metrics.

merlimat: Good point. Also, I think there is an opportunity to add more stats around rejected & failed tasks. It would be very useful to see these in the charts.

merlimat:
> Error: Failures: Error: org.apache.bookkeeper.common.util.TestSingleThreadExecutor.testShutdownNow Error: Run 1: TestSingleThreadExecutor.testShutdownNow:183 expected:<2> but was:<0> Error: Run 2: TestSingleThreadExecutor.testShutdownNow:183 expected:<2> but was:<0> Error: Run 3: TestSingleThreadExecutor.testShutdownNow:183 expected:<2> but was:<0>
> 
> There are some failure tests related to this change.

@zymap Fixed the test, PTAL

hangc0276:
Error:  Failures: 
Error:  org.apache.bookkeeper.common.util.TestSingleThreadExecutor.testRejectWhenQueueIsFull
Error:    Run 1: TestSingleThreadExecutor.testRejectWhenQueueIsFull:100 should have rejected the task
Error:    Run 2: TestSingleThreadExecutor.testRejectWhenQueueIsFull:100 should have rejected the task
Error:    Run 3: TestSingleThreadExecutor.testRejectWhenQueueIsFull:100 should have rejected the task
[INFO] 
Error:  org.apache.bookkeeper.common.util.TestSingleThreadExecutor.testSimple
Error:    Run 1: TestSingleThreadExecutor.testSimple:65 expected:<11> but was:<10>
Error:    Run 2: TestSingleThreadExecutor.testSimple:65 expected:<11> but was:<10>
Error:    Run 3: TestSingleThreadExecutor.testSimple:65 expected:<11> but was:<10>
[INFO] 
Error:  org.apache.bookkeeper.common.util.TestSingleThreadExecutor.testTasksWithNPE
Error:    Run 1: TestSingleThreadExecutor.testTasksWithNPE:265 expected:<1> but was:<0>
Error:    Run 2: TestSingleThreadExecutor.testTasksWithNPE:265 expected:<1> but was:<0>
Error:    Run 3: TestSingleThreadExecutor.testTasksWithNPE:265 expected:<1> but was:<0>
[INFO] 
Warning:  Flakes: 
Warning:  org.apache.bookkeeper.common.util.TestSingleThreadExecutor.testTasksWithException
Error:    Run 1: TestSingleThreadExecutor.testTasksWithException:240 expected:<1> but was:<0>

@merlimat Please take a look at the above-failed tests.

bookkeeper-common/src/test/java/org/apache/bookkeeper/common/util/TestSingleThreadExecutor.java
```diff
@@ -0,0 +1,290 @@
...
+import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicInteger;
+import lombok.Cleanup;
+import org.awaitility.Awaitility;
+import org.junit.Test;
+
+/**
+ * Unit test for {@link SingleThreadExecutor}.
+ */
+public class TestSingleThreadExecutor {
```
hangc0276, line 43: Can we add a test to validate the SingleThreadExecutor metrics?

hangc0276 (approved):
Great job!

bookkeeper-common/src/test/java/org/apache/bookkeeper/common/util/TestSingleThreadExecutor.java
merlimat: I've added assertions to verify the values of the counters. Do you mean to read the gauges out of the StatsLogger?

PR #3547, opened 2022-10-18
Adapt thread number automatically

wolfstudy:
Signed-off-by: xiaolongran <xiaolongran@tencent.com>

### Motivation

In actual production scenarios, the default configuration of the number of read and write threads provided by Bookie is relatively small, and we generally need to configure the number of read and write threads in the configuration file. In fact, for machines with different CPU cores, we can adopt a certain default strategy to automatically adapt the number of threads for reading and writing.

### Changes

 
- Add `ThreadNumUtils` func
- Add test case for this change


shoothzj:
closed for no update.

shoothzj:
also see #3099

PR #3516, opened 2022-10-04
LedgerHandle: do not complete metadata operation on the ZookKeeper/Metadata callback thread

eolivelli:
### Motivation

We are currently completing the creation and opening (also with recovery) of the LedgerHandle in the main ZookKeeper (or metadata driver) callback thread.

This leads to unpredictable code to be executed on that thread:
- possible deadlocks (that cannot be prevented fully from BK code)
- execute heavy weight operations, like opening BK connections

This is a good example of a something that tries to open a connection to a bookie because the application executes a "read" after a "openLedger" operation.

```
at org.apache.bookkeeper.proto.PerChannelBookieClient.connect(PerChannelBookieClient.java:532)
	at org.apache.bookkeeper.proto.PerChannelBookieClient.connectIfNeededAndDoOp(PerChannelBookieClient.java:658)
	at org.apache.bookkeeper.proto.DefaultPerChannelBookieClientPool.initialize(DefaultPerChannelBookieClientPool.java:92)
	at org.apache.bookkeeper.proto.BookieClientImpl.lookupClient(BookieClientImpl.java:217)
	at org.apache.bookkeeper.proto.BookieClientImpl.isWritable(BookieClientImpl.java:170)
	at org.apache.bookkeeper.client.LedgerHandle.isWriteSetWritable(LedgerHandle.java:1227)
	at org.apache.bookkeeper.client.LedgerHandle.waitForWritable(LedgerHandle.java:1249)
	at org.apache.bookkeeper.client.LedgerHandle.readEntriesInternalAsync(LedgerHandle.java:883)
	at org.apache.bookkeeper.client.LedgerHandle.asyncReadEntriesInternal(LedgerHandle.java:800)
	at org.apache.bookkeeper.client.LedgerHandle.asyncReadEntries(LedgerHandle.java:694)
	at org.apache.pulsar.broker.service.schema.BookkeeperSchemaStorage$Functions.getLedgerEntry(BookkeeperSchemaStorage.java:646)
	at org.apache.pulsar.broker.service.schema.BookkeeperSchemaStorage.lambda$readSchemaEntry$33(BookkeeperSchemaStorage.java:524)
	at org.apache.pulsar.broker.service.schema.BookkeeperSchemaStorage$$Lambda$820/0x00000008007e5840.apply(Unknown Source)
	at java.util.concurrent.CompletableFuture$UniCompose.tryFire(java.base@11.0.15.0.1/CompletableFuture.java:1072)
[...]

PR #3548, opened 2022-10-18
Using a separate thread pool to execute openWithMetadata

### LINKED ISSUE #2398: LedgerOpenOp may perform connection to a bookie inside the ZooKeeper client thread
While working on BP-41 I saw this potential problem.
During OpenLedgerOp (openLedger) we are executing blocking I/O operations against the bookies in the main thread of ZooKeeper client,
This may slow down the whole processing of the client because the ZooKeeper client has only one single thread that execute callbacks.
See the stacktrace (it is not an error, it is just a dump)

Suggestions:
- use a separate thread pool in order to execute blocking I/O 

```

2020-08-14 18:46:33,190 - INFO  - [AuditorBookie-127.0.0.1:37597-EventThread:CachingBookieAddressResolver@50] - resolve 192.0.2.1:1000 current cache {127.0.0.1:44827=127.0.0.1:44827, 127.0.0.1:33607=127.0.0.1:33607, 127.0.0.1:37597=127.0.0.1:37597}
	at org.apache.bookkeeper.proto.PerChannelBookieClient.connect(PerChannelBookieClient.java:541)
	at org.apache.bookkeeper.proto.PerChannelBookieClient.connectIfNeededAndDoOp(PerChannelBookieClient.java:666)
	at org.apache.bookkeeper.proto.DefaultPerChannelBookieClientPool.intialize(DefaultPerChannelBookieClientPool.java:92)
	at org.apache.bookkeeper.proto.BookieClientImpl.lookupClient(BookieClientImpl.java:217)
	at org.apache.bookkeeper.proto.BookieClientImpl.readEntry(BookieClientImpl.java:502)
	at org.apache.bookkeeper.proto.BookieClientImpl.readEntry(BookieClientImpl.java:495)
	at org.apache.bookkeeper.proto.BookieClientImpl.readEntry(BookieClientImpl.java:489)
	at org.apache.bookkeeper.client.ReadLastConfirmedOp.initiate(ReadLastConfirmedOp.java:81)
	at org.apache.bookkeeper.client.LedgerHandle.asyncReadLastConfirmed(LedgerHandle.java:1416)
	at org.apache.bookkeeper.client.LedgerOpenOp.openWithMetadata(LedgerOpenOp.java:210)
	at org.apache.bookkeeper.client.LedgerOpenOp.lambda$initiate$0(LedgerOpenOp.java:119)
	at java.util.concurrent.CompletableFuture.uniWhenComplete(CompletableFuture.java:774)
	at java.util.concurrent.CompletableFuture$UniWhenComplete.tryFire(CompletableFuture.java:750)
	at java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:488)
	at java.util.concurrent.CompletableFuture.complete(CompletableFuture.java:1975)
	at org.apache.bookkeeper.meta.AbstractZkLedgerManager$3.processResult(AbstractZkLedgerManager.java:464)
	at org.apache.bookkeeper.zookeeper.ZooKeeperClient$19$1.processResult(ZooKeeperClient.java:994)
	at org.apache.zookeeper.ClientCnxn$EventThread.processEvent(ClientCnxn.java:598)
	at org.apache.zookeeper.ClientCnxn$EventThread.run(ClientCnxn.java:510)
2020-08-14 18:46:43,243 - INFO  - [Time-limited test:Auditor@1946] - Shutting down auditor
```

gaozhangmin:
Fixes #2398

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerOpenOp.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerOpenOp.java
@@ -36,7 +36,6 @@ import org.apache.bookkeeper.client.api.BKException.Code;
 import org.apache.bookkeeper.client.api.LedgerMetadata;
 import org.apache.bookkeeper.client.api.ReadHandle;
 import org.apache.bookkeeper.client.impl.OpenBuilderBase;
-import org.apache.bookkeeper.common.util.OrderedScheduler;
 import org.apache.bookkeeper.stats.OpStatsLogger;
 import org.apache.bookkeeper.util.MathUtils;
 import org.apache.bookkeeper.util.OrderedGenericCallback;
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerOpenOp.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerOpenOp.java
@@ -36,7 +36,6 @@ import org.apache.bookkeeper.client.api.BKException.Code;
 import org.apache.bookkeeper.client.api.LedgerMetadata;
 import org.apache.bookkeeper.client.api.ReadHandle;
 import org.apache.bookkeeper.client.impl.OpenBuilderBase;
-import org.apache.bookkeeper.common.util.OrderedExecutor;
 import org.apache.bookkeeper.common.util.OrderedScheduler;
 import org.apache.bookkeeper.stats.OpStatsLogger;
 import org.apache.bookkeeper.util.MathUtils;
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerOpenOp.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerOpenOp.java
@@ -261,6 +261,7 @@ class LedgerOpenOp {
         } else {
             cb.openComplete(rc, null, ctx);
         }
+        scheduler.shutdown();
     }
 
     static final class OpenBuilderImpl extends OpenBuilderBase {
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerOpenOp.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerOpenOp.java
@@ -36,6 +36,8 @@ import org.apache.bookkeeper.client.api.BKException.Code;
 import org.apache.bookkeeper.client.api.LedgerMetadata;
 import org.apache.bookkeeper.client.api.ReadHandle;
 import org.apache.bookkeeper.client.impl.OpenBuilderBase;
+import org.apache.bookkeeper.common.util.OrderedExecutor;
+import org.apache.bookkeeper.common.util.OrderedScheduler;
 import org.apache.bookkeeper.stats.OpStatsLogger;
 import org.apache.bookkeeper.util.MathUtils;
 import org.apache.bookkeeper.util.OrderedGenericCallback;
@@ -64,6 +66,7 @@ class LedgerOpenOp {
 
     final DigestType suggestedDigestType;
     final boolean enableDigestAutodetection;
+    private OrderedScheduler scheduler;
 
     /**
      * Constructor.
@@ -86,6 +89,7 @@ class LedgerOpenOp {
         this.enableDigestAutodetection = bk.getConf().getEnableDigestTypeAutodetection();
         this.suggestedDigestType = digestType;
         this.openOpLogger = clientStats.getOpenOpLogger();
+        this.scheduler = OrderedScheduler.newSchedulerBuilder().numThreads(1).name("LedgerOpenOp").build();
     }
 
     public LedgerOpenOp(BookKeeper bk, BookKeeperClientStats clientStats,
@@ -100,6 +104,7 @@ class LedgerOpenOp {
         this.enableDigestAutodetection = false;
         this.suggestedDigestType = bk.conf.getBookieRecoveryDigestType();
         this.openOpLogger = clientStats.getOpenOpLogger();
+        this.scheduler = OrderedScheduler.newSchedulerBuilder().numThreads(1).name("LedgerOpenOp").build();
     }
 
     /**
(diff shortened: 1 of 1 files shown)


gaozhangmin:
@eolivelli PTAL

gaozhangmin:
rerun failure checks

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerOpenOp.java
```diff
@@ -86,6 +88,7 @@ public LedgerOpenOp(BookKeeper bk, BookKeeperClientStats clientStats,
         this.enableDigestAutodetection = bk.getConf().getEnableDigestTypeAutodetection();
         this.suggestedDigestType = digestType;
         this.openOpLogger = clientStats.getOpenOpLogger();
+        this.scheduler = OrderedScheduler.newSchedulerBuilder().numThreads(1).name("LedgerOpenOp").build();
```
eolivelli, line 91: No need to create a new pool.
As we know the ledgerId we can probably use the same thread used for all the operations on the ledger

eolivelli (changes requested):

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerOpenOp.java
gaozhangmin: We cannot use executor in bk,  if bk is closed unexpectedly, the operation will never completed.

eolivelli: we cannot start a new thread per each LederOpenOp

if the BK client closes all the pending operations must fail somehow. we should fix that
otherwise there will be other operations that will never complete

eolivelli: "some unit test would stuck forever"

maybe  we are missing to mock the executor

eolivelli:
how this patch related to this ?
https://github.com/apache/bookkeeper/pull/3516

eolivelli (changes requested):
probably this patch is no more needed ?
with https://github.com/apache/bookkeeper/pull/3516 the problem is fixed

can you double check please ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerOpenOp.java
gaozhangmin: fixed. @eolivelli

gaozhangmin:
> probably this patch is no more needed ? with #3516 the problem is fixed
> 
> can you double check please ?

@eolivelli , it's not related,  we should execute `openWithMetadata` in separated thread

gaozhangmin:
rerun failure checks

eolivelli (approved):
Lgtm

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerOpenOp.java
```diff
@@ -112,12 +112,10 @@ public void initiate() {
...
         bk.getLedgerManager().readLedgerMetadata(ledgerId)
-            .whenComplete((metadata, exception) -> {
-                    if (exception != null) {
-                        openComplete(BKException.getExceptionCode(exception), null);
-                    } else {
-                        openWithMetadata(metadata);
-                    }
+                .thenAcceptAsync(this::openWithMetadata, bk.getScheduler().chooseThread(ledgerId))
+                .exceptionally(exception -> {
+                    openComplete(BKException.getExceptionCode(exception), null);
```
eolivelli, line 117: After thinking more...
We have to execute also this callback in a separate thread.
I suggest to use whenCompleteAsync

eolivelli (changes requested):

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerOpenOp.java
gaozhangmin: Then, we need to create a new executor here if bk is closed.

gaozhangmin: Then, we need to create new executor here  for exception case.

eolivelli:
```
.whenCompleteAsyncAsync((m, exception) -> {
if (exception != null) {
 openComplete(BKException.getExceptionCode(exception), null);
}  else {
this.openWithMetadata(m);
}
}, bk.getScheduler().chooseThread(ledgerId))
```

gaozhangmin: `org.apache.bookkeeper.client.BookKeeperCloseTest#testBookKeeperAdmin`  This test will be stuck, if change made

gaozhangmin: This happened if bk is closed, we cannot go inside `whenCompleteAsyncAsync` , because bk's scheduler is shutdown.

dlg99 (approved):

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerOpenOp.java
gaozhangmin: FYI @eolivelli

gaozhangmin:
@eolivelli  PTAL.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerOpenOp.java
eolivelli: So when the BK closed the treadpool is shutdown and the callback isn't executed.
This looks like a problem that can happen on every callback.

Let's keep this patch as it is here but please open a ticket to track this bad problem.

If the BK client is closed then probably the application is shutting down  and it is a critical point in the life cycle

eolivelli (approved):
I am approving this patch.
Please take a look to my comment.
It looks like we have a serious problem.

PR #3554, opened 2022-10-19
In avoid to call zk frequently, using watcher to watch the change

### LINKED ISSUE #3358: ZkLedgerUnderreplicationManager: cache responses / reduce number of zk calls by isLedgerReplicationEnabled()
It looks like ZkLedgerUnderreplicationManager.isLedgerReplicationEnabled() can cache the state and either use watcher (see notifyLedgerReplicationEnabled) + enableLedgerReplication/disableLedgerReplication or just some time interval (10sec or so) to cache the value for.

AFAICT Auditor.checkAllLedgers() calls it rather frequently

gaozhangmin:
Fixes #3358

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/LedgerUnderreplicationManager.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/LedgerUnderreplicationManager.java
@@ -150,6 +150,16 @@ public interface LedgerUnderreplicationManager extends AutoCloseable {
 
     /**
      * Receive notification asynchronously when the ledger replication process
+     * is changed.
+     *
+     * @param cb
+     *            - callback implementation to receive the notification
+     */
+    void notifyLedgerReplicationStatusChanged(GenericCallback<Void> cb)
+            throws ReplicationException.UnavailableException;
+
+    /**
+     * Receive notification asynchronously when the ledger replication process
      * is enabled.
      *
      * @param cb
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/ZkLedgerUnderreplicationManager.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/ZkLedgerUnderreplicationManager.java
@@ -738,6 +738,37 @@ public class ZkLedgerUnderreplicationManager implements LedgerUnderreplicationMa
     }
 
     @Override
+    public void notifyLedgerReplicationStatusChanged(final GenericCallback<Void> cb)
+            throws ReplicationException.UnavailableException {
+        if (LOG.isDebugEnabled()) {
+            LOG.debug("notifyLedgerReplicationStatusChanged()");
+        }
+        Watcher w = e -> {
+            if (e.getType() == Watcher.Event.EventType.NodeDeleted) {
+                LOG.info("LedgerReplication is enabled externally through Zookeeper, "
+                        + "since DISABLE_NODE ZNode is deleted");
+                cb.operationComplete(0, null);
+            } else if (e.getType() == Watcher.Event.EventType.NodeCreated) {
+                LOG.info("LedgerReplication is disabled externally through Zookeeper, "
+                        + "since DISABLE_NODE ZNode is created");
+                cb.operationComplete(1, null);
+            }
+        };
+        try {
+            zkc.addWatch(basePath + '/'
+                    + BookKeeperConstants.DISABLE_NODE, w, AddWatchMode.PERSISTENT);
+        } catch (KeeperException ke) {
+            LOG.error("Error while checking the state of "
+                    + "ledger re-replication", ke);
+            throw ReplicationException.fromKeeperException("Error contacting zookeeper", ke);
+        } catch (InterruptedException ie) {
+            Thread.currentThread().interrupt();
+            throw new ReplicationException.UnavailableException(
+                    "Interrupted while contacting zookeeper", ie);
+        }
+    }
+
+    @Override
     public void notifyLedgerReplicationEnabled(final GenericCallback<Void> cb)
             throws ReplicationException.UnavailableException {
         if (LOG.isDebugEnabled()) {
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
@@ -72,6 +72,7 @@ import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.Semaphore;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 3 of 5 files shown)


gaozhangmin:
@dlg99  PTAL Thx

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/ZkLedgerUnderreplicationManager.java
```diff
@@ -737,6 +737,40 @@ public boolean isLedgerReplicationEnabled()
...
+                    cb.operationComplete(0, null);
+                } else if (e.getType() == Watcher.Event.EventType.NodeCreated) {
+                    LOG.info("bLedgerReplication is disabled externally through Zookeeper, "
+                            + "since DISABLE_NODE ZNode is created");
+                    cb.operationComplete(1, null);
+                }
+            }
+        };
+        try {
+            zkc.addWatch(basePath + '/'
```
eolivelli, line 761: We must implement this for every supported Metadata driver 
What happens if the driver doesn't implement it?

eolivelli (changes requested):

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
```diff
@@ -167,6 +167,7 @@ public class Auditor implements AutoCloseable {
     private final int zkOpTimeoutMs;
     private final Semaphore openLedgerNoRecoverySemaphore;
     private final int openLedgerNoRecoverySemaphoreWaitTimeoutMSec;
+    private boolean isLedgerReplicationEnabled;
```
eolivelli, line 170: This should be a AtomicBoolean as it is accessed by multiple threads

eolivelli (changes requested):

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/ZkLedgerUnderreplicationManager.java
gaozhangmin: If this, We also have this problem  currently.  see `notifyLedgerReplicationEnabled `

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/ZkLedgerUnderreplicationManager.java
```diff
@@ -737,6 +737,40 @@ public boolean isLedgerReplicationEnabled()
         }
     }
 
+    @Override
+    public void notifyLedgerReplicationStatusChanged(final GenericCallback<Void> cb)
+            throws ReplicationException.UnavailableException {
```
dlg99, line 742: ReplicationException.fromKeeperException can throw NonRecoverableReplicationException

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
```diff
@@ -818,11 +821,6 @@ public void run() {
                         LOG.error("Exception running periodic check", bke);
                     } catch (IOException ioe) {
                         LOG.error("I/O exception running periodic check", ioe);
-                    } catch (ReplicationException.NonRecoverableReplicationException nre) {
-                        LOG.error("Non Recoverable Exception while reading from ZK", nre);
-                        submitShutdownTask();
```
dlg99, line 823: why is this removed / where NonRecoverableReplicationException is handled now?

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
```diff
@@ -1328,18 +1326,8 @@ void checkAllLedgers() throws BKException, IOException, InterruptedException {
...
-                try {
-                    if (!ledgerUnderreplicationManager.isLedgerReplicationEnabled()) {
-                        LOG.info("Ledger rereplication has been disabled, aborting periodic check");
-                        FutureUtils.complete(processFuture, null);
-                        return;
-                    }
-                } catch (ReplicationException.NonRecoverableReplicationException nre) {
-                    LOG.error("Non Recoverable Exception while reading from ZK", nre);
-                    submitShutdownTask();
-                    return;
```
dlg99, line 1340: why is this removed / where NonRecoverableReplicationException is handled now?

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/LedgerUnderreplicationManager.java
```diff
@@ -148,6 +148,16 @@ void enableLedgerReplication()
...
             throws ReplicationException.UnavailableException;
 
+    /**
+     * Receive notification asynchronously when the ledger replication process
+     * is changed.
+     *
+     * @param cb
+     *            - callback implementation to receive the notification
+     */
+    void notifyLedgerReplicationStatusChanged(GenericCallback<Void> cb)
```
dlg99, line 158: something like
``` java
default void notifyLedgerReplicationStatusChanged(GenericCallback<Void> cb) throws ReplicationException.UnavailableException {
 throw new RuntimeException("not implemented");
}
```

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
dlg99: +1

dlg99 (changes requested):

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
gaozhangmin: it's handled where `notifyLedgerReplicationStatusChanged` is called

gaozhangmin:
@eolivelli  @dlg99  PTAL

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/LedgerUnderreplicationManager.java
```diff
@@ -148,6 +148,18 @@ void enableLedgerReplication()
...
+    /**
+     * Receive notification asynchronously when the ledger replication process
+     * is changed.
+     *
+     * @param cb
+     *            - callback implementation to receive the notification
+     */
+    default void notifyLedgerReplicationStatusChanged(GenericCallback<Void> cb)
+            throws ReplicationException.UnavailableException {
+        throw new RuntimeException("not implemented");
```
merlimat, line 160: This would be breaking the (runtime) compatibility for an existing `LedgerUnderReplicationManager` that does not implement the method, since it will throw exception.

merlimat (commented):
I think the repeated calls to ZK are not be a problem in Pulsar, since 2.10, because the calls to ZK are already internally cached.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/LedgerUnderreplicationManager.java
gaozhangmin: Remove default  method as current now.

gaozhangmin:
rerun failure checks

PR #3555, opened 2022-10-19
Fix flaky-test testWriteSetWriteableCheck

wenbingshen:
### Motivation
 `testWriteSetWriteableCheck` always failed. After investigation, it is found that `isWritable=false` is set when the simulated channel is not writable, but the unit test cannot fully perceive the channel establishment, because the channel is completely established asynchronously. The isWritable state of false is changed to true again.

you can see the following logs:
image

The smallest change is that we can ensure that the channel is established before changing the channel state, such as executing `bkc.getBookieInfo()`, and then `setTargetChannelState`

### DIFF
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/SlowBookieTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/SlowBookieTest.java
@@ -323,6 +323,10 @@ public class SlowBookieTest extends BookKeeperClusterTestCase {
             // b2 or b3 is no more writeable
             int slowBookieIndex = writeSet.get(ThreadLocalRandom.current().nextInt(writeSet.size()));
             List<BookieId> curEns = lh.getCurrentEnsemble();
+
+            // Trigger connection to the bookie service first
+            bkc.getBookieInfo().get(curEns.get(slowBookieIndex));
+            // then mock channel is not writable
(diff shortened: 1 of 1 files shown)


wenbingshen:
ping @dlg99 @zymap @eolivelli @hangc0276 @StevenLuMT PTAL. Thanks.

wenbingshen:
rerun failure checks

dlg99:
@wenbingshen Do you know what broke the test? I believe it was passing / not flaking on the CI.

wenbingshen:
> I believe it was passing / not flaking on the CI.

@dlg99 Thanks for your comments. I've documented in my notes the failure of testWriteSetWriteableCheck in the following CI test, check it out:
https://github.com/apache/bookkeeper/actions/runs/3125949264/jobs/5070881119
https://github.com/apache/bookkeeper/actions/runs/3124965464/jobs/5068848112

As you can see it often fails. Likewise, when I tested my local, I ran 10 times and failed 7 times.
image


>  Do you know what broke the test? 

Please check the picture below:

I printed the log before and after setting the channel `isWritable` flag . It can be seen that the test thread sets the channel to be unreadable, which occurs before the ChannelFuture listener makeWritable is triggered after the connection is established, causing the result of setTargetChannelState=false to be changed to true again.

image

dlg99 (approved):

hangc0276 (approved):
Nice investigation!

PR #3557, opened 2022-10-20
[Branch-4.14] Fix TLS test local port case problem

horizonzy:

When we config as follow:
```
            conf.setDisableServerSocketBind(true);
            conf.setEnableLocalTransport(true);
```
The BookieNettyServer will create jvmBootstrap, it binds the address by bookieId.

https://github.com/apache/bookkeeper/blob/255416a8552f34e5fd50231bbeed77e0945b63e3/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieNettyServer.java#L374-L428
At line_426, it use bookieId to bind, there is a flag about `BookKeeperClusterTestCase#useUUIDasBookieId`, it set uuid to bookieId.
When the client connects to the server, the remoteAddress `local:a4367111-5740-40dd-a8b2-0fcb2995398b` is strange, the resolve logic  throw java.lang.ArrayIndexOutOfBoundsException.
https://github.com/apache/bookkeeper/blob/255416a8552f34e5fd50231bbeed77e0945b63e3/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java#L1502-L1515

The master branch looks good, The pr #2723 remove logic.
```
        // use a random BookieId
        if (useUUIDasBookieId) {
            conf.setBookieId(UUID.randomUUID().toString());
        }
```

### DIFF
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieNetworkAddressChangeTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieNetworkAddressChangeTest.java
@@ -34,6 +34,7 @@ import org.apache.bookkeeper.conf.ServerConfiguration;
 import org.apache.bookkeeper.discover.ZKRegistrationClient;
 import org.apache.bookkeeper.test.BookKeeperClusterTestCase;
 import org.apache.bookkeeper.util.PortManager;
+import org.junit.Ignore;
 import org.junit.Test;
 
 /**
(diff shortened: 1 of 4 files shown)


horizonzy:
@hangc0276 @dlg99

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/BookieNetworkAddressChangeTest.java
```diff
@@ -88,6 +89,7 @@ public void testFollowBookieAddressChange() throws Exception {
     }
 
     @Test
+    @Ignore("PLSR-1850 Seems like restart of the bookie always comes up on same port hence failing this test")
```
horizonzy, line 92: Copy from master.

hangc0276 (approved):

zymap (approved):

PR #3561, opened 2022-10-20
Skipping placementPolicyCheck when ledger replication disabled

wenbingshen:
### Motivation

When `placementPolicyCheck` is enabled and then `bookkeeper shell autorecovery -disable` is executed, `placementPolicyCheck` will detect ledgers that do not satisfy `placementPolicy` and write to zookeeper, but `ReplicationWorker` no longer obtains ledgers from zookeeper for replication work because autorecovery is disabled, which results in a large number of temporary nodes on zookeeper , when there are more ledgers that do not satisfy placementPolicy, the problem will get worse.

The method of `ReplicationWorker` to get ledger:
```java
private boolean rereplicate() throws InterruptedException, BKException,
            UnavailableException {
        long ledgerIdToReplicate = underreplicationManager
                .getLedgerToRereplicate();
       
       ...
```

So we should also disable `placementPolicyCheck`.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
@@ -878,6 +878,11 @@ public class Auditor implements AutoCloseable {
                 @Override
                 public void run() {
                     try {
+                        if (!ledgerUnderreplicationManager.isLedgerReplicationEnabled()) {
+                            LOG.info("Ledger replication disabled, skipping placementPolicyCheck");
+                            return;
+                        }
+
                         Stopwatch stopwatch = Stopwatch.createStarted();
                         LOG.info("Starting PlacementPolicyCheck");
                         placementPolicyCheck();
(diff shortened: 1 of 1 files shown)


wenbingshen:
ping @zymap @dlg99 @eolivelli @hangc0276 @Shoothzj PTAL. Thanks.

hangc0276 (approved):

dlg99 (approved):
This seems to be in line with the rest of Auditor's logic, looks like all the tasks scheduled by
```
            scheduleBookieCheckTask();
            scheduleCheckAllLedgersTask();
            schedulePlacementPolicyCheckTask();
            scheduleReplicasCheckTask();
```
won't do anything.

This brings another question: we also don't see what's going on in this case (independent of this change).
E.g. pausing AR for some time (maintenance window etc.) is fine but is it ok to not update counts of underreplicated ledgers?

zymap (approved):

wenbingshen:
> This brings another question: we also don't see what's going on in this case (independent of this change). E.g. pausing AR for some time (maintenance window etc.) is fine but is it ok to not update counts of underreplicated ledgers?

@dlg99 Good catch. Thanks for your review and comments. :)
AutoRecovery's responsibilities may need to be divided into the following departments:

Auditor: 1. Different ledger status detection. 
              2. and initiate ledger recovery tasks.
ReplicationWorker: Execute tasks triggered by Auditor.

maybe disable AutoRecovery needs to be divided into 2 parts:
1. Completely prohibit AutoRecovery work, including Auditor, ReplicationWorker
2. Only the replication work of AutoRecovery is prohibited, and the Auditor still performs the detection work and updates the relevant indicators.

This should be a new feature, we need other discussions and PRs.

Issue #3564, opened 2022-10-21
Remove forceWrite call if syncData is disabled during journal creation 


zhaohaidao:
**FEATURE REQUEST**

1. Try to optimize tail latency caused by forceWrite call duringjournal creation

3.  should-have
   Are you currently using any workarounds to address this issue? yes

4. Provide any additional detail on your proposed use case for this feature.

PR #3565, opened 2022-10-21
Do not call forceWrite if syncData is disabled during Journal Creation

zhaohaidao:
Try to optimize tail latency caused by forceWrite calls during journal creation

### Motivation

During the performance test of bookkeeper, it was observed that there was a serious queue backlog in journal, and then the flush time did not increase significantly, maintaining at the normal 100 millisecond level (experience value), but the sync time increased to the second level. Through the analysis of the indicators, it is known that this part of the time-consuming is caused by the creation of the journal. Then, it is found that there is a forceWrite call in the journal creation process. It is well known that forceWrite can cause write blocking.
If I understand correctly, the forceWrite call is not necessary if syncData is turned off. If this call is removed, it is expected to optimize the long tail delay caused by journal creation blocking.

The following are the changes before and after removing forceWrite. It can be seen that the journal creation time is maintained at a relatively low level
BEFORE
image

AFTER
image

### Changes

Do not call forceWrite if syncData is disabled during Journal Creation

Master Issue: #3564

> ---
>
> ---
> quickly and easily:
>
>
>
> Otherwise:
> 
> 
> ---

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/JournalChannel.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/JournalChannel.java
@@ -259,7 +259,9 @@ class JournalChannel implements Closeable {
         fc.write(bb);
 
         bc = bcBuilder.create(fc, writeBufferSize);
-        forceWrite(true);
+        if (configuration.getJournalSyncData()) {
+            forceWrite(true);
+        }
         nextPrealloc = this.preAllocSize;
         fc.write(zeros, nextPrealloc - journalAlignSize);
     }


zhaohaidao:
rerun failure checks

hangc0276 (changes requested):
I'm not sure whether the journal file metadata can be written into the file if we skip the `forceWrite(true)` when new journal file created. Would you please add a test to cover it?

hangc0276:
@zhaohaidao Would you please take a look at the above comments? thanks a lot.

zhaohaidao:
Ok, I will find time to deal with this PR in the near future, thank you for reminding
@hangc0276

Issue #3567, opened 2022-10-24
Rocksdb instance throws Checksum Mismatch exception frequently with high workload in directio mode


zhaohaidao:
**BUG REPORT**

***Describe the bug***

Rocksdb instance throws Checksum Mismatch exception frequently with high workload in directio mode

***To Reproduce***

1. Run with high workload with openmessaging-benchmark(2000MB/s)

- 3 bookies
- 3 benchmark-worker
- 3 replica
- 3 writeQuorumSize
- 2 ackQuorumSize

2. See error

> Oct 20 18:34:46 test_host pulsar[6380]: Caused by: org.rocksdb.RocksDBException: block checksum mismatch: stored = 4048457838, computed = 3129443982, type = 2  in /cache2/storage/5/current/locations/000037.sst offset 0 size 29026
Oct 20 18:34:46 test_host pulsar[6380]: Caused by: org.rocksdb.RocksDBException: block checksum mismatch: stored = 4048457838, computed = 3129443982, type = 2  in /cache2/storage/5/current/locations/000037.sst offset 0 size 29026
Oct 20 18:34:47 test_host pulsar[6380]: Caused by: org.rocksdb.RocksDBException: block checksum mismatch: stored = 4048457838, computed = 3129443982, type = 2  in /cache2/storage/5/current/locations/000037.sst offset 0 size 29026
Oct 20 18:34:47 test_host pulsar[6380]: Caused by: org.rocksdb.RocksDBException: block checksum mismatch: stored = 4048457838, computed = 3129443982, type = 2  in /cache2/storage/5/current/locations/000037.sst offset 0 size 29026
Oct 20 18:34:48 test_host pulsar[6380]: Caused by: org.rocksdb.RocksDBException: block checksum mismatch: stored = 4048457838, computed = 3129443982, type = 2  in /cache2/storage/5/current/locations/000037.sst offset 0 size 29026
Oct 20 18:34:48 test_host pulsar[6380]: Caused by: org.rocksdb.RocksDBException: block checksum mismatch: stored = 4048457838, computed = 3129443982, type = 2  in /cache2/storage/5/current/locations/000037.sst offset 0 size 29026
Oct 20 18:34:49 test_host pulsar[6380]: Caused by: org.rocksdb.RocksDBException: block checksum mismatch: stored = 4048457838, computed = 3129443982, type = 2  in /cache2/storage/5/current/locations/000037.sst offset 0 size 29026
[...]

PR #3568, opened 2022-10-24
Issue 3567: Upgrade rocksdb version to avoid checksum mismatch error

zhaohaidao:
### Motivation
Rocksdb instance throws Checksum Mismatch exception frequently in high workload(2000MB/s with 3 bookies and 3 replica) with directio mode.
I checked the release history of rocksdb and found that checksum mismatch bug maybe fixed after 7.4.0. The fix is followed
* Fixed a bug in `WriteBatchInternal::Append()` where WAL termination point in write batch was not considered and the function appends an incorrect number of checksums.

### Changes

Upgrade rocksdb version to 7.4.5 to avoid checksum mismatch error
After rocksdb upgraded to 7.4.5. Checksum mismatch error no longer appears

Master Issue: #3567

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -165,7 +165,7 @@
     <protoc3.version>3.19.2</protoc3.version>
     <protoc-gen-grpc-java.version>${grpc.version}</protoc-gen-grpc-java.version>
     <reflections.version>0.9.11</reflections.version>
-    <rocksdb.version>6.29.4.1</rocksdb.version>
+    <rocksdb.version>7.4.5</rocksdb.version>
     <shrinkwrap.version>3.0.1</shrinkwrap.version>
     <slf4j.version>1.7.32</slf4j.version>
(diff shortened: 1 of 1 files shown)


hangc0276 (changes requested):
It upgrades the major version of the RocksDB(7.x), and I'm not sure whether it can compatible with the current RocksDB version (6.x). Does RocksDB 6.x contains this fix?

hangc0276:
@zhaohaidao Would you please start a discussion in the dev mail list for upgrading RocksDB to 7.x?

zhaohaidao:
> @zhaohaidao Would you please start a discussion in the dev mail list for upgrading RocksDB to 7.x?

Ok, I will take the time to launch the discussion in the near future, thank you for reminding
@hangc0276

hangc0276:
The RocksDB version has been upgraded to 7.9.2, close this PR.

PR #3570, opened 2022-10-24
Flush time started moved to after lock

zhaohaidao:
### Motivation

In BP-44(Ledger storage metrics enhancement), ledger flush start time is moved to after lock to avoid 200% time
  utilization calculations.
However, 200% time utilization still happens in my test. After reading the source code, I found that this part was rolled back by #3160. 
If I understand correctly, this change is not as expected

### Changes

Flush start time is moved to after lock again.


wenbingshen (approved):

nicoloboschi (approved):
+1

zhaohaidao:
rerun failure checks

zhaohaidao:
rerun failure checks

PR #3575, opened 2022-10-25
Single thread executor should not have max capacity by default

merlimat:
### Motivation

The OrderedExecutor queue must be really unbounded or we are subject to deadlocks when a large number of tasks are submitted and may be depending on tasks that are going to be executed on the same executor.

Renamed the `GrowableArrayBlockingQueue` into `GrowableMpScArrayConsumerBlockingQueue` and optimized for single consumer scenario. 

Added test to ensure the `SingleThreadExecutor` is really unbound.

Added simple benchmark test scenario that uses 2 queues and 2 threads passing a bounded number of messages between each others, as fast as possible:

https://github.com/apache/bookkeeper/compare/master...merlimat:bookkeeper:fix-ordered-executor-max-size?expand=1#diff-e8f29aa89f5bf7b81642d51409ab37e220c9a066c009f24b82bafbfb8bc3e4e6R248
```
ArrayBlockingQueue: Throughput 27.433703 Millions items/s
LinkedBlockingQueue: Throughput 17.744839 Millions items/s
GrowableMpScArrayConsumerBlockingQueue: Throughput 61.425191 Millions items/s
```


hangc0276 (approved):
Great job!

codelipenghui (approved):

PR #2958, opened 2021-12-23
zk client config update and bugfix for ZKMetadataClientDriver

StevenLuMT:
Descriptions of the changes in this PR:

### Motivation

1. bug fix for error config for BoundExponentialBackoffRetryPolicy in class ZKMetadataClientDriver,if set MaxRetries zero, zk client will throw ConnectionLossException when the zk has some changing,for example: zk leader node changed.

2. In Bookie's ZKClient, different BoundExponentialBackoffRetryPolicy set different MaxRetries,so change zkRetryBackoffMaxRetries to config in Bookie's AbstractConfiguration

### Changes

1. update a error config for BoundExponentialBackoffRetryPolicy in class ZKMetadataClientDriver
2. change zkRetryBackoffMaxRetries to config

Master Issue: #2760

Issue #2760, opened 2021-08-09
Bookkeeper shutdown when zookeeper connection just timeout


suiyuzeng:
***Describe the bug***
Bookkeeper begin to shutdown after a zookeeper connection timeout. Session is not expired and it was attempting reconnect.
In the log, there is a ConnectionLossException when a node was deleted in org.apache.bookkeeper.meta.ZkLedgerUnderreplicationManager#releaseUnderreplicatedLedger. It can not be processed in org.apache.bookkeeper.replication.ReplicationWorker#deferLedgerLockRelease and shutdown was called.
![image](https://user-images.githubusercontent.com/25764673/128686126-0788296d-331f-46d9-b35e-96c8c0110ed0.png)
It shoul be retried as syncCallWithRetries was called. One cause is that the maxRetries of the retry policy was setted to be 0. The zk client is initialized in org.apache.bookkeeper.meta.zk.ZKMetadataDriverBase#initialize. And the retrying policy is created in org.apache.bookkeeper.meta.zk.ZKMetadataClientDriver#initialize.
![image](https://user-images.githubusercontent.com/25764673/128686264-5bb34f7d-390e-44e8-863f-28dea71a9409.png)
The maxRetries is setted 0. Can we set it to be Integer.MAX_VALUE? 

***To Reproduce***

1. Run bookkeeper
2. release ledger lock when zk timeout

LOG:
06:18:45.732 [main-SendThread(10.168.114.200:2181)] WARN  org.apache.zookeeper.ClientCnxn - Session 0x37b0a37639d001d for sever xxxx/x.x.x.x:2181, Closing socket connection. Attempting reconnect except it is a SessionExpiredException.
org.apache.zookeeper.ClientCnxn$EndOfStreamException: Unable to read additional data from server sessionid 0x37b0a37639d001d, likely server has closed socket
    at org.apache.zookeeper.ClientCnxnSocketNIO.doIO(ClientCnxnSocketNIO.java:77) ~[org.apache.zookeeper-zookeeper-3.6.3.jar:3.6.3]
    at org.apache.zookeeper.ClientCnxnSocketNIO.doTransport(ClientCnxnSocketNIO.java:350) ~[org.apache.zookeeper-zookeeper-3.6.3.jar:3.6.3]
    at org.apache.zookeeper.ClientCnxn$SendThread.run(ClientCnxn.java:1290) ~[org.apache.zookeeper-zookeeper-3.6.3.jar:3.6.3]

[...]

PR #3578, opened 2022-10-25
zk client config update and bugfix for ZKMetadataClientDriver (#2958)

Shawyeok:
### Motivation

1. bug fix for error config for BoundExponentialBackoffRetryPolicy in class ZKMetadataClientDriver,if set MaxRetries zero, zk client will throw ConnectionLossException when the zk has some changing,for example: zk leader node changed.

2. In Bookie's ZKClient, different BoundExponentialBackoffRetryPolicy set different MaxRetries,so change zkRetryBackoffMaxRetries to config in Bookie's AbstractConfiguration

### Changes

1. update a error config for BoundExponentialBackoffRetryPolicy in class ZKMetadataClientDriver
2. change zkRetryBackoffMaxRetries to config

Master Issue: #2760

(cherry picked from commit 84ddc9116b36221c92df89bc2f55eeff5c0442c5)


hangc0276 (changes requested):
We will cherry-pick that Pr into branch-4.14 directly, and do not need a separate PR.

PR #3581, opened 2022-10-26
Replace sleep with await to avoid flaky test in SlowBookieTest

hangc0276:
### Motivation
There is a lot of `Thread.sleep()` in Bookie tests, which may lead to flaky tests.

### Changes
Replace sleep with await to avoid flaky test in SlowBookieTest

### DIFF
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/SlowBookieTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/SlowBookieTest.java
@@ -421,7 +421,7 @@ public class SlowBookieTest extends BookKeeperClusterTestCase {
             // waitForWritable async
             new Thread(() -> isWriteable.set(lh.waitForWritable(writeSet, 0, timeout))).start();
 
-            Awaitility.await().untilAsserted(() -> assertFalse(isWriteable.get()));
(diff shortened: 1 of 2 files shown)


bookkeeper-server/src/test/java/org/apache/bookkeeper/client/SlowBookieTest.java
```diff
@@ -419,11 +419,9 @@ public void testWaitForWritable() throws Exception {
...
 
             // waitForWritable async
-           new Thread(() -> {
-                isWriteable.set(lh.waitForWritable(writeSet, 0, timeout));
-            }).start();
-            TimeUnit.MILLISECONDS.sleep(5000);
-            assertFalse(isWriteable.get());
+            new Thread(() -> isWriteable.set(lh.waitForWritable(writeSet, 0, timeout))).start();
+
+            Awaitility.await().untilAsserted(() -> assertFalse(isWriteable.get()));
```
zymap, line 424:
```suggestion
            Awaitility.await().pollDelay(5, TimeUnit.SECONDS).untilAsserted(() -> assertFalse(isWriteable.get()));
```

Because the isWritable is false from the beginning, we want to wait a moment to verify it still in `waitForWritable` backoff

zymap (approved):

PR #3584, opened 2022-10-27
[improve] Optimize log4j configuration

Shawyeok:
### Motivation

Log output with [location information is very expensive][1], we should avoid use it in our log configuration. In most cases, we are able to find which line print specific log by logger name and log content.

### Changes

- remove log line information
- use logger name instead of caller class name

[1]: https://logging.apache.org/log4j/2.x/manual/layouts.html#LocationInformation

### DIFF
--- a/conf/log4j2.xml
+++ b/conf/log4j2.xml
@@ -28,13 +28,13 @@
     </Properties>
     <Appenders>
         <Console name="CONSOLE" target="SYSTEM_OUT">
-            <PatternLayout pattern="%d{ISO8601} - %-5p - [%t:%C{1}@%L] - %m%n"/>
+            <PatternLayout pattern="%d{ISO8601} - %-5p - [%t:%c{1}] - %m%n"/>
         </Console>
         <File name="TRACEFILE" fileName="${sys:bookkeeper.log.dir}/bookkeeper-trace.log">
-            <PatternLayout pattern="%d{ISO8601} - %-5p [%t:%C{1}@%L][%ndc] - %m%n"/>
+            <PatternLayout pattern="%d{ISO8601} - %-5p [%t:%c{1}][%ndc] - %m%n"/>
         </File>
         <RollingFile name="ROLLINGFILE" fileName="${sys:bookkeeper.log.dir}/${sys:bookkeeper.log.file}" filePattern="${sys:bookkeeper.log.dir}/${sys:bookkeeper.log.file}%d{.yyyy-MM-dd}">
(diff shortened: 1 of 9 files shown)


Shawyeok:
If this pr merged, I would like to port to other release branches.

conf/log4j2.xml
```diff
@@ -28,13 +28,13 @@
...
-            <PatternLayout pattern="%d{ISO8601} - %-5p - [%t:%C{1}@%L] - %m%n"/>
+            <PatternLayout pattern="%d{ISO8601} - %-5p - [%t:%c{1}] - %m%n"/>
         </Console>
         <File name="TRACEFILE" fileName="${sys:bookkeeper.log.dir}/bookkeeper-trace.log">
-            <PatternLayout pattern="%d{ISO8601} - %-5p [%t:%C{1}@%L][%ndc] - %m%n"/>
+            <PatternLayout pattern="%d{ISO8601} - %-5p [%t:%c{1}][%ndc] - %m%n"/>
         </File>
         <RollingFile name="ROLLINGFILE" fileName="${sys:bookkeeper.log.dir}/${sys:bookkeeper.log.file}" filePattern="${sys:bookkeeper.log.dir}/${sys:bookkeeper.log.file}%d{.yyyy-MM-dd}">
-            <PatternLayout pattern="%d{ISO8601} - %-5p [%t:%C{1}@%L] - %m%n"/>
+            <PatternLayout pattern="%d{ISO8601} - %-5p [%t:%c{1}] - %m%n"/>
```
hangc0276, line 37: Can we include the full package name?

hangc0276:
ping @Shawyeok, would you please take a look at the comments? thanks a lot.

hangc0276:
@horizonzy Would you please help take a look at this PR? Thanks a lot.

conf/log4j2.xml
Shawyeok: `%d{ISO8601} [%t] %-5level %logger - %m%n%ex`

@hangc0276  How about this?
- %-5p and %-5level is identical but later is more readable
- %logger is identical with %c, logger name with full package name, pay a little more disk space
- the `-` after timestamp is redundant
- %ex is to prevent default %xEx, which is to add packaging information about each stack trace element, that means more cost. see more: `org.apache.logging.log4j.core.pattern.ExtendedThrowablePatternConverter`

horizonzy: Without `%L`, we can't identify the place where the log is from. 
In some case, we will record the same info in the same class, without the line number, we can't trace the problem precisely according to the log info.
If we want to remove `%L`, we should ensure can't print same log content in the same class in different place

PR #3585, opened 2022-10-28
Monitor timeout tasks in executor

gaozhangmin:
### Motivation

Monitor slow tasks, add metric `thread_executor_tasks_timeout ` in  `SingleThreadExecutor`

### DIFF
--- a/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/SingleThreadExecutor.java
+++ b/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/SingleThreadExecutor.java
@@ -19,6 +19,7 @@
 package org.apache.bookkeeper.common.util;
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.AbstractExecutorService;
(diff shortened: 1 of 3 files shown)


bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/SingleThreadExecutor.java
```diff
@@ -80,11 +92,14 @@ public SingleThreadExecutor(ThreadFactory tf, int maxQueueCapacity, boolean reje
         } else {
             this.queue = new GrowableMpScArrayConsumerBlockingQueue<>();
         }
+        this.executor = Executors.newSingleThreadScheduledExecutor(new DefaultThreadFactory(EXECUTOR_NAME));
```
merlimat, line 95: This would be doubling the number of threads

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/SingleThreadExecutor.java
```diff
@@ -225,6 +246,26 @@ public void execute(Runnable r) {
         }
     }
 
+    private void monitorTimeoutTasks() {
+        for (Runnable task : queue) {
```
merlimat, line 250: The `queue` is meant to be used by a single consumer, so we cannot iterate over it from a different thread.

merlimat (changes requested):

hangc0276 (changes requested):
Why not check when the task actually runs?

PR #3593, opened 2022-10-29
Removed the necessity of SafeRunnable

merlimat:
### Motivation

`OrderedExecutor` is taking `SafeRunnable` tasks instead of the standard `Runnable`. 

The reason was the Java `ThreadPoolExecutor` are failing when an exception is thrown by one of the tasks and there is no log printed of the exception. 

Right now, this is not needed anymore, because the `SingleThreadExecutor` is already doing try/catch/log: https://github.com/apache/bookkeeper/blob/8ac28db16118e5415c041f8da8cca73777971ce6/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/SingleThreadExecutor.java#L135 

For OrderedScheduler, we can instead wrap the tasks when they are submitted, instead of relying on caller to wrap them.

There are a few reasons not to use `SafeRunnable`: 
 1. It requires 1 extra allocation for each task
 2. It makes the API non-standard, pushing the need to always manually wrap the regular `Runnable`
 3. Even when using it in lambda-form, it still requires extra levels of indentation that make the code more difficult to read.

### DIFF
--- a/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/OrderedExecutor.java
+++ b/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/OrderedExecutor.java
@@ -49,13 +49,7 @@ import org.apache.commons.lang.StringUtils;
 import org.slf4j.MDC;
 
 /**
- * This class provides 2 things over the java {@link ExecutorService}.
- *
- * <p>1. It takes {@link SafeRunnable objects} instead of plain Runnable objects.
- * This means that exceptions in scheduled tasks wont go unnoticed and will be
- * logged.
- *
- * <p>2. It supports submitting tasks with an ordering key, so that tasks submitted
+ * This class supports submitting tasks with an ordering key, so that tasks submitted
  * with the same key will always be executed in order, but tasks across
  * different keys can be unordered. This retains parallelism while retaining the
  * basic amount of ordering we want (e.g. , per ledger handle). Ordering is
@@ -526,7 +520,7 @@ public class OrderedExecutor implements ExecutorService {
      * @param orderingKey
      * @param r
      */
-    public void executeOrdered(Object orderingKey, SafeRunnable r) {
+    public void executeOrdered(Object orderingKey, Runnable r) {
         chooseThread(orderingKey).execute(r);
     }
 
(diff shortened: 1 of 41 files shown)


bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/OrderedExecutor.java
```diff
@@ -526,7 +520,7 @@ public boolean preserveMdc() {
      * @param orderingKey
      * @param r
      */
-    public void executeOrdered(Object orderingKey, SafeRunnable r) {
+    public void executeOrdered(Object orderingKey, Runnable r) {
```
eolivelli, line 523: I am pretty sure that this API is uses by other projects as utility.

Could we still keep the old methods in order to keep compatibility?

eolivelli (changes requested):
Overall LGTM

But I left one comments about backwards compatibility. Please take a look.
Unfortunately BookKeeper consumers are very 'creative' and they tend to use many internal classes of BK (see Pulsar and probably Protocol Handlers)

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/OrderedExecutor.java
merlimat: `SafeRunnable` interface is not being removed and changing these methods to require `Runnable` should be ok since it's the "parent" interface.

merlimat: eg: passing a `SafeRunnable` instance is still valid, because it implements `Runnable`.

hangc0276 (commented):
Overall LGTM, would you please take a look at the failed tests? @merlimat

zymap (commented):
LGTM

hangc0276 (approved):

hangc0276:
@eolivelli Would you please take a look? thanks.

dlg99 (approved):

eolivelli (approved):
Let's try.

The applications will have to recompiled and probably that will be a issue for Pulsar Protocol handlers and other Pulsar extensions.
This is because in Java the compiled call site refers to the actual signature of the method and not handle subclasses/polymorphism

PR #3595, opened 2022-10-31
Allow to use IO uring instead of epoll

merlimat:
### Motivation

Allow users to enable IO_uring for BK network transport.

Since the Netty implementation is still considered experimental, it will not be enabled by default, rather only when the `-Denable.io_uring=true` is passed.

### DIFF
--- a/bookkeeper-common/pom.xml
+++ b/bookkeeper-common/pom.xml
@@ -60,6 +60,18 @@
       <artifactId>jctools-core</artifactId>
     </dependency>
     <dependency>
+      <groupId>io.netty.incubator</groupId>
+      <artifactId>netty-incubator-transport-native-io_uring</artifactId>
+      <version>${netty-iouring.version}</version>
+      <classifier>linux-x86_64</classifier>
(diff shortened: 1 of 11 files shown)


codelipenghui (approved):

bookkeeper-common/pom.xml
```diff
@@ -59,6 +59,18 @@
       <groupId>org.jctools</groupId>
       <artifactId>jctools-core</artifactId>
     </dependency>
+    <dependency>
+      <groupId>io.netty.incubator</groupId>
+      <artifactId>netty-incubator-transport-native-io_uring</artifactId>
+      <version>${netty-iouring.version}</version>
```
hangc0276, line 65: remove the version?

hangc0276 (approved):
Good job!

hangc0276:
@merlimat Would you please take a look at the failed tests?

Shoothzj (approved):

zymap (approved):

PR #3598, opened 2022-11-02
Add ByteBuf refCnt test for AddEntry and use pooled ByteBuf reduce heap usage

wenbingshen:
### Motivation

1. Add ByteBuf refCnt test for `addEntry` and `recoveryAddEntry`
2. use pooled ByteBuf reduce heap usage

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
@@ -931,6 +931,18 @@ public class BookieImpl extends BookieCriticalThread implements Bookie {
         return journals.get(MathUtils.signSafeMod(ledgerId, journals.size()));
     }
 
+    @VisibleForTesting
+    public ByteBuf createMasterKeyEntry(long ledgerId, byte[] masterKey) {
+        // new handle, we should add the key to journal ensure we can rebuild
+        ByteBuffer bb = ByteBuffer.allocate(8 + 8 + 4 + masterKey.length);
+        bb.putLong(ledgerId);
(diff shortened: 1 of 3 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
```diff
@@ -931,6 +931,18 @@ private Journal getJournal(long ledgerId) {
         return journals.get(MathUtils.signSafeMod(ledgerId, journals.size()));
     }
 
+    @VisibleForTesting
+    public ByteBuf createMasterKeyEntry(long ledgerId, byte[] masterKey) {
+        // new handle, we should add the key to journal ensure we can rebuild
+        ByteBuffer bb = ByteBuffer.allocate(8 + 8 + 4 + masterKey.length);
```
merlimat, line 937: If we’re returning a ByteBuf and it’s properly ref counted, then we should just use a pooled ByteBuf instance instead of wrapping a ByteBuffer

merlimat (commented):
Change looks good. I don’t think there’s any memory leak though, because the buffer is just an unspooled buffer from heap

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
wenbingshen: @merlimat Great understanding! Thanks for your review and suggestion. I have used a pooled ByteBuf instead of wrapping a ByteBuffer. :)

wenbingshen:
@StevenLuMT also help take a look. Thanks.

PR #3324, opened 2022-06-09
Apply the backpressure changes on the V2 requests

zymap:
---

*Motivation*

If one bookie is slow (not down, just slow), the BK client
will the acks to the user that the entries are written after
the first 2 acks. In the meantime, it will keep waiting for
the 3rd bookie to respond. If the bookie responds within the
timeout, the entries can now be dropped from memory, otherwise
the write will timeout internally and it will get replayed
to a new bookie.

In the V3 request, we have [server-side backpressure](https://github.com/apache/bookkeeper/pull/1410)
to impact the client-side behaviors. We should apply the same
changes to the V2 request. That would help this [issue](https://github.com/apache/pulsar/issues/14861)
to be resolved.

*Modification*

- Apply the change https://github.com/apache/bookkeeper/pull/1410 to V2 protocol

Descriptions of the changes in this PR:

PR #3599, opened 2022-11-02
Add a judgement of waitTimeoutOnResponseBackpressureMs when throttling read response

gaozhangmin:
#3324 made v2 request support back pressure,  if we set `waitTimeoutOnResponseBackpressureMs`, it won't take effect, because  readWorkerThreadsThrottlingEnabled is enabled by default,


gaozhangmin:
rerun failure checks

zymap (changes requested):
Because it is different logic for them. If you want to use the backpressure, you should know the throttle need to be disabled.

gaozhangmin:
> Because it is different logic for them but they are trying to resolve the same problem. If you want to use the backpressure, you should know the throttle need to be disabled

For most who don't be familiar with the code, they won't know back pressure  need disabling throttle

PR #3600, opened 2022-11-02
waitTimeoutOnResponseBackpressureMs should be gt 0

gaozhangmin:
update `timeOut >= 0` to `timeout > 0`,  it make no sense executing back pressure when  `waitTimeoutOnBackpressureMillis=0`

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PacketProcessorBase.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PacketProcessorBase.java
@@ -85,7 +85,7 @@ abstract class PacketProcessorBase<T extends Request> extends SafeRunnable {
         final long writeNanos = MathUtils.nowInNano();
 
         final long timeOut = requestProcessor.getWaitTimeoutOnBackpressureMillis();
-        if (timeOut >= 0 && !channel.isWritable()) {
+        if (timeOut > 0 && !channel.isWritable()) {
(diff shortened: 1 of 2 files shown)


hangc0276 (changes requested):
0 to allow request to fail immediately

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PacketProcessorBase.java
```diff
@@ -85,7 +85,7 @@ protected void sendResponse(int rc, Object response, OpStatsLogger statsLogger)
         final long writeNanos = MathUtils.nowInNano();
 
         final long timeOut = requestProcessor.getWaitTimeoutOnBackpressureMillis();
-        if (timeOut >= 0 && !channel.isWritable()) {
+        if (timeOut > 0 && !channel.isWritable()) {
```
zymap, line 88: timeout is 0 should fail immediately, it is expected.

zymap (changes requested):

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PacketProcessorBase.java
gaozhangmin: Is it ok when timeout is 0 and blacklistedChannels = Optional.empty() ?

PR #3603, opened 2022-11-04
BookieImpl remove wait until journal quits

wenbingshen:
### Motivation

PR #2887 introduced the feature of shutdown the bookie service after any Journal thread exits, so we don't need to wait in BookieImpl for all Journal threads to exit before shutdown the Bookie service, because this cannot happen.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
@@ -810,30 +810,9 @@ public class BookieImpl extends BookieCriticalThread implements Bookie {
 
     @Override
     public void run() {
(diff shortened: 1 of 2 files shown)


wenbingshen:
We have BookieMultipleJournalsTest#testJournalExit, this test shows that any journal thread exit will cause bookie to exit. So we no longer need BookieImpl to shutdown the bookie service after waiting for all journal threads to exit.

eolivelli (approved):
Good catch

LGTM

wenbingshen:
ping @hangc0276 @dlg99 @zymap PTAL, Thanks.

hangc0276 (approved):

wenbingshen:
ping @zymap @hangc0276 @Shoothzj @dlg99 If you have time, can you help take a look. Thanks.

Issue #3604, opened 2022-11-04
the concurrency problem of the compaction process and the regular write process


1559924775:
**BUG REPORT**

***Describe the bug***

we use SortedLedgerStorage+EntryLogManagerForSingleEntryLog. 
When the entrylog file managed by EntryLogManagerForSingleEntryLog reaches the upper limit, the createNewLog method is executed. This operation is not locked, which will conflict with the addEntry executed in the compaction process, resulting in disordered entryLog file data.
![14e56f8a3e057508f446abb1ccd7b687](https://user-images.githubusercontent.com/35036009/199918343-ce3f5648-6851-4d82-b093-e0605ea9a02a.png)
![ae141da9c0fa6d51373775c19e0860ad](https://user-images.githubusercontent.com/35036009/199918357-d0124ed3-d7d7-4563-bef1-64ab62d553dc.png)
There is a concurrency problem with the two operations shown above.
In our case we found this code execution order:
1.logChannel.flush();
2.logChannel.write(sizeBuffer);
3.logChannel.appendLedgersMap();
4.logChannel.write(entry);

The correct entryLog file format should be like this：
header
entrySize1+entryData1
entrySize2+entryData2
...
entrySizen+entryDatan
ledgersMap

The wrong entrylog is as follows:
header
entrySize1+entryData1
entrySize2+entryData2
...
entrySizen
ledgersMap
entryDatan

@hangc0276

PR #3605, opened 2022-11-04
Fix the concurrency problem of the compaction process and the regular…

1559924775:
… write process

add a synchronized createNewLog method in EntryLogManagerForSingleEntryLog.

### Motivation

we use SortedLedgerStorage+EntryLogManagerForSingleEntryLog. 
When the entrylog file managed by EntryLogManagerForSingleEntryLog reaches the upper limit, the createNewLog method is executed. This operation is not locked, which will conflict with the addEntry executed in the compaction process, resulting in disordered entryLog file data.
14e56f8a3e057508f446abb1ccd7b687
ae141da9c0fa6d51373775c19e0860ad
There is a concurrency problem with the two operations shown above.
In our case we found this code execution order:
1.logChannel.flush();
2.logChannel.write(sizeBuffer);
3.logChannel.appendLedgersMap();
4.logChannel.write(entry);

The correct entryLog file format should be like this：
header
entrySize1+entryData1
entrySize2+entryData2
...
entrySizen+entryDatan
ledgersMap

The wrong entrylog is as follows:
header
entrySize1+entryData1
entrySize2+entryData2
...
entrySizen
ledgersMap
entryDatan

### Changes

add a synchronized createNewLog method in EntryLogManagerForSingleEntryLog.

Master Issue: #3604

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogManagerForSingleEntryLog.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogManagerForSingleEntryLog.java
@@ -119,6 +119,11 @@ class EntryLogManagerForSingleEntryLog extends EntryLogManagerBase {
     }
 
     @Override
+    synchronized void createNewLog(long ledgerId, String reason) throws IOException {
+        super.createNewLog(ledgerId, reason);
+    }
+
+    @Override
     public synchronized void setCurrentLogForLedgerAndAddToRotate(long ledgerId, BufferedLogChannel logChannel) {
         BufferedLogChannel hasToRotateLogChannel = activeLogChannel;
         activeLogChannel = logChannel;


wenbingshen (changes requested):
@1559924775 I'm not sure if I understand it wrong, please let me know if I'm wrong.

In fact, the implementation of entryLogManager used by `DefaultEntryLogger` is `EntryLogManagerForSingleEntryLog`. In Compactor,  `entryLogger.addEntry` actually executes `EntryLogManagerForSingleEntryLog.addEntry` , and this method is locked.
https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogManagerForSingleEntryLog.java#L86
```java
        if (entryLogPerLedgerEnabled) {
            this.entryLogManager = new EntryLogManagerForEntryLogPerLedger(conf, ledgerDirsManager,
                    entryLoggerAllocator, listeners, recentlyCreatedEntryLogsStatus, statsLogger);
        } else {
            this.entryLogManager = new EntryLogManagerForSingleEntryLog(conf, ledgerDirsManager, entryLoggerAllocator,
                    listeners, recentlyCreatedEntryLogsStatus);
        }
```
Similarly, `createNewLog` has also been locked:
https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogManagerForSingleEntryLog.java#L117
```java
    @Override
    synchronized void createNewLog(long ledgerId) throws IOException {
        super.createNewLog(ledgerId);
    }
```

wenbingshen:
@hangc0276 PTAL. Thanks.

wenbingshen (approved):
Sorry, I seem to be mistaken, this is the different overloading method of `createNewLog`.

Shoothzj (approved):

StevenLuMT (changes requested):
I don't think it's nice to lock it
The createNewLog should be a separate process and should eliminate the point of thread insecurity

hangc0276 (changes requested):
Would you please add a test to cover this case?

1559924775:
> I don't think it's nice to lock the method, the createNewLog should be a separate process and should eliminate the point of thread insecurity

Do you have any concerns about this modification? Could you be more specific? @StevenLuMT

wenbingshen (changes requested):
I re-understood this problem. getCurrentLogForLedgerForAddEntry is protected by synchronized. During createNewLog, the compaction process addEntry should not be able to get activeLogChannel until createNewLog is completed. The problem here should be caused by other race conditions.

Even if the compaction process addEntry get the activeLogChannel, and then other threads get synchronized on createNewLog, it is useless.

1559924775:
> getCurrentLogForLedgerForAddEntry

image
The createNewLog does not lock the activeLogChannel

1559924775:
> getCurrentLogForLedgerForAddEntry

image
The createNewLog does not lock the activeLogChannel

hangc0276:
@zymap @horizonzy Please help take a look at this PR, thanks a lot.

horizonzy (approved):
LGTM.

horizonzy (commented):
In `EntryLogManagerForEntryLogPerLedger`, it looks like it has the same problem.

horizonzy (approved):
LGTM

horizonzy:
Could you add a test to cover the function? Thanks.

Issue #3607, opened 2022-11-04
Error in parsing entrylog file causes GC failure


1559924775:
**BUG REPORT**

***Describe the bug***

When bookie GC encounters an entryLog file with disordered data, a IllegalArgumentException are not caught, resulting in the death of the GC thread and the end of the GC process.
This problem is encountered again in the next GC, resulting in the subsequent entrylog being unable to GC.
The problem described in this issue(#3604) will cause the entryLog file to be disordered, and will cause the IllegalArgumentException when parsing the entrylog file.

![image](https://user-images.githubusercontent.com/35036009/199923547-367a1c0c-8eac-426d-ad7d-50769849efce.png)
In our case, due to the disorder of the entrylog file, the exception in the above figure appeared. This exception is not a subclass of IOException and cannot be caught.
@hangc0276

shoothzj:
This only affects branch-4.14. master change is #3263

PR #3609, opened 2022-11-04
Fix the problem that the abnormal file causes the bookie GC to fail

1559924775:

### Motivation
When bookie GC encounters an entryLog file with disordered data, all exceptions are not caught, resulting in the death of the GC thread and the end of the GC process.

image
In our case, due to the disorder of the entrylog file, the exception in the above figure appeared. This exception is not a subclass of IOException and cannot be caught.

This issue (#3604) describes a scenario that causes the entry log file to be disordered.

Master Issue: #3607

@hangc0276

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
@@ -1012,6 +1012,12 @@ public class EntryLogger {
             while (true) {
                 // Check if we've finished reading the entry log file.
                 if (pos >= bc.size()) {
+                    LOG.warn("read position illegal, ledgersMap maybe error. position {} entrylog size {} entrylog {}",
+                            pos, bc.size(), entryLogId);
+                    break;
+                }
+                if (pos < 0) {
+                    LOG.warn("read position illegal, position is negative. position {} entrylog {}", pos, entryLogId);
                     break;
                 }
                 if (readFromLogChannel(entryLogId, bc, headerBuffer, pos) != headerBuffer.capacity()) {


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -1012,6 +1012,12 @@ public void scanEntryLog(long entryLogId, EntryLogScanner scanner) throws IOExce
             while (true) {
                 // Check if we've finished reading the entry log file.
                 if (pos >= bc.size()) {
+                    LOG.warn("read position illegal, ledgersMap maybe error. position {} entrylog size {} entrylog {}",
+                            pos, bc.size(), entryLogId);
+                    break;
+                }
+                if (pos < 0) {
```
hangc0276, line 1019: I'm curious about how `pos < 0` happens

eolivelli: good catch, it looks like it is not possible

dlg99: at least throw in this case?

Shoothzj:
Can the entry log still be readable?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
1559924775: This issue describes a scenario that causes pos to be negative.（#3604）

1559924775:
> Can the entry log still be readable?

Unreadable. The existence of multiple copies avoids subscription exceptions

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
1559924775: > I'm curious about how `pos < 0` happens
This issue describes a scenario where pos is negative due to the disorder of entrylog files

1559924775: > good catch, it looks like it is not possible

This issue describes a scenario where pos is negative due to the disorder of entrylog files

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -1012,6 +1012,12 @@ public void scanEntryLog(long entryLogId, EntryLogScanner scanner) throws IOExce
             while (true) {
                 // Check if we've finished reading the entry log file.
                 if (pos >= bc.size()) {
+                    LOG.warn("read position illegal, ledgersMap maybe error. position {} entrylog size {} entrylog {}",
+                            pos, bc.size(), entryLogId);
+                    break;
+                }
```
wenbingshen, line 1018: `pos >= bc.size` used checking if we've finished reading the entry log file, why need add warn log? Does this mean the warn log is always there?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
wenbingshen: > I'm curious about how `pos < 0` happens

+1

wenbingshen (changes requested):

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
hangc0276: From the code logic, the `pos` won't be less than 0, and I'm not sure how `pos < 0` happens

1559924775: sorry，This problem exists in version 4.14, but not in the latest version.
image

PR #3611, opened 2022-11-05
Fix the problem that the abnormal file causes the bookie GC fail

1559924775:

Motivation
When bookie GC encounters an entryLog file with disordered data, a IllegalArgumentException are not caught, resulting in the death of the GC thread and the end of the GC process.
The problem described in this issue(https://github.com/apache/bookkeeper/issues/3604) will cause the entryLog file to be disordered, and will cause the IllegalArgumentException when parsing the entrylog file.

Master Issue:
https://github.com/apache/bookkeeper/issues/3607 : Describes the problem to be solved by this pr.
https://github.com/apache/bookkeeper/issues/3604 : Describes the phenomenon and causes of disordered entrylog files.


hangc0276 (approved):

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -1000,6 +1000,11 @@ public void scanEntryLog(long entryLogId, EntryLogScanner scanner) throws IOExce
                 long offset = pos;
                 pos += 4;
                 int entrySize = headerBuffer.readInt();
+                if (entrySize <= 0) { // hitting padding
```
zymap, line 1003: Is it possible the entrySize is normal but the ledgerId is abnormal?

1559924775: Skip and continue to read the next entry.
See "scanner.accept(ledgerId)" in line 1011.
The code is consistent with version 4.16.

zymap (approved):

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
horizonzy: I think it can't resolve the disorder problem. After this pr, could you parse the entry file incorrectly?

PR #3612, opened 2022-11-05
[branch-4.15] Fix the problem that the abnormal file causes the bookie GC fail

1559924775:

Motivation
When bookie GC encounters an entryLog file with disordered data, a IllegalArgumentException are not caught, resulting in the death of the GC thread and the end of the GC process.
The problem described in this issue(https://github.com/apache/bookkeeper/issues/3604) will cause the entryLog file to be disordered, and will cause the IllegalArgumentException when parsing the entrylog file.

Master Issue:
https://github.com/apache/bookkeeper/issues/3607 : Describes the problem to be solved by this pr.
https://github.com/apache/bookkeeper/issues/3604 : Describes the phenomenon and causes of disordered entrylog files.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
@@ -1021,6 +1021,11 @@ public class EntryLogger {
                 long offset = pos;
                 pos += 4;
                 int entrySize = headerBuffer.readInt();
+                if (entrySize <= 0) {
+                    LOG.warn("bad read for ledger entry from entryLog {}@{} (entry size {})",
+                            entryLogId, pos, entrySize);
+                    return;
+                }
                 long ledgerId = headerBuffer.readLong();
                 headerBuffer.clear();
 
@@ -1031,11 +1036,6 @@ public class EntryLogger {
                 }
                 // read the entry
                 data.clear();
-                if (entrySize <= 0) {
-                    LOG.warn("bad read for ledger entry from entryLog {}@{} (entry size {})",
-                            entryLogId, pos, entrySize);
-                    return;
-                }
                 data.capacity(entrySize);
                 int rc = readFromLogChannel(entryLogId, bc, data, pos);
                 if (rc != entrySize) {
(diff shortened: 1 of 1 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -1021,6 +1021,11 @@ public void scanEntryLog(long entryLogId, EntryLogScanner scanner) throws IOExce
                 long offset = pos;
                 pos += 4;
                 int entrySize = headerBuffer.readInt();
+                if (entrySize <= 0) {
```
hangc0276, line 1024: Why return directly？

hangc0276 (changes requested):

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
1559924775: I have modified it.

> Why return directly？

hangc0276:
@horizonzy Please help take a look at this Pr, thanks a lot.

horizonzy (approved):
~LGTM.  This problem has already been fixed in the master, but it didn't fix in branch-4.15.
See: https://github.com/apache/bookkeeper/pull/3263/files#diff-efe10967f0d6718093d4f23406e3435c294d0d57ca8461afcda348288f7b193b~

horizonzy:
> LGTM. This problem has already been fixed in the master, but it didn't fix in branch-4.15.

Sorry, it's not the same case. I will test it and reply it again later.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -1021,6 +1021,11 @@ public void scanEntryLog(long entryLogId, EntryLogScanner scanner) throws IOExce
                 long offset = pos;
                 pos += 4;
                 int entrySize = headerBuffer.readInt();
+                if (entrySize <= 0) { // hitting padding
```
horizonzy, line 1024: It can't resolve the problem. It can't skip the ledgerMap.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLogger.java
```diff
@@ -1021,6 +1021,11 @@ public void scanEntryLog(long entryLogId, EntryLogScanner scanner) throws IOExce
                 long offset = pos;
                 pos += 4;
```
horizonzy, line 1022: `pos += 4` should be after line_1030

Issue #3618, opened 2022-11-07
[Bug] NPE throwed when recovering ledger: Cannot read field "distributionSchedule" because "this.lh" is null


zhaohaidao:
**BUG REPORT**
NPE throwed when recovering ledger: Cannot read field "distributionSchedule" because "this.lh" is null
***Describe the bug***

During LedgerRecovery, the safeRun method of AddPendingOp will be executed. When handling failed bookies, bk client will try to replace failed bookies with healthy ones. However, if a NotEnoughBookieException occurred when replacing bookies , the corresponding AddPendingOp instance will be recycled. Then bad things happened as the following logic includes field call on null object.
<img width="1051" alt="image" src="https://user-images.githubusercontent.com/6216286/200281434-117fae3a-5c15-47e5-85f5-45f6e11dbb90.png">

***To Reproduce***

The reproduce details can be seen in PendingAddOpTest#testReadOnlyLedgerHandleWithNotEnoughBookiesExceptionDuringRecoveryAdd
1. Create a ReadOnlyLedgerHandle instance and a PendingAddOp instance
2. Make sure  NotEnoughBookiesException throwed during replace bookie process
3. Execute the safeRun method of the PendingAddOp instance

***Expected behavior***
NPE throwed and the error log is as followed, just the same as the issue title desrcibes

> java.lang.NullPointerException: Cannot read field "distributionSchedule" because "this.lh" is null

***Screenshots***
<img width="1393" alt="image" src="https://user-images.githubusercontent.com/6216286/200280874-e36f3f2a-5dd7-4a8a-8120-1c15130b0318.png">

***Additional context***

PR #3620, opened 2022-11-07
Fix NPE issue during recovery add

zhaohaidao:
Fix NPE issue during recovery add

### Motivation
The issue details can is https://github.com/apache/bookkeeper/issues/3618

### Changes
Check if the op has been recycled after handling bookie failure
image

Master Issue: #3618

> ---
>
> ---
> quickly and easily:
>
>
>
> Otherwise:
> 
> 
> ---

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
@@ -1769,7 +1769,7 @@ public class LedgerHandle implements WriteHandle {
         if (getLedgerMetadata().getState() == LedgerMetadata.State.IN_RECOVERY) {
             // we should not close ledger if ledger is recovery mode
             // otherwise we may lose entry.
-            executeOrdered(() -> errorOutPendingAdds(rc));
+            errorOutPendingAdds(rc);
             return;
         }
         LOG.error("Closing ledger {} due to {}", ledgerId, BKException.codeLogger(rc));
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
@@ -1769,7 +1769,7 @@ public class LedgerHandle implements WriteHandle {
         if (getLedgerMetadata().getState() == LedgerMetadata.State.IN_RECOVERY) {
             // we should not close ledger if ledger is recovery mode
             // otherwise we may lose entry.
-            errorOutPendingAdds(rc);
+            executeOrdered(() -> errorOutPendingAdds(rc));
             return;
         }
         LOG.error("Closing ledger {} due to {}", ledgerId, BKException.codeLogger(rc));
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
@@ -266,10 +266,6 @@ class PendingAddOp implements Runnable, WriteCallback {
         // We are about to send. Check if we need to make an ensemble change
         // because of delayed write errors
         lh.maybeHandleDelayedWriteBookieFailure();
-        // This op maybe recycled during bookie failures
-        if (maybeRecycled()) {
-            return;
-        }
 
         // Iterate over set and trigger the sendWriteRequests
         DistributionSchedule.WriteSet writeSet = lh.distributionSchedule.getWriteSet(entryId);
@@ -496,10 +492,6 @@ class PendingAddOp implements Runnable, WriteCallback {
         if (hasRun && toSend == null && pendingWriteRequests == 0) {
             recyclePendAddOpObject();
         }
-    }
-
-    boolean maybeRecycled() {
-        return lh == null;
     }
 
     public void recyclePendAddOpObject() {
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
@@ -266,6 +266,10 @@ class PendingAddOp implements Runnable, WriteCallback {
         // We are about to send. Check if we need to make an ensemble change
         // because of delayed write errors
         lh.maybeHandleDelayedWriteBookieFailure();
+        // This op maybe recycled during bookie failures
+        if (maybeRecycled()) {
+            return;
+        }
 
         // Iterate over set and trigger the sendWriteRequests
         DistributionSchedule.WriteSet writeSet = lh.distributionSchedule.getWriteSet(entryId);
@@ -492,6 +496,10 @@ class PendingAddOp implements Runnable, WriteCallback {
         if (hasRun && toSend == null && pendingWriteRequests == 0) {
             recyclePendAddOpObject();
         }
+    }
+
+    boolean maybeRecycled() {
+        return lh == null;
     }
 
     public void recyclePendAddOpObject() {
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/PendingAddOpTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/PendingAddOpTest.java
@@ -104,34 +104,6 @@ public class PendingAddOpTest {
     }
 
     @Test
-    public void testLedgerHandleWithDeferredSyncDuringRecoveryAdd() throws Exception {
-        final BookieId b1 = new BookieSocketAddress("b1", 3181).toBookieId();
-        final BookieId b2 = new BookieSocketAddress("b2", 3181).toBookieId();
-        final BookieId b3 = new BookieSocketAddress("b3", 3181).toBookieId();
-        MockClientContext clientCtx = MockClientContext.create();
-        Versioned<LedgerMetadata> md = ClientUtil.setupLedger(clientCtx, 0,
-                LedgerMetadataBuilder.create().withInRecoveryState().newEnsembleEntry(0L,
-                        Lists.newArrayList(b1, b2, b3)));
-        LedgerHandle lh = new LedgerHandle(clientCtx, 0, md, BookKeeper.DigestType.CRC32C,
-                ClientUtil.PASSWD, EnumSet.of(WriteFlag.DEFERRED_SYNC));
-        lh.notifyWriteFailed(0, b1);
-        AtomicInteger rcHolder = new AtomicInteger(-0xdead);
-        PendingAddOp op = PendingAddOp.create(
-                lh, mockClientContext, lh.getCurrentEnsemble(),
-                payload, EnumSet.of(WriteFlag.DEFERRED_SYNC),
-                (rc, handle, entryId, qwcLatency, ctx) -> {
-                    rcHolder.set(rc);
-                }, null).enableRecoveryAdd();
-        assertSame(lh, op.lh);
-        op.setEntryId(0);
-        lh.pendingAddOps.add(op);
-        lh.clientCtx.getMainWorkerPool().submitOrdered(lh.ledgerId, (Callable<Void>) () -> {
-            op.run();
-            return null;
-        }).get();
-    }
-
-    @Test
     public void testReadOnlyLedgerHandleWithNotEnoughBookiesExceptionDuringRecoveryAdd() throws Exception {
         final BookieId b1 = new BookieSocketAddress("b1", 3181).toBookieId();
         final BookieId b2 = new BookieSocketAddress("b2", 3181).toBookieId();
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/PendingAddOpTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/PendingAddOpTest.java
@@ -123,7 +123,6 @@ public class PendingAddOpTest {
                     rcHolder.set(rc);
                 }, null).enableRecoveryAdd();
         assertSame(lh, op.lh);
-        lh.pendingAddOps.add(op);
         op.setEntryId(0);
         lh.pendingAddOps.add(op);
         lh.clientCtx.getMainWorkerPool().submitOrdered(lh.ledgerId, (Callable<Void>) () -> {
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/PendingAddOpTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/PendingAddOpTest.java
@@ -22,24 +22,36 @@ import static java.nio.charset.StandardCharsets.UTF_8;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertSame;
+import static org.junit.Assert.assertTrue;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
+import com.google.common.collect.Lists;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
+
+import java.util.EnumSet;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.concurrent.atomic.AtomicInteger;
 import org.apache.bookkeeper.client.BKException.Code;
+import org.apache.bookkeeper.client.api.LedgerMetadata;
 import org.apache.bookkeeper.client.api.WriteFlag;
 import org.apache.bookkeeper.common.util.OrderedExecutor;
+import org.apache.bookkeeper.conf.ClientConfiguration;
+import org.apache.bookkeeper.discover.MockRegistrationClient;
+import org.apache.bookkeeper.net.BookieId;
+import org.apache.bookkeeper.net.BookieSocketAddress;
 import org.apache.bookkeeper.proto.BookieClient;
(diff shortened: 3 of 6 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
```diff
@@ -266,6 +266,10 @@ public void run() {
         // We are about to send. Check if we need to make an ensemble change
         // because of delayed write errors
         lh.maybeHandleDelayedWriteBookieFailure();
+        // This op maybe recycled during bookie failures
+        if (maybeRecycled()) {
```
dlg99, line 270: So maybeHandleDelayedWriteBookieFailure triggers something that recycles this pending add op (I haven't looked that closely)?

Then the fix masks the problem. 
One side of the problem is that lh becomes null / results in NPE later. 
Another side of the problem is that pending add opp can be recycled and immediately reused by another add.

hangc0276 (changes requested):
@zhaohaidao  Thanks for your contribution.

The root cause of this NPE is the different implementations of `handleBookieFailure` in `LedgerHandle` and `ReadOnlyLedgerHandle`.

In the `LedgerHandle` implementation, it handles ensemble change results in a new task with the same thread
https://github.com/apache/bookkeeper/blob/43bdfb35317d9a69840d17df94c5c801a4cbba62/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java#L1962-L2004

In the `ReadOnlyLedgerHandle`, it handles the ensemble change results and exceptions in the current task and current thread, which will lead to the NPE.
https://github.com/apache/bookkeeper/blob/43bdfb35317d9a69840d17df94c5c801a4cbba62/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadOnlyLedgerHandle.java#L232

I suggest keeping the same with `LedgerHandle` implementation to change `handleUnrecoverableErrorDuringAdd(e.getCode());` to `executeOrdered(() -> handleUnrecoverableErrorDuringAdd(e.getCode()));`

zhaohaidao:
> 

Thanks for your advice. I don't quite understand why processing in a different task and current thread would not be a problem. Could you please explain this in more detail?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java
zhaohaidao: > 

That's right. maybeHandleDelayedWriteBookieFailure will replace failed bookie with healthy one. However if NotEnoughBookieException is throwed. maybeHandleDelayedWriteBookieFailure will recycle this pending add op.
You can see this recycle logic in followed code logic
`

    @Override
    void handleBookieFailure(final Map<Integer, BookieId> failedBookies) {
        // handleBookieFailure should always run in the ordered executor thread for this
        // ledger, so this synchronized should be unnecessary, but putting it here now
        // just in case (can be removed when we validate threads)
        synchronized (metadataLock) {
            ...
            try {
                List<BookieId> newEnsemble = EnsembleUtils.replaceBookiesInEnsemble(
                        clientCtx.getBookieWatcher(), metadata, currentEnsemble, failedBookies, logContext);
            ...
            } catch (BKException.BKNotEnoughBookiesException e) {
                LOG.error("Could not get additional bookie to remake ensemble, closing ledger: {}", ledgerId);

                handleUnrecoverableErrorDuringAdd(e.getCode());
                return;
            }
`

zhaohaidao: > So maybeHandleDelayedWriteBookieFailure triggers something that recycles this pending add op (I haven't looked that closely)?
> 
> Then the fix masks the problem. One side of the problem is that lh becomes null / results in NPE later. Another side of the problem is that pending add opp can be recycled and immediately reused by another add.

- One side of the problem is that lh becomes null / results in NPE later: 
If I understand correctly, only maybeHandleDelayedWriteBookieFailure has a chance to recycle this operation causing lh to become null which leads to this npe issue. Although this behavior is not as described in the followed comments about recycle condition, I think it is reasonable. 
```
    private void maybeRecycle() {
        /**
         * We have opportunity to recycle two objects here.
         * PendingAddOp#toSend and LedgerHandle#pendingAddOp
         *
         * A. LedgerHandle#pendingAddOp: This can be released after all 3 conditions are met.
         *    - After the callback is run
         *    - After safeRun finished by the executor
         *    - Write responses are returned from all bookies
         *      as BookieClient Holds a reference from the point the addEntry requests are sent.
         *
         * B. ByteBuf can be released after 2 of the conditions are met.
         *    - After the callback is run as we will not retry the write after callback
         *    - After safeRun finished by the executor
         * BookieClient takes and releases on this buffer immediately after sending the data.
         *
         * The object can only be recycled after the above conditions are met
         * otherwise we could end up recycling twice and all
         * joy that goes along with that.
         */
        if (hasRun && callbackTriggered) {
            ReferenceCountUtil.release(toSend);
            toSend = null;
        }
        // only recycle a pending add op after it has been run.
        if (hasRun && toSend == null && pendingWriteRequests == 0) {
            recyclePendAddOpObject();
        }
    }

```

- Another side of the problem is that pending add opp can be recycled and immediately reused by another add.: I don't understand this very well, can you explain in more detail?

dlg99: >> Another side of the problem is that pending add opp can be recycled and immediately reused by another add.

> I don't understand this very well, can you explain in more detail?

```java
        lh.maybeHandleDelayedWriteBookieFailure();
        
        // Add Op is recycled at this point, maybeRecycled() has a chance to check lh == null
        // but the recycler can give that op to another thread/request (via RECYCLER.get() call)
        // so it can be re-initialized with different lh and other data
        // so maybeRecycled returns false and the rest of the method is executed against different op
        // to avoid such problems and avoid locks we use ordered executors / guarantee order of calls
        
        if (maybeRecycled()) {
```

@zhaohaidao I hope this helps

zhaohaidao: > > > Another side of the problem is that pending add opp can be recycled and immediately reused by another add.
> 
> > I don't understand this very well, can you explain in more detail?
> 
> ```java
>         lh.maybeHandleDelayedWriteBookieFailure();
>         
>         // Add Op is recycled at this point, maybeRecycled() has a chance to check lh == null
>         // but the recycler can give that op to another thread/request (via RECYCLER.get() call)
>         // so it can be re-initialized with different lh and other data
>         // so maybeRecycled returns false and the rest of the method is executed against different op
>         // to avoid such problems and avoid locks we use ordered executors / guarantee order of calls
>         
>         if (maybeRecycled()) {
> ```
> 
> @zhaohaidao I hope this helps

Thanks a lot for your explanation. I submitted a revision based on your suggestions.
Please have a look if you have time.
@dlg99

zhaohaidao:
> @zhaohaidao Thanks for your contribution.
> 
> The root cause of this NPE is the different implementations of `handleBookieFailure` in `LedgerHandle` and `ReadOnlyLedgerHandle`.
> 
> In the `LedgerHandle` implementation, it handles ensemble change results in a new task with the same thread
> 
> https://github.com/apache/bookkeeper/blob/43bdfb35317d9a69840d17df94c5c801a4cbba62/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java#L1962-L2004
> 
> In the `ReadOnlyLedgerHandle`, it handles the ensemble change results and exceptions in the current task and current thread, which will lead to the NPE.
> 
> https://github.com/apache/bookkeeper/blob/43bdfb35317d9a69840d17df94c5c801a4cbba62/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadOnlyLedgerHandle.java#L232
> 
> I suggest keeping the same with `LedgerHandle` implementation to change `handleUnrecoverableErrorDuringAdd(e.getCode());` to `executeOrdered(() -> handleUnrecoverableErrorDuringAdd(e.getCode()));`

Thanks a lot for your explanation. I submitted a revision based on your suggestions.
Please have a look if you have time.
@hangc0276

hangc0276:
@merlimat @eolivelli @dlg99 @zymap @horizonzy  Please help take a look at this Pr, thanks a lot.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/PendingAddOpTest.java
```diff
@@ -87,4 +103,70 @@ public void testExecuteAfterCancelled() {
...
+        lh.notifyWriteFailed(0, b1);
+        AtomicInteger rcHolder = new AtomicInteger(-0xdead);
+        PendingAddOp op = PendingAddOp.create(
+                lh, mockClientContext, lh.getCurrentEnsemble(),
+                payload, EnumSet.of(WriteFlag.DEFERRED_SYNC),
+                (rc, handle, entryId, qwcLatency, ctx) -> {
+                    rcHolder.set(rc);
+                }, null).enableRecoveryAdd();
+        assertSame(lh, op.lh);
+        lh.pendingAddOps.add(op);
```
horizonzy, line 126: Why add same `PendingAddOp` twice to the `pendingAddOps queue`.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/PendingAddOpTest.java
```diff
@@ -87,4 +103,70 @@ public void testExecuteAfterCancelled() {
...
+                new DefaultBookieAddressResolver(regClient),
+                NullStatsLogger.INSTANCE);
+        ClientContext clientCtx = MockClientContext.create(mockBookies, conf, regClient, placementPolicy,
+                bookieWatcher);
+        Versioned<LedgerMetadata> md = ClientUtil.setupLedger(clientCtx, 0,
+                LedgerMetadataBuilder.create().withInRecoveryState().newEnsembleEntry(0L,
+                        Lists.newArrayList(b1, b2, b3)));
+        ReadOnlyLedgerHandle lh = new ReadOnlyLedgerHandle(clientCtx, 0, md, BookKeeper.DigestType.CRC32C,
+                ClientUtil.PASSWD, true);
+        lh.notifyWriteFailed(0, b1);
```
horizonzy, line 155: In the normal case, it can't happen.  
`lh.notifyWriteFailed(0, b1);` should not before `op.run()`, it just for test I guess.
I'm curious about why the problem happened. Could you add a test to mock the real case? thanks a lot.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/PendingAddOpTest.java
zhaohaidao: Nice catch. This is a redundant one

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/PendingAddOpTest.java
zhaohaidao: I try to give a scenario where the problem might occur.
The client tries to write 945 and 946 to the same ledger (id=1256) successively, and writes 3 copies of each entry. Write 945 is successful but one of the bookies returns an exception, so LedgerHandle.delayedWriteFailedBookies will not be empty. Then when writing 946, if a BKNotEnoughBookiesException is thrown, an NPE issue will occur.

It is a little troublesome to construct the above scenario, so I ensure that delayedWriteFailedBookies is not empty (by lh.notifyWriteFailed(0, b1)) from the beginning of my test case, so that I can quickly verify the npe issue.

It should be noted that this is a scenario I deduced from the error log. In the log, only 946 write failures caused by BKNotEnoughBookiesException can be seen.

horizonzy (commented):
~LGTM. 
I have a topic to discuss: When `errorOutPendingAdds`, it will drain all the PendingAddOps(including the current PendingAddOp). Would you like us to make the current PendingAddOp didn't send any request to the bookie server? ~

horizonzy (commented):
Sorry, this change may introduce a new problem.
There is a case:
The ledger handle asyncAddEntry 10000 times. So there are 10000 PendingAddOp will be sent to the LedgerHandle#pendingAddOps.

The 10000 PendingAddOp will be executed one by one in the future. 
Before this change, `errorOutPendingAdds` is a sync operation, it takes all the PendingAddOp from LedgerHandle#pendingAddOps, then submits callback. After submitting callback, all the 
PendingAddOp#callbackTriggered will be set to `true`. 
Because we execute every PendingAddOp in the thread pool, every PendingAddOp will be run at some time.

https://github.com/apache/bookkeeper/blob/7b5b6b240ca7551da266903078f2dd2ba2906b96/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java#L248-L254
At now, we check PendingAddOp#callbackTriggered value is true or not, if true, it will return directly, didn't send any request to the server; it's quick. 
But if we make `errorOutPendingAdds` async, all the 10000 PendingAddOp will send requests to the server, because the Runnable is ordered. After 10000 PendingAddOp executing, the `errorOutPendingAdds` will be executed. It's not acceptable.

zhaohaidao:
> Sorry, this change may introduce a new problem. There is a case: The ledger handles asyncAddEntry 10000 times. So there are 10000 PendingAddOp will be sent to the LedgerHandle#pendingAddOps.
> 
> The 10000 PendingAddOp will be executed one by one in the future. Before this change, `errorOutPendingAdds` is a sync operation; it takes all the PendingAddOp from LedgerHandle#pendingAddOps, then submits a callback. After submitting callback, all the PendingAddOp#callbackTriggered will be set to `true`. Because we execute every PendingAddOp in the thread pool, every PendingAddOp will be run at some time.
> 
> https://github.com/apache/bookkeeper/blob/7b5b6b240ca7551da266903078f2dd2ba2906b96/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingAddOp.java#L248-L254
> 
> 
> At now, we check PendingAddOp#callbackTriggered value is true or not; if true, it will return directly, didn't send any request to the server; it's quick.
> But if we make `errorOutPendingAdds` async, all the 10000 PendingAddOp will send requests to the server, because the Runnable is ordered. After 10000 PendingAddOp executing, the `errorOutPendingAdds` will be executed. It's not acceptable.

@horizonzy Thank you very much for your careful explanation, your concern is very reasonable. However in the scenario you described, when bkclient writes multiple messages to the same ledger. If the first message fails, LedgerHandle makes subsequent requests fail by calling errorOutPendingAdds asynchronously.
There is a difference between the implementation of ReadOnlyLedgeHandle and the implementation of LedgerHandle. The latter is a sync operation.
The change of this pr is actually referring to the implementation of handleUnrecoverableErrorDuringAdd of LedgerHandle, according to the suggestion of @hangc0276 .
Another fix I can think of is to interrupt the execution of safeRun when NotEnoughException is encountered, but this may require adding an exception signature to handleBookieFailure. I understand that this may violate the original design intention? Do you have any better suggestions for changes? Welcome to continue the discussion.

It is worth mentioning that although LedgerHandle.handleUnrecoverableErrorDuringAdd may also call errorOutPendingAdds synchronously, the logic of this code will never happen, because the ledger recovery process will definitely call the ReadOnlyLedgeHandle method.

https://github.com/apache/bookkeeper/blob/7b5b6b240ca7551da266903078f2dd2ba2906b96/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java#L1767-L1777

horizonzy:
There are some places to invoke `handleUnrecoverableErrorDuringAdd`, the LedgerHandle will also invoke `handleUnrecoverableErrorDuringAdd`. This pr also change the behavior in LedgerHandle.


image

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -1769,7 +1769,7 @@ void handleUnrecoverableErrorDuringAdd(int rc) {
         if (getLedgerMetadata().getState() == LedgerMetadata.State.IN_RECOVERY) {
             // we should not close ledger if ledger is recovery mode
             // otherwise we may lose entry.
-            errorOutPendingAdds(rc);
+            executeOrdered(() -> errorOutPendingAdds(rc));
```
zhaohaidao, line 1772: @horizonzy  I may not have explained it clearly. Although this is changed to an async operation, this code block will never be executed, because when in IN_RECOVERY mode, the method of ReadOnlyLedgerHandle must be called.
What do you think if I roll back the changes to LedgerHandle and only keep the changes to ReadOnlyLedgerHandle

horizonzy: I see.

horizonzy: I see.

zhaohaidao:
rerun failure checks

zhaohaidao:
rerun failure checks

zhaohaidao:
rerun failure checks

zhaohaidao:
rerun failure checks

dlg99 (approved):
LGTM.
Thank you for fixing this and adding the test.

zhaohaidao:
Hey, @hangc0276 Can you help me continue to review this pr when you are free? Also, the ci seems to be stuck and cannot be re-triggered.

zhaohaidao:
rerun failure checks

zhaohaidao:
rerun failure checks

zhaohaidao:
rerun failure checks

zhaohaidao:
rerun failure checks

PR #3622, opened 2022-11-07
Fix GetBookieInfo failed event stats

wenbingshen:
### Motivation

In `GetBookieInfoProcessorV3`, even if throw `IOException` and get DiskSpace failed, it will still `registerSuccessfulEvent`, which is wrong, we should `registerFailedEvent`.

### Changes

1. registerFailedEvent when get DiskSpace failed.
2. add unit test.


wenbingshen:
rerun failure checks

hangc0276 (approved):

zymap (approved):

dlg99 (approved):

PR #3623, opened 2022-11-08
[log] Add journal file path that caused failure in multi-journal config

rdhabalia:
### Motivation

Right now, if bookie fails to replay the journal then it logs journal-txn-id but it doesn't log journal dir path. It will be useful if bookie prints journal path along with journal-dir in case if user is using multi-journal directories so, user can troubleshoot the issue and fix it correctly.

```
2022-11-07T18:29:07,205+0000 [main] ERROR org.apache.bookkeeper.bookie.Bookie - Exception while replaying journals, shutting down
java.io.IOException: Recovery log 1667603421612 is missing
        at org.apache.bookkeeper.bookie.Bookie.replay(Bookie.java:932) ~[org.apache.bookkeeper-bookkeeper-server-4.12.jar:4.12]
        at org.apache.bookkeeper.bookie.Bookie.readJournal(Bookie.java:911) ~[org.apache.bookkeeper-bookkeeper-server-4.12.jar]
```

### Changes

Add journal dir file path in journal replay failure.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
@@ -617,7 +617,8 @@ public class BookieImpl extends BookieCriticalThread implements Bookie {
         // validate filtered log ids only when we have markedLogId
         if (markedLog.getLogFileId() > 0) {
             if (logs.size() == 0 || logs.get(0) != markedLog.getLogFileId()) {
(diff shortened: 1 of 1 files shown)


hangc0276 (approved):

merlimat (approved):

wenbingshen (approved):

dlg99 (approved):

hangc0276:
This Pr can not be cherry-picked to branch-4.14, and I will push another PR to branch-4.14

PR #3630, opened 2022-11-10
Add Http-service to check bookie sanity state

rdhabalia:
### Motivation

Right now, if Bookie startup or process fails due to various reasons then bookie getState() rest API still gives a successful response even though bookie writes are not going through. Therefore. Bookie should have a REST API to check bookie sanity test and that can be used to check bookie liveliness in k8/helm-chart or to plot bookie health graph.

### Changes

Add Bookie REST API to check bookie health to validate bookie liveliness

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/server/http/service/BookieSanityService.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/server/http/service/BookieSanityService.java
@@ -1,0 +1,90 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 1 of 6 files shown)


rdhabalia:
good point. made a change to run max (1) concurrent requests.

rdhabalia:
@eolivelli can you PTAL.

bookkeeper-server/src/main/java/org/apache/bookkeeper/server/http/service/BookieSanityService.java
```diff
@@ -0,0 +1,100 @@
...
+		}
+
+		BookieSanity bs = new BookieSanity();
+		if (config.isForceReadOnlyBookie()) {
+			bs.readOnly = true;
+		} else {
+		    try {
+                // allow max concurrent request as sanity-test check relatively
+                // longer time to complete
+		        lock.acquire();
```
eolivelli, line 87: Please wait at most for X seconds (5?) and if the semaphore cannot be acquired fail the request. We don't want many requests piling up.


Please also fix formatting

eolivelli (approved):
I have left one minor issue please check my comment.
Then the patch is good to go from my POV.
And we should release this with 4.16.

@hangc0276 FYI

hangc0276:
> I have left one minor issue please check my comment. Then the patch is good to go from my POV. And we should release this with 4.16.
> 
> @hangc0276 FYI

@eolivelli Sure

hangc0276 (approved):
Overall LGTM.

bookkeeper-server/src/main/java/org/apache/bookkeeper/server/http/service/BookieSanityService.java
rdhabalia: done

PR #3631, opened 2022-11-10
Add small files check in garbage collection

hangc0276:
### Motivation
When we use `TransactionalEntryLogCompactor` to compact the entry log files, it will generate a lot of small entry log files, and for those files, the file usage is usually greater than 90%, which can not be compacted unless the file usage decreased.

image

### Changes
We introduce the entry log file size check during compaction, and the checker is controlled by `gcEntryLogSizeRatio`. 
If the total entry log file size is less than `gcEntryLogSizeRatio * logSizeLimit`, the entry log file will be compacted even though the file usage is greater than 90%. This feature is disabled by default and the `gcEntryLogSizeRatio` default value is `0.0`

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
@@ -589,7 +589,7 @@ public class GarbageCollectorThread implements Runnable {
 
         final int maxBucket = calculateUsageIndex(numBuckets, threshold);
         stopCompaction:
-        for (int currBucket = 0; currBucket < maxBucket; currBucket++) {
+        for (int currBucket = 0; currBucket <= maxBucket; currBucket++) {
             LinkedList<Long> entryLogIds = compactableBuckets.get(currBucket);
             while (!entryLogIds.isEmpty()) {
                 if (timeDiff.getValue() < maxTimeMillis) {
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
@@ -546,24 +546,14 @@ public class GarbageCollectorThread implements Runnable {
     @VisibleForTesting
     void doCompactEntryLogs(double threshold, long maxTimeMillis) throws EntryLogMetadataMapException {
         LOG.info("Do compaction to compact those files lower than {}", threshold);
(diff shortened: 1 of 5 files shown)


zymap:
Open an email to discuss it?

hangc0276:
> Open an email to discuss it?

@zymap Ok, I send an email.

hangc0276:
ping @merlimat @eolivelli @dlg99 @zymap Please help take a look at this PR, thanks.

eolivelli (approved):
great

+1

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -548,14 +548,24 @@ private boolean removeIfLedgerNotExists(EntryLogMetadata meta) throws EntryLogMe
...
+                + "on garbage collection.", conf.getGcEntryLogSizeRatio(), gcEntryLogSizeRatio,
+                gcEntryLogSizeRatio * conf.getEntryLogSizeLimit());
+        }
 
         final int numBuckets = 10;
         int[] entryLogUsageBuckets = new int[numBuckets];
         int[] compactedBuckets = new int[numBuckets];
 
         ArrayList<LinkedList<Long>> compactableBuckets = new ArrayList<>(numBuckets);
+        ArrayList<LinkedList<Long>> smallFilesCompactableBuckets = new ArrayList<>(numBuckets);
```
dlg99, line 565: I think you can avoid addition of another type of buckets.
Have something like useTargetEntryLogSizeForGc config;

Replace
```java
calculateUsageIndex(numBuckets, meta.getUsage());
```

with something like
```java
double usage = meta.getUsage();

if (conf.getUseTargetEntryLogSizeForGc()) {
  usage = (double) meta.getRemainingSize() / Math.max(meta.getTotalSize(), conf.getEntryLogSizeLimit());
}
```
and the rest will happen naturally.

As an improvement, transactional compaction can fallback to a regular one if the estimated size of entry log after compaction is below some limit.

hangc0276: @dlg99 Good suggestion. I updated the code with `useTargetEntryLogSizeForGc` flag.  Please help take a look, thanks.

> As an improvement, transactional compaction can fallback to a regular one if the estimated size of entry log after compaction is below some limit.

For this one, I will open a new PR to handle it.

zymap (approved):

horizonzy (approved):
LGTM.

hangc0276:
This PR is hard to be cherry-picked to branch-4.14, and I will push another PR for branch-4.14

zymap:
Cherry-pick this change without the tests class. The test class is based on the direct IO improvement.

PR #3632, opened 2022-11-12
Include bkperf into bk all package

hangc0276:
### Motivation
The `bkperf` dependency doesn't include in the bookkeeper-all package, and when we run `bin/bkperf` command, it will throw the following exceptions.
```
$ bin/bkperf journal append -j data -n 100000000 --sync true
Couldn't find module '(org.apache.bookkeeper-)?bookkeeper-perf' jar.
Do you want me to run `mvn package -DskipTests` for you ? (y|n)
```

### Changes
Include the `bookkeeper-perf` dependency into `bookkeeper-all` package.

### DIFF
--- a/bookkeeper-dist/all/pom.xml
+++ b/bookkeeper-dist/all/pom.xml
@@ -104,6 +104,12 @@
     </dependency>
 
     <dependency>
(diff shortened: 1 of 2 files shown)


eolivelli (approved):
Lgtm

PR #3643, opened 2022-11-15
Fix wrong key for getHttpServerTrustStorePassword

### LINKED ISSUE #3639: ServerConfiguration returns incorrect property for httpServerTrustStorePassword
**BUG REPORT**

***Describe the bug***

ServerConfiguration.getHttpServerTrustStorePassword() returns value of HTTP_SERVER_KEY_STORE_PASSWORD. Instead it should return the value of HTTP_SERVER_TRUST_STORE_PASSWORD.

***To Reproduce***

1. Set the following properties, specifying different key store and truststore files with different passwords:
httpServerTlsEnable=true
httpServerKeyStorePath
httpServerKeyStorePassword
httpServerTrustStorePath
httpServerTrustStorePassword
2. Start bookkeeper and observe the truststore password is incorrect.

***Expected behavior***

The truststore should be unlocked with its password.

***Screenshots***

N/A

***Additional context***

N/A

wenbingshen:
### Motivation
Fix #3639 

As the title.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
@@ -3663,7 +3663,7 @@ public class ServerConfiguration extends AbstractConfiguration<ServerConfigurati
      * @return http server trustStore password
      */
     public String getHttpServerTrustStorePassword() {
-        return getString(HTTP_SERVER_KEY_STORE_PASSWORD);
+        return getString(HTTP_SERVER_TRUST_STORE_PASSWORD);
     }
 
     /**


Shoothzj (approved):

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
```diff
@@ -3663,7 +3663,7 @@ public ServerConfiguration setHttpServerTrustStorePath(String trustStorePath) {
      * @return http server trustStore password
      */
     public String getHttpServerTrustStorePassword() {
-        return getString(HTTP_SERVER_KEY_STORE_PASSWORD);
+        return getString(HTTP_SERVER_TRUST_STORE_PASSWORD);
```
nicoloboschi, line 3666: should we handle retrocompatibility here? the risk is that the user upgrades and the webserver doesn't work anymore

at least, we should add highlight in the release notes

nicoloboschi (changes requested):

hangc0276 (commented):
I think we should handle the  retro compatibility

hangc0276:
@merlimat @eolivelli @dlg99 @zymap Please help take a look, thanks.

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
nicoloboschi: Please add fallback to the old property. Also a comment in the code pointing to this pull would be valuable

nicoloboschi (changes requested):

wenbingshen:
Maybe we can mark this config as deprecated and add a new config to restore what it actually does and doc it in the docs and release notes.

shoothzj:
@wenbingshen any updates?

shoothzj:
closed by #4301

PR #3651, opened 2022-11-16
Fix JAVA_HOME would be not set

gaozhangmin:
`source ${BK_CONFDIR}/bkenv.sh` should be precede in order to get proper JAVA_HOME variable

### DIFF
--- a/bin/bookkeeper
+++ b/bin/bookkeeper
@@ -23,6 +23,11 @@ set -e
 BINDIR=`dirname "$0"`
 BK_HOME=`cd ${BINDIR}/..;pwd`
 
+if [ -f $BK_HOME/conf/bkenv.sh ]
+then
+ . $BK_HOME/conf/bkenv.sh
+fi
+
 source ${BK_HOME}/bin/common.sh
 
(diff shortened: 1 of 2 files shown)


zymap (changes requested):
As the comment said: `Set JAVA_HOME here to override the environment setting`. It is used to override the default value, so it shouldn't precede.

gaozhangmin:
There would be a case where we don't invoke common.sh or bookkeeper.sh from bookkeeper-daemon script. so the env settings which are set in  bkenv.sh would be useless.

hangc0276:
> There would be a case where we don't invoke common.sh or bookkeeper.sh from bookkeeper-daemon script

> There would be a case where we don't invoke common.sh or bookkeeper.sh from bookkeeper-daemon script

In bookkeeper-daemon.sh, we will call bookkeeper.sh

hangc0276 (changes requested):
> As the comment said: Set JAVA_HOME here to override the environment setting. It is used to override the default value, so it shouldn't precede.

+1

PR #3662, opened 2022-11-22
[fix][server]fix memory leak when operating ledger metadata

HQebupt:
### Motivation
There is a memory leak point in Bookkeeper client, see the following heap report:
image

It happens when opening a non-existent or expired ledger data. The following screen shot is memeory leak point by [Eclipse Memory Analyzer](https://www.eclipse.org/mat/).
image

### Changes
Handle exception and clean up when promise complete exceptionally.

Master Issue: [#1](https://github.com/HQebupt/bookkeeper/pull/1)

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/CleanupLedgerManager.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/CleanupLedgerManager.java
@@ -34,11 +34,14 @@ import org.apache.bookkeeper.proto.BookkeeperInternalCallbacks.Processor;
 import org.apache.bookkeeper.versioning.Version;
 import org.apache.bookkeeper.versioning.Versioned;
 import org.apache.zookeeper.AsyncCallback;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * A ledger manager that cleans up resources upon closing.
  */
 public class CleanupLedgerManager implements LedgerManager {
+    private static final Logger LOG = LoggerFactory.getLogger(CleanupLedgerManager.class);
 
     private class CleanupGenericCallback<T> implements GenericCallback<T> {
 
@@ -111,7 +114,12 @@ public class CleanupLedgerManager implements LedgerManager {
 
     private void recordPromise(CompletableFuture<?> promise) {
         futures.add(promise);
-        promise.thenRun(() -> futures.remove(promise));
+        promise.whenComplete((result, exception) -> {
+            futures.remove(promise);
+            if (exception != null) {
+                LOG.error("Failed on operating ledger metadata: {}", BKException.getExceptionCode(exception));
+            }
+        });
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


HQebupt:
@StevenLuMT @hangc0276  @eolivelli PTAL

hangc0276 (commented):
Nice catch! 
Would you please add a test to protect this logic? thanks.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/CleanupLedgerManager.java
```diff
@@ -34,11 +34,14 @@
...
 import org.apache.bookkeeper.versioning.Versioned;
 import org.apache.zookeeper.AsyncCallback;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * A ledger manager that cleans up resources upon closing.
  */
 public class CleanupLedgerManager implements LedgerManager {
+    private static final Logger LOG = LoggerFactory.getLogger(CleanupLedgerManager.class);
```
mattisonchao, line 44: It looks like this project supports Lombok.  I'm wondering if we can use it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/CleanupLedgerManager.java
```diff
@@ -111,7 +114,12 @@ private GenericCallback removeCallback(GenericCallback callback) {
 
     private void recordPromise(CompletableFuture<?> promise) {
         futures.add(promise);
-        promise.thenRun(() -> futures.remove(promise));
+        promise.whenComplete((result, exception) -> {
+            futures.remove(promise);
+            if (exception != null) {
+                LOG.error("Failed on operating ledger metadata: {}", BKException.getExceptionCode(exception));
```
mattisonchao, line 120: I'm not sure if we need to log the error message.

mattisonchao (approved):

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/CleanupLedgerManager.java
HQebupt: Sure

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/CleanupLedgerManager.java
HQebupt: LGTM

StevenLuMT (approved):
LGTM

dlg99 (approved):

zymap (approved):

horizonzy (approved):
LGTM.

wenbingshen (approved):

HQebupt:
> Nice catch!
> Would you please add a test to protect this logic? thanks.

added a unit test. PTAL @hangc0276

PR #3664, opened 2022-11-22
Dont validate CompactionInterval when compaction is disabled

gaozhangmin:
### Motivation
Currently, We cannot disable minorCompaction and majorCompaction by setting interval to -1, with `gcWaitTime` is  enabled.

### Changes
Don't validate CompactionInterval if compaction is disabled


dlg99 (approved):

horizonzy (approved):
LGTM

zymap (approved):

PR #3672, opened 2022-11-27
[fix][server]fix memory leak when closeRecovered,failed on clearing newEnsemblesFromRecovery

StevenLuMT:
### Motivation

change Future.thenRun to Future.whenComplete in the bookie project there are places that must be released.
1) the first place is updated in  #3662 
2) the second place is this pr: #3672 

the condition that the bug triggers:
1) when future handle exception, thenRun will be ignored to run

### Changes
1) handle exception and clean up when promise complete exceptionally.
2) org.apache.bookkeeper.client.LedgerRecovery2Test#testFirstWriterCannotCommitWriteAfter2ndWriterCloses has cover this pr, so no need to add a new testcase

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadOnlyLedgerHandle.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadOnlyLedgerHandle.java
@@ -331,8 +331,7 @@ class ReadOnlyLedgerHandle extends LedgerHandle implements LedgerMetadataListene
                 newEnsemblesFromRecovery.clear();
             }
             if (exception != null) {
-                LOG.error("When closeRecovered,failed on clearing newEnsemblesFromRecovery: {}",
-                        BKException.getExceptionCode(exception));
+                LOG.error("When closeRecovered,failed on clearing newEnsemblesFromRecovery.", exception);
             }
         });
         return f;
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadOnlyLedgerHandle.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadOnlyLedgerHandle.java
@@ -326,11 +326,15 @@ class ReadOnlyLedgerHandle extends LedgerHandle implements LedgerMetadataListene
                     return builder.withClosedState().withLastEntryId(lac).withLength(len).build();
                 },
                 this::setLedgerMetadata).run();
-        f.thenRun(() -> {
-                synchronized (metadataLock) {
-                    newEnsemblesFromRecovery.clear();
-                }
-            });
+        f.whenComplete((result, exception) -> {
+            synchronized (metadataLock) {
+                newEnsemblesFromRecovery.clear();
+            }
+            if (exception != null) {
+                LOG.error("When closeRecovered,failed on clearing newEnsemblesFromRecovery: {}",
+                        BKException.getExceptionCode(exception));
+            }
+        });
         return f;
     }
 


HQebupt (approved):
Nice catch. 👍

StevenLuMT:
rerun failure checks

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadOnlyLedgerHandle.java
```diff
@@ -326,11 +326,15 @@ CompletableFuture<Versioned<LedgerMetadata>> closeRecovered() {
...
-        f.thenRun(() -> {
-                synchronized (metadataLock) {
-                    newEnsemblesFromRecovery.clear();
-                }
-            });
+        f.whenComplete((result, exception) -> {
+            synchronized (metadataLock) {
+                newEnsemblesFromRecovery.clear();
+            }
+            if (exception != null) {
```
hangc0276, line 333: The exception has been catched in the outside logic, and where is the memory leak?

hangc0276 (changes requested):

zymap (approved):
LGTM

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadOnlyLedgerHandle.java
horizonzy: The promise may be complete with an exception. https://github.com/apache/bookkeeper/blob/01232c94c91759a175345e8c055951e5cc6091dd/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/MetadataUpdateLoop.java#L125-L178

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadOnlyLedgerHandle.java
```diff
@@ -326,11 +326,15 @@ CompletableFuture<Versioned<LedgerMetadata>> closeRecovered() {
                     return builder.withClosedState().withLastEntryId(lac).withLength(len).build();
                 },
                 this::setLedgerMetadata).run();
-        f.thenRun(() -> {
-                synchronized (metadataLock) {
-                    newEnsemblesFromRecovery.clear();
-                }
-            });
+        f.whenComplete((result, exception) -> {
+            synchronized (metadataLock) {
```
horizonzy, line 330: There is one thing I care about:
Shall we clear `newEnsemblesFromRecovery` when the metadata update fails? 
Line_314 support `newEnsemblesFromRecovery` to do some work. If we clear it, is there some influence for it?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadOnlyLedgerHandle.java
StevenLuMT: yes, MetadataUpdateLoop.run may be completed with an exception. @hangc0276

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadOnlyLedgerHandle.java
StevenLuMT: @horizonzy it's ok, 
I don't think the problem you mentioned has anything to do with this pr?
I understand this logic is
when finished to `closeRecovered`, clean up data for `newEnsemblesFromRecovery` at the end

horizonzy: Fine.

horizonzy (approved):
LGTM.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/ReadOnlyLedgerHandle.java
```diff
@@ -326,11 +326,15 @@ CompletableFuture<Versioned<LedgerMetadata>> closeRecovered() {
...
+                newEnsemblesFromRecovery.clear();
+            }
+            if (exception != null) {
+                LOG.error("When closeRecovered,failed on clearing newEnsemblesFromRecovery: {}",
+                        BKException.getExceptionCode(exception));
```
eolivelli, line 335: Get exception code is only a number? Why not dumping the whole exception (with stacktrace)

hangc0276: +1

StevenLuMT: good suggestion, I have updated
@hangc0276 @eolivelli

eolivelli (approved):
+1

PR #3676, opened 2022-11-30
Add an argument to decide expose the default JVM metrics

zymap:
---

###Motivation

Expose the default JVM Metrics or not. If you are using the BookKeeper as an embedded service and you want to expose metrics in your application, you might need to disable this to avoid the JVM metrics register duplicated.

### Modification

- Add a configuration variable to decide whether or not to expose the JVM metrics


horizonzy (approved):
LGTM.

hangc0276 (approved):

PR #3683, opened 2022-12-01
Not wrap IOException twice form checkpoint

wenbingshen:
### Motivation

IOException may be wrapped by RuntimeException, so don't wrap RuntimeException with IOException.


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
```diff
@@ -839,6 +839,10 @@ public void checkpoint(Checkpoint checkpoint) throws IOException {
             throw e;
         } catch (RuntimeException e) {
             recordFailedEvent(dbLedgerStorageStats.getFlushStats(), startTime);
+            // not wrap IOException again
+            if (e.getCause() != null && e.getCause() instanceof IOException) {
```
gaozhangmin, line 843: Does IOException throw at the first catch block?

wenbingshen: This IOException is wrapped by RuntimeException. not a directly IOException. The first catch block can not caught it.

hangc0276: Could you share the detailed IOException wrapped by RuntimeException?

hangc0276 (changes requested):

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
horizonzy: Change the interface will be better:
```
    public interface EntryConsumer {
        void accept(long ledgerId, long entryId, ByteBuf entry);
    }
```

```
    public interface EntryConsumer {
        void accept(long ledgerId, long entryId, ByteBuf entry) throws IOException;
    }
```
So we needn't catch the inner exception. Throw the IOException directly.

hangc0276: +1

horizonzy (approved):
LGTM.

hangc0276:
@wenbingshen Would you please resolve the conflicts? thanks a lot.

hangc0276 (approved):

PR #3688, opened 2022-12-03
[refactor] [server] Refactor ByteBuf release method in module bookkeeper-server

StevenLuMT:
### Motivation

It may throw an exception when release a ByteBuf object. so the exception in ByteBuf.release should be checked.

### Changes
1) Use ReferenceCountUtil.safeRelease() instead of ByteBuf.release() in module bookkeeper-server


HQebupt (approved):

HQebupt:
LGTM

dlg99 (approved):

PR #3689, opened 2022-12-03
[refactor][bookkeeper] Refactor ByteBuf release method in stream/statelib

HQebupt:
### Motivation
It may throw an exception when release  a `ByteBuf` object.  So the exception in `ByteBuf.release` should be checked.

### Changes
Use `ReferenceCountUtil.safeRelease()` instead of `ByteBuf.release()`.

Master Issue: [#7](https://github.com/HQebupt/bookkeeper/pull/7)


StevenLuMT (approved):
LGTM

dlg99 (approved):

PR #3693, opened 2022-12-06
[refactor] [server] Refactor ByteBuf release method in module distributedlog-core/distributedlog-protocol

StevenLuMT:
### Motivation

It may throw an exception when release a ByteBuf object. so the exception in ByteBuf.release should be checked.

### Changes
1) Use ReferenceCountUtil.safeRelease() instead of ByteBuf.release() in module distributedlog-core/distributedlog-protocol


hangc0276:
@StevenLuMT Please do not tag `cherry-picked/xx` on Prs, which is not cherry-picked to the target branch.

hangc0276 (approved):

PR #3697, opened 2022-12-08
HTTP GET request method is not implemented in the resume_compaction rest api handler

zwOvO:
HTTP GET request method is not implemented in the resume_compaction rest api handler


StevenLuMT:
rerun failure checks

zwOvO:
$ curl localhost:8000/api/v1/bookie/gc/resume_compaction
Not found method. Should be PUT to resume major or minor compaction**, Or GET to get compaction state**.

zwOvO:
https://github.com/apache/bookkeeper/blob/1e3efa112b66aa597def7cc247a1ffa9e7732301/bookkeeper-server/src/main/java/org/apache/bookkeeper/server/http/service/ResumeCompactionService.java#L81

The ResumeCompactionService class does not implement a response from the HTTP GET request method

zwOvO:
`Not found method. Should be PUT to resume major or minor compaction, Or GET /api/v1/bookie/gc_details to get compaction state.`

PR #3703, opened 2022-12-11
[refactor][bookkeeper] Refactor ByteBuf release method in stream/storage

HQebupt:
### Motivation
It may throw an exception when release  a `ByteBuf` object.  So the exception in `ByteBuf.release` should be checked.

### Changes
Use `ReferenceCountUtil.safeRelease()` instead of `ByteBuf.release()`.

Master Issue: [#5](https://github.com/HQebupt/bookkeeper/pull/5)


horizonzy (approved):
LGTM.

zymap (approved):

PR #3706, opened 2022-12-17
[improvement ]Avoid leaks  when use computeIfAbsent in ConcurrentLongHashMap and ConcurrentOpenHashMap

### LINKED ISSUE #3705: [improvement] Avoid leaks  when use computeIfAbsent in ConcurrentLongHashMap and ConcurrentOpenHashMap
# Motivation
When use computeIfAbsent  in ConcurrentLongHashMap and ConcurrentOpenHashMap:
image
There is a problem:
1. According to the get algorithm, it is equivalent to not inserting the key-value pair(Key Not found), and it is meaningless for valueProvider.apply(key) to return null, but we have not prevented such things from happening
2. When valueProvider.apply(key) returns null, unnecessary expansion will be performed at this time, and it cannot be recycled. It can only wait for reuse, which will cause leakage

# Changes
when value is null, should add checkNotNull  
image

Radiancebobo:
Fix  #3705
# Motivation
When use computeIfAbsent  in ConcurrentLongHashMap and ConcurrentOpenHashMap:
image
There are following problems:
1. According to the logic of the get function, it is equivalent to not inserting the key-value pair(Key Not found), and it is meaningless for valueProvider.apply(key) to return null, but we have not prevented such things from happening
2. When valueProvider.apply(key) returns null, unnecessary expansion will be performed at this time, and it cannot be recycled. It can only wait for reuse, which will cause leakage

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/util/collections/ConcurrentLongHashMap.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/util/collections/ConcurrentLongHashMap.java
@@ -382,6 +382,9 @@ public class ConcurrentLongHashMap<V> {
         }
 
         V put(long key, V value, int keyHash, boolean onlyIfAbsent, LongFunction<V> valueProvider) {
+            if (value == null) {
+                checkNotNull(valueProvider.apply(key));
+            }
             int bucket = keyHash;
 
             long stamp = writeLock();
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/util/collections/ConcurrentLongHashMap.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/util/collections/ConcurrentLongHashMap.java
@@ -226,6 +226,12 @@ public class ConcurrentLongHashMap<V> {
         return getSection(h).put(key, null, (int) h, true, provider);
     }
 
+    public V computePut(long key, LongFunction<V> provider, boolean onlyIfAbsent) {
+        checkNotNull(provider);
+        long h = hash(key);
+        return getSection(h).put(key, null, (int) h, onlyIfAbsent, provider);
+    }
+
     public V remove(long key) {
         long h = hash(key);
         return getSection(h).remove(key, null, (int) h);
@@ -382,9 +388,8 @@ public class ConcurrentLongHashMap<V> {
         }
 
         V put(long key, V value, int keyHash, boolean onlyIfAbsent, LongFunction<V> valueProvider) {
-            if (value == null) {
-                checkNotNull(valueProvider.apply(key));
-            }
+            V provideValue = null;
+
             int bucket = keyHash;
 
             long stamp = writeLock();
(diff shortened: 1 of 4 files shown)


Radiancebobo:
If we allow valueProvider.apply(key) to return null, we also need to avoid leaks, maybe change it like this:
image
so.
image

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/collections/ConcurrentLongHashMap.java
```diff
@@ -382,6 +382,9 @@ V get(long key, int keyHash) {
         }
 
         V put(long key, V value, int keyHash, boolean onlyIfAbsent, LongFunction<V> valueProvider) {
+            if (value == null) {
```
zymap, line 385: How about moving the check to the `computeIfAbsent`? I saw it checked whether the provider is null or not, we should also check the value.

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/collections/ConcurrentLongHashMap.java
```diff
@@ -382,6 +382,9 @@ V get(long key, int keyHash) {
...
+                checkNotNull(valueProvider.apply(key));
```
hangc0276, line 386: if `valueProvider.apply(key)` returns an object and is not null, it will create at least two objects, this is not expected.

hangc0276 (changes requested):

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/collections/ConcurrentLongHashMap.java
Radiancebobo: yes, I need a variable to store it。 like:
```
        V put(long key, V value, int keyHash, boolean onlyIfAbsent, LongFunction<V> valueProvider) {
            V provideValue;
            if (value == null) {
                provideValue = valueProvider.apply(key);
                if (provideValue == null) {
                    return null;
                }
            }
```
for the compatibility of the program, I feel that returning null directly is better than throwing NullPointerException .
what do you think?

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/collections/ConcurrentLongHashMap.java
Radiancebobo: Yes, this modification is also correct. maybe I can move to the `computeIfAbsent`

Radiancebobo: But I feel that it is more appropriate to do the verification in the put function, because this problem exists in ```org.apache.bookkeeper.util.collections.ConcurrentLongHashMap.Section#put``` , not the caller ```ConcurrentLongHashMap#computeIfAbsent```

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/collections/ConcurrentLongHashMap.java
Radiancebobo: There are still problems with this modification, let me think about it

Radiancebobo:
Found another bug, when the onlyIfAbsent of ```org.apache.bookkeeper.util.collections.ConcurrentLongHashMap.Section#put``` is false, the following unit test fails, I will fix it together:
image

image

Radiancebobo:
@hangc0276  @zymap  @merlimat  @eolivelli ，PTAL, thanks.

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/collections/ConcurrentLongHashMap.java
```diff
@@ -226,6 +226,12 @@ public V computeIfAbsent(long key, LongFunction<V> provider) {
         return getSection(h).put(key, null, (int) h, true, provider);
     }
 
+    public V computePut(long key, LongFunction<V> provider, boolean onlyIfAbsent) {
```
hangc0276, line 229: Why do we need this public interface?

Radiancebobo: the ```computePut```  is similar to ``` computeIfAbsent``` ,  computePut is currently only used for testing. 


> Found another bug, when the onlyIfAbsent of `org.apache.bookkeeper.util.collections.ConcurrentLongHashMap.Section#put` is false, the following unit test fails, I will fix it together: image
> 
> image

PR #3707, opened 2022-12-19
[improve] replace bookie from different rack of other quorum when do recovery

TakaHiR07:
### Motivation

If we need to do recovery when one bookie down, in the current RackAwarePolicy, it would try picking a candidate from same rack to replace the down bookie firstly. There is a risk that if many bookie in the same rack shutdown, it would result in replacing bookie to the several bookie remain in this rack, making these bookies become hot spot.  

### Changes

choose the replaced bookie from different rack of other quorum when do recovery or do ensemble change.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
@@ -471,9 +471,8 @@ public class RackawareEnsemblePlacementPolicyImpl extends TopologyAwareEnsembleP
                 LOG.debug("Try to choose a new bookie to replace {} from ensemble {}, excluding {}.",
                     bookieToReplace, ensembleNodes, excludeNodes);
             }
-            // pick a candidate from same rack to replace
+            // pick a candidate from different rack of excludeNodes to replace
             BookieNode candidate = selectFromNetworkLocation(
-                    bn.getNetworkLocation(),
                     networkLocationsToBeExcluded,
                     excludeNodes,
                     TruePredicate.INSTANCE,


dlg99 (changes requested):
Please make this configurable with current behavior enabled by default + add tests

TakaHiR07:
> For `RackAwarePolicy`, it will try its best effort to pick bookies from different racks. When doing recovery, it is safe to replace bookies from the same rack.

In some extreme case, such as 5 bookies in 1 rack, 4 bookies down, 1 bookie survive. The recovery throughput would concentrate on this bookie. After recovery, it would result in data skew in this survived bookie. I think it is not really safe, maybe cause some problem.
And when do recovery, it seems data locality is not exist. Because the recover bk client in ReplicationWorker is random distribution, it would read data from rack 2 and then recover data to rack 1. So it is not neccessary to ensure the replaced bookie in the same rack

hangc0276:
> > For `RackAwarePolicy`, it will try its best effort to pick bookies from different racks. When doing recovery, it is safe to replace bookies from the same rack.
> 
> In some extreme case, such as 5 bookies in 1 rack, 4 bookies down, 1 bookie survive. The recovery throughput would concentrate on this bookie. After recovery, it would result in data skew in this survived bookie. I think it is not really safe, maybe cause some problem. And when do recovery, it seems data locality is not exist. Because the recover bk client in ReplicationWorker is random distribution, it would read data from rack 2 and then recover data to rack 1. So it is not neccessary to ensure the replaced bookie in the same rack

@TakaHiR07 If we not ensure the replaced bookie in the same rack, it will break the placement policy in ensemble change and auto recovery.

hangc0276:
> > > > For `RackAwarePolicy`, it will try its best effort to pick bookies from different racks. When doing recovery, it is safe to replace bookies from the same rack.
> > > 
> > > 
> > > In some extreme case, such as 5 bookies in 1 rack, 4 bookies down, 1 bookie survive. The recovery throughput would concentrate on this bookie. After recovery, it would result in data skew in this survived bookie. I think it is not really safe, maybe cause some problem. And when do recovery, it seems data locality is not exist. Because the recover bk client in ReplicationWorker is random distribution, it would read data from rack 2 and then recover data to rack 1. So it is not neccessary to ensure the replaced bookie in the same rack
> > 
> > 
> > @TakaHiR07 If we not ensure the replaced bookie in the same rack, it will break the placement policy in ensemble change and auto recovery.
> 
> @hangc0276 This pr only use the implementation in
> 
> https://github.com/apache/bookkeeper/blob/e6722848e0a599f215e68e1340196c9dfe8e906e/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java#L558-L562
> 
> instead of
> https://github.com/apache/bookkeeper/blob/e6722848e0a599f215e68e1340196c9dfe8e906e/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java#L531-L536
> 
> 
> I think this implementation should not break the RackawarePolicy

@TakaHiR07 The `replaceBookie` will be used in ensemble change and auto recovery.

StevenLuMT:
rerun failure checks

PR #3709, opened 2022-12-20
[improve] Add journal retention time policy

### LINKED ISSUE #3708: Journal retains some backups that have expired for a long time
**BUG REPORT**

***Describe the bug***

1. We have a bookie work environment, which is deployed with other services mixed together.
2. Machine disk space allocation is relatively tight.
3. Bookie is not always in traffic mode, sometimes it has short traffic write peaks, sometimes it is always idle.
4. We hope that when the traffic peaks, a few more journal file backups can help us restore the bookie in abnormal situations, but we also hope that when the traffic is low, we can clean up the journal files that have been backed up for a long time.

As shown in the figure below, the oldest journal file is already on June 23, 2022, and now it is December 20, and during this period, our bookie has been running, and these journals have existed for a long time along with the backup strategy.

image

We can add a retention mechanism about time to help us guarantee the requirement of point 4.

wenbingshen:
### Motivation

Fixes #3708 

### Changes
Add two configs:

`journalRetentionTimeEnable`
`maxJournalRetentionTimeMs`

`journalRetentionTimeEnable` is disabled by default, consistent with existing logic.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
@@ -674,6 +674,9 @@ public class Journal extends BookieCriticalThread implements CheckpointSource {
     // number journal files kept before marked journal
     final int maxBackupJournals;
 
+    final long maxJournalRetentionTimeMs;
+    final boolean journalRetentionTimeEnable;
+
     final File journalDirectory;
     final ServerConfiguration conf;
     final ForceWriteThread forceWriteThread;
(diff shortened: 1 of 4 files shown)


hangc0276:
You can set `maxBackupJournals` to `0` to achieve your goal

wenbingshen:
> You can set `maxBackupJournals` to `0` to achieve your goal

No. my goal is not maxBackupJournals=0, you can check the detailed description in the issue #3708

StevenLuMT (approved):
LGTM

hangc0276:
> > You can set `maxBackupJournals` to `0` to achieve your goal
> 
> No. my goal is not maxBackupJournals=0, you can check the detailed description in the issue #3708

@wenbingshen 
> We hope that when the traffic peaks, a few more journal file backups can help us restore the bookie in abnormal situations, but we also hope that when the traffic is low, we can clean up the journal files that have been backed up for a long time.

This is your requirement, right?  If you set maxBackupJournals to `0`, the journal files will be deleted immediately after the journal checkpoint is completed. No matter if there is high traffic or not, the policy can ensure no data loss.

The journal files can be cleaned up based on the maxBackup policy, and won't cost too much disk space. I'm not sure whether it is valuable to introduce the time based retention policy.

hangc0276:
@wenbingshen Do you have any updates?

PR #3710, opened 2022-12-22
[feature] Added api/v1/bookie/cluster_info REST API

dlg99:

### Motivation

Information provided by current REST API is not enough (and cumbersome to combine) to answer such question as "is any data in danger if I shut down one more bookie".
E.g. getting list of underreplicated ledgers can get some info but it is either fast (no ledgers) or can be super slow on large cluster with some bookies lost (it retrieves full list of ledgers).
Even if there are no UR ledgers it still possible that the problem is that Auditor is down etc.

### Changes

 Added api/v1/bookie/cluster_info REST API

```
curl -s 127.0.0.1:8080/api/v1/bookie/cluster_info
{
  "auditorElected" : false,
  "auditorId" : "",
  "clusterUnderReplicated" : false,
  "ledgerReplicationEnabled" : true,
  "totalBookiesCount" : 1,
  "writableBookiesCount" : 1,
  "readonlyBookiesCount" : 0,
  "unavailableBookiesCount" : 0
}%
```

Side-fix:
`org.apache.bookkeeper.stream.cluster.StandaloneStarter` (used by bookie standalone) did not pass `LedgerManagerFactory` to the http server thus REST calls that needed it didn't work.

### DIFF
--- a/stream/server/src/main/java/org/apache/bookkeeper/stream/server/StorageServer.java
+++ b/stream/server/src/main/java/org/apache/bookkeeper/stream/server/StorageServer.java
@@ -260,10 +260,13 @@ public class StorageServer {
 
             // Build http service
             if (bkServerConf.isHttpServerEnabled()) {
-                MetadataBookieDriver metadataDriver = BookieResources.createMetadataDriver(bkServerConf, rootStatsLogger);
-                serverBuilder.addComponent(new AutoCloseableLifecycleComponent("metadataDriver", metadataDriver));
+                MetadataBookieDriver metadataDriver = BookieResources.createMetadataDriver(bkServerConf,
+                        rootStatsLogger);
+                serverBuilder.addComponent(new AutoCloseableLifecycleComponent("metadataDriver",
+                        metadataDriver));
                 LedgerManagerFactory ledgerManagerFactory = metadataDriver.getLedgerManagerFactory();
-                serverBuilder.addComponent(new AutoCloseableLifecycleComponent("lmFactory", ledgerManagerFactory));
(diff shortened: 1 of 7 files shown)


StevenLuMT:
rerun failure checks

StevenLuMT:
rerun failure checks

stream/server/src/main/java/org/apache/bookkeeper/stream/server/StorageServer.java
```diff
@@ -256,10 +260,19 @@ public static LifecycleComponent buildStorageServer(CompositeConfiguration conf,
 
             // Build http service
             if (bkServerConf.isHttpServerEnabled()) {
+                MetadataBookieDriver metadataDriver = BookieResources.createMetadataDriver(bkServerConf,
```
eolivelli, line 263: are we going to create an additional connection to ZooKeeper ?

eolivelli (changes requested):
very useful !
in BKVM we had to call many internal BookKeeperAdmin APIs to get these pieces of information

@nicoloboschi @mino181295 @aluccaroni @dianacle you may be interested

nicoloboschi (approved):

stream/server/src/main/java/org/apache/bookkeeper/stream/server/StorageServer.java
dlg99: AFAICT, there is an internal cache in MetadataDrivers so if there is one already it will be reused

dlg99:
@StevenLuMT dead link checker failure is fixed at https://github.com/apache/bookkeeper/pull/3712

eolivelli (approved):
Lgtm

PR #3715, opened 2022-12-23
[improve][doc] improve bk_server.conf docs

pgier:
Add default value of `allowMultipleDirsUnderSameDiskPartition`

Signed-off-by: Paul Gier <paul.gier@datastax.com>

### Motivation

Just a minor doc fix

### Changes

Updated the bk_server.conf and the website config doc

### DIFF
--- a/conf/bk_server.conf
+++ b/conf/bk_server.conf
@@ -41,7 +41,8 @@
 bookiePort=3181
 
 # Configure the bookie to allow/disallow multiple ledger/index/journal directories
-# in the same filesystem disk partition
+# in the same filesystem disk partition.
+# Defaults to true.
 # allowMultipleDirsUnderSameDiskPartition=false
 
 # Set the network interface that the bookie should listen on.
--- a/conf/bk_server.conf
+++ b/conf/bk_server.conf
@@ -43,7 +43,7 @@ bookiePort=3181
 # Configure the bookie to allow/disallow multiple ledger/index/journal directories
 # in the same filesystem disk partition.
 # Defaults to true.
-# allowMultipleDirsUnderSameDiskPartition=false
+# allowMultipleDirsUnderSameDiskPartition=true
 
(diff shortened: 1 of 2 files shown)


conf/bk_server.conf
```diff
@@ -41,7 +41,8 @@
 bookiePort=3181
 
 # Configure the bookie to allow/disallow multiple ledger/index/journal directories
-# in the same filesystem disk partition
+# in the same filesystem disk partition.
+# Defaults to true.
 # allowMultipleDirsUnderSameDiskPartition=false
```
nicoloboschi, line 46: Can you change also this line?

eolivelli (approved):

conf/bk_server.conf
eolivelli: If we change this value then we will change the default behaviour

@RaulGracia please also take a look. I think that in Pravega we rely on this file for default values

pgier: @eolivelli I thought that wouldn't affect the default value since it's commented out?  Wouldn't it still default to true based on the [sources](https://github.com/apache/bookkeeper/blob/032aef7e75f6b3487a6611a2b8af8306b9a3f7bb/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java#L3499-L3500).

I wasn't sure if the commented out examples were supposed to match the default values, or were to there to provide an each way to uncomment and override the default value.  I think there are examples of both cases currently in the config file.  Setting it to match the default value as @nicoloboschi suggested does seem a bit more intuitive to me.

pgier: @nicoloboschi I updated the example config to match the default `true`

StevenLuMT (approved):
LGTM

hangc0276 (approved):

nicoloboschi (approved):

PR #3719, opened 2022-12-26
BP-60:Change PCBC limitStatsLogging default value to true

### LINKED ISSUE #3718: BP-60:Change PCBC limitStatsLogging default value to true
### Motivation

We have an efficient online bookie cluster with hundreds of bookie nodes deployed on SSD disks.

We separate the AutoRecovery cluster and the Bookie cluster for independent deployment.

I observed that our AutoRecovery cluster GC is very frequent. After investigation, I found that the `limitStatsLogging` of the bookkeeper client PCBC is disabled by default, and a large number of channel monitoring indicators are generated. Due to the large number of bookie cluster nodes, this metric data occupies a large amount of heap memory.

A single `StringWriter` object occupies 16MB of memory, of which nearly 70 `StringWriter` objects are waiting for the next GC to be destroyed, occupying 1GB+ heap memory.

image

image

image

### Proposal

In my use, I haven't found any usefulness of these PCBC monitoring metrics data, at least so far, I haven't used it effectively.

If our AutoRecovery and Bookie cluster are mixed in one process, these large objects will affect the performance and stability of Bookie cluster.

Since I can't find the meaning of these metrics by default, I suggest to adjust the default value of `limitStatsLogging` to true.

Everyone can choose to turn it on or off, but by default, it is difficult for users to find out what effect this parameter will have, so that when their cluster grows to hundreds or thousands, when they realize the problem sometimes, it is necessary to restart hundreds to thousands of bookies in a rolling manner.

At the same time, I observed that in pulsar, various monitoring of the bookkeeper client is turned off by default, because they really affect the performance of the pulsar service, which is enough to show that we should try to change it, especially some very redundant metrics created based on channels.

### Compatibility, Deprecation, and Migration Plan

Clients that rely on PCBC metrics monitoring need to pay attention to this upgrade, but this will not affect the actual functions of the client, only the metrics data, and users can choose to open it again.

Proposal PR - #3719

wenbingshen:
### Motivation

Fixes #3718

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
@@ -1190,7 +1190,7 @@ public abstract class AbstractConfiguration<T extends AbstractConfiguration>
      *      the boolean flag indicating whether to limit stats logging
      */
     public boolean getLimitStatsLogging() {
-        return getBoolean(LIMIT_STATS_LOGGING, false);
+        return getBoolean(LIMIT_STATS_LOGGING, true);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


wenbingshen:
rerun failure checks

hangc0276 (approved):

horizonzy (approved):
LGTM.

StevenLuMT (approved):
LGTM

StevenLuMT (changes requested):
is this the BP?
BP needs to wait for the vote to be completed before merging
@eolivelli @dlg99

dlg99 (approved):
LGTM

wenbingshen:
> is this the BP? BP needs to wait for the vote to be completed before merging @eolivelli @dlg99

@StevenLuMT PTAL. Thanks. 
@eolivelli and @dlg99 have agreed to this change.

StevenLuMT (approved):
LGTM

PR #3733, opened 2023-01-11
Use ChannelVoidPromise to avoid useless promise objects creation

nicoloboschi:
### Motivation

In the bookie some writes callback are not considered. In those cases we can save useless allocations of default channel promises by using the voidPromise() method.

### Changes

* Use VoidChannelPromise if the promise result is never checked


bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
```diff
@@ -705,4 +713,8 @@ public boolean isBlacklisted(Channel channel) {
     public void handleNonWritableChannel(Channel channel) {
         onResponseTimeout.accept(channel);
     }
+
+    private static void writeAndFlush(Channel channel, Object msg) {
```
horizonzy, line 717: no need static.

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/NettyChannelUtil.java
```diff
@@ -0,0 +1,62 @@
...
+     * The promise is an instance of {@link VoidChannelPromise} that properly propagates exceptions up to the pipeline.
+     * Netty has many ad-hoc optimization if the promise is an instance of {@link VoidChannelPromise}.
+     * Lastly, it reduces pollution of useless {@link io.netty.channel.ChannelPromise} objects created
+     * by the default write and flush method {@link ChannelOutboundInvoker#writeAndFlush(Object)}.
+     * See https://stackoverflow.com/q/54169262 and https://stackoverflow.com/a/9030420 for more details.
+     *
+     * @param ctx channel's context
+     * @param msg buffer to write in the channel
+     */
+    public static void writeAndFlushWithVoidPromise(ChannelOutboundInvoker ctx, Object msg) {
```
horizonzy, line 45: `ctx` -> `out`

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/NettyChannelUtil.java
```diff
@@ -0,0 +1,62 @@
...
+    /**
+     * Write and flush the message to the channel.
+     *
+     * The promise is an instance of {@link VoidChannelPromise} that properly propagates exceptions up to the pipeline.
+     * Netty has many ad-hoc optimization if the promise is an instance of {@link VoidChannelPromise}.
+     * Lastly, it reduces pollution of useless {@link io.netty.channel.ChannelPromise} objects created
+     * by the default write and flush method {@link ChannelOutboundInvoker#writeAndFlush(Object)}.
+     * See https://stackoverflow.com/q/54169262 and https://stackoverflow.com/a/9030420 for more details.
+     *
+     * @param ctx channel's context
```
horizonzy, line 42: In some cases, the `ctx` also be channel

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/NettyChannelUtil.java
horizonzy, line 45: In the netty. `channel#write` is different from `ctx#write`, the `channel#write` will send the msg from tail to head. `context.write` only sends the msg to the before ctx.
Here transfer a static method; I'm afraid it will be abused in the future. I suggest didn't transfer an util to handle it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
nicoloboschi: It's best practice to make a method static if possible. 
Look at this reference for example: https://stackoverflow.com/questions/2671496/when-to-use-static-methods

horizonzy: I got, thanks.

StevenLuMT:
rerun failure checks

StevenLuMT:
test case run failed , maybe have a look @nicoloboschi 
image

horizonzy:
@nicoloboschi This change is useful. Could you have a look at the CI?

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/NettyChannelUtil.java
nicoloboschi: what's your concern? even if different, the right implementation method will be called based on the pass argument

horizonzy: Some developers may didn't notice the difference between ctx and channel. If they want to writeAndFlush the data. They didn't know the intput param should be channel or ctx.

hangc0276 (approved):
Nice catch!

PR #3736, opened 2023-01-13
[improve] Optimize masterKeyCache StorageNotificationListener

wenbingshen:
### Motivation

Related to PR #3522, in order to implement masterKeyCache to clean up the deleted ledgerId, PR #3522 introduced LedgerStorageNotificationListener, which is used to monitor the cleanup action of calling back masterKeyCache when the ledger is deleted. I noticed that before this, LedgerStorage already has The purpose of the implementation of LedgerDeletionListener is to call back the corresponding user to do some cleaning operations after the ledger is deleted, which fully meets the requirements of LedgerStorageNotificationListener.

The function of the additionally introduced LedgerStorageNotificationListener is the same as that of LedgerDeletionListener, but a strange operation L is introduced in the GC thread: the GC thread needs to process a LedgerStorageNotificationListener notification, and the GC thread calls ledgerStorage.deleteLedger(ledgerId), and it will be executed  LedgerDeletionListener notification again during the processing deleteLedger.

### Changes

Replace LedgerStorageNotificationListener with existing LedgerDeletionListener to achieve this requirement.


wenbingshen:
@StevenLuMT @eolivelli PTAL. Thanks.

wenbingshen:
rerun failure checks

wenbingshen:
rerun failure checks

eolivelli (approved):
Good catch !

very nice

hangc0276 (approved):

StevenLuMT (approved):
good work

PR #3738, opened 2023-01-16
Pin order executor in LedgerHandle

AnonHxy:

### Motivation

Pin the order executor in `LedgerHandle`  to save the choose thread overhead. For details reason:

We could call `executeOrdered()` frequently, e.g., when in `asyncReadEntries`. And we would compute the thread index from ledgerId every time when call `executeOrdered()`. So pin the order executor will save the thread index compution time.

### Changes

- Choose the order executor in `LedgerHandle`  constructor method and save it as a final field
- Override `getMainWorkerPool()` in `MockBookKeeper` to fix UT NPE exception when call `this.executor = clientCtx.getMainWorkerPool().chooseThread(ledgerId);`

### DIFF
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/MockBookKeeper.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/client/MockBookKeeper.java
@@ -25,8 +25,6 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
@@ -37,6 +35,7 @@ import org.apache.bookkeeper.client.api.BKException.Code;
 import org.apache.bookkeeper.client.api.OpenBuilder;
 import org.apache.bookkeeper.client.api.ReadHandle;
 import org.apache.bookkeeper.client.impl.OpenBuilderBase;
+import org.apache.bookkeeper.common.util.OrderedExecutor;
 import org.apache.bookkeeper.conf.ClientConfiguration;
 import org.apache.zookeeper.ZooKeeper;
 import org.slf4j.Logger;
@@ -49,7 +48,10 @@ import org.slf4j.LoggerFactory;
  */
 public class MockBookKeeper extends BookKeeper {
 
-    final ExecutorService executor = Executors.newFixedThreadPool(1, new DefaultThreadFactory("mock-bookkeeper"));
+    final OrderedExecutor orderedExecutor = OrderedExecutor.newBuilder()
+            .numThreads(1)
+            .threadFactory(new DefaultThreadFactory("mock-bookkeeper"))
+            .build();
     final ZooKeeper zkc;
 
     @Override
@@ -66,6 +68,11 @@ public class MockBookKeeper extends BookKeeper {
 
     public MockBookKeeper(ZooKeeper zkc) throws Exception {
         this.zkc = zkc;
+    }
+
+    @Override
+    public OrderedExecutor getMainWorkerPool() {
+        return orderedExecutor;
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


gaozhangmin (approved):

AnonHxy:
rerun failure checks

AnonHxy:
PTAL @StevenLuMT  @eolivelli

hangc0276:
@AnonHxy Would you please provide more context about the reason to pin the executor? thanks.

AnonHxy:
> @AnonHxy Would you please provide more context about the reason to pin the executor? thanks.

Sure. It was inspired by this PR https://github.com/apache/pulsar/pull/18078  @hangc0276

AnonHxy:
PTAL @hangc0276 @eolivelli , thanks

StevenLuMT (changes requested):
when the execution mode changes, the performance change significantly ? ledgerID will block reading. Is it expected?

AnonHxy:
> when the execution mode changes, the performance change significantly ? ledgerID will block reading. Is it expected?

I think this patch will not change the executeion mode.  It just make the execute thread as a final field,  which is a obvious improvement. But I am not sure how much can be improved. @StevenLuMT

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/MockBookKeeper.java
```diff
@@ -49,7 +48,10 @@
  */
 public class MockBookKeeper extends BookKeeper {
 
-    final ExecutorService executor = Executors.newFixedThreadPool(1, new DefaultThreadFactory("mock-bookkeeper"));
+    final OrderedExecutor orderedExecutor = OrderedExecutor.newBuilder()
```
hangc0276, line 51: Why change this thread pool?

hangc0276 (commented):
Overall LGTM. This change saves the cost of choosing threads in executeOrdered, and I wonder if choosing threads costs a lot of CPU resources and whether this change can improve the performance or not.

AnonHxy:
> Overall LGTM. This change saves the cost of choosing threads in executeOrdered, and I wonder if choosing threads costs a lot of CPU resources and whether this change can improve the performance or not.

I tested it using one bookie client to add entry async like following.  It looks that it was not a significant improvement from the result.

```
 for (int i = 0; i < Integer.MAX_VALUE; ++i) {
                handle.asyncAddEntry("Some entry data".getBytes(), new AsyncCallback.AddCallback() {
                    @Override
                    public void addComplete(int rc, LedgerHandle lh, long entryId, Object ctx) {

                    }
                }, null);
 }
```

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/MockBookKeeper.java
AnonHxy: We need call `getMainWorkerPool()`  in `LedgerHandle` constructor method. If we didn't change it, the `getMainWorkerPool()` will return null and throw NPE in the constructor method:

[Line199](https://github.com/apache/bookkeeper/pull/3738/files#diff-1d893bb31553b5e1f55c8301d04ae15f38e0d35f531f9dd22475128b7972ddf9R197-R199)

hangc0276:
> > Overall LGTM. This change saves the cost of choosing threads in executeOrdered, and I wonder if choosing threads costs a lot of CPU resources and whether this change can improve the performance or not.
> 
> I tested it using one bookie client to add entry async like following. It looks that it was not a significant improvement from the result.
> 
> ```
>  for (int i = 0; i < Integer.MAX_VALUE; ++i) {
>                 handle.asyncAddEntry("Some entry data".getBytes(), new AsyncCallback.AddCallback() {
>                     @Override
>                     public void addComplete(int rc, LedgerHandle lh, long entryId, Object ctx) {
> 
>                     }
>                 }, null);
>  }
> ```

@AnonHxy If there is no significant performance improvement, I think we can hold this change.

hangc0276:
@AnonHxy Do you have any updates?

AnonHxy:
> 



> I think we can hold on this change.

OK. But I think this change is a positive optimization at least, while the proform result is not obvious ... @hangc0276

hangc0276 (approved):
LGTM

StevenLuMT (approved):
Good catch,LGTM

AnonHxy:
@eolivelli @dlg99 @merlimat PTAL, thanks:)

eolivelli (approved):
LGTM

PR #3740, opened 2023-01-17
Fix entryId in ReadRequest when call readLac

AnonHxy:

### Motivation

The `entryId` of `ReadRequest` should be -1 instead of 0 when call `PerChannelBookieClient#readLac`. 
Luckly we will call `PerChannelBookieClient#readLac` only when `useV2WireProtocol` is false,  so it will not cause problem in production environment.  But it's a potential bug if we refactor the readLac in v2 protocol, so it's necessary to fix it I think

### Changes
Changing `entryId` of `ReadRequest` in `PerChannelBookieClient#readLac` from 0 to `-1`


eolivelli (changes requested):
So this code is never executed.

We should drop it and not try to fix

AnonHxy:
> So this code is never executed.
> 
> We should drop it and not try to fix

Thanks for your reply @eolivelli . I have modified this patch to cleanup it

StevenLuMT (changes requested):
useV2WireProtocol has judgment in other methods, you modify this, I feel a little change semantics, why not fix useV2WireProtocol to false logic?

hangc0276 (changes requested):
Currently, the `PerChannelBookieClient#readLac` is only called by V3 protocol, but we can't ensure someone who doesn't know the context called this method in V2 protocol.  

IMO, we'd better add the V2 protocol support back and fix it. Please add a test to cover your fix, because the code block you changed never runs into, and we need to ensure your change works as expected.

hangc0276:
@AnonHxy Do you have any updates?

AnonHxy:
> 

OK. I will add a test to verfiy it later

AnonHxy:
@StevenLuMT @eolivelli @hangc0276 PTAL, thanks

PR #3747, opened 2023-01-18
Fix QueueEntry recycle problem.

horizonzy:
In the QueueEntry recycle, it only recycles itself to the object pool, but didn't reset some properties.
Like entry, cb, etc. We should reset the filed before recycles itself.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
@@ -365,6 +365,14 @@ public class Journal extends BookieCriticalThread implements CheckpointSource {
         };
 
         private void recycle() {
+            this.entry = null;
+            this.cb = null;
+            this.ctx = null;
+            this.journalAddEntryStats = null;
+            this.journalCbQueuedLatency = null;
+            this.journalCbQueueSize = null;
+            this.cbThreadPoolQueueSize = null;
+            this.callbackTime = null;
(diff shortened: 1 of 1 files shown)


hangc0276 (approved):
Nice catch!

nicoloboschi (approved):

StevenLuMT (approved):
Great work

wenbingshen (approved):
nice work!

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -347,8 +347,8 @@ public void run() {
             cbThreadPoolQueueSize.dec();
             journalAddEntryStats.registerSuccessfulEvent(MathUtils.elapsedNanos(enqueueTime), TimeUnit.NANOSECONDS);
             cb.writeComplete(0, ledgerId, entryId, null, ctx);
-            recycle();
             callbackTime.addLatency(MathUtils.elapsedNanos(startTime), TimeUnit.NANOSECONDS);
+            recycle();
```
lhotari, line 351: @horizonzy This change wasn't mentioned in the PR description. Why was this missing before? Was the recycling completely unused before? /cc @hangc0276

horizonzy: Cause we already reset `callbackTime` to null. And we should ensure do the all works before recycle self.

PR #3749, opened 2023-01-19
Fix ForceWriteRequest recycle problem

gaozhangmin:
The ForceWriteRequest is recycled without resetting properties.

image


eolivelli (approved):
LGTM

horizonzy (commented):
The basic type will be overridden at the next reuse. I think we needn't to reset the basic type value. It doesn't help GC work. And the old value can help to trace problems when we dump.

StevenLuMT (approved):
LGTM

hangc0276 (changes requested):
+1 for no need to reset the basic type values.

PR #3750, opened 2023-01-19
Fix PendingAddOp recycle problem

gaozhangmin:
Fix PendingAddOp and  V2CompletionKey  recycle problem


horizonzy (commented):
The basic type will be overridden at the next reuse. I think we needn't to reset the basic type value. It doesn't help GC work. And the old value can help to trace problems when we dump.

hangc0276 (changes requested):
+1 for no need to reset basic type value.

PR #3372, opened 2022-06-28
add and check indexDirs in Cookie meta

StevenLuMT:
Descriptions of the changes in this PR:

### Motivation

planning for index dir: [mail talking](https://lists.apache.org/thread/r657jf55khl59bbqltj2s95107lbkr0w)
stage 2: add and check indexDirs in cookie meta

### Changes

1. add indexDirs in cookie meta
2. check indexDirs in cookie meta
3. add a testcase for cookie

PR #3762, opened 2023-02-02
[improve] Fix indexDirs upgrade failed

wenbingshen:
### Motivation

Now we have written indexDir into Cookie in #3372 .

When we configure an independent indexDir, the `FileSystemUpgrade` tool will fail when the cookie is an old version or a new cookie version will be added in the future, because it does not consider indexDirs when upgrading, so we should add indexDirs to the upgrade step.

### Changes
You can use the following two implementations of getAllDirectories in FileSystemUpgrade, and then run some tests added by this PR in UpgradeTest, and you will find that the first implementation will fail the test, while the second implementation will succeed.
1:
```java
    private static List<File> getAllDirectories(ServerConfiguration conf) {
        List<File> dirs = new ArrayList<>();
        dirs.addAll(Lists.newArrayList(conf.getJournalDirs()));
        Collections.addAll(dirs, conf.getLedgerDirs());
        return dirs;
    }
```
image

2:
```java
    private static List<File> getAllDirectories(ServerConfiguration conf) {
        List<File> dirs = new ArrayList<>();
        dirs.addAll(Lists.newArrayList(conf.getJournalDirs()));
        final File[] ledgerDirs = conf.getLedgerDirs();
        final File[] indexDirs = conf.getIndexDirs();
        if (indexDirs != null && indexDirs != ledgerDirs) {
            dirs.addAll(Lists.newArrayList(indexDirs));
        }
        Collections.addAll(dirs, ledgerDirs);
        return dirs;
    }
```
image

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileSystemUpgrade.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileSystemUpgrade.java
@@ -99,7 +99,12 @@ public class FileSystemUpgrade {
     private static List<File> getAllDirectories(ServerConfiguration conf) {
         List<File> dirs = new ArrayList<>();
         dirs.addAll(Lists.newArrayList(conf.getJournalDirs()));
-        Collections.addAll(dirs, conf.getLedgerDirs());
+        final File[] ledgerDirs = conf.getLedgerDirs();
+        final File[] indexDirs = conf.getIndexDirs();
+        if (indexDirs != null && indexDirs != ledgerDirs) {
+            dirs.addAll(Lists.newArrayList(indexDirs));
+        }
+        Collections.addAll(dirs, ledgerDirs);
         return dirs;
     }
 
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileSystemUpgrade.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileSystemUpgrade.java
@@ -24,6 +24,7 @@ package org.apache.bookkeeper.bookie;
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static org.apache.bookkeeper.meta.MetadataDrivers.runFunctionWithRegistrationManager;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Lists;
 import com.google.common.util.concurrent.UncheckedExecutionException;
 import java.io.File;
@@ -97,13 +98,14 @@ public class FileSystemUpgrade {
             }
         };
 
-    private static List<File> getAllDirectories(ServerConfiguration conf) {
+    @VisibleForTesting
+    public static List<File> getAllDirectories(ServerConfiguration conf) {
         List<File> dirs = new ArrayList<>();
         dirs.addAll(Lists.newArrayList(conf.getJournalDirs()));
         final File[] ledgerDirs = conf.getLedgerDirs();
         final File[] indexDirs = conf.getIndexDirs();
-        if (indexDirs != null
-                && !Arrays.asList(indexDirs).equals(Arrays.asList(ledgerDirs))) {
+        if (indexDirs != null && indexDirs.length == ledgerDirs.length
+                && !Arrays.asList(indexDirs).containsAll(Arrays.asList(ledgerDirs))) {
             dirs.addAll(Lists.newArrayList(indexDirs));
         }
         Collections.addAll(dirs, ledgerDirs);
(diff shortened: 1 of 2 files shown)


eolivelli (approved):
+1

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileSystemUpgrade.java
```diff
@@ -99,7 +99,12 @@ public boolean accept(File dir, String name) {
     private static List<File> getAllDirectories(ServerConfiguration conf) {
         List<File> dirs = new ArrayList<>();
         dirs.addAll(Lists.newArrayList(conf.getJournalDirs()));
-        Collections.addAll(dirs, conf.getLedgerDirs());
+        final File[] ledgerDirs = conf.getLedgerDirs();
+        final File[] indexDirs = conf.getIndexDirs();
+        if (indexDirs != null && indexDirs != ledgerDirs) {
```
hangc0276, line 104: We can't use `!=` to check if two array equals or not

hangc0276 (changes requested):

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileSystemUpgrade.java
wenbingshen: @hangc0276 I have addressed your comments. Please take a look. Thanks.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/FileSystemUpgrade.java
```diff
@@ -99,7 +100,13 @@ public boolean accept(File dir, String name) {
     private static List<File> getAllDirectories(ServerConfiguration conf) {
         List<File> dirs = new ArrayList<>();
         dirs.addAll(Lists.newArrayList(conf.getJournalDirs()));
-        Collections.addAll(dirs, conf.getLedgerDirs());
+        final File[] ledgerDirs = conf.getLedgerDirs();
+        final File[] indexDirs = conf.getIndexDirs();
+        if (indexDirs != null
+                && !Arrays.asList(indexDirs).equals(Arrays.asList(ledgerDirs))) {
```
StevenLuMT, line 106: `Arrays.asList(indexDirs).equals(Arrays.asList(ledgerDirs))` 
How to judge the correctness of this sentence? I think you need to add two more examples to cover this code

wenbingshen: @StevenLuMT You are right! I made a mistake here, I updated the method and added the test to cover it, please help to review, thank you!

StevenLuMT: it is not right
ledgerDirs and indexDirs are one-to-one mappings
If this state occurs, it feels that there will be an error in judgment, which will lead to more additions?

ledgerDirs=/t1,/t2,/t3
indexDirs=/t1,/i1,/i3 

> @StevenLuMT You are right! I made a mistake here, I updated the method and added the test to cover it, please help to review, thank you!

wenbingshen: @StevenLuMT I have considered the 1-to-1 situation you mentioned, but I think that in this tool, it does not care whether it is 1-to-1, but only focuses on whether the number of ledgerDirs and indexDirs and whether the absolute paths are the same.

wenbingshen: If you are worried about the existence of duplicate directories, resulting in multiple executions, then we can use the Set collection, but the original implementation did not consider whether JournalDirs and LedgerDirs are duplicated? Also, 1 vs 1 doesn't make a difference here.

StevenLuMT: ok, LGTM, let Enrico or others review it again @wenbingshen 

> If you are worried about the existence of duplicate directories, resulting in multiple executions, then we can use the Set collection, but the original implementation did not consider whether JournalDirs and LedgerDirs are duplicated? Also, 1 vs 1 doesn't make a difference here.

wenbingshen: @StevenLuMT Thanks. :)

wenbingshen:
@dlg99 @eolivelli @hangc0276 Please help take a look. Thanks!

StevenLuMT (approved):
LGTM

hangc0276 (approved):

PR #3596, opened 2022-11-01
Optimized debug log enable

wenbingshen:
### Motivation

As the title.

PR #3763, opened 2023-02-02
[improve] [Follow-PR#3596] Optimize logger debug enable to reduce allocating objects

wenbingshen:
### Motivation

Following PR #3596


wenbingshen:
@hangc0276 PTAL. Thanks!

PR #3764, opened 2023-02-02
[improve] [Follow-PR#3596] Optimize log trace enable to reduce allocating objects

wenbingshen:
### Motivation

Following PR #3596

### DIFF
--- a/stream/distributedlog/core/src/main/java/org/apache/distributedlog/BKAsyncLogReader.java
+++ b/stream/distributedlog/core/src/main/java/org/apache/distributedlog/BKAsyncLogReader.java
@@ -364,7 +364,9 @@ class BKAsyncLogReader implements AsyncLogReader, Runnable, AsyncNotification {
 
         Throwable cause = lastExceptionUpdater.get(this);
         if (null != cause) {
-            LOG.trace("Cancelling pending reads");
+            if (LOG.isTraceEnabled()) {
+                LOG.trace("Cancelling pending reads");
+            }
             cancelAllPendingReads(cause);
             return true;
         }
@@ -593,7 +595,9 @@ class BKAsyncLogReader implements AsyncLogReader, Runnable, AsyncNotification {
 
                     // Queue is empty, nothing to read, return
                     if (null == nextRequest) {
-                        LOG.trace("{}: Queue Empty waiting for Input", readHandler.getFullyQualifiedName());
+                        if (LOG.isTraceEnabled()) {
+                            LOG.trace("{}: Queue Empty waiting for Input", readHandler.getFullyQualifiedName());
+                        }
                         scheduleCountUpdater.set(this, 0);
                         backgroundReaderRunTime.registerSuccessfulEvent(
                             runTime.stop().elapsed(TimeUnit.MICROSECONDS), TimeUnit.MICROSECONDS);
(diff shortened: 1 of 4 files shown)


wenbingshen:
@hangc0276 PTAL. Thanks!

stream/distributedlog/core/src/main/java/org/apache/distributedlog/BKAsyncLogReader.java
```diff
@@ -364,7 +364,9 @@ private boolean checkClosedOrInError(String operation) {
 
         Throwable cause = lastExceptionUpdater.get(this);
         if (null != cause) {
-            LOG.trace("Cancelling pending reads");
+            if (LOG.isTraceEnabled()) {
```
nicoloboschi, line 367: this is not useful

stream/distributedlog/core/src/main/java/org/apache/distributedlog/impl/metadata/ZKLogStreamMetadataStore.java
```diff
@@ -269,10 +269,14 @@ public void processResult(final int rc, final String path, Object ctx, String na
                                             logMetadata.getFullyQualifiedName())));
                         } else if (KeeperException.Code.OK.intValue() == rc) {
                             FutureUtils.complete(promise, null);
-                            LOG.trace("Created path {}.", path);
+                            if (LOG.isTraceEnabled()) {
```
nicoloboschi, line 272: this is not useful

stream/distributedlog/core/src/main/java/org/apache/distributedlog/impl/metadata/ZKLogStreamMetadataStore.java
```diff
@@ -269,10 +269,14 @@ public void processResult(final int rc, final String path, Object ctx, String na
...
+                                LOG.trace("Created path {}.", path);
+                            }
                         } else if (KeeperException.Code.NODEEXISTS.intValue() == rc) {
                             FutureUtils.complete(promise, null);
-                            LOG.trace("Path {} is already existed.", path);
+                            if (LOG.isTraceEnabled()) {
```
nicoloboschi, line 277: this is not useful

stream/distributedlog/core/src/main/java/org/apache/distributedlog/lock/ZKDistributedLock.java
```diff
@@ -181,7 +181,9 @@ void doAsyncAcquireWithSemaphore(final CompletableFuture<ZKDistributedLock> acqu
 
     void doAsyncAcquire(final CompletableFuture<ZKDistributedLock> acquirePromise,
                         final long lockTimeout) {
-        LOG.trace("Async Lock Acquire {}", lockPath);
+        if (LOG.isTraceEnabled()) {
```
nicoloboschi, line 184: this is not useful

stream/distributedlog/core/src/main/java/org/apache/distributedlog/lock/ZKSessionLock.java
```diff
@@ -770,7 +770,9 @@ public void execute() {
 
                                         currentNode = name;
                                         currentId = getLockIdFromPath(currentNode);
-                                        LOG.trace("{} received member id for lock {}", lockId, currentId);
+                                        if (LOG.isTraceEnabled()) {
```
nicoloboschi, line 773: this is not useful

nicoloboschi (changes requested):
If there's no computation to get the value to print, there's no need to add those conditions

stream/distributedlog/core/src/main/java/org/apache/distributedlog/BKAsyncLogReader.java
wenbingshen: @nicoloboschi I think we should be consistent in style, consistent in form. What do you think?

nicoloboschi: I think it's fine to leave the logging libraries to perform optimization if needed.
In this case there's no improvement.
Also, the code is less readable

It's fine to have cases where we check if some level is enabled to not perform values computation if not needed but in general it's not a good practice to guard every logging call

PR #3767, opened 2023-02-03
[fix] Cover negative numReadWorkerThreads

wenbingshen:
### Motivation

As the title.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
@@ -213,7 +213,7 @@ public class DbLedgerStorage implements LedgerStorage {
                     new DefaultThreadFactory("EntryLoggerFlush"));
 
                 int numReadThreads = conf.getNumReadWorkerThreads();
-                if (numReadThreads == 0) {
+                if (numReadThreads <= 0) {
                     numReadThreads = conf.getServerNumIOThreads();
                 }
 


hangc0276 (commented):
Setting the number of read threads to negative is unreasonable.

wenbingshen:
> Setting the number of read threads to negative is unreasonable.

@hangc0276 Yes. So we should check or cover it. I see that in other places where numReadWorkerThreads is used, negative numbers are covered.

Shoothzj (approved):

Shoothzj:
+0 to this change. it doesn't feel very necessary.

wenbingshen:
> +0 to this change. it doesn't feel very necessary.

@Shoothzj Can you explain why it's not necessary? From the code, numReadWorkerThreads covers negative values in other places where it is used

```java
    private OrderedExecutor createExecutor(
            int numThreads,
            String nameFormat,
            int maxTasksInQueue,
            StatsLogger statsLogger) {
        if (numThreads <= 0) {.         <==   here
            return null;
        } else {
            return OrderedExecutor.newBuilder()
                    .numThreads(numThreads)
                    .name(nameFormat)
                    .traceTaskExecution(serverCfg.getEnableTaskExecutionStats())
                    .preserveMdcForTaskExecution(serverCfg.getPreserveMdcForTaskExecution())
                    .statsLogger(statsLogger)
                    .maxTasksInQueue(maxTasksInQueue)
                    .enableThreadScopedMetrics(true)
                    .build();
        }
    }
```

And here, an `IllegalArgumentException` will be thrown in `DirectEntryLogger`

```java
    Cache<Integer, LogReader> cache = CacheBuilder.newBuilder()
                    .maximumWeight(perThreadBufferSize)   <== here
                    .weigher((key, value) -> readBufferSize)
                    .removalListener(rl)
                    .expireAfterAccess(maxFdCacheTimeSeconds, TimeUnit.SECONDS)
                    .concurrencyLevel(1) // important to avoid too aggressive eviction
                    .build();

    public CacheBuilder<K, V> maximumWeight(long maximumWeight) {
        Preconditions.checkState(this.maximumWeight == -1L, "maximum weight was already set to %s", this.maximumWeight);
        Preconditions.checkState(this.maximumSize == -1L, "maximum size was already set to %s", this.maximumSize);
        Preconditions.checkArgument(maximumWeight >= 0L, "maximum weight must not be negative");    <== here will throw IllegalArgumentException
        this.maximumWeight = maximumWeight;
        return this;
    }
```

Shoothzj:
@wenbingshen 
Personally, I don't think this is crucial. While this isn't a reasonable config value, the patch does indeed provide directly clearified error output when configured with negative values. Howerver, if others agree, I won't object.

StevenLuMT (changes requested):
configure negative numbers? please describe the motivation

hangc0276:
@wenbingshen Would you please provide more motivation? If no strong motivations, I suggest closing this PR, thanks.

PR #3772, opened 2023-02-09
Execute clean indexes in finally

### LINKED ISSUE #3771: db-storage-cleanup would impact write request
I have a question about db-storage-cleanup executor, Why  indexes and ledgers metadata are cleaned Only when `writeCacge.isEmpty()`
https://github.com/apache/bookkeeper/blob/4ca4b4e4845ec1a3d5b17c96c99e97a62b5ec6c7/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java#L742-L816

There is a case, if bookie transits to readonly mode for a long time, during readonly, lots of ledgers would be deleted, but because `writeCacge.isEmpty()`,  the index would remain. 
After we make the bookie writable,  a lot of  indexes gathered that need be deleted.  put heavy load on RocksDB database

gaozhangmin:
### Motivation
Fixes  #3771

### Changes
Move cleanupExecutor to finally block to make sure it can be executed every checkpoint process.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
@@ -800,21 +800,6 @@ public class SingleDirectoryDbLedgerStorage implements CompactableLedgerStorage 
             ledgerIndex.flush();
             recordSuccessfulEvent(dbLedgerStorageStats.getFlushLedgerIndexStats(), ledgerIndexStartTime);
 
-            cleanupExecutor.execute(() -> {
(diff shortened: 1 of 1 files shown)


hangc0276 (changes requested):
What's the difference between moving the code block into finally?

StevenLuMT (changes requested):
I think it doesn't solve the fundamental problem.

gaozhangmin:
> What's the difference between moving the code block into finally?

Moving to finally, The index clean up is going on when there are no data writing to bookie.
Which can address the issue i mentioned in #3771, When a bookie transits from readonly to write-read mode. lots of index cleanup would gather @hangc0276  @StevenLuMT

hangc0276 (approved):

StevenLuMT (approved):
lgtm

PR #3773, opened 2023-02-09
[improve] Let the replication worker can rerplicate the fresh ledgers at first.

Nicklee007:
### Motivation

The old replication worker always rereplicate the ledger which wait to rereplicate is random. But the oldest ledger maybe easier to reach the retention time and be deleted after we rereplicated them, so the better way is to choice the fresh ledger to rereplicate at first. In this way we can decrease the rereplicate data size and not need to rereplicate the older ledger which be deleted before we rereplicate it.

### Changes

1. Changed the replicate worker get ledger to rereplicate order.


bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/ZkLedgerUnderreplicationManager.java
```diff
@@ -557,10 +557,15 @@ private long getLedgerToRereplicateFromHierarchy(String parent, long depth)
...
+        if (depth == 3) {
+            // Avoid workers want to get the same ledger lock.
+            Collections.shuffle(children);
+        } else {
+            Collections.sort(children);
+        }
 
         while (children.size() > 0) {
-            String tryChild = children.get(0);
+            String tryChild = children.get(children.size() - 1);
```
dlg99, line 568: can simply reverse sort order above

dlg99 (changes requested):
Do you have any metrics / proof that the change actually improves something? 
If you do, please add tests and share the metrics.

hangc0276:
Replicating the fresh ledgers first may lead to a lot of unnecessary replication throughput when `lostBookieRecoveryDelay` configured too short.

If the user set `lostBookieRecoveryDelay` to a short value, such as `0` or `60`, and a bookie restarted for more than 60s, the auditor will mark all the ledgers which located on the restarted bookie as under-replicated. If the replication worker replicated the fresh ledgers first, there will be a lot of fresh ledgers being replicated. However, after the bookie restart succced, those ledgers' replication is unnecessary.

hangc0276:
> Replicating the fresh ledgers first may lead to a lot of unnecessary replication throughput when `lostBookieRecoveryDelay` configured too short.
> 
> If the user set `lostBookieRecoveryDelay` to a short value, such as `0` or `60`, and a bookie restarted for more than 60s, the auditor will mark all the ledgers which located on the restarted bookie as under-replicated. If the replication worker replicated the fresh ledgers first, there will be a lot of fresh ledgers being replicated. However, after the bookie restart succced, those ledgers' replication is unnecessary.

Sorry, I misunderstood this part. I support this change.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/ZkLedgerUnderreplicationManager.java
```diff
@@ -557,10 +557,15 @@ private long getLedgerToRereplicateFromHierarchy(String parent, long depth)
             return -1;
         }
 
-        Collections.shuffle(children);
+        if (depth == 3) {
+            // Avoid workers want to get the same ledger lock.
+            Collections.shuffle(children);
+        } else {
+            Collections.sort(children);
```
hangc0276, line 564: It will sort according to the `String` type order instead of the LedgerId long type order, which can't more sure the sorted list is ordered by ledger creation time.

hangc0276 (changes requested):

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/ZkLedgerUnderreplicationManager.java
Nicklee007: The node of `subTreeCache.getChildren(parent)`  is 4 hex bytes，` Collections.sort()` could get the correct order.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/ZkLedgerUnderreplicationManager.java
Nicklee007: addressed

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/ZkLedgerUnderreplicationManager.java
```diff
@@ -557,7 +557,13 @@ private long getLedgerToRereplicateFromHierarchy(String parent, long depth)
...
+            Collections.reverse(children);
```
hangc0276, line 565: Collections.sort(children, Collections.reverseOrder());

Nicklee007: addressed

hangc0276 (commented):
I have one concern about the sort. If we have multiple replication workers, and each replication worker sorts the ledgers before getting the target ledger to replicate. Would this PR introduce more Zookeeper try-lock operation conflicts?

Nicklee007:
> I have one concern about the sort. If we have multiple replication workers, and each replication worker sorts the ledgers before getting the target ledger to replicate. Would this PR introduce more Zookeeper try-lock operation conflicts?

@hangc0276 we set a condition `if (depth == 3)` to avoid the conflicts. 
The zk under replicate ledger dir have 4 layers under /unserreplication/ledgers,  like `/aaaa/bbbb/cccc/dddd`. So, in the PR only sorted the first three layers `/aaaa/bbbb/cccc` to keep we can replicate the fresh ledger at first as far as possible,  and the last layer `/dddd` each worker shuffled znode to avoid workers always  try-lock the same znode and  operation conflicts.

hangc0276 (approved):

hangc0276:
@dlg99 Would you please help review this PR again? Thanks.

Nicklee007:
@dlg99 Would you please help review this PR? Thx.

zymap (commented):
I understand the things you want to resolve, but it seems to introduce another issue. In some cases, increasing the recovery process number might not speed up the ledger recovery.

For example, if I have two ledgers and two auto-recovery. But the ledgers are in different child paths. Let me say /0000/0001 and /0001/0001. It will always go to replicate /0000/0001 first, so the /0001/0001 need to wait for the /0000/0001 to replicate completely. That looks like not expected.

PR #3777, opened 2023-02-13
Add test to cover ReferenceCountUtil.touch in BookieProtocol.

horizonzy:

### Motivation
In #3528, the ReadResponse#touch returns the ByteBuf#touch, the return type is ByteBuf. We should ensure the return type is self.
Add test to cover ReferenceCountUtil.touch in BookieProtocol.

Refer to the Netty usage:
https://github.com/netty/netty/blob/4475b5c5719e7e7f36f2f01e25c139bcb2b048d4/codec-http/src/main/java/io/netty/handler/codec/http/multipart/AbstractMemoryHttpData.java#L297-L301

https://github.com/netty/netty/blob/4475b5c5719e7e7f36f2f01e25c139bcb2b048d4/codec-http/src/main/java/io/netty/handler/codec/http/multipart/AbstractHttpData.java#L128-L131

The wrapper class retain() and touch() return self(); So we check after touch() and retain(), whether the return object is changed or not.

### DIFF
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/proto/BookieProtocolTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/proto/BookieProtocolTest.java
@@ -54,6 +54,7 @@ public class BookieProtocolTest {
         BookieProtocol.ParsedAddRequest parsedAddRequest = BookieProtocol.ParsedAddRequest.create(
                 BookieProtocol.CURRENT_PROTOCOL_VERSION, 0, 0, (short) 1, "test".getBytes(UTF_8), byteBuf);
         assertEquals(ReferenceCountUtil.touch(parsedAddRequest).hashCode(), parsedAddRequest.hashCode());
+        parsedAddRequest.release();
         parsedAddRequest.recycle();
         byteBuf.release();
     }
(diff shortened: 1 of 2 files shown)


horizonzy:
rerun failure checks

horizonzy:
rerun failure checks

wenbingshen:
rerun failure checks

bookkeeper-server/src/test/java/org/apache/bookkeeper/proto/BookieProtocolTest.java
```diff
@@ -0,0 +1,124 @@
...
+    @Test
+    public void testAddRequestTouch() {
+        byte[] data = new byte[4];
+        ByteBuf byteBuf = Unpooled.wrappedBuffer(data);
+
+        ByteBufList byteBufList = ByteBufList.get(byteBuf.retainedSlice());
+
+        BookieProtocol.AddRequest addRequest = BookieProtocol.AddRequest.create(BookieProtocol.CURRENT_PROTOCOL_VERSION,
+                0, 0, (short) 1, "test".getBytes(UTF_8), byteBufList);
+        assertEquals(ReferenceCountUtil.touch(addRequest).hashCode(), addRequest.hashCode());
```
nicoloboschi, line 44: we should verify that the underlying `ByteBuf.retain/touch` is called, not the returned object. It can be easily addressed using `Mockito.verify()`

ReferenceCountUtil methods don't change the object you pass so these assertions aren't useful

nicoloboschi (changes requested):

PR #3721, opened 2022-12-27
New ensemble choose different rack first.

horizonzy:
Descriptions of the changes in this PR:
Fixes #3720

### Motivation
When new ensemble, choose the different rack node first, if there are no more different rack nodes, step down the origin logic to pick nodes.

PR #3779, opened 2023-02-14
Add logs for ensemble select failed

hangc0276:
### Motivation
We have 3 bookies in the same rack, and configured `E = 2, W = 2, A = 2`. When one bookie restarted, we found the ledger select bookie from the same rack failed in the ensemble change replacing the failed bookie step.

Due to there being no log information in `selectRandomFromRack`, it's hard to debug the root cause.
https://github.com/apache/bookkeeper/blob/02e64a4b97e03afc9993ab227f82a5956965c03f/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java#L630-L679

### Modification
Add one warn log in `selectRandomFromRack` when selecting a new bookie node failed.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
@@ -675,6 +675,8 @@ public class RackawareEnsemblePlacementPolicyImpl extends TopologyAwareEnsembleP
             }
             return bn;
         }
+        LOG.warn("Failed to select bookie node from path: {}, leaves: {}, exclude Bookies: {}, ensemble: {}",
+            netPath, leaves, excludeBookies, ensemble);
         throw new BKNotEnoughBookiesException();
     }
 
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
@@ -519,7 +519,7 @@ public class RackawareEnsemblePlacementPolicyImpl extends TopologyAwareEnsembleP
                         networkLoc, excludeBookies);
                 throw e;
             }
-            LOG.warn("Failed to choose a bookie from {} : leaves {}, excluded {}, "
+            LOG.warn("Failed to choose a bookie from {} : leaves {}, excluded bookies {}, "
                     + "current ensemble {}, fallback to choose bookie randomly from the cluster.",
                      networkLoc, topology.getLeaves(networkLoc), excludeBookies, ensemble);
             // randomly choose one from whole cluster, ignore the provided predicate.
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
@@ -519,7 +519,8 @@ public class RackawareEnsemblePlacementPolicyImpl extends TopologyAwareEnsembleP
                         networkLoc, excludeBookies);
                 throw e;
             }
-            LOG.warn("Failed to choose a bookie from {} : leaves {}, excluded bookies {}, "
+            LOG.warn("Failed to choose a bookie from network location {}, "
+                    + "the bookies in the network location are {}, excluded bookies {}, "
                     + "current ensemble {}, fallback to choose bookie randomly from the cluster.",
                      networkLoc, topology.getLeaves(networkLoc), excludeBookies, ensemble);
             // randomly choose one from whole cluster, ignore the provided predicate.
@@ -544,7 +545,8 @@ public class RackawareEnsemblePlacementPolicyImpl extends TopologyAwareEnsembleP
              * the whole cluster and exclude the racks specified at
              * <tt>excludeRacks</tt>.
              */
-            LOG.warn("Failed to choose a bookie node from {} : leaves {}, exclude Bookies {}, "
+            LOG.warn("Failed to choose a bookie node from network location {}, "
+                    + "the bookies in the network location are {}, excluded bookies {}, "
                     + "current ensemble {}, fallback to choose bookie randomly from the cluster.",
                 networkLoc, topology.getLeaves(networkLoc), excludeBookies, ensemble);
             return selectFromNetworkLocation(excludeRacks, excludeBookies, predicate, ensemble, fallbackToRandom);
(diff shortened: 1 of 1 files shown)


hangc0276:
@horizonzy Please help take a look at this Pr, thanks.

horizonzy (commented):
After https://github.com/apache/bookkeeper/pull/3721, It may invoke RackawareEnsemblePlacementPolicyImpl#selectRandomFromRack many times. 
It may a little noise.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
```diff
@@ -675,6 +675,8 @@ protected BookieNode selectRandomFromRack(String netPath, Set<Node> excludeBooki
             }
             return bn;
         }
+        LOG.warn("Failed to select bookie node from path: {}, leaves: {}, exclude Bookies: {}, ensemble: {}",
```
asafm, line 678: I wouldn't log that deep.

1st caller to this method is:
```java
// first attempt to select one from local rack
        try {
            return selectRandomFromRack(networkLoc, excludeBookies, predicate, ensemble);
        } catch (BKNotEnoughBookiesException e) {
            /*
             * there is no enough bookie from local rack, select bookies from
             * the whole cluster and exclude the racks specified at
             * <tt>excludeRacks</tt>.
             */
            return selectFromNetworkLocation(excludeRacks, excludeBookies, predicate, ensemble, fallbackToRandom);
        }
```
For this caller, it's not a WARN yet, since we may still choose a bookie another rack which satisfies rack awareness.
We can print INFO in the catch here.

2nd caller:
```java
        // select one from local rack
        try {
            return selectRandomFromRack(networkLoc, excludeBookies, predicate, ensemble);
        } catch (BKNotEnoughBookiesException e) {
            if (!fallbackToRandom) {
                LOG.error(
                        "Failed to choose a bookie from {} : "
                                + "excluded {}, enforceMinNumRacksPerWriteQuorum is enabled so giving up.",
                        networkLoc, excludeBookies);
                throw e;
            }
            LOG.warn("Failed to choose a bookie from {} : "
                     + "excluded {}, fallback to choose bookie randomly from the cluster.",
                     networkLoc, excludeBookies);
            // randomly choose one from whole cluster, ignore the provided predicate.
            return selectRandom(1, excludeBookies, predicate, ensemble).get(0);
```
Here we have the error or WARN anyway, so no need to add any log here.

Since there are so many ways to this method to throw an exception, I would add a reason message at each throw, and include it in the logs printed which receive it.

hangc0276:
> After #3721, It may invoke RackawareEnsemblePlacementPolicyImpl#selectRandomFromRack many times. It may a little noise.

@horizonzy Thanks for your review, I moved the log out of the method and put it in the catch block

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
hangc0276: @asafm Good point, thanks for your suggestion. I moved the log out of the method and put it in the catch block. Please help take a look again, thanks a lot.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
```diff
@@ -519,9 +519,9 @@ public BookieNode selectFromNetworkLocation(
                         networkLoc, excludeBookies);
                 throw e;
             }
-            LOG.warn("Failed to choose a bookie from {} : "
-                     + "excluded {}, fallback to choose bookie randomly from the cluster.",
-                     networkLoc, excludeBookies);
+            LOG.warn("Failed to choose a bookie from {} : leaves {}, excluded bookies {}, "
```
horizonzy, line 522: The `leaves` may confuse the user.

Failed to choose a bookie from {} : leaves -> Failed to choose a bookie from network location {}, the network location corresponding bookies  {}

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
```diff
@@ -544,6 +544,9 @@ public BookieNode selectFromNetworkLocation(String networkLoc,
              * the whole cluster and exclude the racks specified at
              * <tt>excludeRacks</tt>.
              */
+            LOG.warn("Failed to choose a bookie node from {} : leaves {}, exclude Bookies {}, "
```
horizonzy, line 547: The `leaves` may confuse the user.

Failed to choose a bookie from {} : leaves -> Failed to choose a bookie from network location {}, the network location corresponding bookies  {}

hangc0276: make sense

horizonzy (approved):
LGTM.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
asafm: What do you say about :
> Since there are so many ways to this method to throw an exception, I would add a reason message at each throw, and include it in the logs printed which receive it.

Since `selectRandomFromRack` has several ways to fail and throw an exception, how about we include an error message in the exception and log it as well?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
```diff
@@ -544,6 +545,10 @@ public BookieNode selectFromNetworkLocation(String networkLoc,
              * the whole cluster and exclude the racks specified at
              * <tt>excludeRacks</tt>.
              */
+            LOG.warn("Failed to choose a bookie node from network location {}, "
```
asafm, line 548: Don't you think here it is actually an INFO?
What if we have 5 racks. If we failed choosing from same rack (say rack 1) and we can't choose from rack 2 and 3 since they already have other copies, we can still choose from rack 4, and it's ok, no?

hangc0276: Yes, it still can choose from rack 4. But if we only have 3 racks (rack1, rack2, and rack3), and we failed to choose any bookie nodes from rack1, it will randomly choose one from rack2 or rack3, which will lead to 3 replicas located on 2 racks. 

IMO, we'd better use the `WARN` level, because it will choose one bookie node from the whole bookie cluster randomly and it has the risk of the chosen bookie being located on the same rack with existing replicas.

asafm (approved):

StevenLuMT (approved):
LGTM

zymap (approved):

PR #3784, opened 2023-02-18
Made PendingAddOp thread safe

merlimat:
### Motivation

`PendingAddOp` is not thread-safe and the BK client instead follows a pattern of "jumping" to a particular thread for both the requests and the callback path. 

This is in practice much more expensive than acquiring a mutex since in most cases the caller will have already serialized the calls to `ledger.append()`. 

With this change, we can avoid 2 context switches and the very big contention on the executor queues used between the threads.


hangc0276 (approved):
Good job!

hangc0276:
Hi @merlimat, I found one test failed related to this PR. https://github.com/apache/bookkeeper/issues/3805

Issue #3734, opened 2023-01-11
RocksDB: segfault in org.rocksdb.WriteBatch::delete called from org.apache.bookkeeper.bookie.storage.ldb.EntryLocationIndex#removeOffsetFromDeletedLedgers


dlg99:
**BUG REPORT**

***Describe the bug***

A prod server crashed because of the segfault in the RocksDB. 
Unfortunately, the crash dump is lost. Logs point to org.rocksdb.WriteBatch::delete called from org.apache.bookkeeper.bookie.storage.ldb.EntryLocationIndex#removeOffsetFromDeletedLedgers

It is hard to pinpoint the issue / match it to a specific rocksDB bug without the crash dump. I cannot repro the problem in unit test and even if I repro it I won't know if that's the exact problem.

So far the crash happened only one time, roughly the timing and code correlate with upgrade to a (internal) version (BK 4.14.x uses rocksdb 6.16.4) with change bringing the use of range deletion w/rocksDB https://github.com/apache/bookkeeper/pull/3653 

After some research I have a gut feeling that the problem is related to fix of "a bug in iterator refresh which could segfault for DeleteRange users" https://github.com/facebook/rocksdb/pull/10739
This should be included into RocksDB 7.8.0, I do not see it in 6.x versions. Instead i see 6.29.0 has "Added API warning against using Iterator::Refresh() together with DB::DeleteRange(), which are incompatible and have always risked causing the refreshed iterator to return incorrect results." 

With that said, we have the following options:

1. do nothing, hope the problem is extremely rare. Collect more info if/when it reoccurs. 
2. revert https://github.com/apache/bookkeeper/pull/3653 cc @hangc0276 - do you have any perf test results that show how much this PR improved performance to help decide why we may want to not revert this?
3. upgrade RocksDB to 7.8.0+. Upgrade to 7.x as attempted at https://github.com/apache/bookkeeper/pull/3568 but will need more work for backwards compat tests (at least) assuming there is no data incompatibility. I see some changes around dropping some data format options that may affect downgrade, so there is a risk. 
4. Upgrade to the RocksDB 6.29.5. It sounds like option 1 with extra steps but there are multiple fixes between 6.16.4 (or even 6.29.4.1 used by BK 4.16) and 6.29.5 that might reduce chances of the problem to surface, e.g.:

```
Fixed a bug caused by race among flush, incoming writes and taking snapshots. Queries to snapshots created with these race condition can return incorrect result, e.g. resurfacing deleted data.
[...]

PR #3793, opened 2023-02-21
Make RrocksDB checksum type configurable

hangc0276:
### Motivation
Fix https://github.com/apache/bookkeeper/issues/3734#issuecomment-1407626941

We have two rocksDB tables, one for the ledger index, and another for the entry log location.
 - ledger index RocksDB table: Use the default table option, and the checksum is `kCRC32c`
 - entry log location RocksDb table: Use configured table option, and the checksum is `kxxHash` 

When we upgrade the RocksDB version from 6.10.2 to 7.9.2, the new RocksDB version's default table checksum has changed from `kCRC32c` to `kXXH3`, and `kXXH3` only supported since RocksDB 6.27. The RocksDB version rollback to 6.10.2 will be failed due to RocksDB 6.10.2 doesn't support the `kXXH3` checksum type.

### Modifications
In this PR, I make the RocksDB checksum type configurable. But there is one change that will change the ledger index RocksDB table's checksum type from the default `kCRC32c` to `kxxHash`. I have tested the compatibility of the two checksum types in and between multiple RocksDB versions, it works fine.

After setting the two RocksDB table's checksum type to `kxxHash`, the RocksDB's version upgraded from 6.10.2 to 7.9.2, and rolling back to 6.10.2 works fine.

### More to discuss
When writing the unit test to read the table checksum type from RocksDB configuration files, it failed. I found the related issue on RocksDB: https://github.com/facebook/rocksdb/issues/5297
The related PR: https://github.com/facebook/rocksdb/pull/10826

It means we still can't load RocksDB table options from configuration files. Maybe I missed some parts about reading RocksDB table options from the configuration file. 

If this issue exists, we do **NOT** recommend users configure RocksDB configurations through configuration files.

@merlimat @eolivelli @dlg99 Please help take a look, thanks.

### DIFF
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageRocksDBTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageRocksDBTest.java
@@ -84,7 +84,6 @@ public class KeyValueStorageRocksDBTest {
     @Test
     public void testReadChecksumTypeFromBookieConfiguration() throws Exception {
         ServerConfiguration configuration = new ServerConfiguration();
-        configuration.setProperty("dbStorage_rocksDB_checksum_type", "kxxHash");
         configuration.setEntryLocationRocksdbConf("entry_location_rocksdb.conf");
         File tmpDir = Files.createTempDirectory("bk-kv-rocksdbtest-conf").toFile();
         Files.createDirectory(Paths.get(tmpDir.toString(), "subDir"));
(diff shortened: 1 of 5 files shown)


merlimat (approved):

zymap (approved):

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageRocksDBTest.java
```diff
@@ -78,4 +80,38 @@ public void testRocksDBInitiateWithConfigurationFile() throws Exception {
         assertEquals(1, familyOptions.maxWriteBufferNumber());
         rocksDB.close();
     }
+
+    @Test
+    public void testReadChecksumTypeFromBookieConfiguration() throws Exception {
+        ServerConfiguration configuration = new ServerConfiguration();
+        configuration.setProperty("dbStorage_rocksDB_checksum_type", "kxxHash");
```
horizonzy, line 87: We can remove  `configuration.setProperty("dbStorage_rocksDB_checksum_type", "kxxHash");` to verify the default value is `kxxHash`.

hangc0276: done

PR #3374, opened 2022-06-28
Shut down ReplicationWorker and Auditor on non-recoverable ZK error

dlg99:
Descriptions of the changes in this PR:

Shut down Replication Worker and Auditor on non-recoverable ZK error

### Motivation

Some errors require one to re-create Zk client.
Currently BK and underlying components cannot do that transparently.
When running AutoRecovery as a separate service it does not seem to crash in such cases and keeps on running while unable to do anything useful and requires operator restarting the service manually.

One can see such log messages like 
 ```
ReplicationWorker] ERROR org.apache.bookkeeper.replication.ReplicationWorker - UnavailableException while replicating fragments
org.apache.bookkeeper.replication.ReplicationException$UnavailableException: Error contacting zookeeper
	at org.apache.bookkeeper.meta.ZkLedgerUnderreplicationManager.getLedgerToRereplicate(ZkLedgerUnderreplicationManager.java:610) ~[com.datastax.oss-bookkeeper-server-4.14.5.1.0.0.jar:4.14.5.1.0.0]
	at org.apache.bookkeeper.replication.ReplicationWorker.rereplicate(ReplicationWorker.java:264) ~[com.datastax.oss-bookkeeper-server-4.14.5.1.0.0.jar:4.14.5.1.0.0]
	at org.apache.bookkeeper.replication.ReplicationWorker.run(ReplicationWorker.java:230) [com.datastax.oss-bookkeeper-server-4.14.5.1.0.0.jar:4.14.5.1.0.0]
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30) [io.netty-netty-common-4.1.76.Final.jar:4.1.76.Final]
	at java.lang.Thread.run(Thread.java:829) [?:?]
Caused by: org.apache.zookeeper.KeeperException$ConnectionLossException: KeeperErrorCode = ConnectionLoss for /ledgers/underreplication/ledgers/0000/0001/0ebb
	at org.apache.zookeeper.KeeperException.create(KeeperException.java:102) ~[com.datastax.oss-pulsar-zookeeper-2.7.2.1.1.33.jar:2.7.2.1.1.33]
	at org.apache.zookeeper.KeeperException.create(KeeperException.java:54) ~[com.datastax.oss-pulsar-zookeeper-2.7.2.1.1.33.jar:2.7.2.1.1.33]
[...]

PR #3799, opened 2023-02-22
Support retry logic for auto recovery

wolfstudy:

### Motivation

In the current implementation, when the Session between Bookie and ZK expires, Auto Recovery does not support automatic reconnection. However, the bookie server itself supports reconnection logic, so when Auto Recovery and Bookie Server are deployed together, once the Session expires with ZK, Auto Recovery will automatically stop and cause the entire Bookie service to go down.

However, in a production environment, session expiration between Bookie and ZK itself is a high-frequency operation, such as the following scenario:

From the ZK side, we can see that the current ZK zxid has overflowed, triggering XidRolloverException:

image
image

When the above situation occurs, we can see that the Session between Bookie and ZK expires:

image
image

Since Auto Recovery has no reconnection logic, the entire Bookie service is down.

### Changes

- Add retry logic for auto recovery

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/AuditorElector.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/AuditorElector.java
@@ -40,6 +40,7 @@ import org.apache.bookkeeper.replication.ReplicationException.UnavailableExcepti
 import org.apache.bookkeeper.stats.NullStatsLogger;
 import org.apache.bookkeeper.stats.StatsLogger;
 import org.apache.bookkeeper.stats.annotations.StatsDoc;
+import org.apache.zookeeper.KeeperException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -191,8 +192,14 @@ public class AuditorElector {
                         Thread.currentThread().interrupt();
                         submitShutdownTask();
                     } catch (Exception e) {
-                        LOG.error("Exception while performing auditor election", e);
-                        submitShutdownTask();
+                        if (e.getCause() instanceof KeeperException.ConnectionLossException
+                                || e.getCause() instanceof KeeperException.SessionExpiredException) {
+                            LOG.error("ConnectionLossException while performing auditor election, we will retry", e);
+                            submitElectionTask();
+                        } else {
+                            LOG.error("Exception while performing auditor election", e);
+                            submitShutdownTask();
+                        }
                     }
                 }
             };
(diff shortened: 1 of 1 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/AuditorElector.java
```diff
@@ -191,8 +192,14 @@ public void run() {
                         Thread.currentThread().interrupt();
                         submitShutdownTask();
                     } catch (Exception e) {
-                        LOG.error("Exception while performing auditor election", e);
-                        submitShutdownTask();
+                        if (e.getCause() instanceof KeeperException.ConnectionLossException
```
zymap, line 195: It's better to retry with a delay time I think.

dlg99: IIRC ConnectionLoss is a non-recoverable exception.
To handle it properly, one has to re-create ZK client and deal with the watchers/cached state. Ephemeral nodes expire as result of such error too.
Re-initializing the client at random place when error happens is a rather big chunk of work IIRC and it easier to bounce the service in this case, hence https://github.com/apache/bookkeeper/pull/3374

dlg99 (changes requested):
1. Please see https://github.com/apache/bookkeeper/pull/3374 and https://cwiki.apache.org/confluence/display/ZOOKEEPER/ErrorHandling
2. IIRC ConnectionLoss (or SessionExpiredEx or both) is a non-recoverable exception. To handle it properly, one has to re-create ZK client and deal with the watchers/cached state. Ephemeral nodes expire as result of such error too. Re-initializing the client at random place when error happens is a rather big chunk of work IIRC and it easier to bounce the service in this case, hence https://github.com/apache/bookkeeper/pull/3374
3. Change without unit tests/repro

If you disagree with my assessment please start a discussion in the mailing list.

@eolivelli has deeper knowledge of zookeeper, let's double-check with him too.

It is not a good idea in general to run the autorecovery as part of the bookie in prod. The best practice is to run it as a separate service on either another host (which allows one to have fewer AR service nodes than bookies) or alongside with the bookie as a separate process. Running AR separately also helps with such issues like AR activity resulting in a more frequent java GC cycles. HTH.

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/AuditorElector.java
zymap: @dlg99 I saw it said connection loss is a recoverable exception.
```
Recoverable errors: the disconnected event, connection timed out, and the connection loss exception are examples of recoverable errors
```

And the code
```
       public static boolean isRecoverableException(int rc) {
            return KeeperException.Code.CONNECTIONLOSS.intValue() == rc
                    || KeeperException.Code.OPERATIONTIMEOUT.intValue() == rc
                    || KeeperException.Code.SESSIONMOVED.intValue() == rc
                    || KeeperException.Code.SESSIONEXPIRED.intValue() == rc;
        }
```

PR #3810, opened 2023-02-26
Improved efficiency in DigestManager.verify()

merlimat:
### Motivation

In `DigestManager.verifyDigestAndReturnData()` there are a couple of simple improvements: 

 1. For the temp buffer where to store the digest, do not acquire a buffer from the pool each time. Instead, use a thread-local buffer.
 2. Instead of taking slices() at each time, extend the interface to take a "offset, length" pair and save the allocations.
 

#### Before

```
Benchmark                            (entrySize)   Mode  Cnt   Score   Error   Units
DigestManagerBenchmark.verifyDigest           64  thrpt    3  13.450 ± 3.634  ops/us
DigestManagerBenchmark.verifyDigest         1024  thrpt    3   7.908 ± 2.637  ops/us
DigestManagerBenchmark.verifyDigest         4086  thrpt    3   3.233 ± 0.882  ops/us
DigestManagerBenchmark.verifyDigest         8192  thrpt    3   1.846 ± 0.047  ops/us
```

Allocations: 230 bytes per operation

#### After

```
Benchmark                            (entrySize)   Mode  Cnt   Score   Error   Units
DigestManagerBenchmark.verifyDigest           64  thrpt    3  46.312 ± 7.414  ops/us
DigestManagerBenchmark.verifyDigest         1024  thrpt    3  13.379 ± 1.069  ops/us
DigestManagerBenchmark.verifyDigest         4086  thrpt    3   3.787 ± 0.059  ops/us
DigestManagerBenchmark.verifyDigest         8192  thrpt    3   1.956 ± 0.052  ops/us
```

Allocations: 64 bytes per operation

Note: these remaining 64 bytes are only due to the Java9 reflection access. It would not be present on the JNI-based CRC32c backend (though that does not work on Mac M1).

image

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/DigestManager.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/DigestManager.java
@@ -19,9 +19,12 @@ package org.apache.bookkeeper.proto.checksum;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufAllocator;
+import io.netty.buffer.ByteBufUtil;
 import io.netty.buffer.CompositeByteBuf;
+import io.netty.buffer.PooledByteBufAllocator;
 import io.netty.buffer.Unpooled;
 import io.netty.util.ReferenceCountUtil;
+import io.netty.util.concurrent.FastThreadLocal;
 import java.security.GeneralSecurityException;
 import java.security.NoSuchAlgorithmException;
 import org.apache.bookkeeper.client.BKException.BKDigestMatchException;
@@ -51,10 +54,10 @@ public abstract class DigestManager {
     abstract int getMacCodeLength();
 
     void update(byte[] data) {
-        update(Unpooled.wrappedBuffer(data, 0, data.length));
+        update(Unpooled.wrappedBuffer(data), 0, data.length);
     }
 
-    abstract void update(ByteBuf buffer);
+    abstract void update(ByteBuf buffer, int offset, int len);
 
     abstract void populateValueAndReset(ByteBuf buffer);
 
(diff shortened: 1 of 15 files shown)


hangc0276:
Please rebase the master @merlimat

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/DigestManager.java
```diff
@@ -158,6 +161,18 @@ private void verifyDigest(long entryId, ByteBuf dataReceived) throws BKDigestMat
         verifyDigest(entryId, dataReceived, false);
     }
 
+    private static final FastThreadLocal<ByteBuf> DIGEST_BUFFER = new FastThreadLocal<ByteBuf>() {
+        @Override
+        protected ByteBuf initialValue() throws Exception {
+            return PooledByteBufAllocator.DEFAULT.directBuffer(1024);
```
hangc0276, line 167: The required ByteBuf size for each digest manager:
CRC32CDigestManager -> 4
CRC32DigestManager -> 8
DummyDigestManager -> 0
MacDigestManager -> 20
Do we need to allocate 1024 bytes?

hangc0276 (approved):

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/DigestManager.java
merlimat: Just in case. Though this is only used for MAC at this point, and one per thread. It shouldn't be any noticeable waste.

merlimat: I actually was referring to my next change in the DigestManager :)

Issue #3812, opened 2023-02-27
[Bug] Always one orphan ledger is created


Reidddddd:
**BUG REPORT**

***Describe the bug***

Each time after a ledger is created, there will be one more orphan ledger with size 0 created as well, which can't be tracked.

***To Reproduce***

1. list ledgers before creation
2. DistributedLogManager#openLogWriter
3. LogWriter#write(LogRecord)
4. LogWriter#close
5. list ledgers. Compared with 1. you will find 2 more ledgers

***Expected behavior***

Should be only 1 ledger

***Screenshots***

N/A

***Root Cause***

The root cause lies in the **BKLogWriteHandler#doStartLogSegment**
line 571: it calls **logSegmentAllocator.allocate();**, which is, in fact, SimpleLedgerAllocator#allocate.
     then inside SimpleLedgerAllocator#allocate, whenComplete -> onSuccess, it will mark phase: **ALLOCATED**,
     but just one line after, inside **completeAllocation**, it adds one ZKSetVersionTxn (together with SimpleLedgerAllocator as listner) and marks the phase back to **HANDING_OVER**
So here now we back to **BKLogWriteHandler#doStartLogSegment**,
line 579, started a ZK Transactiion
line 661, txn.execute -> ZKTransaction#processResult -> ZKSetVersionTxn#commitOpResult -> listener#onCommit() (listener is SimpleLedgerAllocator, so it is SimpleLedgerAllocator#onCommit) --> SimpleLedgerAllocator#confirmObtain.
In side, SimpleLedgerAllocator#confirmObtain, as the phase is changed back to **HANDING_OVER**, there **shouldAllocate = true;**, and this is the where the orphan ledger comes from

PR #3813, opened 2023-02-27
[Bug] Always one orphan ledger is created

Reidddddd:
Issue #3812

Fix bug and one UT for verification

### Motivation

Fix the bug

### Changes

To avoid the ledger being created twice, add a state check before setting a new state

### DIFF
--- a/stream/distributedlog/core/src/main/java/org/apache/distributedlog/bk/SimpleLedgerAllocator.java
+++ b/stream/distributedlog/core/src/main/java/org/apache/distributedlog/bk/SimpleLedgerAllocator.java
@@ -382,9 +382,12 @@ public class SimpleLedgerAllocator implements LedgerAllocator, FutureEventListen
                 allocatePath, DistributedLogConstants.EMPTY_BYTES, (int) version.getLongVersion());
         ZKVersionedSetOp commitOp = new ZKVersionedSetOp(zkSetDataOp, this);
         tryObtainTxn.addOp(commitOp);
-        if (this.phase != Phase.ALLOCATED) {
-            setPhase(Phase.HANDING_OVER);
-        }
+        if (this.phase == Phase.ALLOCATED) {
+            // please check issue: https://github.com/apache/bookkeeper/issues/3812
+            allocatePromise.complete(lh);
+            return;
+        } // otherwise, it is ALLOCATING phase
+        setPhase(Phase.HANDING_OVER);
         allocatePromise.complete(lh);
     }
 
--- a/stream/distributedlog/core/src/main/java/org/apache/distributedlog/bk/SimpleLedgerAllocator.java
+++ b/stream/distributedlog/core/src/main/java/org/apache/distributedlog/bk/SimpleLedgerAllocator.java
@@ -382,7 +382,9 @@ public class SimpleLedgerAllocator implements LedgerAllocator, FutureEventListen
                 allocatePath, DistributedLogConstants.EMPTY_BYTES, (int) version.getLongVersion());
         ZKVersionedSetOp commitOp = new ZKVersionedSetOp(zkSetDataOp, this);
         tryObtainTxn.addOp(commitOp);
-        setPhase(Phase.HANDING_OVER);
+        if (this.phase != Phase.ALLOCATED) {
+            setPhase(Phase.HANDING_OVER);
+        }
         allocatePromise.complete(lh);
     }
 
--- a/stream/distributedlog/core/src/main/java/org/apache/distributedlog/bk/SimpleLedgerAllocator.java
+++ b/stream/distributedlog/core/src/main/java/org/apache/distributedlog/bk/SimpleLedgerAllocator.java
@@ -310,7 +310,9 @@ public class SimpleLedgerAllocator implements LedgerAllocator, FutureEventListen
                 tryObtainTxn = null;
                 tryObtainListener = null;
                 // mark flag to issue an allocation request
-                shouldAllocate = true;
+                if (lhToNotify == null) {
+                    shouldAllocate = true;
+                }
             }
         }
         if (null != listenerToNotify && null != lhToNotify) {
@@ -382,11 +384,6 @@ public class SimpleLedgerAllocator implements LedgerAllocator, FutureEventListen
                 allocatePath, DistributedLogConstants.EMPTY_BYTES, (int) version.getLongVersion());
         ZKVersionedSetOp commitOp = new ZKVersionedSetOp(zkSetDataOp, this);
         tryObtainTxn.addOp(commitOp);
-        if (this.phase == Phase.ALLOCATED) {
-            // please check issue: https://github.com/apache/bookkeeper/issues/3812
-            allocatePromise.complete(lh);
-            return;
-        } // otherwise, it is ALLOCATING phase
         setPhase(Phase.HANDING_OVER);
         allocatePromise.complete(lh);
     }
--- a/stream/distributedlog/core/src/test/java/org/apache/distributedlog/TestBKLogWriteHandler.java
+++ b/stream/distributedlog/core/src/test/java/org/apache/distributedlog/TestBKLogWriteHandler.java
@@ -23,8 +23,13 @@ import static org.junit.Assert.fail;
 
 import java.io.IOException;
 import java.net.URI;
+import java.nio.charset.StandardCharsets;
+import java.util.HashSet;
+import java.util.Set;
+import org.apache.bookkeeper.client.BookKeeperAdmin;
 import org.apache.distributedlog.api.AsyncLogWriter;
 import org.apache.distributedlog.api.DistributedLogManager;
+import org.apache.distributedlog.api.LogWriter;
 import org.apache.distributedlog.api.namespace.NamespaceBuilder;
 import org.apache.distributedlog.bk.LedgerAllocator;
 import org.apache.distributedlog.bk.LedgerAllocatorPool;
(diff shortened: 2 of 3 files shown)


Reidddddd:
I described the root cause in https://github.com/apache/bookkeeper/issues/3812

stream/distributedlog/core/src/main/java/org/apache/distributedlog/bk/SimpleLedgerAllocator.java
```diff
@@ -382,7 +382,9 @@ private synchronized void completeAllocation(LedgerHandle lh) {
                 allocatePath, DistributedLogConstants.EMPTY_BYTES, (int) version.getLongVersion());
         ZKVersionedSetOp commitOp = new ZKVersionedSetOp(zkSetDataOp, this);
         tryObtainTxn.addOp(commitOp);
-        setPhase(Phase.HANDING_OVER);
+        if (this.phase != Phase.ALLOCATED) {
```
eolivelli, line 385: It is common to get here when phase is ALLOCATED
see here

https://github.com/apache/bookkeeper/blob/e6722848e0a599f215e68e1340196c9dfe8e906e/stream/distributedlog/core/src/main/java/org/apache/distributedlog/bk/SimpleLedgerAllocator.java#L432

or you can get to here from this path
https://github.com/apache/bookkeeper/blob/e6722848e0a599f215e68e1340196c9dfe8e906e/stream/distributedlog/core/src/main/java/org/apache/distributedlog/bk/SimpleLedgerAllocator.java#L287

maybe it is clearer to not use "!=" but to explicitly enumerate the "phases" in which you are allowed to transition to `HANDING_OVER` .

Could you please change the condition to a "positive" condition ?
also we could add assertions and fail in case we reach to this point in an unexpected case

if using the phase is not enough to discriminate the reason why to set HANDING_OVER maybe we could change the signature of this method and pass the expected "nextPhase" or something like that

eolivelli (changes requested):
good work.

I have left some feedback.
I am not used to this code, so I am asking a few questions in order to understand better

stream/distributedlog/core/src/main/java/org/apache/distributedlog/bk/SimpleLedgerAllocator.java
Reidddddd: > It is common to get here when phase is ALLOCATED see here
> 
> https://github.com/apache/bookkeeper/blob/e6722848e0a599f215e68e1340196c9dfe8e906e/stream/distributedlog/core/src/main/java/org/apache/distributedlog/bk/SimpleLedgerAllocator.java#L432
> 
> or you can get to here from this path
> 
> https://github.com/apache/bookkeeper/blob/e6722848e0a599f215e68e1340196c9dfe8e906e/stream/distributedlog/core/src/main/java/org/apache/distributedlog/bk/SimpleLedgerAllocator.java#L287
> 
> maybe it is clearer to not use "!=" but to explicitly enumerate the "phases" in which you are allowed to transition to `HANDING_OVER` .
> 
> Could you please change the condition to a "positive" condition ? also we could add assertions and fail in case we reach to this point in an unexpected case
> 
> if using the phase is not enough to discriminate the reason why to set HANDING_OVER maybe we could change the signature of this method and pass the expected "nextPhase" or something like that

https://github.com/apache/bookkeeper/blob/e6722848e0a599f215e68e1340196c9dfe8e906e/stream/distributedlog/core/src/main/java/org/apache/distributedlog/bk/SimpleLedgerAllocator.java#L287
Do you feel this piece is weird? ERROR, HANDLING_OVER, HANDED_OVER will lead to exception, so only ALLOCATING  and ALLOCATED will pass the check, then the **completeAllocation(allocatedLh);**, will change the ALLOCATING/ALLOCATED to HANDLING_OVER. I'm not familiar with the codes, but is the state machine buggy? Normally, I feel should be like this:
HANDLING_OVER > HANDED_OVER > ALLOCATING > ALLOCATED, all these 4 states can point to ERROR. But when i went through the codes, seems it is not, just as I mentioned.


Are you suggesting to a big changes? (mine is too simple)

Reidddddd:
New commit use positive condition check as review

Reidddddd:
I want to have a whole review about the Phase's state machine and refine it after this quick fix.
What do you think of the idea

Reidddddd:
Failed UTs seem related, working on

eolivelli (approved):
LGTM

eolivelli (changes requested):
Many tests are failing now

Failures: 
Error:  org.apache.distributedlog.TestRollLogSegments.testLastDLSNInRollingLogSegments
Error:    Run 1: TestRollLogSegments.testLastDLSNInRollingLogSegments:164
Error:    Run 2: TestRollLogSegments.testLastDLSNInRollingLogSegments:164
Error:    Run 3: TestRollLogSegments.testLastDLSNInRollingLogSegments:164
[INFO] 
Error:  org.apache.distributedlog.TestRollLogSegments.testRollingLogSegments
Error:    Run 1: TestRollLogSegments.testRollingLogSegments:297
Error:    Run 2: TestRollLogSegments.testRollingLogSegments:297
Error:    Run 3: TestRollLogSegments.testRollingLogSegments:297
[INFO] 
Error:  org.apache.distributedlog.bk.TestLedgerAllocatorPool.testConcurrentAllocation
Error:    Run 1: TestLedgerAllocatorPool.testConcurrentAllocation:309 expected:<0> but was:<5>
Error:    Run 2: TestLedgerAllocatorPool.testConcurrentAllocation:309 expected:<0> but was:<5>
Error:    Run 3: TestLedgerAllocatorPool.testConcurrentAllocation:309 expected:<0> but was:<5>
[INFO] 
Error:  Errors: 
Error:  org.apache.distributedlog.TestAppendOnlyStreamWriter.testOffsetGapAfterSegmentWriterFailure
Error:    Run 1: TestAppendOnlyStreamWriter.testOffsetGapAfterSegmentWriterFailure » ConcurrentObtain
Error:    Run 2: TestAppendOnlyStreamWriter.testOffsetGapAfterSegmentWriterFailure » ConcurrentObtain
Error:    Run 3: TestAppendOnlyStreamWriter.testOffsetGapAfterSegmentWriterFailure » ConcurrentObtain
[INFO] 
Error:  org.apache.distributedlog.bk.TestLedgerAllocator.testObtainMultipleLedgers
Error:    Run 1: TestLedgerAllocator.testObtainMultipleLedgers:383 » ConcurrentObtain Ledger ha...
Error:    Run 2: TestLedgerAllocator.testObtainMultipleLedgers:383 » ConcurrentObtain Ledger ha...
Error:    Run 3: TestLedgerAllocator.testObtainMultipleLedgers:383 » ConcurrentObtain Ledger ha...
[INFO] 
Error:  org.apache.distributedlog.bk.TestLedgerAllocatorPool.testAllocateMultipleLedgers
Error:    Run 1: TestLedgerAllocatorPool.testAllocateMultipleLedgers:257 » IO No ledger allocat...
Error:    Run 2: TestLedgerAllocatorPool.testAllocateMultipleLedgers:257 » IO No ledger allocat...
Error:    Run 3: TestLedgerAllocatorPool.testAllocateMultipleLedgers:257 » IO No ledger allocat...
[INFO] 
Error:  org.apache.distributedlog.bk.TestLedgerAllocatorPool.testRescueAllocators
Error:    Run 1: TestLedgerAllocatorPool.testRescueAllocators:194 » TestTimedOut test timed out...
Error:    Run 2: TestLedgerAllocatorPool.testRescueAllocators:194 » TestTimedOut test timed out...
Error:    Run 3: TestLedgerAllocatorPool.testRescueAllocators:194 » TestTimedOut test timed out...
[INFO] 
Warning:  Flakes: 
Warning:  org.apache.distributedlog.TestBKLogWriteHandler.testLedgerNumber
Error:    Run 1: TestBKLogWriteHandler.testLedgerNumber:130 expected:<0> but was:<1>

Reidddddd:
This fix affects the segments rolling function.

Reidddddd:
Still working, the fix probably should be more complex than i thought, im still reading source codes.. 

BTW, it is not only about DL, since issue lies in BKLogWriteHandler, i think BK should have this orphan ledger issue as well..

Reidddddd:
Let's wait for the tests results before i explain the fix

Reidddddd:
org.apache.bookkeeper.client.HandleFailuresTest.testHandleFailureBookieNotInWriteSet

can pass on local, flaky?

Reidddddd:
The new fix is about the null check of `lhToNotify`

it is assigned allocatedLh(LedgerHandle), and allocatedLh is only assigned in SimpleLedgerAllocator#completeAllocation, which means ledger allocation is completed. So if this is null, it means allocation doesn't happen, `shouldAllocate` should be true. If it is not null, set false and avoid orphan ledger allocation

Reidddddd:
The problem of previous fix was, if we setPhase(Allocated), the segment rolling will not take effect as because of https://github.com/apache/bookkeeper/blob/915deb035b618a50915e438be2c2a0bc08a10323/stream/distributedlog/core/src/main/java/org/apache/distributedlog/bk/SimpleLedgerAllocator.java#L267

eolivelli (approved):
LGTM

the final solution is better

I have restarted the failed job, the failure is not related

eolivelli:
@diegosalvi @hangc0276 @nicoloboschi  @sijie  PTAL

diegosalvi:
LGTM
The latest solution is way cleaner!

dlg99 (approved):
makes sense but I am not very familiar with dlog

hangc0276 (approved):

zymap (approved):

PR #3821, opened 2023-02-27
Support update ledger metadata option bk-cli

rdhabalia:
### Motivation

Bk-CLI api supports creating ledger metadata from the metadata binary payload-file but it doesn't support updating metadata for existing ledger which will require some time during updating metadata in some operational update usecases. Therefore, support update metadata if ledger already exists by adding a flag.

### Changes

`bookkeeper shell ledgermetadata -restorefromfile` command can update ledger-metadata by passing additional `--update flag`.


eolivelli (approved):
Lgtm

merlimat (approved):

PR #3824, opened 2023-02-28
Use JNI directly for posix_fadvise

merlimat:
### Motivation

For calling `posix_fadvise`, we are using JNA which is a wrapper on top of JNI. While JNA is convenient because it's dynamically creating a stub for accessing native methods, it's also very inefficient compared to direct JNI access.

Since we already have JNI modules, we should add `posix_fadvise` there and remove the dependency on JNA.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/util/PageCacheUtil.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/util/PageCacheUtil.java
@@ -34,7 +34,7 @@ public final class PageCacheUtil {
 
     private static final int POSIX_FADV_DONTNEED = 4; /* fadvise.h */
 
-    private static boolean fADVISE_POSSIBLE = true;
+    private static boolean fadvisePossible = true;
 
     private static final NativeIO NATIVE_IO;
 
@@ -44,7 +44,7 @@ public final class PageCacheUtil {
             nativeIO = new NativeIOImpl();
         } catch (Exception e) {
             log.warn("Unable to initialize NativeIO for posix_fdavise: {}", e.getMessage());
-            fADVISE_POSSIBLE = false;
+            fadvisePossible = false;
         }
 
         NATIVE_IO = nativeIO;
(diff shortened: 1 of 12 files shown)


zymap (approved):

hangc0276 (approved):
Great job!

hangc0276:
Please fix the failed check:
```
Error:  /home/runner/work/bookkeeper/bookkeeper/bookkeeper-server/src/main/java/org/apache/bookkeeper/util/PageCacheUtil.java:37:28: Name 'FADVISE_POSSIBLE' must match pattern '^[a-z][a-zA-Z0-9]*_?$'. [StaticVariableName]
Audit done.
[INFO] There is 1 error reported by Checkstyle 9.3 with buildtools/src/main/resources/bookkeeper/checkstyle.xml ruleset.
Error:  src/main/java/org/apache/bookkeeper/util/PageCacheUtil.java:[37,28] (naming) StaticVariableName: Name 'FADVISE_POSSIBLE' must match pattern '^[a-z][a-zA-Z0-9]*_?$'.
```

codelipenghui (approved):

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/PageCacheUtil.java
```diff
@@ -18,41 +18,37 @@
...
         try {
-            Native.register("c");
-            initialized = true;
-        } catch (NoClassDefFoundError e) {
-            LOG.info("JNA not found. Native methods will be disabled.");
-        } catch (UnsatisfiedLinkError e) {
-            LOG.info("Unable to link C library. Native methods will be disabled.");
-        } catch (NoSuchMethodError e) {
-            LOG.warn("Obsolete version of JNA present; unable to register C library");
+            nativeIO = new NativeIOImpl();
```
horizonzy, line 44: We should check if `nativeIO.posix_fdavise` works or not. `new NativeIOImpl()` din't ensure `posix_fdavise` work

wenbingshen: @horizonzy I also found this when I run unit tests on mac os.

We don't need to check whether nativeIO.posix_fdavise is valid, because in use, once an error occurs, we will close posix_fdavise, so we can just check whether the library is normal, this is the error I encountered.

```java
    public static void bestEffortRemoveFromPageCache(int fd, long offset, long len) {
        if (!fadvisePossible || fd < 0) {
            return;
        }
        try {
            NATIVE_IO.posix_fadvise(fd, offset, len, POSIX_FADV_DONTNEED);
        } catch (Exception e) {
            log.warn("Failed to perform posix_fadvise: {}", e.getMessage());
            fadvisePossible = false;
        }
    }
```

PR #3830, opened 2023-03-01
Simplified the logic for ForceWriteThread after we introduced queue.drainTo()

merlimat:
### Motivation

In #3545 we have switched the `ForceWriteThread` to take advantage o `BlockingQueue.drainTo()` method for reducing contention, though the core logic of the force-write was not touched at the time.

The logic of force-write is quite complicated because it tries to group multiple force-write requests in the queue by sending a new marker and grouping them when the marker is received. This also leads to a bit of lag when there are many requests coming in and the IO is stressed, as we're waiting a bit more before issuing the fsync.

Instead, with the `drainTo()` approach we can greatly simplify the logic and maintain a strict fsync grouping: 
 1. drain all the force-write-requests available in the queue into a local array list
 2. perform the fsync
 3. update the journal log mark to the position of the last fw request
 4. trigger send-responses for all the requests
 5. go back to read from the queue

This refactoring will also enable further improvements, to optimize how the send responses are prepared, since we have now a list of responses ready to send.


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -534,11 +484,26 @@ public void run() {
                         requestsCount = 1;
                     }
 
+                    journalStats.getForceWriteQueueSize().addCount(-requestsCount);
+
+                    // Sync and mark the journal up to the position of the last entry in the batch
+                    ForceWriteRequest lastRequest = localRequests.get(requestsCount - 1);
+                    syncJournal(lastRequest);
```
zymap, line 491: Is it possible that there have two log files in the batch?

hangc0276: +1
If the localRequests queue contains multiple journal files and we only sync the lastRequest's journal file, other journal files will skip sync.

merlimat: Actually, you're correct. We need to ensure all the files are closed before the response is triggered. Fixed it.

zymap: Original logic will execute forceWrite before close, which will run bestEffortRemoveFromPageCache. If there have two different journal files in the batch, we only force write for the last file, do we need to do the force write for the another journal file? Does the `close` will do that?

hangc0276: +1
It has two bugs:
- Those non-last journal files in the batch won't be removed from OS PageCache
- That non-last journal files channel in the batch will just call `close` instead of `fore(false)`, which can't ensure the data is flushed to the disk.

hangc0276:
Please rebase the master after https://github.com/apache/bookkeeper/pull/3836 is merged to trigger the CI.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
merlimat: I was actually 100% sure that close() implied an fsync, but that's not really the case.

zymap (approved):
LGTM

hangc0276 (approved):

PR #3833, opened 2023-03-02
[improve] Enhance LedgerFragment toString method

wenbingshen:
### Motivation

LedgerFragment adds the state of replicateType, but it has not been added to the toString method. I think the replicateType should be added to the toString method to track the state of LF and troubleshoot problems.

### Changes

LedgerFragment is generally in `DATA_LOSS` or `DATA_NOT_ADHERING_PLACEMENT` state.
But it shouldn't have status when it's in detection progress or used elsewhere, so I added `NULL` status.

The following log shows the result:

replicating `DATA_LOSS` LedgerFragment
image

replicating `DATA_NOT_ADHERING_PLACEMENT` LedgerFragment
image

checking LedgerFragment in `NULL` type
image

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerFragment.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerFragment.java
@@ -229,9 +229,9 @@ public class LedgerFragment {
     @Override
     public String toString() {
         return String.format("Fragment(LedgerID: %d, FirstEntryID: %d[%d], "
-                + "LastKnownEntryID: %d[%d], Host: %s, Closed: %s)", ledgerId, firstEntryId,
+                        + "LastKnownEntryID: %d[%d], Host: %s, Closed: %s, Type: %s)", ledgerId, firstEntryId,
                 getFirstStoredEntryId(), lastKnownEntryId, getLastStoredEntryId(),
-                getAddresses(), isLedgerClosed);
+                getAddresses(), isLedgerClosed, replicateType);
     }
 
     /**


wenbingshen:
@horizonzy @hangc0276 @zymap Could you help take a look? Thanks.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerFragment.java
```diff
@@ -229,15 +229,16 @@ public void setReplicateType(ReplicateType replicateType) {
...
                 getFirstStoredEntryId(), lastKnownEntryId, getLastStoredEntryId(),
-                getAddresses(), isLedgerClosed);
+                getAddresses(), isLedgerClosed, replicateType);
     }
 
     /**
      * ReplicateType.
      */
     public enum ReplicateType {
+        NULL,
```
hangc0276, line 241: Would you please explain why adding the `NULL` state and what's the meaning of the `NULL` replication type? I found the all code parts are related to `DATA_LOSS` or `DATA_NOT_ADHERING_PLACEMENT`

hangc0276 (changes requested):

hangc0276:
@horizonzy Please help take a look, thanks.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerFragment.java
horizonzy: +1

wenbingshen: @hangc0276 @horizonzy Yes, when I first looked at the use of LedgerFragment, I also thought it only had two states, `DATA_LOSS` or `DATA_NOT_ADHERING_PLACEMENT`, when I read the code again, I felt something was different.

`LedgerFragmentReplicator` is used to copy and restore those copies of a ledger segment in a failed bookie. This state initialization is `DATA_LOSS`.

`ReplicationWorker` will first recover the LedgerFragment of `DATA_LOSS` state, and then recover the LedgerFragment of `DATA_NOT_ADHERING_PLACEMENT`.

When `BookKeeperAdmin` is recovering Ledger, LedgerFragment is in `DATA_LOSS` state.

The above LedgerFragment is mainly in recover or replicate, and LedgerFragment has an obvious state.

In addition to the above, there is also a LedgerFragment, which is in an unclear state and mainly occurs inside `LedgerChecker`. `AuditorCheckAllLedgersTask` and `ReplicationWorker` use `LedgerChecker` checkLedger. By default, it will check the first and last Entry of each LedgerFragment of the ledger before the check is completed. , LedgerFragment should not be in `DATA_LOSS` or `DATA_NOT_ADHERING_PLACEMENT`, because after the inspection is completed, the LedgerFragment may be normal, or the replicate may be missing; before and after the inspection is completed, we sometimes print the LedgerFragment through debug, which may cause a wrong state, so I added the `NULL` state is used to identify this situation.

https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerChecker.java#L476-L492
```java
        for (LedgerFragment r : fragments) {
            if (LOG.isDebugEnabled()) {
                LOG.debug("Checking fragment {}", r);
            }
            try {
                verifyLedgerFragment(r, allFragmentsCb, percentageOfLedgerFragmentToBeVerified);
            } catch (InvalidFragmentException ife) {
                LOG.error("Invalid fragment found : {}", r);
                allFragmentsCb.operationComplete(
                        BKException.Code.IncorrectParameterException, r);
            } catch (BKException e) {
                LOG.error("BKException when checking fragment : {}", r, e);
                LOG.error("InterruptedException when checking fragment : {}", r, e);
            }
        }
```

`NULL` may be difficult to understand, maybe we can add `INIT` before checking, and `NORMAL` state if the segment not data loss after checking?

wenbingshen: @horizonzy I understand what you mean enough now, I misunderstood earlier, I understood replicateType as a piecemeal state, as you said, it does not represent state.

Now I want to add replicateType to the toString method of LedgerFragment. When the user opens the debug logger, we can see this situation:
image

When checking a fragment, the fragment replicateType is DATA_LOSS, which may be a bit strange for users.

This is only a small ambiguity here, but I think it is not a big problem now, we keep DATA_LOSS and DATA_NOT_ADHERING_PLACEMENT enough, it does not affect the result of program execution.

horizonzy (approved):
LGTM.

wenbingshen:
rerun failure checks

wenbingshen:
rerun failure checks

hangc0276 (approved):

StevenLuMT (approved):
LGTM

PR #3835, opened 2023-03-02
Pass BookieRequestHandler instead of Channel to the request processors

merlimat:
### Motivation

This is just a preparatory refactoring change. The final goal is to streamline the send-response path from Journal to the client connections.

We're now passing the Netty Channel object as context for the requests, though that prevents us from adding more state and logic when tracking back to the original connection.

Instead, we can pass `BookieRequestHandler` where we can potentially add new fields.


hangc0276 (approved):
Nice job!

PR #3838, opened 2023-03-04
Added BatchedArrayBlockingQueue

merlimat:
### Motivation

One of the main contention points in bookies is the Journal queue. There are multiple threads posting entries to be added on the journal on the queue. 

In order to reduce contention and increase overall throughput (when there are many small entries), we can adopt a strategy of "bulk" inserting to the queue.

`BatchedArrayBlockingQueue` is a fixed size blocking queue, very similar to Java `ArrayBlockingQueue` with 2 additional methods: 
 1. `putAll(T[] array, int offset, int len)` --> Add all the items from the array in one shot, block if full
 2. `takeAll(T[] dest)` --> Take all the available items in the queue that fit into the destination array, block if empty

We use arrays here instead of List/Collection because it enables to do `System.arrayCopy()` which is way faster than a for loop and assignments.

This PR only introduces the queue, subsequent work will make use of it for the journal.

### Benchmarks

When inserting items in bulk, this queue is 10x to 20x faster than `ArrayBlockingQueue`

```
// 4 producing threads

Benchmark                                                      Mode  Cnt    Score    Error   Units
MpScQueueBenchmark.arrayBlockingQueue                         thrpt    5   46.844 ±  3.445  ops/us
MpScQueueBenchmark.batchAwareArrayBlockingQueueBatch          thrpt    5  846.931 ± 36.330  ops/us
MpScQueueBenchmark.batchAwareArrayBlockingQueueSingleEnqueue  thrpt    5   51.522 ±  4.505  ops/us

// 8 producing threads

Benchmark                                                      Mode  Cnt    Score     Error   Units
MpScQueueBenchmark.arrayBlockingQueue                         thrpt    3   43.090 ±   7.548  ops/us
MpScQueueBenchmark.batchAwareArrayBlockingQueueBatch          thrpt    3  825.470 ± 293.480  ops/us
MpScQueueBenchmark.batchAwareArrayBlockingQueueSingleEnqueue  thrpt    3   47.250 ±  16.575  ops/us


// 16 producing threads

Benchmark                                                      Mode  Cnt    Score     Error   Units
MpScQueueBenchmark.arrayBlockingQueue                         thrpt    3   44.185 ±  12.999  ops/us
MpScQueueBenchmark.batchAwareArrayBlockingQueueBatch          thrpt    3  655.073 ± 565.844  ops/us
MpScQueueBenchmark.batchAwareArrayBlockingQueueSingleEnqueue  thrpt    3   47.575 ±  37.798  ops/us
```

### DIFF
--- a/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/collections/BatchedArrayBlockingQueue.java
+++ b/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/collections/BatchedArrayBlockingQueue.java
@@ -341,7 +341,9 @@ public class BatchedArrayBlockingQueue<T>
                 if (waitForever) {
                     notEmpty.await();
                 } else {
-                    notEmpty.await(timeout, unit);
+                    if (!notEmpty.await(timeout, unit)) {
+                        return 0;
+                    }
                 }
             }
 
--- a/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/collections/BatchedArrayBlockingQueue.java
+++ b/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/collections/BatchedArrayBlockingQueue.java
@@ -333,7 +333,8 @@ public class BatchedArrayBlockingQueue<T>
         return internalTakeAll(array, false, timeout, unit);
     }
 
-    private int internalTakeAll(T[] array, boolean waitForever, long timeout, TimeUnit unit) throws InterruptedException {
+    private int internalTakeAll(T[] array, boolean waitForever, long timeout, TimeUnit unit)
+            throws InterruptedException {
         lock.lockInterruptibly();
         try {
             while (size == 0) {
--- a/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/collections/BatchedArrayBlockingQueue.java
+++ b/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/collections/BatchedArrayBlockingQueue.java
@@ -323,20 +323,25 @@ public class BatchedArrayBlockingQueue<T>
         }
     }
 
-    /**
-     * Drain the queue into an array.
-     * Wait if there are no items in the queue.
-     *
-     * @param array
-     * @return
-     * @throws InterruptedException
-     */
     @Override
     public int takeAll(T[] array) throws InterruptedException {
+        return internalTakeAll(array, true, 0, TimeUnit.SECONDS);
+    }
+
+    @Override
+    public int pollAll(T[] array, long timeout, TimeUnit unit) throws InterruptedException {
+        return internalTakeAll(array, false, timeout, unit);
+    }
+
+    private int internalTakeAll(T[] array, boolean waitForever, long timeout, TimeUnit unit) throws InterruptedException {
         lock.lockInterruptibly();
         try {
             while (size == 0) {
-                notEmpty.await();
+                if (waitForever) {
+                    notEmpty.await();
+                } else {
+                    notEmpty.await(timeout, unit);
+                }
             }
 
             int toDrain = Math.min(size, array.length);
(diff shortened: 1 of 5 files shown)


bookkeeper-common/src/main/java/org/apache/bookkeeper/common/collections/BatchedArrayBlockingQueue.java
```diff
@@ -0,0 +1,409 @@
...
+            }
+
+            enqueueOne(e);
+        } finally {
+            lock.unlock();
+        }
+    }
+
+    public int putAll(List<T> c) throws InterruptedException {
+        lock.lockInterruptibly();
```
hangc0276, line 149: Do we need to check whether the items in the list is `null`?

hangc0276 (approved):

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/collections/BatchedArrayBlockingQueue.java
```diff
@@ -0,0 +1,409 @@
...
+                data[consumerIdx] = null;
+                c.add(item);
+
+                if (++consumerIdx == capacity) {
+                    consumerIdx = 0;
+                }
+            }
+
+            this.consumerIdx = consumerIdx;
+            if (size == capacity) {
```
horizonzy, line 315: If the `maxElements` is 0, we can't trigger `notFull.signalAll();`. Shall we check the `maxElements` param?

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/collections/BatchedArrayBlockingQueue.java
```diff
@@ -0,0 +1,409 @@
...
+            }
+
+            return dequeueOne();
+        } finally {
+            lock.unlock();
+        }
+    }
+
+    @Override
+    public int remainingCapacity() {
```
horizonzy, line 288: The method `size()` is locked, and the method `remainingCapacity` should also be locked.

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/collections/BatchedArrayBlockingQueue.java
```diff
@@ -0,0 +1,409 @@
...
+                data[producerIdx] = c.get(i);
+                if (++producerIdx == capacity) {
+                    producerIdx = 0;
+                }
+            }
+
+            this.producerIdx = producerIdx;
+
+            if (size == 0) {
+                notEmpty.signalAll();
```
horizonzy, line 171: No needn't trigger `notEmpty.signalAll();` if the target collection `c` is empty.

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/collections/BatchedArrayBlockingQueue.java
```diff
@@ -0,0 +1,409 @@
...
+                notFull.await();
+            }
+
+            int availableCapacity = capacity - size;
+            int toInsert = Math.min(availableCapacity, len);
+            int producerIdx = this.producerIdx;
+
+            // First span
+            int firstSpan = Math.min(toInsert, capacity - producerIdx);
+            System.arraycopy(a, offset, data, producerIdx, firstSpan);
```
horizonzy, line 205: It may throw ArrayIndexOutOfBoundsException when the param is illegal.
```
        Integer[] integers = new Integer[] {1, 2, 3, 4, 5, 6, 7, 8, 9,10,11,12,13,14,15,16,17,18,19,20};
        queue.putAll(integers, 10, 11);
```

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/collections/BatchedArrayBlockingQueue.java
```diff
@@ -0,0 +1,409 @@
...
+
+            size -= toDrain;
+            return toDrain;
+        } finally {
+            lock.unlock();
+        }
+    }
+
+    @Override
+    public void clear() {
```
horizonzy, line 383: And the clear() may not reach the expectation, it may clear the queue.

```
        new Thread(()-> {
            Integer[] integers = new Integer[] {1, 2, 3, 4, 5, 6, 7, 8, 9,10,11,12,13,14,15,16,17,18,19,20};
            try {
                queue.putAll(integers, 0, 20);
                throw new RuntimeException(e);
            }
        }).start();
        
        
        queue.clear();
        //After clear, the queue is not empty.
        System.out.println(queue.size());
```

PR #3841, opened 2023-03-06
Fix website deploy build failed

hangc0276:
### Motivation
Website deployment build failed
https://github.com/apache/bookkeeper/actions/runs/4340797094/jobs/7579680220

It may be caused by the maven 3.9.0 bug
https://issues.apache.org/jira/browse/MNG-7679

### Modifications
Set the maven version to 3.8.7 instead of the default 3.9.0 to fix the issue.

### DIFF
--- a/.github/workflows/website-deploy.yaml
+++ b/.github/workflows/website-deploy.yaml
@@ -46,6 +46,11 @@ jobs:
           distribution: 'temurin'
           java-version: 11
 
+      - name: Set up Maven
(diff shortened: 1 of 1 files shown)


nicoloboschi (commented):
Good finding! I think it's better to use in every CI job at this point

hangc0276:
> Good finding! I think it's better to use in every CI job at this point

@nicoloboschi Yes, I have already add to other CI jobs.

nicoloboschi (approved):

PR #3842, opened 2023-03-07
Make read entry request recyclable

codelipenghui:
### Motivation

Make the read entry request object recyclable like the add request to reduce the object creation when doing catchup read from Pulsar.

### Changes

Add recycler to the read request.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadEntryProcessor.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadEntryProcessor.java
@@ -122,6 +122,7 @@ class ReadEntryProcessor extends PacketProcessorBase<ReadRequest> {
         if (LOG.isTraceEnabled()) {
             LOG.trace("Read entry rc = {} for {}", errorCode, request);
         }
+        request.recycle();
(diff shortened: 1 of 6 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadEntryProcessor.java
```diff
@@ -122,6 +122,7 @@ protected void processPacket() {
         if (LOG.isTraceEnabled()) {
             LOG.trace("Read entry rc = {} for {}", errorCode, request);
         }
+        request.recycle();
```
zymap, line 125: sendResponse still using request. Maybe we should recycle at the ReadEntryProcessor#recycle()?

zymap (commented):
We should also recycle the read request in BookieRequestProcessor#processReadRequest() if it is rejected to execution.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadEntryProcessor.java
codelipenghui: Yes, we can only recycle the request here.

PR #3843, opened 2023-03-07
Use BatchedArrayBlockingQueue in Journal

merlimat:
### Motivation

Introduced `BatchedArrayBlockingQueue` for Journal and ForceWriteThread. 

Refactored the code to take advantage of `takeAll()` and `pollAll()` methods, this simplifies the logic of doing `drainTo()`, `take()` and `poll()`.


horizonzy:
@merlimat I fixed the failed test in https://github.com/horizonzy/bookkeeper/pull/2, Please help take a look. If you have no objections, I will push the fix to your branch, thanks.

merlimat:
@horizonzy Yes, please push the fix, thanks!

hangc0276 (approved):
Nice job!

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -472,35 +472,30 @@ public void run() {
...
                     // responses
                     for (int i = 0; i < requestsCount; i++) {
-                        ForceWriteRequest req = localRequests.get(i);
-                        numReqInLastForceWrite += req.process();
+                        ForceWriteRequest req = localRequests[i];
+                        numEntriesInLastForceWrite += req.process();
                         req.recycle();
                     }
-
+                    Arrays.fill(localRequests, 0, requestsCount, null);
```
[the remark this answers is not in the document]
zymap: At least we can do it in the line 491 loop to avoid introducing another loop

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -512,16 +507,8 @@ public void run() {
                 } catch (InterruptedException e) {
                     Thread.currentThread().interrupt();
                     LOG.info("ForceWrite thread interrupted");
-                    // close is idempotent
-                    if (!localRequests.isEmpty()) {
-                        ForceWriteRequest req = localRequests.get(localRequests.size() - 1);
-                        req.shouldClose = true;
-                        req.closeFileIfNecessary();
-                    }
```
zymap, line 520: Won't this cause the journal to discard one flush?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -1197,7 +1177,13 @@ journalFormatVersionToWrite, getBufferedChannelBuilder(),
 
                 toFlush.add(qe);
                 numEntriesToFlush++;
-                qe = null;
+
+                if (localQueueEntriesIdx < localQueueEntriesLen) {
+                    qe = localQueueEntries[localQueueEntriesIdx++];
+                } else {
+                    Arrays.fill(localQueueEntries, 0, localQueueEntriesLen, null);
```
zymap, line 1184: Same question with above. At least we can do it in the same loop when you take the queue from the array.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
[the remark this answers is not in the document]
horizonzy: > Why do we need to fill the array with null? It will overwrite by the next, no?

At the before time, we read the 100 elements from the queue. the localRequests array elements will be filled from 0-100.
But the next time, if the queue only has 50 elements, the the localRequests array 0-50 will be overridden, the 51-100 elements still be the before-time elements.

The reference is strong, it's bad for GC.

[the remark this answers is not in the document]
horizonzy: > At least we can do it in the line 491 loop to avoid introducing another loop

Make sense.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
horizonzy: Here we catch InterruptedException, the InterruptedException is thrown by `forceWriteRequests.takeAll(localRequests)`.
So if it InterruptedException, that means we didn't read the data from the queue. So we don't need to handle the data.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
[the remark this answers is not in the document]
merlimat: > Why do we need to fill the array with null? It will overwrite by the next, no?

While not strictly necessary, it is mostly to make sure the object can be GCed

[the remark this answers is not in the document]
zymap: I see.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
horizonzy: @merlimat, cc

zymap: But it also interrupted the force write thread. We need to ensure the journal data is flushed into the disk, right?

horizonzy: Maybe not, only the `forceWriteRequests.takeAll(localRequests)` throw the InterruptedException. If it throw InterruptedException, we didn't have data to flush.

PR #3846, opened 2023-03-07
Streamline batch add request

hangc0276:
### Motivation
The bookie server process add-entry requests pipeline:
- Get one request from the Netty socket channel
- Choose one thread to process the written request
- Write the entry into the target ledger entry logger's write cache(memory table)
- Put the entry into the journal's pending queue
- Journal thread takes the entry from the pending queue and writes it into PageCache/Journal Disk
- Write the callback response to the Netty buffer and flush to the bookie client side.

For every add entry request, the bookie server needs to go through the above steps one by one. It will introduce a lot of thread context switch. 

We can batch the add requests according to the Netty socket channel, and write a batch of entries into the ledger entry logger and journal disk.

### Modifications
The PR will change the add requests processing pipeline into the following steps.
- Get a batch of add-entry requests from the socket channel until the socket channel is empty or reached the max capacity (default is 10_000)
- Choose one thread to process the batch of add-entry requests.
- Write the entries into the target ledger entry logger's write cache one by one
- Put the batch of entries into the journal's pending queue
- Journal thread drain a batch of entries from the pending queue and write them into PageCache/Journal disk
- Write the callback response to the Netty buffer and flush to the bookie client side.

With this change, we can save a lot of thread context switch.


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -885,6 +888,25 @@ public void logAddEntry(ByteBuf entry, boolean ackBeforeSync, WriteCallback cb,
...
+            ByteBuf entry = entries.get(i);
+            long ledgerId = entry.getLong(entry.readerIndex());
+            long entryId = entry.getLong(entry.readerIndex() + 8);
+            entry.retain();
+            reserveMemory.addAndGet(entry.readableBytes());
+            queueEntries[i] = QueueEntry.create(entry, ackBeforeSync, ledgerId, entryId, cb, ctx,
+                MathUtils.nowInNano(), journalStats.getJournalAddEntryStats(), callbackTime);
+        }
+
+        memoryLimitController.releaseMemory(reserveMemory.get());
```
zymap, line 905: reserveMemory?

bookkeeper-server/src/main/java/org/apache/bookkeeper/processor/RequestProcessor.java
```diff
@@ -47,4 +49,11 @@ public interface RequestProcessor extends AutoCloseable {
      * Flush any pending response staged on all the client connections.
      */
     void flushPendingResponses();
+
+    /**
+     * Process a list of ParsedAddRequests.
+     * @param r
+     * @param channel
+     */
+    void processAddRequest(List<BookieProtocol.ParsedAddRequest> r, BookieRequestHandler channel);
```
zymap, line 58: Looks like `void processRequest(Object r, BookieRequestHandler channel);` is the same as the new one. Because the argument is an object, we can reuse that method

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestHandler.java
```diff
@@ -87,7 +97,32 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
...
             return;
         }
-        requestProcessor.processRequest(msg, this);
+
+        if (msg instanceof BookieProtocol.ParsedAddRequest
+            && ADDENTRY == ((BookieProtocol.ParsedAddRequest) msg).getOpCode()
+            && !((BookieProtocol.ParsedAddRequest) msg).isHighPriority()
+            && ((BookieProtocol.ParsedAddRequest) msg).getProtocolVersion() == BookieProtocol.CURRENT_PROTOCOL_VERSION
+            && !((BookieProtocol.ParsedAddRequest) msg).isRecoveryAdd()) {
+            msgs.put((BookieProtocol.ParsedAddRequest) msg);
```
zymap, line 106: Put is a blocking operation. We should check the size first and then put the message into the queue.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestHandler.java
```diff
@@ -87,7 +97,32 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
...
+            if (msgs.size() >= maxCapacity) {
+                int count = msgs.size();
+                List<BookieProtocol.ParsedAddRequest> c = new ArrayList<>(count);
+                msgs.drainTo(c, count);
+                requestProcessor.processAddRequest(c, this);
+            }
```
zymap, line 112:
```suggestion
            if (!msgs.offer(msg)) {
                ctx.fireChannelReadComplete();
            }
            msgs.put(msg); 
```

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
```diff
@@ -229,6 +230,16 @@ protected void onAddRequestFinish() {
...
 
+    protected void onAddRequestFinishWithoutUnTrack() {
+        if (addsSemaphore != null) {
+            addsSemaphore.release();
+        }
+    }
+
+    protected void onAddRequestUnTrack() {
+        requestStats.untrackAddRequest();
+    }
```
zymap, line 241: You need to track N add requests at onAddRequestStart, and untrack it when onAddRequestFinish.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
```diff
@@ -729,4 +740,28 @@ public boolean isBlacklisted(Channel channel) {
     public void handleNonWritableChannel(Channel channel) {
         onResponseTimeout.accept(channel);
     }
+
+    @Override
+    public void processAddRequest(List<BookieProtocol.ParsedAddRequest> msgs, BookieRequestHandler requestHandler) {
+        WriteBatchEntryProcessor write = WriteBatchEntryProcessor.create(msgs, requestHandler, this);
```
zymap, line 746: You can use[ the original method](https://github.com/apache/bookkeeper/pull/3846/files#diff-380ca68ed5ce21fb226c59b8f9c9bd7c7be70da37091ea2c55db89cc4125e578R318) and when the Object msg is `List<BookieProtocol.ParsedAddRequest>` then call here. They do the same thing just in different ways, so let's reuse the existing method.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
hangc0276: We batched the add requests, but each request callback one by one. So I tracked N add requests at onAddRequestStart, but I should not call the onAddRequestFinish after all the entries callback reached. We need to release the addsSemaphore when one callback is reached and un-track the add requests after all the entries callback reached.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -885,6 +887,25 @@ public void logAddEntry(ByteBuf entry, boolean ackBeforeSync, WriteCallback cb,
...
+            ByteBuf entry = entries.get(i);
+            long ledgerId = entry.getLong(entry.readerIndex());
+            long entryId = entry.getLong(entry.readerIndex() + 8);
+            entry.retain();
+            reserveMemory += entry.readableBytes();
+            queueEntries[i] = QueueEntry.create(entry, ackBeforeSync, ledgerId, entryId, cb, ctx,
+                MathUtils.nowInNano(), journalStats.getJournalAddEntryStats(), callbackTime);
+        }
+
+        memoryLimitController.releaseMemory(reserveMemory);
```
horizonzy, line 904: -> memoryLimitController.reserveMemory(reserveMemory);

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
```diff
@@ -1087,6 +1096,80 @@ public void addEntry(ByteBuf entry, boolean ackBeforeSync, WriteCallback cb, Obj
...
+                request.recycle();
+            }
+        }
+        handleMap.clear();
+
+        if (hasFailedRequests && requestProcessor != null) {
+            requestProcessor.flushPendingResponses();
+        }
+
+        if (writeDataToJournal && !requests.isEmpty()) {
```
horizonzy, line 1161: Whatever `writeDataToJournal` is true or false, we should release the request byteBuf.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/WriteBatchEntryProcessor.java
```diff
@@ -0,0 +1,143 @@
...
+        requestHandler = null;
+        requestProcessor = null;
+        startTimeNanos = -1L;
+    }
+
+    public static WriteBatchEntryProcessor create(List<ParsedAddRequest> requests, BookieRequestHandler requestHandler,
+                                                  BookieRequestProcessor requestProcessor) {
+        WriteBatchEntryProcessor wbep = RECYCLER.get();
+        wbep.init(requests, requestHandler, requestProcessor);
+        requestProcessor.onAddRequestStart(requestHandler.ctx().channel(), requests.size());
```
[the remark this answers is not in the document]
hangc0276: I add the requests.size() instead of 1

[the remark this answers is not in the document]
horizonzy: The `addsSemaphore` use requests.size().
The `addsInProgress` use 1.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/WriteBatchEntryProcessor.java
```diff
@@ -0,0 +1,144 @@
...
+    long startTimeNanos;
+    List<ParsedAddRequest> requests;
+    AtomicInteger requestCount;
+
+    @Override
+    protected void reset() {
+        requests = null;
+        requestHandler = null;
+        requestProcessor = null;
+        requestCount = null;
```
horizonzy, line 49: Here we need not set `requestCount = null`, we can set requestCount.set(request.size()) in the init method to override the value.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/WriteBatchEntryProcessor.java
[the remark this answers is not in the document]
hangc0276: got it

horizonzy (approved):
LGTM.

zymap (approved):

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
```diff
@@ -1087,6 +1096,80 @@ public void addEntry(ByteBuf entry, boolean ackBeforeSync, WriteCallback cb, Obj
...
+            List<ByteBuf> entries = requests.stream()
+                .map(ParsedAddRequest::getData).collect(Collectors.toList());
+            getJournal(requests.get(0).getLedgerId()).logAddEntry(entries, ackBeforeSync, cb, ctx);
```
eolivelli, line 1164: here we are assuming that all the entries are for the same ledger.
Can we add some assertions ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
```diff
@@ -1087,6 +1096,80 @@ public void addEntry(ByteBuf entry, boolean ackBeforeSync, WriteCallback cb, Obj
...
+        }
+
+        requests.forEach(t -> {
```
eolivelli, line 1167: put this into a "finally" block ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -885,6 +887,26 @@ public void logAddEntry(ByteBuf entry, boolean ackBeforeSync, WriteCallback cb,
         logAddEntry(ledgerId, entryId, entry, ackBeforeSync, cb, ctx);
     }
 
+    public void logAddEntry(List<ByteBuf> entries, boolean ackBeforeSync, WriteCallback cb, Object ctx)
```
eolivelli, line 890: nit: logAddEntries

eolivelli (changes requested):
Very interesting work

@merlimat @dlg99 PTAL

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestHandler.java
```diff
@@ -34,20 +40,24 @@
...
 
     private ByteBuf pendingSendResponses = null;
     private int maxPendingResponsesSize;
 
     BookieRequestHandler(ServerConfiguration conf, RequestProcessor processor, ChannelGroup allChannels) {
         this.requestProcessor = processor;
         this.allChannels = allChannels;
+
+        int maxCapacity = conf.getMaxAddsInProgressLimit() > 0 ? conf.getMaxAddsInProgressLimit() : DEFAULT_CAPACITY;
+        this.msgs = new ArrayBlockingQueue<>(maxCapacity);
```
wenbingshen, line 60: I'm not sure if it's going to happen, but I wanted to ask this question: Do we need to add a byte size limit? How to prevent the channel from reading a large amount of data from the socket and causing the OOM of the direct memory?

wenbingshen (commented):
Nice work!

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
hangc0276: We don't assume all the entries belong to the same ledger. When we write entries in the journal, we don't need to write all the ledger's data into the same journal directory. The goal of `getJournal(requests.get(0).getLedgerId())` is to balance the write throughput between journals.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestHandler.java
hangc0276: I think maybe not. The megs queue is also controlled by the Netty request handle. Netty also has a read batch control and once the batch limit is reached, it will call `channelReadComplete`, and we will flush the msgs queue requests to the storage when the `channelReadComplete` is called.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestHandler.java
```diff
@@ -87,7 +97,33 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
...
+        } else {
+            requestProcessor.processRequest(msg, this);
+        }
+    }
+
+    @Override
+    public void channelReadComplete(ChannelHandlerContext ctx) {
+        if (!msgs.isEmpty()) {
+            int count = msgs.size();
+            log.info("[hangc] count: {}", count);
```
horizonzy, line 120: maybe it's a test log.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestHandler.java
wenbingshen: I got it. Thanks!

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/WriteBatchEntryProcessor.java
```diff
@@ -0,0 +1,143 @@
...
+                log.warn("BookieServer is running in readOnly mode, so rejecting the request from the client!");
+                for (ParsedAddRequest r : requests) {
+                    writeComplete(BookieProtocol.EREADONLY, r.getLedgerId(), r.getEntryId(), null,
+                        requestHandler.ctx());
+                    r.release();
+                    r.recycle();
+                }
+                return;
+        }
+
```
wenbingshen, line 105: When batch add, do we need to update the `writeThreadQueuedLatency` metric? like: https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PacketProcessorBase.java#L176-L184

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestHandler.java
hangc0276: removed it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/WriteBatchEntryProcessor.java
hangc0276: Good point, added it.

wenbingshen (approved):

dlg99 (commented):
Overall a good idea, but I have concerns related to handling of
* backpressure
* memoryLimitController's enforced limits
* hitting memtable's flush limit in the middle of the batch

I think these areas require more testing and perf testing to make sure that perf improvements won't become perf regression under peak load

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java
```diff
@@ -205,21 +206,21 @@ public BookieRequestProcessor(ServerConfiguration serverCfg, Bookie bookie, Stat
...
+    protected void onAddRequestStart(Channel channel, int permits) {
         if (addsSemaphore != null) {
-            if (!addsSemaphore.tryAcquire()) {
+            if (!addsSemaphore.tryAcquire(permits)) {
                 final long throttlingStartTimeNanos = MathUtils.nowInNano();
                 channel.config().setAutoRead(false);
                 LOG.info("Too many add requests in progress, disabling autoread on channel {}", channel);
                 requestStats.blockAddRequest();
-                addsSemaphore.acquireUninterruptibly();
+                addsSemaphore.acquireUninterruptibly(permits);
```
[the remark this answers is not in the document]
horizonzy: +1

XLzed:
Hello! I'm a beginner of bookkeeper. Could you share more details about the benchmark like bk configuration and hardware enviroment? I want to use this change and learn the optimization method within it. But when I tried to reproduce the result of your benchmark, the results looked a little odd, the optimized version of protocolV2 performed about as well as the unoptimized version of protocolV3. What's the difference between V2 and V3？And why this optimization supports V2 only?
Here's my hardware enviroment and benchmark results, thank you!

server cmd：OPTS='-XX:MaxDirectMemorySize=10G -Xms40G -Xmx40G' ./bin/bookkeeper localbookie 1
client cmd：OPTS='-Dorg.apache.bookkeeper.conf.readsystemproperties=true -DuseV2WireProtocol=true' ./bookkeeper-benchmark/bin/benchmark writes -ensemble 1 -quorum 1 -ackQuorum 1 -ledgers 50 -throttle 10000

image

Before Change（useV2WireProtocol=false）
<html>
<body>
<div data-sabo-line="true" style="margin: 0px; white-space: pre;">

times | ops/sec | p95 latency | p99 latency
-- | -- | -- | --
1 | 212980 | 41.68268 | 44.85271
2 | 202530 | 42.99525 | 46.63221
3 | 212138 | 42.30779 | 45.08659
4 | 213154 | 42.22976 | 44.76916
5 | 198376 | 42.3406 | 47.08146
Avg | 207835.6 | 42.31122 | 45.68443

</div><div data-sabo-app-id="note.temu.team"></div>
</body>
</html>

Before Change (useV2WireProtocol=true)
<html>
<body>
<div data-sabo-line="true" style="margin: 0px; white-space: pre;">

times | ops/sec | p95 latency | p99 latency
-- | -- | -- | --
1 | 195363 | 46.08798 | 49.72923
2 | 198560 | 45.37565 | 48.83366
3 | 198490 | 45.53701 | 49.00147
4 | 202162 | 44.70965 | 48.1483
5 | 197736 | 45.71856 | 49.23503
Avg | 198462.2 | 45.48577 | 48.98954

</div><div data-sabo-app-id="note.temu.team"></div>
</body>
</html>

After Change (useV2WireProtocol=true)
<html>
<body>
<div data-sabo-line="true" style="margin: 0px; white-space: pre;">

times | ops/sec | p95 latency | p99 latency
-- | -- | -- | --
1 | 211978 | 41.49986 | 44.78894
2 | 216729 | 41.85365 | 44.18242
3 | 208047 | 42.2268 | 45.47032
4 | 204107 | 41.99345 | 46.13609
5 | 203803 | 42.82836 | 46.37437
Avg | 208932.8 | 42.08043 | 45.39043

</div><div data-sabo-app-id="note.temu.team"></div>
</body>
</html>

PR #3848, opened 2023-03-08
Improve group and flush add-responses after journal sync

horizonzy:
This is an improvement for #3837

### Motivation
1. Now if the maxPendingResponsesSize is expanded large, it will not decrease. => We should make it flexible.
2. Now after prepareSendResponseV2 to the channel, then we trigger all channels to flush pendingSendResponses, maybe there is only a few channels that need to flush, but if we trigger all channels, it's a waste. => We only flush the channel which prepareSendResponseV2.


horizonzy:
rerun failure checks

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -495,6 +504,7 @@ public void run() {
 
                     journalStats.getForceWriteQueueSize().addCount(-requestsCount);
 
+                    Set<BookieRequestHandler> writeHandlers = new HashSet<>();
```
merlimat, line 507: Using the set has its own overhead too. In this case it will allocate a `Node` object for the linked list for each entry added here.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -503,17 +513,26 @@ public void run() {
                     // responses
                     for (int i = 0; i < requestsCount; i++) {
                         ForceWriteRequest req = localRequests.get(i);
+                        req.getForceWriteWaiters().forEach(ele -> {
```
merlimat, line 516: `req.process()` is already iterating over all the entries for each `ForceWriteRequest`. It would be better to reuse the same loop.

Additionally using `forEach` has 2 problems: 
 1. allocates an iterator, unlike doing `for (int i =0; i... .)` 
 2. the lambda is also capturing variables and thus it will allocate a temporary object

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
merlimat: Perhaps we could use `ObjectHashSet` from carrotsearch https://github.com/carrotsearch/hppc
It's already used in Pulsar too. 

We should also reuse the same set each time.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestHandler.java
```diff
@@ -92,31 +92,22 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
...
-                    } else {
-                        pendingSendResponses.release();
-                    }
-
-                    pendingSendResponses = null;
-                }
+    public synchronized void flushPendingResponse() {
+        if (pendingSendResponses != null) {
+            maxPendingResponsesSize = (int) Math.max(
+                    maxPendingResponsesSize * 0.9 + 0.1 * pendingSendResponses.readableBytes(),
```
merlimat, line 103: This looks like it's going to take quite a bit of iterations to expand the cluster size. While in the meantime we have to keep expanding the `ByteBuf`, which will incur in multiple allocations and memory copies

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -503,17 +513,18 @@ public void run() {
...
                             .registerSuccessfulValue(numReqInLastForceWrite);
-
-                    if (requestProcessor != null) {
-                        requestProcessor.flushPendingResponses();
+                    if (writeHandlers.size() > 0) {
+                        for (ObjectCursor<BookieRequestHandler> writeHandler : writeHandlers) {
+                            writeHandler.value.flushPendingResponse();
+                        }
+                        writeHandlers.clear();
                     }
```
merlimat, line 527: In this case it would efficient to use the lambda approach because it would be non-capturing and `ObjectHashSet` is not going to create an iterator either.

```suggestion
                    writeHandlers.forEach(wh -> wh.flushPendingResponse());
                    writeHandlers.clear();
```

merlimat (approved):

horizonzy:
rerun failure checks

zymap (approved):

horizonzy:
rerun failure checks

wenbingshen:
rerun failure checks

wenbingshen (approved):
Good work!

hangc0276 (approved):
Good job!

PR #3863, opened 2023-03-14
Disable code coverage and Codecov report

yaalsn:
### Motivation

There're two reasons that we want to disable the code coverage.

1. The current report result is not accurate.
2. We can't get the PR's unit test's code coverage because of the apache Codecov permission.


yaalsn:
rerun failure checks

Shoothzj (approved):

wenbingshen (approved):

PR #3869, opened 2023-03-16
[fix][gc] Change order of doGcLedgers and extractMetaFromEntryLogs

dlg99:
### Motivation

Garbage collector does
```
 extractMetaFromEntryLogs();
 doGcLedgers();
```

where extractMetaFromEntryLogs checks if ledger exists in the storage and doGcLedgers removes deleted ledgers from the storage.

Logically, these should be reordered.
Currently it means that deleted ledgers will be either re-verified during compaction and skipped (if verifyMetadataOnGC is true) or picked up on the next GC run.  
In prod it means data released later.

The tests pass as they tend to do
```
        getGCThread().enableForceGC();
        getGCThread().triggerGC().get();
```
where both of these calls trigger gc.

### Changes

reordered operations


hangc0276 (approved):

wenbingshen (approved):
Good catch!

PR #3872, opened 2023-03-16
Add log for entry log file delete.

horizonzy:
When deleting the entry log file, we would better add a log for it.
```
2023-03-17T00:25:02,069 - INFO  - [main:GarbageCollectorThread@748] - Entry log file 2 is empty, delete it from disk.
```

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
@@ -745,6 +745,7 @@ public class GarbageCollectorThread implements Runnable {
                 EntryLogMetadata entryLogMeta = entryLogger.getEntryLogMetadata(entryLogId, throttler);
                 removeIfLedgerNotExists(entryLogMeta);
                 if (entryLogMeta.isEmpty()) {
+                    LOG.info("Entry log file {} is empty, delete it from disk.", entryLogId);
                     entryLogger.removeEntryLog(entryLogId);
                     // remove it from entrylogmetadata-map if it is present in
                     // the map


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -745,6 +745,7 @@ protected void extractMetaFromEntryLogs() throws EntryLogMetadataMapException {
                 EntryLogMetadata entryLogMeta = entryLogger.getEntryLogMetadata(entryLogId, throttler);
                 removeIfLedgerNotExists(entryLogMeta);
                 if (entryLogMeta.isEmpty()) {
+                    LOG.info("Entry log file {} is empty, delete it from disk.", entryLogId);
```
merlimat, line 748: `entryLogId` is not directly the same (eg: it gets converted to hex)

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/directentrylogger/DirectEntryLogger.java
```diff
@@ -377,6 +381,9 @@ public boolean removeEntryLog(long entryLogId) {
         checkArgument(entryLogId < Integer.MAX_VALUE, "Entry log id must be an int [%d]", entryLogId);
         File file = logFile(ledgerDir, (int) entryLogId);
         boolean result = file.delete();
+        if (!result) {
+            LOG.warn("Could not delete entry log file {}", file);
```
hangc0276, line 385: Please use slog to print logs.

horizonzy:
rerun failure checks

horizonzy:
rerun failure checks

PR #3874, opened 2023-03-17
[bookie-shell] fix readLedger with parameter -bookie and -ledgerid can not read any entries in ledger

TakaHiR07:
### Motivation

If we use bookieshell readLedger only with parameter -ledgerid, we can read all entries in this ledger, but if use with parameter -ledgerid and --bookie, we cant read any entries. 

This is because it try to reverse entries from firstentry(default 0)  to lastEntry(default -1). Therefore it would directly finish without read any entries, misleading us that ledgerId in our selected bookie has no entry.

### Changes

when set -ledgerid and -bookie and not set -entry, continue to read next entry until get BKNoSuchEntryException

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/commands/bookie/ReadLedgerCommand.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/commands/bookie/ReadLedgerCommand.java
@@ -31,7 +31,6 @@ import java.util.Iterator;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
-import java.util.stream.LongStream;
 import lombok.Setter;
 import lombok.experimental.Accessors;
 import org.apache.bookkeeper.client.BKException;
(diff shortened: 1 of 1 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/commands/bookie/ReadLedgerCommand.java
```diff
@@ -189,33 +188,39 @@ private boolean readledger(ServerConfiguration serverConf, ReadLedgerFlags flags
...
+                                }
+
+                                future.complete(null);
+                            }, null, BookieProtocol.FLAG_NONE);
 
                     try {
                         future.get();
                     } catch (Exception e) {
                         LOG.error("Error future.get while reading entries from ledger {}", flags.ledgerId, e);
+                        if (e.getCause() instanceof BKException.BKNoSuchEntryException && lastEntry == -1) {
```
horizonzy, line 218: If the ledger only has 100 entries, but the input param `lastEntry` is 10000, it will still read 100-10000 entries. 
There are two ways to solve it.
1. remove the `lastEntry == -1` judgment, once meet BKNoSuchEntryException, end the loop.
2. Add more limitations for the loop. (lastEntry == -1 || nextEntryId <= lastEntry && nextEntryId <= LAC), we can use bookieClient.readLac() to get the LAC.

TakaHiR07: Second way is more reasonable, I will implement it.

shoothzj:
@TakaHiR07 Are you still working on this?

TakaHiR07:
> @TakaHiR07 Are you still working on this?

Sorry for forget this work. If needing, I can finish this in a few day.

shoothzj:
@TakaHiR07 Thanks for your contribution, I want to move this forward in the next release.

shoothzj (changes requested):
Can we add some tests?

TakaHiR07:
> Can we add some tests?

Ok. I would try it today.

TakaHiR07:
> Can we add some tests?

@shoothzj @horizonzy Have done. Can you help review again. Thx

shoothzj (approved):

PR #3876, opened 2023-03-18
Try to use jdk api to create hardlink when rename file when compaction.

lifepuzzlefun:
### Motivation

Current HardLink will create a process to execute mv like command to rename file in compaction logic.

maybe we can just use jdk api to do this with lower overhead.

see javadoc: https://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#createLink(java.nio.file.Path,%20java.nio.file.Path)

### Changes

test if `Files.createLink` is available in hardlink static code block.
if test fails means `Files.createLink` is not available.

else will use `Files.createLink` when call createHardlink


lifepuzzlefun:
@hangc0276 @eolivelli @StevenLuMT can you take a look ?

Shoothzj:
IIUC, hardlink will result in two files with same inode, which is different from *mv* command?

lifepuzzlefun:
> IIUC, hardlink will result in two files with same inode, which is different from *mv* command?

thanks for review. the current code will create a process
 to do the whole thing which is much slower than in-process api call.

Shoothzj (approved):
LGTM

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/HardLink.java
```diff
@@ -416,6 +471,19 @@ public static void createHardLink(File file, File linkName)
       throw new IOException(
           "invalid arguments to createHardLink: link name is null");
     }
+
+    // if createLink available try first, else fall back to shell command.
+    if (CREATE_LINK_SUPPORTED.get()) {
```
horizonzy, line 476: We can make the `CREATE_LINK_SUPPORTED` default value true, and remove the test code.
If use createLink failed, change `CREATE_LINK_SUPPORTED` to false.

lifepuzzlefun: fixed.

hangc0276 (approved):

zymap (approved):

hangc0276:
@lifepuzzlefun Thanks for your contribution. Would you please add a test to cover this change?

lifepuzzlefun:
> @lifepuzzlefun Thanks for your contribution. Would you please add a test to cover this change?

sorry for late reply. a unit test to cover this change is uploaded. :-)

zymap:
@lifepuzzlefun Could you please try rebase master to retrigger the ci?

lifepuzzlefun:
> @lifepuzzlefun Could you please try rebase master to retrigger the ci?

rebased. : -)

hangc0276:
@lifepuzzlefun Would you please fix the license issue?

hangc0276:
Move to the next release first. If this license can be fixed before the 4.16.4 release is triggered, I will include this PR.

PR #3884, opened 2023-03-24
Fix write entry failed in without writing journal case

hangc0276:
### Motivation
https://github.com/apache/bookkeeper/pull/3837 introduced group flush add responses triggered by journal sync. However, if we skip writing journals, the add responses won't be flushed to the netty channel and the client will receive write entries timeout.

### Changes
- Flush the add responses when skipping writing journals
- Add tests to cover V2 protocol and skip writing journal cases.


eolivelli (approved):
Great catch

It's weird that no test failed!

hangc0276:
@merlimat @eolivelli @zymap @horizonzy Please help take a look at this fix, and it is a blocker for BookKeeper 4.16.0 release.

horizonzy (approved):
LGTM

PR #3383, opened 2022-07-03
Consolidate Netty channel flushes to mitigate syscall overhead

merlimat:
### Motivation

When we are writing/reading a lot of small entries, one of the main bottleneck in BK client/server becomes the CPU usage. A big part of it is caused by the syscall overhead when writing to the socket. 

The reason is that we have many independent (and small) operations happening on the connection and each time we call `writeAndFlush()` on each of them, causing many `write()` calls on the socket.

Netty has a mechanism to consolidate the flushes on the channel and it improves the handling of many small entries.

PR #3886, opened 2023-03-26
[Improve][Client] Group writing for per bookie client

hangc0276:
### Motivation
When the BookKeeper client writes an entry to the BookKeeper server, it runs with the following steps:
- Step 1: Initiate a PendingAddOp object.
- Step 2: For each replica, select a bookie client channel according to the ledgerId
- Step 3: Write the entry to the bookie client channel, and flush it.
- Step 4: The entry was added to Netty's pending queue, and processed with the configured Netty pipeline, such as `bookieProtoEncoder`,  `lengthbasedframedecoder`, and `consolidation`
- Step 5: Waiting for the written response

If the bookie client writes small entries with high ops and the Netty's pending queue will be full and the Netty thread will be busy with processing entries and flushing them into the socket channel. The CPU will switch between the user mode and the kernel mode in high frequency.

#3383 introduced Netty channel flushes consolidation to mitigate syscall overhead. But it can not reduce the overhead on the Netty threads. 

We can tune it one `Step 3` to group the small entries into one ByteBuf and flush it into the Netty pending queue when conditions are met.

### Design
When a new entry comes to the bookie client channel, we add it into one ByteBuf and check whether the ByteBuf exceeds the max threshold, the default is 1MB. 

In order to avoid entry staying in the Bookie client channel ByteBuf for a long time causing high write latency, we schedule a timer task to flush the ByteBuf every 1 ms.

### Performance
We test the write performance on my laptop with the following command.
```bash
bin/benchmark writes -ensemble 1 -quorum 1 -ackQuorum 1 -ledgers 100 -throttle 300000 -entrysize 60 -useV2 -warmupMessages 1000000
```
The performance result.
|  Writer ledgers | batched write ops/s | non-batched write ops/s | improved |
| --- | --- | --- | --- | 
| 1 | 333238 | 335970 | 0% |
| 50 | 261605 | 153011 |  71% |
| 100 | 260650 | 126331 |  100% |
| 500 | 265628 | 164393 | 62%  |


bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
```diff
@@ -1154,26 +1162,80 @@ private void writeAndFlush(final Channel channel,
...
+                        }
+                    } else {
+                        nettyOpLogger.registerFailedEvent(MathUtils.elapsedNanos(startTime), TimeUnit.NANOSECONDS);
+                    }
+                });
+                channel.writeAndFlush(request, promise);
+            }
         } catch (Throwable e) {
             LOG.warn("Operation {} failed", StringUtils.requestToString(request), e);
             errorOut(key);
```
horizonzy, line 1192: The catch code block is not suitable for the new logic. It just handle the single key case, and didn't handle the pending keys case.
And in the try code block, it looks like there is no exception that will be thrown. I think we can remove the try catch block.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
```diff
@@ -1154,26 +1162,80 @@ private void writeAndFlush(final Channel channel,
...
+            pendingSendRequests = channel.alloc().directBuffer(maxPendingRequestsSize);
+        }
+        BookieProtoEncoding.RequestEnDeCoderPreV3.serializeAddRequests(request, pendingSendRequests);
+        pendingSendKeys.add(key);
+        return pendingSendRequests.readableBytes() > MAX_PENDING_REQUEST_SIZE;
+    }
+
+    public synchronized void flushPendingRequests() {
+        final long startTime = MathUtils.nowInNano();
+        Set<CompletionKey> keys = new HashSet<>(pendingSendKeys);
```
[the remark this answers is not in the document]
horizonzy: +1.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
```diff
@@ -1154,26 +1162,80 @@ private void writeAndFlush(final Channel channel,
...
+    public synchronized boolean prepareSendRequests(Channel channel, Object request, CompletionKey key) {
+        if (pendingSendRequests == null) {
+            pendingSendRequests = channel.alloc().directBuffer(maxPendingRequestsSize);
+        }
+        BookieProtoEncoding.RequestEnDeCoderPreV3.serializeAddRequests(request, pendingSendRequests);
+        pendingSendKeys.add(key);
+        return pendingSendRequests.readableBytes() > MAX_PENDING_REQUEST_SIZE;
+    }
+
+    public synchronized void flushPendingRequests() {
```
horizonzy, line 1205: The flushPendingRequests trigger at every millisecond. It may exacerbate the lock race between `prepareSendRequests` and `flushPendingRequests.`

horizonzy: Here we can refer the write cache and flush cache to make flushPendingRequests lock free.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
```diff
@@ -1154,26 +1162,80 @@ private void writeAndFlush(final Channel channel,
...
-                if (future.isSuccess()) {
-                    nettyOpLogger.registerSuccessfulEvent(MathUtils.elapsedNanos(startTime), TimeUnit.NANOSECONDS);
-                    CompletionValue completion = completionObjects.get(key);
-                    if (completion != null) {
-                        completion.setOutstanding();
-                    }
-                } else {
-                    nettyOpLogger.registerFailedEvent(MathUtils.elapsedNanos(startTime), TimeUnit.NANOSECONDS);
+                if (nextScheduledFlush == null) {
+                    nextScheduledFlush = channel.eventLoop().scheduleWithFixedDelay(this::flushPendingRequests,
```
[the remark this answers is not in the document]
horizonzy: > Do we need to make the delay time configurable?

We should ensure the flush operation as soon as possible, or it may lead to the addEntry operation timeout. So if we make it configurable, the user may config it with a wrong value.

dlg99:
I think this should be configurable. 
It improves throughput for some workloads, it is possible that more latency-sensitive workloads would want to disable this

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtoEncoding.java
```diff
@@ -215,6 +215,14 @@ private static byte[] readMasterKey(ByteBuf packet) {
 
             return masterKey;
         }
+
+        public static void serializeAddRequests(Object request, ByteBufList buf) {
+            if (request instanceof ByteBuf) {
+                buf.add((ByteBuf) request);
+            } else if (request instanceof ByteBufList) {
+                buf.add((ByteBufList) request);
+            }
```
eolivelli, line 224: else throw IllegalStateException ?

eolivelli (changes requested):

PR #3889, opened 2023-03-27
[improve] [client] Group read/write response when jumping to the ordered executor

hangc0276:
### Motivation
When the PerChannelBookieClient receives read/write responses from the Bookie server, it processes those responses one by one with the following pipeline
- Get the CompletionValue according to the CompletionKey
- Jump to the ordered executor the run the callback

https://github.com/apache/bookkeeper/blob/cbd634dbe0c894474a934fab6d8711588ad94114/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java#L1353-L1374

If there are 1000 responses, we need to jump 1000 times into the ordered executor and cause 1000 times thread context switches.

### Design
We can group those responses according to whether the Netty socket channel is empty. We initiate a Blocking Queue to buffer all the responses, and the default capacity is 1_000. There are two conditions that will trigger the response process.
- The Blocking queue is full
- The Netty socket channel is empty and the `channelReadComplete` is called.

If we buffered 1_000 responses in the queue, we will reduce 1_000 times thread context switch if those responses are hashed to the same thread to execute.


bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
```diff
@@ -173,6 +178,7 @@ public class PerChannelBookieClient extends ChannelInboundHandlerAdapter {
                         BKException.Code.DuplicateEntryIdException,
                         BKException.Code.WriteOnReadOnlyBookieException));
     private static final int DEFAULT_HIGH_PRIORITY_VALUE = 100; // We may add finer grained priority later.
+    private static final int DEFAULT_CAPACITY = 1_000;
```
horizonzy, line 181: DEFAULT_CAPACITY > BATCH_RESPONSE_SIZE

horizonzy (approved):

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
```diff
@@ -1341,7 +1349,10 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
 
         if (msg instanceof BookieProtocol.Response) {
             BookieProtocol.Response response = (BookieProtocol.Response) msg;
-            readV2Response(response);
+            if (!responses.offer(response)) {
```
dlg99, line 1352: It will increase the latency of the latency-sensitive workloads.
At least this has to be behind a feature flag.

hangc0276: Good point. I add a flag to enable this feature.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
hangc0276: done

ytong01:
I see this is only for v2,  it does not work for v3?

PR #3894, opened 2023-03-28
Return activeLogChannel if new create

AnonHxy:
### Motivation

If the activeLogChannel is new create, we should fast return it

### Changes

Fast return if new create activeLogChannel


horizonzy (approved):
LGTM.

AnonHxy:
rerun failure checks

AnonHxy:
rerun failure checks

hangc0276 (commented):
Would you please add a test to cover this case?

AnonHxy:
> Would you please add a test to cover this case?

Sure. I have add a test to verfiy this. PTAL @hangc0276

wenbingshen (approved):
LGTM

PR #3895, opened 2023-03-28
[feature] [server] add dbStorage_readAheadCacheBatchBytesSize properties when read ahead entries

StevenLuMT:

### Motivation

read ahead only limits the number of entries, which is not precise enough. so we need to add the bytes limit to make the prefetch to be more controllable

### Changes

1.  add dbStorage_readAheadCacheBatchBytesSize properties 
2. when dbStorage_readAheadCacheBatchBytesSize is setted, use it to limit the bytes for read ahead entries.

### TestCase 

1. Regression test case:  DbLedgerStorageReadCacheTest#chargeReadAheadCacheRegressionTest
2. Unit test case:   DbLedgerStorageReadCacheTest#chargeReadAheadCacheUnitTest
3. Perf test case:   DbLedgerStorageReadCacheTest#compareDiffReadAheadPerfTest

perf test results demonstrating improvements with this option:
case(read ahead cache by limit bytes size) get less cachemiss,it is suitable for large messages, reduce the pollution of readAhead large messages to readCache
image

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
@@ -700,11 +700,9 @@ public class SingleDirectoryDbLedgerStorage implements CompactableLedgerStorage 
         // compatible with old logic
         boolean chargeSizeCondition = currentReadAheadCount < readAheadCacheBatchSize
                 && currentReadAheadBytes < maxReadAheadBytesSize;
-        if (readAheadCacheBatchBytesSize > 0) {
+        if (chargeSizeCondition && readAheadCacheBatchBytesSize > 0) {
             // exact limits limit the size and count for each batch
-            chargeSizeCondition = currentReadAheadCount < readAheadCacheBatchSize
-                    && currentReadAheadBytes < readAheadCacheBatchBytesSize
-                    && currentReadAheadBytes < readCacheMaxSize;
+            chargeSizeCondition = currentReadAheadBytes < readAheadCacheBatchBytesSize;
         }
         return chargeSizeCondition;
     }
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
@@ -132,6 +132,7 @@ public class SingleDirectoryDbLedgerStorage implements CompactableLedgerStorage 
     private final long writeCacheMaxSize;
     private final long readCacheMaxSize;
     private final int readAheadCacheBatchSize;
+    private final long readAheadCacheBatchBytesSize;
 
     private final long maxThrottleTimeNanos;
 
@@ -146,7 +147,8 @@ public class SingleDirectoryDbLedgerStorage implements CompactableLedgerStorage 
     public SingleDirectoryDbLedgerStorage(ServerConfiguration conf, LedgerManager ledgerManager,
                                           LedgerDirsManager ledgerDirsManager, LedgerDirsManager indexDirsManager,
                                           EntryLogger entryLogger, StatsLogger statsLogger, ByteBufAllocator allocator,
-                                          long writeCacheSize, long readCacheSize, int readAheadCacheBatchSize)
+                                          long writeCacheSize, long readCacheSize, int readAheadCacheBatchSize,
+                                          long readAheadCacheBatchBytesSize)
             throws IOException {
         checkArgument(ledgerDirsManager.getAllLedgerDirs().size() == 1,
                 "Db implementation only allows for one storage dir");
@@ -173,6 +175,7 @@ public class SingleDirectoryDbLedgerStorage implements CompactableLedgerStorage 
 
         readCacheMaxSize = readCacheSize;
         this.readAheadCacheBatchSize = readAheadCacheBatchSize;
+        this.readAheadCacheBatchBytesSize = readAheadCacheBatchBytesSize;
 
         // Do not attempt to perform read-ahead more than half the total size of the cache
         maxReadAheadBytesSize = readCacheMaxSize / 2;
@@ -655,9 +658,7 @@ public class SingleDirectoryDbLedgerStorage implements CompactableLedgerStorage 
             long currentEntryLogId = firstEntryLogId;
             long currentEntryLocation = firstEntryLocation;
 
-            while (count < readAheadCacheBatchSize
-                    && size < maxReadAheadBytesSize
-                    && currentEntryLogId == firstEntryLogId) {
+            while (chargeReadAheadCache(count, size) && currentEntryLogId == firstEntryLogId) {
                 ByteBuf entry = entryLogger.readEntry(orginalLedgerId,
                         firstEntryId, currentEntryLocation);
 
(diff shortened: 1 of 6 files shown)


StevenLuMT:
@eolivelli @dlg99 @zymap @Shoothzj @HQebupt @AnonHxy 
needs review, thanks

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
```diff
@@ -695,6 +696,18 @@ private void fillReadAheadCache(long orginalLedgerId, long firstEntryId, long fi
         }
     }
 
+    private boolean chargeReadAheadCache(int currentReadAheadCount, long currentReadAheadBytes) {
```
eolivelli, line 699: can we add unit tests for this ? you can make it a static method, it will be very easy to write an unit test.
the logic is not straightforward, a test with edge cases will help

eolivelli (changes requested):
Perfect.

I added an ask about writing an unit test

Shoothzj (approved):

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
```diff
@@ -695,6 +696,18 @@ private void fillReadAheadCache(long orginalLedgerId, long firstEntryId, long fi
...
+        // compatible with old logic
+        boolean chargeSizeCondition = currentReadAheadCount < readAheadCacheBatchSize
+                && currentReadAheadBytes < maxReadAheadBytesSize;
+        if (readAheadCacheBatchBytesSize > 0) {
```
zymap, line 703: We should also check it is smaller than the max read cache size.

zymap (commented):
Would it be more clearly to add a configuration to change the percentage of the readCacheMaxSize? Currently, it is 50% of readCacheMaxSize. Using a bytes' configuration, you need to calculate the size and take care of the readCacheMaxSize. If we use percentage, we don't need to consider that.

hangc0276 (changes requested):
We'd better add the performance result to show this change is valuable.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
StevenLuMT: > can we add unit tests for this ? you can make it a static method, it will be very easy to write an unit test.
> the logic is not straightforward, a test with edge cases will help

Thanks for reviewing , I will add unit test today @eolivelli

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
StevenLuMT: Good catch, I will update it @zymap

StevenLuMT:
> Do you have perf test results demonstrating improvements with this option?

Thanks for reviewing , I will add unit test today @dlg99

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
```diff
@@ -695,6 +696,18 @@ private void fillReadAheadCache(long orginalLedgerId, long firstEntryId, long fi
...
+            // exact limits limit the size and count for each batch
+            chargeSizeCondition = currentReadAheadCount < readAheadCacheBatchSize
+                    && currentReadAheadBytes < readAheadCacheBatchBytesSize;
+        }
```
AnonHxy, line 707: If the `readAheadCacheBatchBytesSize` must smaller than `maxReadAheadBytesSize`, maybe we could change    like this:
```suggestion
        if (chargeSizeCondition && readAheadCacheBatchBytesSize > 0) {
            // exact limits limit the size and count for each batch
            chargeSizeCondition = currentReadAheadBytes < readAheadCacheBatchBytesSize;
        }
```

conf/bk_server.conf
```diff
@@ -749,6 +749,9 @@ gcEntryLogMetadataCacheEnabled=false
 # By default it will be allocated to 25% of the available direct memory
 # dbStorage_readAheadCacheMaxSizeMb=
 
+# How many entries' bytes to pre-fill in cache after a read cache miss
+# dbStorage_readAheadCacheBatchBytesSize=-1
```
AnonHxy, line 753: How about change the default value to 0?  Because 0 will also disable this feature.

And we'd better add a comments like this:
`Default is 0. 0 or less disables this feature`

conf/bk_server.conf
```diff
@@ -749,6 +749,9 @@ gcEntryLogMetadataCacheEnabled=false
 # By default it will be allocated to 25% of the available direct memory
 # dbStorage_readAheadCacheMaxSizeMb=
 
+# How many entries' bytes to pre-fill in cache after a read cache miss,default is 0 or less disables this feature
```
AnonHxy, line 752:
```suggestion
# How many entries' bytes to pre-fill in cache after a read cache miss. Default is -1. 0 or less disables this feature
```

conf/bk_server.conf
StevenLuMT: > How about change the default value to 0? Because 0 will also disable this feature.
> 
> And we'd better add a comments like this: `Default is 0. 0 or less disables this feature`

updated,thanks

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
StevenLuMT: > If the `readAheadCacheBatchBytesSize` must smaller than `maxReadAheadBytesSize`, maybe we could change like this:

The logic is not inherited, it cannot be changed like this, thanks

StevenLuMT:
I have add a unit test(TestBookieImpl) to cover this function,and add the performance result to show this change is valuable.
help me review,thanks @dlg99 @eolivelli @hangc0276 @zymap

AnonHxy (approved):

eolivelli (approved):
Lgtm

StevenLuMT:
> perf test

@dlg99 @hangc0276 add three testcase: 
1. Regression test case: DbLedgerStorageReadCacheTest#chargeReadAheadCacheRegressionTest
2. Unit test case: DbLedgerStorageReadCacheTest#chargeReadAheadCacheUnitTest
3. Perf test case: DbLedgerStorageReadCacheTest#compareDiffReadAheadPerfTest

StevenLuMT:
rerun failure checks

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
StevenLuMT: > I saw the unit tests, very good.
> 
> But here I meant to add a "unit test" that tests this method (you can make it "static", no need to start a Bookie)

updated,thanks

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
```diff
@@ -695,6 +696,19 @@ private void fillReadAheadCache(long orginalLedgerId, long firstEntryId, long fi
         }
     }
 
+    protected boolean chargeReadAheadCache(int currentReadAheadCount, long currentReadAheadBytes) {
+        // compatible with old logic
+        boolean chargeSizeCondition = currentReadAheadCount < readAheadCacheBatchSize
+                && currentReadAheadBytes < maxReadAheadBytesSize;
+        if (readAheadCacheBatchBytesSize > 0) {
+            // exact limits limit the size and count for each batch
+            chargeSizeCondition = currentReadAheadCount < readAheadCacheBatchSize
```
hangc0276, line 705: if (readAheadCacheBatchBytesSize > 0 && chargeSizeCondition) {
    chargeSizeCondition = chargeSizeCondition && currentReadAheadBytes < readCacheMaxSize;
}

hangc0276 (approved):
Good job!

PR #3783, opened 2023-02-17
Single buffer for small add requests

merlimat:
### Motivation

Similar to #3597, this is targeting AddRequest for using a single buffer when serializing them.

Modifications: 
 * Changed `DigestManager` to not return a `ByteBufList` in all conditions
 * If the payload is small, allocate a buffer large enough for the headers and the payload.

PR #3902, opened 2023-04-02
Fix promise lost in ClientSide AuthHandler

hangc0276:
### Motivation
#3783 changed the AddEntry Request type from `BookieProtocol.AddRequest` to `ByteBuf` or `ByteBufList`, leading to the AddEntry request being added into `waitingForAuth` queue in the client-side AuthHandler
https://github.com/apache/bookkeeper/blob/0171a408e21a51eb74e18a07df1b0ea71b7638ff/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java#L787-L791

https://github.com/apache/bookkeeper/blob/0171a408e21a51eb74e18a07df1b0ea71b7638ff/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/AuthHandler.java#L363-L364

For the requests in the `waitingForAuth`, they will be polled out by `AuthHandshakeCompleteCallback#operationComplete`. 
https://github.com/apache/bookkeeper/blob/0171a408e21a51eb74e18a07df1b0ea71b7638ff/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/AuthHandler.java#L426-L436

However, the `waitingForAuth` queue only stores the AddEntry request, and it ignores the related ChannelPromise. When polling requests from the `waitingForAuth` queue in `AuthHandshakeCompleteCallback#operationComplete`, those requests will be written and flushed into the Netty channel without passing any ChannelPromises, and Netty will use `voidPromise` by default.

In a word, the AddEntry request side passed ChannelPromise will be replaced with `voidPromise`, and the original promise will never be complete and the following operations won't execute.
https://github.com/apache/bookkeeper/blob/0171a408e21a51eb74e18a07df1b0ea71b7638ff/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java#L1159-L1169

There are two impactions:
- The `nettyOpLogger ` metric won't update
- The `addEntryOutstanding` counter metric won't update

### Changes
Store the ChannelPromise along with the request in the client-side AuthHandle and pass it to the Netty channel when polling out.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/AuthHandler.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/AuthHandler.java
@@ -41,6 +41,7 @@ import org.apache.bookkeeper.auth.ClientAuthProvider;
 import org.apache.bookkeeper.client.BKException;
 import org.apache.bookkeeper.proto.BookkeeperProtocol.AuthMessage;
 import org.apache.bookkeeper.util.ByteBufList;
+import org.apache.commons.lang3.tuple.Pair;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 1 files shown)


zymap (approved):
Do we need to have a test?

horizonzy (approved):
Nice catch

dlg99 (commented):
Please add a test

wenbingshen (approved):
Nice Catch!

It doesn't seem very easy to add an AddEntry request before the auth handshake is complete, especially getting it into the waitingForAuth queue, and the auth request will also trigger the promise execution.

hangc0276:
We need to add a unit test for this change, move to the next release

shoothzj:
As dicussed in #4293 I think we can close this PR now.

PR #3903, opened 2023-04-03
Fix keys leak in EntryLocationIndex when ledgersToDelete is empty

hangc0276:
### Motivation
In EntryLocationIndex#removeOffsetFromDeletedLedgers, if there are no ledgers to delete, the key pair will leak.

### Changes
Check the ledgers to deletion Set first, and if the set is not empty, then generate the keys.


AnonHxy (approved):

AnonHxy:
LGTM

zymap (approved):

horizonzy (approved):
Nice catch

dlg99 (approved):
good catch

PR #3904, opened 2023-04-03
Skip sync the RocksDB when no changes

hangc0276:
### Motivation
For the `LedgerMetadataIndex#removeDeletedLedgers` and `LedgerMetadataIndex#flush`, it will call ledgersDB sync whether the ledgersDB has changed or not. We can skip the sync call when nothing changed in the ledgersDB.

### Changes
- Check whether pendingLedgersUpdates is empty in `flush()` and `pendingDeletedLedgers` is empty in removeDeletedLedgers
- Move the `key.recycle()` in finally to cover keys leak when the ledgersDB operations throw an exception.


zymap (approved):

horizonzy (approved):
LGTM

wenbingshen (approved):
LGTM

merlimat (approved):

PR #3913, opened 2023-04-11
Enable PCBC completionObjects autoShrink to reduce memory usage and gc

wenbingshen:
### Motivation

PerChannelBookieClient completionObjects occupy a lot of heap space and cannot be recycled.
The figure below shows that the internal table array of ConcurrentOpenHashMap has used space size=0, but the array length is still 16384, and the memory overhead is 65552bytes.
image

image

ConcurrentOpenHashMap default DefaultConcurrencyLevel=16. We have hundreds of bookie nodes. Due to the feature of bookie polling and writing, the client and server have long connection characteristics, which will as a result, the memory usage of about 65552 * 16 * 1776 = 1.74GB cannot be recycled, and the space take up by these tables is all size=0 (The broker's owner topic has drifted to other brokers due to Full GC).
image

When the throughput of the pulsar cluster increases and the bookie cluster expands, these memory usage will also increase. Coupled with the unreasonable memory usage in other aspects of pulsar that we know, this will cause the pulsar broker to continuously generate full gc.

### Changes
I think adding autoShrink to completionObjects can reduce this part of memory usage and reduce the frequency of Full GC.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
@@ -187,7 +187,7 @@ public class PerChannelBookieClient extends ChannelInboundHandlerAdapter {
     final int startTLSTimeout;
 
     private final ConcurrentOpenHashMap<CompletionKey, CompletionValue> completionObjects =
-            ConcurrentOpenHashMap.<CompletionKey, CompletionValue>newBuilder().build();
+            ConcurrentOpenHashMap.<CompletionKey, CompletionValue>newBuilder().autoShrink(true).build();
 
     // Map that hold duplicated read requests. The idea is to only use this map (synchronized) when there is a duplicate
     // read request for the same ledgerId/entryId


horizonzy:
> ConcurrentOpenHashMap default DefaultConcurrencyLevel=16. We have hundreds of bookie nodes. Due to the feature of bookie polling and writing, the client and server have long connection characteristics, which will as a result, the memory usage of about 65552 * 16 * 1776 = 1.74GB cannot be recycled, and the space take up by these tables is all size=0 (The broker's owner topic has drifted to other brokers due to Full GC).

I have a question about this. The memory occupation is not about the array size. The key(CompletionKey) and the value(CompletionValue) is the occupier. As long as the key and the value is removed, the memory occupation will be decrease.
We make the array size autoShrink didn't help the GC

wenbingshen:
> I have a question about this. The memory occupation is not about the array size. The key(CompletionKey) and the value(CompletionValue) is the occupier. As long as the key and the value is removed, the memory occupation will be decrease. We make the array size autoShrink didn't help the GC

@horizonzy The layout of the array object in memory, when pointer compression is enabled, includes
array object header + length * 4 (reference) + length * (single element size)

We have an object array with a size of 16384, and the space it occupies in memory is = 8 + 4 + 4 + 4 * 16384 = 65552
In our pulsar broker, such an array would occupy about = 65552 * 16 * 1776 = 1.74GB

If we can turn on autoShrink, in the case of size=0, the array size will shrink to 24. 
And the space it occupies in memory is = 8 + 4 + 4 + 4 * 24 = 112
In our pulsar broker, such an array would occupy about = 112 * 16 * 1776 = 3108KB

This way we can reclaim a lot of space in memory.

horizonzy:
> In our pulsar broker, such an array would occupy about = 65552 * 16 * 1776 = 1.74GB

Why multiply 16.

wenbingshen:
> Why multiply 16.

@horizonzy ConcurrentOpenHashMap default DefaultConcurrencyLevel=16

image
image

horizonzy:
Thanks, I got it.

horizonzy (approved):
Nice improvement. LGTM

hangc0276 (approved):
Nice Catch!

wenbingshen:
@merlimat @eolivelli @dlg99 @zymap Can you help take a look at this pr. Thanks.

zymap (approved):

dlg99 (approved):
LGTM

PR #3918, opened 2023-04-14
Fix isWritable incorrect problem.

horizonzy:
### Motivation
The `isWritable` should be false in some places, correct it.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
@@ -631,11 +631,11 @@ public class PerChannelBookieClient extends ChannelInboundHandlerAdapter {
     public boolean isWritable() {
         return isWritable;
     }
-    
+
     public void makeWritable() {
         this.isWritable = true;
     }
-    
+
     public void makeUnWritable() {
         this.isWritable = false;
     }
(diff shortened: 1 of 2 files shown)


horizonzy:
@dlg99 cc.

hangc0276 (approved):

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
```diff
@@ -1097,7 +1097,7 @@ private void closeInternal(boolean permanent, boolean wait) {
             }
             toClose = channel;
             channel = null;
-            makeWritable();
+            makeUnWritable();
```
zymap, line 1100: The writable should be a reflection value of the Channel's isWritable. So we make it default to true. So it should always be true except when the channel is non-writable.

zymap (commented):
@michaeljmarshall Could you please take a look at this change?

dlg99 (changes requested):
I see couple of problems with the PR:

1. It claims the X is a problem, fixes X, does not explain what's the problem, what it affects etc. The PR changes existing test to match but does not add a test reproducing the problem it was actually fixing. I don't understand what led you to this change, did you just looked at he code and decided to modify it or there is actual problem you have encountered?

2. I don't remember the call chain that well, but II think the idea is that PCBC with disconnected channel must be writable. Otherwise whoever needs to write to the PCBC may decide to skip at non-writable instead of just writing and triggering reconnect.

Issue #3923, opened 2023-04-17
Add support to include the bookie sanity state in the bookie metrics.


rdhabalia:
### Motivation
Right now, Bookie metrics (Prometheus) don't show if Bookie's sanity is passing or not. and with container-based Prometheus bookie metrics which are used for monitoring, it will be really useful if metrics also show the bookie sanity state so, Prometheus can automatically scrape the bookie sanity state and one can set up monitoring on it. 
Therefore, add support to include the bookie sanity state in the bookie metrics.
For example:
```
URL: http://localhost:8000/metrics
:
# 1 shows bookie sanity is passing, 0 failed, -1 unknown
bookie_SERVER_SANITY 1
```

PR #3924, opened 2023-04-17
[stats][bookie] Add bookie sanity test metrics

rdhabalia:
### Motivation

It solves #3923 

Right now, Bookie metrics (Prometheus) don't show if Bookie's sanity is passing or not. and with container-based Prometheus bookie metrics which are used for monitoring, it will be really useful if metrics also show the bookie sanity state so, Prometheus can automatically scrape the bookie sanity state and one can set up monitoring on it. 
Therefore, add support to include the bookie sanity state in the bookie metrics.
For example:
```
URL: http://localhost:8000/metrics
:
# 1 shows bookie sanity is passing, 0 failed, -1 unknown
bookie_SERVER_SANITY 1
```

### Changes

- Add periodic task which performs sanity command on bookie and captures bookie sanity test periodically
- Reuse the `SanityTestCommnad` for this purpose but SanityTestCommand has lot of blocking IO which can cause thread blocking for metrics collection.
- Therefore, implement async method into `SanityTestCommnad` to avoid thread blocking and use it in metrics collection.

Master Issue: #3923

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStateManager.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStateManager.java
@@ -173,7 +173,7 @@ public class BookieStateManager implements StateManager {
         };
         statsLogger.registerGauge(SERVER_SANITY, serverSanityGauge);
         stateService.scheduleAtFixedRate(() -> {
-            if (forceReadOnly.get() || bookieStatus.isInReadOnlyMode()) {
+            if (isReadOnly()) {
                 sanityPassed.set(1);
                 return;
             }
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStateManager.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStateManager.java
@@ -23,22 +23,23 @@ package org.apache.bookkeeper.bookie;
 
 import static org.apache.bookkeeper.bookie.BookKeeperServerStats.BOOKIE_SCOPE;
 import static org.apache.bookkeeper.bookie.BookKeeperServerStats.CATEGORY_SERVER;
+import static org.apache.bookkeeper.bookie.BookKeeperServerStats.SERVER_SANITY;
 import static org.apache.bookkeeper.bookie.BookKeeperServerStats.SERVER_STATUS;
 
-import com.google.common.annotations.VisibleForTesting;
-import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import java.io.File;
 import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.net.UnknownHostException;
 import java.util.List;
 import java.util.concurrent.Callable;
-import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Supplier;
-import lombok.extern.slf4j.Slf4j;
+
 import org.apache.bookkeeper.conf.ServerConfiguration;
 import org.apache.bookkeeper.discover.BookieServiceInfo;
 import org.apache.bookkeeper.discover.RegistrationManager;
@@ -47,9 +48,15 @@ import org.apache.bookkeeper.stats.Gauge;
 import org.apache.bookkeeper.stats.NullStatsLogger;
 import org.apache.bookkeeper.stats.StatsLogger;
 import org.apache.bookkeeper.stats.annotations.StatsDoc;
+import org.apache.bookkeeper.tools.cli.commands.bookie.SanityTestCommand;
 import org.apache.bookkeeper.util.DiskChecker;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
+
+import lombok.extern.slf4j.Slf4j;
 
 /**
  * An implementation of StateManager.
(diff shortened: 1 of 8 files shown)


zymap (approved):

bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/commands/bookie/SanityTestCommand.java
```diff
@@ -81,59 +86,126 @@ public boolean apply(ServerConfiguration conf, SanityFlags cmdFlags) {
...
+                        return;
+                    }
+                    long lac = olh.getLastAddConfirmed();
+                    if (lac != (cmdFlags.entries - 1)) {
+                        lhFuture.completeExceptionally(new Exception("Invalid last entry found on ledger. expecting: "
+                                + (cmdFlags.entries - 1) + " -- found: " + lac));
+                        return;
+                    }
+                    lhFuture.complete(lh);
+                }, null);
```
hangc0276, line 152: We'd better add the `customMetadata` for the created ledger in order to find out orphan ledgers if the ledger delete failed.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStateManager.java
```diff
@@ -149,6 +160,30 @@ public Number getSample() {
...
+                return sanityPassed.get();
+            }
+        };
+        statsLogger.registerGauge(SERVER_SANITY, serverSanityGauge);
+        stateService.scheduleAtFixedRate(() -> {
+            if (forceReadOnly.get()) {
+                sanityPassed.set(1);
+                return;
+            }
+            SanityTestCommand.handleAsync(conf, new SanityTestCommand.SanityFlags()).thenAccept(__ -> {
```
hangc0276, line 180: Do we need to check whether the bookie is in ReadOnly mode? so that we can skip running the Sanity test.

hangc0276 (approved):
I'm not sure whether we need a flag to turn on/off the sanity test

bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/commands/bookie/SanityTestCommand.java
rdhabalia: this is existing code where we don't have handling for failed delete. this change only changes blocking IO to async to avoid thread-blocking while checking sanity. 
So, if we need any enhancement into existing codebase then we can address it in separate PR.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStateManager.java
rdhabalia: yes, it's checking at line#176

hangc0276: Line #176 only checks forceReadOnly, not check the bookie runs into ReadOnly due to the disk being full.

bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/commands/bookie/SanityTestCommand.java
hangc0276: This PR makes the case worse. The existing code only triggers the sanity test by the bookkeeper shell, but from this PR on, the bookie will trigger sanity test every 60s

rdhabalia: well running sanity is executing one operation of what bookie suppose to do with high throughput : creating ledger/add entries/read entries/delete ledger. so, one operation per minute should be fine. the only thing is we should not have any blocking IO and that's the reason, I have changed sanityTestCommand implementation with async behavior.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStateManager.java
rdhabalia: i see. sure, let me add readOnly check as well.

bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/commands/bookie/SanityTestCommand.java
rdhabalia: > I'm not sure whether we need a flag to turn on/off the sanity test

sure, we can add the flag and keep it disable by default and user can enable based on the need.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStateManager.java
rdhabalia: done. added the read-only check.

bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/commands/bookie/SanityTestCommand.java
rdhabalia: added the flag `sanityCheckMetricsEnabled` and disabled the behavior by default.

rdhabalia:
@hangc0276 comments are addressed. can you please merge if it looks good to you.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStateManager.java
rdhabalia: sure, done.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieStateManager.java
```diff
@@ -149,6 +160,30 @@ public Number getSample() {
...
+                return sanityPassed.get();
+            }
+        };
+        statsLogger.registerGauge(SERVER_SANITY, serverSanityGauge);
+        stateService.scheduleAtFixedRate(() -> {
+            if (isReadOnly()) {
+                sanityPassed.set(1);
+                return;
+            }
+            SanityTestCommand.handleAsync(conf, new SanityTestCommand.SanityFlags()).thenAccept(__ -> {
```
horizonzy, line 180: It looks like the config `sanityCheckMetricsEnabled` doesn't work. The scheduled task only works when `sanityCheckMetricsEnabled` is `true`.

Issue #3931, opened 2023-04-24
Stuck recovery read when bookie is not available.


hanmz:
**BUG REPORT**

***Describe the bug***
When recovery reads, we should return BookieHandleNotAvailableException immediately if bookie is not available. This is because, in the case of network isolation, reading a faulty bookie node will time out, which will cause a slow recovery or even timeout.

***Additional context***

<img width="1034" alt="image" src="https://user-images.githubusercontent.com/16524922/233945206-39b89591-db26-43ea-b4a9-7f06f371f98f.png">

PR #3932, opened 2023-04-24
Improve recovery read when bookie is not available.

hanmz:
### Motivation
It solves https://github.com/apache/bookkeeper/issues/3931

### Changes
Return BookieHandleNotAvailableException immediately if bookie is not available when recovery read. In the case of network isolation, reading a faulty bookie node will time out, which will cause a slow recovery or even timeout.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
@@ -576,6 +576,14 @@ class PendingReadOp implements ReadEntryCallback, Runnable {
         }
 
         if (isRecoveryRead) {
+            // Return BookieHandleNotAvailableException immediately if bookie is not available when recovery read.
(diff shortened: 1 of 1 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/PendingReadOp.java
```diff
@@ -576,6 +576,14 @@ void sendReadTo(int bookieIndex, BookieId to, LedgerEntryRequest entry) throws I
         }
 
         if (isRecoveryRead) {
+            // Return BookieHandleNotAvailableException immediately if bookie is not available when recovery read.
+            // In the case of network isolation, reading a faulty bookie node will time out,
+            // which will cause a slow recovery or even timeout.
+            if (clientCtx.getBookieWatcher().isBookieUnavailable(to)) {
```
merlimat, line 582: One issue here is if the bookie is partitioned from ZK though it's effectively online and reachable from the BK client. This change will make it fail while today it's succeeding.

PR #3938, opened 2023-04-25
drop invalid entryFormat arg from shell command

wenbingshen:
### Motivation

According to the configuration of ENTRY_FORMATTER_CLASS in the bookkeeper shell command, it will never take effect, we should remove it,

We can use -entryformat or configure in bookie config to take effect.

### DIFF
--- a/bin/bookkeeper
+++ b/bin/bookkeeper
@@ -87,7 +87,6 @@ Environment variables:
    BOOKIE_ZK_CONF         Configuration file for zookeeper (default: $DEFAULT_ZK_CONF)
    BOOKIE_EXTRA_OPTS      Extra options to be passed to the jvm
    BOOKIE_EXTRA_CLASSPATH Add extra paths to the bookkeeper classpath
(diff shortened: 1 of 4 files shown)


wenbingshen:
rerun failure checks

wenbingshen:
rerun failure checks

hangc0276 (approved):

Shoothzj (approved):

Shoothzj (approved):

StevenLuMT (changes requested):
Will there be compatibility issues, will it be compatible with old usage? i'm a bit worried about this

wenbingshen:
> Will there be compatibility issues, will it be compatible with old usage? i'm a bit worried about this

@StevenLuMT I don't think there is an old compatibility issue here, because `ENTRY_FORMATTER_CLASS` is set as a jvm system property value, and It is not referenced anywhere in the bookkeeper's existing source code, even though the system the attribute has been configured, but it will not take effect. It is safe for us to delete it. PTAL. Thanks.

StevenLuMT (approved):
great

PR #3946, opened 2023-05-06
When the executor has been shut down, do not schedule task

wenbingshen:
### Motivation

We should return, when an exception is encountered, the executor should not be scheduled tasks after it has been shutdown.


wenbingshen:
rerun failure checks

hangc0276 (approved):

Shoothzj (approved):

StevenLuMT (approved):
good catch

Issue #2220, opened 2019-12-18
Exception while replaying journals, shutting down


andrekramer1:
We're running Apache Pulsar with bookkeeper on a Kubernetes cluster and we have bookies (2/3) failing to restart with the following exception:
09:49:25.511 [main] ERROR org.apache.bookkeeper.bookie.Bookie - Exception while replaying journals, shutting down
java.io.IOException: org.apache.bookkeeper.bookie.BookieException$OperationRejectedException
        at org.apache.bookkeeper.bookie.Bookie$4.process(Bookie.java:895) ~[org.apache.bookkeeper-bookkeeper-server-4.9.2.jar:4.9.2]
        at org.apache.bookkeeper.bookie.Journal.scanJournal(Journal.java:821) ~[org.apache.bookkeeper-bookkeeper-server-4.9.2.jar:4.9.2]
        at org.apache.bookkeeper.bookie.Journal.replay(Journal.java:866) ~[org.apache.bookkeeper-bookkeeper-server-4.9.2.jar:4.9.2]
        at org.apache.bookkeeper.bookie.Bookie.readJournal(Bookie.java:901) ~[org.apache.bookkeeper-bookkeeper-server-4.9.2.jar:4.9.2]
        at org.apache.bookkeeper.bookie.Bookie.start(Bookie.java:922) ~[org.apache.bookkeeper-bookkeeper-server-4.9.2.jar:4.9.2]
        at org.apache.bookkeeper.proto.BookieServer.start(BookieServer.java:141) ~[org.apache.bookkeeper-bookkeeper-server-4.9.2.jar:4.9.2]
        at org.apache.bookkeeper.server.service.BookieService.doStart(BookieService.java:58) ~[org.apache.bookkeeper-bookkeeper-server-4.9.2.jar:4.9.2]
        at org.apache.bookkeeper.common.component.AbstractLifecycleComponent.start(AbstractLifecycleComponent.java:78) ~[org.apache.bookkeeper-bookkeeper-common-4.9.2.jar:4.9.2]
        at org.apache.bookkeeper.common.component.LifecycleComponentStack.lambda$start$2(LifecycleComponentStack.java:113) ~[org.apache.bookkeeper-bookkeeper-common-4.9.2.jar:4.9.2]
        at com.google.common.collect.ImmutableList.forEach(ImmutableList.java:407) [com.google.guava-guava-25.1-jre.jar:?]
[...]

PR #3437, opened 2022-08-01
Issue #2220: support skip invalid record in recovery

leizhiyuan:
Descriptions of the changes in this PR:

### Motivation
fix #2220

### Changes

(Describe: what changes you have made)

Master Issue: #<master-issue-number>

> ---
> In order to uphold a high standard for quality for code contributions, Apache BookKeeper runs various precommit
> checks for pull requests. A pull request can only be merged when it passes precommit checks.
>
> ---
> Be sure to do all of the following to help us incorporate your contribution
> quickly and easily:
>
> If this PR is a BookKeeper Proposal (BP):
>
> - [ ] Make sure the PR title is formatted like:
>     `<BP-#>: Description of bookkeeper proposal`
>     `e.g. BP-1: 64 bits ledger is support`
> - [ ] Attach the master issue link in the description of this PR.
> - [ ] Attach the google doc link if the BP is written in Google Doc.
>
> Otherwise:
> 
> - [ ] Make sure the PR title is formatted like:
>     `<Issue #>: Description of pull request`
>     `e.g. Issue 123: Description ...`
> - [ ] Make sure tests pass via `mvn clean apache-rat:check install spotbugs:check`.
> - [ ] Replace `<Issue #>` in the title with the actual Issue number.
> 
> ---

PR #3956, opened 2023-05-15
Support skip invalid journal record in replying journal stage

hangc0276:
### Motivation
This PR is generated from https://github.com/apache/bookkeeper/pull/3437
Fix https://github.com/apache/bookkeeper/issues/2220

When we use `kill -9` command to stop one bookie pod, the journal file may be broken. If we try to start up the bookie pod again, the pod will startup to fail with the following exception.
```
10:15:55.026 [main] ERROR org.apache.bookkeeper.bookie.Bookie - Exception while replaying journals, shutting down
java.io.IOException: Invalid record found with negative length -448299468
        at org.apache.bookkeeper.bookie.Journal.scanJournal(Journal.java:821) ~[org.apache.bookkeeper-bookkeeper-server-4.12.0.jar:4.12.0]
        at org.apache.bookkeeper.bookie.Bookie.replay(Bookie.java:945) ~[org.apache.bookkeeper-bookkeeper-server-4.12.0.jar:4.12.0]
        at org.apache.bookkeeper.bookie.Bookie.readJournal(Bookie.java:911) ~[org.apache.bookkeeper-bookkeeper-server-4.12.0.jar:4.12.0]
        at org.apache.bookkeeper.bookie.Bookie.start(Bookie.java:965) ~[org.apache.bookkeeper-bookkeeper-server-4.12.0.jar:4.12.0]
        at org.apache.bookkeeper.proto.BookieServer.start(BookieServer.java:156) ~[org.apache.bookkeeper-bookkeeper-server-4.12.0.jar:4.12.0]
        at org.apache.bookkeeper.server.service.BookieService.doStart(BookieService.java:68) ~[org.apache.bookkeeper-bookkeeper-server-4.12.0.jar:4.12.0]
        at org.apache.bookkeeper.common.component.AbstractLifecycleComponent.start(AbstractLifecycleComponent.java:83) ~[org.apache.bookkeeper-bookkeeper-common-4.12.0.jar:4.12.0]
        at org.apache.bookkeeper.common.component.LifecycleComponentStack.lambda$start$4(LifecycleComponentStack.java:144) ~[org.apache.bookkeeper-bookkeeper-common-4.12.0.jar:4.12.0]
        at com.google.common.collect.ImmutableList.forEach(ImmutableList.java:406) [com.google.guava-guava-30.1-jre.jar:?]
        at org.apache.bookkeeper.common.component.LifecycleComponentStack.start(LifecycleComponentStack.java:144) [org.apache.bookkeeper-bookkeeper-common-4.12.0.jar:4.12.0]
        at org.apache.bookkeeper.common.component.ComponentStarter.startComponent(ComponentStarter.java:85) [org.apache.bookkeeper-bookkeeper-common-4.12.0.jar:4.12.0]
        at org.apache.bookkeeper.server.Main.doMain(Main.java:234) [org.apache.bookkeeper-bookkeeper-server-4.12.0.jar:4.12.0]
        at org.apache.bookkeeper.server.Main.main(Main.java:208) [org.apache.bookkeeper-bookkeeper-server-4.12.0.jar:4.12.0]
10:15:55.134 [main] INFO  org.apache.zookeeper.ZooKeeper - Session: 0x200064a14681be2 closed
```

We should have a way to make the bookie pod startup instead of decommissioning it.

### Changes
1. Add a configuration to allow skipping the invalid entries when journal replying to journal files.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
@@ -862,16 +862,17 @@ public class Journal extends BookieCriticalThread implements CheckpointSource {
                     scanner.process(journalVersion, offset, recBuff);
                 }
             }
+            return recLog.fc.position();
         } catch (IOException e) {
             if (skipInvalidRecord) {
                 LOG.warn("Failed to parse journal file, and skipInvalidRecord is true, skip this journal file reply");
             } else {
                 throw e;
             }
+            return recLog.fc.position();
         } finally {
             recLog.close();
         }
-        return recLog.fc.position();
     }
 
     /**
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
@@ -829,27 +829,31 @@ public class Journal extends BookieCriticalThread implements CheckpointSource {
                 }
                 boolean isPaddingRecord = false;
                 if (len < 0) {
-                    if (len == PADDING_MASK && journalVersion >= JournalChannel.V5) {
-                        // skip padding bytes
-                        lenBuff.clear();
-                        fullRead(recLog, lenBuff);
-                        if (lenBuff.remaining() != 0) {
-                            break;
+                    try {
+                        if (len == PADDING_MASK && journalVersion >= JournalChannel.V5) {
+                            // skip padding bytes
+                            lenBuff.clear();
+                            fullRead(recLog, lenBuff);
+                            if (lenBuff.remaining() != 0) {
+                                break;
+                            }
+                            lenBuff.flip();
+                            len = lenBuff.getInt();
+                            if (len == 0) {
+                                continue;
+                            }
+                            isPaddingRecord = true;
+                        } else {
+                            LOG.error("Invalid record found with negative length: {}", len);
+                            throw new IOException("Invalid record found with negative length " + len);
                         }
-                        lenBuff.flip();
-                        len = lenBuff.getInt();
-                        if (len == 0) {
-                            continue;
+                    } catch (IOException e) {
+                        if (skipInvalidRecord) {
+                            LOG.warn("Invalid record found with negative length: {},because of " +
+                                    "skipInvalidRecord is true,we skip the next data", len);
+                        } else {
+                            throw e;
                         }
-                        isPaddingRecord = true;
-                    } else if (skipInvalidRecord){
-                        LOG.warn("Invalid record found with negative length: {},because of " +
-                                "skipInvalidRecord is true,we skip the next data", len);
-                        break;
-                    }
-                    else {
-                        LOG.error("Invalid record found with negative length: {}", len);
-                        throw new IOException("Invalid record found with negative length " + len);
                     }
                 }
                 recBuff.clear();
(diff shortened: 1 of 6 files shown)


hangc0276:
@frankjkelly I have created a new PR to fix the bookie that can't startup due to the journal reply failing.

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
```diff
@@ -4010,6 +4012,25 @@ public boolean isDataIntegrityStampMissingCookiesEnabled() {
...
+        this.setProperty(SKIP_REPLAY_JOURNAL_INVALID_RECORD,
+                Boolean.toString(skipReplayJournalInvalidRecord));
+        return this;
+    }
+
+    /**
+     * @see #isSkipReplayJournalInvalidRecord .
+     */
+    public boolean isSkipReplayJournalInvalidRecord() {
+        return this.getBoolean(SKIP_REPLAY_JOURNAL_INVALID_RECORD, false);
```
eolivelli, line 4031: the previous behaviour was to skip errors, we should keep the same value

eolivelli (changes requested):
Overall LGTM, but I would set the property to true by default

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
hangc0276: The previous behavior is to throw the exception out and block the bookie start-up. We should keep the same behavior.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -827,22 +828,32 @@ public long scanJournal(long journalId, long journalPos, JournalScanner scanner)
...
-                            break;
+                    try {
+                        if (len == PADDING_MASK && journalVersion >= JournalChannel.V5) {
+                            // skip padding bytes
+                            lenBuff.clear();
+                            fullRead(recLog, lenBuff);
+                            if (lenBuff.remaining() != 0) {
+                                break;
+                            }
+                            lenBuff.flip();
```
wenbingshen, line 839: line 820 may throw IOException, this may cause bookie startup failed ==>  fullRead(recLog, lenBuff);
In this PR, line 835 also may throw IOException, because this future the bookie can continue starting.
Same behavior here, two results, right?

https://github.com/apache/bookkeeper/pull/3956/files#diff-0a79d54e3dfd4bab1d85510ef957f8f4b7ab6f6acd09aadd634b4d7bbee6e1f4R815-R821
```java
            while (true) {
                // entry start offset
                long offset = recLog.fc.position();
                // start reading entry
                lenBuff.clear();
                fullRead(recLog, lenBuff);
```

hangc0276: I caught all the `fullRead` throw exceptions.

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
hangc0276: Even though the previous default behavior is `false`, I changed the default value to `true`, because I think it's valuable to skip the the invalid record instead of make the bookie node can't start up.

wenbingshen (approved):
LGTM

wenbingshen:
rerun failure checks

frankjkelly:
Hi folks - thanks for the efforts here - what are the remaining issues on this? Thanks!

zymap:
@eolivelli Would you like to take another look at this PR?

hangc0276:
> Hi folks - thanks for the efforts here - what are the remaining issues on this? Thanks!

@frankjkelly I addressed all the comments, please help take a look, thanks.

eolivelli (approved):
+1

vitalii-buchyn-exa:
hello community,
any documentation update for this feature?

PR #3958, opened 2023-05-18
[improve] Limit decommissionBookie http request

wenbingshen:
### Motivation

1. DecommissionService may receive multiple repeated decommission tasks, because it does not check the current task execution progress, which will bring dos and ddos attack vulnerabilities.
2. After DecommissionService receives a task, it cannot tell the task initiator whether the task is still in progress.


wenbingshen:
rerun failure checks

PR #3961, opened 2023-05-19
When the thread has been interrupted, there is no need for waitBackOffTime

wenbingshen:
### Motivation

When the thread has been interrupted, there is no need for waitBackOffTime, we can directly shutdown ReplicationWorker.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/ReplicationWorker.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/ReplicationWorker.java
@@ -265,12 +265,12 @@ public class ReplicationWorker implements Runnable {
             } catch (UnavailableException e) {
                 LOG.error("UnavailableException "
                         + "while replicating fragments", e);
-                waitBackOffTime(rwRereplicateBackoffMs);
                 if (Thread.currentThread().isInterrupted()) {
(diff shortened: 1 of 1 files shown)


hangc0276 (approved):

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/ReplicationWorker.java
```diff
@@ -265,12 +265,12 @@ public void run() {
             } catch (UnavailableException e) {
                 LOG.error("UnavailableException "
                         + "while replicating fragments", e);
-                waitBackOffTime(rwRereplicateBackoffMs);
                 if (Thread.currentThread().isInterrupted()) {
                     LOG.error("Interrupted  while replicating fragments");
                     shutdown();
                     return;
                 }
+                waitBackOffTime(rwRereplicateBackoffMs);
```
zymap, line 273: We already handled the InterruptedException at line 251. It looks like there want to check if the thread is interrupted after waiting the time. In the waitBackOffTime, it may interrupt the current thread.

zymap (changes requested):

hangc0276:
@wenbingshen Close this PR. If you have more updates, please reopne it.

PR #3962, opened 2023-05-19
[fix] Fix wrong key for getHttpServerTrustStorePassword 

### LINKED ISSUE #3639: ServerConfiguration returns incorrect property for httpServerTrustStorePassword

wenbingshen:
### Motivation

Here's a second approach based on this PR #3643 
Fix #3639 

### Changes

1. Add a configuration to compatible with historical configuration.
2. This is a compatibility fix for the previous error of pointing HttpServerTrustStorePassword to HttpServerKeyStorePassword.
3. If you want to fix this problem, you can set this parameter to false.
4. Default value is true, keep the original behavior.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
@@ -278,6 +278,8 @@ public class ServerConfiguration extends AbstractConfiguration<ServerConfigurati
     protected static final String HTTP_SERVER_KEY_STORE_PASSWORD = "httpServerKeyStorePassword";
     protected static final String HTTP_SERVER_TRUST_STORE_PATH = "httpServerTrustStorePath";
     protected static final String HTTP_SERVER_TRUST_STORE_PASSWORD = "httpServerTrustStorePassword";
+    protected static final String HTTP_SERVER_TRUST_ROLLBACK_KEY_STORE_PASSWORD =
+            "httpServerTrustRollbackKeyStorePassword";
 
     // Lifecycle Components
     protected static final String EXTRA_SERVER_COMPONENTS = "extraServerComponents";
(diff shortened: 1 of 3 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
```diff
@@ -3698,7 +3700,7 @@ public ServerConfiguration setHttpServerTrustStorePath(String trustStorePath) {
      * @return http server trustStore password
      */
     public String getHttpServerTrustStorePassword() {
-        return getString(HTTP_SERVER_KEY_STORE_PASSWORD);
+        return getString(HTTP_SERVER_TRUST_STORE_PASSWORD);
```
zymap, line 3703: An easier way to keep the compatibility and avoid introducing a new configuration:
```
return getString(HTTP_SERVER_TRUST_STORE_PASSWORD, getString(HTTP_SERVER_KEY_STORE_PASSWORD))
```

wenbingshen: We need to consider that the user configures the wrong trust store password, and then uses the correct key store password to successfully start the bookie. After the upgrade, the wrong trust store password takes effect, causing the bookie upgrade to fail?

This way like PR #3643, does not seem to be fully compatible.

Maybe we need more people to see this PR.

shoothzj:
closed by #4301

PR #3965, opened 2023-05-20
[fix] Keep entryLogMeta when entryLogFile deletion fails

wenbingshen:
### Motivation

When the DefaultEntryLogger fails to delete the entryLogFile, it will still clear the corresponding entryLogId from the entryLogMetaMap. When the entryLogFile file exists, the GC thread will scan all the entryLogFile files and rebuild the LogMeta information of the entryLogFile that is not in the entryLogMetaMap when the next GC is executed. This will bring to unnecessary additional disk IO, it is necessary to scan the entire entryLogFile file in severe cases.

```java
    public void runWithFlags(boolean force, boolean suspendMajor, boolean suspendMinor) {
            ......
            extractMetaFromEntryLogs();  // 1
            ......
    }
```

```java
    public EntryLogMetadata getEntryLogMetadata(long entryLogId, AbstractLogCompactor.Throttler throttler)
        throws IOException {
        // First try to extract the EntryLogMetadata from the index, if there's no index then fallback to scanning the
        // entry log
        try {
            return extractEntryLogMetadataFromIndex(entryLogId);    2 <== here scan from index
        } catch (Exception e) {
            LOG.info("Failed to get ledgers map index from: {}.log : {}", entryLogId, e.getMessage());

            // Fall-back to scanning
            return extractEntryLogMetadataByScanning(entryLogId, throttler);    3 <== here scan all entryLogFile
        }
    }

```

Another point is that we should only update `reclaimedSpaceViaDeletes` after successfully deleting the entryLogFile file.


horizonzy (approved):
LGTM. You should modify the PR description to match the changes.

horizonzy:
> When the DefaultEntryLogger fails to delete the entryLogFile, it will still clear the corresponding entryLogId from the entryLogMetaMap. 

For this part, I have noticed that it has already been covered.

horizonzy:
> > > When the DefaultEntryLogger fails to delete the entryLogFile, it will still clear the corresponding entryLogId from the entryLogMetaMap.
> > 
> > 
> > For this part, I have noticed that it has already been covered.
> 
> @horizonzy No, the previous logic, when the file deletion fails, the method will always return true. https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/DefaultEntryLogger.java#L530
> 
> ```java
>     public boolean removeEntryLog(long entryLogId) {
>         removeFromChannelsAndClose(entryLogId);
>         File entryLogFile;
>         try {
>             entryLogFile = findFile(entryLogId);
>         } catch (FileNotFoundException e) {
>             LOG.error("Trying to delete an entryLog file that could not be found: "
>                     + entryLogId + ".log");
>             return false;
>         }
>         if (!entryLogFile.delete()) {
>             LOG.warn("Could not delete entry log file {}", entryLogFile);  <== here always return true
>         }
>         return true;
>     }
> ```

Got it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -755,8 +759,9 @@ protected void extractMetaFromEntryLogs() throws EntryLogMetadataMapException {
                     // ledgers anymore.
                     // We can remove this entry log file now.
                     LOG.info("Deleting entryLogId {} as it has no active ledgers!", entryLogId);
-                    removeEntryLog(entryLogId);
-                    gcStats.getReclaimedSpaceViaDeletes().addCount(entryLogMeta.getTotalSize());
+                    if (removeEntryLog(entryLogId)) {
+                        gcStats.getReclaimedSpaceViaDeletes().addCount(entryLogMeta.getTotalSize());
+                    }
```
dlg99, line 764: we should increment some stats (reclaimFailedToDelete) in case of error/else block to enable monitoring of such cases

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -497,8 +497,9 @@ private void doGcEntryLogs() throws EntryLogMetadataMapException {
                     // ledgers anymore.
                     // We can remove this entry log file now.
                     LOG.info("Deleting entryLogId {} as it has no active ledgers!", entryLogId);
-                    removeEntryLog(entryLogId);
-                    gcStats.getReclaimedSpaceViaDeletes().addCount(meta.getTotalSize());
+                    if (removeEntryLog(entryLogId)) {
+                        gcStats.getReclaimedSpaceViaDeletes().addCount(meta.getTotalSize());
```
dlg99, line 501: we should increment some stats (reclaimFailedToDelete) in case of error/else block to enable monitoring of such cases

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/DefaultEntryLogger.java
```diff
@@ -528,6 +528,7 @@ public boolean removeEntryLog(long entryLogId) {
         }
         if (!entryLogFile.delete()) {
```
dlg99, line 529: above we return false in case of file not found
```java
        } catch (FileNotFoundException e) {
            LOG.error("Trying to delete an entryLog file that could not be found: "
                    + entryLogId + ".log");
            return false;
        }
```
Would it make sense to simply log and return true; after all, we wanted to delete the entry log, and let the `entryLogMetaMap.remove` run.
Alternatively, throw IllegalStateException to force an investigation into how we got into such a state (is there a concurrent process that could delete the entry log?) though this feels like an overkill.

dlg99 (commented):
overall LGTM with couple of suggestions

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
wenbingshen: make sense to me. I will add the stats for this.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/DefaultEntryLogger.java
wenbingshen: I think the first method is better, simply log and return true, throw IllegalStateException changes the previous behavior, users may be less willing to care about the reason for its non-existence when cleaning up the outdated entryLogFile. And keep the original behavior  can avoid introducing possible bugs. Here, we just make `entryLogMetaMap.remove` work

wenbingshen:
@dlg99 I have addressed your comments. PTAL. Thanks.

dlg99 (approved):
LGTM, thank you!
CI is still running so let's wait for the tests to pass before merging.

hangc0276 (approved):
Is it easy to add a test to cover this change?

wenbingshen:
> Is it easy to add a test to cover this change?

@hangc0276 In the modification of PR, there are two cases.
In case1, we return true when the entryLog file does not exist, which helps us remove the corresponding meta information from the entryLogMeta of the GC object. For the entryLogFile file to be deleted, the impact of the change is small.

In case2, when we perform file deletion operations, deletion failures may occur. This result depends more on the impact between multiple processes, or deletion failures caused by misoperations by system maintenance personnel, such as permissions, file locks changed, etc. It is not easy to simulate this situation within a single bookie process.

Executing the system `remove` function in java native method has a `boolean` return value.

```java
JNIEXPORT jboolean JNICALL
Java_java_io_UnixFileSystem_delete0(JNIEnv *env, jobject this,
                                    jobject file)
{
    jboolean rv = JNI_FALSE;

    WITH_FIELD_PLATFORM_STRING(env, file, ids.path, path) {
        if (remove(path) == 0) {
            rv = JNI_TRUE;
        }
    } END_PLATFORM_STRING(env, path);
    return rv;
}
```

StevenLuMT (approved):
Great work

wenbingshen:
4+ approved, merging...

PR #3968, opened 2023-05-24
Added support for OpenTelemetry metrics backend

merlimat:
### Motivation

Added support for collecting and exporting BK metrics through OpenTelemetry SDK. 

The instruments are using counters and histogram. Summaries with pre-computed quantiles are not used anymore. Histograms have a much lower overhead in the bookie process.

The SDK is set to use the autoconfigure. The exporter can be configured through env variables as described in https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/autoconfigure/README.md 

eg: simplest case to export in Prometheus format: 

```
OTEL_METRICS_EXPORTER=prometheus
```

This will make the metrics available on `http://0.0.0.0:9464/metrics`


eolivelli (approved):
+1

Shoothzj (approved):

stats/bookkeeper-stats-providers/otel-metrics-provider/src/main/java/org/apache/bookkeeper/stats/otel/OtelStatsLogger.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    @Override
+    public OpStatsLogger getOpStatsLogger(String name) {
+        return provider.opStats.computeIfAbsent(scopeContext(name), sc -> new OtelOpStatsLogger(provider.meter, sc));
+    }
+
+    @Override
+    public OpStatsLogger getThreadScopedOpStatsLogger(String name) {
+        return getOpStatsLogger(name);
```
asafm, line 45: In the original implementation, they made a great effort to take attributes from the current thread. Isn't it helpful?

stats/bookkeeper-stats-providers/otel-metrics-provider/src/main/java/org/apache/bookkeeper/stats/otel/OtelMetricsProvider.java
```diff
@@ -0,0 +1,150 @@
...
+    public void stop() {
+    }
+
+    @Override
+    public StatsLogger getStatsLogger(String scope) {
+        return new OtelStatsLogger(this, scope, Attributes.empty());
+    }
+
+    /*
+     * Try to get Netty counter of used direct memory. This will be correct, unlike the JVM values.
```
asafm, line 123: Did you ever find out why the JVM report lies? :)

stats/bookkeeper-stats-providers/otel-metrics-provider/src/main/java/org/apache/bookkeeper/stats/otel/OtelStatsLogger.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    @Override
+    public Counter getThreadScopedCounter(String name) {
+        return getCounter(name);
+    }
+
+    @Override
+    public <T extends Number> void registerGauge(String name, Gauge<T> gauge) {
+        provider.meter.gaugeBuilder(completeName(name))
```
asafm, line 60: This should be async upDownCounter

PR #3959, opened 2023-05-18
Avoid compaction to trigger extra flushes DbLedgerStorage

merlimat:
### Motivation

This PR solves the problem described #3940. 

Because compaction is triggering a flush after each entry log file is compacted, the flush on the DbLedgerStorage causes the write cache to be underutilized and ultimately it will slow down the writes.

### Changes

The call to `flush()` in `updateEntriesLocations()` was there to ensure that every entry log file we might be compacting has the indexes already flushed into RocksDB. If that's not the case, the compaction might first write the new location and the regular flush will override it with the old location, to a file that is already deleted.

In practice, we don't actually need to trigger a new flush for that to be valid, we just need to ensure that if there is a pending `flush()`, we wait until it's done. If there are no pending flushes, we're also good to go.

PR #2963, opened 2021-12-25
Add throttle for rebuild entryMetadataMap

hangc0276:
### Motivation
When a bookie restart, the garbageCollectorThread will rebuild entryMetadataMap from all the entry log files in ledger directory. For normal case, it will extract the EntryLogMetadata from the index in entry log file. However, if there's no index, then fallback to scanning the entry log file. 

In user's production environment, the log files without index occupied 4%. The total entry log files is 80000, and the log files without index is 3000. The default entry log file size is 2GB, and the garbageCollectorThread will read 3000 * 2GB = 6TB data without speed limit, which will cause ledger disk IO util runs high for dozens of minutes and affect ledger read and write latency.

### Modification
1. Add read speed rate limiter for scanning entry log file in entryMetadataMap rebuild.

PR #3976, opened 2023-05-31
[improve] add read rate limter for compaction

HQebupt:
### Motivation
The purpose of this feature is to provide more flexibility and control over the compaction process in the Bookkeeper server by allowing users to configure the read rate of compaction according to their specific requirements.
It adds support for configuring the read rate of compaction in bytes per second.

In our production environment of Bookkeeper, we have set compactionRateByBytes=30M. However, we have observed that it only limits the write speed to disk and does not limit the read speed from disk. In our environment, the continuous saturation of disk IO for reads in compaction is causing issues with data consumption from the disk.

```
// Set the rate at which compaction will readd entries. The unit is bytes added per second.
 compactionRateByBytes=31457280
```

image

image

### Changes

- In the `GarbageCollectorThread` class, a new field `compactionReadByteRateLimiter` of type RateLimiter is added to control the read rate of compaction.
- In the `ServerConfiguration` class, a new configuration property `COMPACTION_READ_RATE_BY_BYTES` is defined to specify the read rate of compaction in bytes per second. 

Master Issue: #[8](https://github.com/HQebupt/bookkeeper/pull/8)

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
@@ -25,6 +25,7 @@ import static org.apache.bookkeeper.util.BookKeeperConstants.METADATA_CACHE;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Strings;
+import com.google.common.util.concurrent.RateLimiter;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import io.netty.util.concurrent.DefaultThreadFactory;
 import java.io.IOException;
@@ -123,6 +124,7 @@ public class GarbageCollectorThread implements Runnable {
 
     private static final AtomicLong threadNum = new AtomicLong(0);
     final AbstractLogCompactor.Throttler throttler;
+    private final RateLimiter compactionReadByteRateLimiter;
 
     /**
      * Create a garbage collector thread.
@@ -266,6 +268,7 @@ public class GarbageCollectorThread implements Runnable {
                + majorCompactionThreshold + ", interval=" + majorCompactionInterval);
 
         lastMinorCompactionTime = lastMajorCompactionTime = System.currentTimeMillis();
+        compactionReadByteRateLimiter = RateLimiter.create(conf.getCompactionReadRateByBytes());
     }
 
     private EntryLogMetadataMap createEntryLogMetadataMap() throws IOException {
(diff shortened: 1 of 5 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -619,9 +622,10 @@ void doCompactEntryLogs(double threshold, long maxTimeMillis) throws EntryLogMet
                                 meta.getEntryLogId(), meta.getUsage(), threshold);
                     }
 
-                    long priorRemainingSize = meta.getRemainingSize();
+                    long compactSize = meta.getTotalSize() - meta.getRemainingSize();
+                    compactionReadByteRateLimiter.acquire((int) (compactSize));
                     compactEntryLog(meta);
```
wenbingshen, line 627: If I understand correctly, when the compact is in progress, if compactionRateBytes is configured, the compact will be limited when writing the disk. Since read and write are processed strictly serially, the next read needs to wait for the write operation to complete. Does it mean that read is also limited to compactionRateByBytes？

I suspect that your problem here is because the disk read rate is too fast caused by getting entryLogMeta from entryLogFile, you can look at this PR: #2963 

@hangc0276 Please help take a look. Thanks.

wenbingshen: @HQebupt Can you search in your bookie service log to see if there is the following log when the read IO is high: “Extracting entry log meta from entryLogId”

StevenLuMT: Good catch, please see this comment @HQebupt

HQebupt: @wenbingshen It did not help after I cherry pick https://github.com/apache/bookkeeper/pull/2963. We can discuss it when you are available tomorrow.
```
// Set the rate at which compaction will readd entries. The unit is bytes added per second.
 compactionRateByBytes=31457280
```
image

image

wenbingshen:
> Please check the following two PRs. #3959, #3192

@HQebupt I think compaction high read IO can be solved by #2963 #3192 and the high write IO can be solved by #3959. Please look at @hangc0276‘s comments.

HQebupt:
It works, thank you @wenbingshen @hangc0276 @StevenLuMT  I closed the PR.

Issue #3971, opened 2023-05-25
[Bug] repaired not adhering placement ledger feature can not repair opened ledger


TakaHiR07:
**BUG REPORT**

I also do a disaster recovery test, and try to use this feature to repair ledger. https://github.com/apache/bookkeeper/pull/3359

But it can not repair the ledger of pulsar "__change_event_" topic, because its ledger is opened status.

***Expected behavior***

Now my solution is also repair the opened ledger, and ensure not write the opened ledger by do not modify the topic metadata. 

But I wonder if there have better way ? 

By the way, now we can not repair not adhering RegionAware placement ledger. Can we add a config minNumRegionsPerWriteQuorum, just as minNumRacksPerWriteQuorum, to ensure quorum must be distributed in at least 3 regions ?

hangc0276:
@horizonzy Please help take a look, thanks.

horizonzy:
> @horizonzy Please help take a look, thanks.

Fine

horizonzy:
> Now my solution is also repair the opened ledger, and ensure not write the opened ledger by do not modify the topic metadata.

Could you please tell me how you implemented it? Now we check the ledger placement policy when its metadata is closed. Did you change the source code to implement it?

https://github.com/apache/bookkeeper/blob/575a81c79e90e256e919287ef9762bc14d15789a/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/AuditorPlacementPolicyCheckTask.java#L260

horizonzy:
> > Could you please tell me how you implemented it? Now we check the ledger placement policy when its metadata is closed. Did you change the source code to implement it?
> > https://github.com/apache/bookkeeper/blob/575a81c79e90e256e919287ef9762bc14d15789a/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/AuditorPlacementPolicyCheckTask.java#L260
> 
> I change two part of source code, remove "if (!metadata.isClosed())"
> 
> https://github.com/apache/bookkeeper/blob/575a81c79e90e256e919287ef9762bc14d15789a/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/AuditorPlacementPolicyCheckTask.java#L260
> 
> https://github.com/apache/bookkeeper/blob/575a81c79e90e256e919287ef9762bc14d15789a/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/ReplicationWorker.java#L391

[...]

PR #3977, opened 2023-06-05
Do not skip opened ledger in repair not adhering placement ledger

TakaHiR07:
### Motivation

In order to completely fix not adhering placement ledgers, also repair the opened ledger in feature "auto recover support repaired not adhering placement ledger"

### Changes

remove "if (!metadata.isClosed())" in two function. 

It is ok to recover opened ledger because the ledger would become fence when do recovery.

Master Issue: #3971

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/AuditorPlacementPolicyCheckTask.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/AuditorPlacementPolicyCheckTask.java
@@ -257,66 +257,66 @@ public class AuditorPlacementPolicyCheckTask extends AuditorTask {
         LedgerMetadata metadata = metadataVer.getValue();
         int writeQuorumSize = metadata.getWriteQuorumSize();
         int ackQuorumSize = metadata.getAckQuorumSize();
-        if (metadata.isClosed()) {
-            boolean foundSegmentNotAdheringToPlacementPolicy = false;
-            boolean foundSegmentSoftlyAdheringToPlacementPolicy = false;
-            for (Map.Entry<Long, ? extends List<BookieId>> ensemble : metadata
-                    .getAllEnsembles().entrySet()) {
-                long startEntryIdOfSegment = ensemble.getKey();
-                List<BookieId> ensembleOfSegment = ensemble.getValue();
-                EnsemblePlacementPolicy.PlacementPolicyAdherence segmentAdheringToPlacementPolicy = admin
-                        .isEnsembleAdheringToPlacementPolicy(ensembleOfSegment, writeQuorumSize,
-                                ackQuorumSize);
-                if (segmentAdheringToPlacementPolicy == EnsemblePlacementPolicy.PlacementPolicyAdherence.FAIL) {
-                    foundSegmentNotAdheringToPlacementPolicy = true;
-                    LOG.warn(
-                            "For ledger: {}, Segment starting at entry: {}, with ensemble: {} having "
-                                    + "writeQuorumSize: {} and ackQuorumSize: {} is not adhering to "
-                                    + "EnsemblePlacementPolicy",
-                            ledgerId, startEntryIdOfSegment, ensembleOfSegment, writeQuorumSize,
-                            ackQuorumSize);
-                } else if (segmentAdheringToPlacementPolicy
-                        == EnsemblePlacementPolicy.PlacementPolicyAdherence.MEETS_SOFT) {
-                    foundSegmentSoftlyAdheringToPlacementPolicy = true;
-                    if (LOG.isDebugEnabled()) {
-                        LOG.debug(
-                                "For ledger: {}, Segment starting at entry: {}, with ensemble: {}"
-                                        + " having writeQuorumSize: {} and ackQuorumSize: {} is"
-                                        + " softly adhering to EnsemblePlacementPolicy",
-                                ledgerId, startEntryIdOfSegment, ensembleOfSegment, writeQuorumSize,
-                                ackQuorumSize);
-                    }
-                }
-            }
-            if (foundSegmentNotAdheringToPlacementPolicy) {
-                numOfLedgersFoundNotAdheringInPlacementPolicyCheck.incrementAndGet();
-                //If user enable repaired, mark this ledger to under replication manager.
-                if (conf.isRepairedPlacementPolicyNotAdheringBookieEnable()) {
-                    ledgerUnderreplicationManager.markLedgerUnderreplicatedAsync(ledgerId,
-                            Collections.emptyList()).whenComplete((res, e) -> {
-                        if (e != null) {
-                            LOG.error("For ledger: {}, the placement policy not adhering bookie "
-                                            + "storage, mark it to under replication manager failed.",
-                                    ledgerId, e);
-                            return;
-                        }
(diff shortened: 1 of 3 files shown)


TakaHiR07:
@horizonzy can you take a look of this ?

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/AuditorPlacementPolicyCheckTask.java
```diff
@@ -257,66 +257,66 @@ void doPlacementPolicyCheck(Long ledgerId,
...
-                EnsemblePlacementPolicy.PlacementPolicyAdherence segmentAdheringToPlacementPolicy = admin
-                        .isEnsembleAdheringToPlacementPolicy(ensembleOfSegment, writeQuorumSize,
-                                ackQuorumSize);
-                if (segmentAdheringToPlacementPolicy == EnsemblePlacementPolicy.PlacementPolicyAdherence.FAIL) {
-                    foundSegmentNotAdheringToPlacementPolicy = true;
-                    LOG.warn(
-                            "For ledger: {}, Segment starting at entry: {}, with ensemble: {} having "
-                                    + "writeQuorumSize: {} and ackQuorumSize: {} is not adhering to "
-                                    + "EnsemblePlacementPolicy",
+        if (!metadata.isClosed()) {
```
hangc0276, line 260: We'd better skipping check the last segment if the ledger is `OPEN`, otherwise, the replication worker will fence a lot of ledgers can lead to client recreate new ledgers.

hangc0276 (changes requested):

horizonzy (commented):
I checked the code and found that the ReplicaitonWorker didn't fence the ledger. So If we skip the open ledger, this ledger will be written simultaneously by two clients. One is the original client who created the ledger, and the other is a client replicating the data. 

Example:
The E:W:A=3:2:2, and there are 3 bookies [bk0, bk1, bk2]. 
Rack info: bk0 -> rack1, bk1 -> rack1, bk2 -> rack2.
The client creates ledger 1 and writes entries to bk0, bk1. 
At the same time, the ReplicaitonWorker found the ledger 1 ensemble [bk0, bk1] is not adhere to the placement policy.
The ReplicationWorker opens ledger 1 with no recovery and replicates the data to bk2, but the origin client is still to write data to ledger 1. 

The final result is very messy

horizonzy:
> It is ok to recover opened ledger because the ledger would become fence when do recovery.

I didn't notice this logic, can you help describe it?

TakaHiR07:
> I didn't notice this logic, can you help describe it?

When do rereplicate, if foundOpenFragments = true, it goto deferLedgerLockRelease(). The ledger would be fenced. And then wait openLedgerRereplicationGracePeriod(default 30s) to trigger rereplicate again. 
I guess  we are able to recover opened ledger with this logic?

https://github.com/apache/bookkeeper/blob/c8c593ee6d3e1f8970bc90b4b6785ff405c81cc4/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/ReplicationWorker.java#L461-463

horizonzy:
> > I didn't notice this logic, can you help describe it?
> 
> When do rereplicate, if foundOpenFragments = true, it goto deferLedgerLockRelease(). The ledger would be fenced. And then wait openLedgerRereplicationGracePeriod(default 30s) to trigger rereplicate again. I guess we are able to recover opened ledger with this logic?
> 
> https://github.com/apache/bookkeeper/blob/c8c593ee6d3e1f8970bc90b4b6785ff405c81cc4/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/ReplicationWorker.java#L461-L463

We'd better skip the last ensemble of the current ledger, as it may be writing data. Let's only copy the ensembles that have already been written before it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/AuditorPlacementPolicyCheckTask.java
wenbingshen: +1

If there is a rack failure and enforceMinNumRacksPerWriteQuorum=false, the open segment will never satisfy the rack distribution, and the auditor will cause all open ledger fences in the cluster. We can try to repair the closed ledger or segment, but for the open segment, the frequent fence ledgers can be a bit bad when rack failures.

We should be consistent with the behavior of the bookie client write side. Especially for open segments.

wenbingshen:
> > > I didn't notice this logic, can you help describe it?
> > 
> > 
> > When do rereplicate, if foundOpenFragments = true, it goto deferLedgerLockRelease(). The ledger would be fenced. And then wait openLedgerRereplicationGracePeriod(default 30s) to trigger rereplicate again. I guess we are able to recover opened ledger with this logic?
> > https://github.com/apache/bookkeeper/blob/c8c593ee6d3e1f8970bc90b4b6785ff405c81cc4/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/ReplicationWorker.java#L461-L463
> 
> We'd better skip the last ensemble of the current ledger, as it may be writing data. Let's only copy the ensembles that have already been written before it.

+1

TakaHiR07:
@horizonzy @hangc0276 @wenbingshen The code has been modified to skip the last ensemble of the current ledger. Can you help review again?
To avoid fence a lot of ledgers, actually some opened ledgers still can not be recover. Maybe we can add a config to enableForceRecoverOpenLedger,  I think in some scene we just fence a little of ledgers, such as only recover "__change_event" topic.  what do you think?

wenbingshen:
> @horizonzy @hangc0276 @wenbingshen The code has been modified to skip the last ensemble of the current ledger. Can you help review again? To avoid fence a lot of ledgers, actually some opened ledgers still can not be recover. Maybe we can add a config to enableForceRecoverOpenLedger, I think in some scene we just fence a little of ledgers, such as only recover "__change_event" topic. what do you think?

All ledgers are equal in the bookie. For the `__change_event` topic, the rack allocation cannot be satisfied. I think we need to do some investigations in the broker.

horizonzy:
> > @horizonzy @hangc0276 @wenbingshen The code has been modified to skip the last ensemble of the current ledger. Can you help review again? To avoid fence a lot of ledgers, actually some opened ledgers still can not be recover. Maybe we can add a config to enableForceRecoverOpenLedger, I think in some scene we just fence a little of ledgers, such as only recover "__change_event" topic. what do you think?
> 
> All ledgers are equal in the bookie. For the `__change_event` topic, the rack allocation cannot be satisfied. 

+1.

StevenLuMT:
I feel that this change is quite complicated. Can I elaborate on it, do I need to write a BP?

TakaHiR07:
> I feel that this change is quite complicated. Can I elaborate on it, do I need to write a BP?

you can.

dlg99:
Fencing is not exactly a cheap operation for application using BK.
E.g. in case of BK as a WAL it may cause app crash/restart/recovery from WAL etc, in other cases rolling restart/upgrade of bookies may result in a lot of smaller ledgers than anticipated (growth of metadata size).

There is a config parameter https://github.com/apache/bookkeeper/blob/b8cc1fb10a45073ad38ca26b566f01afb3a31f99/conf/bk_server.conf#L1063-L1065 that you can tune down for your usecase. I think it should work there though I don't remember all intricacies  of replication worker.

PR #3980, opened 2023-06-06
Add #3979 unit test

gaozhangmin:
Add Unit test for #3979

### DIFF
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieJournalForceTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieJournalForceTest.java
@@ -21,6 +21,7 @@
 package org.apache.bookkeeper.bookie;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.atLeast;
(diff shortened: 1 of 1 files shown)


gaozhangmin:
rerun failure checks

gaozhangmin:
rerun failure checks

eolivelli (changes requested):
Thank for working on this.

I think that the new test seems too much an "artifact":
- it is full of reflective call
- full of mocks

I wonder if the test would be useful in the future.

Here we are testing some values in the internal representation, if in the future we would change that data structure this test may become useless or misleading.

It would be better to have a test that reproduces a problem and that fails without the fix.

which is the original problem that you was ? that the Bookie didn't call "force" ? or "flush" ?
We can intercept those calls and verify that they are called the expected number of times

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieJournalForceTest.java
```diff
@@ -402,4 +406,70 @@ public void testFileChannelProvider() throws Exception {
...
+
+        field = forceWriteRequestClass.getDeclaredField("forceWriteWaiters");
+        field.setAccessible(true);
+        assertEquals(forceWriteWaiters, field.get(request));
+
+        field = forceWriteRequestClass.getDeclaredField("shouldClose");
+        field.setAccessible(true);
+        assertTrue((boolean) field.get(request));
+
+        field = forceWriteRequestClass.getDeclaredField("flushed");
```
horizonzy, line 470: We shouldn't use particular filed by name. If we add a new field in the future, it can't cover it.

gaozhangmin:
> Thank for working on this.
> 
> I think that the new test seems too much an "artifact":
> 
> * it is full of reflective call
> * full of mocks
> 
> I wonder if the test would be useful in the future.
> 
> Here we are testing some values in the internal representation, if in the future we would change that data structure this test may become useless or misleading.
> 
> It would be better to have a test that reproduces a problem and that fails without the fix.
> 
> which is the original problem that you was ? that the Bookie didn't call "force" ? or "flush" ? We can intercept those calls and verify that they are called the expected number of times

Sure, I will add another test

StevenLuMT (changes requested):
I also feel that this testcase is very strange, can you add the motivation of the testcase?

PR #3985, opened 2023-06-13
Unify ByteBufAllocator for the DirectIO component

hangc0276:
### Motivation
Some classes in the DirectIO component use `PooledByteBufAllocator` to create ByteBuf instead of using the BookKeeper-initiated allocator. When we configured the BookKeeper allocator OOM policy to shut down, the OOM policy can't apply to those classes which use `PooledByteBufAllocator` to create ByteBuf. 

### Modifications
- Unify ByteBufAllocator for the DirectIO component.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/directentrylogger/Buffer.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/directentrylogger/Buffer.java
@@ -67,10 +67,10 @@ class Buffer {
         checkArgument(isAligned(bufferSize),
                       "Buffer size not aligned %d", bufferSize);
 
+        this.allocator = allocator;
         this.buffer = allocateAligned(ALIGNMENT, bufferSize);
         this.nativeIO = nativeIO;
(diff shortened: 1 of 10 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/directentrylogger/Buffer.java
```diff
@@ -60,21 +60,23 @@ class Buffer {
...
 
-    Buffer(NativeIO nativeIO, int bufferSize) throws IOException {
+    Buffer(NativeIO nativeIO, ByteBufAllocator allocator, int bufferSize) throws IOException {
         checkArgument(isAligned(bufferSize),
                       "Buffer size not aligned %d", bufferSize);
 
         this.buffer = allocateAligned(ALIGNMENT, bufferSize);
         this.nativeIO = nativeIO;
         this.bufferSize = bufferSize;
+        this.allocator = allocator;
```
zymap, line 73: Need to declare at line 69. The `allocateAligned` uses the allocator.

zymap (approved):

PR #3988, opened 2023-06-15
Fix Journal Sync wrong stats

wenbingshen:
### Motivation

In Journal, we process `ForceWriteRequest` with batch as the smallest unit. When we rotate a Journal File, in the forceWriteRequests queue, we may contain a ForceWriteRequest of an old file, and shouldClose is true, and the last ForceWriteRequest in the queue is created by the new Jouranl file. When we batch process fwr, we only sync the last ForceWriteRequest to the disk, so that we can flash the previous fwr data of the file to the disk together. But for old files, when we process the fwr response, we will still process `closeFileIfNecessary`. Here we also sync the Journal files that are waiting for close, but here, we do not update `JournalSyncStats`.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
@@ -372,8 +372,8 @@ public class Journal extends BookieCriticalThread implements CheckpointSource {
         private long logId;
         private boolean flushed;
 
-        public int process(ObjectHashSet<BookieRequestHandler> writeHandlers) {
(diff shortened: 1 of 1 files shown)


wenbingshen:
rerun failure checks

wenbingshen:
rerun failure checks

wenbingshen:
rerun failure checks

gaozhangmin (approved):

StevenLuMT (changes requested):
Two questions:
1. what is the impact of the old logic?
2. can you add a testcase to verify that the new logic to check expectations?

PR #3992, opened 2023-06-19
Upgrade grpc and protobuf to address CVE-2023-32732

lhotari:
### Motivation

OWASP dependency check fails because of CVE-2023-32732 in grpc.

### Changes

* Upgrade grpc to 1.56.0
* Upgrade protobuf to 3.22.3 to match the version used in grpc 1.56.0
* Upgrade other grpc/protobuf related libs

PR #3997, opened 2023-06-20
Fix issue with binary compatibility with older grpc versions at runtime in the client

lhotari:
### Motivation

- grpc version was upgraded to 1.56.0 in #3992
- that breaks binary compatibility for DnsNameResolverProvider class
  - see https://github.com/grpc/grpc-java/commit/fcb5c54e4b82d354f42ced0121928fabce9ef53f#diff-b04e884de51ed12ff79482f600a2d4ec18e405ee189a4952ae35f4d2742b7160L50

### Changes

- make the field type NameResolverProvider instead of DnsNameResolverProvider
  - this prevents possible NoSuchMethodError errors about DnsNameResolverProvider.newNameResolver method

### DIFF
--- a/stream/common/src/main/java/org/apache/bookkeeper/common/resolver/ServiceNameResolverProvider.java
+++ b/stream/common/src/main/java/org/apache/bookkeeper/common/resolver/ServiceNameResolverProvider.java
@@ -40,7 +40,7 @@ import org.apache.bookkeeper.common.util.SharedResourceManager.Resource;
 @Slf4j
 public final class ServiceNameResolverProvider extends NameResolverFactoryProvider {
 
-    private final DnsNameResolverProvider dnsProvider;
+    private final NameResolverProvider dnsProvider;
     private final Resource<ExecutorService> executorResource;
 
     public ServiceNameResolverProvider() {


zymap (approved):

hangc0276 (approved):

wenbingshen (approved):

lhotari:
This PR will also be useful to cherry-pick on maintenance branches to support grpc-java >= 1.54.0

lhotari:
In Pulsar we need Bookkeeper 4.16.2 release with this fix until we can make the [switch to grpc-java 1.56 on Pulsar side](https://github.com/apache/pulsar/pull/20602). @zymap @hangc0276 @eolivelli @nicoloboschi @dlg99  Any timeline for 4.16.2 release?

zymap:
@lhotari  I am preparing the release, and I think we can release it in the next week. But I met [some failures in Pulsar's CI ](https://github.com/zymap/pulsar/pull/46)and need to take a look before sending the vote. It looks like something is wrong with the bookie that causes the pulsar's tests to fail.

zymap:
It seems I met the same issue mentioned here https://github.com/protocolbuffers/protobuf/issues/11986

Pulsar errors:

```
Caused by: java.lang.IllegalAccessError: class org.apache.bookkeeper.proto.DataFormats$BookieServiceInfoFormat$Endpoint tried to access method 'com.google.protobuf.LazyStringArrayList com.google.protobuf.LazyStringArrayList.emptyList()' (org.apache.bookkeeper.proto.DataFormats$BookieServiceInfoFormat$Endpoint and com.google.protobuf.LazyStringArrayList are in unnamed module of loader 'app')
```

lhotari:
> It seems I met the same issue mentioned here [protocolbuffers/protobuf#11986](https://github.com/protocolbuffers/protobuf/issues/11986)
> 
> Pulsar errors:
> 
> Caused by: java.lang.IllegalAccessError: class org.apache.bookkeeper.proto.DataFormats$BookieServiceInfoFormat$Endpoint tried to access method 'com.google.protobuf.LazyStringArrayList com.google.protobuf.LazyStringArrayList.emptyList()' (org.apache.bookkeeper.proto.DataFormats$BookieServiceInfoFormat$Endpoint and com.google.protobuf.LazyStringArrayList are in unnamed module of loader 'app')

@zymap I wonder what the resolution is? Perhaps there's an option to downgrade protobuf to 3.21.x ?

lhotari:
@zymap I made an attempt in #4000 to fix the issue.

zymap:
> > It seems I met the same issue mentioned here [protocolbuffers/protobuf#11986](https://github.com/protocolbuffers/protobuf/issues/11986)
> > Pulsar errors:
> > Caused by: java.lang.IllegalAccessError: class org.apache.bookkeeper.proto.DataFormats$BookieServiceInfoFormat$Endpoint tried to access method 'com.google.protobuf.LazyStringArrayList com.google.protobuf.LazyStringArrayList.emptyList()' (org.apache.bookkeeper.proto.DataFormats$BookieServiceInfoFormat$Endpoint and com.google.protobuf.LazyStringArrayList are in unnamed module of loader 'app')
> 
> @zymap I wonder what the resolution is? Perhaps there's an option to downgrade protobuf to 3.21.x ?

@lhotari I found the reason for this error. It is caused by the protobuf-java version. In the generated code, it uses `emptyList()`
```
auth_ =
            com.google.protobuf.LazyStringArrayList.emptyList();
        extensions_ =
            com.google.protobuf.LazyStringArrayList.emptyList();
```

In the pulsar side, we are using 3.19.6, which the `emptyList()` method is package private https://github.com/protocolbuffers/protobuf/blob/5cba162a5d93f8df786d828621019e03e50edb4f/java/core/src/main/java/com/google/protobuf/LazyStringArrayList.java#L70

They have exposed the method to the public since 3.22.0 https://github.com/protocolbuffers/protobuf/commit/c658e27529ccf4a000724ab3622f1b807c85449b

This means if we want to change upgrade protobuf to 3.22.0+, we must require all the sub-project, which depends on the bookkeeper to upgrade protobuf to 3.22.0+. It should not be acceptable in a minor release. 
So I will exclude this PR from the 4.16.2 release and include it in the next major release.

PR #3998, opened 2023-06-22
Fix trigger GC not work

hangc0276:
### Motivation
When I use `curl -XPUT http://<ip>:<port>/api/v1/bookie/gc` command to trigger garbage compaction, the bookie server returns `Internal Server Error`. Then I checked the bookie server's log, and nothing was found.

After a deep debugging of the server, I found the following exception.
```
    org.apache.bookkeeper.common.util.JsonUtil$ParseJsonException: Failed to deserialize Object from Json string
	at org.apache.bookkeeper.common.util.JsonUtil.fromJson(JsonUtil.java:41)
	at org.apache.bookkeeper.server.http.service.TriggerGCService.handle(TriggerGCService.java:71)
	at org.apache.bookkeeper.http.vertx.VertxAbstractHandler.processRequest(VertxAbstractHandler.java:55)
	at org.apache.bookkeeper.http.vertx.VertxHttpHandlerFactory$1.handle(VertxHttpHandlerFactory.java:47)
	at org.apache.bookkeeper.http.vertx.VertxHttpHandlerFactory$1.handle(VertxHttpHandlerFactory.java:43)
	at io.vertx.ext.web.impl.RouteState.handleContext(RouteState.java:1038)
	at io.vertx.ext.web.impl.RoutingContextImplBase.iterateNext(RoutingContextImplBase.java:137)
	at io.vertx.ext.web.impl.RoutingContextImpl.next(RoutingContextImpl.java:132)
	at io.vertx.ext.web.handler.impl.BodyHandlerImpl$BHandler.doEnd(BodyHandlerImpl.java:296)
	at io.vertx.ext.web.handler.impl.BodyHandlerImpl$BHandler.end(BodyHandlerImpl.java:276)
	at io.vertx.ext.web.handler.impl.BodyHandlerImpl.lambda$handle$0(BodyHandlerImpl.java:87)
	at io.vertx.core.http.impl.HttpServerRequestImpl.onEnd(HttpServerRequestImpl.java:525)
	at io.vertx.core.http.impl.HttpServerRequestImpl.handleEnd(HttpServerRequestImpl.java:511)
	at io.vertx.core.http.impl.Http1xServerConnection.handleEnd(Http1xServerConnection.java:176)
	at io.vertx.core.http.impl.Http1xServerConnection.handleMessage(Http1xServerConnection.java:138)
	at io.vertx.core.impl.ContextImpl.executeTask(ContextImpl.java:366)
	at io.vertx.core.impl.EventLoopContext.execute(EventLoopContext.java:43)
```

This bug is introduced by https://github.com/apache/bookkeeper/pull/3205. When we use `curl -XPUT http://<ip>:<port>/api/v1/bookie/gc` command to trigger gc, the request body is empty string, but not `null`. 
https://github.com/apache/bookkeeper/blob/ffc8e8bec19e54dc76710c3de133401248f243d6/bookkeeper-server/src/main/java/org/apache/bookkeeper/server/http/service/TriggerGCService.java#L67-L75

It will enter Line#68 but goes into Line#71. It uses `JsonUtil.fromJson` to parse an empty string will throw the above exception and will return `500` to the client but no logs in the server side.

### Changes
- Fix the request body check bug
- Add a unit test to cover this logic

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/server/http/service/TriggerGCService.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/server/http/service/TriggerGCService.java
@@ -22,6 +22,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
 
 import java.util.HashMap;
 import java.util.Map;
+import org.apache.bookkeeper.bookie.LedgerStorage;
 import org.apache.bookkeeper.common.util.JsonUtil;
 import org.apache.bookkeeper.conf.ServerConfiguration;
 import org.apache.bookkeeper.http.HttpServer;
@@ -70,9 +71,13 @@ public class TriggerGCService implements HttpEndpointService {
                 } else {
                     @SuppressWarnings("unchecked")
                     Map<String, Object> configMap = JsonUtil.fromJson(requestBody, HashMap.class);
-                    Boolean forceMajor = (Boolean) configMap.getOrDefault("forceMajor", null);
-                    Boolean forceMinor = (Boolean) configMap.getOrDefault("forceMinor", null);
-                    bookieServer.getBookie().getLedgerStorage().forceGC(forceMajor, forceMinor);
+                    LedgerStorage ledgerStorage = bookieServer.getBookie().getLedgerStorage();
+                    boolean forceMajor = !ledgerStorage.isMajorGcSuspended();
+                    boolean forceMinor = !ledgerStorage.isMinorGcSuspended();
+
+                    forceMajor = Boolean.parseBoolean(configMap.getOrDefault("forceMajor", forceMajor).toString());
+                    forceMinor = Boolean.parseBoolean(configMap.getOrDefault("forceMinor", forceMinor).toString());
+                    ledgerStorage.forceGC(forceMajor, forceMinor);
                 }
 
                 String output = "Triggered GC on BookieServer: " + bookieServer.getBookieId();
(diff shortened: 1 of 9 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/server/http/service/TriggerGCService.java
```diff
@@ -61,40 +62,46 @@ public TriggerGCService(ServerConfiguration conf, BookieServer bookieServer) {
...
         HttpServiceResponse response = new HttpServiceResponse();
+        try {
+            if (HttpServer.Method.PUT == request.getMethod()) {
+                String requestBody = request.getBody();
+                if (StringUtils.isBlank(requestBody)) {
+                    bookieServer.getBookie().getLedgerStorage().forceGC();
+                } else {
+                    @SuppressWarnings("unchecked")
+                    Map<String, Object> configMap = JsonUtil.fromJson(requestBody, HashMap.class);
+                    Boolean forceMajor = (Boolean) configMap.getOrDefault("forceMajor", null);
```
zymap, line 73: Shouldn't the default value be a boolean? If it is null, it would be easy to cause NPE in the inner code. We can use Boolean.parseBoolean() to make sure it is a boolean and avoid a cast error happening.

bookkeeper-server/src/main/java/org/apache/bookkeeper/server/http/service/TriggerGCService.java
```diff
@@ -61,40 +62,46 @@ public TriggerGCService(ServerConfiguration conf, BookieServer bookieServer) {
...
+                response.setBody(jsonResponse);
+                response.setCode(HttpServer.StatusCode.OK);
+                return response;
             } else {
-                @SuppressWarnings("unchecked")
-                Map<String, Object> configMap = JsonUtil.fromJson(requestBody, HashMap.class);
-                Boolean forceMajor = (Boolean) configMap.getOrDefault("forceMajor", null);
-                Boolean forceMinor = (Boolean) configMap.getOrDefault("forceMinor", null);
-                bookieServer.getBookie().getLedgerStorage().forceGC(forceMajor, forceMinor);
+                response.setCode(HttpServer.StatusCode.NOT_FOUND);
```
zymap, line 97: The status should be NOT_ALLOWED, see here https://httpwg.org/specs/rfc9110.html#status.405

eolivelli: +1

eolivelli (approved):
+1

bookkeeper-server/src/main/java/org/apache/bookkeeper/server/http/service/TriggerGCService.java
hangc0276: Done.

bookkeeper-server/src/main/java/org/apache/bookkeeper/server/http/service/TriggerGCService.java
hangc0276: Updated. It will change the current status code.

zymap (approved):

PR #4001, opened 2023-06-25
[Branch-4.16] Downgrade grpc and protobuf to avoid introducing breaking change

zymap:
---

### Motivation

We upgrade grpc and protobuf to address CVE-2023-32732. But it requires the protobuf 3.22+. In protobuf 3.22.0, it introduces a breaking change. It requires all the sub-project, which depend on the bookkeeper to upgrade protobuf to 3.22.0+. It should not be acceptable in a minor release.

So we use a lower version of grpc and protobuf to address the CVE issue.

See more context: https://github.com/apache/bookkeeper/pull/3997


zymap:
@lhotari PTAL, thanks

lhotari (approved):

hangc0276 (approved):
Good job!

zymap:
https://github.com/protocolbuffers/protobuf/issues/11393 Another compatible issue with jdk8, I need to downgrade the protobuf.
They have fixed but not released https://github.com/protocolbuffers/protobuf/pull/12036

hangc0276:
LGTM

PR #4004, opened 2023-06-28
[improve] Add ClusterInfo shell Command

wenbingshen:
### Motivation
PR #3710 introduced a future for get ClusterInfo about bookies cluster.
This is convenient for us to obtain the basic information of the cluster through the rest api, which is very useful.

When we disable the http port, it will be very troublesome to obtain these cluster information through the command tool, because most of the command tools print the output through the log4j info log, and it will be very troublesome for us to aggregate these results. So I hope to introduce a shell command for this feature.


wenbingshen:
@dlg99 @eolivelli @hangc0276 PTAL. Thanks!

dlg99 (approved):

dlg99:
Great work!
There is a checkstyle violation that needs a fix; the rest lgtm.

hangc0276 (approved):

wenbingshen:
rerun failure checks

PR #4005, opened 2023-06-29
Optimize getActiveLedgersInRange

AnonHxy:

### Motivation

Optimize `Iterable<Long>  getActiveLedgersInRange(0, Long.MAX_VALUE)` by  avoid applying the filter predicate when traverse the `Iterable<Long>`.

There are many places call `getActiveLedgersInRange(0, Long.MAX_VALUE)`  to get the `Iterable` of all the `ledgerId`,  for example:
1. https://github.com/apache/bookkeeper/blob/147b5bf846189584d6a9b4e20ac9a73e6e617255/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/ScanAndCompareGarbageCollector.java#L124-L125
2. https://github.com/apache/bookkeeper/blob/147b5bf846189584d6a9b4e20ac9a73e6e617255/bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/commands/bookie/ConvertToDBStorageCommand.java#L88
3. https://github.com/apache/bookkeeper/blob/147b5bf846189584d6a9b4e20ac9a73e6e617255/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/LocationsIndexRebuildOp.java#L160

We could return the keys of `ledgers` directly to avoid applying the filter predicate when traverse the `Iterable<Long>`

### Changes

Return the keys of `ledgers` directly if `firstLedgerId <=0 && lastLedgerId == Long.MAX_VALUE`

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/LedgerMetadataIndex.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/LedgerMetadataIndex.java
@@ -160,6 +160,9 @@ public class LedgerMetadataIndex implements Closeable {
 
     public Iterable<Long> getActiveLedgersInRange(final long firstLedgerId, final long lastLedgerId)
             throws IOException {
+        if (firstLedgerId <=0 && lastLedgerId == Long.MAX_VALUE) {
+            return ledgers.keys();
+        }
         return Iterables.filter(ledgers.keys(), new Predicate<Long>() {
             @Override
             public boolean apply(Long ledgerId) {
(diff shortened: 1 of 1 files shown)


StevenLuMT (approved):
good catch,LGTM

horizonzy (commented):
We would better add a new method `getAllActiveLedgers`.

StevenLuMT:
> We would better add a new method `getAllActiveLedgers`.

Whether to change the semantics of the original method, if it is changed, it is recommended to initiate a BP to add

hangc0276 (approved):

hangc0276:
> We would better add a new method `getAllActiveLedgers`.

@horizonzy This change is low risk, and I think it's ok to be put in the original method.

horizonzy:
That's fine.&nbsp;

------------------ Original ------------------
From: Hang Chen ***@***.***&gt;
Date: Mon,Jul 3,2023 3:57 PM
To: apache/bookkeeper ***@***.***&gt;
Cc: Yan Zhao ***@***.***&gt;, Mention ***@***.***&gt;
Subject: Re: [apache/bookkeeper] Optimize getActiveLedgersInRange (PR #4005)





  
We would better add a new method getAllActiveLedgers.
  
@horizonzy This change is low risk, and I think it's ok to be put in the original method.

horizonzy (approved):
LGTM

PR #4009, opened 2023-07-01
Fix replace result adhere result wrong problem.

horizonzy:

When replacing a bookie, if there are two or more bookies that need to be replaced, our logic is to replace them one by one. However, after the replacement, we do not use the new bookies to replace the old ones in the ensemble. Therefore, the return result of isEnsembleAdheringToPlacementPolicy may be incorrect.

The code:
```
        for (Integer bookieIndex : bookieIndexesToRereplicate) {
            BookieId oldBookie = newEnsemble.get(bookieIndex);
            EnsemblePlacementPolicy.PlacementResult<BookieId> replaceBookieResponse =
                    bkc.getPlacementPolicy().replaceBookie(
                            lh.getLedgerMetadata().getEnsembleSize(),
                            lh.getLedgerMetadata().getWriteQuorumSize(),
                            lh.getLedgerMetadata().getAckQuorumSize(),
                            lh.getLedgerMetadata().getCustomMetadata(),
                            ensemble,
                            oldBookie,
                            bookiesToExclude);
            BookieId newBookie = replaceBookieResponse.getResult();
                LOG.debug(
                        "replaceBookie for bookie: {} in ensemble: {} "
                                + "is not adhering to placement policy and chose {}",
                        oldBookie, newEnsemble, newBookie);
            }
            targetBookieAddresses.put(bookieIndex, newBookie);
            bookiesToExclude.add(newBookie);
            newEnsemble.set(bookieIndex, newBookie);
        }
```

The 5th param `ensemble` in EnsemblePlacementPolicy#replaceBookie should be updated after replace.

Example:
The ensemble [0,1,2], we want to replace [1,2].

Step1. replace 1 by 3, after replace, the 5th param ensemble should be [0,3,2], instead still [0,1,2].

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperAdmin.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperAdmin.java
@@ -1084,17 +1084,17 @@ public class BookKeeperAdmin implements AutoCloseable {
             BookieId bookie = ensemble.get(bookieIndex);
             bookiesToExclude.add(bookie);
         }
-
+        List<BookieId> newEnsemble = new ArrayList<>(ensemble) ;
         // allocate bookies
         for (Integer bookieIndex : bookieIndexesToRereplicate) {
-            BookieId oldBookie = ensemble.get(bookieIndex);
+            BookieId oldBookie = newEnsemble.get(bookieIndex);
             EnsemblePlacementPolicy.PlacementResult<BookieId> replaceBookieResponse =
                     bkc.getPlacementPolicy().replaceBookie(
                             lh.getLedgerMetadata().getEnsembleSize(),
                             lh.getLedgerMetadata().getWriteQuorumSize(),
                             lh.getLedgerMetadata().getAckQuorumSize(),
                             lh.getLedgerMetadata().getCustomMetadata(),
-                            ensemble,
+                            newEnsemble,
                             oldBookie,
                             bookiesToExclude);
             BookieId newBookie = replaceBookieResponse.getResult();
@@ -1103,10 +1103,11 @@ public class BookKeeperAdmin implements AutoCloseable {
                 LOG.debug(
                         "replaceBookie for bookie: {} in ensemble: {} "
                                 + "is not adhering to placement policy and chose {}",
-                        oldBookie, ensemble, newBookie);
+                        oldBookie, newEnsemble, newBookie);
             }
             targetBookieAddresses.put(bookieIndex, newBookie);
             bookiesToExclude.add(newBookie);
+            newEnsemble.set(bookieIndex, newBookie);
         }
 
         return targetBookieAddresses;


zymap (commented):
Better to have a test. So it will easier to understand the situation.

horizonzy:
> Better to have a test. So it will easier to understand the situation.

Make sense.

StevenLuMT (approved):
Good catch

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/TestRackawareEnsemblePlacementPolicy.java
```diff
@@ -636,6 +638,75 @@ public void testIsEnsembleAdheringToPlacementPolicy() throws Exception {
...
+
+        for (Integer bookieIndex : bookieIndexesToRereplicate) {
+            BookieId bookie = ensemble.get(bookieIndex);
+            bookiesToExclude.add(bookie);
+        }
+
+        List<BookieId> newEnsemble = new ArrayList<>(ensemble);
+
+        int i = 0;
+        for (Integer bookieIndex : bookieIndexesToRereplicate) {
```
hangc0276, line 691: Can you check the logic by calling getReplacementBookiesByIndexes directly or indirectly? By copy this changed logic to the test can't protect the original code logic.

horizonzy: For getReplacementBookiesByIndexes, this change doesn't influence the result. https://github.com/apache/bookkeeper/blob/52e780f326c6cb5f303b51aa6ab5363c8e94b074/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperAdmin.java#L1088-L1110

Line_1109 already add the new bookie into `bookiesToExclude`.

It just didn't update the new bookie to the ensemble. It's different to test for `getReplacementBookiesByIndexes `, so I add a test to mock the internal logic in the `getReplacementBookiesByIndexes`.

horizonzy:
rerun failure checks

horizonzy:
rerun failure checks

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperAdmin.java
```diff
@@ -1084,17 +1084,17 @@ private Map<Integer, BookieId> getReplacementBookiesByIndexes(
             BookieId bookie = ensemble.get(bookieIndex);
             bookiesToExclude.add(bookie);
         }
-
+        List<BookieId> newEnsemble = new ArrayList<>(ensemble);
```
zymap, line 1087: If I understand correctly, this isn't a bug, right? I mean, this hasn't any impact on the existing logic and result, right?

horizonzy: Yes, it just makes the log incorrect.

zymap: Got it

PR #4010, opened 2023-07-01
Add read failed log for ledger checker.

horizonzy:
The ledger checker will try to read entries from the bookie. If the read failed, mark the ledger as under-replicated. 

For users, they didn't know the reason why the ledger mark as under-replicated. Is it because the entry is missing, the ledger does not exist, or because the bookie is under heavy load and the read request times out?

So here we add debug log for ledger checker. If read failed, log the error code.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerChecker.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerChecker.java
@@ -87,6 +87,9 @@ public class LedgerChecker {
                     cb.operationComplete(rc, fragment);
                 }
             } else if (!completed.getAndSet(true)) {
+                if (LOG.isDebugEnabled()) {
+                    LOG.debug("Read {}:{} from {} failed, the error code: {}", ledgerId, entryId, ctx, rc);
+                }
                 cb.operationComplete(rc, fragment);
             }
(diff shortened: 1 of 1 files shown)


horizonzy:
rerun failure checks

horizonzy:
rerun failure checks

horizonzy:
rerun failure checks

zymap (approved):

hangc0276 (approved):

wenbingshen (approved):

StevenLuMT (changes requested):
What is the reason for modifying the fifth parameter of bookieClient.readEntry? 
I couldn't see it's motivation in this PR

horizonzy:
> What is the reason for modifying the fifth parameter of bookieClient.readEntry? I couldn't see it's motivation in this PR

In the callback, we need to print the bookie. So we need the bookie as the ctx.

wenbingshen:
> > What is the reason for modifying the fifth parameter of bookieClient.readEntry? I couldn't see it's motivation in this PR
> 
> In the callback, we need to print the bookie. So we need the bookie as the ctx.

@StevenLuMT Can you help take a look at this comment? Thanks.

StevenLuMT (approved):
LGTM

PR #4015, opened 2023-07-06
Fix getLedgerDirsListener

gaozhangmin:
### Motivation
The ledgerDirsListener in SingleDirectoryDbLedgerStorage is currently triggered by a ledger directory that does not belong to this storage.

### Changes

To address this issue, we can introduce a comparison step to verify the ownership of the ledger directory before triggering the listeners. This way, the ledgerDirsListener will only be triggered for the directories that are associated with the storage.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
@@ -144,8 +144,8 @@ public class SingleDirectoryDbLedgerStorage implements CompactableLedgerStorage 
 
     private final Counter flushExecutorTime;
     private final boolean singleLedgerDirs;
-
-    private final String ledgerBaseDir;
+    private final LedgerDirsManager ledgerDirsManager;
+    private final LedgerDirsManager indexDirsManager;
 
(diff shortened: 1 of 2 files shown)


gaozhangmin:
rerun failure checks

gaozhangmin:
rerun failure checks

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
```diff
@@ -1147,21 +1151,28 @@ private LedgerDirsManager.LedgerDirsListener getLedgerDirsListener() {
 
             @Override
             public void diskAlmostFull(File disk) {
-                if (gcThread.isForceGCAllowWhenNoSpace()) {
-                    gcThread.enableForceGC();
-                } else {
-                    gcThread.suspendMajorGC();
+                if (ledgerDirsManager.getAllLedgerDirs().contains(disk)
+                        || indexDirsManager.getAllLedgerDirs().contains(disk)) {
```
hangc0276, line 1155: We'd better not take indexDirs into consideration. If the index dir shares the same disk with the ledger dir, the LedgerDirsManager can cover the event. If the index dir uses different disks with ledger dir, it doesn't make sense to suspend ledger disk garbage collection.

hangc0276 (changes requested):
Would you please add a test to cover the case you described?

gaozhangmin:
@hangc0276  PTAL

PR #4018, opened 2023-07-06
Add diskUnderWarnThreshold check event to prevent gcThread suspended forever

gaozhangmin:
### Motivation
This  code snippet  implements disk check event trigger if disk usage is above warnThreshold, either marjor is suspend or forcegc is enabled. 

There exists a case that if usgae descends to warn threshold, the gc thread is suspend forever.  until it goes to full and becomes writable again.
```
@Override
                if (gcThread.isForceGCAllowWhenNoSpace) {
                    gcThread.enableForceGC();
                } else {
                    gcThread.suspendMajorGC();
                }
            }
```

### Changes

Add diskUnderWarnThreshold check event to prevent gcThread  supended forever

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDirsManager.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDirsManager.java
@@ -47,6 +47,7 @@ public class LedgerDirsManager {
     private static final Logger LOG = LoggerFactory.getLogger(LedgerDirsManager.class);
 
     private volatile List<File> filledDirs;
+    private volatile List<File> warnDirs;
     private final List<File> ledgerDirectories;
     private volatile List<File> writableLedgerDirectories;
     private final List<LedgerDirsListener> listeners;
@@ -234,6 +235,13 @@ public class LedgerDirsManager {
     }
 
     /**
+     * @return warn threshold ledger dirs.
+     */
+    public List<File> getWarnLedgerDirs() {
+        return warnDirs;
+    }
+
+    /**
      * Get dirs, which are full more than threshold.
      */
     public boolean isDirFull(File dir) {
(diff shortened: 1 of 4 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDirsManager.java
```diff
@@ -233,6 +235,13 @@ public List<File> getFullFilledLedgerDirs() {
         return filledDirs;
     }
 
+    /**
+     * @return warn threshold ledger dirs.
+     */
+    public List<File> getWarnLedgerDirs() {
```
hangc0276, line 241: What's the difference with `diskWritable` of the newly introduced interface?

hangc0276 (changes requested):

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDirsManager.java
gaozhangmin: `warnLedgerDirs` is only used for trigger under warn threshold event. has no effects on writing and reading.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDirsManager.java
```diff
@@ -290,6 +299,39 @@ public void addToWritableDirs(File dir, boolean underWarnThreshold) {
...
     }
 
+    /**
+     * Add the dir to warn dirs list.
+     */
+    @VisibleForTesting
+    public void addToWarnDirs(File dir) {
+        if (!warnDirs.contains(dir)) {
+            LOG.warn(dir + " usages is above warn threshold. Adding it to warn dirs list");
+            List<File> updatedFilledDirs = new ArrayList<File>(warnDirs);
```
hangc0276, line 309: Why not use `warnDirs.add(dir)` directly? Due to the warnDirs is a list, does it will happen the same dir being add to the list twice?

hangc0276 (commented):
Would you please add a test to protect the case you described?

gaozhangmin:
> Would you please add a test to protect the case you described?

@gaozhangmin @hangc0276  done, PTAL again

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDirsManager.java
gaozhangmin: Yes, it would be addToWarnDirs every time its usage exceeds the warn threshold

gaozhangmin:
rerun failure checks

zhannao456:
rerun failure checks

gaozhangmin:
rerun failure checks

gaozhangmin:
rerun failure checks

PR #4020, opened 2023-07-08
PerChannelBookieClient Enhancement: Self-Recovery Support in Corrupte…

shoothzj:
### Motivation
In the context of popular protocols like LengthBasedFrame, which typically determine total message length based on the first 0-4 bytes (for example, if the initial bytes are 00000014, the message length is 1*16+4 = 20), message tampering can cause serious issues. Specifically, any modification to the higher order bytes can lead to miscalculation of message length and leave the communication end waiting indefinitely without any ability to self-recover.

In summary, in scenarios with message tampering, communication timeout can lead to an unrecoverable link.
I think we can rebuild the client when timeout.

### Changes
rebuild the **PerChannelBookieClient** when timeout.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
@@ -1043,6 +1043,9 @@ public class PerChannelBookieClient extends ChannelInboundHandlerAdapter {
         if (timedOutOperations > 0) {
             LOG.info("Timed-out {} operations to channel {} for {}",
                      timedOutOperations, channel, bookieId);
+            // In scenarios with message tampering, communication timeout can lead to an unrecoverable state.
+            this.disconnect();
+            this.connect();
         }
     }
 


bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
```diff
@@ -1043,6 +1043,9 @@ public void checkTimeoutOnPendingOperations() {
         if (timedOutOperations > 0) {
             LOG.info("Timed-out {} operations to channel {} for {}",
                      timedOutOperations, channel, bookieId);
+            // In scenarios with message tampering, communication timeout can lead to an unrecoverable state.
+            this.disconnect();
```
hangc0276, line 1047: If the timeout is not caused by message tampering, will the reconnect operation cause any side effects?

hangc0276 (changes requested):
Do we need any tests to cover your case?
Need more eyes @eolivelli @dlg99 @merlimat @zymap

Shoothzj:
@hangc0276 Yes, I'm still working on it.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
Shoothzj: @hangc0276 I believe there shouldn't be an issue here. However, if everyone feels that its impact is significant or it's too sensitive, we might consider adding a threshold for reconnection.

shoothzj:
Recently, I have no time to work on this, closed for now.

PR #4023, opened 2023-07-08
Fix headerBuffer not released in `DefaultEntryLogger.scanEntryLog`

lifepuzzlefun:
it should be released

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/DefaultEntryLogger.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/DefaultEntryLogger.java
@@ -1027,6 +1027,7 @@ public class DefaultEntryLogger implements EntryLogger {
                 pos += entrySize;
             }
         } finally {
+            ReferenceCountUtil.release(headerBuffer);
             ReferenceCountUtil.release(data);
         }
     }


horizonzy (commented):
We should also release the `headerBuffer` when `getChannelForLogId` throw exception.

hangc0276:
> We should also release the `headerBuffer` when `getChannelForLogId` throw exception.

+1

lifepuzzlefun:
Nice suggestion! The change has been committed.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/DefaultEntryLogger.java
```diff
@@ -968,12 +968,18 @@ public void scanEntryLog(long entryLogId, EntryLogScanner scanner) throws IOExce
         // Buffer where to read the entrySize (4 bytes) and the ledgerId (8 bytes)
         ByteBuf headerBuffer = Unpooled.buffer(4 + 8);
         BufferedReadChannel bc;
+        boolean throwExceptionWhenGetChannel = false;
         // Get the BufferedChannel for the current entry log file
         try {
             bc = getChannelForLogId(entryLogId);
         } catch (IOException e) {
             LOG.warn("Failed to get channel to scan entry log: " + entryLogId + ".log");
+            throwExceptionWhenGetChannel = true;
```
AnonHxy, line 977: Why not call `headerBuffer.release()` here?

lifepuzzlefun: Thanks for review. I think the code function is the same. release resource in finally code block is more like the java style.

horizonzy: agree with @AnonHxy

hangc0276 (commented):
Would you please add a test to cover this change? thanks.

PR #4030, opened 2023-07-15
Release readBuffer when close `BufferedReadChannel` & `BufferedChannel`

lifepuzzlefun:
### Motivation

Release readBuffer when close `BufferedReadChannel` & `BufferedChannel`

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedChannel.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedChannel.java
@@ -101,6 +101,11 @@ public class BufferedChannel extends BufferedReadChannel implements Closeable {
         if (closed) {
             return;
         }
+
+        super.close();
+        writeBufferStartPosition.set(0);
+        writeBuffer.clear();
+
         ReferenceCountUtil.release(writeBuffer);
         fileChannel.close();
         closed = true;
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedReadChannel.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedReadChannel.java
@@ -46,8 +46,6 @@ public class BufferedReadChannel extends BufferedChannelBase  {
 
     private boolean closed = false;
 
-    private boolean isFileChannelOwner = false;
-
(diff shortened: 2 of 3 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedChannel.java
```diff
@@ -101,6 +101,11 @@ public synchronized void close() throws IOException {
         if (closed) {
             return;
         }
+
+        super.close();
+        writeBufferStartPosition.set(0);
+        writeBuffer.clear();
```
hangc0276, line 107: The writeBuffer will be released, do we need to clear it?

hangc0276 (changes requested):
Would you please provide more context about this change?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedChannel.java
lifepuzzlefun: Agree no need to clear buffer, clear statement removed.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedReadChannel.java
```diff
@@ -103,4 +105,17 @@ public synchronized void clear() {
         readBuffer.clear();
     }
 
+    public synchronized void close() throws IOException {
+        if (closed) {
+            return;
+        }
+
+        readBufferStartPosition = Long.MIN_VALUE;
+        readBuffer.release();
```
hangc0276, line 114: We'd better use `ReferenceCountUtil.release(readBuffer);` instead of `readBuffer.release();`

hangc0276 (approved):
I support this change. Would you please add a test to cover this change? Especially cover the `readBuffer` leak case.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedReadChannel.java
lifepuzzlefun: Thanks for review, fixed.

zymap (changes requested):

dlg99 (commented):
In general LGTM, but needs a test and more context on scenarios when this is a problem

PR #4032, opened 2023-07-17
Support markLedgerReplicated command.

horizonzy:
Support markLedgerReplicated command.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
@@ -48,6 +48,7 @@ import org.apache.bookkeeper.meta.LedgerUnderreplicationManager;
 import org.apache.bookkeeper.replication.ReplicationException;
 import org.apache.bookkeeper.tools.cli.commands.autorecovery.ListUnderReplicatedCommand;
 import org.apache.bookkeeper.tools.cli.commands.autorecovery.LostBookieRecoveryDelayCommand;
+import org.apache.bookkeeper.tools.cli.commands.autorecovery.MarkLedgerReplicatedCommand;
 import org.apache.bookkeeper.tools.cli.commands.autorecovery.QueryAutoRecoveryStatusCommand;
 import org.apache.bookkeeper.tools.cli.commands.autorecovery.ToggleCommand;
 import org.apache.bookkeeper.tools.cli.commands.autorecovery.TriggerAuditCommand;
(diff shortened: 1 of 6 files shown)


hangc0276:
@horizonzy Would you please add more background of adding this command?

horizonzy:
> Support markLedgerReplicated command.

horizonzy:
> @horizonzy Would you please add more background of adding this command?

Done

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieShell.java
```diff
@@ -702,6 +704,48 @@ int runCmd(CommandLine cmdLine) throws Exception {
...
     }
 
+    /**
+     * Command to delete a given ledger.
+     */
+    class MarkLedgerReplicatedCmd extends MyCommand {
+
+        MarkLedgerReplicatedCmd() {
+            super(CMD_MARKLEDGERREPLICATED);
+            opts.addOption("l", "ledgerid", true, "Ledger ID");
```
zymap, line 714: It's better to provide a list or range option to allow given a group of ledger id to mark. For example, we can use `,` to give a list of ledgers, "1,2,3,4,5,6". We also can use a range to specify the ledger, "1...10". 
This can avoid calling this command multiple times when there has many ledgers that need to be marked.

horizonzy: Make sense.

hangc0276: +1

hangc0276 (commented):
Please add the documents for the newly added command. Please refer to https://bookkeeper.apache.org/docs/reference/cli#bookkeeper-shell

bookkeeper-server/src/main/java/org/apache/bookkeeper/tools/cli/commands/autorecovery/MarkLedgerReplicatedCommand.java
```diff
@@ -0,0 +1,135 @@
+/*
```
captain-hoity-toity, line 1: The code is generally well-written and follows good practices. However, there are a few points that could be improved:

1. Exception Handling: You are throwing `UncheckedExecutionException` for almost all types of exceptions. It would be better to handle different exceptions separately and provide more specific messages for each one.

2. Method `markLedgerReplicated()`: It is doing multiple operations like getting user confirmation, marking ledger replicated etc. Consider breaking it down into multiple smaller methods for better readability and maintainability.

3. Method `initLedgerIdFormatter()`: The if-else conditions can be simplified. You can initialize `ledgerIdFormatter` to a default value and only change it if `flags.ledgerIdFormatter` is not equal to `DEFAULT`.

4. Magic Number: Avoid using magic numbers in your code (like -1 in `flags.ledgerId < 0`). Instead, define them as constants with meaningful names.

5. `LedgerIdFormatter` initialization: The `initLedgerIdFormatter()` method is called from `apply()`, and then `markLedgerReplicated()` is called. It's not clear why the `LedgerIdFormatter` can't be initialized directly in the `markLedgerReplicated()` method.

6. Commenting: Comments explaining the purpose of each method will make the code easier to understand for other developers. 

7. Use of `force` flag: The `force` flag is used to bypass the user confirmation. This could be risky if used unintentionally. It might be good to add a warning or additional confirmation when `force` is used.

8. Variable naming: The variable `confirm` is not very descriptive. Consider renaming it to something like `userConfirmed`.

PR #4038, opened 2023-07-26
Improve DefaultEntryLogger read performance.

horizonzy:
If the BufferChannel is BufferedReadChannel, the size won't change. We can cache the fileSize to avoid time-consuming operation.
 
### Performance tests.

The entry log `f` file holds 1GB data. 
```
-rw-r--r--@  1 horizon  staff   1.0G  7 26 17:47 f.log
```

#### 1. Test DefaultEntryLogger.scanEntryLog (10x performance improvement)
```
    public static void main(String[] args) throws IOException {
        ServerConfiguration serverConfiguration = new ServerConfiguration();
        serverConfiguration.setLedgerDirNames(new String[]{"/Users/horizon/Downloads/bk1/bk-data"});
        DefaultEntryLogger defaultEntryLogger = new DefaultEntryLogger(serverConfiguration);
        long l = System.currentTimeMillis();
        AtomicInteger entryCount = new AtomicInteger();
        defaultEntryLogger.scanEntryLog(15, new EntryLogScanner() {
            @Override
            public boolean accept(long ledgerId) {
                return true;
            }

            @Override
            public void process(long ledgerId, long offset, ByteBuf entry) throws IOException {
                entryCount.incrementAndGet();
            }
        });
        long spend = System.currentTimeMillis() - l;
        System.out.println("Scan entry end, entry count: " + entryCount.get());
        System.out.println("Spend: " + spend);
    }

```
Before this pr:
```
Scan entry end, entry count: 17886769
Spend: 37273
```
After this pr:
```
Scan entry end, entry count: 17886769
Spend: 3363
```

#### 2. Test DefaultEntryLogger.readFromLogChannel (30x performance improvement)

```
    public static void main(String[] args) throws IOException {
        ServerConfiguration serverConfiguration = new ServerConfiguration();
        serverConfiguration.setLedgerDirNames(new String[]{"/Users/horizon/Downloads/bk1/bk-data"});
        DefaultEntryLogger defaultEntryLogger = new DefaultEntryLogger(serverConfiguration);
        long l = System.currentTimeMillis();
        BufferedReadChannel channel = defaultEntryLogger.getChannelForLogId(15);
        int pos = 0;
        //Read total 1gb data from the BufferedReadChannel.
        long gb = 1024 * 1024 * 1024;
        ByteBuf byteBuf = ByteBufAllocator.DEFAULT.heapBuffer(10);
        while (pos < gb) {
            defaultEntryLogger.readFromLogChannel(15, channel, byteBuf, 0);
            pos += byteBuf.readableBytes();
            byteBuf.clear();
        }
        long spend = System.currentTimeMillis() - l;
        System.out.println("Spend: " + spend);
    }
```

Before this pr:
```
Spend: 69810
```

After this pr:
```
Spend: 2338
```

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedReadChannel.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedReadChannel.java
@@ -43,6 +43,7 @@ public class BufferedReadChannel extends BufferedChannelBase  {
 
     long invocationCount = 0;
     long cacheHitCount = 0;
+    private long fileSize = -1;
 
     public BufferedReadChannel(FileChannel fileChannel, int readCapacity) {
         super(fileChannel);
@@ -64,10 +65,18 @@ public class BufferedReadChannel extends BufferedChannelBase  {
         return read(dest, pos, dest.writableBytes());
     }
 
+    @Override
+    public long size() throws IOException {
+        if (fileSize == -1) {
+            fileSize = validateAndGetFileChannel().size();
+        }
+        return fileSize;
+    }
+
     public synchronized int read(ByteBuf dest, long pos, int length) throws IOException {
         invocationCount++;
         long currentPosition = pos;
-        long eof = validateAndGetFileChannel().size();
+        long eof = size();
         // return -1 if the given position is greater than or equal to the file's current size.
         if (pos >= eof) {
             return -1;
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedReadChannel.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedReadChannel.java
@@ -44,7 +44,7 @@ public class BufferedReadChannel extends BufferedChannelBase {
     long invocationCount = 0;
     long cacheHitCount = 0;
     private long fileSize = -1;
-    private final boolean sealed;
+    final boolean sealed;
 
     public BufferedReadChannel(FileChannel fileChannel, int readCapacity) {
         this(fileChannel, readCapacity, false);
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedReadChannel.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedReadChannel.java
@@ -46,6 +46,10 @@ public class BufferedReadChannel extends BufferedChannelBase {
     private long fileSize = -1;
     private final boolean sealed;
 
+    public BufferedReadChannel(FileChannel fileChannel, int readCapacity) {
+        this(fileChannel, readCapacity, false);
+    }
+
     public BufferedReadChannel(FileChannel fileChannel, int readCapacity, boolean sealed) {
         super(fileChannel);
         this.sealed = sealed;
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedReadChannel.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedReadChannel.java
@@ -43,7 +43,7 @@ public class BufferedReadChannel extends BufferedChannelBase {
 
     long invocationCount = 0;
     long cacheHitCount = 0;
-    private long fileSize = -1;
+    private volatile long fileSize = -1;
     final boolean sealed;
 
     public BufferedReadChannel(FileChannel fileChannel, int readCapacity) {
@@ -75,7 +75,11 @@ public class BufferedReadChannel extends BufferedChannelBase {
     public long size() throws IOException {
         if (sealed) {
             if (fileSize == -1) {
-                fileSize = validateAndGetFileChannel().size();
+                synchronized (this) {
+                    if (fileSize == -1) {
+                        fileSize = validateAndGetFileChannel().size();
+                    }
+                }
             }
             return fileSize;
         } else {
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedReadChannel.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedReadChannel.java
@@ -30,7 +30,7 @@ import java.nio.channels.FileChannel;
 /**
  * A Buffered channel without a write buffer. Only reads are buffered.
  */
-public class BufferedReadChannel extends BufferedChannelBase  {
+public class BufferedReadChannel extends BufferedChannelBase {
 
     // The capacity of the read buffer.
     protected final int readCapacity;
@@ -44,9 +44,11 @@ public class BufferedReadChannel extends BufferedChannelBase  {
     long invocationCount = 0;
     long cacheHitCount = 0;
     private long fileSize = -1;
+    private final boolean sealed;
 
-    public BufferedReadChannel(FileChannel fileChannel, int readCapacity) {
+    public BufferedReadChannel(FileChannel fileChannel, int readCapacity, boolean sealed) {
         super(fileChannel);
+        this.sealed = sealed;
         this.readCapacity = readCapacity;
         this.readBuffer = Unpooled.buffer(readCapacity);
     }
@@ -67,10 +69,14 @@ public class BufferedReadChannel extends BufferedChannelBase  {
 
     @Override
     public long size() throws IOException {
-        if (fileSize == -1) {
-            fileSize = validateAndGetFileChannel().size();
+        if (sealed) {
+            if (fileSize == -1) {
+                fileSize = validateAndGetFileChannel().size();
+            }
+            return fileSize;
+        } else {
+            return validateAndGetFileChannel().size();
         }
-        return fileSize;
     }
 
     public synchronized int read(ByteBuf dest, long pos, int length) throws IOException {
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLoggerAllocator.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLoggerAllocator.java
@@ -90,6 +90,10 @@ class EntryLoggerAllocator {
         return preallocatedLogId;
     }
 
+    public long getCurrentWritingLogId() {
+        return entryLogPreAllocationEnabled ? preallocatedLogId - 1 : preallocatedLogId;
+    }
+
     BufferedLogChannel createNewLog(File dirForNextEntryLog) throws IOException {
         synchronized (createEntryLogLock) {
             BufferedLogChannel bc;
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLoggerAllocator.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLoggerAllocator.java
@@ -63,8 +63,8 @@ class EntryLoggerAllocator {
     private final boolean entryLogPreAllocationEnabled;
     private final ByteBufAllocator byteBufAllocator;
     final ByteBuf logfileHeader = Unpooled.buffer(DefaultEntryLogger.LOGFILE_HEADER_SIZE);
-    private long writingLogId = -1;
-    private long writingCompactingLogId = -1;
+    private volatile long writingLogId = -1;
+    private volatile long writingCompactingLogId = -1;
 
     EntryLoggerAllocator(ServerConfiguration conf, LedgerDirsManager ledgerDirsManager,
                          DefaultEntryLogger.RecentEntryLogsStatus recentlyCreatedEntryLogsStatus, long logId,
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLoggerAllocator.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLoggerAllocator.java
@@ -101,19 +101,15 @@ class EntryLoggerAllocator {
             BufferedLogChannel bc;
             if (!entryLogPreAllocationEnabled) {
                 // create a new log directly
-                bc = allocateNewLog(dirForNextEntryLog);
-                writingLogId = bc.getLogId();
-                return bc;
+                return allocateNewLog(dirForNextEntryLog);
             } else {
                 // allocate directly to response request
                 if (null == preallocation) {
                     bc = allocateNewLog(dirForNextEntryLog);
-                    writingLogId = bc.getLogId();
                 } else {
                     // has a preallocated entry log
                     try {
                         bc = preallocation.get();
-                        writingLogId = bc.getLogId();
                     } catch (ExecutionException ee) {
                         if (ee.getCause() instanceof IOException) {
                             throw (IOException) (ee.getCause());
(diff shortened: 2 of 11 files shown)


dlg99:
Good idea, great results.

Can you please check that these "Bookie Tests" failures aren't related the change:
```
Error:  Errors: 
Error:  org.apache.bookkeeper.bookie.DefaultEntryLogTest.testAppendLedgersMapOnCacheRemoval
Error:    Run 1: DefaultEntryLogTest.testAppendLedgersMapOnCacheRemoval:1223 » IO Not all ledge...
Error:    Run 2: DefaultEntryLogTest.testAppendLedgersMapOnCacheRemoval:1223 » IO Not all ledge...
Error:    Run 3: DefaultEntryLogTest.testAppendLedgersMapOnCacheRemoval:1223 » IO Not all ledge...
[INFO] 
Error:  org.apache.bookkeeper.bookie.storage.ldb.DbLedgerStorageTest.testEntriesOutOfOrderWithFlush
Error:    Run 1: DbLedgerStorageTest.testEntriesOutOfOrderWithFlush:400 » IO Bad entry read fro...
Error:    Run 2: DbLedgerStorageTest.testEntriesOutOfOrderWithFlush:400 » IO Bad entry read fro...
Error:    Run 3: DbLedgerStorageTest.testEntriesOutOfOrderWithFlush:400 » IO Bad entry read fro...
[INFO] 
Warning:  Flakes: 
Warning:  InterleavedLedgerStorageTest.testConsistencyCheckConcurrentGC
[INFO]   Run 1: PASS
Error:    Run 2: expected:<0> but was:<8354>
Error:    Run 3: expected:<0> but was:<7255>
Error:    Run 4: expected:<0> but was:<7712>
```

I'll restart the job to see if these are flaps though I don't think these tests flapped before.

horizonzy:
> Good idea, great results.
> 
> Can you please check that these "Bookie Tests" failures aren't related the change:
> 
> ```
> Error:  Errors: 
> Error:  org.apache.bookkeeper.bookie.DefaultEntryLogTest.testAppendLedgersMapOnCacheRemoval
> Error:    Run 1: DefaultEntryLogTest.testAppendLedgersMapOnCacheRemoval:1223 » IO Not all ledge...
> Error:    Run 2: DefaultEntryLogTest.testAppendLedgersMapOnCacheRemoval:1223 » IO Not all ledge...
> Error:    Run 3: DefaultEntryLogTest.testAppendLedgersMapOnCacheRemoval:1223 » IO Not all ledge...
> [INFO] 
> Error:  org.apache.bookkeeper.bookie.storage.ldb.DbLedgerStorageTest.testEntriesOutOfOrderWithFlush
> Error:    Run 1: DbLedgerStorageTest.testEntriesOutOfOrderWithFlush:400 » IO Bad entry read fro...
> Error:    Run 2: DbLedgerStorageTest.testEntriesOutOfOrderWithFlush:400 » IO Bad entry read fro...
> Error:    Run 3: DbLedgerStorageTest.testEntriesOutOfOrderWithFlush:400 » IO Bad entry read fro...
> [INFO] 
> Warning:  Flakes: 
> Warning:  InterleavedLedgerStorageTest.testConsistencyCheckConcurrentGC
> [INFO]   Run 1: PASS
> Error:    Run 2: expected:<0> but was:<8354>
> Error:    Run 3: expected:<0> but was:<7255>
> Error:    Run 4: expected:<0> but was:<7712>
> ```
> 
> I'll restart the job to see if these are flaps though I don't think these tests flapped before.

I have checked the testAppendLedgersMapOnCacheRemoval and testEntriesOutOfOrderWithFlush. The problem is that the BufferedReadChannel and BufferedLogChannel share the same FileChannel. When the BufferedReadChannel finishes reading the fileSize, the BufferedLogChannel writes new data into the file. I will verify if this situation exists. In theory, the FileChannel read by the BufferedLogChannel should not write any new data.

Shoothzj (approved):

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/EntryLogger.java
```diff
@@ -129,4 +129,6 @@ default EntryLogMetadata getEntryLogMetadata(long entryLogId) throws IOException
      * @return false if the entrylog doesn't exist.
      */
     boolean removeEntryLog(long entryLogId);
+
+    String name();
```
eolivelli, line 133: Is this a 'name' or a 'type' ?

eolivelli (changes requested):
Overall I support this change.

But I am not sure about the 'name' name.
Can we fins something that explains better the purpose of this property?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/EntryLogger.java
horizonzy: make sense.

Reidddddd:
I feel Enrico's purpose is to require some method docs to explain why it needs the method `name` or `type` (anyway)

And i skimmed the changes, seems `if (xxx instance of DefaultEntryLogger)` will resolve it? no need to add this method

Reidddddd:
> > And i skimmed the changes, seems `if (xxx instance of DefaultEntryLogger)` will resolve it? no need to add this method
> 
> Using instanceOf is not a good practice, It violates polymorphism.

`compactingLogWriteDone` is only a private method in `TransactionalEntryLogCompactor.java`, not a common method in `AbstractLogCompactor`.

What you did, in order to serve a private function in a concrete sub class, you added a function and parameters for other classes which don't need them

When you are talking about polymorphism, polymorphism allows each sub-class has its own behaviour, you are enforcing all the sub-class to satisfy a concrete need not a common need, you are the one who breaks polymorphism...

horizonzy:
> > > And i skimmed the changes, seems `if (xxx instance of DefaultEntryLogger)` will resolve it? no need to add this method
> > 
> > 
> > Using instanceOf is not a good practice, It violates polymorphism.
> 
> `compactingLogWriteDone` is only a private method in `TransactionalEntryLogCompactor.java`, not a common method in `AbstractLogCompactor`.
> 
> What you did, in order to serve a private function in a concrete sub class, you added a function and parameters for other classes which don't need them
> 
> When you are talking about polymorphism, polymorphism allows each sub-class has its own behaviour, you are enforcing all the sub-class to satisfy a concrete need not a common need, you are the one who breaks polymorphism...

Ok, that makes sense. I would like to public an interface to do the work.

horizonzy:
> I feel Enrico's purpose is to require some method docs to explain why it needs the method `name` or `type` (anyway)
> 
> And i skimmed the changes, seems `if (xxx instance of DefaultEntryLogger)` will resolve it? no need to add this method

Finally, take your suggestion, thanks.

horizonzy:
rerun failure checks

horizonzy:
rerun failure checks

horizonzy:
rerun failure checks

Reidddddd:
+1 from me

horizonzy:
rerun failure checks

hangc0276 (commented):
Great job! Would you please add a test to protect this change? Especially for the sealed flag

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLoggerAllocator.java
```diff
@@ -127,10 +136,16 @@ BufferedLogChannel createNewLog(File dirForNextEntryLog) throws IOException {
...
         synchronized (createCompactionLogLock) {
-            return allocateNewLog(dirForNextEntryLog, COMPACTING_SUFFIX);
+            BufferedLogChannel bc = allocateNewLog(dirForNextEntryLog, COMPACTING_SUFFIX);
+            writingCompactingLogId = bc.getLogId();
+            return bc;
         }
     }
 
+    void clearCompactingLogId() {
+        writingCompactingLogId = -1;
```
zymap, line 146: Do we need to guard this by `synchronized (createCompactionLogLock)`?  It seems not possible to have race conditions, but I see the `createNewLogForCompaction(File dirForNextEntryLog)` is guarded.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedReadChannel.java
```diff
@@ -43,9 +43,12 @@ public class BufferedReadChannel extends BufferedChannelBase  {
 
     long invocationCount = 0;
     long cacheHitCount = 0;
+    private long fileSize = -1;
+    private final boolean sealed;
 
-    public BufferedReadChannel(FileChannel fileChannel, int readCapacity) {
+    public BufferedReadChannel(FileChannel fileChannel, int readCapacity, boolean sealed) {
```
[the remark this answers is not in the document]
horizonzy: Make sense.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLoggerAllocator.java
horizonzy: No need this, the `createNewLogForCompaction` and `clearCompactingLogId` are serial.

horizonzy:
From the user report, they cherry-pick this commit, and they found that there is a problem about it.

```
Caused by: org.apache.bookkeeper.bookie.DefaultEntryLogger$EntryLookupException$MissingEntryException: pos 1027337410 (entry 135 for ledgerId 574) past end of entryLog 46

```

I will try to find the problem with them.

horizonzy:
> From the user report, they cherry-pick this commit, and they found that there is a problem about it.
> 
> ```
> Caused by: org.apache.bookkeeper.bookie.DefaultEntryLogger$EntryLookupException$MissingEntryException: pos 1027337410 (entry 135 for ledgerId 574) past end of entryLog 46
> ```
> 
> I will try to find the problem with them.

The user config `setEntryLogPerLedgerEnabled=true`, so the defaultEntryLogger will write multi entry log file at the same time, so the sealed flag is not set correctly.

If the config `setEntryLogPerLedgerEnabled=true`, the sealed flag always be false.

horizonzy:
> Great job! Would you please add a test to protect this change? Especially for the sealed flag

Done.

Reidddddd:
Hope to see this improvement in the next release (thumb up

eolivelli (approved):
Lgtm

hangc0276:
@merlimat @lhotari Please help take a look at this PR, thanks.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedReadChannel.java
```diff
@@ -64,10 +71,22 @@ public int read(ByteBuf dest, long pos) throws IOException {
...
+    public long size() throws IOException {
+        if (sealed) {
+            if (fileSize == -1) {
+                fileSize = validateAndGetFileChannel().size();
+            }
+            return fileSize;
+        } else {
+            return validateAndGetFileChannel().size();
+        }
+    }
```
lhotari, line 84: Would there be a benefit to override this logic in `BufferedChannel` class where the `position` (and possibly `unpersistedBytes`) field information would be leveraged? Doesn't `position` contain the file size when it's writable?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLoggerAllocator.java
```diff
@@ -90,16 +92,19 @@ synchronized long getPreallocatedLogId() {
         return preallocatedLogId;
     }
 
+    public boolean isSealed(long logId) {
+        return logId != writingLogId && logId != writingCompactingLogId;
+    }
```
lhotari, line 97: Isn't this misleading if it is called "isSealed"?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLoggerAllocator.java
lhotari: > No need this, the `createNewLogForCompaction` and `clearCompactingLogId` are serial.

What do you mean?

lhotari: Just wondering if there's a chance that writingLogId and writingCompactingLogId field changes wouldn't be visible for another thread that reads these fields.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedReadChannel.java
hangc0276: The FileChnanel doesn't have `position` field and the `position()` method will trigger the OS system call

lhotari: I was referring to https://github.com/apache/bookkeeper/blob/22097343ceed1312964ad077cdcd9fbd7e3e026b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedChannel.java#L49-L52 in BufferedChannel class. In the case that the instance is a BufferedChannel class, perhaps it could use that information?

Is the challenge in that case that the same file could have a BufferedReadChannel instance and a BufferedChannel instance at the same time?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLoggerAllocator.java
horizonzy: `isSealed` means that the entry log file will no longer be written to, we needn't care that the file sile change.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLoggerAllocator.java
horizonzy: > Just wondering if there's a chance that writingLogId and writingCompactingLogId field changes wouldn't be visible for another thread that reads these fields.

Make sense, make it volatile.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedReadChannel.java
horizonzy: I'm not got you fully, the BufferedChannel means the corresponding file is writing, it override the `BufferedReadChannel#read` method, it know the file end position.

> Is the challenge in that case that the same file could have a BufferedReadChannel instance and a BufferedChannel instance at the same time?

https://github.com/apache/bookkeeper/pull/4038#issuecomment-1652685967 
The test contains the case.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedReadChannel.java
```diff
@@ -64,10 +71,22 @@ public int read(ByteBuf dest, long pos) throws IOException {
         return read(dest, pos, dest.writableBytes());
     }
 
+    @Override
+    public long size() throws IOException {
+        if (sealed) {
+            if (fileSize == -1) {
```
merlimat, line 77: nit: if the file is already "sealed", should we read the size in the constructor? This way we wouldn't have any concurrency issues here if 2 threads call `size()` at the size time.

merlimat (approved):

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BufferedReadChannel.java
horizonzy: The `size()` would throw IOException, I don't want to make the constructor announce the exception, so make the fileSize double check to avoid concurrency.

PR #4044, opened 2023-08-07
Allow to set max operation numbers in a single rocksdb batch

zymap:
---

## Motivation

In rocksdb, memory usage is related to batch size. The more operations in a single batch, the more memory is consumed. Expose the configuration to allow control the batch size.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageRocksDB.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageRocksDB.java
@@ -84,6 +84,7 @@ public class KeyValueStorageRocksDB implements KeyValueStorage {
     private final ReadOptions optionCache;
     private final ReadOptions optionDontCache;
     private final WriteBatch emptyBatch;
+    private final int writeBatchMaxSize;
 
     private String dbPath;
 
@@ -143,6 +144,8 @@ public class KeyValueStorageRocksDB implements KeyValueStorage {
 
         optionCache.setFillCache(true);
         optionDontCache.setFillCache(false);
+
+        this.writeBatchMaxSize = conf.getMaxOperationNumbersInSingleRocksDBBatch();
     }
 
     private RocksDB initializeRocksDBWithConfFile(String basePath, String subPath, DbConfigType dbConfigType,
@@ -516,21 +519,29 @@ public class KeyValueStorageRocksDB implements KeyValueStorage {
 
     @Override
     public Batch newBatch() {
-        return new RocksDBBatch();
+        return new RocksDBBatch(writeBatchMaxSize);
     }
 
     private class RocksDBBatch implements Batch {
         private final WriteBatch writeBatch = new WriteBatch();
+        private final int batchSize;
+        private int batchCount = 0;
+
+        RocksDBBatch(int batchSize) {
+            this.batchSize = batchSize;
+        }
 
         @Override
         public void close() {
             writeBatch.close();
+            batchCount = 0;
         }
 
         @Override
         public void put(byte[] key, byte[] value) throws IOException {
             try {
                 writeBatch.put(key, value);
+                countBatchAndFlushIfNeeded();
             } catch (RocksDBException e) {
                 throw new IOException("Failed to flush RocksDB batch", e);
             }
@@ -540,6 +551,7 @@ public class KeyValueStorageRocksDB implements KeyValueStorage {
         public void remove(byte[] key) throws IOException {
             try {
                 writeBatch.delete(key);
+                countBatchAndFlushIfNeeded();
             } catch (RocksDBException e) {
                 throw new IOException("Failed to flush RocksDB batch", e);
             }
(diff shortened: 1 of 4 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageRocksDB.java
```diff
@@ -548,17 +560,31 @@ public void remove(byte[] key) throws IOException {
         @Override
         public void clear() {
             writeBatch.clear();
+            batchCount = 0;
```
horizonzy, line 563: If the user invoke flush() directly, we should also reset the batchCount

bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageTest.java
```diff
@@ -170,4 +170,41 @@ public void simple() throws Exception {
...
+        assertEquals(2, batch.batchCount());
+
+        batch.put(toArray(3), toArray(3));
+        batch.put(toArray(4), toArray(4));
+        batch.put(toArray(5), toArray(5));
+        assertEquals(0, batch.batchCount());
+        batch.put(toArray(6), toArray(6));
+        assertEquals(1, batch.batchCount());
+
+        batch.flush();
```
horizonzy, line 202: Here trigger flush() directly, the `batchCount `should be 0.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageRocksDB.java
```diff
@@ -548,17 +560,31 @@ public void remove(byte[] key) throws IOException {
...
         }
 
         @Override
         public void deleteRange(byte[] beginKey, byte[] endKey) throws IOException {
             try {
                 writeBatch.deleteRange(beginKey, endKey);
+                countBatchAndFlushIfNeeded();
```
horizonzy, line 570: We need to discuss whether deleteRange should also be performed within a batch. A deleteRange operation can potentially delete a large number of keys. If we include multiple deleteRange operations within a single batch, I am concerned that it may introduce issues. The previous core dump issue could be related to this.

zymap:
I was tested with this code to verify the batch impaction on memory.
https://gist.github.com/zymap/19249ab35bb0f64c55cbf7f2e8356cb3
I found the memory keeps increasing with the batch size. And if the batch is not flushed into the sst, it will save into the WAL file, and the WAL file will not be limited by `max_total_wal_size`. 
If it was OOM killed because of the large batch and the batch was saved in the WAL. The only way to reopen the rocksDB is to add more memory for the bookie.
I also talk this issue with rocksDB community, they said:
>when the batch size is so large (esp if you run multiple batches together) the wal size may reach (limit + batch-size * number of open batches). We have a project opened by our friends from Kafka streams to handle huge batch size. In the meanwhile can you restrict the size of your batch ?

--
In the Pulsar, the compacted ledger hasn't a rollover policy or retention policy. If the user has tons of keys in the compaction, that would make the compacted ledger bigger and bigger. In our environment, a compacted ledger reached 200G. It contains lots of entries in a single ledger, which makes the batch very large.
In release 4.14.7 and branch-4.15, we didn't limit the delete operation numbers in a single batch. 
https://github.com/apache/bookkeeper/blob/c22136b03489db1643521f586e9cae2c4a511e10/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/EntryLocationIndex.java#L241
Finally, when bookie runs the garbage collection and removes the ledger, it will be OOM killed because of the large batch.

Pulsar already has a proposal about configuring the compacted topic ledger retention, https://github.com/apache/pulsar/issues/19665. 
But I think we also need to have a way to control the batch size to make sure we have a way to limit the memory.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageRocksDB.java
zymap: That should be another issue. This PR aims to limit the batch size. We can handle that with another PR.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageRocksDB.java
zymap: I don't think so. Users can flush multiple times using the same batch. Even you flush the batch, the batch's content doesn't clean up. If we reset it after flushing, the count is not accurate. So I follow the API to reset the batch.

horizonzy: If manual flush is completed and the batchCount reaches the batchSize again, does it mean that the operations in this batch will be applied to the database twice?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageRocksDB.java
```diff
@@ -540,6 +551,7 @@ public void put(byte[] key, byte[] value) throws IOException {
         public void remove(byte[] key) throws IOException {
             try {
                 writeBatch.delete(key);
+                countBatchAndFlushIfNeeded();
```
hangc0276, line 554: When we try to remove a key from the writeBatch, but the writeBatch has been triggered flushed in `put` method, will it impact the correctness of the key value stored in the RocksDB?

zymap: It won't. The batch should only affect when the action starts to do in the rocksdb, we're just pending our operations into the batch to avoid calling the real action method multiple times. 
See here: https://github.com/facebook/rocksdb/blob/9a034801cead6421bcf82b506b77e3b2251f1edb/include/rocksdb/write_batch.h#L9

hangc0276 (approved):
Nice catch!

dlg99 (approved):

horizonzy (approved):
LGTM.

PR #4046, opened 2023-08-11
Fix bookie shell getUsage msg

AnonHxy:

### Motivation
Fix bookie shell getUsage message.

When we execute the bookkeeper shell `bin/bookkeeper shell`,  it will print the help messages:
image

However the long options  should be `--`, instead of`-`;

### Changes

Change `-` to `--` in the `getUsage()` method.


AnonHxy:
rerun failure checks

AnonHxy:
rerun failure checks

AnonHxy:
Oh..  The long cmd with `-` could also work.  So it looks that there is no need to change the usage messages.  I will close this patch. Thanks all

PR #4051, opened 2023-08-15
BP62-New API for batched reads

horizonzy:
### Motivation
Introduce the batch read API to improve read performance.


site3/website/src/pages/bps/BP-62-new-API-for-batched-reads.md
```diff
@@ -0,0 +1,305 @@
...
+Non-Batch: Read 1000130 entries in 199928ms
+```
+The key difference with the batch read is that the BookKeeper client sends only 10,000 requests to the server 
+(compared to 1,000,000 in non-batch read), and the server sends 10,000 responses back to the client (versus 1,000,000 in non-batch read).
+This approach reduces the number of RPC calls made and eliminates the need for the server to process a large number of 
+requests from the IO thread and the entry read thread pool, resulting in a significant improvement in performance.
+
+Hence, the objective of this proposal is to enhance the performance of entry reading by introducing a batch entry reading protocol 
+that takes into account the expected count and size of entries.
+- **Optimize entry reading performance:** By reading multiple entries in a single RPC request, the network communication and RPC call overhead can be reduced, thereby optimizing the reading performance.
```
merlimat, line 38: We should also add the effect of 1 failed read operation in a batch of 100 (eg: because the bookie rejects the operation when the executor queue is full), where the entire batch is failed in the BK client and we have to perform the entire batch read again.

merlimat (approved):
👍

zymap (approved):

hangc0276 (approved):

wenbingshen (approved):
Nice work!

site3/website/src/pages/bps/BP-62-new-API-for-batched-reads.md
horizonzy: See section `Loose constraints`. If the batch read request read entries [0-100]. Then the bookie received the request, If reading entries [0-50] is successful, but reading entry 51 fails, we will return entries [0-50] to the client and trigger the callback. The data returned to the user will only include entries [0-50]. The user will need to read the data starting from entry 51 in the next attempt.

PR #4055, opened 2023-08-21
Add explanatory messages to the asserts

Taher-Ghaleb:
I am working on research that investigates test smell refactoring in which we identify alternative implementations of test cases, study how commonly used these refactorings are, and assess how acceptable they are in practice. This smell is known as _Assertion Roulette_, where a test method has multiple asserts without explanatory messages, which can sometimes make it difficult to identify which assert has failed. We would like to get your feedback on this particular test refactoring. Thanks in advance for your input.


hangc0276 (approved):

Taher-Ghaleb:
Hi, any updates or feedback on the changes, please? Thanks.

PR #4058, opened 2023-08-25
Remove underreplicaiton callback

horizonzy:
Address https://lists.apache.org/thread/1xl3hr2cpyd5xh9kozbx5xlfsjsg3f4h
Deprecate underreplicaiton callback to relieve zookeeper pressure.


bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/LedgerUnderreplicationManager.java
```diff
@@ -240,6 +240,7 @@ boolean initializeLostBookieRecoveryDelay(int lostBookieRecoveryDelay)
      * @param cb
      * @throws ReplicationException.UnavailableException
      */
+    @Deprecated
```
hangc0276, line 243: We'd better add a default empty implementation for the interface, otherwise, all the implementation classes need to implement this method

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
```diff
@@ -558,14 +556,26 @@ private void scheduleReplicasCheckTask() {
...
-    private class UnderReplicatedLedgersChangedCb implements GenericCallback<Void> {
-        @Override
-        public void operationComplete(int rc, Void result) {
+    private void scheduleReplicateStatTask() {
+        long interval = conf.getAuditorUnderReplicasStatInterval();
+        if (interval <= 0) {
+            LOG.info("Under Replicas Stats disabled");
+            return;
+        }
+        auditorStatExecutor = Executors.newSingleThreadScheduledExecutor(
```
hangc0276, line 565: Why not reuse the existing executor?

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
```diff
@@ -558,14 +556,26 @@ private void scheduleReplicasCheckTask() {
...
+                    public Thread newThread(Runnable r) {
+                        Thread t = new Thread(r, "AuditorStat-" + bookieIdentifier);
+                        t.setDaemon(true);
+                        return t;
+                    }
+                });
+        auditorStatExecutor.scheduleAtFixedRate(() -> {
             Iterator<UnderreplicatedLedger> underreplicatedLedgersInfo = ledgerUnderreplicationManager
                     .listLedgersToRereplicate(null);
             auditorStats.getUnderReplicatedLedgersGuageValue().set(Iterators.size(underreplicatedLedgersInfo));
```
hangc0276, line 577: We update the Gauge metric with a specific interval, and I'm not sure whether it is meaningful for the metric on the dashboard.

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/LedgerUnderreplicationManager.java
horizonzy: Make sense.

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
horizonzy: I don't want it to block the auditor task. If there are too many under replication ledgers, the iteration will take too much time. And the existing executor is a single thread executor, it will block the auditor task.

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/Auditor.java
horizonzy: I think this metric doesn't need to be highly real-time. We just need to give users an approximate value and indicate whether it has changed. If real-time values are required, we can retrieve the current data of under-replicated ledgers using the BookKeeper shell.

hangc0276: If it is better to use `bin/bookkeeper shell` to get the count of under-replicated ledgers directly?

horizonzy: Yes, it is. If so, we can remove the auditorStatExecutor. If the user wants to know the under-replication ledger stats, use the bookkeeper shell instead of auditor metrics.

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
```diff
@@ -214,6 +214,7 @@ public class ServerConfiguration extends AbstractConfiguration<ServerConfigurati
     protected static final String UNDERREPLICATED_LEDGER_RECOVERY_GRACE_PERIOD =
             "underreplicatedLedgerRecoveryGracePeriod";
     protected static final String AUDITOR_REPLICAS_CHECK_INTERVAL = "auditorReplicasCheckInterval";
+    protected static final String AUDITOR_UNDER_REPLICAS_STAT_INTERVAL = "auditorUnderReplicasStatInterval";
```
hangc0276, line 217: Please remove this configuration

hangc0276 (approved):

zymap (approved):

PR #4060, opened 2023-08-28
[release] Force to use linux/amd64 to build release

zymap:
---

### Motivation

Because we have a Native IO that needs to be built from cpp code. The build platform will impact the target library. People are using different laptop to execute the release process with docker, we need to force the image to use linux/amd64 platform to ensure the .so lib build out the correct platform lib.
Since most of time we are using amd64 to run the bookie, making the dev docker image build from amd64.

### Changes

Update the dev Dockerfile to use the specified platform.


eolivelli (approved):
Lgtm

Shoothzj (approved):

hangc0276 (approved):

PR #4062, opened 2023-08-28
Entry read support local node rack awareness

hangc0276:
### Motivation
- The entry write supports the local node Rack-Aware placement policy but does not support the local node Region-Aware placement policy
- The entry read supports the local node Region-Aware placement policy but does not support the local node Rack-Aware placement policy

In order to match the local node region/rack awareness both on entry write and read, we need to support the following feature
- Entry write supports local node region awareness placement policy
- Entry read supports local node rack awareness placement policy

### Modification
The PR aims to support entry read local node awareness placement policy.


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
```diff
@@ -1318,4 +1322,11 @@ private BookieNode replaceToAdherePlacementPolicyInternal(
         }
         throw new BKNotEnoughBookiesException();
     }
+
+    protected String getLocalRack(BookieNode node) {
+        if (null == node || null == node.getAddr()) {
+            return UNKNOWN_RACK;
+        }
+        return node.getNetworkLocation();
```
horizonzy, line 1330: If the network location is `/default-rack`, shall we think of it as UNKNOWN_RACK?

PR #4063, opened 2023-08-29
Entry write support local node region aware placement policy

hangc0276:
### Motivation
The entry write supports the local node Rack-Aware placement policy but does not support the local node Region-Aware placement policy
The entry read supports the local node Region-Aware placement policy but does not support the local node Rack-Aware placement policy
In order to match the local node region/rack awareness both on entry write and read, we need to support the following feature

Entry write supports local node region awareness placement policy
Entry read supports local node rack awareness placement policy

### Modification
- The PR aims to support the entry write local node region awareness placement policy. 
- It will select one bookie that is located in the same region and rack with the local node to form the ledger ensemble.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RegionAwareEnsemblePlacementPolicy.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RegionAwareEnsemblePlacementPolicy.java
@@ -388,9 +388,20 @@ public class RegionAwareEnsemblePlacementPolicy extends RackawareEnsemblePlaceme
                 remainingEnsembleBeforeIteration = remainingEnsemble;
                 int regionsToAllocate = numRemainingRegions;
                 int startRegionIndex = lastRegionIndex % numRegionsAvailable;
+                int localRegionIndex = -1;
+                if (myRegion != null && !UNKNOWN_REGION.equals(myRegion)) {
+                    localRegionIndex = availableRegions.indexOf(myRegion);
+                }
+                String region = myRegion;
                 for (int i = 0; i < numRegionsAvailable; ++i) {
-                    String region = availableRegions.get(startRegionIndex % numRegionsAvailable);
-                    startRegionIndex++;
+                    // select the local region first, and for the rest region select, use round-robin selection.
+                    if (i > 0 || localRegionIndex == -1) {
+                        if (startRegionIndex % numRegionsAvailable == localRegionIndex) {
+                            startRegionIndex++;
+                        }
+                        region = availableRegions.get(startRegionIndex % numRegionsAvailable);
+                        startRegionIndex++;
+                    }
                     final Pair<Integer, Integer> currentAllocation = regionsWiseAllocation.get(region);
                     TopologyAwareEnsemblePlacementPolicy policyWithinRegion = perRegionPlacement.get(region);
                     if (!regionsReachedMaxAllocation.contains(region)) {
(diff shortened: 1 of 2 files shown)


hangc0276:
@horizonzy Please help take a look at this PR, thanks.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/TestRegionAwareEnsemblePlacementPolicy.java
```diff
@@ -1880,4 +1880,111 @@ public void testNotifyRackChangeWithNewRegion() throws Exception {
...
+                bookie1Count++;
+            }
+            if (ensemble.contains(addr8.toBookieId())) {
+                bookie8Count++;
+            }
+            if (ensemble.contains(addr9.toBookieId())) {
+                bookie9Count++;
+            }
+
+            LOG.info("[hangc] ensemble: {}", ensemble);
```
horizonzy, line 1953: We can remove it.

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/TestRegionAwareEnsemblePlacementPolicy.java
```diff
@@ -1880,4 +1880,111 @@ public void testNotifyRackChangeWithNewRegion() throws Exception {
...
+                fail("Failed to select bookie located on the same region and rack with bookie client");
+            }
+            if (ensemble.contains(addr2.toBookieId()) && ensemble.contains(addr3.toBookieId())) {
+                fail("addr2 and addr3 is same rack.");
+            }
+        }
+        LOG.info("Bookie1 Count: {}, Bookie8 Count: {}, Bookie9 Count: {}", bookie1Count, bookie8Count, bookie9Count);
+
+        //addr4 shutdown.
+        addrs.remove(addr5.toBookieId());
```
horizonzy, line 1964: What's the purpose of this operation(remove addr5)

horizonzy (approved):
LGTM.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RegionAwareEnsemblePlacementPolicy.java
```diff
@@ -388,9 +388,20 @@ public PlacementResult<List<BookieId>> newEnsemble(int ensembleSize, int writeQu
                 remainingEnsembleBeforeIteration = remainingEnsemble;
                 int regionsToAllocate = numRemainingRegions;
                 int startRegionIndex = lastRegionIndex % numRegionsAvailable;
+                int localRegionIndex = -1;
```
zymap, line 391: Maybe a more clear way to do this is to remove the localRegion from the availableRegions then everything happen smoothly with original logic.

zymap (commented):
Because it changes the current behaviors, do we need to allow it to configure?

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RegionAwareEnsemblePlacementPolicy.java
horizonzy: Maybe not. If so, the `localRegion` node can't pick again. If we want to pick 5 nodes, and there are only two regions. If we remove `localRegion`, then we can only pick nodes from the remaining one region.

merlimat (approved):

PR #4069, opened 2023-09-02
Parallel checkpoint when dbledgerStorage config multi ledgerStorageDir

lifepuzzlefun:
### Motivation

Our production env config multi journal and with one ledgerStorage on nvme disk.
but the problem occurs with checkpoint which is slow. and disk have much more performance to use.

after we changed to multi ledgerStorageDir, the SyncThread will become bottleneck, which will try to checkpoint one by one for all the ledgerStorage.

### Changes

if the DbLedgerStorage is config with multi ledgerStorage Dir try to submit each dirs task to executor to speed up checkpoint

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
@@ -39,6 +39,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.PrimitiveIterator.OfLong;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


hangc0276:
>after we changed to multi ledgerStorageDir, the SyncThread will become bottleneck

@lifepuzzlefun Could you provide more detail about the SyncThread bottleneck?
Checkpoint will triggered in two ways:
- Triggered by addEntry in SingleDirectoryDbLedgerStorage, which is executed by the thread in SingleDirectoryDbLedgerStorage
- Triggered by SyncThread every 60s by default

When high throughput writes in to bookie, the SyncThread won't be the bottleneck for the throughput because the checkpoint will be triggered by the thread in SingleDirectoryDbLedgerStorage

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/DbLedgerStorage.java
```diff
@@ -388,8 +404,41 @@ public void flush() throws IOException {
 
     @Override
     public void checkpoint(Checkpoint checkpoint) throws IOException {
+        if (checkpointExecutor != null) {
```
hangc0276, line 407: The checkpoint is executed by SyncThread

hangc0276 (changes requested):

PR #4077, opened 2023-09-10
Fix bug of negative JournalQueueSize

XLzed:

### Motivation

While shouldRolloverJournal is true，the QueueEntry qe is not added to the toFlush queue but we decrease the JournalQueueSize incorrectly.

### Changes

Decrease the JournalQueueSize after correct NonNull check of qe.


hangc0276 (approved):
Nice catch!

horizonzy (approved):
Nice catch. It solves my confusion

PR #4085, opened 2023-09-16
Fix verifyDirsForNewEnvironment

AnonHxy:

### Motivation

When verify dires for new environment,  we should make sure that all directories should be empty, instead of `missedCookieDirs`

### Changes

`verifyDirsForNewEnvironment(missedCookieDirs)` -> `verifyDirsForNewEnvironment(directories)`

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LegacyCookieValidation.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LegacyCookieValidation.java
@@ -94,7 +94,7 @@ public class LegacyCookieValidation implements CookieValidation {
             if (!missedCookieDirs.isEmpty()) {
                 if (rmCookie == null) {
                     // 5.1 new environment: all directories should be empty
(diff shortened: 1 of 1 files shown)


AnonHxy:
rerun failure checks

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LegacyCookieValidation.java
```diff
@@ -94,7 +94,7 @@ public void checkCookies(List<File> directories) throws BookieException {
             if (!missedCookieDirs.isEmpty()) {
                 if (rmCookie == null) {
                     // 5.1 new environment: all directories should be empty
-                    verifyDirsForNewEnvironment(missedCookieDirs);
+                    verifyDirsForNewEnvironment(directories);
```
hangc0276, line 97: Have you encountered any exceptions? The missedCookieDirs is the result of checking remote cookies and directories

hangc0276 (changes requested):

PR #3433, opened 2022-07-28
indexDir compatible fix for issue #3430

StevenLuMT:
Descriptions of the changes in this PR:

### Motivation
Fixes https://github.com/apache/bookkeeper/issues/3430, make cookie validation is compatible with old version.
I think we should check indexDirs when configuration‘s indexDirectories is setted  rather than avoiding the check 

### Changes
1. add compatible check 
2. add testcase to cover the new code

PR #4087, opened 2023-09-25
Make after version5 cookie index dir compatible with before version5.

horizonzy:

### Motivation
Fixes https://github.com/apache/bookkeeper/pull/3433#issuecomment-1729231396

(


horizonzy:
rerun failure checks

PR #4095, opened 2023-10-03
[Bug] LedgerManagerFactory Info storing in LAYOUT znode is not correct

Reidddddd:

### Motivation

Fix a bug described in #<4094>

### Changes

When executing `format` method, as `layoutManager` is inited already as user configured, the `LedgerManagerFactory` info should be read from znode, instead of configuration.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractZkLedgerManagerFactory.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractZkLedgerManagerFactory.java
@@ -62,9 +62,11 @@ public abstract class AbstractZkLedgerManagerFactory implements LedgerManagerFac
 
         Class<? extends LedgerManagerFactory> factoryClass;
         try {
-            factoryClass = conf.getLedgerManagerFactoryClass();
-        } catch (ConfigurationException e) {
-            throw new IOException("Failed to get ledger manager factory class from configuration : ", e);
+            factoryClass = (Class<? extends LedgerManagerFactory>)
+                    Class.forName(layoutManager.readLedgerLayout().getManagerFactoryClass());
+        } catch (ClassNotFoundException e) {
+            // should not reach here, as LayoutManager was successfully inited
+            throw new IOException("Failed to read ledger manager factory class from LAYOUT : ", e);
         }
 
         layoutManager.deleteLedgerLayout();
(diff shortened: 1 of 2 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractZkLedgerManagerFactory.java
```diff
@@ -62,9 +62,11 @@ public void format(AbstractConfiguration<?> conf, LayoutManager layoutManager)
...
-        } catch (ConfigurationException e) {
-            throw new IOException("Failed to get ledger manager factory class from configuration : ", e);
+            factoryClass = (Class<? extends LedgerManagerFactory>)
+                    Class.forName(layoutManager.readLedgerLayout().getManagerFactoryClass());
+        } catch (ClassNotFoundException e) {
+            // should not reach here, as LayoutManager was successfully inited
+            throw new IOException("Failed to read ledger manager factory class from LAYOUT : ", e);
         }
 
         layoutManager.deleteLedgerLayout();
```
Reidddddd, line 72: as here it wipes the LAYOUT in znode, `factoryClass` by default is null, which will use `HierarchicalLedgerManager` as default in the following `createNewLMFactory`

Shoothzj (approved):

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractZkLedgerManagerFactory.java
hangc0276: The LAYOUT is `HierarchicalLedgerManager` but the user configuration is `LongHierarchicalLedgerManagerFactory`, I think we should fail the format. Otherwise, the user may think the LAYOUT  has been formatted to `LongHierarchicalLedgerManagerFactory`, but actually it doesn't

hangc0276 (changes requested):

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractZkLedgerManagerFactory.java
Reidddddd: > The LAYOUT is `HierarchicalLedgerManager` but the user configuration is `LongHierarchicalLedgerManagerFactory`, I think we should fail the format. Otherwise, the user may think the LAYOUT has been formatted to `LongHierarchicalLedgerManagerFactory`, but actually it doesn't

emm, don't get your point, seems to be some misunderstanding here

Reidddddd: > The LAYOUT is `HierarchicalLedgerManager` but the user configuration is `LongHierarchicalLedgerManagerFactory`, I think we should fail the format. Otherwise, the user may think the LAYOUT has been formatted to `LongHierarchicalLedgerManagerFactory`, but actually it doesn't

Let's put it straight, user configured `longhierarchical` and executed `format` which means using `LongHierarchicalLedgerManagerFactory. 

But the fact,  znode is written `HierarchicalLedgerManager` which is not configured, it is just a value from `if null then by default`

So in the UT, if w/o this fix, the second time `format` will fail because of mismatching `HierarchicalLedgerManager` and `LongHierarchicalLedgerManagerFactory`. Normally, it should be good to execute `format` as much as many time as configurations are the same

PR #4104, opened 2023-10-15
Fix ledgerHandle pendingAddEntries stats

lifepuzzlefun:
### Motivation
Current LedgerHandle stats NUM_PENDING_ADD only record the first LedgerHandle create by bookkeeper instance.
if we need record all LedgerHandle at bookkeeper instance level the goal fail.

### Changes
1. Record all LedgerHandle instance in static field.
2. register NUM_PENDING_ADD gauge by sum all ledgerHandle pendingAppOp queue size.
3. remove LedgerHandle from static field when handle is closed.
4. add new metric NUM_OPEN_LEDGER_HANDLE to expose all open LedgerHandle numbers by this bookkeeper client.


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
```diff
@@ -92,6 +92,8 @@
  * write operations to a ledger.
  */
 public class LedgerHandle implements WriteHandle {
+    public static Set<LedgerHandle> INSTANCES = ConcurrentHashMap.newKeySet();
```
hangc0276, line 95: Store the LedgerHandle object will bring extra memory overload for the client.

hangc0276 (changes requested):

PR #4111, opened 2023-10-18
Fix bookie process still alive when bookie startup is not successful and shutdown

TakaHiR07:
### Motivation

When bookie startup encounter IOException in BookieImpl#readJournal(), bookie startup is not successful and then trigger shutdown(). However, bookie process is still alive. 

The reason is IOException is caught in BookieImpl and trigger shutdown. Bookie actually is not running. But the exception do not throw to startComponent. 

So BookieServer.main / server.Main.doMain still wait for the startComponent future to complete.

The relevant code is as following:

https://github.com/apache/bookkeeper/blob/3221aa30924825cb4c1a5b00fb68dec44712946e/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java#L654-L660

https://github.com/apache/bookkeeper/blob/3221aa30924825cb4c1a5b00fb68dec44712946e/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieServer.java#L123-L131

https://github.com/apache/bookkeeper/blob/3221aa30924825cb4c1a5b00fb68dec44712946e/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/component/ComponentStarter.java#L63-L87

### Changes

throw exception to startComponent when bookie start encounter error.


bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/BookieInitializationTest.java
```diff
@@ -821,6 +882,8 @@ public void testBookieStartException() throws Exception {
         loggerOutput.expect((List<LoggingEvent> logEvents) -> {
```
hangc0276, line 882: We'd better assert startFuture and startFuture2 separately.

TakaHiR07: have done.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieServer.java
```diff
@@ -127,7 +127,7 @@ public void start() throws InterruptedException, IOException {
         if (!this.bookie.isRunning()) {
             exitCode = bookie.getExitCode();
             this.requestProcessor.close();
-            return;
+            throw new InterruptedException("fail fast, bookie startup is not successful");
```
horizonzy, line 130: We would better define another Exception; the `InterruptedException` is for the `thread` case.

TakaHiR07: Can we define it "BookieStartException", extends the BookieException ?

horizonzy: Sounds good.

TakaHiR07: have done.

hangc0276:
@TakaHiR07 Could you take a look at the check style issue? Thanks.

hangc0276 (approved):

TakaHiR07:
> @TakaHiR07 Could you take a look at the check style issue? Thanks.

have fix checkstyle.

horizonzy (approved):
LGTM.

hangc0276:
@TakaHiR07 Would you please help fix the failed tests? Thanks.

TakaHiR07:
> @TakaHiR07 Would you please help fix the failed tests? Thanks.

@hangc0276 I have fix the test since powermock is removed in master branch.

shoothzj (approved):

shoothzj:
update the branch

xiang092689:
Normally, "Main" thread will wait for "component-shutdown-thread" thread to complete the "component start future". Your edition will cause future complete too soon, and main thread exit, which may cause shutdown not finished and process exit?

shoothzj (changes requested):
I think it's not an APIException, should we define this exception in https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/ExitCode.java ?

TakaHiR07:
> Normally, "Main" thread will wait for "component-shutdown-thread" thread to complete the "component start future". Your edition will cause future complete too soon, and main thread exit, which may cause shutdown not finished and process exit?

"component start future" would complete only when exception throw. if bookie.start() encounter exception, it may catch some exception and go into shutdown(). only shutdown finish, then bookie.isRunning() become false, then go into the edition code.

TakaHiR07:
> I think it's not an APIException, should we define this exception in https://github.com/apache/bookkeeper/blob/master/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/ExitCode.java ?

ExitCode looks like pointing out the exact reason why bookie start fail. It seems not appropriate to put this exception here, this exception only point out that bookie start fail.

TakaHiR07:
Many failed test shows that in previous design, bookieServer start failed, but not throw exception. I think these test is also unreasonable and should be fixed.  Looking forward to more views, then I would fix these test.

PR #4118, opened 2023-10-24
Improve auto-recovery noise log when some bookie down.

horizonzy:
When some bookie is down, the resolver will resolve failed, it will print many logs. 

1. In RackawareEnsemblePlacementPolicyImpl#replaceBookie, reduce the repeat resolve operation.
2. In TopologyAwareEnsemblePlacementPolicy#resolveNetworkLocation, avoid to print the stack info.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
@@ -507,13 +507,27 @@ public class RackawareEnsemblePlacementPolicyImpl extends TopologyAwareEnsembleP
         try {
             excludeBookies = addDefaultRackBookiesIfMinNumRacksIsEnforced(excludeBookies);
             excludeBookies.addAll(currentEnsemble);
+    
+            Set<Node> ensembleNodes = new HashSet<>();
+            Set<Node> excludeNodes = new HashSet<>();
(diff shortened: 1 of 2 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
```diff
@@ -507,13 +507,27 @@ public PlacementResult<BookieId> replaceBookie(int ensembleSize, int writeQuorum
...
             BookieNode bn = knownBookies.get(bookieToReplace);
             if (null == bn) {
                 bn = createBookieNode(bookieToReplace);
             }
-
-            Set<Node> ensembleNodes = convertBookiesToNodes(currentEnsemble);
-            Set<Node> excludeNodes = convertBookiesToNodes(excludeBookies);
+            for (BookieId bookieId : currentEnsemble) {
+                if (bookieId.equals(bookieToReplace)) {
+                    ensembleNodes.add(bn);
```
hangc0276, line 519: Can we skip the `bn` directly, because the following code is
```
excludeNodes.addAll(ensembleNodes);
excludeNodes.add(bn);
ensembleNodes.remove(bn);
```

horizonzy: Make sense.

merlimat (approved):

PR #4119, opened 2023-10-26
Fix shell start localbookie if not specify NUMBER

AnonHxy:

If we  run  the cmd `bin/bookkeeper localbookie` without specify the $NUMBER, it will exit without any message.  It's better print the help message 

### Motivation


AnonHxy:
@zymap @hangc0276 @eolivelli PTAL

bin/bookkeeper
```diff
@@ -162,6 +162,10 @@ elif [ ${COMMAND} == "autorecovery" ]; then
   exec "${JAVA}" ${OPTS} ${JMX_ARGS} org.apache.bookkeeper.replication.AutoRecoveryMain --conf ${BOOKIE_CONF} $@
 elif [ ${COMMAND} == "localbookie" ]; then
   NUMBER=$1
+  if [[ ! "$NUMBER" =~ ^[0-9]+$ ]]; then
+      bookkeeper_help;
+      exit 1;
+  fi
```
zymap, line 168: The local bookie already handles the argument checks in the code. https://github.com/apache/bookkeeper/blob/acde49c349b6856f25ab0b00ef50412b844182a0/bookkeeper-server/src/main/java/org/apache/bookkeeper/util/LocalBookKeeper.java#L358
So we can change this to 

```
args="$1 ${BOOKIE_CONF} $@"
  exec "${JAVA}" ${OPTS} ${JMX_ARGS} -Dzookeeper.4lw.commands.whitelist='*' org.apache.bookkeeper.util.LocalBookKeeper ${args}
```

AnonHxy: Thanks for your review @zymap 
Do you mean:
```
  args="$1"
  shift
  args="$args ${BOOKIE_CONF} $@"
  exec "${JAVA}" ${OPTS} ${JMX_ARGS} -Dzookeeper.4lw.commands.whitelist='*' org.apache.bookkeeper.util.LocalBookKeeper $args
```
I don't think this is a solution. The same problem still exists if we just run the cmd following:
```
bin/bookkeeper localbookie
```

PR #4120, opened 2023-10-29
Increase journal thread priority

hangc0276:
### Motivation
When deploying Bookie on K8s with limited CPU resources, once the CPU is throttled, the journal write latency will increase.
One case of causing CPU throttle is garbage collection triggered.
image

image

The journal write latency is sensitive with the journal CPU, especially when the CPU is throttled.

### Modifications
Increase the JournalThread and ForceWriteThread priority to reduce the latency impaction when the CPU is being throttled.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
@@ -471,6 +471,7 @@ public class Journal extends BookieCriticalThread implements CheckpointSource {
                                 boolean enableGroupForceWrites,
                                 StatsLogger statsLogger) {
             super("ForceWriteThread");
+            this.setPriority(Thread.MAX_PRIORITY);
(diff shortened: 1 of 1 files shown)


hangc0276:
@merlimat @eolivelli @zymap @dlg99 @lhotari Please help take a look at this PR, thanks.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/Journal.java
```diff
@@ -471,6 +471,7 @@ public ForceWriteThread(Thread threadToNotifyOnEx,
                                 boolean enableGroupForceWrites,
                                 StatsLogger statsLogger) {
             super("ForceWriteThread");
+            this.setPriority(Thread.MAX_PRIORITY);
```
zymap, line 474: How about using a configuration to specify it?

merlimat (approved):

PR #4122, opened 2023-11-01
Fix RegionAwarePolicy can not update rackInfo between bookie left and join

TakaHiR07:
### Motivation

When using regionAwarePolicy in pulsar, we encounter a case as following:
1. bookie shutdown, trigger handleBookiesThatLeft()
2. change bookie's rackInfo, trigger onBookieRackChange()
3. bookie start, trigger handleBookiesThatJoined()

However, the rackInfo is not updated actually, broker still write ensemble with old rackInfo. 

### Changes

1. The Hashmap "address2Region" is local variable. It should be updated in handleBookiesThatJoined()
2. add a test to verify

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RegionAwareEnsemblePlacementPolicy.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RegionAwareEnsemblePlacementPolicy.java
@@ -134,7 +134,8 @@ public class RegionAwareEnsemblePlacementPolicy extends RackawareEnsemblePlaceme
             topology.add(node);
             knownBookies.put(addr, node);
             historyBookies.put(addr, node);
-            String region = getLocalRegion(node);
+            String region = parseBookieRegion(addr);
+            address2Region.put(addr, region);
             if (null == perRegionPlacement.get(region)) {
                 perRegionPlacement.put(region, new RackawareEnsemblePlacementPolicy()
                         .initialize(dnsResolver, timer, this.reorderReadsRandom, this.stabilizePeriodSeconds,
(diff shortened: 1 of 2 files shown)


hangc0276:
@horizonzy Would you please help take a look at this PR? Thanks.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RegionAwareEnsemblePlacementPolicy.java
```diff
@@ -134,7 +134,8 @@ public void handleBookiesThatJoined(Set<BookieId> joinedBookies) {
             topology.add(node);
             knownBookies.put(addr, node);
             historyBookies.put(addr, node);
-            String region = getLocalRegion(node);
+            String region = parseBookieRegion(addr);
```
horizonzy, line 137: Shall we update the `address2Region` in the `onBookieRackChange`. 
Even if the bookie already has been shutdown, we can still update the `address2Region` in the `onBookieRackChange`

TakaHiR07: I think current code has updated address2Region in onBookieRackChange(). If oldRegion == newRegion, no need to update. If oldRegion != newRegion, would put in newRegion.

 https://github.com/apache/bookkeeper/blob/9aad0453f0f17f45ea673e82fb63762156315e8b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RegionAwareEnsemblePlacementPolicy.java#L188-L194

horizonzy: > I think current code has updated address2Region in onBookieRackChange(). If oldRegion == newRegion, no need to update. If oldRegion != newRegion, would put in newRegion.
> 
> https://github.com/apache/bookkeeper/blob/9aad0453f0f17f45ea673e82fb63762156315e8b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RegionAwareEnsemblePlacementPolicy.java#L188-L194

If the line_178 node is null, it does nothing.

TakaHiR07: Is we need to get correct rack of a shutdown bookie ? It seem is enough to update bookie's rack once it restart.

horizonzy: I check the code. It's hard to get the rack info when the bookie is shutdown, because we can't resolve the address.

Your solution is good enough.

horizonzy (approved):
LGTM.

PR #4123, opened 2023-11-02
Change callbacks in CleanupLedgerManager as a Set

AnonHxy:

### Motivation

Currently  the type of `CleanupLedgerManager#callbacks` is map, and the key and value are both the same `callbacks`. We just used it as a set. So this patch will change the `callbacks`  from map to set

### Changes

Change `callbacks` in `CleanupLedgerManager` as a Set


AnonHxy:
rerun failure checks

AnonHxy:
rerun failure checks

AnonHxy:
@zymap @eolivelli @hangc0276  PTAL

hangc0276 (approved):
LGTM

PR #4124, opened 2023-11-03
Cleanup layoutManager null-check when AbstractZkLedgerManagerFactory#newLedgerManagerFactory

AnonHxy:

### Motivation

Remove the null-check from `AbstractZkLedgerManagerFactory#newLedgerManagerFactory`. The reason is that:
1. the parameter `layoutManager` in `newLedgerManagerFactory()` method can never be null. So this check is unnecessory
2. If the `layoutManager` is null and we return  `new FlatLedgerManagerFactory().initialize(conf, null, FlatLedgerManagerFactory.CUR_VERSION)`,  the 'zk' in `FlatLedgerManagerFactory` is null and could throw NPE when invoke the methods in the `FlatLedgerManagerFactory`.

### Changes

1. Cleanup layoutManager null-check when `newLedgerManagerFactory()`
2. Let `layoutManager` could never be null in `FlatLedgerManagerFactory`

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractZkLedgerManagerFactory.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/AbstractZkLedgerManagerFactory.java
@@ -164,12 +164,6 @@ public abstract class AbstractZkLedgerManagerFactory implements LedgerManagerFac
             throw new IOException("Empty Ledger Root Path.");
         }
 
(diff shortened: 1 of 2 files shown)


AnonHxy:
rerun failure checks

AnonHxy:
rerun failure checks

AnonHxy:
rerun failure checks

AnonHxy:
rerun failure checks

Issue #4126, opened 2023-11-09
GC thread shuts down slowly when compacting


erobot:
**BUG REPORT**

***Describe the bug***

GC thread shuts down slowly when compacting.

***To Reproduce***

1. A bookie is major compacting.
2. Stop the bookie.
3. The bookie does not exit in a reasonable time.

***Expected behavior***

Bookie exits in a reasonable time.

***Additional context***

#### Version

4.15.4

#### Problem

I think the problem is that the stop flag `running` has been moved after `compacting.compareAndSet`. When `running` is not set to false, entry log continues to compact one after one and the shutdown thread is hard to set the `compacting`.

shutdown from 4.14.8:
https://github.com/apache/bookkeeper/blob/965c3627328787e2750f41c2a4eda6c1d709d7c6/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java#L528-L535

shutdown from 4.15.4:
https://github.com/apache/bookkeeper/blob/44f87efb8121a9f5428e7ddc70656aee85f563bf/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java#L652-L665

#### Stack dump

component-shutdown-thread is waiting gc thread to shutdown and has been waiting for 5153.82s.

```
"component-shutdown-thread" #177 prio=5 os_prio=0 cpu=3148.35ms elapsed=5153.82s tid=0x00007fac601f4880 nid=0x14dee waiting on condition  [0x00007fb97d0e4000]
   java.lang.Thread.State: TIMED_WAITING (sleeping)
        at java.lang.Thread.sleep(java.base@17.0.9/Native Method)
        at org.apache.bookkeeper.bookie.GarbageCollectorThread.shutdown(GarbageCollectorThread.java:662)
        - locked <0x00007fbcce3885f0> (a org.apache.bookkeeper.bookie.GarbageCollectorThread)
        at org.apache.bookkeeper.bookie.storage.ldb.SingleDirectoryDbLedgerStorage.shutdown(SingleDirectoryDbLedgerStorage.java:267)
        at org.apache.bookkeeper.bookie.storage.ldb.DbLedgerStorage.shutdown(DbLedgerStorage.java:222)
        at org.apache.bookkeeper.bookie.BookieImpl.shutdown(BookieImpl.java:865)
        at org.apache.bookkeeper.bookie.BookieImpl.shutdown(BookieImpl.java:834)
[...]

PR #4127, opened 2023-11-09
Issue 4126: Fix a slow gc thread shutdown when compacting

erobot:

### Motivation

Fix a slow gc thread shutdown when compacting.

The problem here is that the stop flag `running` has been moved after `compacting.compareAndSet`. When `running` is not set to false, entry log continues to compact one after one and the shutdown thread is hard to set the `compacting`.

### Changes

Set `running` to false first and then check `compacting`.

Master Issue: #4126


zymap (approved):

horizonzy (approved):
LGTM.

hangc0276 (approved):
Good Catch!

PR #4132, opened 2023-11-21
Readledger command did not read data when specifying the bookie address

liubingxing:
When we use the **bookkeeper shell readledger -bookie** command to read a ledger, and specifying the bookie address without specifying the entry range, no data return.
Because the lastEntry is -1 and no entry will be read.
image
So, I think we should open the ledger first to get the LAC and then read the entry.


liubingxing:
It is not a problem. 
Because bookie may not have the full entry data for the ledger. So if we specify bookie addr to read the ledger, preferably specify the scope of entry.

Issue #1898, opened 2019-01-10
implement isEnsembleAdheringToPlacementPolicy method for RegionAwareEnsemblePlacementPolicy


reddycharan:
**FEATURE REQUEST**

As part of https://github.com/apache/bookkeeper/pull/1883, EnsemblePlacementPolicy implementations are expected to be enhanced to log relevant metrics. This uses/depends on isEnsembleAdheringToPlacementPolicy method for each EnsemblePlacementPolicy. But for RegionAwareEnsemblePlacementPolicy this method is quite complicated considering the number of config parameters at play for RegionAwareEnsemblePlacementPolicy. So creating separate issue to implement this.

PR #4133, opened 2023-11-23
Issue 1898: Implement `isEnsembleAdheringToPlacementPolicy` in `RegionAwareEnsemblePlacementPolicy`

dragonls:
Master Issue: #1898

### Motivation
Before this PR, `isEnsembleAdheringToPlacementPolicy` in `RegionAwareEnsemblePlacementPolicy` always return `PlacementPolicyAdherence.MEETS_STRICT`, which is not good to keep data highly available while using `RegionAwareEnsemblePlacementPolicy` , especially when one region is down.

### Changes

Implement `isEnsembleAdheringToPlacementPolicy` in `RegionAwareEnsemblePlacementPolicy`.

The main implementation idea is that when all allocations are satisfied in different regions, it is considered `PlacementPolicyAdherence.MEETS_STRICT`, otherwise `PlacementPolicyAdherence.FAIL`.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RegionAwareEnsemblePlacementPolicy.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RegionAwareEnsemblePlacementPolicy.java
@@ -41,6 +41,7 @@ import org.apache.bookkeeper.net.NodeBase;
 import org.apache.bookkeeper.proto.BookieAddressResolver;
 import org.apache.bookkeeper.stats.StatsLogger;
 import org.apache.bookkeeper.util.BookKeeperConstants;
+import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.lang3.tuple.Pair;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


hangc0276:
@horizonzy Please help take a look at this PR, thanks.

hangc0276:
@dragonls Thanks for your contribution. Would you please fix the failed check style? thanks.

dragonls:
> @dragonls Thanks for your contribution. Would you please fix the failed check style? thanks.

Fixed.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RegionAwareEnsemblePlacementPolicy.java
```diff
@@ -644,12 +645,75 @@ public final DistributionSchedule.WriteSet reorderReadLACSequence(
...
+            if (effectiveMinRegionsForDurability > 0 && regionsInQuorum.size() < effectiveMinRegionsForDurability) {
+                if (LOG.isDebugEnabled()) {
+                    LOG.debug("For ensemble {}, write set starting at {} are from {} regions, "
+                                    + "less than effectiveMinRegionsForDurability: {}.",
+                            ensembleList, i, regionsInQuorum.size(), effectiveMinRegionsForDurability);
+                }
+                return PlacementPolicyAdherence.FAIL;
+            }
+
+            if (regionsInQuorum.size() < writeQuorumSize) {
```
horizonzy, line 706: We can check it before line_679

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RegionAwareEnsemblePlacementPolicy.java
```diff
@@ -644,12 +645,75 @@ public final DistributionSchedule.WriteSet reorderReadLACSequence(
...
+        int effectiveMinRegionsForDurability = disableDurabilityFeature.isAvailable() ? 1 : minRegionsForDurability;
+
+        int ensembleSize = ensembleList.size();
+        Map<String, Set<BookieId>> regionsInQuorum = new HashMap<>();
+        BookieId bookie;
+        for (int i = 0; i < ensembleList.size(); i++) {
+            regionsInQuorum.clear();
+            for (int j = 0; j < writeQuorumSize; j++) {
+                bookie = ensembleList.get((i + j) % ensembleSize);
+                if (knownBookies.containsKey(bookie)) {
```
horizonzy, line 661: we didn't lock this method, so the result may be different between line_661 and line_662, line_661 contains the bookie, but at the line_662, the bookie may be removed. Use `knownBookies.get(bookie);` directly would be better.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RegionAwareEnsemblePlacementPolicy.java
```diff
@@ -644,12 +645,75 @@ public final DistributionSchedule.WriteSet reorderReadLACSequence(
...
+                } else if (LOG.isDebugEnabled()) {
+                    LOG.debug("bookie {} is not in the list of knownBookies", bookie);
+                }
+            }
+
+            if (regionsInQuorum.isEmpty()) {
+                return PlacementPolicyAdherence.FAIL;
+            }
+
+            if (regionsInQuorum.size() < 2) {
```
horizonzy, line 679: I have a question: why judge the `regionsInQuorum.size() < 2`?

dragonls: The purpose of doing this here is to align the implement in `org.apache.bookkeeper.client.RegionAwareEnsemblePlacementPolicy#newEnsemble`:
https://github.com/apache/bookkeeper/blob/185fef6df92713a92f1531a4044f75ffebcc6e93/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RegionAwareEnsemblePlacementPolicy.java#L356-L366

When there is only one available region, fall back to `RackAwareEnsemblePlacement`.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RegionAwareEnsemblePlacementPolicy.java
dragonls: You're right, fixed.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RegionAwareEnsemblePlacementPolicy.java
dragonls: We should judge `regionsInQuorum.size() < 2` first, in case that it falls back to `RackawareEnsemblePlacementPolicy`.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RegionAwareEnsemblePlacementPolicy.java
horizonzy: It's a little weird. 
Based on: writeQuorumSize = 3.

If regionsInQuorum < 2, the result may be `MEETS_STRICT`.

If regionsInQuorum = 2, the result is `FAIL`.

dragonls: Yes, a little weird.

Should we just remove the `numRegionsAvailable < 2` judge in `isEnsembleAdheringToPlacementPolicy`? May I have your suggestion?

horizonzy: The PlacementPolicyAdherence enum has three values `FAIL(1), MEETS_SOFT(3), MEETS_STRICT(5)`.
If there are enough different region, the result could be `MEETS_STRICT`. If there aren't enough different region, the each RackAwarePLacementPolicy `MEETS_STRICT`, the result could be `MEETS_SOFT`

dragonls: Updated.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RegionAwareEnsemblePlacementPolicy.java
```diff
@@ -360,8 +361,12 @@ public PlacementResult<List<BookieId>> newEnsemble(int ensembleSize, int writeQu
                         effectiveMinRegionsForDurability, minNumRacksPerWriteQuorum);
                 TopologyAwareEnsemblePlacementPolicy nextPolicy = perRegionPlacement.get(
                         availableRegions.iterator().next());
-                return nextPolicy.newEnsemble(ensembleSize, writeQuorumSize, writeQuorumSize,
+
+                PlacementResult<List<BookieId>> placementResult = nextPolicy.newEnsemble(ensembleSize, writeQuorumSize, writeQuorumSize,
                         comprehensiveExclusionBookiesSet, ensemble, ensemble);
+                return PlacementResult.of(placementResult.getResult(),
+                        isEnsembleAdheringToPlacementPolicy(
```
horizonzy, line 368: Here, we can simplify the logic. We already know there is only one region. 
Just need check the  result of `nextPolicy.newEnsemble`.
1. If the  result is PlacementPolicyAdherence.FAIL, return directly. 
2. If the result is  PlacementPolicyAdherence.MEETS_STRICT, then check whether minNumRacksPerWriteQuorum is more than 1. If it is, return PlacementPolicyAdherence.SOFT, otherwise return PlacementPolicyAdherence.MEETS_STRICT.

PR #4137, opened 2023-11-27
Issue 4136: Fix logging configurations are broken in docker image

### LINKED ISSUE #4136: Logging configurations are broken in the docker image
**BUG REPORT**
The logging configuration in the Docker image is broken after #3225, which switched to using `log4j2.xml` instead of `log4j.properties`.

In the Bookkeeper Docker image entrypoint, configuration files in the `conf` directory are overridden with environment variables prefixed with `BK_`. This configuration override mechanism ([apply-config-from-env.py][1]) affects all files in the `conf` directory, including `log4j2*.xml`.
However, `apply-config-from-env.py` was not intended to override XML configurations, or at least it was not originally designed for this purpose. As a result, some lines are unexpectedly dropped after the configuration override.

https://github.com/apache/bookkeeper/blob/c07e72aabd68abec18ac761e61aec289814390c4/docker/scripts/apply-config-from-env.py#L62-L70

[1]: https://github.com/apache/bookkeeper/blob/c07e72aabd68abec18ac761e61aec289814390c4/docker/scripts/apply-config-from-env.py

***Describe the bug***
After start a bookie from docker image could see error message in the console.
```
[Fatal Error] log4j2.xml:30:11: The element type "Appenders" must be terminated by the matching end-tag "</Appenders>".
ERROR StatusLogger Error parsing /opt/bookkeeper/conf/log4j2.xml
 org.xml.sax.SAXParseException; systemId: file:///opt/bookkeeper/conf/log4j2.xml; lineNumber: 30; columnNumber: 11; The element type "Appenders" must be terminated by the matching end-tag "</Appenders>".
	at java.xml/com.sun.org.apache.xerces.internal.parsers.DOMParser.parse(DOMParser.java:262)
	at java.xml/com.sun.org.apache.xerces.internal.jaxp.DocumentBuilderImpl.parse(DocumentBuilderImpl.java:342)
	at org.apache.logging.log4j.core.config.xml.XmlConfiguration.<init>(XmlConfiguration.java:94)
	at org.apache.logging.log4j.core.config.xml.XmlConfigurationFactory.getConfiguration(XmlConfigurationFactory.java:46)
	at org.apache.logging.log4j.core.config.ConfigurationFactory$Factory.getConfiguration(ConfigurationFactory.java:499)
	at org.apache.logging.log4j.core.config.ConfigurationFactory$Factory.getConfiguration(ConfigurationFactory.java:474)
	at org.apache.logging.log4j.core.config.ConfigurationFactory$Factory.getConfiguration(ConfigurationFactory.java:398)
	at org.apache.logging.log4j.core.config.ConfigurationFactory.getConfiguration(ConfigurationFactory.java:318)
	at org.apache.logging.log4j.core.LoggerContext.reconfigure(LoggerContext.java:690)
	at org.apache.logging.log4j.core.LoggerContext.reconfigure(LoggerContext.java:711)
	at org.apache.logging.log4j.core.LoggerContext.start(LoggerContext.java:253)
	at org.apache.logging.log4j.core.impl.Log4jContextFactory.getContext(Log4jContextFactory.java:155)
	at org.apache.logging.log4j.core.impl.Log4jContextFactory.getContext(Log4jContextFactory.java:47)
	at org.apache.logging.log4j.LogManager.getContext(LogManager.java:196)
	at org.apache.logging.log4j.spi.AbstractLoggerAdapter.getContext(AbstractLoggerAdapter.java:137)
	at org.apache.logging.slf4j.Log4jLoggerFactory.getContext(Log4jLoggerFactory.java:55)
	at org.apache.logging.log4j.spi.AbstractLoggerAdapter.getLogger(AbstractLoggerAdapter.java:47)
	at org.apache.logging.slf4j.Log4jLoggerFactory.getLogger(Log4jLoggerFactory.java:33)
	at org.slf4j.LoggerFactory.getLogger(LoggerFactory.java:363)
	at org.slf4j.LoggerFactory.getLogger(LoggerFactory.java:388)
	at org.apache.bookkeeper.server.Main.<clinit>(Main.java:51)
ERROR StatusLogger No logging configuration
```

***To Reproduce***

Just start a bookie from docker image, here's an example in _docker-compose_:
```yaml
version: '3'
services:
  zk:
    image: zookeeper:3.8.1
  bookie0:
    image: apache/bookkeeper:4.16.3
    environment:
      - BK_zkServers=zk:2181
    depends_on:
      - zk
```

Shawyeok:
### Motivation

Fix #4136 

### Changes

Modify `apply-config-from-env.py` to accept individual files as parameters instead of a configuration folder. This change will enable the script to selectively apply the configuration override mechanism only to the necessary files.

### DIFF
--- a/docker/scripts/apply-config-from-env.py
+++ b/docker/scripts/apply-config-from-env.py
@@ -23,21 +23,28 @@
 ## based on the ENV variables
 ## export my-key=new-value
 ##
-## ./apply-config-from-env config_dir
+## ./apply-config-from-env file ...
 ##
 
 import os, sys
 
-if len(sys.argv) != 2:
-    print('Usage: %s ' + 'config_dir' % (sys.argv[0]))
+if len(sys.argv) < 2:
+    print('Usage: %s file ...' % (sys.argv[0]))
     sys.exit(1)
 
-def mylistdir(dir):
-    return [os.path.join(dir, filename) for filename in os.listdir(dir)]
+def prepare_conf_files(files):
+    conf_files = []
+    for f in files:
+        if os.path.isfile(f):
+            if not os.path.isabs(f):
+                f = os.path.join(os.getcwd(), f)
+            conf_files.append(f)
+        else:
+            print('%s is not a readable file' % f)
+            sys.exit(1)
+    return conf_files
 
-# Always apply env config to all the files under conf
-conf_dir = sys.argv[1]
-conf_files = mylistdir(conf_dir)
+conf_files = prepare_conf_files(sys.argv[1:])
 print('conf files: ')
 print(conf_files)
 
--- a/docker/scripts/common.sh
+++ b/docker/scripts/common.sh
@@ -71,7 +71,7 @@ echo "  BK_STREAM_STORAGE_ROOT_PATH is ${BK_STREAM_STORAGE_ROOT_PATH}"
 echo "  BK_NUM_STORAGE_CONTAINERS is ${BK_NUM_STORAGE_CONTAINERS}"
 echo "  BOOKIE_GRPC_PORT is ${BOOKIE_GRPC_PORT}"
 
-python scripts/apply-config-from-env.py ${BK_HOME}/conf
+python scripts/apply-config-from-env.py ${BK_HOME}/conf/*.conf
 
 export BOOKIE_CONF=${BK_HOME}/conf/bk_server.conf
 export SERVICE_PORT=${PORT0}
--- a/docker/scripts/init_zookeeper.sh
+++ b/docker/scripts/init_zookeeper.sh
@@ -63,7 +63,7 @@ function create_zk_dynamic_conf() {
 function init_zookeeper() {
 
     # apply zookeeper envs
-    python scripts/apply-config-from-env.py ${BK_HOME}/conf
+    python scripts/apply-config-from-env.py ${BK_HOME}/conf/zookeeper.conf
 
     # create dirs if they don't exist
     create_zk_dirs


hangc0276:
@Shawyeok Please rebase the master, thanks.

Shawyeok:
> @Shawyeok Please rebase the master, thanks.

Done, thanks.

hangc0276:
@zymap Please help take a look at this PR, thanks.

zymap (changes requested):
Actually, the apply-config-from-env will go through the configuration file and only update the existing key with the format k=v. So I don't think it will update the log4j configuration file. Could you please paste your log4j configuration file as well?

Shawyeok:
> Actually, the apply-config-from-env will go through the configuration file and only update the existing key with the format k=v. So I don't think it will update the log4j configuration file. Could you please paste your log4j configuration file as well?

@zymap 
There is an edge case here. For example, with the log4j2.xml configuration below, <Console name="CONSOLE" and <Console name="CONSOLEMDC" are treated as having the same key due to their k=v pattern (here, it’s <Consolename). As a result, any configuration lines that appear later are discarded, thus breaking the validity of the XML format.
https://github.com/apache/bookkeeper/blob/5be03cfdfb4dbe4ad37bd2502fe514d9b347cd26/conf/log4j2.xml#L30-L35

Here is a docker-compose file that can quickly reproduce the issue. You will see the error information mentioned in #4136.
```yaml
version: '3'
services:
  zk:
    image: zookeeper:3.8.1
  bookie0:
    image: apache/bookkeeper:4.16.3
    environment:
      - BK_zkServers=zk:2181
    depends_on:
      - zk
```

Shawyeok:
@zymap 
The current capability of `apply-config-from-env.py` is limited to handling configuration files in the `k=v` format, like `properties` files. Applying it to XML format configurations is meaningless and may lead to unexpected configuration replacements.

zymap:
okay. How about letting the apply-conf-from-env support use the specified files to update? Most of the time, we only need to update the bookkeeper configuration file and zookeeper configuration file. I think we always manually update the rocksdb and log4j configuration files.

Shawyeok:
> okay. How about letting the apply-conf-from-env support use the specified files to update? Most of the time, we only need to update the bookkeeper configuration file and zookeeper configuration file. I think we always manually update the rocksdb and log4j configuration files.

Yes, this is indeed what this PR does: it modifies `apply-config-from-env.py` to accept individual files as parameters. The usage would be like this:

```
$ ./apply-config-from-env file ...
```

zymap:
Ah. Sorry. I misunderstood this pr. LGTM

zymap (approved):

zwOvO (approved):

raymondBourges:
Thanks @Shawyeok for this PR :-)

I use the three files it contains via docker volumes to get around the log configuration problem on my 4.16.3 image.

What do you think @zymap about also pushing this PR on the 4.16 branch so that it will be present in future 4.16.x images?

THANKS

zymap:
@raymondBourges Sure, will include this in the next release.

raymondBourges:
Thanks but 4.16.4 was already released on January 29 (but image on dockerhub still in 4.16.3).
This is why I talk about an 4.16 branch so that the PR will be included in a possible 4.16.5.
Thanks

lhotari:
I'm including this fix in #4207 and #4206 while fixing CI for branch-4.15 and branch-4.16 .

PR #4161, opened 2023-12-21
[Improvement] skip read rebundunct content when scanning entry log.

thetumbled:
Remove the rebundunt disk read when scannig the entry log file, which could reduce the pressure of disk a lot.

### Motivation
We have `org.apache.bookkeeper.bookie.storage.EntryLogScanner` to process the entry data when scanning the entry log file, in which `process` method is used to process the entry data.
We have a lot of implementations for `EntryLogScanner` for various reason.
image
But some of them do not need to access all data in the entry log file.
For example:
- scanner implemented in method `extractEntryLogMetadataByScanning` only need to know the `entrySize` field, in such case we do not need to read any data in the entry.
- scanner implemented in method `org.apache.bookkeeper.bookie.InterleavedStorageRegenerateIndexOp#initiate` only need to know the `entryId` field, in such case we only need to read the first 16 bytes(ledgerId+entryId).
- many more simillar situations...

So, we **do not need to read the entry data when scanning the entry log in some cases.**

### Changes

Add method `getLengthToRead` in `EntryLogScanner` to indicate how much data do we need to read, and read this mount of data only.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/DefaultEntryLogger.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/DefaultEntryLogger.java
@@ -1036,15 +1036,20 @@ public class DefaultEntryLogger implements EntryLogger {
                 }
                 // read the entry
                 data.clear();
-                data.capacity(entrySize);
-                int rc = readFromLogChannel(entryLogId, bc, data, pos);
-                if (rc != entrySize) {
-                    LOG.warn("Short read for ledger entry from entryLog {}@{} ({} != {})",
-                            entryLogId, pos, rc, entrySize);
-                    return;
+                int capacity = Math.min(scanner.getLengthToRead(), entrySize);
+                // skip read when scanner.getLengthToRead() == 0.
+                if (capacity > 0) {
+                    data.capacity(capacity);
+                    int rc = readFromLogChannel(entryLogId, bc, data, pos);
+                    if (rc != entrySize) {
+                        LOG.warn("Short read for ledger entry from entryLog {}@{} ({} != {})",
+                                entryLogId, pos, rc, entrySize);
+                        return;
+                    }
                 }
+
                 // process the entry
-                scanner.process(ledgerId, offset, data);
+                scanner.process(ledgerId, offset, data, entrySize);
 
                 // Advance position to the next entry
                 pos += entrySize;
@@ -1164,17 +1169,23 @@ public class DefaultEntryLogger implements EntryLogger {
         // Read through the entry log file and extract the entry log meta
         scanEntryLog(entryLogId, new EntryLogScanner() {
             @Override
-            public void process(long ledgerId, long offset, ByteBuf entry) throws IOException {
+            public void process(long ledgerId, long offset, ByteBuf entry, int entrySize) throws IOException {
                 if (throttler != null) {
                     throttler.acquire(entry.readableBytes());
                 }
                 // add new entry size of a ledger to entry log meta
-                meta.addLedgerSize(ledgerId, entry.readableBytes() + 4);
+                meta.addLedgerSize(ledgerId, entrySize + 4);
             }
 
             @Override
             public boolean accept(long ledgerId) {
                 return ledgerId >= 0;
+            }
+
+            @Override
+            public int getLengthToRead(){
+                // we only need to read the entry size.
+                return EntryLogScanner.READ_NOTHING;
             }
         });
 
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/EntryLogScanner.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/EntryLogScanner.java
@@ -27,6 +27,11 @@ import java.io.IOException;
  * Scan entries in a entry log file.
  */
 public interface EntryLogScanner {
+    public static final int READ_ALL = Integer.MAX_VALUE;
+    public static final int READ_NOTHING = 0;
+    public static final int READ_ENTRY_ID = 8;
+    public static final int READ_LEDGER_ENTRY_ID = 16;
+
     /**
      * Tests whether or not the entries belongs to the specified ledger
      * should be processed.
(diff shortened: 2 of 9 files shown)


thetumbled:
Could you help to reviw this PR? thanks!
@eolivelli @dlg99 @hangc0276 @nicoloboschi @shoothzj @zymap @wenbingshen

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/EntryLogScanner.java
```diff
@@ -27,6 +27,10 @@
  * Scan entries in a entry log file.
  */
 public interface EntryLogScanner {
+    public static final int READ_ALL = Integer.MAX_VALUE;
+    public static final int READ_NOTHING = 0;
+    public static final int READ_LEDGER_ENTRY_ID = 16;
```
eolivelli, line 32: what about `READ_LEDGER_ENTRY_ID_LENGTH` ?

but as we basically expect only these 3 values we can use an "enum" and ensure that the code is not called with other combinations (like negative numbers or non sense numbers caused by bugs)

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/DefaultEntryLogger.java
```diff
@@ -1036,15 +1036,20 @@ public void scanEntryLog(long entryLogId, EntryLogScanner scanner) throws IOExce
                 }
                 // read the entry
                 data.clear();
-                data.capacity(entrySize);
-                int rc = readFromLogChannel(entryLogId, bc, data, pos);
-                if (rc != entrySize) {
-                    LOG.warn("Short read for ledger entry from entryLog {}@{} ({} != {})",
-                            entryLogId, pos, rc, entrySize);
-                    return;
+                int capacity = Math.min(scanner.getLengthToRead(), entrySize);
```
eolivelli, line 1039: if getLengthToRead returns READ_NOTHING then maybe "data" can be passed as "null"
if getLengthToRead returns READ_LEDGER_ENTRY_ID then the caller is interested only in the initial bytes.

Please consider changing the interface and have 3 different "process" methods

READ_NOTHING:
process(ledgerId, entrySize)

READ_LEDGER_ENTRY_ID;
process(ledgerId, entryId, entrySize)
in this case the entryId is extracted by this code and not by the code downstream

READ_ALL
process(ledgerId, offset, data)

This way it is clearer what is happening, and the code is safer, as the implementation of the interface doesn't need to understand the internal format of the "data" buffer when it doesn't need it.
Using the enum (as suggested below) does not introduce extra costs at runtime

eolivelli (changes requested):

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/EntryLogScanner.java
thetumbled: good idea, i have introduced the enum.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/DefaultEntryLogger.java
thetumbled: Good idea, i have refactored the code, please help to review it again, thanks.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/DefaultEntryLogger.java
```diff
@@ -1036,16 +1037,33 @@ public void scanEntryLog(long entryLogId, EntryLogScanner scanner) throws IOExce
...
+                        break;
+                    case READ_ALL:
+                        data.capacity(entrySize);
+                        rc = readFromLogChannel(entryLogId, bc, data, pos);
+                        if (rc != entrySize) {
+                            LOG.warn("Short read for ledger entry id from entrylog {}", entryLogId);
+                            return;
+                        }
+                        scanner.process(ledgerId, offset, data);
+                        break;
```
eolivelli, line 1065: please add the "default" case

eolivelli (changes requested):
very good, I left one final comment about the "switch" statement

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/DefaultEntryLogger.java
thetumbled: Done, PTAL.

hangc0276 (changes requested):
I'm unsure of the extent of the improvements that can be achieved with this change, given the following concerns:

- All the changes in this PR are specific to uncommon cases, occurring only when the entrylog file's index is missing.
- The readFromLogChannel function utilizes BufferedLogChannel, which is a RandomAccessFile channel. When reading data from the file channel, the data is pre-fetched into the OS PageCache, with a default pre-fetch size of 4KB. Even though we only retrieve the entryId, the actual data read from the disk is a minimum of 4KB.

There is a potential risk associated with this PR:

- We solely read the entryID without validating the entry data. If the file is corrupted, the scan operation will be unable to detect it and will incorrectly populate the index with the wrong ledger size. This introduces a high level of risk.

thetumbled:
> All the changes in this PR are specific to uncommon cases, occurring only when the entrylog file's index is missing.

We have encountered cases that the ledger map in entry log is missed, maybe because the bookie crashed before flushing entry log. As long as the corrupted entry exists, bookie will scan the entry log by `extractEntryLogMetadataByScanning` to generate `EntryLogMetadataMap` when doing gc every `gcWaitTime` milliseconds (default 15min). 
And other cases is related to index rebuilding, which may be rarely used.

> The readFromLogChannel function utilizes BufferedLogChannel, which is a RandomAccessFile channel. When reading data from the file channel, the data is pre-fetched into the OS PageCache, with a default pre-fetch size of 4KB. Even though we only retrieve the entryId, the actual data read from the disk is a minimum of 4KB.

It is common that the size of one entry reach 4MB (  we set the max batch size of Pulsar Client to be 4MB), so we can decrease 90% of the disk read with this enhancement.

> There is a potential risk associated with this PR:
> We solely read the entryID without validating the entry data. If the file is corrupted, the scan operation will be unable to detect it and will incorrectly populate the index with the wrong ledger size. This introduces a high level of risk.

The fault detecting logic is not changed, in the old logic we just read `entrySize` amount of data and check if we have read such amount of data. In the new read type, `READ_NOTHING` or `READ_LEDGER_ENTRY_ID`, we still move the read position in the loop and check if we have read expected amount of data. I don't think it is a break change.

PR #4163, opened 2023-12-24
Fix the autorecovery failed replicate by add entry fenced error

zymap:
---

### Motivation

#3783 changed the recovery add request flags. It causes when replication happening, if the ledger was fenced, the ledger can not be recovered. Because the recovery add flag is not added then the normal add is not allowed to a fenced ledger.


hangc0276:
@zymap Please fix the check style, thanks.

hangc0276 (approved):
Nice Catch!

Issue #4143, opened 2023-12-04
[improve][stats] Enhance bookie metrics to support URL parameter filtering


zwOvO:
### Feature Request for Bookkeeper Issue

**Feature Request:**
I would like to request an improvement to the bookie metrics interface, allowing it to filter metrics by URL parameters. This is important to me as we are currently facing an issue with an excessive amount of temporarily unused metrics being collected by our alerting platform. These metrics are useful for monitoring purposes, but not for the alerting platform.

**Proposed Improvement:**
Allow filtering of metrics using URL parameters, for example:
   - `/metrics?m_regex=bookkeeper_server_thread_executor_queue`
   - `/metrics?m_regex=.*_thread_executor_queue`

**Are you willing to submit a PR?**
 I'm willing to submit a PR!

miton18:
Can't you filtre metrics on the collector side ?

zwOvO:
> I would like to request an improvement to the bookie metrics interface, allowing it to filter metrics by URL parameters. This is important to me as we are currently facing an issue with an excessive amount of temporarily unused metrics being collected by our alerting platform. These metrics are useful for monitoring purposes, but not for the alerting platform.

Our alarm platform only needs to configure a Prometheus export port to complete indicator collection, and the large number of indicators currently exposed by Bookie do not need to be collected and saved in the alarm platform.

PR #4164, opened 2023-12-27
[improve][stats] Enhance bookie metrics to support URL parameter filtering

zwOvO:
### Motivation
Allow filtering of metrics using URL parameters, for example:

/metrics?m_regex=bookkeeper_server_thread_executor_queue
/metrics?m_regex=.*_thread_executor_queue

Master Issue: #4143

### DIFF
--- a/stats/bookkeeper-stats-api/src/main/java/org/apache/bookkeeper/stats/StatsProvider.java
+++ b/stats/bookkeeper-stats-api/src/main/java/org/apache/bookkeeper/stats/StatsProvider.java
@@ -48,6 +48,16 @@ public interface StatsProvider {
     }
 
     /**
+     *
+     * @param writer
+     * @param filterRegexStr
+     * @throws IOException
+     */
+    default void writeAllMetrics(Writer writer, String filterRegexStr) throws IOException {
+        throw new UnsupportedOperationException("writeAllMetrics is not implemented yet");
+    }
+
+    /**
      * Return the stats logger to a given <i>scope</i>.
      * @param scope
      *          Scope for the given stats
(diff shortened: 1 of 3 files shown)


hangc0276:
@asafm Would you please help take a look at this improvement? Thanks.

asafm:
@hangc0276 `/metrics` should be Prometheus compatible. 
If you look at the Prometheus Java client, setting up a server, exposing `/metrics` you'll see it only supports specifying the metric names, using the query param `name[]`.

```
    protected static Set<String> parseQuery(String query) throws IOException {
        Set<String> names = new HashSet<String>();
        if (query != null) {
            String[] pairs = query.split("&");
            for (String pair : pairs) {
                int idx = pair.indexOf("=");
                if (idx != -1 && URLDecoder.decode(pair.substring(0, idx), "UTF-8").equals("name[]")) {
                    names.add(URLDecoder.decode(pair.substring(idx + 1), "UTF-8"));
                }
            }
        }
        return names;
    }
```

Besides that, I presume the idea is that BookKeeper users will transition into OpenTelemetry which is already available. For now, even OpenTelemetry Java SDK doesn't support filtering YET.
I added a [proposal](https://github.com/open-telemetry/opentelemetry-specification/pull/3566) to OTel specifications and it was approved, and hopefully in the following months it will be implemented. 

For now I recommend not adding it, and using OTel collector to scrape `/metrics` and filter out any unwanted metrics.

zwOvO:
OK, Now I am filtering JVM metrics by writing a Python server proxy
——————
```python
#!/bin/python

import SimpleHTTPServer
import SocketServer
import subprocess
import requests

class CustomHandler(SimpleHTTPServer.SimpleHTTPRequestHandler):
    def do_GET(self):
        if self.path == '/metrics':
            
            response = requests.get('http://localhost:8080/metrics/')
            output = '\n'.join(line for line in response.text.split('\n') if 'jvm' in line)
            self.send_response(200)
            self.send_header('Content-type', 'text/plain')
            self.end_headers()
            self.wfile.write(output)

PORT = 9001
httpd = SocketServer.TCPServer(("", PORT), CustomHandler)
print "Server running on port", PORT
httpd.serve_forever()
```

asafm:
> OK, Now I am filtering JVM metrics by writing a Python server proxy ——————
> 
> ```python
> #!/bin/python
> 
> import SimpleHTTPServer
> import SocketServer
> import subprocess
> import requests
> 
> class CustomHandler(SimpleHTTPServer.SimpleHTTPRequestHandler):
>     def do_GET(self):
>         if self.path == '/metrics':
>             
>             response = requests.get('http://localhost:8080/metrics/')
>             output = '\n'.join(line for line in response.text.split('\n') if 'jvm' in line)
>             self.send_response(200)
>             self.send_header('Content-type', 'text/plain')
>             self.end_headers()
>             self.wfile.write(output)
> 
> PORT = 9001
> httpd = SocketServer.TCPServer(("", PORT), CustomHandler)
> print "Server running on port", PORT
> httpd.serve_forever()
> ```

You can just use OTel Collector. It has built-in functionality to do just that.

PR #4165, opened 2023-12-27
Fix auditor elector executor block problem.

horizonzy:
### Motivation
Now, when we shut down the auditor elector. The shutdown behavior via submitting a shutdown task to do it. 
In some cases, the follower auditor elector executor is always blocked due to waiting leader election, so the shutdown task will lie in the task queue forever, get no chance to execute.

In the pulsar, the case happen. See https://github.com/apache/pulsar/pull/21797#discussion_r1436278822

So in the auditor elector shutdown, if the remain task can't done in time, we should invoke shutdownNow to interrupt the blocked thread.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/AuditorElector.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/AuditorElector.java
@@ -151,24 +151,26 @@ public class AuditorElector {
      * Run cleanup operations for the auditor elector.
      */
     private Future<?> submitShutdownTask() {
-        return executor.submit(new Runnable() {
-                @Override
-                public void run() {
-                    if (!running.compareAndSet(true, false)) {
-                        return;
-                    }
-
-                    try {
-                        ledgerAuditorManager.close();
-                    } catch (InterruptedException ie) {
-                        Thread.currentThread().interrupt();
-                        LOG.warn("InterruptedException while closing ledger auditor manager", ie);
-                    } catch (Exception ke) {
-                        LOG.error("Exception while closing ledger auditor manager", ke);
-                    }
-                }
-            });
+        return executor.submit(shutdownTask);
     }
+
+    Runnable shutdownTask = new Runnable() {
+        @Override
+        public void run() {
+            if (!running.compareAndSet(true, false)) {
+                return;
+            }
+
+            try {
+                ledgerAuditorManager.close();
+            } catch (InterruptedException ie) {
+                Thread.currentThread().interrupt();
+                LOG.warn("InterruptedException while closing ledger auditor manager", ie);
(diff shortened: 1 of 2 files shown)


hangc0276:
@horizonzy The failed tests looks like related to this change, please help take a look, thanks.

horizonzy:
> @horizonzy The failed tests looks like related to this change, please help take a look, thanks.

Fixed.

lhotari (changes requested):
I think it's better to make submitShutdownTask return `Future<?>` and wait for that to complete before shutting down the executor using `MoreExecutors.shutdownAndAwaitTermination`. Something like this: https://github.com/lhotari/bookkeeper/commit/670c992607d38172b45193228b8bb82eb851ef11 . 
If the "submitShutdownTask" doesn't run, there could be a race condition and the running status isn't changed to false. That will then shutdown the other autorecovery threads in https://github.com/apache/bookkeeper/blob/405e72acf42bb1104296447ea8840d805094c787/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/AutoRecoveryMain.java#L234 .

horizonzy:
> I think it's better to make submitShutdownTask return `Future<?>` and wait for that to complete before shutting down the executor using `MoreExecutors.shutdownAndAwaitTermination`. Something like this: [lhotari@670c992](https://github.com/lhotari/bookkeeper/commit/670c992607d38172b45193228b8bb82eb851ef11) . If the "submitShutdownTask" doesn't run, there could be a race condition and the running status isn't changed to false. That will then shutdown the other autorecovery threads in
> 
> https://github.com/apache/bookkeeper/blob/405e72acf42bb1104296447ea8840d805094c787/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/AutoRecoveryMain.java#L234
> 
> .

Good idea. That makes sense.

lhotari (approved):
LGTM

bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/AuditorElector.java
```diff
@@ -238,8 +241,16 @@ public void shutdown() throws InterruptedException {
...
-            executor.shutdown();
+            try {
+                submitShutdownTask().get(10, TimeUnit.SECONDS);
+                executor.shutdown();
+            } catch (ExecutionException e) {
+                LOG.warn("Failed to close auditor manager", e);
+                executor.shutdownNow();
+            } catch (TimeoutException e) {
+                LOG.warn("Failed to close auditor manager in 10 seconds", e);
+                executor.shutdownNow();
```
lhotari, line 252: I wonder if it would be necessary to set running to false in these exceptional cases so that the other thread that checks that value would exit?

hangc0276 (approved):

PR #4175, opened 2024-01-11
Fix race condition in LedgerHandle in drainPendingAddsAndAdjustLength & sendAddSuccessCallbacks

lhotari:
### Motivation

There's a race condition in LedgerHandle between drainPendingAddsAndAdjustLength & sendAddSuccessCallbacks method calls. The methods could get called simultaneously which would cause inconsistency. 
The changingEnsemble field has a thread safety issue and the changes might not be visible to the reading thread.

### Changes

- use `synchronized (pendingAddOps)` as the object monitor for serializing drainPendingAddsAndAdjustLength & sendAddSuccessCallbacks method calls.
- make changingEnsemble field volatile
- remove the too strict rule for pendingAddsSequenceHead which breaks things after failures

This rule is removed:
https://github.com/apache/bookkeeper/blob/13e7efaa971cd3613b065ac50836c5ee98985d13/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java#L1822-L1829
The current assumption is that this doesn't hold under failure conditions.

### Additional Context

- this is an alternative fix for https://github.com/apache/pulsar/issues/21860 
- the other fix is #4171

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerHandle.java
@@ -1827,14 +1827,6 @@ public class LedgerHandle implements WriteHandle {
                     }
                     return;
                 }
-                // Check if it is the next entry in the sequence.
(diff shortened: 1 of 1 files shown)


lhotari:
Closing in favor of the original PR #4171.

PR #4176, opened 2024-01-12
Support read timeout on BK server

315157973:
### Motivation
When client read request has timed out, the read request in Bookie's thread pool queue cannot time out, and will continue to read disk, causing unnecessary IO.

### Changes
Add read request timeout, and read requests that have timed out returned directly.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
@@ -229,6 +229,7 @@ public class ServerConfiguration extends AbstractConfiguration<ServerConfigurati
     protected static final String NUM_LONG_POLL_WORKER_THREADS = "numLongPollWorkerThreads";
     protected static final String NUM_HIGH_PRIORITY_WORKER_THREADS = "numHighPriorityWorkerThreads";
     protected static final String READ_WORKER_THREADS_THROTTLING_ENABLED = "readWorkerThreadsThrottlingEnabled";
+    protected static final String READ_ENTRY_PENDING_TIMEOUT_MILLIS = "readEntryPendingTimeoutMillis";
 
     // Long poll parameters
     protected static final String REQUEST_TIMER_TICK_DURATION_MILLISEC = "requestTimerTickDurationMs";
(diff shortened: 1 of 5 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadEntryProcessor.java
```diff
@@ -71,6 +73,11 @@ protected void processPacket() {
         int errorCode = BookieProtocol.EOK;
         long startTimeNanos = MathUtils.nowInNano();
         ByteBuf data = null;
+        if (readTimeoutNanos >= 0 && startTimeNanos - enqueueNanos > readTimeoutNanos) {
```
AnonHxy, line 76:
```suggestion
        if (readTimeoutNanos > 0 && startTimeNanos - enqueueNanos > readTimeoutNanos) {
```
I thinks <=0 for disabled is better:)

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
```diff
@@ -2030,7 +2031,23 @@ public boolean isReadWorkerThreadsThrottlingEnabled() {
...
+    public ServerConfiguration setReadEntryPendingTimeoutMillis(long timeoutMillis) {
+        setProperty(READ_ENTRY_PENDING_TIMEOUT_MILLIS, timeoutMillis);
+        return this;
+    }
 
+    /**
+     * Get timout millis for pending read request.
+     * @return timout millis
+     */
+    public long getReadEntryPendingTimeoutMillis() {
```
hangc0276, line 2048: Why set the timeout to 5 seconds by default? Pulsar sets the timeout to 30s https://github.com/apache/pulsar/blob/3fa85f038aac0bb416d6821a61709adfdc1a1c49/conf/broker.conf#L955

Adding the timeout on the server side will save IOs to disk, but it will make the configuration more complex because the user needs to care about the configuration on both the server and client side.

What's more, if Pulsar different namespace configures different bk client read timeout configurations, it will be hard for the server to match the readTimeout with the client.

hangc0276 (changes requested):

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
315157973: The server should set a maximum timeout, and we cannot rely on the time set by the client.
Like Zookeeper, the server can also set a maximum Session timeout.

PR #4177, opened 2024-01-14
[BP-62] Batch read protocol change: Support encoding and decoding for batchReadRequest and batchReadResponse

horizonzy:
### Motivation
This is the first PR for the batch read(#4051) feature. 

Protocol change: Support encoding and decoding for batchReadRequest and batchReadResponse

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtoEncoding.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtoEncoding.java
@@ -110,7 +110,30 @@ public class BookieProtoEncoding {
                 return msg;
             }
             BookieProtocol.Request r = (BookieProtocol.Request) msg;
-            if (r instanceof BookieProtocol.ReadRequest) {
+            if (r instanceof BookieProtocol.BatchedReadRequest) {
+                int totalHeaderSize = 4 // for request type
+                        + 8 // for ledger id
+                        + 8 // for entry id
+                        + 8 // for request id
+                        + 4 // for max count
+                        + 8; // for max size
+                if (r.hasMasterKey()) {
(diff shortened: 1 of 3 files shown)


hangc0276:
@horizonzy Please fix the failed check style, thanks.

hangc0276:
@lhotari Please help take a look at this PR, thanks.

lhotari (approved):
LGTM

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtoEncoding.java
```diff
@@ -181,6 +204,21 @@ public Object decode(ByteBuf packet)
                 } else {
                     return BookieProtocol.ReadRequest.create(version, ledgerId, entryId, flags, null);
                 }
+            case BookieProtocol.BATCH_READ_ENTRY:
+                ledgerId = packet.readLong();
+                entryId = packet.readLong();
+                long requestId = packet.readLong();
+                int maxCount = packet.readInt();
+                long maxSize = packet.readLong();
+                if ((flags & BookieProtocol.FLAG_DO_FENCING) == BookieProtocol.FLAG_DO_FENCING
```
zymap, line 213: FENCING is a special reading that will fence the ledger, and no more write allowed; will it allow in the batch read?

hangc0276 (approved):

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtoEncoding.java
eolivelli: If we move the recovery to use batch reads then this may be useful

eolivelli (approved):
Lgtm

Issue #4185, opened 2024-01-20
[improve][docker] fix yaml and dockerfile


zwOvO:
**FEATURE REQUEST**
![image](https://github.com/apache/bookkeeper/assets/18533252/ef8194e6-7815-4dff-aa91-24300fbac6a5)

![image](https://github.com/apache/bookkeeper/assets/18533252/daff156f-f03f-49e3-b4d7-c7058f0e858b)

lhotari:
This issue can be closed since the PR has been merged. /cc @hangc0276

PR #4186, opened 2024-01-20
[improve][docker] fix yaml and dockerfile

zwOvO:
1. There are multiple spec.ports fields in sts yaml
2. The RUN command specified in dockerfile is not standardized

Master Issue: #4185


shoothzj (approved):

hangc0276 (approved):

lhotari:
More Dockerfile improvements in this PR: #4197 . For example, the base image `ubuntu:22.10` is no longer usable since updates aren't available. That is fixed by switching to `ubuntu:22.04` which is a LTS version.

PR #2701, opened 2021-04-23
Fixed unnecessary copy to heap, see https://github.com/apache/pulsar/pull/10330

dlg99:
Descriptions of the changes in this PR:

Handling CompositeByteBuf in a way that avoids unnecessary data copy.

### Motivation

https://github.com/apache/pulsar/pull/10330

https://github.com/apache/pulsar/pull/10330#issuecomment-825619753

### Changes

Handling CompositeByteBuf in a way that avoids unnecessary data copy.

> ---
> In order to uphold a high standard for quality for code contributions, Apache BookKeeper runs various precommit
> checks for pull requests. A pull request can only be merged when it passes precommit checks.
>
> ---
> Be sure to do all of the following to help us incorporate your contribution
> quickly and easily:
>
> If this PR is a BookKeeper Proposal (BP):
>
> - [ ] Make sure the PR title is formatted like:
>     `<BP-#>: Description of bookkeeper proposal`
>     `e.g. BP-1: 64 bits ledger is support`
> - [ ] Attach the master issue link in the description of this PR.
> - [ ] Attach the google doc link if the BP is written in Google Doc.
>
> Otherwise:
> 
> - [ ] Make sure the PR title is formatted like:
>     `<Issue #>: Description of pull request`
>     `e.g. Issue 123: Description ...`
> - [ ] Make sure tests pass via `mvn clean apache-rat:check install spotbugs:check`.
> - [ ] Replace `<Issue #>` in the title with the actual Issue number.
> 
> ---

PR #4196, opened 2024-01-31
Fix checksum calculation bug when the payload is a CompositeByteBuf with readerIndex > 0

lhotari:
### Motivation

There's a checksum calculation bug when the payload is a CompositeByteBuf with readerIndex > 0.
This bug is demonstrated in the included repro test case in commit 9fb47f5dcf3848320991c7a2cf602eb325893a42.

The problematic solution to unwrap CompositeByteBuf was introduced in PRs #2701. 
PR #2701 was added to resolve a problem described in this Apache Pulsar PR: https://github.com/apache/pulsar/pull/10330 . _"... BookKeeper's checksum logic, the nioBuffer() method will be called and the CompositeByteBuf's internal buffers will be concatenated into a single buffer in heap memory..."_

### Changes

Revert the unsafe ways to "unwrap" Netty ByteBuf instances and access CompositeByteBuf directly without taking the readerIndex state into account. This approach was introduced in PRs #2701.

Netty doesn't have a public API for visiting the sub buffers, but there is a way to do this using the "ByteBuf.getBytes" method. That is the approach used in this PR.

The reason to access the underlying buffers is to ensure that no extra copies will have to be made when calculating the checksum. This PR doesn't cause performance regressions, but improves performance since there were code paths in DigestManager where the previous buggy solution for checksum calculation optimization wasn't used.

### DIFF
--- a/.github/workflows/bk-ci.yml
+++ b/.github/workflows/bk-ci.yml
@@ -200,6 +200,10 @@ jobs:
           path: surefire-reports
           retention-days: 7
 
+      - name: print JVM thread dumps when cancelled
+        if: cancelled()
+        run: ./dev/ci-tool print_thread_dumps
+
   integration-tests:
     name: Integration Tests
     runs-on: ubuntu-latest
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/DigestManager.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/DigestManager.java
@@ -20,7 +20,6 @@ package org.apache.bookkeeper.proto.checksum;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufAllocator;
 import io.netty.buffer.ByteBufUtil;
-import io.netty.buffer.DuplicatedByteBuf;
 import io.netty.buffer.PooledByteBufAllocator;
 import io.netty.buffer.Unpooled;
 import io.netty.util.ReferenceCounted;
@@ -68,7 +67,7 @@ public abstract class DigestManager {
             return digest;
         }
         if (depth < MAX_SUB_BUFFER_VISIT_RECURSION_DEPTH && !buffer.hasMemoryAddress() && !buffer.hasArray()) {
-            return visitWrappedBuffersAndCallInternalUpdateForEach(digest, buffer, offset, len, depth);
+            return visitChildBuffersAndCallInternalUpdateForEach(digest, buffer, offset, len, depth);
         }
         return internalUpdate(digest, buffer, offset, len);
     }
@@ -87,46 +86,16 @@ public abstract class DigestManager {
      * @param depth the recursion depth of the wrapped buffer visit
      * @return updated digest value
      */
-    private int visitWrappedBuffersAndCallInternalUpdateForEach(int digest, ByteBuf buffer, int offset, int len,
-                                                                int depth) {
+    private int visitChildBuffersAndCallInternalUpdateForEach(int digest, ByteBuf buffer, int offset, int len,
+                                                              int depth) {
         // hold the digest in a MutableInt so that it can be updated in the wrapped buffer visit
         MutableInt digestRef = new MutableInt(digest);
-        ByteBuf sliced = buffer.slice(offset, len);
-        // call getBytes to trigger the wrapped buffer visit
-        sliced.getBytes(0, new DuplicatedByteBuf(Unpooled.EMPTY_BUFFER) {
-            @Override
-            public ByteBuf setBytes(int index, ByteBuf src, int srcIndex, int length) {
-                if (length > 0) {
-                    // recursively visit the sub buffer and update the digest
-                    int updatedDigest = recursiveSubBufferVisitForDigestUpdate(digestRef.intValue(), src, srcIndex,
-                            length, depth + 1);
-                    digestRef.setValue(updatedDigest);
-                }
-                return this;
-            }
-
-            @Override
-            public boolean hasArray() {
-                // return false so that the wrapped buffer is visited
-                return false;
-            }
-
-            @Override
-            public boolean hasMemoryAddress() {
-                // return false so that the wrapped buffer is visited
-                return false;
-            }
-
-            @Override
-            public int nioBufferCount() {
-                // return 0 so that the wrapped buffer is visited
-                return 0;
-            }
-
-            @Override
-            public int capacity() {
-                // should return sufficient capacity for the total length
-                return len;
+        ChildByteBufVisitor.visitChildBuffers(buffer, offset, len, (ByteBuf childBuffer, int srcIndex, int length) -> {
+            if (length > 0) {
+                // recursively visit the sub buffer and update the digest
+                int updatedDigest =
+                        recursiveSubBufferVisitForDigestUpdate(digestRef.intValue(), childBuffer, srcIndex, length, depth + 1);
+                digestRef.setValue(updatedDigest);
             }
         });
         // return updated digest value
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/DigestManager.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/DigestManager.java
@@ -61,7 +61,7 @@ public abstract class DigestManager {
     abstract int internalUpdate(int digest, byte[] buffer, int offset, int len);
 
     final int update(int digest, ByteBuf buffer, int offset, int len) {
-        if (buffer.hasMemoryAddress()) {
+        if (buffer.hasMemoryAddress() && acceptsMemoryAddressBuffer()) {
             return internalUpdate(digest, buffer, offset, len);
         } else if (buffer.hasArray()) {
             return internalUpdate(digest, buffer.array(), buffer.arrayOffset() + offset, len);
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/DigestManager.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/DigestManager.java
@@ -61,18 +61,7 @@ public abstract class DigestManager {
 
     abstract int internalUpdate(int digest, byte[] buffer, int offset, int len);
 
-    protected int initialDigest() {
-        return 0;
-    };
-
-    protected int finalizeDigest(int digest) {
-        return digest;
-    }
-
     final int update(int digest, ByteBuf buffer, int offset, int len) {
-        if (len == 0) {
-            return digest;
-        }
         if (buffer.hasMemoryAddress()) {
             return internalUpdate(digest, buffer, offset, len);
         } else if (buffer.hasArray()) {
@@ -82,14 +71,6 @@ public abstract class DigestManager {
             ByteBufVisitor.visitBuffers(buffer, offset, len, byteBufVisitorCallback, digestRef);
             return digestRef.intValue();
         }
-    }
-
-    protected final int initializeAndUpdate(ByteBuf buf, int offset, int len) {
-        return update(initialDigest(), buf, offset, len);
-    }
-
-    private int updateAndFinalize(ByteBuf data, int digest, int offset, int len) {
-        return finalizeDigest(update(digest, data, offset, len));
     }
 
     abstract void populateValueAndReset(int digest, ByteBuf buffer);
@@ -170,8 +151,8 @@ public abstract class DigestManager {
         buf.writeLong(length);
 
         // Compute checksum over the headers
-        int digest = initializeAndUpdate(buf, buf.readerIndex(), buf.readableBytes());
-        digest = updateAndFinalize(data, digest, data.readerIndex(), data.readableBytes());
+        int digest = update(0, buf, buf.readerIndex(), buf.readableBytes());
+        digest = update(digest, data, data.readerIndex(), data.readableBytes());
 
         populateValueAndReset(digest, buf);
 
@@ -195,8 +176,8 @@ public abstract class DigestManager {
         headersBuffer.writeLong(lastAddConfirmed);
         headersBuffer.writeLong(length);
 
-        int digest = initializeAndUpdate(headersBuffer, 0, METADATA_LENGTH);
-        digest = updateAndFinalize(data, digest, data.readerIndex(), data.readableBytes());
+        int digest = update(0, headersBuffer, 0, METADATA_LENGTH);
+        digest = update(digest, data, data.readerIndex(), data.readableBytes());
         populateValueAndReset(digest, headersBuffer);
         return ByteBufList.get(headersBuffer, data);
     }
@@ -218,7 +199,7 @@ public abstract class DigestManager {
         headersBuffer.writeLong(ledgerId);
         headersBuffer.writeLong(lac);
 
-        int digest = finalizeDigest(initializeAndUpdate(headersBuffer, 0, LAC_METADATA_LENGTH));
+        int digest = update(0, headersBuffer, 0, LAC_METADATA_LENGTH);
         populateValueAndReset(digest, headersBuffer);
 
         return ByteBufList.get(headersBuffer);
@@ -254,10 +235,10 @@ public abstract class DigestManager {
                     this.getClass().getName(), dataReceived.readableBytes());
             throw new BKDigestMatchException();
         }
-        int digest = initializeAndUpdate(dataReceived, 0, METADATA_LENGTH);
+        int digest = update(0, dataReceived, 0, METADATA_LENGTH);
 
         int offset = METADATA_LENGTH + macCodeLength;
-        digest = updateAndFinalize(dataReceived, digest, offset, dataReceived.readableBytes() - offset);
+        digest = update(digest, dataReceived, offset, dataReceived.readableBytes() - offset);
 
         if (isInt32Digest()) {
             int receivedDigest = dataReceived.getInt(METADATA_LENGTH);
@@ -302,7 +283,7 @@ public abstract class DigestManager {
             throw new BKDigestMatchException();
         }
 
-        int digest = finalizeDigest(initializeAndUpdate(dataReceived, 0, LAC_METADATA_LENGTH));
+        int digest = update(0, dataReceived, 0, LAC_METADATA_LENGTH);
 
         if (isInt32Digest()) {
             int receivedDigest = dataReceived.getInt(LAC_METADATA_LENGTH);
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/DigestManager.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/DigestManager.java
@@ -370,18 +370,14 @@ public abstract class DigestManager {
 
         @Override
         public void visitBuffer(UpdateContext context, ByteBuf visitBuffer, int visitIndex, int visitLength) {
-            if (visitLength > 0) {
-                // recursively visit the sub buffer and update the digest
-                context.digest = internalUpdate(context.digest, visitBuffer, visitIndex, visitLength);
-            }
+            // recursively visit the sub buffer and update the digest
+            context.digest = internalUpdate(context.digest, visitBuffer, visitIndex, visitLength);
         }
 
         @Override
         public void visitArray(UpdateContext context, byte[] visitArray, int visitIndex, int visitLength) {
-            if (visitLength > 0) {
-                // update the digest with the array
-                context.digest = internalUpdate(context.digest, visitArray, visitIndex, visitLength);
-            }
+            // update the digest with the array
+            context.digest = internalUpdate(context.digest, visitArray, visitIndex, visitLength);
         }
 
         @Override
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/DigestManager.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/DigestManager.java
@@ -57,14 +57,30 @@ public abstract class DigestManager {
 
     abstract int internalUpdate(int digest, ByteBuf buffer, int offset, int len);
 
+    abstract int internalUpdate(int digest, byte[] buffer, int offset, int len);
+
     final int update(int digest, ByteBuf buffer, int offset, int len) {
         MutableInt digestRef = new MutableInt(digest);
         ByteBufVisitor.visitBuffers(buffer, offset, len,
-                (ByteBuf childBuffer, int childIndex, int childLength) -> {
-                    if (childLength > 0) {
-                        // recursively visit the sub buffer and update the digest
-                        int updatedDigest = internalUpdate(digestRef.intValue(), childBuffer, childIndex, childLength);
-                        digestRef.setValue(updatedDigest);
+                new ByteBufVisitor.ByteBufVisitorCallback() {
+                    @Override
+                    public void visitBuffer(ByteBuf visitBuffer, int visitIndex, int visitLength) {
+                        if (visitLength > 0) {
+                            // recursively visit the sub buffer and update the digest
+                            int updatedDigest =
+                                    internalUpdate(digestRef.intValue(), visitBuffer, visitIndex, visitLength);
+                            digestRef.setValue(updatedDigest);
+                        }
+                    }
+
+                    @Override
+                    public void visitArray(byte[] visitArray, int visitIndex, int visitLength) {
+                        if (visitLength > 0) {
+                            // update the digest with the array
+                            int updatedDigest =
+                                    internalUpdate(digestRef.intValue(), visitArray, visitIndex, visitLength);
+                            digestRef.setValue(updatedDigest);
+                        }
                     }
                 });
         return digestRef.intValue();
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/DigestManager.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/DigestManager.java
@@ -73,9 +73,15 @@ public abstract class DigestManager {
         if (len == 0) {
             return digest;
         }
-        MutableInt digestRef = new MutableInt(digest);
-        ByteBufVisitor.visitBuffers(buffer, offset, len, byteBufVisitorCallback, digestRef);
-        return digestRef.intValue();
+        if (buffer.hasMemoryAddress()) {
+            return internalUpdate(digest, buffer, offset, len);
+        } else if (buffer.hasArray()) {
+            return internalUpdate(digest, buffer.array(), offset, len);
+        } else {
+            MutableInt digestRef = new MutableInt(digest);
+            ByteBufVisitor.visitBuffers(buffer, offset, len, byteBufVisitorCallback, digestRef);
+            return digestRef.intValue();
+        }
     }
 
     protected final int initializeAndUpdate(ByteBuf buf, int offset, int len) {
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/DigestManager.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/DigestManager.java
@@ -20,6 +20,7 @@ package org.apache.bookkeeper.proto.checksum;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufAllocator;
 import io.netty.buffer.ByteBufUtil;
+import io.netty.buffer.DuplicatedByteBuf;
 import io.netty.buffer.PooledByteBufAllocator;
 import io.netty.buffer.Unpooled;
 import io.netty.util.ReferenceCounted;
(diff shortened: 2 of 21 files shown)


lhotari:
@hangc0276 @poorbarcode @BewareMyPower Please review

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/DigestManager.java
```diff
@@ -49,14 +49,89 @@ public abstract class DigestManager {
...
+     * @param depth the recursion depth of the wrapped buffer visit
+     * @return updated digest value
+     */
+    private int visitWrappedBuffersAndCallInternalUpdateForEach(int digest, ByteBuf buffer, int offset, int len,
+                                                                int depth) {
+        // hold the digest in a MutableInt so that it can be updated in the wrapped buffer visit
+        MutableInt digestRef = new MutableInt(digest);
+        ByteBuf sliced = buffer.slice(offset, len);
+        // call getBytes to trigger the wrapped buffer visit
+        sliced.getBytes(0, new DuplicatedByteBuf(Unpooled.EMPTY_BUFFER) {
```
BewareMyPower, line 96: `DuplicatedByteBuf` is deprecated, so it could be removed in a future release of Netty.

This implementation is tricky and hard to understand without diving into the internal. After debugging step by step, I found the correctness depends on the implementation details in `UnsafeByteBufUtil`.

image

Is it possible to achieve the same goal with the following logic? 

image

lhotari: > `DuplicatedByteBuf` is deprecated, so it could be removed in a future release of Netty.

That's true. However if that would happen, the usage could be replaced. I could also replace that in this PR in further commits. I was hoping to explain the solution for achieving checksum calculations without extra copies by minimizing the changes in this PR in the first place. 

I also looked in the Netty source code while trying to find the correct "unwrapping" logic to replace the invalid solution that was introduced by #2701 . 
This is the comment I found when I knew that this doesn't exist:
https://github.com/netty/netty/blob/2c8b3fb2dea00903843a6f78ce8990b06e0bff96/handler/src/main/java/io/netty/handler/ssl/SslHandler.java#L1118-L1120

I was planning to propose a public API to Netty ByteBuf for visiting the internal buffers. This PR could be a good starting point.

> This implementation is tricky and hard to understand without diving into the internal. After debugging step by step, I found the correctness depends on the implementation details in UnsafeByteBufUtil.

I don't think that there's a way around it. There is a specific implicit contract that Netty ByteBufs follow regarding the `hasArray`, `hasMemoryAddress` and `nioBufferCount` and how that impacts what method in the destination ByteBuf the call to the source ByteBuf's `getBytes` (with `ByteBuf` dest) will be calling.

> Is it possible to achieve the same goal with the following logic?

The Netty ByteBuf contains a lot of classes that are package private. The example of `CompositeByteBuf.getBytes` will be used in the solution that this PR provides.

We have a failing test case that doesn't touch the internals. I'm open to any solution that fixes the checksum calculation bug, that the test case (the original commit works for unchanged master branch and possibly other branches, that is commit https://github.com/lhotari/bookkeeper/commit/9fb47f5dcf3848320991c7a2cf602eb325893a42). 

@BewareMyPower do you have another proposal for the direction to take in fixing this issue?

lhotari: @BewareMyPower I have address your review feedback and attempted to cover your concerns by refactoring the ByteBuf visiting solution. It no longer uses `DuplicatedByteBuf` as the base class. Instead a full custom `ByteBuf` implementation is used with `throw new UnsupportedOperationException()` bodies for all other methods than the ones that are part of the `getBytes`/`setBytes` interface contract. This will ensure that things will break if something unexpected is passed as input.

lhotari:
@merlimat @horizonzy @wenbingshen @eolivelli @dlg99 @nicoloboschi Please review

BewareMyPower (commented):
How about this implementation?

```java
    final int update(int digest, ByteBuf buffer, int offset, int len) {
        if (buffer instanceof CompositeByteBuf) {
            int readerIndex = buffer.readerIndex();
            CompositeByteBuf compositeByteBuf = (CompositeByteBuf) buffer;
            for (int i = 0; i < compositeByteBuf.numComponents(); i++) {
                ByteBuf component = compositeByteBuf.component(i);
                int readableBytes = compositeByteBuf.writerIndex() - readerIndex;
                if (readableBytes > 0) {
                    digest = internalUpdate(digest, component.asByteBuf(), readerIndex, readableBytes);
                    readerIndex = 0;
                } else {
                    readerIndex -= compositeByteBuf.writerIndex();
                }
            }
            return digest;
        } else {
            return internalUpdate(digest, buffer, offset, len);
        }
    }
```

lhotari:
> How about this implementation?

@BewareMyPower The readerIndex state that is lost is in the wrappers. In many cases, there could be multiple layers of wrappers and the main issue is that there's no way to determine whether it's ok or not to call "unwrap". In a lot of cases, it's not ok to call "unwrap". 

It's possible to use the test case in this PR and do attempts in the context of the test case to see what really happens with different approaches. The test case doesn't have all of the possible problem cases. I think that I'll add a scenario for "slice" too.

I'm fairly confident that the approach in the PR is one of rare safe ways to do ["internal iteration" for ByteBuf as the Netty maintainers call it](https://github.com/netty/netty/pull/13820#issuecomment-1919790889). I got some high level feedback from Netty maintainers about the approach: https://github.com/netty/netty/pull/13820 . This is a gap in Netty 4. [In Netty 5, there's a nice API to traverse all components](https://github.com/netty/netty/blob/main/buffer/src/main/java/io/netty5/buffer/Buffer.java#L1086-L1127).

lhotari:
@BewareMyPower I have refactored the solution to use a stateless visitor callback implementation where the state is passed around in a method call parameter. The benefit of this is that it reduces unnecessary object allocations. I'm not concerned about that specificly, but object allocations by callbacks was brought up in the discussion with the Netty maintainer. That's why I decided to handle that aspect too.

lhotari:
There's currently a failing test with heap buffers (byte[] backed buffer) and this PR is still WIP.

eolivelli:
as @merlimat and @dlg99 working on this code recently maybe it is a good idea to wait for their opinion

bookkeeper-server/src/test/java/org/apache/bookkeeper/proto/checksum/CompositeByteBufUnwrapBugReproduceTest.java
```diff
@@ -0,0 +1,241 @@
...
+        assertDigestAndPackageScenario(intHash, payload2, referenceOutput,
+                "payload with prefix wrapped in CompositeByteBuf with readerIndex state");
+
+        ByteBuf payload3 = wrapWithPrefixAndMultipleCompositeByteBufWithReaderIndexStateAndMultipleLayersOfDuplicate(
+                payload.retainedDuplicate());
+        assertDigestAndPackageScenario(intHash, payload3, referenceOutput,
+                "payload with prefix wrapped in 2 layers of CompositeByteBuf with readerIndex state in the outer "
+                        + "composite. In addition, the outer composite is duplicated twice.");
+
+        /*
```
eolivelli, line 151: nit: remove commented code

dlg99 (commented):
Bunch of test timed out so something is wrong.

Would use of CompositeByteBuf's decompose() solve this issue?
Like
((CompositeByteBuf) unwrapped).decompose(data.offset(), data.length()).forEach(..)

graysonzeng:
This issue may also be related ? https://github.com/apache/pulsar/issues/21892 I got some exceptions and tracked it for quite some time @lhotari 

```java
14:29:19.961 [BookKeeperClientWorker-OrderedExecutor-2-0] ERROR org.apache.bookkeeper.common.util.SingleThreadExecutor - Error while running task: readerIndex: 31215, writerIndex: 21324 (expected: 0 <= readerIndex <= writerIndex <= capacity(65536))
java.lang.IndexOutOfBoundsException: readerIndex: 31215, writerIndex: 21324 (expected: 0 <= readerIndex <= writerIndex <= capacity(65536))
	at io.netty.buffer.AbstractByteBuf.checkIndexBounds(AbstractByteBuf.java:112) ~[io.netty-netty-buffer-4.1.104.Final.jar:4.1.104.Final]
	at io.netty.buffer.AbstractByteBuf.setIndex(AbstractByteBuf.java:144) ~[io.netty-netty-buffer-4.1.104.Final.jar:4.1.104.Final]
	at io.netty.buffer.DuplicatedByteBuf.<init>(DuplicatedByteBuf.java:56) ~[io.netty-netty-buffer-4.1.104.Final.jar:4.1.104.Final]
	at io.netty.buffer.DuplicatedByteBuf.<init>(DuplicatedByteBuf.java:42) ~[io.netty-netty-buffer-4.1.104.Final.jar:4.1.104.Final]
	at io.netty.buffer.UnpooledDuplicatedByteBuf.<init>(UnpooledDuplicatedByteBuf.java:24) ~[io.netty-netty-buffer-4.1.104.Final.jar:4.1.104.Final]
	at io.netty.buffer.AbstractPooledDerivedByteBuf$PooledNonRetainedDuplicateByteBuf.<init>(AbstractPooledDerivedByteBuf.java:164) ~[io.netty-netty-buffer-4.1.104.Final.jar:4.1.104.Final]
	at io.netty.buffer.AbstractPooledDerivedByteBuf.duplicate0(AbstractPooledDerivedByteBuf.java:157) ~[io.netty-netty-buffer-4.1.104.Final.jar:4.1.104.Final]
	at io.netty.buffer.PooledSlicedByteBuf.duplicate(PooledSlicedByteBuf.java:118) ~[io.netty-netty-buffer-4.1.104.Final.jar:4.1.104.Final]
	at io.netty.buffer.CompositeByteBuf$Component.duplicate(CompositeByteBuf.java:1947) ~[io.netty-netty-buffer-4.1.104.Final.jar:4.1.104.Final]
	at io.netty.buffer.CompositeByteBuf.component(CompositeByteBuf.java:1556) ~[io.netty-netty-buffer-4.1.104.Final.jar:4.1.104.Final]
	at org.apache.bookkeeper.proto.checksum.DigestManager.computeDigestAndPackageForSendingV2(DigestManager.java:149) ~[bookkeeper-server-4.17.0-SNAPSHOT.jar:4.17.0-SNAPSHOT]
	at org.apache.bookkeeper.proto.checksum.DigestManager.computeDigestAndPackageForSending(DigestManager.java:106) ~[bookkeeper-server-4.17.0-SNAPSHOT.jar:4.17.0-SNAPSHOT]
	at org.apache.bookkeeper.client.PendingAddOp.initiate(PendingAddOp.java:246) ~[bookkeeper-server-4.17.0-SNAPSHOT.jar:4.17.0-SNAPSHOT]
	at org.apache.bookkeeper.client.LedgerHandle.doAsyncAddEntry(LedgerHandle.java:1363) ~[bookkeeper-server-4.17.0-SNAPSHOT.jar:4.17.0-SNAPSHOT]
	at org.apache.bookkeeper.client.LedgerHandle.asyncAddEntry(LedgerHandle.java:1061) ~[bookkeeper-server-4.17.0-SNAPSHOT.jar:4.17.0-SNAPSHOT]
	at org.apache.bookkeeper.mledger.impl.OpAddEntry.initiate(OpAddEntry.java:144) ~[org.apache.pulsar-managed-ledger-3.3.0-SNAPSHOT.jar:3.3.0-SNAPSHOT]
	at org.apache.bookkeeper.mledger.impl.ManagedLedgerImpl.internalAsyncAddEntry(ManagedLedgerImpl.java:862) ~[org.apache.pulsar-managed-ledger-3.3.0-SNAPSHOT.jar:3.3.0-SNAPSHOT]
	at org.apache.bookkeeper.mledger.impl.ManagedLedgerImpl.lambda$asyncAddEntry$3(ManagedLedgerImpl.java:794) ~[org.apache.pulsar-managed-ledger-3.3.0-SNAPSHOT.jar:3.3.0-SNAPSHOT]
	at org.apache.bookkeeper.common.util.SingleThreadExecutor.safeRunTask(SingleThreadExecutor.java:137) ~[org.apache.bookkeeper-bookkeeper-common-4.16.3.jar:4.16.3]
	at org.apache.bookkeeper.common.util.SingleThreadExecutor.run(SingleThreadExecutor.java:107) ~[org.apache.bookkeeper-bookkeeper-common-4.16.3.jar:4.16.3]
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30) ~[io.netty-netty-common-4.1.104.Final.jar:4.1.104.Final]
	at java.lang.Thread.run(Thread.java:833) ~[?:?]
```

lhotari:
> Bunch of test timed out so something is wrong.

Yes, that was the WIP that I was working to fix. I have a fix, but since there are a lot of changes, it's possible that this will have be split into multiple PRs. I found an issue with Java9IntHash, but I haven't yet added a unit test to demonstrate the issue. 

> Would use of CompositeByteBuf's decompose() solve this issue?
> Like
> ((CompositeByteBuf) unwrapped).decompose(data.offset(), data.length()).forEach(..)

No. There are multiple layers of wrappers and it's not safe to unwrap the wrappers. I hope the previous comment https://github.com/apache/bookkeeper/pull/4196#issuecomment-1920714720 clarifies this.

lhotari:
> as @merlimat and @dlg99 working on this code recently maybe it is a good idea to wait for their opinion

Sure, makes sense. This PR isn't yet ready for merging since CI isn't passing yet. :)

dlg99:
> No. There are multiple layers of wrappers and it's not safe to unwrap the wrappers. I hope the previous comment [#4196 (comment)](https://github.com/apache/bookkeeper/pull/4196#issuecomment-1920714720) clarifies this.

as I understood, if you take offset (readerIndex) and length from the "wrapped" buffer, unwrapped.decompose(offset, len) will use these to provide proper buffers (slices if needed). I just briefly looked at the code of decompose, please double-check.

lhotari:
> as I understood, if you take offset (readerIndex) and length from the "wrapped" buffer, unwrapped.decompose(offset, len) will use these to provide proper buffers (slices if needed). I just briefly looked at the code of decompose, please double-check.


@dlg99 won't work. You would need to know implementation details. If you check the implementations, there's also other internal state in wrappers that cannot be observed. The Netty 4 API is useless in many ways for decomposing buffers. This problem is resolved in Netty 5 API. The previous discussion has a link to the draft PR where I asked about this from Netty maintainers.

The ByteBufVisitor solution in this PR is a result of many iterations. I actually tried the solution you are describing as one of the first attempts and ran into a dead-end. I'm confident that there isn't many other safe approaches for Netty 4, other than the getBytes/setBytes hack used in ByteBufVisitor. It's a safe hack.

The stateless visitor minimises object allocations and that won't become a bottleneck.

The other benefit of ByteBufVisitor is that it will be able to iterate through any buffer hierarchy. That will also improve performance, although it's not likely that there are many of these cases in the current hotpaths.

lhotari:
@dlg99 @eolivelli @merlimat This is now ready for review.

lhotari:
I have significantly improvement this PR. I hope to finally receive review feedback about the code in this PR. 
Please review @dlg99 @eolivelli @merlimat @hangc0276 @BewareMyPower . Thanks!

As a reminder, there's a test that reproduces a problem with unwrapping and the checksum calculation. That is the very first commit in this PR, it is https://github.com/apache/bookkeeper/pull/4196/commits/9fb47f5dcf3848320991c7a2cf602eb325893a42 .

Here's one way to reproduce on the command line:
```
gh pr checkout 4196
git checkout 9fb47f5d
mvn clean install -DskipTests
mvn -pl bookkeeper-server test -Dtest=CompositeByteBufUnwrapBugReproduceTest
```

Besides fixing problems, this PR will also improve performance across all checksum algorithms since input ByteBufs will be processed in most optimal ways across all implementations.

bookkeeper-server/src/test/java/org/apache/bookkeeper/proto/checksum/CompositeByteBufUnwrapBugReproduceTest.java
lhotari: done

.github/workflows/bk-ci.yml
```diff
@@ -200,6 +200,10 @@ jobs:
           path: surefire-reports
           retention-days: 7
 
+      - name: print JVM thread dumps when cancelled
+        if: cancelled()
+        run: ./dev/ci-tool print_thread_dumps
+
```
lhotari, line 206: One of the build jobs gets stuck once in a while. I added the thread dump printing to the build to find out what is causing that. It seems to be existing test flakiness, but this way we will find out sooner or later.

lhotari: DeferredSyncTest.testForceWillAdvanceLacOnlyUpToLastAcknoledgedWrite gets stuck. I think that the test if flaky itself.
```
"main" #1 prio=5 os_prio=0 cpu=1865.56ms elapsed=1477.95s tid=0x00007f34b4028000 nid=0x149f5 waiting on condition  [0x00007f34bcffc000]
   java.lang.Thread.State: TIMED_WAITING (parking)
	at jdk.internal.misc.Unsafe.park(java.base@11.0.22/Native Method)
	- parking to wait for  <0x000000008ad136e8> (a java.util.concurrent.CompletableFuture$Signaller)
	at java.util.concurrent.locks.LockSupport.parkNanos(java.base@11.0.22/LockSupport.java:234)
	at java.util.concurrent.CompletableFuture$Signaller.block(java.base@11.0.22/CompletableFuture.java:1798)
	at java.util.concurrent.ForkJoinPool.managedBlock(java.base@11.0.22/ForkJoinPool.java:3128)
	at java.util.concurrent.CompletableFuture.timedGet(java.base@11.0.22/CompletableFuture.java:1871)
	at java.util.concurrent.CompletableFuture.get(java.base@11.0.22/CompletableFuture.java:2027)
	at org.apache.bookkeeper.common.concurrent.FutureUtils.result(FutureUtils.java:80)
	at org.apache.bookkeeper.common.concurrent.FutureUtils.result(FutureUtils.java:61)
	at org.apache.bookkeeper.client.DeferredSyncTest.testForceWillAdvanceLacOnlyUpToLastAcknoledgedWrite(DeferredSyncTest.java:178)
```

lhotari: Reported as #4200

eolivelli (approved):
LGTM

great work

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/ByteBufVisitor.java
```diff
@@ -0,0 +1,1132 @@
...
+     * See @{@link #visitBuffers(ByteBuf, int, int, ByteBufVisitorCallback, Object)}. This method
+     * allows to specify the maximum depth of recursion for visiting wrapped buffers.
+     */
+    public static <T> void visitBuffers(ByteBuf buffer, int offset, int length, ByteBufVisitorCallback<T> callback,
+                                        T context, int maxDepth) {
+        if (length == 0) {
+            // skip visiting empty buffers
+            return;
+        }
+        InternalContext<T> internalContext = new InternalContext<>();
```
hangc0276, line 104: `visitBuffers` is a high-frequency call. Can we reuse the InternalContext to reduce the object allocation?

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/ByteBufVisitor.java
```diff
@@ -0,0 +1,1132 @@
...
+        private void passBufferToCallback(ByteBuf visitBuffer, int visitIndex, int visitLength) {
+            if (callback.preferArrayOrMemoryAddress(callbackContext)) {
+                if (visitBuffer.hasArray()) {
+                    handleArray(visitBuffer.array(), visitBuffer.arrayOffset() + visitIndex, visitLength);
+                } else if (visitBuffer.hasMemoryAddress() && callback.acceptsMemoryAddress(callbackContext)) {
+                    callback.visitBuffer(callbackContext, visitBuffer, visitIndex, visitLength);
+                } else if (callback.acceptsMemoryAddress(callbackContext) && visitBuffer.isDirect()
+                        && visitBuffer.alloc().isDirectBufferPooled()) {
+                    // read-only buffers need to be copied before they can be directly accessed
+                    ByteBuf copyBuffer = visitBuffer.copy(visitIndex, visitLength);
```
hangc0276, line 168: We will copy the ByteBuf here, will this path will high-frequency called?

hangc0276:
Do we need a CPU and object alloc profile to see if this change introduces any performance issues?

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/ByteBufVisitor.java
lhotari: The reason to have that InternalContext is to enable sharing of all other instances. There will be only a single instance for each root level buffer so object allocation isn't a problem.

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/ByteBufVisitor.java
lhotari: When it gets here, the source buffer (`visitBuffer`) will need to be copied in any case.
Since this checks that the source buffer is a direct buffer and that the callback accepts a memory address buffer (direct buffer), it's better to make a direct copy than to use the thread local copy buffer (byte[]) for copying.

The last resort is copying the source buffer using the byte[] copying. 

Netty won't allow visiting the backing direct buffer or backing array when the buffer is a read-only buffer. In all other cases, the ByteBufVisitor is able to visit all source buffers without copies. This is the most efficient solution that there is available.

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/ByteBufVisitor.java
lhotari: This PR removes some allocations that were happening in the old solution so there won't be additional allocations caused by this PR. One example of removed allocations is the reduced use of `slice`. When you call `slice` on a Netty ByteBuf, it will create an instance. The solution will be passing a byte[] or a direct buffer to the checksum calculations and the existing solutions to extract to use `slice` to get the buffer content won't be used any more. 

For example this has been creating a lot of objects when CRC32 is used:
https://github.com/apache/bookkeeper/blob/22097343ceed1312964ad077cdcd9fbd7e3e026b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/checksum/DirectMemoryCRC32Digest.java#L55-L58

Similarly Java9IntHash was using `slice` here:
https://github.com/apache/bookkeeper/blob/22097343ceed1312964ad077cdcd9fbd7e3e026b/circe-checksum/src/main/java/com/scurrilous/circe/checksum/Java9IntHash.java#L123

JniIntHash here:
https://github.com/apache/bookkeeper/blob/22097343ceed1312964ad077cdcd9fbd7e3e026b/circe-checksum/src/main/java/com/scurrilous/circe/checksum/JniIntHash.java#L51

All of this is now optimized and won't end up using this code since ByteBufVisitor will be extracting (or copying) byte[] or direct buffers (memory address) without any object allocations. Previously there has been object allocations.

lhotari:
> This issue may also be related ? [apache/pulsar#21892](https://github.com/apache/pulsar/issues/21892) I got some exceptions and tracked it for quite some time @lhotari
> 
> ```java
> 14:29:19.961 [BookKeeperClientWorker-OrderedExecutor-2-0] ERROR org.apache.bookkeeper.common.util.SingleThreadExecutor - Error while running task: readerIndex: 31215, writerIndex: 21324 (expected: 0 <= readerIndex <= writerIndex <= capacity(65536))
> java.lang.IndexOutOfBoundsException: readerIndex: 31215, writerIndex: 21324 (expected: 0 <= readerIndex <= writerIndex <= capacity(65536))
> 	at io.netty.buffer.AbstractByteBuf.checkIndexBounds(AbstractByteBuf.java:112) ~[io.netty-netty-buffer-4.1.104.Final.jar:4.1.104.Final]
> 	at io.netty.buffer.AbstractByteBuf.setIndex(AbstractByteBuf.java:144) ~[io.netty-netty-buffer-4.1.104.Final.jar:4.1.104.Final]
> 	at io.netty.buffer.DuplicatedByteBuf.<init>(DuplicatedByteBuf.java:56) ~[io.netty-netty-buffer-4.1.104.Final.jar:4.1.104.Final]
> 	at io.netty.buffer.DuplicatedByteBuf.<init>(DuplicatedByteBuf.java:42) ~[io.netty-netty-buffer-4.1.104.Final.jar:4.1.104.Final]
> 	at io.netty.buffer.UnpooledDuplicatedByteBuf.<init>(UnpooledDuplicatedByteBuf.java:24) ~[io.netty-netty-buffer-4.1.104.Final.jar:4.1.104.Final]
> 	at io.netty.buffer.AbstractPooledDerivedByteBuf$PooledNonRetainedDuplicateByteBuf.<init>(AbstractPooledDerivedByteBuf.java:164) ~[io.netty-netty-buffer-4.1.104.Final.jar:4.1.104.Final]
> 	at io.netty.buffer.AbstractPooledDerivedByteBuf.duplicate0(AbstractPooledDerivedByteBuf.java:157) ~[io.netty-netty-buffer-4.1.104.Final.jar:4.1.104.Final]
> 	at io.netty.buffer.PooledSlicedByteBuf.duplicate(PooledSlicedByteBuf.java:118) ~[io.netty-netty-buffer-4.1.104.Final.jar:4.1.104.Final]
> 	at io.netty.buffer.CompositeByteBuf$Component.duplicate(CompositeByteBuf.java:1947) ~[io.netty-netty-buffer-4.1.104.Final.jar:4.1.104.Final]
> 	at io.netty.buffer.CompositeByteBuf.component(CompositeByteBuf.java:1556) ~[io.netty-netty-buffer-4.1.104.Final.jar:4.1.104.Final]
> 	at org.apache.bookkeeper.proto.checksum.DigestManager.computeDigestAndPackageForSendingV2(DigestManager.java:149) ~[bookkeeper-server-4.17.0-SNAPSHOT.jar:4.17.0-SNAPSHOT]
> 	at org.apache.bookkeeper.proto.checksum.DigestManager.computeDigestAndPackageForSending(DigestManager.java:106) ~[bookkeeper-server-4.17.0-SNAPSHOT.jar:4.17.0-SNAPSHOT]
> 	at org.apache.bookkeeper.client.PendingAddOp.initiate(PendingAddOp.java:246) ~[bookkeeper-server-4.17.0-SNAPSHOT.jar:4.17.0-SNAPSHOT]
> 	at org.apache.bookkeeper.client.LedgerHandle.doAsyncAddEntry(LedgerHandle.java:1363) ~[bookkeeper-server-4.17.0-SNAPSHOT.jar:4.17.0-SNAPSHOT]
> 	at org.apache.bookkeeper.client.LedgerHandle.asyncAddEntry(LedgerHandle.java:1061) ~[bookkeeper-server-4.17.0-SNAPSHOT.jar:4.17.0-SNAPSHOT]
> 	at org.apache.bookkeeper.mledger.impl.OpAddEntry.initiate(OpAddEntry.java:144) ~[org.apache.pulsar-managed-ledger-3.3.0-SNAPSHOT.jar:3.3.0-SNAPSHOT]
> 	at org.apache.bookkeeper.mledger.impl.ManagedLedgerImpl.internalAsyncAddEntry(ManagedLedgerImpl.java:862) ~[org.apache.pulsar-managed-ledger-3.3.0-SNAPSHOT.jar:3.3.0-SNAPSHOT]
> 	at org.apache.bookkeeper.mledger.impl.ManagedLedgerImpl.lambda$asyncAddEntry$3(ManagedLedgerImpl.java:794) ~[org.apache.pulsar-managed-ledger-3.3.0-SNAPSHOT.jar:3.3.0-SNAPSHOT]
> 	at org.apache.bookkeeper.common.util.SingleThreadExecutor.safeRunTask(SingleThreadExecutor.java:137) ~[org.apache.bookkeeper-bookkeeper-common-4.16.3.jar:4.16.3]
> 	at org.apache.bookkeeper.common.util.SingleThreadExecutor.run(SingleThreadExecutor.java:107) ~[org.apache.bookkeeper-bookkeeper-common-4.16.3.jar:4.16.3]
> 	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30) ~[io.netty-netty-common-4.1.104.Final.jar:4.1.104.Final]
> 	at java.lang.Thread.run(Thread.java:833) ~[?:?]
> ```

@graysonzeng This looks like a multi-threading / thread-safety issue. Possibly multiple threads are mutating the same Netty ByteBuf instance. ByteBuf isn't thread safe. It might not be related to the problem that this PR is fixing.

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/ByteBufVisitor.java
hangc0276: I see. This PR won't introduce more object allocation. I'm not blocking this PR, and it will be perfect if we can reduce the InternalContext object allocation. For each entry, it will create two InternalContext objects.

hangc0276 (approved):

merlimat (approved):

StevenLuMT:
Good work

PR #4199, opened 2024-02-04
log error when ReadEntryProcessor IOException

AnonHxy:

### Motivation

Change the log-level to error when `ReadEntryProcessor` throw `IOException`.   The reason is that:
1. `IOException` should not happen frequently，so the error log will not print too much
2. It will be convenient for troubleshooting  if we print the  error log.
3. Also we can see that `ReadEntryProcessorV3` has printed the error log

https://github.com/apache/bookkeeper/blob/dc2bb1dfed38d7f9ad7a29fc92bd907e51af5b21/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadEntryProcessorV3.java#L233-L236

### Changes

`log.debug` -> `log.error`

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadEntryProcessor.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadEntryProcessor.java
@@ -105,9 +105,7 @@ class ReadEntryProcessor extends PacketProcessorBase<ReadRequest> {
             }
             errorCode = BookieProtocol.ENOENTRY;
         } catch (IOException e) {
-            if (LOG.isDebugEnabled()) {
-                LOG.debug("Error reading {}", request, e);
-            }
+            LOG.error("Error reading {}", request, e);
             errorCode = BookieProtocol.EIO;
         } catch (BookieException.DataUnknownException e) {
             LOG.error("Ledger {} is in an unknown state", request.getLedgerId(), e);


bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadEntryProcessor.java
```diff
@@ -105,9 +105,7 @@ protected void processPacket() {
             }
             errorCode = BookieProtocol.ENOENTRY;
         } catch (IOException e) {
-            if (LOG.isDebugEnabled()) {
-                LOG.debug("Error reading {}", request, e);
-            }
+            LOG.error("IOException while reading {}", request, e);
```
merlimat, line 108: I believe this would print the whole stack trace every time a client gets a connectivity error with any bookie, for each entry that we're trying to read.

AnonHxy: > whole stack trace every time a client gets a connectivity error with any bookie

Thanks for your review @merlimat . But I don't get your point. This method runs only on serever side and probability throw `IOException` if read an entry from storage. So I think it doesn't matter with client print logs

hangc0276: Pring `ledgerId:entryId` instead of the whole request content?

hangc0276 (approved):
Overall LGTM.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadEntryProcessor.java
AnonHxy: I think print  whole `request ` is OK.  because `ReadRequest` has overwrite the `toString` method:
https://github.com/apache/bookkeeper/blob/34bd4b38fa13f105608ae4e7b4c749124c271e1e/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieProtocol.java#L249-L251
@hangc0276

hangc0276: >I believe this would print the whole stack trace every time a client gets a connectivity error with any bookie, for each entry that we're trying to read.

If the target entry does not exist on the bookie, it will throw NoLedgerException or NoEntryException and won't go into this code path. @merlimat Do you have any concern?

horizonzy: > > I believe this would print the whole stack trace every time a client gets a connectivity error with any bookie, for each entry that we're trying to read.
> 
> If the target entry does not exist on the bookie, it will throw NoLedgerException or NoEntryException and won't go into this code path. @merlimat Do you have any concern?

The NoLedgerException and NoEntryException extend IOException, it should be caught.

horizonzy (approved):
LGTM.

shoothzj:
@merlimat Could you please take a look again?

PR #4208, opened 2024-02-09
Issue 2161: set metrics endpoint content-type

### LINKED ISSUE #2161: When httpServerEnabled=true, Prometheus reports doesn't set content-type
**BUG REPORT**

***Describe the bug***

When running with `httpServerEnabled` set to true, `/metrics` endpoint stops sending a `content-type` header (this uses the `MetricsService`, see code here https://github.com/apache/bookkeeper/blob/28dee8464764b0edceeb31ce24424f0947b5789b/bookkeeper-server/src/main/java/org/apache/bookkeeper/server/http/service/MetricsService.java).
With the `httpServerEnabled=false` the `PrometheusServlet` is used, and it does set a content-type header (https://github.com/apache/bookkeeper/blob/28dee8464764b0edceeb31ce24424f0947b5789b/bookkeeper-stats-providers/prometheus-metrics-provider/src/main/java/org/apache/bookkeeper/stats/prometheus/PrometheusServlet.java#L44)

 The content-type header is expected by some prometheus tools to differentiate between plaintext and protobuf formatted metrics and no content-type is considered an error.

***To Reproduce***

1. Set `httpServerEnabled=false` and enable prometheus reporter
2. run `curl localhost:8000/metrics` and notice that a content-type header is set
3. Set `httpServerEnabled=true` and enable prometheus reporter
4. run `curl localhost:8000/metrics -v` notice that there is no content-type header set

***Expected behavior***

I expect a content-type header

It seems like the statsProvider interface should have a method for allowing to set headers on the response. Default to a given content type might cause problems with other providers.

tomjo:
### Motivation

Fixes #2161

### Changes

Add contentType to **HttpServiceResponse** and map it in **VertxAbstractHandler** which allows
**MetricsService** (which is used instead of **PrometheusServlet** when bookie http server is enabled) to set the http **Content-Type** header to the right value which allow successful scraping of the endpoint by e.g. New Relic

Master Issue: #2161

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/server/http/service/MetricsService.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/server/http/service/MetricsService.java
@@ -35,6 +35,8 @@ import org.apache.bookkeeper.stats.StatsProvider;
  */
 public class MetricsService implements HttpEndpointService {
 
+    public static final String PROMETHEUS_CONTENT_TYPE_004 = "text/plain; version=0.0.4; charset=utf-8";
+
     private final ServerConfiguration conf;
     private final StatsProvider statsProvider;
 
@@ -65,6 +67,7 @@ public class MetricsService implements HttpEndpointService {
             statsProvider.writeAllMetrics(writer);
             writer.flush();
             response.setCode(StatusCode.OK);
+            response.setContentType(PROMETHEUS_CONTENT_TYPE_004);
             response.setBody(writer.getBuffer().toString());
         } catch (UnsupportedOperationException uoe) {
             response.setCode(StatusCode.INTERNAL_ERROR);
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/server/http/service/MetricsService.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/server/http/service/MetricsService.java
@@ -18,7 +18,6 @@
 
 package org.apache.bookkeeper.server.http.service;
 
-import io.prometheus.client.exporter.common.TextFormat;
 import java.io.IOException;
 import java.io.StringWriter;
 import org.apache.bookkeeper.conf.ServerConfiguration;
@@ -35,6 +34,8 @@ import org.apache.bookkeeper.stats.StatsProvider;
  * <p>The GET method will return all the emtrics collected at stats provider.
  */
 public class MetricsService implements HttpEndpointService {
+
+    public static final String PROMETHEUS_CONTENT_TYPE_004 = "text/plain; version=0.0.4; charset=utf-8";
 
     private final ServerConfiguration conf;
     private final StatsProvider statsProvider;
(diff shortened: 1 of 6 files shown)


merlimat (approved):

shoothzj (approved):

bookkeeper-server/src/main/java/org/apache/bookkeeper/server/http/service/MetricsService.java
```diff
@@ -65,6 +67,7 @@ public HttpServiceResponse handle(HttpServiceRequest request) throws Exception {
             statsProvider.writeAllMetrics(writer);
             writer.flush();
             response.setCode(StatusCode.OK);
+            response.setContentType(PROMETHEUS_CONTENT_TYPE_004);
```
hangc0276, line 70: Use `TextFormat.CONTENT_TYPE_004` instead of defining a new one. Just like: https://github.com/apache/bookkeeper/blob/eb27e6ed3ccd1ee46f372ae6fd0a53ccea81908f/stats/bookkeeper-stats-providers/prometheus-metrics-provider/src/main/java/org/apache/bookkeeper/stats/prometheus/PrometheusServlet.java#L42

hangc0276 (approved):

bookkeeper-server/src/main/java/org/apache/bookkeeper/server/http/service/MetricsService.java
tomjo: This will require dependency changes wasn't sure of that. I can either change the scope of the org.apache.bookkeeper.stats:prometheus-metrics-provider artifact to non-test or add an explicit dependency on io.prometheus:simpleclient_common

tomjo: I chose the former approach.

bookkeeper-server/src/main/java/org/apache/bookkeeper/server/http/service/MetricsService.java
```diff
@@ -65,6 +66,7 @@ public HttpServiceResponse handle(HttpServiceRequest request) throws Exception {
             statsProvider.writeAllMetrics(writer);
             writer.flush();
             response.setCode(StatusCode.OK);
+            response.setContentType(TextFormat.CONTENT_TYPE_004);
```
eolivelli, line 69: we really have to add a new jar only for a string constant ?
I think that it is overkill

can we drop the jar and use a normal string here ?

eolivelli (changes requested):
LGTM

bookkeeper-server/src/main/java/org/apache/bookkeeper/server/http/service/MetricsService.java
tomjo: I can revert those changes, are the maintainers in agreement? @hangc0276

hangc0276: Sorry, I forgot it will introduce a new dependency. I agree to revert those changes.

hangc0276:
@eolivelli Would you please review this PR again? Thanks.

eolivelli (approved):
Looks great 

+1

lhotari:
Until this change is available, it's possible to configure `fallbackScrapeProtocol` in Prometheus at multiple levels. Here's an example for Apache Pulsar Helm chart https://github.com/apache/pulsar-helm-chart/pull/577 .

PR #4212, opened 2024-02-19
Use mockito instead of powermock

### LINKED ISSUE #4201: Replace the usage of PowerMock with Mockito in the Bookkeeper tests
[PowerMock](https://github.com/powermock/powermock/releases) is no longer maintained. The last release way in January 2020.
It should be replaced with Mockito. A similar replacement was done in Pulsar in these PRs:

- https://github.com/apache/pulsar/pull/14098
- https://github.com/apache/pulsar/pull/14250
- https://github.com/apache/pulsar/pull/17696

PowerMock / PowerMockito is holding back from upgrading to Mockito 4.11.0 . Mockito 4.11.x is the last version to support Java 8.
Getting rid of PowerMock will be needed for upgrading some other libraries for supporting newer JDK versions (Java 17 & Java 21).

nodece:
Closes #4201

### Motivation

[PowerMock](https://github.com/powermock/powermock/releases) is no longer maintained. The last release way in January 2020.

### Changes

- Use mockito instead of powermock in the codebase
- Remove powermock library from pom.xml
- Change some code is used for mock/spy object


lhotari:
Great job @nodece

lhotari:
@nodece the next step would be to upgrade Mockito to 4.11. That's the last version that supports Java 8. Would you like to handle it. There are some minor breaking changes in Mockito between 3.x -> 4.x. the upgrade PR in Pulsar could be a useful reference

PR #4217, opened 2024-02-21
Fix getHosts by ip

nodece:
### Motivation

- The PTR result will end with a `.`, we need to trim that, please see https://datatracker.ietf.org/doc/html/rfc1035#autoid-37
   
- In testing, we use the loopback net interface, which doesn't need to do PTR parse by DNS: 

```
java.lang.RuntimeException: java.net.UnknownHostException: Unable to resolve default hostname: 1.0.0.127.in-addr.arpa. for interface: lo0

 at org.apache.bookkeeper.proto.SimpleBookieServiceInfoProvider.<init>(SimpleBookieServiceInfoProvider.java:41)
 at org.apache.bookkeeper.bookie.TestBookieImpl.<init>(TestBookieImpl.java:71)
 at org.apache.bookkeeper.bookie.TestBookieImpl.<init>(TestBookieImpl.java:46)
 at org.apache.bookkeeper.bookie.BookieWriteToJournalTest.testForceLedger(BookieWriteToJournalTest.java:167)
```

### Changes

- Trim the PTR result.
- Skip PTR parsing of loopback address by DNS.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/net/DNS.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/net/DNS.java
@@ -99,7 +99,13 @@ public class DNS {
             throw new NamingException("PTR attribute value is null");
         }
 
-        return ptrAttr.get().toString();
+        String ptr = ptrAttr.get().toString();
+        int ptrLen = ptr.length();
+        if (ptr.charAt(ptrLen - 1) == '.') {
+            ptr = ptr.substring(0, ptrLen - 1);
+        }
+
+        return ptr;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


shoothzj (changes requested):
If it's a given that the result will end with a dot, could the if check be potentially redundant?
Additionally, I wonder if the getDefaultHost code has been consistently encountering resolution errors, might it indicate that this block of code has not been effectively utilized? Perhaps we can remove this method?
cc @eolivelli @merlimat

shoothzj:
We has been config like that:

```bash
bash-5.0$ cat conf/bk_server.conf |grep -i listen
listeningInterface=eth0
```

nodece:
Alternative approach: #4588

PR #4219, opened 2024-02-23
BenchReadThroughputLatency support config the way of gen ledger path.

horizonzy:
### Motivation
In BenchReadThroughputLatency, it generates the ledger zk node path using 
`String.format("/ledgers/L%010d", ledger.get());`

But if the user uses `HierarchicalLedgerManager`, `LegacyHierarchicalLedgerManager`, or `LongHierarchicalLedgerManager`, the ledger zk node path is not the same as the way in the BenchReadThroughputLatency.

So we need support config the `genLedgerWay` to match the custom LedgerManager.

### DIFF
--- a/bookkeeper-benchmark/src/main/java/org/apache/bookkeeper/benchmark/BenchReadThroughputLatency.java
+++ b/bookkeeper-benchmark/src/main/java/org/apache/bookkeeper/benchmark/BenchReadThroughputLatency.java
@@ -36,6 +36,7 @@ import org.apache.bookkeeper.client.BookKeeper;
 import org.apache.bookkeeper.client.LedgerEntry;
 import org.apache.bookkeeper.client.LedgerHandle;
 import org.apache.bookkeeper.conf.ClientConfiguration;
+import org.apache.bookkeeper.util.StringUtils;
 import org.apache.bookkeeper.zookeeper.ZooKeeperClient;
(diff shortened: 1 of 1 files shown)


bookkeeper-benchmark/src/main/java/org/apache/bookkeeper/benchmark/BenchReadThroughputLatency.java
```diff
@@ -153,6 +154,9 @@ public static void main(String[] args) throws Exception {
         Options options = new Options();
         options.addOption("ledger", true, "Ledger to read. If empty, read all ledgers which come available. "
                           + " Cannot be used with -listen");
+        //How to generate ledger id
+        options.addOption("genLedgerWay", true, "The way of generating ledgerId. "
```
hangc0276, line 158: Use `ledgerManagerType`? https://github.com/apache/bookkeeper/blob/34bd4b38fa13f105608ae4e7b4c749124c271e1e/conf/bk_server.conf#L827-L828

horizonzy: > ledgerManagerType

Make sense.

hangc0276 (approved):

PR #4220, opened 2024-02-23
Fix BenchReadThroughputLatency batch read can't stop problems.

horizonzy:
Fix BenchReadThroughputLatency batch read can't stop problems.

### DIFF
--- a/bookkeeper-benchmark/src/main/java/org/apache/bookkeeper/benchmark/BenchReadThroughputLatency.java
+++ b/bookkeeper-benchmark/src/main/java/org/apache/bookkeeper/benchmark/BenchReadThroughputLatency.java
@@ -102,7 +102,7 @@ public class BenchReadThroughputLatency {
                 }
                 long starttime = System.nanoTime();
 
-                while (entriesRead < lastConfirmed) {
+                while (lastRead < lastConfirmed) {
                     long nextLimit = lastRead + 100000;
                     Enumeration<LedgerEntry> entries;
                     if (batchEntries <= 0) {
--- a/bookkeeper-benchmark/src/main/java/org/apache/bookkeeper/benchmark/BenchReadThroughputLatency.java
+++ b/bookkeeper-benchmark/src/main/java/org/apache/bookkeeper/benchmark/BenchReadThroughputLatency.java
@@ -78,7 +78,7 @@ public class BenchReadThroughputLatency {
         BookKeeper bk = null;
         long time = 0;
         long entriesRead = 0;
-        long lastRead = 0;
+        long lastRead = -1;
         int nochange = 0;
 
         long absoluteLimit = 5000000;
@@ -103,7 +103,7 @@ public class BenchReadThroughputLatency {
                 long starttime = System.nanoTime();
 
                 while (entriesRead < lastConfirmed) {
-                    long nextLimit = lastRead + 100000;
+                    long nextLimit = lastRead + 10;
                     Enumeration<LedgerEntry> entries;
                     if (batchEntries <= 0) {
                         long readTo = Math.min(nextLimit, lastConfirmed);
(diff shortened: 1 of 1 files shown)


bookkeeper-benchmark/src/main/java/org/apache/bookkeeper/benchmark/BenchReadThroughputLatency.java
```diff
@@ -89,7 +89,7 @@ private static void readLedger(ClientConfiguration conf, long ledgerId, byte[] p
                 lh = bk.openLedgerNoRecovery(ledgerId, BookKeeper.DigestType.CRC32,
                                              passwd);
                 long lastConfirmed = Math.min(lh.getLastAddConfirmed(), absoluteLimit);
-                if (lastConfirmed == lastRead) {
+                if (lastConfirmed <= lastRead) {
```
horizonzy, line 92: if the read is batched, the lastRead may be greater than lastConfirmed.

bookkeeper-benchmark/src/main/java/org/apache/bookkeeper/benchmark/BenchReadThroughputLatency.java
```diff
@@ -102,14 +102,14 @@ private static void readLedger(ClientConfiguration conf, long ledgerId, byte[] p
                 }
                 long starttime = System.nanoTime();
 
-                while (entriesRead <= lastConfirmed) {
+                while (entriesRead < lastConfirmed) {
                     long nextLimit = lastRead + 100000;
                     Enumeration<LedgerEntry> entries;
                     if (batchEntries <= 0) {
                         long readTo = Math.min(nextLimit, lastConfirmed);
                         entries = lh.readEntries(lastRead + 1, readTo);
```
horizonzy, line 110: If the read is single read, we should make sure that entriesRead is less than lastConfirmed. Or the `lastRead + 1` will be greater than `readTo`.

bookkeeper-benchmark/src/main/java/org/apache/bookkeeper/benchmark/BenchReadThroughputLatency.java
```diff
@@ -102,14 +102,14 @@ private static void readLedger(ClientConfiguration conf, long ledgerId, byte[] p
...
                     } else {
-                        entries = lh.batchReadEntries(lastRead, batchEntries, -1);
+                        entries = lh.batchReadEntries(lastRead + 1, batchEntries, -1);
```
horizonzy, line 112: The `lastRead` entry already be read, so we should read next entry at the next read.

hangc0276: Do we need init the `lastRead` to `-1` instead of `0`, otherwise the entry `0` can't be read

horizonzy: Nice catch.

bookkeeper-benchmark/src/main/java/org/apache/bookkeeper/benchmark/BenchReadThroughputLatency.java
```diff
@@ -102,14 +102,14 @@ private static void readLedger(ClientConfiguration conf, long ledgerId, byte[] p
                 }
                 long starttime = System.nanoTime();
 
-                while (entriesRead <= lastConfirmed) {
-                    long nextLimit = lastRead + 100000;
+                while (entriesRead < lastConfirmed) {
+                    long nextLimit = lastRead + 10;
```
hangc0276, line 106: Why change it to 10?

bookkeeper-benchmark/src/main/java/org/apache/bookkeeper/benchmark/BenchReadThroughputLatency.java
```diff
@@ -89,7 +89,7 @@ private static void readLedger(ClientConfiguration conf, long ledgerId, byte[] p
                 lh = bk.openLedgerNoRecovery(ledgerId, BookKeeper.DigestType.CRC32,
                                              passwd);
                 long lastConfirmed = Math.min(lh.getLastAddConfirmed(), absoluteLimit);
-                if (lastConfirmed == lastRead) {
+                if (lastConfirmed <= lastRead + 1) {
```
horizonzy, line 92: Because we from the entry 0 to start read, so here we should make `lastRead + 1` to compare with lastConfirmed.

hangc0276 (approved):

PR #4225, opened 2024-03-06
Fix the bookie doesn't transition to writable

zymap:
---

### Motivation

When bookie started with readonly, and the disk usage is smaller than the warning threshold, it won't transition to write mode. Notify the disk writable when init the bookie

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDirsMonitor.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDirsMonitor.java
@@ -66,7 +66,7 @@ class LedgerDirsMonitor {
         this.dirsManagers = dirsManagers;
     }
 
-    private void check(final LedgerDirsManager ldm) {
+    void check(final LedgerDirsManager ldm, boolean notifyDiskWritable) {
         final ConcurrentMap<File, Float> diskUsages = ldm.getDiskUsages();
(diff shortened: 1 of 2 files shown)


hangc0276 (commented):
The bookie starts with ReadOnly has three cases:
- Case 1: Bookie starts in ReadOnly mode by startup command
- Case 2: Bookie transmits to ReadOnly mode due to disk usage before the restart, and the ReadOnly mode persists in storage. The next time, it starts with the ReadOnly mode
- Case 3: Bookie changed to ReadOnly mode by REST API and persisted the ReadOnly mode to storage before the restart. In the next time, it starts with ReadOnly mode.

In this PR, you add the notify flag in the init stage, which means the bookie starts with ReadOnly mode, but if disk usage is below the disk WARN threshold, it will transmit to ReadWrite mode. It fixes some situations in Case 2, but it will change the behavior of Case 1 and Case 3.

zymap:
For the first case, I think if you start bookie with readonly, the bookie will in forceReadonly and won't be writable anymore. So even if the disk has space and it won't transition to the write mode.

For the third case, yes. It is very difficult to know if it is controlled by the rest API or the disk threshold for now. So if you want to make bookie into readonly forever, you need to make it to the forceReadonly, not the readonly state. So i think it should be acceptable?

mattisonchao:
image
Do we need safe run here?

Issue #4238, opened 2024-03-21
Bookie could coredump when it is shutting down


AnonHxy:
**BUG REPORT**

***Describe the bug***

When we kill the bookie processor and shut it down, sometimes core dump could happen.  Core dump log：
```
---------------  T H R E A D  ---------------

Current thread (0x00007f14a7128350):  JavaThread "qtp1834755909-1155" [_thread_in_native, id=7358, stack(0x00007f053cbfc000,0x00007f053ccfd000)]

Stack: [0x00007f053cbfc000,0x00007f053ccfd000],  sp=0x00007f053ccfb130,  free space=1020k
Native frames: (J=compiled Java code, j=interpreted, Vv=VM code, C=native code)
C  [librocksdbjni3874548073448578178.so+0x37b79b]  rocksdb::DBImpl::GetIntProperty(rocksdb::ColumnFamilyHandle*, rocksdb::Slice const&, unsigned long*)+0x2b
C  [librocksdbjni3874548073448578178.so+0x291bb1]  Java_org_rocksdb_RocksDB_getLongProperty+0x171
J 7450  org.rocksdb.RocksDB.getLongProperty(JJLjava/lang/String;I)J (0 bytes) @ 0x00007f1495901a0d [0x00007f1495901920+0x00000000000000ed]
J 9392 c2 org.apache.bookkeeper.bookie.storage.ldb.EntryLocationIndexStats$1.getSample()Ljava/lang/Number; (5 bytes) @ 0x00007f1495344cac [0x00007f1495344be0+0x00000000000000cc]
J 6119 c2 org.apache.bookkeeper.stats.prometheus.SimpleGauge.getSample()Ljava/lang/Number; (10 bytes) @ 0x00007f1495829f84 [0x00007f1495829f20+0x0000000000000064]
J 5356 c2 org.apache.bookkeeper.stats.prometheus.PrometheusMetricsProvider$$Lambda$310+0x0000000800fa6460.accept(Ljava/lang/Object;Ljava/lang/Object;)V (20 bytes) @ 0x00007f14957245f4 [0x0000
7f14957234a0+0x0000000000001154]
J 6418 c2 java.util.concurrent.ConcurrentHashMap.forEach(Ljava/util/function/BiConsumer;)V java.base@17.0.5 (65 bytes) @ 0x00007f149554f0d8 [0x00007f149554ee40+0x0000000000000298]
[...]

PR #4243, opened 2024-03-24
Check rocksdb closed before operating

AnonHxy:

### Motivation

Fix https://github.com/apache/bookkeeper/issues/4238

From the issue above we can see that the core dump could happen when operate Rocksdb after it has been closed.  So we  should check  if closed before operating rocksdb.

### Changes

Checking if closed before operation RocksDB.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageRocksDB.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageRocksDB.java
@@ -39,7 +39,6 @@ import java.util.List;
 import java.util.Map.Entry;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
-
 import org.apache.bookkeeper.bookie.storage.ldb.KeyValueStorageFactory.DbConfigType;
 import org.apache.bookkeeper.conf.ServerConfiguration;
 import org.rocksdb.BlockBasedTableConfig;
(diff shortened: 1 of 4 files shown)


dlg99 (approved):
LGTM

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/KeyValueStorageRocksDB.java
```diff
@@ -285,41 +290,57 @@ private RocksDB initializeRocksDBWithBookieConf(String basePath, String subPath,
...
+            optionDontCache.close();
+            emptyBatch.close();
+        } finally {
+            writeLock.unlock();
+        }
     }
 
     @Override
     public void put(byte[] key, byte[] value) throws IOException {
+        readLock.lock();
```
merlimat, line 315: Acquiring the read lock for each put/get operation might be quite expensive, since the lock has to maintain the state of threads and the order they try to acquire the lock. 

We could use a different approach, like a reference counter on the entire object. When everyone is done using it, then rocksdb is really closed

lhotari: @AnonHxy It seems that a StampedLock could be sufficient here since there doesn't seem to be a need for re-entrancy of the lock. @merlimat would using a StampedLock address your concern?

lhotari: There's an alternative PR #4581 which simply handles `close` and `count`

hangc0276:
Since https://github.com/apache/bookkeeper/pull/4581 has fixed this issue, close this PR. @AnonHxy If you encountered this issue again, feel free the reopen it.

PR #4248, opened 2024-03-27
dependency: bump slf4j log4j2 version

shoothzj:
### Motivation

Update slf4j to 2.x. This is a backward-compatible change. 

> Log4j just update minor version. Logging configuration files don't need to change.

> SLF4J API version 2.0.0 requires Java 8 and introduces a backward-compatible fluent logging API. By backward-compatible, we mean that existing logging frameworks do not have to be changed in order for the user to benefit from the fluent logging API.

### Modifications

- Upgrade slf4j from 1.7.32 to 2.0.12
- Use log4j-slf4j2-impl instead log4j-slf4j-impl

See also: https://www.slf4j.org/manual.html#fluent


shoothzj:
@nicoloboschi @eolivelli PTAL :)

eolivelli (changes requested):
It is this a breaking change for the users ?
can you please add some description ?

do the user have to update their logging configuration files ?

shoothzj:
@eolivelli It's not a breaking change. I have updated descriptions. PTAL

eolivelli (approved):
Lgtm

Thanks

PR #4259, opened 2024-04-02
[improve] Prevent memory copy when read entry

dao-jun:
### Motivation

Prevent memory copy when read entry.

### Changes

> ---
>
> ---
> quickly and easily:
>
>
>
> Otherwise:
> 
> 
> ---

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadEntryProcessorV3.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadEntryProcessorV3.java
@@ -18,7 +18,7 @@
 package org.apache.bookkeeper.proto;
 
 import com.google.common.base.Stopwatch;
-import com.google.protobuf.ByteString;
+import com.google.protobuf.UnsafeByteOperations;
 import io.netty.buffer.ByteBuf;
 import io.netty.channel.Channel;
 import io.netty.util.ReferenceCountUtil;
@@ -51,6 +51,8 @@ class ReadEntryProcessorV3 extends PacketProcessorBaseV3 {
     protected final ReadRequest readRequest;
     protected final long ledgerId;
     protected final long entryId;
+
+    private volatile ByteBuf body;
 
     // Stats
     protected final OpStatsLogger readStats;
@@ -92,8 +94,6 @@ class ReadEntryProcessorV3 extends PacketProcessorBaseV3 {
     /**
      * Handle read result for fence read.
      *
-     * @param entryBody
-     *          read result
      * @param readResponseBuilder
      *          read response builder
      * @param entryId
@@ -102,7 +102,6 @@ class ReadEntryProcessorV3 extends PacketProcessorBaseV3 {
      *          timer for the read request
      */
     protected void handleReadResultForFenceRead(
-        final ByteBuf entryBody,
         final ReadResponse.Builder readResponseBuilder,
         final long entryId,
         final Stopwatch startTimeSw) {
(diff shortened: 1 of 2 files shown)


dao-jun:
@lhotari @eolivelli @hangc0276 @zymap Could you please take a look?

shoothzj:
I am not familiar with netty reference count. I have two questions.
1. why we need `onSendResponseFinished`? Can we just change to `UnsafeByteOperations.unsafeWrap(entryBody.nioBuffer())`? It may leak memory?
2. I think extract a class field is a little hard to maintenance, can we still pass it by method?

dao-jun:
> I am not familiar with netty reference count. I have two questions.
> 
> 1. why we need `onSendResponseFinished`? Can we just change to `UnsafeByteOperations.unsafeWrap(entryBody.nioBuffer())`? It may leak memory?
> 2. I think extract a class field is a little hard to maintenance, can we still pass it by method?

@shoothzj 
1. `UnsafeByteOperations` is protobuf‘s API, it will not release the memory, so we have to release it manually.
2. Pass `body` by method will lead to more changes, this is the most simple way.

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadEntryProcessorV3.java
```diff
@@ -170,25 +169,21 @@ protected ReadResponse readEntry(ReadResponse.Builder readResponseBuilder,
...
-                } else {
-                    long knownLAC = requestProcessor.getBookie().readLastAddConfirmed(ledgerId);
-                    readResponseBuilder.setMaxLAC(knownLAC);
-                }
-                registerSuccessfulEvent(readStats, startTimeSw);
-                readResponseBuilder.setStatus(StatusCode.EOK);
-                return readResponseBuilder.build();
-            } finally {
-                ReferenceCountUtil.release(entryBody);
+            readResponseBuilder.setBody(UnsafeByteOperations.unsafeWrap(entryBody.nioBuffer()));
```
lhotari, line 177: Calling `.nioBuffer()` might result in memory copies when the buffer is a composite buffer.
Something in this direction would result in zero copies:
https://github.com/apache/bookkeeper/pull/4293/files#diff-7c7c625da2136a47ce06a1753046baa3e3aeea006ef537b3fb6efd34d25000ddR851-R872

dao-jun: @lhotari Thanks for reminding, fixed

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadEntryProcessorV3.java
```diff
@@ -350,6 +344,43 @@ protected void registerEvent(boolean failed, OpStatsLogger statsLogger, Stopwatc
         }
     }
 
+    ByteString parseToByteString(ByteBuf entryBody) {
```
lhotari, line 347: The `parse` naming isn't optimal. A better name is `unsafeWrapAsByteString` etc. It's also worth adding a comment that the lifecycle of the ByteString continues to be managed with the ByteBuf lifecycle (retain/release).

bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadEntryProcessorV3.java
```diff
@@ -350,6 +344,43 @@ protected void registerEvent(boolean failed, OpStatsLogger statsLogger, Stopwatc
...
+        ByteString aggregated = null;
+        ByteString piece;
```
lhotari, line 349: Having this variable introduced at this level isn't needed.

dao-jun: makes sense to me, fixed

liangyepianzhou (approved):

PR #4266, opened 2024-04-03
Use vertx blockingHandlers to run Bookkeeper http handlers which could be blocking

lhotari:
### Motivation

- The http handler implementations in Bookkeeper aren't necessarily non-blocking. That's why they should be executed on the blocking thread pool in vertx.

### Changes

Executing on blocking thread pool in vertx can be achieved by registering the handlers using the `blockingHandler` method instead of the `handler` method.


dao-jun (approved):

Ghatage (approved):

hangc0276 (approved):

shoothzj (approved):

dragonls:
May I ask that why change http handlers to bloking handlers?

In our case, we are using `/api/v1/ledger/list` to get the ledgers of bookie, which can run more than 30 seconds due to lots of ledgers. So if we call `/api/v1/ledger/list`, the bookie can not response the `/metrics` in time, which cause the prometheus can not get the metrics of this bookie correctly.

PR #4282, opened 2024-04-11
chore: sort pom property version by alphabet

shoothzj:
### Motivation
sort pom property version by alphabet for better maintenance


shoothzj:
@nicoloboschi @eolivelli @zhaijack @dlg99 @hangc0276 PTAL

nicoloboschi (changes requested):
what is the purpose of this change ? 

I don't see any advantage

shoothzj:
@nicoloboschi I think alphabetically sorted property versions can be easier to read and understand. And the code was essentially sorted alphabetically before, and I only made a few changes to make it orderly.

PR #4278, opened 2024-04-11
fix: No need to trigger ensembleChangeLoop if all failed bookies are not in current ensemble

M1eyu2018:
Descriptions of the changes in this PR:

No need to trigger ensembleChangeLoop if all failed bookies are not in current ensemble.

### Motivation

Fix the bug issue.
https://github.com/apache/bookkeeper/issues/4261

### Changes

No need to trigger ensembleChangeLoop if all failed bookies are not in current ensemble.

Master Issue: #<master-issue-number>

> ---
> In order to uphold a high standard for quality for code contributions, Apache BookKeeper runs various precommit
> checks for pull requests. A pull request can only be merged when it passes precommit checks.
>
> ---
> Be sure to do all of the following to help us incorporate your contribution
> quickly and easily:
>
> If this PR is a BookKeeper Proposal (BP):
>
> - [ ] Make sure the PR title is formatted like:
>     `<BP-#>: Description of bookkeeper proposal`
>     `e.g. BP-1: 64 bits ledger is support`
> - [ ] Attach the master issue link in the description of this PR.
> - [ ] Attach the google doc link if the BP is written in Google Doc.
>
> Otherwise:
> 
> - [ ] Make sure the PR title is formatted like:
>     `<Issue #>: Description of pull request`
>     `e.g. Issue 123: Description ...`
> - [ ] Make sure tests pass via `mvn clean apache-rat:check install spotbugs:check`.
> - [ ] Replace `<Issue #>` in the title with the actual Issue number.
> 
> ---

PR #4285, opened 2024-04-11
Fix the completionObjects leak problem.

horizonzy:
Fixes #4278, you need to go through #4278 for the context.

When a connection is broken, it will trigger PerChannelBookieClient#channelInactive. https://github.com/apache/bookkeeper/blob/8516d0a0f1dbd352d3ef6455c2d41c94661ce1ce/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java#L1292-L1312

Point 1. line_1303, it will drain all completionObjects, and complete with BookieHandleNotAvailableException.

Point 2. line_1310, set PerChannelBookieClient#channel = null.

When PerChannelBookieClient#addEntry. https://github.com/apache/bookkeeper/blob/8516d0a0f1dbd352d3ef6455c2d41c94661ce1ce/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java#L776-L854

Point 3. line_840, it will put  completionKeyValue to completionObjects.

Point 4. line_852, if the channel is not null, invoke writeAndFlush.

There is a race condition between PerChannelBookieClient#channelInactive and PerChannelBookieClient#addEntry.

There is the timeline.

Point 1 -> Point 3 -> Point 4 -> Point 2.

It will write and flush the AddRequest to the netty channel. In the bookkeeper, there is a weakness in PerChannelBookieClient#writeAndFlush. https://github.com/apache/bookkeeper/blob/8516d0a0f1dbd352d3ef6455c2d41c94661ce1ce/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java#L1176-L1279

At line_1230, we define a promise for the netty write and flush, if the write and flush failed, we only record the metrics at line_1211, not remove the completionKey from completionObjects. The completionKey will leak in the completionObjects.
If the PerChannelBookieClient#addEntryTimeoutNanos is disabled, the timeoutCheck won't work, so the completionKey exists in the completionObjects forever.

In #4278, the PerChannelBookieClient#addEntryTimeoutNanos is enabled, So you will see that 5 seconds(default timeout) passed after bk1 died, the step 9 trigger PendingAddOp timeout, then cause the issue in #4278.

So, as long as we remove the completionKey from completionObjects when write and flush AddRequest failed, we can solve the problem.


bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PerChannelBookieClient.java
```diff
@@ -1209,6 +1209,7 @@ private void writeAndFlush(final Channel channel,
                     }
                 } else {
                     nettyOpLogger.registerFailedEvent(MathUtils.elapsedNanos(startTime), TimeUnit.NANOSECONDS);
+                    errorOut(key);
```
wenbingshen, line 1212: > 
> In #4278, the PerChannelBookieClient#addEntryTimeoutNanos is enabled, So you will see that 5 seconds(default timeout) passed after bk1 died, the step 9 trigger PendingAddOp timeout, then cause the issue in #4278.
> 

I have a question, since we have a default request to detect timeout every 5 seconds and remove the timed `CompletionKey` from `completionObjects`, even if the write fails, will the timeout detection task not remove the key?

horizonzy: Sure, the timeout detection will remove the key.

hangc0276 (approved):

zymap (approved):

shoothzj (approved):

shoothzj:
Nice Catch!

PR #4288, opened 2024-04-16
Fix some obvious typos

hudeqi:
While reading and learning the code, I found some obvious typos and fixed them.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
@@ -1098,7 +1098,7 @@ public abstract class AbstractConfiguration<T extends AbstractConfiguration>
      *            the concurrency level to use for the allocator pool
      * @return configuration object.
      */
-    public T setAllocatorPoolingConcurrenncy(int concurrency) {
+    public T setAllocatorPoolingConcurrency(int concurrency) {
         this.setProperty(ALLOCATOR_POOLING_POLICY, concurrency);
         return getThis();
     }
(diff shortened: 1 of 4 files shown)


shoothzj (approved):

shoothzj:
I am not sure if we can directly rename `setAllocatorPoolingConcurrenncy` to `setAllocatorPoolingConcurrency`. Should we mark it deprecated and remove in next minor versions? cc @nicoloboschi @zhaijack @eolivelli @merlimat

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
```diff
@@ -1098,7 +1098,7 @@ public int getAllocatorPoolingConcurrency() {
      *            the concurrency level to use for the allocator pool
      * @return configuration object.
      */
-    public T setAllocatorPoolingConcurrenncy(int concurrency) {
+    public T setAllocatorPoolingConcurrency(int concurrency) {
```
eolivelli, line 1101: This is a public API change and it will break users.

In order to fix this property we have to:
- add the new method
- mark the existing method as Deprecated
- wait for a minor (4.18) version release
-  send a new a patch that removes the method

eolivelli (changes requested):

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/AbstractConfiguration.java
hudeqi: @eolivelli Thanks for your review and suggestions, I have updated and I will follow the removal for old method.

shoothzj:
@hudeqi I will merge this once check passed. Thanks for your interests in the Apache BookKeeper project, Looking forward to your future contributions.

shoothzj:
@hudeqi Well, I think we need a more precise PR title and content to description this change. People should notice this change not only as typos.

shoothzj:
@hudeqi Could you please also update the PR content? thanks

hudeqi:
> @hudeqi Could you please also update the PR content? thanks

updated.

hudeqi:
> > @hudeqi Could you please also update the PR content? thanks
> 
> updated. @shoothzj

hudeqi:
@hangc0276 hi, I see that version 4.17 has been released recently. According to the previous discussion on this PR, should I remove the `setAllocatorPoolingConcurrency` method now, or wait until after the release of version 4.18 to remove it?

shoothzj:
@hudeqi I think we can remove it in master branch, it can be released in 4.18.0

PR #4299, opened 2024-04-19
improve: use SLASHSTAR_STYLE for licenses in Java files

hezhangjian:
### Motivation

When reformatting Java files, JAVADOC_STYLE cause the license header be
treated as doc comments and it will be affected from:

```java
/**
* ...
* with the License. You may obtain a copy of the License at
*
* http://www.apache.org/licenses/LICENSE-2.0
*
* Unless required by applicable law or agreed to in writing,
* ...
*/
```

to

```java
/**
* ...
* with the License. You may obtain a copy of the License at
* <p>
* http://www.apache.org/licenses/LICENSE-2.0
* <p>
* Unless required by applicable law or agreed to in writing,
* ...
*/
```

Also can see https://lists.apache.org/thread/gwy3ycgsq24r784mk3yv6cbppmbfxgjd

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/AuditorBookieCheckTask.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/AuditorBookieCheckTask.java
@@ -1,4 +1,4 @@
-/**
+/*
  * Licensed to the Apache Software Foundation (ASF) under one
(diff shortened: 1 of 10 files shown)


shoothzj:
@nicoloboschi @eolivelli @hangc0276 @merlimat @zhaijack @dlg99 PTAL

jiazhai (approved):

PR #4300, opened 2024-04-19
 Fix ThreadRegistry#register behavior to ensure correct Prom metrics 

michaeljmarshall:

There is an issue in the `ThreadRegistry` that causes incomplete Prometheus metrics reporting. It is trivial to reproduce:

1. Start bookies with multiple journal directories.
2. Start multiple clients writing to the bookies.
3. Observe that the `bookkeeper_server_ADD_ENTRY_count` does not increment at a similar rate to the journal's add entry rate.

The bug can be understood abstractly by considering the following:

1. The `threadPoolMap` in `ThreadRegistry` goes from thread id to `ThreadPoolThread`.
2. The `ThreadPoolThread` is defined by the thread pool and the configured id.
3. Every time we set the id for step 2, we use `0`.
4. When we have multiple journals, we call the same `ThreadRegistry.register(journalThreadName, 0);` method multiple times.
5. The `threadPoolMap` ends up with multiple identical values for different keys.
6. `ThreadScopedDataSketchesStatsLogger#getStatsLogger` stores the identical values in the `provider.opStats` map and ultimately overwrites map values, which leads to an under-reporting of metrics.

### Motivation

Correct bookkeeper metrics reporting by ensuring that threads are properly registered in the `ThreadRegistry`.

### Changes

I replaced the static map in `ThreadRegistry#register` with a static `ThreadLocal` instance. While the original work in #2839, which was part of #2835, specifically indicated it did not want to use thread local storage (TLS) due to its complexity, I think it is worth it because of the added flexibility. Specifically, it is much simpler to handle cleanup of unnecessary metadata by using thread locals. Note that the thread id, previously used as the key in the static map, is not unique to the thread after the thread goes away, which poses problems if any threads recycle. While BK doesn't generally recycle these threads, the test framework does, so it helps to make tests pass while still being able to make useful assertions about state.

An interesting design detail that I noticed and did not change is documented in the ThreadRegistry. Specifically, it is that threads that go away will continue to report metrics. I think this is fine, but it's worth calling out.

### DIFF
--- a/stats/bookkeeper-stats-api/src/main/java/org/apache/bookkeeper/stats/ThreadRegistry.java
+++ b/stats/bookkeeper-stats-api/src/main/java/org/apache/bookkeeper/stats/ThreadRegistry.java
@@ -40,7 +40,8 @@ public class ThreadRegistry {
      */
     public static void register(String threadPool, int threadPoolThread, long threadId) {
         ThreadPoolThread tpt = new ThreadPoolThread(threadPool, threadPoolThread, threadId);
-        threadPoolMap.put(threadId, tpt);
+        ThreadPoolThread previous = threadPoolMap.put(threadId, tpt);
+        assert previous == null : "Thread " + threadId + " was already registered in thread pool " + previous.threadPool + " as thread " + previous.ordinal;
     }
 
     /*
--- a/stats/bookkeeper-stats-api/src/main/java/org/apache/bookkeeper/stats/ThreadRegistry.java
+++ b/stats/bookkeeper-stats-api/src/main/java/org/apache/bookkeeper/stats/ThreadRegistry.java
@@ -18,6 +18,9 @@ package org.apache.bookkeeper.stats;
 
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 /**
  * For mapping thread ids to thread pools and threads within those pools
  * or just for lone named threads. Thread scoped metrics add labels to
@@ -25,6 +28,7 @@ import java.util.concurrent.ConcurrentMap;
  * For flexibility, this registry is not based on TLS.
  */
 public class ThreadRegistry {
+    private static Logger logger = LoggerFactory.getLogger(ThreadRegistry.class);
     private static ConcurrentMap<Long, ThreadPoolThread> threadPoolMap = new ConcurrentHashMap<>();
     private static ConcurrentMap<String, Integer> threadPoolThreadMap = new ConcurrentHashMap<>();
 
@@ -34,8 +38,31 @@ public class ThreadRegistry {
     for the given thread pool.
     */
     public static void register(String threadPool) {
+        register(threadPool, false);
+    }
+
+    public static void register(String threadPool, boolean force) {
         Integer threadPoolThread = threadPoolThreadMap.compute(threadPool, (k, v) -> v == null ? 0 : v + 1);
+        if (force) {
+            threadPoolMap.remove(Thread.currentThread().getId());
+        }
         register(threadPool, threadPoolThread, Thread.currentThread().getId());
+    }
+
+    /**
+     * In some tests we run in the same thread activities that should
+     * run in different threads from different thread-pools
+     * this would trigger assertions to fail.
+     * This is a convenience method to work around such cases.
+     * This method shouldn't be used in production code.
+     */
+    public static void forceClearRegistrationForTests(long threadId) {
+        threadPoolMap.compute(threadId, (id, value) -> {
+            if (value !=  null) {
+                logger.info("Forcibly clearing registry entry {} for thread id {}", value, id);
+            }
+           return null;
+        });
     }
 
     /*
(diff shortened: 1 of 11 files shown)


stats/bookkeeper-stats-api/src/main/java/org/apache/bookkeeper/stats/ThreadRegistry.java
```diff
@@ -40,14 +51,16 @@ public static void register(String threadPool, int threadPoolThread) {
      */
     public static void register(String threadPool, int threadPoolThread, long threadId) {
         ThreadPoolThread tpt = new ThreadPoolThread(threadPool, threadPoolThread, threadId);
-        threadPoolMap.put(threadId, tpt);
+        ThreadPoolThread previous = threadPoolMap.put(threadId, tpt);
+        assert previous == null : "Thread " + threadId + " was already registered in thread pool " + previous.threadPool + " as thread " + previous.ordinal;
```
eolivelli, line 55: I am not sure that assertions are enabled in BK, maybe an old style "if (....) throw new IllegalStateException is better here

eolivelli (approved):
Great Catch

This is a bad bug affecting also 4.16. we will discuss cutting an hotfix release also for that branch

stats/bookkeeper-stats-api/src/main/java/org/apache/bookkeeper/stats/ThreadRegistry.java
michaeljmarshall: Sure, I'll update that now.

lhotari:
Great catch @michaeljmarshall

lhotari:
checkstyle error
```
Error:  /home/runner/work/bookkeeper/bookkeeper/stats/bookkeeper-stats-api/src/main/java/org/apache/bookkeeper/stats/ThreadRegistry.java:55: Line is longer than 120 characters (found 156). [LineLength]
Audit done.
[INFO] There is 1 error reported by Checkstyle 9.3 with buildtools/src/main/resources/bookkeeper/checkstyle.xml ruleset.
Error:  src/main/java/org/apache/bookkeeper/stats/ThreadRegistry.java:[55] (sizes) LineLength: Line is longer than 120 characters (found 156).
```

stats/bookkeeper-stats-api/src/main/java/org/apache/bookkeeper/stats/ThreadRegistry.java
shoothzj: @eolivelli assert is not encouraged by ivankelly #995

shoothzj (approved):
Nice Catch.

nicoloboschi (approved):
+1 Great work

eolivelli:
in order to let the tests pass I suggest to remove the assertions and simply use "log.error". In production you will see ERRORs.
The problem is that in tests we launch multiple bookies in the same JVM

nicoloboschi (commented):
@michaeljmarshall tests are failing because most of them spin up N bookies in the same JVM, and since ThreadRegistry is static they overlap each other.

I suggest we replace the assertion with a log.error

shoothzj:
@eolivelli @nicoloboschi Agreed, I think we can have an issue to track subsequent refactorings to non-static.

dlg99 (approved):

stats/bookkeeper-stats-providers/prometheus-metrics-provider/src/main/java/org/apache/bookkeeper/stats/prometheus/ThreadScopedDataSketchesStatsLogger.java
```diff
@@ -96,15 +96,25 @@ private DataSketchesOpStatsLogger getStatsLogger() {
             ThreadRegistry.ThreadPoolThread tpt = ThreadRegistry.get();
             if (tpt == null) {
                 statsLoggers.set(defaultStatsLogger);
-                provider.opStats.put(new ScopeContext(scopeContext.getScope(), originalLabels), defaultStatsLogger);
+                DataSketchesOpStatsLogger previous = provider.opStats
+                        .put(new ScopeContext(scopeContext.getScope(), originalLabels), defaultStatsLogger);
+                // If we overwrite a logger, metrics will not be collected correctly
+                if (previous != null && previous != defaultStatsLogger) {
+                    throw new IllegalStateException("Invalid state. Overwrote a stats logger.");
+                }
```
eolivelli, line 104: I would add a log line about this case, because we should notice when something is using the "default" statslogger, its matrics don't make much sense and they won't probably meaningful for administrators.
So it is fine to not break but we should log a warning

stats/bookkeeper-stats-providers/prometheus-metrics-provider/src/main/java/org/apache/bookkeeper/stats/prometheus/ThreadScopedLongAdderCounter.java
```diff
@@ -92,15 +92,25 @@ private LongAdderCounter getCounter() {
...
+                threadScopedLabels.put("thread", String.valueOf(tpt.getOrdinal()));
 
-                counter.initializeThread(threadScopedlabels);
-                provider.counters.put(new ScopeContext(scopeContext.getScope(), threadScopedlabels), counter);
+                counter.initializeThread(threadScopedLabels);
+                LongAdderCounter previous = provider.counters
+                        .put(new ScopeContext(scopeContext.getScope(), threadScopedLabels), counter);
+                // If we overwrite a counter, metrics will not be collected correctly
+                if (previous != null) {
+                    throw new IllegalStateException("Invalid state. Overwrote a counter.");
```
eolivelli, line 112: what about reporting the labels of what is being overwritten ? it will help trhoubleshooting

eolivelli (approved):
LGTM

I added two minor suggestions

stats/bookkeeper-stats-providers/prometheus-metrics-provider/src/main/java/org/apache/bookkeeper/stats/prometheus/ThreadScopedDataSketchesStatsLogger.java
michaeljmarshall: > its matrics don't make much sense

Can just remove these labels? It's a slight breaking change, but they seem unnecessary.

michaeljmarshall: Also, many metrics were removed in the 4.16 release, implying there isn't a strict sense of compatibility here.

hangc0276:
Hi @michaeljmarshall Could you add a test to protect this change?

PR #4305, opened 2024-04-22
Prevent double flush due to race in SingleDirectoryDbLedgerStorage

michaeljmarshall:
Disclaimer: I am not very familiar with this code, so please review carefully.

### Motivation

We use the `isFlushOngoing` and the `hasFlushBeenTriggered` atomic booleans to prevent concurrent flushes. See:

https://github.com/apache/bookkeeper/blob/2986e04f1057ba3da7c549b1479c187c9ba167f2/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java#L514-L530

By my reading of the code, it seems there is a brief period of time where the calling code will think it is valid to flush, even though there is a flush in progress.

### Changes

In the `SingleDirectoryDbLedgerStorage#swapWriteCache` method modified by this PR, I propose a change to set `isFlushOngoing` to true before setting `hasFlushBeenTriggered` to false.

Master Issue: none


hezhangjian (approved):
LGTM

dlg99 (approved):

dlg99:
closing to re-trigger CI

PR #4306, opened 2024-04-22
SingleDirectoryDbLedgerStorage skip optimistic cache put sometimes

michaeljmarshall:
### Motivation

When tryOptimisticRead returns 0, the lock is acquired by another thread and we will fail the validation step. We can therefore skip the eager cache insertion.

Relevant javadocs from `StampedLock`:

```java
    /**
     * Returns a stamp that can later be validated, or zero
     * if exclusively locked.
     *
     * @return a valid optimistic read stamp, or zero if exclusively locked
     */
    public long tryOptimisticRead() {
        long s;
        return (((s = state) & WBIT) == 0L) ? (s & SBITS) : 0L;
    }

    /**
     * Returns true if the lock has not been exclusively acquired
     * since issuance of the given stamp. Always returns false if the
     * stamp is zero. Always returns true if the stamp represents a
     * currently held lock. Invoking this method with a value not
     * obtained from {@link #tryOptimisticRead} or a locking method
     * for this lock has no defined effect or result.
     *
     * @param stamp a stamp
     * @return {@code true} if the lock has not been exclusively acquired
     * since issuance of the given stamp; else false
     */
    public boolean validate(long stamp) {
        VarHandle.acquireFence();
        return (stamp & SBITS) == (state & SBITS);
    }
```

### Changes

Check for `stamp == 0` since it is a special case in the `StampedLock` class.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
@@ -480,7 +480,7 @@ public class SingleDirectoryDbLedgerStorage implements CompactableLedgerStorage 
         long stamp = writeCacheRotationLock.tryOptimisticRead();
         boolean inserted = false;
 
-        // If the stamp is 0, the lock is already acquired, validation will fail, and we can skip this put.
+        // If the stamp is 0, the lock was exclusively acquired, validation will fail, and we can skip this put.
         if (stamp != 0) {
             inserted = writeCache.put(ledgerId, entryId, entry);
         }
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
@@ -480,8 +480,12 @@ public class SingleDirectoryDbLedgerStorage implements CompactableLedgerStorage 
         long stamp = writeCacheRotationLock.tryOptimisticRead();
         boolean inserted = false;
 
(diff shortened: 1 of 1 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
```diff
@@ -480,8 +480,12 @@ public long addEntry(ByteBuf entry) throws IOException, BookieException {
         long stamp = writeCacheRotationLock.tryOptimisticRead();
         boolean inserted = false;
 
-        inserted = writeCache.put(ledgerId, entryId, entry);
-        if (!writeCacheRotationLock.validate(stamp)) {
+        // If the stamp is 0, the lock was exclusively acquired, validation will fail, and we can skip this put.
```
michaeljmarshall, line 483: I looked closer, and it's only 0 when the lock is exclusively held, which only happens when we call `swapWriteCache()`. That does not seem frequent enough to bother with this special case, but I will wait for others to review to confirm.

hezhangjian (approved):

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
lhotari: It seems like there's invalid use of the StampedLock before this PR. Read operations are used for write operations. That's not supported by StampedLock. I don't see how that could be useful.

lhotari: I think it's better to just use the StampedLock API to do that validation. It's not an expensive operation.

dlg99:
closing to re-trigger CI

dlg99 (approved):

PR #4307, opened 2024-04-22
Prevent race in SingleDirectoryDbLedgerStorage metrics collection

michaeljmarshall:
### Motivation

We swap the `writeCache` and the `writeCacheBeingFlushed` on flush. Therefore, it is not safe to call each to get the size/count metrics. Instead, if we expose the underlying accumulators, we can safely get the stats.

### Changes

Expose the `AtomicLong` and the `LongAdder` objects in the `WriteCache` for safer metrics reporting.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
@@ -46,6 +46,8 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.LongAdder;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.concurrent.locks.StampedLock;
 import org.apache.bookkeeper.bookie.Bookie;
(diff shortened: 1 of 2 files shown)


shoothzj:
How do you guarantee they won't swap during L214-L217? It seems problem is not solved.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
```diff
@@ -207,10 +209,17 @@ public SingleDirectoryDbLedgerStorage(ServerConfiguration conf, LedgerManager le
         gcThread = new GarbageCollectorThread(conf,
                 ledgerManager, ledgerDirsManager, this, entryLogger, ledgerIndexDirStatsLogger);
 
+        // Because the writeCache and the writeCacheBeingFlushed are swapped during flushes, it is
+        // safer to acquire the underlying cache size and count references and then use those for metrics.
+        final AtomicLong finalWriteCacheSize = writeCache.cacheSize;
+        final AtomicLong finalWriteCacheBeingFlushedSize = writeCacheBeingFlushed.cacheSize;
+        final LongAdder finalWriteCacheCounter = writeCache.cacheCount;
+        final LongAdder finalWriteCacheBeingFlushedCounter = writeCacheBeingFlushed.cacheCount;
+
```
dlg99, line 218: I don't think this is right.
Lambdas in the DbLedgerStorageStats() constructor used to provide a way to access to metrics of current writeCache/.. instance.
With this change it will provide metrics for the first used instance, one before the very first swap, so the stats will stop changing after the cache swaps couple of times.

dlg99 (changes requested):
See my comment.
Overall, I don't think there is a problem.

michaeljmarshall:
> How do you guarantee they won't swap during L214-L217? It seems problem is not solved.

We're in the constructor. I don't think anything could have access to trigger the swap yet.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
michaeljmarshall: Notice that we're aggregating the `writeCache` and `writeCacheBeingFlushed` size and count. Given that the swap is just to replace references for `writeCache` and `writeCacheBeingFlushed`, I am not sure I understand what is wrong here. In this PR, we grab the references once and remove the possibility of `writeCache.size() + writeCacheBeingFlushed.size()` double counting the same value from `writeCache` in the event the caches are swapped after `writeCache.size()` and before `writeCacheBeingFlushed.size()`

liangyepianzhou: +1, I don't t think this is right. 

The original logic used dynamic binding, where each lambda invocation dynamically retrieves the current instances of `writeCache` and `writeCacheBeingFlushed`. The modified approach statically binds by capturing the initial instance references of `writeCache` and `writeCacheBeingFlushed` in the constructor. This error severely distorts the storage monitoring data.

PR #4310, opened 2024-04-23
test: use junit assert instead of java assert statement in test

shoothzj:
### Motivation

See #995

### Changes

1. use junit assert instead of java assert statement in test
2. delete a unused util class also uses assert statement


shoothzj:
@dlg99 @eolivelli @hangc0276 @merlimat @nicoloboschi @zhaijack PTAL

shoothzj:
@dlg99 @eolivelli @hangc0276 @merlimat @nicoloboschi @zhaijack PTAL

bookkeeper-server/src/main/java/org/apache/bookkeeper/util/DaemonThreadFactory.java
jiazhai, line 1: Seems this change is not related with the target. Is this class not used anywhere?  why this is removed?

shoothzj: @jiazhai This class is not used anywhere even in pulsar project. And this class also uses assert statement.

jiazhai (approved):

hangc0276:
This PR can only be released in major releases because we removed DaemonThreadFactory in the util package and it may be used in other projects.

PR #4314, opened 2024-04-25
[improve] different bookies do not trigger gcOverreplicatedLedger at the same time

TakaHiR07:
### Motivation

In the `gcOverReplicatedLedger` function, there is a config `gcOverreplicatedLedgerMaxConcurrentRequests` to control the zk concurrent request of one bookie, which aim to avoid too much pressure on zk.

However, if we have so much bookies, and different bookies start at similar time, different bookies would trigger gcOverreplicatedLedger at the same time. This make the request pressure on zk become `maxConcurrentRequests * bookieNums`. 

I think we can improve this process, make different bookies not trigger gcOverreplicatedLedger at the same time. So that we can avoid `gcOverreplicatedLedgerMaxConcurrentRequests` be a very small value.

### Changes

define lastOverReplicatedLedgerGcTimeMillis based on `the bookie start time + randomly time` or `the bookie start time - randomly time`


shoothzj:
Thanks for your contribution. I have triggered the CI. BTW, how many bookies do you maintain?

TakaHiR07:
> Thanks for your contribution. I have triggered the CI. BTW, how many bookies do you maintain?

The largest cluster contains 400+ bookies.
image

jiazhai (approved):

hangc0276 (approved):

TakaHiR07:
Have fixed the test. Can you trigger CI again, thx. @shoothzj @hangc0276 @jiazhai

PR #4325, opened 2024-04-30
fix: bookie http endpoint info always return 0.0.0.0

nicoloboschi:
### Motivation

If you set `httpServerHost`, the bookie http info endpoint doesn't expose the host correctly but it always return 0.0.0.0 

### Changes

Use `httpServerHost` configuration


shoothzj (approved):
nice catch

PR #4332, opened 2024-05-03
Fix improper negative key generation and update assertions in tests

### LINKED ISSUE #2077: ConcurrentLongHashSetTest.java: Math.abs does not always give a positive result (errorprone warning)
**BUG REPORT**

***Describe the bug***

When compiling the project using the ErrorProne compiler, the following warning is emitted.

[MathAbsoluteRandom] Math.abs does not always give a positive result. Please consider other methods for positive random numbers. (See https://errorprone.info/bugpattern/MathAbsoluteRandom)

It is related to code at

https://github.com/apache/bookkeeper/blob/28dee8464764b0edceeb31ce24424f0947b5789b/bookkeeper-server/src/test/java/org/apache/bookkeeper/util/collections/ConcurrentLongHashSetTest.java#L162-L171

***To Reproduce***

1. Install the ErrorProne IntelliJ plugin https://plugins.jetbrains.com/plugin/7349-error-prone-compiler-integration
2. Enable it in Project settings
2. Rebuild the project
3. See error in log

***Additional context***

```
jshell> import java.lang.Math;

jshell> Math.abs(Long.MIN_VALUE)
$5 ==> -9223372036854775808
```

One solution is to use the `java.util.concurrent.ThreadLocalRandom#nextLong` which has variant to return only positive numbers.

I had a brief look at the test and I don't see why it is necessary for the random number to be positive, to tell the truth.

shoothzj:
Fix #2077 

By the way, I have been tested, negative number will fail.

### Motivation

In the existing tests for `ConcurrentLongHashSet` and `ConcurrentLongLongHashMap`, the method `Math.abs(random.nextLong())` is used to generate keys. However, `Math.abs` can return a negative number if `Long.MIN_VALUE` is generated, which is problematic for tests that assume non-negative keys. Additionally, the assertions in the tests are using the older JUnit 4 style, which is less readable and not as flexible as JUnit 5.

### Changes

1. **Key Generation Fix**: Replaced `Math.abs(random.nextLong())` with `ThreadLocalRandom.current().nextLong(Long.MAX_VALUE)` to ensure that all generated keys are non-negative.
2. **Assertion Updates**: Updated all assertions in the `ConcurrentLongHashSetTest` and `ConcurrentLongLongHashMapTest` to use JUnit 5 style. This includes using `assertEquals`, `assertTrue`, `assertFalse`, and `assertNull` for better readability and consistency.

### DIFF
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/util/collections/ConcurrentLongHashSetTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/util/collections/ConcurrentLongHashSetTest.java
@@ -20,23 +20,24 @@
  */
 package org.apache.bookkeeper.util.collections;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.fail;
 
 import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
 import java.util.ArrayList;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


shoothzj:

zymap (approved):
LGTM

PR #4333, opened 2024-05-04
build: fix JAVA_HOME not properly set in docker image

hezhangjian:
### Motivation
`JAVA_HOME` env variable is not properly set.

I have tested `ARG TARGETARCH` works well, when we run `docker build`, it will automatically set the value.

error log below:
```
bookie3_1    | JAVA_HOME not set, using java from PATH. (/usr/bin/java)
bookie1_1    | JAVA_HOME not set, using java from PATH. (/usr/bin/java)
bookie3_1    | Unrecognized VM option 'PrintGCApplicationStoppedTime'
bookie3_1    | Error: Could not create the Java Virtual Machine.
```

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -20,6 +20,7 @@
 FROM ubuntu:22.04
 MAINTAINER Apache BookKeeper <dev@bookkeeper.apache.org>
 
+ARG TARGETARCH
 ARG BK_VERSION=4.15.1
 ARG DISTRO_NAME=bookkeeper-server-${BK_VERSION}-bin
 ARG DISTRO_URL=https://archive.apache.org/dist/bookkeeper/bookkeeper-${BK_VERSION}/${DISTRO_NAME}.tar.gz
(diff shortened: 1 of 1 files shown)


shoothzj:

docker/Dockerfile
```diff
@@ -61,13 +62,15 @@ RUN set -x \
...
     && rm -rf "$DISTRO_NAME.tar.gz" "$DISTRO_NAME.tar.gz.asc" "$DISTRO_NAME.tar.gz.sha512" \
-    && pip install zk-shell \
-    && JAVA_HOME=$(dirname $(dirname $(readlink -f $(which java)))) \
-    && echo networkaddress.cache.ttl=1 >> $JAVA_HOME/conf/security/java.security \
-    && echo networkaddress.cache.negative.ttl=1 >> $JAVA_HOME/conf/security/java.security
+    && pip install zk-shell
 
 WORKDIR /opt/bookkeeper
 
+ENV JAVA_HOME /usr/lib/jvm/java-17-openjdk-$TARGETARCH
```
zhiheng123, line 69: can we use ```ENV JAVA_HOME $(dirname $(dirname $(readlink -f $(which java)))) ``` instead of ```ENV JAVA_HOME 
 /usr/lib/jvm/java-17-openjdk-$TARGETARCH```

shoothzj: I have tested on my laptop with my personal image. It won't work.

```
└─[$] <> cat Dockerfile
FROM shoothzj/base:jdk17


ENV JAVA1_HOME $(dirname $(dirname $(readlink -f $(which java))))
```
```
docker run -it --rm test bash
root@f28cdaac85a6:/# echo $JAVA1_HOME
$(dirname $(dirname $(readlink -f $(which java))))
root@f28cdaac85a6:/#
exit
```

shoothzj:

jiazhai (approved):

docker/Dockerfile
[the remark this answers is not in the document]
shoothzj: @jiazhai Agreed, but now it's impossible to put shell command during ENV statement.

hangc0276: +1 for not hard-coding the path.
Can we merge these two changed command into the previous one? The new one will generate more layers in docker

hangc0276:
-1 for this change.

shoothzj:
@hangc0276 Could you please share more information about why against this change? :)

shoothzj:
>> +1 for not hard-coding the path.
>> Can we merge these two changed command into the previous one? The new one will generate more layers in docker

Thank you for raising the concern regarding the setting of the `JAVA_HOME` environment variable in our Dockerfile. I’d like to address the points raised and explain the reasoning behind the current approach.

#### 1. Setting `JAVA_HOME` with `ENV`
The suggestion to dynamically set the `JAVA_HOME` using commands like `dirname` and `readlink` within the `ENV` directive is not feasible. Docker does not support executing commands within the `ENV` directive. This limitation means we cannot use:

Like I commented in issue before, I have tested on my laptop with my personal image. It won't work.

```
└─[$] <> cat Dockerfile
FROM shoothzj/base:jdk17


ENV JAVA1_HOME $(dirname $(dirname $(readlink -f $(which java))))
```
```
docker run -it --rm test bash
root@f28cdaac85a6:/# echo $JAVA1_HOME
$(dirname $(dirname $(readlink -f $(which java))))
root@f28cdaac85a6:/#
exit
```

### 2. Impact of an Extra Layer

Regarding the concern about adding an extra layer in the Docker image, while it is ideal to minimize layers for efficiency, in this case, the impact is negligible. Our Dockerfile already includes several `ENV` directives, and one additional layer for correctly setting `JAVA_HOME` ensures the functionality is as expected without introducing complexity or potential errors.

PR #4334, opened 2024-05-04
chore: remove assert statement in LedgerDescriptor

shoothzj:
See #995 

### Motivation

Remove assert in LedgerDescriptor these are disabled at runtime and never works

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDescriptorImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDescriptorImpl.java
@@ -40,10 +40,10 @@ import org.slf4j.LoggerFactory;
 public class LedgerDescriptorImpl extends LedgerDescriptor {
     private static final Logger LOG = LoggerFactory.getLogger(LedgerDescriptorImpl.class);
     final LedgerStorage ledgerStorage;
-    private long ledgerId;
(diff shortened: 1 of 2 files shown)


shoothzj:

shoothzj:

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/LedgerDescriptorImpl.java
```diff
@@ -40,10 +40,10 @@
 public class LedgerDescriptorImpl extends LedgerDescriptor {
     private static final Logger LOG = LoggerFactory.getLogger(LedgerDescriptorImpl.class);
     final LedgerStorage ledgerStorage;
-    private long ledgerId;
+    private final long ledgerId;
```
jiazhai, line 43: why do we need these two changes?

shoothzj: @jiazhai When I was submitting this patch, I found this place could do some optimize, marking field final has two benefits:
- For people reading this code, clarify this field never change.
- For jvm, it could do some optimize since it's immutable.

I suggest we can include this in to this patch, it's not worth opening a new pr(I think).

cc @eolivelli @nicoloboschi @merlimat

jiazhai (approved):

PR #4342, opened 2024-05-06
[client] new API to check if Bookkeeper client is connected to metadata service

congbobo184:
# Motivation
when zookeeper disconnect, use bookkeeper api to create new ledger will fail. add a new api to check metadata driver is available, if not available, use current ledger continue and don't rollover ledger

## ** tip **
Although it may not solve all issues since it's not atomic, it will reduce the losses caused by zk unavailability.
# Modification
add a new api named isDriverMetadataServiceAvailable


bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/zk/ZKMetadataDriverBase.java
```diff
@@ -212,6 +217,10 @@ protected void initialize(AbstractConfiguration<?> conf,
                     .sessionTimeoutMs(conf.getZkTimeout())
                     .operationRetryPolicy(zkRetryPolicy)
                     .requestRateLimit(conf.getZkRequestRateLimit())
+                    .watchers(Collections.singleton(watchedEvent -> {
+                        log.info("Got ZK session watch event: {}", watchedEvent);
```
eolivelli, line 221: This is probably adding noise, we should move this to "debug" level

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/zk/ZKMetadataDriverBase.java
```diff
@@ -247,6 +256,19 @@ protected void initialize(AbstractConfiguration<?> conf,
             acls);
     }
 
+    private synchronized void handleState(Watcher.Event.KeeperState zkClientState) {
+        switch (zkClientState) {
+            case Expired:
+            case Disconnected:
+                this.metadataServiceAvailable = false;
+                break;
+
```
eolivelli, line 265: nit: remove line break

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/zk/ZKMetadataDriverBase.java
```diff
@@ -247,6 +256,19 @@ protected void initialize(AbstractConfiguration<?> conf,
             acls);
     }
 
+    private synchronized void handleState(Watcher.Event.KeeperState zkClientState) {
```
eolivelli, line 259: this is `synchronized` so we must add a 
synchronized boolean isMetadataServiceAvailable()

methodm otherwise we are not accessing the field properly

eolivelli (changes requested):

eolivelli (approved):
LGTM

in theory we should require a BP for this, but the new API is pretty straightforward and it is not a breaking change

@merlimat @dlg99 @hangc0276

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/zk/ZKMetadataDriverBase.java
shoothzj: @congbobo184 looks like line break is not removed?

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/zk/ZKMetadataDriverBase.java
```diff
@@ -64,6 +66,8 @@
 public class ZKMetadataDriverBase implements AutoCloseable {
 
     protected static final String SCHEME = "zk";
+
+    protected volatile boolean metadataServiceAvailable;
```
shoothzj, line 70: nits: add blank line after

hezhangjian (approved):
LGTM, I find some nits.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/BookKeeper.java
```diff
@@ -81,6 +81,13 @@ static BookKeeperBuilder newBuilder(final ClientConfiguration clientConfiguratio
      */
     CompletableFuture<LedgerMetadata> getLedgerMetadata(long ledgerId);
 
+    /**
+     * Return driver metadata service is available.
+     *
+     * @return the metadata service is available.
+     */
+    CompletableFuture<Boolean> isDriverMetadataServiceAvailable();
```
zymap, line 89: Add a default implementation to avoid breaking the interface

zymap (approved):

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/api/DriverMetadataServiceAvailableTest.java
```diff
@@ -0,0 +1,53 @@
...
+import org.apache.bookkeeper.conf.ClientConfiguration;
+import org.apache.bookkeeper.test.BookKeeperClusterTestCase;
+import org.awaitility.Awaitility;
+import org.junit.Test;
+
+/**
+ * Bookkeeper Client API driver metadata service available test.
+ */
+public class DriverMetadataServiceAvailableTest extends BookKeeperClusterTestCase {
+
```
shoothzj, line 34: nit: we only need one blank space here.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/BookKeeper.java
shoothzj: I doubt that add default implementation is a good idea. First, we don't guarantee/make ABI comptabile between minor releases; Second, People whole implement metadata driver should implement this.

cc @eolivelli @dlg99

congbobo184:
rerun failure checks

congbobo184:
rerun failure checks

hezhangjian (changes requested):
I think the default impl needs discuss, I have started a mail thread: https://lists.apache.org/thread/fk02bz4lggz086kgwwxdw2yv1ktn7x35

cc @eolivelli @zymap

bookkeeper-server/src/main/java/org/apache/bookkeeper/meta/zk/ZKMetadataDriverBase.java
shoothzj: @congbobo184 line 256 blank line?

congbobo184: sorry, I thought it was to delete break

dlg99 (approved):

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/api/BookKeeper.java
zymap: Since it is not a required thing for the bookkeeper. I think it should be ok to implement a default handle. @shoothzj

shoothzj: @zymap This default interface might be error prone. I have started a mail thread. I think we can continue discuss in https://lists.apache.org/thread/fk02bz4lggz086kgwwxdw2yv1ktn7x35

eolivelli (changes requested):
as discussed in the mailing list, having a default implementation is error prone
it is better to not have a default implementation

congbobo184:
@shoothzj @eolivelli  please review again, thanks

hezhangjian (approved):

eolivelli (approved):
+1

shoothzj:
Thanks for your contribution. :)

PR #4345, opened 2024-05-06
Bump jackson from 2.13.4.20221013 to 2.17.1 to address CVE list

shoothzj:
### changes
- CVE-2023-5072(7.5), CVE-2022-45688(7.5)
- Also bump apache from 29 to 32, it will update some maven plugins(like maven-shade-plugin), to support jdk21


shoothzj:
we might need to switch to jdk17 for build. I will open a discuss latter

shoothzj:
> we might need to switch to jdk17 for build. I will open a discuss latter

maven-shade-plugin needs to update to support jdk21. `org.apache:apache` needs to be updated. #4352

eolivelli (approved):
LGTM

PR #4292, opened 2024-04-16
build: add java21 and windows daily build

shoothzj:
### Motivation

Add `java21` and `windows` daily build to prepare future support. We already do some works in java21 and windows. See alos #4189 

### Changes

Add daily github actions to trigger `mvn clean install` on java21 and windows

PR #4347, opened 2024-05-07
Disable two tests in BookKeeperTest on JDK 17+ due to UnsupportedOperationException

shoothzj:
### Motivation

JDK 17 and later throw `UnsupportedOperationException` for the `suspend` and `resume` methods, causing the `testConstructionZkDelay` and `testConstructionNotConnectedExplicitZk` tests to fail.

See jdk21 daily build error: https://github.com/apache/bookkeeper/actions/runs/8962118072

### Changes

- Disabled the two affected test methods for JDK 17 and above using the `@EnabledForJreRange` annotation.
- Transitioned to JUnit 5 annotations and assertions.

### Q&A

Q: why there are so many line changes? like format change
A: In junit5, the msg should be put in the latest param, related to some changes.


shoothzj:

dlg99 (commented):
Thank you for picking this up!

Overall this LGTM but the change calls for addition of the "Compatibility Check" CI job with JDK 21 (LTS), otherwise I can't really look at the PR and tell that the fix works. Also this way we won't break this in the future.
See https://github.com/apache/bookkeeper/blob/1ae4be04e20eca4490b52da474d455a2414168b4/.github/workflows/bk-ci.yml#L441-L448
and https://github.com/apache/bookkeeper/blob/1ae4be04e20eca4490b52da474d455a2414168b4/.asf.yaml#L54-L56

shoothzj:
@dlg99 IIUC, if you mean add jdk21 CI, I think we can open a new PR. And shall we postpone it after jdk21 daily build sucesses?
We have added jdk 21 daily build in #4292

dlg99 (approved):

PR #4371, opened 2024-05-17
build: fix owasp check broken

### LINKED ISSUE #4370: Owasp CI is broken
I found that the OWASP check is broken; it doesn't trigger during the change of #4344. It should be triggered when these files are changed.

https://github.com/apache/bookkeeper/blob/c2e8037baa5ad2beaab29c519a5a54a893d00585/.github/workflows/bk-ci.yml#L482-L487

https://github.com/apache/bookkeeper/blob/c2e8037baa5ad2beaab29c519a5a54a893d00585/.github/changes-filter.yaml#L12-L16

shoothzj:
Fix #4370

### DIFF
--- a/.github/changes-filter.yaml
+++ b/.github/changes-filter.yaml
@@ -9,6 +9,7 @@ docs:
   - '**/*.md'
   - '.github/changes-filter.yaml'
   - '.github/ISSUE_TEMPLATE/**'
+  - 'src/owasp-dependency-check-suppressions.xml'
 need_owasp:
   - 'pom.xml'
   - '**/pom.xml'
--- a/.github/workflows/bk-ci.yml
+++ b/.github/workflows/bk-ci.yml
@@ -60,6 +60,7 @@ jobs:
         id: check_changes
         run: |
           echo "docs_only=${{ fromJSON(steps.changes.outputs.all_count) == fromJSON(steps.changes.outputs.docs_count) && fromJSON(steps.changes.outputs.docs_count) > 0 }}" >> $GITHUB_OUTPUT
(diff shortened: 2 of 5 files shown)


StevenLuMT (approved):
LGTM

.github/workflows/bk-ci.yml
```diff
@@ -505,7 +509,8 @@ jobs:
       - name: Set up JDK 11
         uses: actions/setup-java@v4
         with:
-          java-version: 11
+          distribution: 'temurin'
+          java-version: 21
```
hangc0276, line 513: Why upgrade the java version to jdk21? All the current CI's java version is jdk11

shoothzj: @hangc0276 I previously thought that using JDK 21 for OWASP might be potentially faster and help expose more problems.

Additionally, considering that OpenJDK 11 will reach its end of life in October 2024, do we have plan to upgrade CI to jdk17?

.github/changes-filter.yaml
```diff
@@ -9,6 +9,7 @@ docs:
   - '**/*.md'
   - '.github/changes-filter.yaml'
   - '.github/ISSUE_TEMPLATE/**'
+  - 'src/owasp-dependency-check-suppressions.xml'
```
nicoloboschi, line 12: this is wrong, it should go on "need_owasp" section

shoothzj: @nicoloboschi It's already in the "need_owasp` section, I think it's also no need to run tests, so I put it in here.

.github/workflows/bk-ci.yml
hangc0276: At least, we should keep the same with other CI's versions

Issue #4373, opened 2024-05-17
Bookie server lost some metrics of OrderedExecutor


TakaHiR07:
**BUG REPORT**

***Describe the bug***

We found that bookie lost some metrics of OrderedExecutor. The problem is caused after https://github.com/apache/bookkeeper/pull/3546

The root cause is : If we disable traceTaskExecution, it is ok. But if we enable traceTaskExecution, SingleThreadExecutor would do addExecutorDecorators(). Then it can't go through the judge  `if (thread instanceof SingleThreadExecutor)`. Therefore, singleThreadExecutor can not do registerMetrics.

![企业微信截图_2afea188-7d24-4be9-be45-0fba5a517eb9](https://github.com/apache/bookkeeper/assets/13505225/51dc9168-4dc7-4c89-8a27-d06847dee781)

![企业微信截图_8439ec29-1307-4d62-ad26-7753c038aeb9](https://github.com/apache/bookkeeper/assets/13505225/87c494b3-3540-4111-8de4-a7d47d9e029c)

After diving into the code. I found that there are several areas need to be improved in OrderedExecutor and SingleThreadExecutor.

1. fix the lost metrics
2. no need to keep the code relevant to `if (thread instanceof ThreadPoolExecutor)`. This part of code never trigger
3. SingleThreadExecutor#registerMetrics register duplicate metric. This is redundant
4. SingleThreadExecutor#registerMetrics use different format to register metric, which is different from the previous version. I think        it's better the keep the previous metric format

Duplicate metric and different metric format: 
![企业微信截图_c48701cd-9da2-496d-abac-c30d288b5edb](https://github.com/apache/bookkeeper/assets/13505225/fc72d08d-f855-4a9a-b2b9-4075fad4f9c9)

![企业微信截图_402b81a0-58aa-40e9-9729-d95095e04b08](https://github.com/apache/bookkeeper/assets/13505225/8c855778-67c4-4b6e-bdae-cb5bf7a0c543)

PR #4374, opened 2024-05-17
fix lost prometheus metric in OrderedExecutor  

TakaHiR07:

Fix [#4373](https://github.com/apache/bookkeeper/issues/4373)

### Motivation

As shown in the issue.

This is the first pr to fix the lost metric problem. And the other pr improve the metric in SingleThreadExecutor.

### Changes

If we only fix the problem but do not change the metric , the metric is as follow, the metric become visible, though the format is not same.

企业微信截图_179bc46b-3d59-427b-957b-ec185eff3e38
企业微信截图_ac0c1ec5-6363-48e9-9378-a82d95e3af0d

### DIFF
--- a/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/OrderedExecutor.java
+++ b/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/OrderedExecutor.java
@@ -391,6 +391,10 @@ public class OrderedExecutor implements ExecutorService {
             ExecutorService thread = createSingleThreadExecutor(
                     new ThreadFactoryBuilder().setNameFormat(name + "-" + getClass().getSimpleName() + "-" + i + "-%d")
                     .setThreadFactory(threadFactory).build());
+            SingleThreadExecutor ste = null;
+            if (thread instanceof SingleThreadExecutor) {
+                ste = (SingleThreadExecutor) thread;
+            }
 
             if (traceTaskExecution || preserveMdcForTaskExecution) {
                 thread = addExecutorDecorators(thread);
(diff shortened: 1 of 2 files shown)


TakaHiR07:
@merlimat Could you take a look of this problem?

shoothzj (commented):
Nice catch.

I think we can change `createSingleThreadExecutor` return SingleThreadExecutor directly.
BTW, I prefer split this pr, one to fix problem, one to modify the metric name.

StevenLuMT (commented):
LGTM,
I think this kind of metric change require adding a testcase to cover it

TakaHiR07:
> Nice catch.
> 
> I think we can change `createSingleThreadExecutor` return SingleThreadExecutor directly. BTW, I prefer split this pr, one to fix problem, one to modify the metric name.

ok, I would split to 2 pr. But I guess we can't change `createSingleThreadExecutor` return SingleThreadExecutor directly, because OrderedScheduler#createSingleThreadExecutor do not return SingleThreadExecutor

shoothzj (approved):

shoothzj (changes requested):
Would you mind update the description since we split the PR? I would be glad to proceed with the merge once the updates are made.

shoothzj (approved):

shoothzj:
Thanks for your contribution. :)

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/OrderedExecutor.java
```diff
@@ -391,6 +391,10 @@ protected OrderedExecutor(String baseName, int numThreads, ThreadFactory threadF
             ExecutorService thread = createSingleThreadExecutor(
                     new ThreadFactoryBuilder().setNameFormat(name + "-" + getClass().getSimpleName() + "-" + i + "-%d")
                     .setThreadFactory(threadFactory).build());
+            SingleThreadExecutor ste = null;
+            if (thread instanceof SingleThreadExecutor) {
```
hangc0276, line 395: We can registerMetrics here instead of the instanceOf check

TakaHiR07: Do you mean that remove the check `if (thread instanceof SingleThreadExecutor)` directly ?
But createSingleThreadExecutor() is also triggered in OrderedScheduler. In OrderedScheduler, createSingleThreadExecutor() do not return SingleThreadExecutor, so I keep the check here.

PR #4381, opened 2024-05-22
BP-67: Support skipping compaction at busy times

thetumbled:
Implementation PR for BP-67.
Master Issue: 
Proposal PR:

### DIFF
--- a/site3/website/src/pages/bps/BP-67-support-skipping-compaction-at-busy-times.md
+++ b/site3/website/src/pages/bps/BP-67-support-skipping-compaction-at-busy-times.md
@@ -1,0 +1,31 @@
+# BP-67: Support skipping compaction at busy times
+
+### Motivation
+
+Compaction mechanism is a critical part of the system. It is responsible for compacting the entry log file to reclaim disk space.
+But it is a IO intensive operation. While compaction is running, it need to copy the valid data to a new file and delete the old file.
+While the system is under heavy load, it is not a good idea to run compaction as it will consume more IO and CPU resources.
+So, it is better to skip compaction at busy times.
+For many use cases, the busy time is predictable, we can skip compaction at those times.
+
+### Configuration
+
+add the following configuration:
+
+```
+skipCompactionHourRange
+```
+
+default value is empty, which means do not skip compaction.
+Value format is `startHour-endHour`, for example, `0-6,18-24` means skip compaction from 0 to 6 and 18 to 24.
+
+
+### Proposed Changes
+
+Add a new configuration `skipCompactionHourRange` to control the time range of skipping compaction.
+
+
+### Compatibility, Deprecation, and Migration Plan
+
+Full compatibility with the existing behavior. No deprecation needed.
+


thetumbled:
PTAL, thanks. @shoothzj @eolivelli @dlg99 @horizonzy

thetumbled:
For example, the cpu.idle will decrease significantly while doing compaction.
image

eolivelli (approved):
I support this proposal

site3/website/src/pages/bps/BP-67-support-skipping-compaction-at-busy-times.md
```diff
@@ -0,0 +1,30 @@
+# BP-67: Support skipping compaction at busy times
+
+### Motivation
+
+Compaction mechanism is a critical part of the system. It is responsible for compacting the entry log file to reclaim disk space.
+But it is a IO intensive operation. While compaction is running, it need to copy the valid data to a new file and delete the old file.
```
shoothzj, line 6:
```suggestion
But it is an IO intensive operation. While compaction is running, it need to copy the valid data to a new file and delete the old file.
```

shoothzj (approved):
LGTM, I found two nits

dlg99 (commented):
I am rather indifferent about the change.
There was an attempt in the to implement something similar: https://github.com/apache/bookkeeper/pull/851

It was abandoned, as I remember, because the company where the author worked had chosen to use REST API and cron to run enable/disable compaction. 

So you already can disable the compaction on schedule just not via bookie configuration.

I'd support a mechanism that would dynamically pause/resume major/minor compaction if the load on bookie is high (e.g. as tracked by pending requests or request queue latency) if the disk space is enough.

thetumbled:
> I am rather indifferent about the change. There was an attempt in the to implement something similar: #851
> 
> It was abandoned, as I remember, because the company where the author worked had chosen to use REST API and cron to run enable/disable compaction.
> 
> So you already can disable the compaction on schedule just not via bookie configuration.
> 
> I'd support a mechanism that would dynamically pause/resume major/minor compaction if the load on bookie is high (e.g. as tracked by pending requests or request queue latency) if the disk space is enough.

To support dynamically pause/resume major/minor compaction, we need to introduce two interface:
- `LoadStatCollector` collect the load stat for other modules.
- `CompactionStrategy` decide whether do compaction based on load and any other stats, conf.
WDYT @dlg99 @eolivelli @shoothzj

site3/website/src/pages/bps/BP-67-support-skipping-compaction-at-busy-times.md
```diff
@@ -0,0 +1,30 @@
...
+### Configuration
+
+add the following configuration:
+
+```
+skipCompactionHourRange
+```
+
+default value is empty, which means do not skip compaction.
+Value format is `startHour-endHour`, for example, `0-6,18-24` means skip compaction from 0 to 6 and 18 to 24.
```
hangc0276, line 19: If the compaction has already started, does this proposal suspend it when it enters the skip-hour range? What's more, the specific hour depends on the time zone

thetumbled: We will check the hour range at the beginning of the compaction execution and each time it compact a entry log completly.
We use `LocalTime.now().getHour()` to get the hour, implementation refer to: https://github.com/apache/bookkeeper/pull/4385

hangc0276:
> > I am rather indifferent about the change. There was an attempt in the to implement something similar: #851
> > It was abandoned, as I remember, because the company where the author worked had chosen to use REST API and cron to run enable/disable compaction.
> > So you already can disable the compaction on schedule just not via bookie configuration.
> > I'd support a mechanism that would dynamically pause/resume major/minor compaction if the load on bookie is high (e.g. as tracked by pending requests or request queue latency) if the disk space is enough.
> 
> To support dynamically pause/resume major/minor compaction, we need to introduce two interface:
> 
> * `LoadStatCollector` collect the load stat for other modules.
> * `CompactionStrategy` decide whether do compaction based on load and any other stats, conf.
>   WDYT @dlg99 @eolivelli @shoothzj

@thetumbled suspend and resume compaction based on load will make the system more complex and hard to configure a reasonable threshold. Moreover, if we debug the removed ledger not cleaned up, it's hard to explain why the compaction service was suspended.  To reduce the compaction impact in bookie service, we can:
- Introduce accurate throttle for compaction service
- Avoid or suspend the compaction service in high-traffic time

horizonzy (approved):
Good idea, LGTM.

thetumbled:
There is one point need to be discussed: we should check it at the beginning of the compaction, but should we check it each time before we try to compact a new entry log? If we stop the compaction at the running time, should we resume this major/minor compaction or just start a new compaction?
@hangc0276 @eolivelli @shoothzj @horizonzy 
﻿

site3/website/src/pages/bps/BP-67-support-skipping-compaction-at-busy-times.md
hangc0276: If the compaction starter hits the time range, it will skip. When will be the next time trigger compaction?

liangyepianzhou (approved):
I agree with this good idea

PR #4383, opened 2024-05-22
Remove unused code from ByteBufVisitor

lhotari:
### Motivation

ByteBufVisitor added in #4196 contains some code that isn't used and covered by unit tests. 
It's better to remove such code since it makes it harder to reason about the solution. The ByteBufVisitor solution will unwrap all direct buffers that can be unwrapped without the code that is to be removed. I believe that I forgot this code from some earlier phase of the solution where it was necessary to include this.
I ran some local tests and didn't see that the code was used for the original purpose that it was added for (supporting read-only buffers).

### Changes

Remove the unused code.


shoothzj (approved):

lhotari:
It seems that this might be needed. Will add a test case after checking

lhotari:
It's safe to remove the code. I guess I had added the code originally as a performance optimization, but it's not useful in the end. I'm marking this PR as ready to review. @shoothzj Please merge this and cherry-pick to branch-4.16 and branch-4.17.

Issue #4382, opened 2024-05-22
BP-67: Support skipping compaction at busy times


thetumbled:
BP

This is the master ticket for tracking BP-67:

This proposal describes changes to BookKeeper to support throttling for zookeeper read of rereplication.

Proposal PR - https://github.com/apache/bookkeeper/pull/4381
Implement PR -

PR #4385, opened 2024-05-23
BP-67: Support skipping compaction at busy times

thetumbled:
Implementation PR for BP-67.
Master Issue: https://github.com/apache/bookkeeper/issues/4382
Proposal PR: https://github.com/apache/bookkeeper/pull/4381


hangc0276 (changes requested):
Waiting for the proposal being finalized

PR #4393, opened 2024-05-27
improve registerMetric in orderExecutor

TakaHiR07:
Main Issue: https://github.com/apache/bookkeeper/issues/4373

### Motivation

As is shown in the issue, This is the second pr to improve the registerMetric() in orderExecutor.

### Changes

1. remove the duplicate metric
2. modify the metric name format, make it compatible with previous version.

after improvement, the metric is :
企业微信截图_be1a641e-e969-45db-b661-72122d6c1dd6

image

### DIFF
--- a/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/OrderedExecutor.java
+++ b/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/OrderedExecutor.java
@@ -391,9 +391,9 @@ public class OrderedExecutor implements ExecutorService {
             ExecutorService thread = createSingleThreadExecutor(
                     new ThreadFactoryBuilder().setNameFormat(name + "-" + getClass().getSimpleName() + "-" + i + "-%d")
                     .setThreadFactory(threadFactory).build());
-            SingleThreadExecutor ste = null;
             if (thread instanceof SingleThreadExecutor) {
-                ste = (SingleThreadExecutor) thread;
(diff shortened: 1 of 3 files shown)


bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/OrderedExecutor.java
```diff
@@ -391,9 +391,9 @@ protected OrderedExecutor(String baseName, int numThreads, ThreadFactory threadF
             ExecutorService thread = createSingleThreadExecutor(
                     new ThreadFactoryBuilder().setNameFormat(name + "-" + getClass().getSimpleName() + "-" + i + "-%d")
                     .setThreadFactory(threadFactory).build());
-            SingleThreadExecutor ste = null;
             if (thread instanceof SingleThreadExecutor) {
-                ste = (SingleThreadExecutor) thread;
+                SingleThreadExecutor ste = (SingleThreadExecutor) thread;
+                ste.registerMetrics(statsLogger, name, i);
```
shoothzj, line 396: Did you change this like @hangc0276 commented? https://github.com/apache/bookkeeper/pull/4374#discussion_r1614973418

If  we can move registry in here, I think we can change `createSingleThreadExecutor ` to return SingleThreadExecutor directly. Like I saiy in https://github.com/apache/bookkeeper/pull/4374#pullrequestreview-2069989221

PR #4403, opened 2024-05-28
Configure maven-source-plugin to skip in OWASP check profile

shoothzj:
### Motivation

During OWASP checks, the `maven-source-plugin` raises error. It's hard to figure out try at least I try almost two hours. I think we can skip in OWASP check.

```
Error: Failed to execute goal org.apache.maven.plugins:maven-source-plugin:3.3.0
(attach-sources) on project buildtools: Presumably you have configured maven-source-plugin to execute twice times in your build. You have to configure a classifier for at least one of them.
```

### Changes

- Configured `maven-source-plugin` to skip during OWASP checks in the `pom.xml` to prevent the error and ensure seamless build processes during security assessments.

PR #4406, opened 2024-05-30
Correct Configuration Error in OWASP Dependency Check Plugin Setup

shoothzj:
### Motivation
The purpose of this update is to correct a configuration error in our OWASP Dependency Check Maven plugin setup. 
I appreciate @xiezhx9 for opening the issue on [jeremylong/DependencyCheck #6697](https://github.com/jeremylong/DependencyCheck/issues/6697). Also, a special thanks to @jeremylong for his guidance and response.

From dependency-check-plugin maintainer jeremylong
> With the setup specified ODC would run too many times - once with the specified configuration and once without. You can either put the plugin into the build/plugins or in the reporting - don't do both.
> Regarding it re-running on each child module - you likely want to specify <inherited>false</inherited>.

And I believe that cause error I meet before
```
Error: Failed to execute goal org.apache.maven.plugins:maven-source-plugin:3.3.0
(attach-sources) on project buildtools: Presumably you have configured maven-source-plugin to execute twice times in your build. You have to configure a classifier for at least one of them.
```
#4403 is not required, I will close it after this being merged

### Changes
- **Version Update:** The `dependency-check-maven.version` in `pom.xml` has been updated from `9.1.0` to `9.2.0`. This update incorporates the latest vulnerability definitions and enhancements from OWASP.
- **Fix Configuration Inheritance:** Added `<inherited>false</inherited>` to the dependency-check plugin configuration under the build section. This correction prevents the plugin from inheriting settings unnecessarily across multi-module projects, reducing superfluous scans and speeding up builds.
- **Remove Duplicate Configuration:** Eliminated the OWASP plugin configuration from the reporting section. This change fixes a prior oversight where the plugin was configured to execute twice, once in the build and once in the reporting phase, leading to unnecessary build delays.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -190,7 +190,7 @@
     <maven-checkstyle-plugin.version>3.3.1</maven-checkstyle-plugin.version>
     <maven-compiler-plugin.version>3.12.1</maven-compiler-plugin.version>
     <maven-surefire-plugin.version>3.2.5</maven-surefire-plugin.version>
-    <dependency-check-maven.version>9.1.0</dependency-check-maven.version>
+    <dependency-check-maven.version>9.2.0</dependency-check-maven.version>
     <nar-maven-plugin.version>3.10.1</nar-maven-plugin.version>
     <os-maven-plugin.version>1.4.1.Final</os-maven-plugin.version>
     <protobuf-maven-plugin.version>0.6.1</protobuf-maven-plugin.version>
@@ -1091,6 +1091,7 @@
             <groupId>org.owasp</groupId>
             <artifactId>dependency-check-maven</artifactId>
             <version>${dependency-check-maven.version}</version>
+            <inherited>false</inherited>
             <configuration>
               <suppressionFiles>
                 <suppressionFile>src/owasp-dependency-check-suppressions.xml</suppressionFile>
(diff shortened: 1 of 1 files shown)


dlg99 (approved):

lhotari:
I noticed this in the referenced issue:
> Furthermore, after adding the following pom.xml configuration, I noticed that the OWASP Dependency Check Plugin is downloading NVD (National Vulnerability Database) information for each module during the build process, which significantly slows it down.

In Apache Pulsar, this downloading problem is resolved by caching the NVD information. Here's the solution in Pulsar to restore and update the cache: https://github.com/apache/pulsar/blob/c39f9f82b425c66c899f818583714c9c98d3e213/.github/workflows/ci-owasp-dependency-check.yaml#L61-L113 . Restoring the cache: https://github.com/apache/pulsar/blob/c39f9f82b425c66c899f818583714c9c98d3e213/.github/workflows/pulsar-ci.yaml#L1462-L1477 . Please notice that the dir is excluded from maven caching: https://github.com/apache/pulsar/blob/c39f9f82b425c66c899f818583714c9c98d3e213/.github/workflows/pulsar-ci.yaml#L1439 .

PR #4323, opened 2024-04-29
build: disable testNewEnsembleBookieWithOneEmptyRegion method in TestRegionAwareEnsemblePlacementPolicy beyond jdk11

shoothzj:
### Motivation

In JDK21, it's is no way to change private static final variable currently, before figure out how to replace, disable it first

```
        Field logField = repp.getClass().getDeclaredField("LOG");
        Logger mockLogger = mock(Logger.class);
        Field modifiers = Field.class.getDeclaredField("modifiers");
        modifiers.setAccessible(true);
        modifiers.setInt(logField, logField.getModifiers() & ~Modifier.FINAL);
```

See also https://issues.apache.org/jira/browse/CASSANDRA-18181

### Changes

1. Added junit5 annotation `@EnabledForJreRange(max = JRE.JAVA_17)`(It's modern and better than junit4) to tests that use `suspend` and `resume`, effectively disabling these tests on JDK versions higher than 17 where these methods are not supported.
2. BookieZKExpireTest should be annotated as junit5 test.
3. Make `BookKeeperClusterTestCase` compatible to junit5.

PR #4309, opened 2024-04-23
build: disable BookieZKExpireTest beyond jdk17

shoothzj:
### Motivation

As of JDK 21, the methods `suspend` and `resume` on the `Thread` class throw `UnsupportedOperationException`.

### Changes

1. Added junit5 annotation `@EnabledForJreRange(max = JRE.JAVA_17)`(It's modern and better than junit4) to tests that use `suspend` and `resume`, effectively disabling these tests on JDK versions higher than 17 where these methods are not supported.
2. Migrate this test file to junit5

PR #4420, opened 2024-06-05
[WIP][Branch 4.16]ci: fix ci test 

zhiheng123:
- Downgraded Netty to 4.1.100.Final to match the version required by gRPC 1.54.1
- Cherry-picked #4323 and #4309 to make tests running on OpenJDK 17+

### Changes
> I found that this exception was caused by the upgrade of netty. If you
> revert netty to 4.1.100.Final, the use case passes. Because in
> 4.1.101.Final, netty modified the Http2 related interfaces, the connection
> is as follows: https://netty.io/news/2023/11/09/4-1-101-Final.html
>
> The main reason may be incompatibility caused by (#13651). I recommended
> to roll back netty to 4.1.100.Final, which can also address some CVE
> vulnerabilities related to netty. This will allow the ci of branch-4.1.16
> to return to normal without having to upgrade the version of grpc. The
> corresponding components related to netty such as vertx also need to be
> rolled back to the version that matches netty.

### DIFF
--- a/bookkeeper-dist/src/main/resources/LICENSE-all.bin.txt
+++ b/bookkeeper-dist/src/main/resources/LICENSE-all.bin.txt
@@ -217,32 +217,32 @@ Apache Software License, Version 2.
 - lib/commons-io-commons-io-2.7.jar [8]
 - lib/commons-lang-commons-lang-2.6.jar [9]
 - lib/commons-logging-commons-logging-1.1.1.jar [10]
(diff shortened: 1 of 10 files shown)


shoothzj (changes requested):
I suggest we split the netty downgrade and jdk21 support.

dlg99 (changes requested):
Please fill in PR description to explain what is the problem you are trying to solve, how netty downgrade fits in it, etc.

PR #4425, opened 2024-06-09
Fix TimedRunnable log NPE

AnonHxy:

### Motivation

When the `ReadEntryProcessor` runs slow, the `TimedRunnable` will print the warn log like below:
```
Runnable xx:xx took too long xx micros to execute.
```
https://github.com/apache/bookkeeper/blob/fd2c7c6f632deb0130b9c8fc506a6b66ac02b341/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/OrderedExecutor.java#L197-L208

However we found that NPE happened in this log:
```
WARN  [BookieReadThreadPool-OrderedExecutor-1-0:OrderedExecutor$TimedRunnable@206] - Runnable [!!!org.apache.bookkeeper.proto.ReadEntryProcessor@1dd094f3=>java.lang.NullPointerException:Cannot invoke "org.apache.bookkeeper.proto.BookieProtocol$ReadRequest.getLedgerId()" because "this.request" is null!!!]:class org.apache.bookkeeper.proto.ReadEntryProcessor took too long 1125904 micros to execute.
```

The root cause is that the  `ReadEntryProcessor` will be recycled after calling `run` method,  which means we shouldn't call any method of the `ReadEntryProcessor` after calling `ReadEntryProcessor#run()`.   So line206 calling  `toString()` method  of `runnable`(`ReadEntryProcessor`) will cause NPE or any other wrong  message.

### Changes

Save runnable  string name and class when create `TimedRunnable` and `TimedCallable`

### DIFF
--- a/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/OrderedExecutor.java
+++ b/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/OrderedExecutor.java
@@ -187,13 +187,11 @@ public class OrderedExecutor implements ExecutorService {
     protected class TimedRunnable implements Runnable {
         final Runnable runnable;
         final long initNanos;
-        final String runnableString;
         final Class<?> runnableClass;
 
         TimedRunnable(Runnable runnable) {
             this.runnable = runnable;
             this.initNanos = MathUtils.nowInNano();
-            this.runnableString = runnable.toString();
             this.runnableClass = runnable.getClass();
          }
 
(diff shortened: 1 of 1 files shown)


bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/OrderedExecutor.java
```diff
@@ -187,10 +187,14 @@ public T build() {
     protected class TimedRunnable implements Runnable {
         final Runnable runnable;
         final long initNanos;
+        final String runnableString;
+        final Class<?> runnableClass;
 
         TimedRunnable(Runnable runnable) {
             this.runnable = runnable;
             this.initNanos = MathUtils.nowInNano();
+            this.runnableString = runnable.toString();
```
dlg99, line 196: toString() can be expensive; maybe skip this and only log the class?

dlg99 (commented):
Good catch!
One note about perf implications of the change

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/OrderedExecutor.java
AnonHxy: Nice suggestion. Updated @dlg99

shoothzj (approved):

dlg99 (approved):

PR #4430, opened 2024-06-12
[branch-4.17]fix: build release in ubuntu error

zhiheng123:

When building in ubuntu, there will be an error prompting ```bash: [: /home: integer expression expected```, because in bash and zsh, the string equality cannot use ```-eq```, but should use ```=```


shoothzj (changes requested):
I think we should fix this in master branch first, and I will cherry-pick it to branch-4.17 and branch-4.16

Issue #617, opened 2017-10-17
Split bookkeeper client into a separate maven module


ivankelly:
JIRA: https://issues.apache.org/jira/browse/BOOKKEEPER-792

Reporter: Ivan Kelly @ivankelly

Right now, to include bookkeeper as a dependency, you need to include org.apache.bookkeeper:bookkeeper-server

This pulls in all the bookkeeper-server code, and all the server dependencies as well as the client. The server dependencies are larger than the client dependencies. For one thing, it pulls in jna and log4j which should never be pulled in from the client.

### Comments from JIRA

---
*Ivan Kelly* 2014-10-13T16:12:44.295+0000

Patch renames a lot of files, so I'd like to get it in as soon as possible to avoid a nightmare rebase if other changes go in.

---
*Hadoop QA* 2014-10-13T16:47:35.965+0000

Testing JIRA BOOKKEEPER-792

Patch [0001-Split-the-bookkeeper-client-from-the-server.patch|https://issues.apache.org/jira/secure/attachment/12674526/0001-Split-the-bookkeeper-client-from-the-server.patch] downloaded at Mon Oct 13 16:20:56 UTC 2014

----------------------------

{color:green}+1 PATCH_APPLIES{color}
{color:green}+1 CLEAN{color}
{color:red}-1 RAW_PATCH_ANALYSIS{color}
.    {color:green}+1{color} the patch does not introduce any @author tags
.    {color:green}+1{color} the patch does not introduce any tabs
.    {color:green}+1{color} the patch does not introduce any trailing spaces
.    {color:green}+1{color} the patch does not introduce any line longer than 120
.    {color:red}-1{color} the patch does not add/modify any testcase
{color:green}+1 RAT{color}
.    {color:green}+1{color} the patch does not seem to introduce new RAT warnings
{color:green}+1 JAVADOC{color}
.    {color:green}+1{color} the patch does not seem to introduce new Javadoc warnings
{color:green}+1 COMPILE{color}
.    {color:green}+1{color} HEAD compiles
.    {color:green}+1{color} patch compiles
.    {color:green}+1{color} the patch does not seem to introduce new javac warnings
{color:red}-1 FINDBUGS{color}
.    {color:red}-1{color} the patch seems to introduce 23 new Findbugs warning(s) in module(s) [bookkeeper-client]
{color:red}-1 TESTS{color}
.    Tests run: 1113
.    Tests failed: 191
.    Tests errors: 202

.    The patch failed the following testcases:

[...]

PR #4446, opened 2024-06-24
Require Java 17 or Java 21 for building Bookkeeper

lhotari:
### Motivation

There doesn't seem to be a need to retain support for Java 8 and Java 11 in new releases of Bookkeeper.
Java 11 is EOL in October. 
Please see mailing list discussion https://lists.apache.org/thread/gtp4j5y9txzlz0pfgfhy3gz4q4ptnlo8

### Changes

- Enforce Java 17 or Java 21 for building Bookkeeper
- Set `<release>17</release>` for `maven-compiler-plugin` by default
  - keep Java 8 compatibility in `cpu-affinity` and `circe-checksum` since Pulsar's Java client uses them.
- Replace Java 11 usage with Java 17 in CI
- Remove CI jobs using Java 8 or Java 11

### DIFF
--- a/.github/workflows/bk-ci.yml
+++ b/.github/workflows/bk-ci.yml
@@ -78,7 +78,7 @@ jobs:
         uses: actions/setup-java@v4
         with:
           distribution: 'temurin'
-          java-version: 11
+          java-version: 17
 
       - name: Validate pull request
         if: steps.check_changes.outputs.docs_only != 'true'
@@ -156,7 +156,7 @@ jobs:
         uses: actions/setup-java@v4
         with:
           distribution: 'temurin'
-          java-version: 11
+          java-version: 17
 
       - name: Build
         run: |
@@ -230,7 +230,7 @@ jobs:
         uses: actions/setup-java@v4
         with:
           distribution: 'temurin'
-          java-version: 11
+          java-version: 17
 
       - name: Pick ubuntu mirror for the docker image build
         run: |
@@ -315,7 +315,7 @@ jobs:
         uses: actions/setup-java@v4
         with:
           distribution: 'temurin'
-          java-version: 8
+          java-version: 17
 
       - name: Pick ubuntu mirror for the docker image build
         run: |
@@ -397,7 +397,7 @@ jobs:
         uses: actions/setup-java@v4
         with:
           distribution: 'temurin'
-          java-version: 11
+          java-version: 17
 
       - name: mvn package
         run: mvn -B -nsu clean package -DskipTests
@@ -429,7 +429,7 @@ jobs:
         uses: actions/setup-java@v4
         with:
           distribution: 'temurin'
-          java-version: 11
+          java-version: 17
 
       - name: mvn package
         run: mvn -B -nsu clean package -DskipTests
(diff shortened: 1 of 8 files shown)


dlg99:
I think we should first refactor BK client into a separate module that will be built with JDK 8, then we can move BK server to JDK 17

https://github.com/apache/bookkeeper/issues/617

shoothzj:
@lhotari As far as I know, [Pravega](https://cncf.pravega.io/) uses BookKeeper and their minimum jdk is jdk11.
IMHO, even we decide to require jdk17 for BookKeeper client, we need to prepare a mjor version update.

lhotari:
> @lhotari As far as I know, [Pravega](https://cncf.pravega.io/) uses BookKeeper and their minimum jdk is jdk11.
> IMHO, even we decide to require jdk17 for BookKeeper client, we need to prepare a mjor version update.

@shoothzj they won't be staying on jdk11 forever. Java 11 is EOL very soon. Current maintenance branches will be maintained for some time. It doesn't mean that we cut support for Java 8 or Java 11 when we decide to switch to Java 17 in the master branch.

XuQianJin-Stars:
I think upgrading to jdk17 or jdk21 would be more mainstream.

nodece:
Maintaining compatibility with lower versions of Java is a good idea, but more and more libraries are constantly upgrading Java versions, which also forces us to force upgrades.

Using Java 17 or 21 is a goo idea, +1.


> I think we should first refactor BK client into a separate module that will be built with JDK 8, then we can move BK server to JDK 17

It seems that we need to do that first.

liangyuanpeng:
> I think we should first refactor BK client into a separate module that will be built with JDK 8, then we can move BK server to JDK 17

My two cents
working for java11 and not java8 if we want to refactor BK client into a separate module,java8 is becoming a liability.

lhotari:
> > I think we should first refactor BK client into a separate module that will be built with JDK 8, then we can move BK server to JDK 17
> 
> My two cents working for java11 and not java8 if we want to refactor BK client into a separate module,java8 is becoming a liability.

There's also maintenance branches with Java 8 support, I don't see a point keeping support for JDK 8 in the future versions of Bookkeeper. Existing Java 8 users can continue to use maintenance branches until they have migrated to newer JDK versions. I think that master branch should set the minimum JDK version to either Java 17 or Java 21 and leave the previous JDK versions to be handled by maintenance branches. 

Can we even find a single BK users that runs a recent BK version on Java 8? It's definitely a completely unnecessary liability to keep supporting Java 8 in the master branch and future BK versions.

liangyepianzhou:
I agree to update jdk version in 4.18.0. Generally, we should align major JDK upgrades with a major version bump like 5.0.0 for the client. However, considering the imminent EOL of Java 11 and Java 8, it might still be acceptable to perform the JDK upgrade in a feature release like 4.18.0. After all, users who are reluctant to upgrade their JDKs are likely less interested in new features—as long as we continue providing bug fixes for the older JDK 8/11 branches, the impact would be minimal.

StevenLuMT:
PRs that have not been processed for a long time will be closed first. Please open them yourself if necessary.

rdhabalia (approved):

merlimat:
> I guess, it's better to keep jdk-11 support for sometime as some legacy systems are still running old bookkeeper version, and upgrading bookie is also upgrades rocksDB version where I saw some incompatibility between rocksDB-5.x and rocksDB-7.x where it doesn't allow to downgrade bookkeeper because rocksDB-7.x introduced unknown tags in MANIFEST file which is not recoverable by rocksDB-5.x if one wants to downgrade bookie.

There's no requirement to upgrade the bookies.

merlimat (approved):

lhotari:
I have rebased this PR so that we could merge this for 4.18.0 release

PR #4448, opened 2024-06-25
[fix] fix string format for exception in RocksdbKVStore

dk2k:
Fixed string format for exception


shoothzj (approved):
Thanks for your contribution.

dlg99 (approved):

PR #4449, opened 2024-06-25
[BugFix] Fix to prevent resource leaks in 3 classes 

dk2k:
Implemented try-with-resources for instances of Autocloseable

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLoggerAllocator.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLoggerAllocator.java
@@ -180,24 +180,23 @@ class EntryLoggerAllocator {
         } while (testLogFile == null);
 
         File newLogFile = new File(dirForNextEntryLog, logFileName);
-        try (FileChannel channel = new RandomAccessFile(newLogFile, "rw").getChannel()) {
+        FileChannel channel = new RandomAccessFile(newLogFile, "rw").getChannel();
 
-            BufferedLogChannel logChannel = new BufferedLogChannel(byteBufAllocator, channel, conf.getWriteBufferBytes(),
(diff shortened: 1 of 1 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLoggerAllocator.java
```diff
@@ -180,23 +180,24 @@ private synchronized BufferedLogChannel allocateNewLog(File dirForNextEntryLog,
         } while (testLogFile == null);
 
         File newLogFile = new File(dirForNextEntryLog, logFileName);
-        FileChannel channel = new RandomAccessFile(newLogFile, "rw").getChannel();
+        try (FileChannel channel = new RandomAccessFile(newLogFile, "rw").getChannel()) {
```
dlg99, line 183: is it a good idea to close the channel here? it is used by the logChannel that is being returned

dk2k: Reverted changes for this file

dlg99 (approved):

shoothzj (commented):
It seems we have been leaked some resources for long, really? Need more eyes. @nicoloboschi @hangc0276 @nodece

nodece (approved):

nodece:
> It seems we have been leaked some resources for long, really? Need more eyes. @nicoloboschi @hangc0276 @nodece

Looks like it. We should take care of it.

shoothzj (approved):

shoothzj (changes requested):
LGTM, but I think we need to change the commit and PR title. It's a fix of leak

fu-turer (approved):

nicoloboschi (approved):

shoothzj (approved):

shoothzj:
Thanks for your contribution!

PR #4450, opened 2024-06-25
[Improvement] Removed unused/non-updated objects

dk2k:


tests/integration-tests-topologies/src/main/java/org/apache/bookkeeper/tests/containers/ChaosContainer.java
```diff
@@ -85,7 +85,6 @@ public void tailContainerLog() {
     }
 
     public String getContainerLog() {
```
shoothzj, line 87: This might hides a bug? After refactor, it seems meaningless.

dk2k: Indeed, the nethod was unused. I can remove it if you agree

PR #4456, opened 2024-06-26
improve: use jlink to reduce jdk size in the image

nodece:
### Motivation

Reference: https://adoptium.net/blog/2021/08/using-jlink-in-dockerfiles

Reduce size: 398.23 MB -> 245.23 MB

New image:https://hub.docker.com/layers/nodece/bk/4.17.1-jlink/images/sha256-73c8354a981d7dfefcb1247cf84f6c3e91db2342431749e528ba8468c65be1f5?context=explore

Old image: https://hub.docker.com/layers/apache/bookkeeper/4.17.1/images/sha256-342f384bc21b4d31d5349f741cf831f3c5bb47dfea34557104528ccd9597810f?context=explore

### Changes

- Using jlink to compress the JDK and then copy that to the base image

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -17,7 +17,7 @@
 # under the License.
 #
 
-FROM eclipse-temurin:11 as jre-build
+FROM eclipse-temurin:17 as jre-build
 
 # Create a custom Java runtime
 RUN $JAVA_HOME/bin/jlink \
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -17,6 +17,20 @@
 # under the License.
 #
 
+FROM eclipse-temurin:11 as jre-build
+
+# Create a custom Java runtime
+RUN $JAVA_HOME/bin/jlink \
+         --add-modules ALL-MODULE-PATH \
+         --strip-debug \
+         --no-man-pages \
+         --no-header-files \
+         --compress=2 \
+         --output /javaruntime
+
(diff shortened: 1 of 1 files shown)


nodece:
Could you help to review? @shoothzj @lhotari

shoothzj:
Does this only include the JRE, and exclude JDK tools like jmap, etc.?

nodece:
> Does this only include the JRE, and exclude JDK tools like jmap, etc.?

No. This image includes the JDK tools.

hezhangjian (approved):

nodece:
Could you help to review? @dlg99

docker/Dockerfile
```diff
@@ -42,7 +56,6 @@ RUN set -x \
     && adduser "${BK_USER}" \
     && apt-get update \
     && apt-get install -y ca-certificates apt-transport-https \
-    && apt-get install -y --no-install-recommends openjdk-17-jdk \
```
dlg99, line 45: as I understand, this silently/without discussion downgrades jdk from 17 to 11 (included in `eclipse-temurin:11`)

dlg99 (commented):
Am I right that this change will downgrade jdk from 17 to 11 in the docker image?

docker/Dockerfile
nodece: Good catch!

nodece:
> Am I right that this change will downgrade jdk from 17 to 11 in the docker image?

You are right, I should use jdk 17.

dlg99 (approved):

PR #4462, opened 2024-07-08
[fix] Fix data lost after when writing ledger and deleting legder execute concurrency

poorbarcode:
### Motivation

| step | `BK client 1` | `BK client 2` |
| --- | --- | --- |
| 1 | create ledger `1` | 
| 2 | | open ledger `1` |
| 3 | | delete ledger `1` |
| 4 | write data to ledger `1` |

At the step `4`, the write should fail, but it succeeds. It leads users to assume the data has been written, but it can not be read.

You can reproduce the issue by `testWriteAfterDeleted`

There is a scenario that will lead to Pulsar loss messages

- `broker-2` created a ledger
- `broker-2`'s ZK session is expired, which will lead the topic it owned to be assigned to other brokers
- `broker-0` owned the topic again
  - it will delete the last empty ledger
- consumers connected to `broker-0`
- producers connected to `broker-2`
  - send messages to the topic
- on `broker-2`, the ledger can not be closed due to the ledger metadata has been deleted 

### Changes

Once the ledger is fenced, it can not be wrote anymore.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/HandleFactory.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/HandleFactory.java
@@ -24,7 +24,7 @@ package org.apache.bookkeeper.bookie;
 import java.io.IOException;
 
 interface HandleFactory {
-    LedgerDescriptor getHandle(long ledgerId, byte[] masterKey)
+    LedgerDescriptor getHandle(long ledgerId, byte[] masterKey, boolean journalReplay)
             throws IOException, BookieException;
 
     LedgerDescriptor getReadOnlyHandle(long ledgerId)
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/HandleFactoryImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/HandleFactoryImpl.java
@@ -79,20 +79,20 @@ class HandleFactoryImpl implements HandleFactory, LedgerDeletionListener {
         return handle;
     }
 
-    private void markIfConflictWritingOccurs(long ledgerId) throws IOException {
+    private void markIfConflictWritingOccurs(long ledgerId) {
         LedgerDescriptor ledgerDescriptor = ledgers.get(ledgerId);
         try {
             if (ledgerDescriptor != null && ledgerDescriptor.isFenced()) {
                 recentlyFencedAndDeletedLedgers.put(ledgerId, true);
             }
-        } catch (BookieException ex) {
+        } catch (IOException | BookieException ex) {
             // The ledger is in limbo state.
             recentlyFencedAndDeletedLedgers.put(ledgerId, true);
         }
     }
 
     @Override
-    public void ledgerDeleted(long ledgerId) throws IOException{
+    public void ledgerDeleted(long ledgerId) {
         markIfConflictWritingOccurs(ledgerId);
         // Do delete.
         ledgers.remove(ledgerId);
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/HandleFactoryImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/HandleFactoryImpl.java
@@ -21,13 +21,24 @@
 
 package org.apache.bookkeeper.bookie;
 
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
 import java.io.IOException;
+import java.time.Duration;
 import org.apache.bookkeeper.bookie.LedgerStorage.LedgerDeletionListener;
 import org.apache.bookkeeper.util.collections.ConcurrentLongHashMap;
 
 class HandleFactoryImpl implements HandleFactory, LedgerDeletionListener {
     private final ConcurrentLongHashMap<LedgerDescriptor> ledgers;
     private final ConcurrentLongHashMap<LedgerDescriptor> readOnlyLedgers;
+
+    /**
+     * Once the ledger was marked "fenced" before, the ledger was accessed by multi clients. One client is calling
+     * "delete" now, and other clients may call "write" continuously later. We mark these ledgers can not be written
+     * anymore. And maintains the state for 7 days is safety.
+     */
+    private final Cache<Long, Boolean> recentlyFencedAndDeletedLedgers = CacheBuilder.newBuilder()
+            .expireAfterAccess(Duration.ofDays(7)).build();
 
     final LedgerStorage ledgerStorage;
 
@@ -40,10 +51,14 @@ class HandleFactoryImpl implements HandleFactory, LedgerDeletionListener {
     }
 
     @Override
-    public LedgerDescriptor getHandle(final long ledgerId, final byte[] masterKey) throws IOException, BookieException {
+    public LedgerDescriptor getHandle(final long ledgerId, final byte[] masterKey, boolean journalReplay)
+            throws IOException, BookieException {
         LedgerDescriptor handle = ledgers.get(ledgerId);
 
         if (handle == null) {
+            if (!journalReplay && recentlyFencedAndDeletedLedgers.getIfPresent(ledgerId) != null) {
+                throw BookieException.create(BookieException.Code.LedgerFencedException);
+            }
             handle = LedgerDescriptor.create(masterKey, ledgerId, ledgerStorage);
             ledgers.putIfAbsent(ledgerId, handle);
         }
@@ -64,8 +79,22 @@ class HandleFactoryImpl implements HandleFactory, LedgerDeletionListener {
         return handle;
     }
 
+    private void markIfConflictWritingOccurs(long ledgerId) throws IOException {
+        LedgerDescriptor ledgerDescriptor = ledgers.get(ledgerId);
+        try {
+            if (ledgerDescriptor != null && ledgerDescriptor.isFenced()) {
+                recentlyFencedAndDeletedLedgers.put(ledgerId, true);
+            }
+        } catch (BookieException ex) {
+            // The ledger is in limbo state.
+            recentlyFencedAndDeletedLedgers.put(ledgerId, true);
+        }
+    }
+
     @Override
-    public void ledgerDeleted(long ledgerId) {
+    public void ledgerDeleted(long ledgerId) throws IOException{
+        markIfConflictWritingOccurs(ledgerId);
+        // Do delete.
         ledgers.remove(ledgerId);
         readOnlyLedgers.remove(ledgerId);
     }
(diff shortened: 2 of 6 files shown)


horizonzy (commented):
The Patch is LGTM, but I have a concern about the intent. 
Now we only forbid write data to a fenced ledger, but if the ledger(not fenced) is deleted, should we allow it to write data?

poorbarcode:
> Now we only forbid write data to a fenced ledger, but if the ledger(not fenced) is deleted, should we allow it to write data?

If it has not been fenced before, it means only one client has the ledger handler, and if it was written after it was deleted, this means there is a bug on the client side.

hangc0276 (changes requested):

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/HandleFactoryImpl.java
```diff
@@ -64,8 +79,22 @@ public LedgerDescriptor getReadOnlyHandle(final long ledgerId) throws IOExceptio
...
+            }
+        } catch (IOException | BookieException ex) {
+            // The ledger is in limbo state.
+            recentlyFencedAndDeletedLedgers.put(ledgerId, true);
+        }
+    }
+
     @Override
     public void ledgerDeleted(long ledgerId) {
+        markIfConflictWritingOccurs(ledgerId);
```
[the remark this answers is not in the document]
poorbarcode: > It depends on the garbage collection triggered. If the gc is not triggered, we also should fail the case, because the ledger metadata has already been deleted, correct?

- Before garbage collection: the variable `fenced` of the result of `LedgerHandleFactory.getHandle( ledgerId )` is `fenced`, so the BK client will get a `ledger fenced` error.
- After garbage collection: the ledger will be marked as `fenced & deleted`, so the BK client will also get a `ledger fenced` error.

So the current change is fine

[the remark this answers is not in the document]
hangc0276: If the bk client opens the ledger by `openLedgerNoRecovery`, the service won't mark the ledger as a `fenced` state. The `fenced` state is caused by bk client open the ledger with recover mode. Of cause, this fix can fix the issue we encountered in Pulsar.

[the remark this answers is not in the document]
poorbarcode: > If the bk client opens the ledger by openLedgerNoRecovery, the service won't mark the ledger as a fenced state. The fenced state is caused by bk client open the ledger with recover mode. Of cause, this fix can fix the issue we encountered in Pulsar.

If the BK client opens the ledger by `openLedgerNoRecovery`, does it mean this client will only read data from the ledger and never `close/delete` it?

If yes, only the first writable ledger handler will delete the ledger, so we can confirm there is a bug in the client who opens the ledger by `openLedgerNoRecovery` if it deleted the ledger. Please correct the behavior if my understanding is wrong.

zymap (approved):

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/HandleFactory.java
```diff
@@ -24,7 +24,7 @@
 import java.io.IOException;
 
 interface HandleFactory {
-    LedgerDescriptor getHandle(long ledgerId, byte[] masterKey)
+    LedgerDescriptor getHandle(long ledgerId, byte[] masterKey, boolean journalReplay)
```
shoothzj, line 27: will `readOnly` better than `journalReplay`?

hezhangjian (commented):
Is there contains Race Condition between `deleteLedger` and `getHandle` with same ledgerId. cc @eolivelli @nicoloboschi

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/HandleFactory.java
poorbarcode: - Get a handle that is read-only will call `getReadOnlyHandle`
- There is only one scenario that will call `getHandle` to read, it is journal replay.

poorbarcode:
> Is there contains Race Condition between deleteLedger and getHandle with same ledgerId. cc @eolivelli @nicoloboschi

There is no race condition issue. See details below

There are two thread safety maps: `ledgers` and `recentlyFencedAndDeletedLedgers`
- `OPS: deleteLedger` adds ledger to `recentlyFencedAndDeletedLedgers` first, then removes ledger from `ledgers`.
- `OPS: getHandle` get ledger from `ledgers` first, then get ledger from `recentlyFencedAndDeletedLedgers`

---

**Scenario 1**
| `getHandle` | `deleteLedger` |
| --- | --- |
| `write`: get an existing handle |
| | fence |
| | `delete ledger`: add ledger into `recentlyFencedAndDeletedLedgers` |
| `write`: get an existing handle(which is fenced), client will get a fenced error |
| | `delete ledger`: remove ledger from `ledgers` |

---

**Scenario 2**
| `getHandle` | `deleteLedger` |
| --- | --- |
| `write`: get an existing handle |
| | fence |
| `write`: get an existing handle(which is fenced), client will get a fenced error |
| | `delete ledger`: add ledger into `recentlyFencedAndDeletedLedgers` |
| | `delete ledger`: remove ledger from `ledgers` |

---

**Scenario 3**
| `getHandle` | `deleteLedger` |
| --- | --- |
| `write`: get an existing handle |
| | fence |
| | `delete ledger`: add ledger into `recentlyFencedAndDeletedLedgers` |
| | `delete ledger`: remove ledger from `ledgers` |
| `write`: get an error since it is in `recentlyFencedAndDeletedLedgers` |

---

**Scenario 4**
| `getHandle` | `deleteLedger` |
| --- | --- |
| | fence |
| `write`: get an existing handle(which is fenced), client will get a fenced error |
| | `delete ledger`: add ledger into `recentlyFencedAndDeletedLedgers` |
| | `delete ledger`: remove ledger from `ledgers` |

---

**Scenario 5**
| `getHandle` | `deleteLedger` |
| --- | --- |
| `write: get handle`: get an existing handle |
| | fence |
| | `delete ledger`: add ledger into `recentlyFencedAndDeletedLedgers` |
| | `delete ledger`: remove ledger from `ledgers` |
| `write: do write`: update LAC |

**Scenario 5 additional note**: Since `getHandle` and `deleteLedger` execute at the same thread, this scenario will not occur.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/HandleFactoryImpl.java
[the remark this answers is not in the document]
hangc0276: You are correct! I checked all the use case of `openLedgerNoRecovery` in BookKeeper and Pulsar, we won't delete ledgers with openLedgerNoRecovery LedgerHandle.

hangc0276 (approved):

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/TestFencing.java
```diff
@@ -87,6 +94,61 @@ public void testBasicFencing() throws Exception {
...
+            triggerGC(server.getServer().getBookie());
+        }
+
+        try {
+            long entryId = writeLedger.addEntry(tmp.getBytes());
+            LOG.info("Not expected: entryId: {}", entryId);
+            LOG.error("Should have thrown an exception");
+            fail("Should have thrown an exception when trying to write");
+        } catch (Exception e) {
+            e.printStackTrace();
```
eolivelli, line 124: Please use logger and catch the expected exception

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/HandleFactoryImpl.java
```diff
@@ -21,14 +21,25 @@
...
 class HandleFactoryImpl implements HandleFactory, LedgerDeletionListener {
     private final ConcurrentLongHashMap<LedgerDescriptor> ledgers;
     private final ConcurrentLongHashMap<LedgerDescriptor> readOnlyLedgers;
 
+    /**
+     * Once the ledger was marked "fenced" before, the ledger was accessed by multi clients. One client is calling
+     * "delete" now, and other clients may call "write" continuously later. We mark these ledgers can not be written
+     * anymore. And maintains the state for 7 days is safety.
+     */
+    private final Cache<Long, Boolean> recentlyFencedAndDeletedLedgers = CacheBuilder.newBuilder()
```
eolivelli, line 40: This is in memory, what happens if you restart the bookie ?

eolivelli (changes requested):

bookkeeper-server/src/test/java/org/apache/bookkeeper/client/TestFencing.java
poorbarcode: Fixed

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/HandleFactoryImpl.java
poorbarcode: > This is in memory, what happens if you restart the bookie?

I think scenarios `1` and `2` below are fine, and scenario `3` below is not fine, do you have any suggestions?

---

**Scenario 1: the journal log is still there**
- Deleted the ledger.
- Restart BK.
- Replay journal logs.
- Reset the ledger as fenced.
  - Since the ledger is fenced, the client will get a fenced error when it tries to write. 
- Delete the ledger after a GC
  - Move the ledger into  `recentlyFencedAndDeletedLedgers`
  - Since the ledger is in the variable `recentlyFencedAndDeletedLedgers`, the client will get a fenced error when it tries to write.

---

**Scenario 2: the journal log was removed & the client switched the ensemble due to pending operations failed**
- Deleted the ledger.
  - Journal logs were removed.
- Restart BK.
- The client tries to switch the ensemble due to previous operation fails
  - Since the ledger metadata was removed, this operation will fail.

---

 **Scenario 3: the journal log was removed & the client did not switch the ensemble because there are no pending operations**
- Deleted the ledger.
  - Journal logs were removed.
- Restart BK.
- The client reconnect successfully.
- If the client tries to write, it will be successful.

poorbarcode: @eolivelli How about we split this into 2 PRs
- Fix it in memory, which is what the current PR did. And this PR can be cherry-picked into the previous branches, this fix can fix almost all the scenarios(the probably of brokers and Bks crashes at the same time is quite few)
- Fix it with persistent storage, we can persist the states into a file, and add a config to define which file will be used, this will be released with the next feature release

eolivelli (approved):
LGTM

The is a small nit about a printStackTrace, please remove it before committing the patch

eolivelli:
Then I expect to see the follow up work:

> Fix it with persistent storage, we can persist the states into a file, and add a config to define which file will be used, this will be released with the next feature release

dlg99 (approved):

PR #4473, opened 2024-07-27
[fix][ci] Fix OWASP Dependency Check download by using NVD API key

shoothzj:
See more in https://github.com/apache/pulsar/pull/22999. Thanks for @lhotari 
INFRA-TICKET https://issues.apache.org/jira/projects/INFRA/issues/INFRA-26000

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -192,7 +192,7 @@
     <maven-checkstyle-plugin.version>3.3.1</maven-checkstyle-plugin.version>
     <maven-compiler-plugin.version>3.12.1</maven-compiler-plugin.version>
     <maven-surefire-plugin.version>3.2.5</maven-surefire-plugin.version>
-    <dependency-check-maven.version>9.2.0</dependency-check-maven.version>
+    <dependency-check-maven.version>10.0.2</dependency-check-maven.version>
     <nar-maven-plugin.version>3.10.1</nar-maven-plugin.version>
     <os-maven-plugin.version>1.4.1.Final</os-maven-plugin.version>
     <protobuf-maven-plugin.version>0.6.1</protobuf-maven-plugin.version>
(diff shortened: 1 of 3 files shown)


shoothzj:
@lhotari It seems download is still failed, do we need to merge this first?

pom.xml
```diff
@@ -192,7 +192,7 @@
     <maven-checkstyle-plugin.version>3.3.1</maven-checkstyle-plugin.version>
     <maven-compiler-plugin.version>3.12.1</maven-compiler-plugin.version>
     <maven-surefire-plugin.version>3.2.5</maven-surefire-plugin.version>
-    <dependency-check-maven.version>9.2.0</dependency-check-maven.version>
+    <dependency-check-maven.version>10.0.1</dependency-check-maven.version>
```
lhotari, line 195: Should use at least 10.0.2 version

lhotari (changes requested):

lhotari:
> @lhotari It seems download is still failed, do we need to merge this first?

Yes. Another option would be to create the PR branch to apache/bookkeeper so that it has access to the secret. The dependency check version must be at least 10.0.2 (there's another Pulsar PR explaining that).

shoothzj:
@lhotari Thanks, I see your description in https://github.com/apache/pulsar/pull/23012.

I suggest we can merge this pr first.

> There's a [mandatory upgrade notice](https://github.com/jeremylong/DependencyCheck?tab=readme-ov-file#mandatory-upgrade-notice) about upgrading to 10.0.2 version.

lhotari:
> I suggest we can merge this pr first.

That works for me, but why not just upgrade to 10.0.2 in this PR?

lhotari (approved):

lhotari:
Looks like 10.0.2 is used in this PR, LGTM.

lhotari (approved):

lhotari:
There's a error building the docker image
```
[INFO] Package netcat is a virtual package provided by:
[INFO]   netcat-traditional 1.10-48
[INFO]   netcat-openbsd 1.226-1ubuntu2
[INFO] 
[INFO] E: Package 'netcat' has no installation candidate
```
I believe `netcat-openbsd` should be used.

lhotari (approved):

PR #4475, opened 2024-07-27
Fix async log appender not print error log when bookie starting expectionally

### LINKED ISSUE #4474: Async log appender not print log when bookie starting error
**BUG REPORT**

***Describe the bug***

Async log appender not print log when bookie starting error

***To Reproduce***

1. Start a new bookie with java properties: `-DLog4jContextSelector=org.apache.logging.log4j.core.async.AsyncLoggerContextSelector`,  so that we using the asyc log appender.
2. Note that don't format the metadata in advance, so that the bookie can start error.

***Expected behavior***

The log should print `Failed to build bookie server` and the error stack trace.  But we can not see any error log.

AnonHxy:
### Motivation

Fix https://github.com/apache/bookkeeper/issues/4474

The root cause is that when bookie starting error,  the java runtime will exit  immediately. And the log4j can not shutdown gracefully:
https://github.com/apache/bookkeeper/blob/e5c418e06e1cdba1c4681159c7266a43fa403cb1/bookkeeper-server/src/main/java/org/apache/bookkeeper/server/Main.java#L199-L201

### Changes

Call `LogManager#shutdown`  in jvm shutdown hook.

### DIFF
--- a/bookkeeper-common/pom.xml
+++ b/bookkeeper-common/pom.xml
@@ -26,6 +26,10 @@
   <name>Apache BookKeeper :: Common</name>
   <dependencies>
     <dependency>
+      <groupId>org.apache.logging.log4j</groupId>
+      <artifactId>log4j-api</artifactId>
+    </dependency>
+    <dependency>
       <groupId>org.apache.bookkeeper.stats</groupId>
       <artifactId>bookkeeper-stats-api</artifactId>
       <version>${project.parent.version}</version>
--- a/bookkeeper-common/pom.xml
+++ b/bookkeeper-common/pom.xml
@@ -26,10 +26,6 @@
   <name>Apache BookKeeper :: Common</name>
   <dependencies>
     <dependency>
-      <groupId>org.apache.logging.log4j</groupId>
-      <artifactId>log4j-api</artifactId>
-    </dependency>
-    <dependency>
       <groupId>org.apache.bookkeeper.stats</groupId>
       <artifactId>bookkeeper-stats-api</artifactId>
       <version>${project.parent.version}</version>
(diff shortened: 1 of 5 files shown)


hezhangjian (approved):

shoothzj:
IIUC, you mean call `LogManager#shutdown` will flush the log to the disk?

dlg99 (changes requested):
What happens if the user doesn't include logs4j's jars and uses logback or something else?
All prod components of BK should only depend on slf4j.

AnonHxy:
> What happens if the user doesn't include logs4j's jars and uses logback or something else? All prod components of BK should only depend on slf4j.

Good point.  But it seems that there is no way to shutdown or force flush the log through slf4j api.  So some log will be lost if user using async logger appender.

I think that we could print the key error logs to `stderr`.  So that users can  find cause for abnormal exit from the termianl  or stderr logs.    @dlg99 @shoothzj

AnonHxy:
rerun failure checks

shoothzj:
Thank you for the update. I understand the need to ensure logs are flushed properly. However, adding multiple `System.err` outputs might not be the most elegant solution. Could we consider pushing for adding this hook to log4j2 instead?

BTW, I am trying to fix the ci in #4476

dlg99 (approved):

StevenLuMT:

StevenLuMT (approved):
LGTM

lhotari:
> Thank you for the update. I understand the need to ensure logs are flushed properly. However, adding multiple `System.err` outputs might not be the most elegant solution. Could we consider pushing for adding this hook to log4j2 instead?

When using Log4J2, it would be necessary to call `org.apache.logging.log4j.LogManager.shutdown()` to ensure that logs get flushed before a forceful JVM exit with `Runtime.getRuntime().halt(int)`.

hangc0276:
@AnonHxy  Would you please check Lari's comment? Just printing `System.err.println(e.getMessage());` is not a standard solution.

PR #4476, opened 2024-07-28
fix: install netcat-openbsd instead of netcat in test image build

hezhangjian:


lhotari (approved):

lhotari:
```
 The command '/bin/sh -c wget -nv https://archive.apache.org/dist/zookeeper/bookkeeper/bookkeeper-4.1.0/bookkeeper-server-4.1.0-bin.tar.gz{,.sha1,.md5,.asc}' returned a non-zero code: 8
```

It looks like the command doesn't get properly expanded. I wonder if using bash shell would address the problem.

shoothzj:
@lhotari I change `FROM eclipse-temurin:8-jdk` to `FROM eclipse-temurin:8-jdk-jammy`, to keep style consistent with other ubuntu docker image.

lhotari:
> @lhotari I change `FROM eclipse-temurin:8-jdk` to `FROM eclipse-temurin:8-jdk-jammy`, to keep style consistent with other ubuntu docker image.

Oh yes, makes sense. It seems that things broken when eclipse-temurin switched to use Ubuntu 24.04 (noble) base image. Switching to jammy (22.04) makes the existing commands to work.

lhotari (approved):

PR #4478, opened 2024-08-13
[improve] Optimize reorderReadSequence to Check WriteSet

ange-k:
### Motivation

`RackawareEnsemblePlacementPolicyImpl.reorderReadSequenc` has a feature to sort Bookies with poor performance.

Currently, the code checks the entire Ensemble for unavailable Bookies and performs reordering if any are found. However, since the sorting is done within the WriteSet, checking the Ensemble is overkill (except when the sizes are the same).

This change modifies the code to check and reorder only the Bookies within the WriteSet. This prevents wasteful sorting processes at the same time, future developers can understand the logic more easily without wondering if there is a specific reason for checking the entire Ensemble.

### Changes

- Modified `RackawareEnsemblePlacementPolicyImpl.reorderReadSequence` to check and reorder only the Bookies within the WriteSet instead of the entire Ensemble.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
@@ -920,8 +920,9 @@ public class RackawareEnsemblePlacementPolicyImpl extends TopologyAwareEnsembleP
         if (useRegionAware || reorderReadsRandom) {
             isAnyBookieUnavailable = true;
         } else {
-            for (int i = 0; i < ensemble.size(); i++) {
-                BookieId bookieAddr = ensemble.get(i);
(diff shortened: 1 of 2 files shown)


eolivelli (approved):
+1

Out of curiosity, how big is your "ensemble" ? I am used to values from 3 to 7, and with those numbers this patch is not a big deal

ange-k:
> +1
> 
> 
> 
> Out of curiosity, how big is your "ensemble" ? I am used to values from 3 to 7, and with those numbers this patch is not a big deal


Thank you for your feedback.
In our environment, the ensemble size is set to 6, and the WriteSet size is 2.

As you pointed out, we also recognize that the performance improvement from this code change may be minimal.
Nevertheless, I found the unnecessary sorting to be somewhat inefficient. While it may not have a significant impact, I felt it was worth addressing for code cleanliness, which led me to submit this PR.

hangc0276 (approved):
Nice catch!

PR #4482, opened 2024-08-16
[improve] when rackaware failed to choose a bookie, print out the list of ensemble.

liudezhi2098:
### Motivation

In file RackawareEnsemblePlacementPolicyImpl.java 
In the log.warn below, we should print out the list of ensemble, instead of the object. The “ensemble” in line-619 should be changed into “ensemble.toList()”.

https://github.com/apache/bookkeeper/blob/999cd0f2ab14404be4d6c24e388456dbe56bb1a8/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java#L600C3-L619C47
### Changes
The “ensemble” in line-619 changed into “ensemble.toList()”.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
@@ -590,7 +590,7 @@ public class RackawareEnsemblePlacementPolicyImpl extends TopologyAwareEnsembleP
             LOG.warn("Failed to choose a bookie from network location {}, "
                     + "the bookies in the network location are {}, excluded bookies {}, "
                     + "current ensemble {}, fallback to choose bookie randomly from the cluster.",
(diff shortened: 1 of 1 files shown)


shoothzj:
Thank you for the contribution! Is this the only place where this needs to be changed? If there are multiple instances, would it make more sense to implement a `toString()` method for the ensemble?

eolivelli (approved):
+1

liudezhi2098:
> Thank you for the contribution! Is this the only place where this needs to be changed? If there are multiple instances, would it make more sense to implement a `toString()` method for the ensemble?

toString is used for other purposes, so considering the minimum change, this method is not adjusted. 
https://github.com/apache/bookkeeper/blob/f048e7a3b127cfa3a7e9761d589225f078d6712c/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/TopologyAwareEnsemblePlacementPolicy.java#L494-L498

comparison with other places has not found any need for adjustment.

hangc0276 (approved):

dlg99 (approved):

shoothzj (approved):

PR #4485, opened 2024-08-21
[fix] Change log level to debug when check parameters in BatchRead.

dao-jun:
### Motivation

Change log-level to debug when checking param `maxSize` in LedgerHandle#batchReadEntriesInternalAsync.

In some conditions, we don't know what's the size of the entries to be read, so we pass `0` to `batchReadAsync` method,
if log the message in each invoke, it may impact the performance and cause too much noise.


StevenLuMT:

StevenLuMT (approved):
LGTM

PR #4488, opened 2024-08-24
fix: throw reject when SingleThreadExecutor drainTo in progress and queue is empty

### LINKED ISSUE #4465: The maxPendingAddRequestsPerThread configuration is inconsistent with the actual behavior, resulting in netty direct memory OOM
**BUG REPORT**

***Describe the bug***

The maxPendingAddRequestsPerThread configuration in bookkeeper.conf is inconsistent with the actual behavior, resulting in netty direct memory OOM.

***To Reproduce***

1) set config: maxPendingAddRequestsPerThread=1000
2. simulate rocksdb failure, flush is blocked for a long time
3. addEntry requests are queued in the thread pool, and the size of queues is twice the maxPendingAddRequestsPerThread configuration
4. the size of queued requests doubled, resulting in netty direct memory OOM

***Expected behavior***

If we configure maxPendingAddRequestsPerThread to 1000, the maximum number of thread pool queues should not exceed 1000, otherwise it will cause unexpected memory usage and may cause OOM.

By reading the source code, I found that the root cause of the problem is that there is a localTasks logic in SingleThreadExecutor. Before each execution, it tries to drain all tasks from the thread pool queue to localTasks. At this time, the thread pool queue is equivalent to being cleared, but the tasks in localTasks have not yet been executed. In extreme cases, the number of tasks queued by each SingleThreadExecutor will double. I guess the purpose of using localTasks is to avoid lock contention and improve performance, which is fine. But perhaps we should consider solving the problem of doubling the number of queued tasks.

***Screenshots***
image
image
image

***Additional context***

nodece:
Fix #4465

### Motivaction

In `SingleThreadExecutor`, the `runner` drains all tasks from the `queue` into `localTasks`. Although the `queue` is empty in memory at this point, the tasks are still pending execution in `localTasks`—so logically, the queue is still "full."

Calling `execute()` during this phase should *not* enqueue a new `runnable` into the `queue`, as doing so would exceed the intended capacity. This can lead to increased memory usage and potential OutOfMemory (OOM) issues.

### Changes

To address this, we introduce a variable to track the total number of pending runnables. This counter is used to control whether a new task should be added:

- The counter is incremented when a runnable is added to the queue.

- The counter is decremented when a runnable is actually executed.

This ensures accurate tracking of pending tasks and prevents overfilling the logical task queue.

### DIFF
--- a/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/SingleThreadExecutor.java
+++ b/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/SingleThreadExecutor.java
@@ -245,15 +245,15 @@ public class SingleThreadExecutor extends AbstractExecutorService implements Exe
                     }
                 }
             } else {
-                int permits = runnable != null ? 1 : runnableList.size();
-                incrementPendingTaskCount(permits);
+                int count = runnable != null ? 1 : runnableList.size();
+                incrementPendingTaskCount(count);
                 boolean success = hasSingle
                         ? queue.offer(runnable)
                         : queue.addAll(runnableList);
                 if (success) {
-                    tasksCount.add(permits);
+                    tasksCount.add(count);
                 } else {
-                    decrementPendingTaskCount(permits);
+                    decrementPendingTaskCount(count);
                     reject();
                 }
             }
@@ -271,8 +271,10 @@ public class SingleThreadExecutor extends AbstractExecutorService implements Exe
             throw new IllegalArgumentException("Count must be non-negative");
         }
 
-        if (pendingTaskCountUpdater.addAndGet(this, count) > maxQueueCapacity) {
-            decrementPendingTaskCount(count);
+        int oldPendingTaskCount = pendingTaskCountUpdater.getAndAccumulate(this, count,
+                (curr, inc) -> (curr + inc > maxQueueCapacity) ? curr : curr + inc);
+
+        if (oldPendingTaskCount + count > maxQueueCapacity) {
             reject();
         }
     }
--- a/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/SingleThreadExecutor.java
+++ b/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/SingleThreadExecutor.java
@@ -57,9 +57,9 @@ public class SingleThreadExecutor extends AbstractExecutorService implements Exe
     private final LongAdder tasksFailed = new LongAdder();
 
     private final int maxQueueCapacity;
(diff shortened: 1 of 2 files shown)


dlg99 (approved):

nodece:
@lhotari Could you have a chance to review?

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/SingleThreadExecutor.java
```diff
@@ -213,18 +233,30 @@ public void execute(Runnable r) {
                 queue.put(r);
                 tasksCount.increment();
             } else {
-                if (queue.offer(r)) {
-                    tasksCount.increment();
+                int delta = r != null ? 1 : runnableList.size();
+                validateQueueCapacity(delta);
```
StevenLuMT, line 237: image
If a null is passed in here, will runnableList.size() report a NullPointerException? This may affect the function of public void execute(Runnable r)

StevenLuMT (changes requested):
Boundary value confirmation is required

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/SingleThreadExecutor.java
nodece: Done.

StevenLuMT (approved):
LGTM

xiezhx9 (commented):
squash into one commit might be better

nodece:
> squash into one commit might be better

The code owner can use squash merge to resolve this.

bookkeeper-common/src/main/java/org/apache/bookkeeper/common/util/SingleThreadExecutor.java
```diff
@@ -204,27 +216,89 @@ public long getFailedTasksCount() {
 
     @Override
     public void execute(Runnable r) {
+        executeRunnableOrList(r, null);
```
xiezhx9, line 219: How about making `executeRunnableOrList(r, null)` to `executeRunnableOrList(null, toList(r))`. This may avoid special-case handling for single runnable.

nodece: @xiezhx9 `toList()` is an expensive operation and should be avoided in performance-critical paths, `executeRunnableOrList()` is used only for testing purposes, so this usage is acceptable.

xiezhx9 (approved):

nodece:
@lhotari @merlimat Could you have a chance to review?

StevenLuMT (approved):

jiazhai (approved):
+1. LTGM. Good Catch.

PR #4490, opened 2024-08-25
[fix]Decommission command throws KeeperErrorCode exception when autoRecovery is disabled

### LINKED ISSUE #4489: Decommission command throws KeeperErrorCode exception when autoRecovery is disabled
**BUG REPORT**

***Describe the bug***

When autoRecovery has never been enabled, the autoRecovery component will not be loaded. Audit nodes will not be created on Zookeeper, so an exception will occur when executing Decommission.

***To Reproduce***

1. Deploy a Bookkeeper cluster but do not start the process
2. Disabled AutoRecovery
3. Start the process
4. Execute decommission command

***Expected behavior***
Print `Autorecovery is disabled. So giving up`  instead of Exception

***Screenshots***
QQ_1724591198253

315157973:
Fix #4489

### Motivation
 Fix decommission command throws KeeperErrorCode exception when autoRecovery is disabled

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperAdmin.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperAdmin.java
@@ -1553,7 +1553,16 @@ public class BookKeeperAdmin implements AutoCloseable {
             throw new UnavailableException("Autorecovery is disabled. So giving up!");
         }
 
-        BookieId auditorId = getLedgerAuditorManager().getCurrentAuditor();
+        BookieId auditorId = null;
+        try {
+            auditorId = getLedgerAuditorManager().getCurrentAuditor();
+        } catch (IOException e) {
+            if (e.getCause() instanceof KeeperException.NoNodeException) {
+                LOG.error("Can not find node for {}", e.getCause().getMessage());
+                throw new UnavailableException("Autorecovery is disabled. So giving up!");
+            }
+            throw e;
+        }
         if (auditorId == null) {
             LOG.error("No auditor elected, though Autorecovery is enabled. So giving up.");
             throw new UnavailableException("No auditor elected, though Autorecovery is enabled. So giving up.");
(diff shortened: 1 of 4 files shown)


dlg99 (approved):

shoothzj (approved):

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperAdmin.java
```diff
@@ -1553,7 +1553,16 @@ public void triggerAudit()
             throw new UnavailableException("Autorecovery is disabled. So giving up!");
         }
 
-        BookieId auditorId = getLedgerAuditorManager().getCurrentAuditor();
+        BookieId auditorId = null;
+        try {
+            auditorId = getLedgerAuditorManager().getCurrentAuditor();
+        } catch (IOException e) {
+            if (e.getCause() instanceof KeeperException.NoNodeException) {
+                LOG.error("Can not find node for {}", e.getCause().getMessage());
```
shoothzj, line 1561:
```suggestion
                LOG.error("Unable to find Zookeeper node: {}", e.getCause().getMessage());
```

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeperAdmin.java
```diff
@@ -1553,7 +1553,16 @@ public void triggerAudit()
...
+                throw new UnavailableException("Autorecovery is disabled. So giving up!");
```
shoothzj, line 1562:
```suggestion
                throw new UnavailableException("Autorecovery is disabled due to missing Zookeeper node. Aborting recovery!");
```

shoothzj (commented):
I suggest to modify the error message to be more formal.

shoothzj (approved):

315157973:
@dlg99 PTAL

PR #4494, opened 2024-08-29
[Cleanup] Definition of groupId is redundant, because it's inherited from the parent

liangyepianzhou:
### Motivation
Clean up the pom files. Definition of groupId is redundant, because it's inherited from the parent.

### Changes
Delete redundant group id.


StevenLuMT (approved):
good job, LGTM

dlg99 (approved):

PR #4513, opened 2024-10-10
Fix check read failed entry memory leak issue.

horizonzy:
When checking the read failed entry, it doesn't release the data when read successfully.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/ReplicationWorker.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/ReplicationWorker.java
@@ -61,6 +61,7 @@ import org.apache.bookkeeper.client.BookKeeper;
 import org.apache.bookkeeper.client.BookKeeperAdmin;
 import org.apache.bookkeeper.client.EnsemblePlacementPolicy;
 import org.apache.bookkeeper.client.LedgerChecker;
+import org.apache.bookkeeper.client.LedgerEntry;
 import org.apache.bookkeeper.client.LedgerFragment;
 import org.apache.bookkeeper.client.LedgerHandle;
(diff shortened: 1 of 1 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/ReplicationWorker.java
```diff
@@ -351,6 +352,8 @@ private boolean tryReadingFaultyEntries(LedgerHandle lh, LedgerFragment ledgerFr
             lh.asyncReadEntries(entryIdToRead, entryIdToRead, (rc, ledHan, seq, ctx) -> {
                 long thisEntryId = (Long) ctx;
                 if (rc == BKException.Code.OK) {
+                    LedgerEntry entry = seq.nextElement();
```
liangyepianzhou, line 355: Is it possible for `seq` to have multiple elements? It would be better to check with `hasMoreElements` here.

StevenLuMT: I think 
1. hasMoreElements needs to be judged.
2. at the same time, if (rc == BKException.Code.OK) is true or not, this release needs to be added.
3. then I suggest adding a testcase to cover it.

horizonzy: > Is it possible for seq to have multiple elements? It would be better to check with hasMoreElements here.

Make sense.

> at the same time, if (rc == BKException.Code.OK) is true or not, this release needs to be added.

For this case, if the `rc == BKException.Code.OK` is false, the seq will be null, we don't need to release the seq.

StevenLuMT (approved):
Good jobs

hangc0276 (approved):
Good job!

PR #4520, opened 2024-10-31
[improve][conf] Change the default value of readOnlyModeOnAnyDiskFullEnabled to true

liangyepianzhou:
### Motivation
The https://github.com/apache/bookkeeper/pull/3212  import a feature: convert bookie to read-only mode when any ledger disks is full. 
In our practice, we found that when the bookie disk is full, if it is not changed to Read Only mode, a large number of messages will fail to be written to the ledger disk. And because the bookie status is still available at this time, no alarm is given to the maintenance personnel, resulting in slow access for the maintenance personnel.

That feature has been in place for some time. It is safe to enable it default.

PS: It was mentioned in the PR discussion that an email discussion should be submitted, but no follow-up was done. [0]
[0] - https://github.com/apache/bookkeeper/pull/3212/files#r857122376

### Changes
Change the default value of readOnlyModeOnAnyDiskFullEnabled to true

### DIFF
--- a/conf/bk_server.conf
+++ b/conf/bk_server.conf
@@ -182,7 +182,7 @@ extraServerComponents=
 # to read-only mode and serve only read requests. When all disks recovered,
 # the bookie will be converted to read-write mode.Otherwise it will obey the `readOnlyModeEnabled` behavior.
 # By default this will be disabled.
-# readOnlyModeOnAnyDiskFullEnabled=false
+# readOnlyModeOnAnyDiskFullEnabled=true
 
 #############################################################################
 ## Netty server settings
--- a/conf/bk_server.conf
+++ b/conf/bk_server.conf
@@ -181,7 +181,7 @@ extraServerComponents=
 # If "readOnlyModeOnAnyDiskFullEnabled=true" then on any ledger disks full, bookie will be converted
 # to read-only mode and serve only read requests. When all disks recovered,
 # the bookie will be converted to read-write mode.Otherwise it will obey the `readOnlyModeEnabled` behavior.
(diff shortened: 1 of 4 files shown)


eolivelli:
I agree with this change but I cannot find the discussion on the mailing list

liangyepianzhou:
> I agree with this change but I cannot find the discussion on the mailing list

Thank you for your support. I apologize for not starting the discussion on the mailing list earlier due to my busy schedule. I have just sent out the discussion email, and you can find it here: https://lists.apache.org/thread/cvspxb4d1ryy4wo1j2k5r1qr9fcjsd42

StevenLuMT (approved):
LGTM

AnonHxy (approved):

hezhangjian (approved):

shoothzj:
@lhotari currently, is there any labels to track release note updates?

lhotari (approved):
LGTM

StevenLuMT:
image
please update the testcase : org.apache.bookkeeper.bookie.LedgerDirsManagerTest

conf/bk_server.conf
```diff
@@ -182,7 +182,7 @@ extraServerComponents=
 # to read-only mode and serve only read requests. When all disks recovered,
 # the bookie will be converted to read-write mode.Otherwise it will obey the `readOnlyModeEnabled` behavior.
 # By default this will be disabled.
```
StevenLuMT, line 184: # By default this will be enable.

StevenLuMT (approved):

PR #4522, opened 2024-11-11
Fix region aware placement policy disk weight dose not update.

dragonls:
### Motivation

We are using region aware policy in our Pulsar cluster and enable the disk weight base placement feature.
> `ensemblePlacementPolicy=org.apache.bookkeeper.client.RegionAwareEnsemblePlacementPolicy`
> `bookkeeperDiskWeightBasedPlacementEnabled=true`

There are different CVMs with different free disk space in a region, but we found that the traffic still seems to be balanced and is not distributed according to the weight of the free disk space.

The root cause of this case is that the disk weight information actually used to allocate bookies in `RegionAwareEnsemblePlacementPolicy` has not been updated at all.

The `RegionAwareEnsemblePlacementPolicy` actually choose bookies for ensemble by `RackawareEnsemblePlacementPolicy` of each region in `perRegionPlacement`, see `org.apache.bookkeeper.client.RegionAwareEnsemblePlacementPolicy#newEnsemble`. `RegionAwareEnsemblePlacementPolicy.updateBookieInfo` dose not update the `bookieInfoMap` in `RackawareEnsemblePlacementPolicy` of each region in `perRegionPlacement`, which results that the disk weight information actually used to allocate bookies in `RegionAwareEnsemblePlacementPolicy` has not been updated at all.

### Changes

1. Overwrites the `updateBookieInfo` in `RegionAwareEnsemblePlacementPolicy`, update the bookie info in `RackawareEnsemblePlacementPolicy` of each region in `perRegionPlacement`.
2. Add test cases `org.apache.bookkeeper.client.TestRegionAwareEnsemblePlacementPolicy#testRegionsWithDifferentDiskWeight`.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RegionAwareEnsemblePlacementPolicy.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RegionAwareEnsemblePlacementPolicy.java
@@ -663,4 +663,12 @@ public class RegionAwareEnsemblePlacementPolicy extends RackawareEnsemblePlaceme
          */
         return PlacementPolicyAdherence.MEETS_STRICT;
     }
+
+    @Override
+    public void updateBookieInfo(Map<BookieId, BookieInfoReader.BookieInfo> bookieInfoMap) {
+        super.updateBookieInfo(bookieInfoMap);
+        for (TopologyAwareEnsemblePlacementPolicy policy : perRegionPlacement.values()) {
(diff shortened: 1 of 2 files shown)


dlg99 (approved):
Looks good.
Thank you for the fix.

StevenLuMT (approved):
Good jobs

StevenLuMT:

liangyepianzhou (approved):
LGTM

hangc0276 (commented):
Good job!

PR #3989, opened 2023-06-15
Change pendingDeletedLedgers as ConcurrentHashSet

AnonHxy:
Descriptions of the changes in this PR:

### Motivation

Improve remove performance  of `LedgerMetadataIndex#pendingDeletedLedgers`.

Currently the `pendingDeletedLedgers` is a `ConcurrentLinkedQueue`,  and we just call `remove` and `add` and `poll` method of this queue.  And it seems that we need not keep the order when poll items from it. 

Removing items from  `ConcurrentLinkedQueue` need traversal the queue. Changing the `LedgerMetadataIndex#pendingDeletedLedgers` as a `ConcurrentHashSet` will improve the remove performance.

### Changes

Changing the `LedgerMetadataIndex#pendingDeletedLedgers` as a `ConcurrentHashSet`

PR #4525, opened 2024-11-15
fix pendingDeletedLedgers do not remove ledger error

TakaHiR07:
### Motivation

fix this pr https://github.com/apache/bookkeeper/pull/3989 cause ledger not remove in pendingDeletedLedgers.

Previous would do poll(), but later just do forEach()

### Changes

keep the same logic before pr-3989.

Alternative fix is revert previous pr.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/LedgerMetadataIndex.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/LedgerMetadataIndex.java
@@ -339,6 +339,7 @@ public class LedgerMetadataIndex implements Closeable {
                 key.set(ledgerId);
                 ledgersDb.delete(key.array);
                 ++deletedLedgers;
+                pendingDeletedLedgers.remove(ledgerId);
             }
 
             if (log.isDebugEnabled()) {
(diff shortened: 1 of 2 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/LedgerMetadataIndex.java
```diff
@@ -345,6 +345,7 @@ public void removeDeletedLedgers() throws IOException {
                 log.debug("Persisting deletes of ledgers {}", deletedLedgers);
             }
 
+            pendingDeletedLedgers.clear();
```
AnonHxy, line 348: Thanks for your fix.  But I  think  we should call   `pendingDeletedLedgers.remove(ledgerId)`  after line342.
`pendingDeletedLedgers` cloud being adding when call `clear()`,  cause some ledgers not have  deleted form ledgersDB but  clear from `pendingDeletedLedgers `.

By the way.  `pendingDeletedLedgers ` is a `Sets.newConcurrentHashSet()`.   I think it will not concurrent error.

TakaHiR07: that's ok

AnonHxy (approved):

dlg99 (approved):

hezhangjian (commented):
is there any chance we can add unit test for this?

TakaHiR07:
> is there any chance we can add unit test for this?

you can see the new code

hezhangjian (approved):

StevenLuMT (approved):
Good jobs

StevenLuMT:
image
Checkstyle fail,please update @TakaHiR07

hezhangjian (approved):

PR #4500, opened 2024-09-02
BP-68: Delete cookie as part of decommission API.

Karthickramk:
Descriptions of the changes in this PR:

New BP for deleting cookie as part of bookkeeper decommission API
https://github.com/apache/bookkeeper/issues/4499
Proposal PR: WIP

Issue #4499, opened 2024-09-02
BP-68: Delete cookie as part of decommission API.


Karthickramk:
**BP**

This is the master ticket for tracking BP-68 :

Adding cookie deletion login in Bookkeeper decommission API.

Proposal PR - TBD

SongOf:
Is the code development completed?

SongOf:
Is there any plan to add this part?

SongOf:
Hello, the issure proposed by BP-68 can solve the problem of cookie cleanup in our production environment. So I implemented the code according to your BP-68. The code for BP-68 is implemented in this pr(https://github.com/apache/bookkeeper/pull/4592).

PR #4541, opened 2024-12-30
Adds the bookie cookie service for http api

SongOf:

Motivation
It would be convenient to support get, or delete the cookie of bookie with REST API, just like what does in the shell command.

Changes
Adds an API to get, delete the cookie of bookie at /api/v1/bookie/cookie.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/server/http/service/BookieCookieService.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/server/http/service/BookieCookieService.java
@@ -1,0 +1,99 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.bookkeeper.server.http.service;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+import java.net.UnknownHostException;
+import java.util.Map;
+import org.apache.bookkeeper.bookie.BookieException;
+import org.apache.bookkeeper.bookie.Cookie;
+import org.apache.bookkeeper.conf.ServerConfiguration;
+import org.apache.bookkeeper.http.HttpServer;
+import org.apache.bookkeeper.http.service.HttpEndpointService;
+import org.apache.bookkeeper.http.service.HttpServiceRequest;
+import org.apache.bookkeeper.http.service.HttpServiceResponse;
+import org.apache.bookkeeper.meta.MetadataDrivers;
+import org.apache.bookkeeper.net.BookieId;
+import org.apache.bookkeeper.net.BookieSocketAddress;
+import org.apache.bookkeeper.versioning.LongVersion;
+import org.apache.bookkeeper.versioning.Version;
+import org.apache.bookkeeper.versioning.Versioned;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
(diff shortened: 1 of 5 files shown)


congbobo184:
@zymap @eolivelli @hangc0276 @horizonzy  please help review this pr, thanks.

liangyuanpeng:
seems like it is working for BP-68 https://github.com/apache/bookkeeper/pull/4500

SongOf:
> seems like it is working for BP-68
> 
> * [BP-68: Delete cookie as part of decommission API. #4499](https://github.com/apache/bookkeeper/issues/4499)
> * [BP-68: Delete cookie as part of decommission API. #4500](https://github.com/apache/bookkeeper/pull/4500)



> seems like it is working for BP-68
> 
> * [BP-68: Delete cookie as part of decommission API. #4499](https://github.com/apache/bookkeeper/issues/4499)
> * [BP-68: Delete cookie as part of decommission API. #4500](https://github.com/apache/bookkeeper/pull/4500)

Yes, there are two http apis. But both are used to delete the decommissioned bookie's cookie information. BP-68 will delete cookie as part of decommission API. This API("/api/v1/bookie/cookie") is invoked by the operation and maintenance platform to delete cookies after detecting that the bookie is decommissioned.

bookkeeper-server/src/main/java/org/apache/bookkeeper/server/http/service/BookieCookieService.java
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        BookieId bookieId = BookieId.parse(bookieIdStr);
+        return MetadataDrivers.runFunctionWithRegistrationManager(conf, registrationManager -> {
+            try {
+                switch (request.getMethod()) {
+                    case GET:
+                        Versioned<Cookie> cookie = Cookie.readFromRegistrationManager(registrationManager, bookieId);
+                        return new HttpServiceResponse(cookie.toString(), HttpServer.StatusCode.OK);
+                    case DELETE:
```
StevenLuMT, line 51: There is nothing wrong with adding new interfaces, 

but I think deleting cookies is a bit risky and does not comply with data integrity specifications. Although this is available in the old Command command, I suggest you still go through the node offline process `./bin/bookkeeper shell decommissionbookie [-bookieid <bookieaddress>]`

StevenLuMT (approved):
LGTM

bookkeeper-server/src/main/java/org/apache/bookkeeper/server/http/service/BookieCookieService.java
liangyepianzhou, line 51: Why not add a method to add a cookie while we have a method to delete a cookie.

SongOf: Agreed, the caller should call this api only after bookie decommission is completed.

SongOf: Already added

StevenLuMT:
ci failed,please check @SongOf

dlg99 (commented):
Deleting a cookie without validation that decomm is done can be an expensive operator error. I think the command must check for that (as with node decommission, make sure node is offline). So I fully support this concern https://github.com/apache/bookkeeper/pull/4541/files#r1955442058 and disagree with idea of making it an operator's problem.

Alternatively, we can expose `recover` command as a sequence of
* switch target node into the read-only mode
* run recover command
* run recover command (double-check) with the cookie delete option

the gotcha here is that only one recover must run at a time in my experience, but I think the same is valid for decomm command.

eolivelli (changes requested):
This PR is adding an API that can execute  only if the service is up and running. 


It seems dangerous to override the cookies or even deleting it while the service is alive. 

What's your usecase?

SongOf:
> This PR is adding an API that can execute only if the service is up and running.
> 
> It seems dangerous to override the cookies or even deleting it while the service is alive.
> 
> What's your usecase?

my user case:
1、bookie A has been decommissioned by /api/v1/autorecovery/decommission.
2、check decommission has been completed by /api/v1/autorecovery/list_under_replicated_ledger. But the cookie of bookie A is not deleted.
3、The new bookie container is expanded, but has the same IP as bookie A. As a result, the new bookie cannot run normally.
Therefore, after bookie A completes decommission, its cookies should also be deleted.

StevenLuMT:
I think this may be an abnormal case of decommission, but you want to open the cookie interface now, which is not a good way. Covering up the error will only cover up the problem. It is better to check why decommission did not clean up the cookies and solve the problem from the source.

StevenLuMT:
I suggest temporarily closing this PR and reopening it later if necessary.

PR #3653, opened 2022-11-18
Bring back deleteRange for RocksDB to improve location delete performance

hangc0276:
### Motivation
The entry log location index deletion is deleted in batches one by one currently, and it will have low performance. Refer to: https://github.com/apache/bookkeeper/pull/3646

Matteo has introduced deleteRange API a few years ago, but rollback due to RocksDB delete ranges bug. https://github.com/apache/bookkeeper/pull/1620.  The RocksDB bug has been addressed since 5.18.0 https://github.com/facebook/rocksdb/blob/main/HISTORY.md#5180-2018-11-30. We can bring the `deleteRange` API back to improve the entry log location deletion performance.

### Changes
Bring `deleteRange` API back for entry log location deletion.

PR #4555, opened 2025-02-20
Fix SST files not being cleaned up in the locations folder

### LINKED ISSUE #4554: [BUG]Large number of SST files in locations not being cleared after upgrade
**BUG REPORT**

***Describe the bug***
After upgrading from Pulsar 2.8 to LTS 3.0+, with the corresponding BookKeeper version upgraded from 4.14.4 to 4.16.6, the system initially runs well. However, when the throughput increases to a certain level, such as 100MB of writes per node, and after running for a period of time, I observe that the system's physical memory usage increases linearly. Meanwhile, the BookKeeper JVM itself does not show significant memory changes, and there are a large number of SST files in BookKeeper's locations that remain uncleared.

***To Reproduce***

1. Deploy a minimal Pulsar cluster.
2. Configure the message retention time in the Pulsar broker for the namespace, setting both retention and TTL to 6 hours.
3. Use the Pulsar perf tool to write data, maintaining a throughput of 30MB/s or higher per BookKeeper node. The higher the throughput, the easier it is to reproduce the issue.
4. After running continuously for a few days, you will observe that the earliest *.log files in BookKeeper are from 6 hours ago, but the earliest files in the locations directory are still from several days ago and cannot be cleared.

***Expected behavior***
The SST files and entry logs should both be retained for only 6 hours, as configured. There should not be a large accumulation of SST files, which can negatively impact query performance, increase disk storage pressure, and lead to physical memory usage issues.

***Screenshots***
As shown in the figure, the entryLog files are retained for a maximum of 6 hours.
Image
However, the oldest SST files in the locations directory remain uncleared.
Image

***Additional context***

danpi:

Fix #4554

### Motivation
Resolve issue with SST files remaining uncleared in the locations folder.

### Changes

1. Reuse the cleanup logic of `GarbageCollectorThread` to trigger entryLocationCompact() when a major compaction occurs.
2. Add the `entryLocationCompaction` configuration option to control the above behavior, disabled by default.


danpi:
I believe this issue is related to optimization #3653.

When using `deleteRange` to replace the one-by-one deletion method, only one action is written to RocksDB. If RocksDB is performing tiered compaction, this action is not guaranteed to be picked up, causing SST files to remain undeleted.

To resolve this issue:
Option 1: Revert to the previous logic of iteratively deleting each key.
Option 2: Use compactRange to forcefully clean up levels L0 to L6.
I chose to perform a compactRange synchronously when executing a major compaction.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -470,6 +473,10 @@ public void runWithFlags(boolean force, boolean suspendMajor, boolean suspendMin
                     gcStats.getMajorCompactionCounter().inc();
                     majorCompacting.set(false);
                 }
+                if (entryLocationCompaction) {
```
hangc0276, line 476: I suggest moving the entryLocationCompact out of the major or minor compaction scope and using a separate interval to control it.  Remove the `entryLocationCompaction` flag and use the entryLocationCompactionInterval instead. The default value is `-1`, which means disable. We can set it to 1 week by default in the future after verified this operation doesn't introduce a huge impact on performance.

The major or minor compaction has many impact factors, such as suspension, force, and interval.  It will make the entryLocation  compaction behavior unpredictable

hangc0276:
> I believe this issue is related to optimization #3653.
> 
> When using `deleteRange` to replace the one-by-one deletion method, only one action is written to RocksDB. If RocksDB is performing tiered compaction, this action is not guaranteed to be picked up, causing SST files to remain undeleted.
> 
> To resolve this issue: Option 1: Revert to the previous logic of iteratively deleting each key. Option 2: Use compactRange to forcefully clean up levels L0 to L6. I chose to perform a compactRange synchronously when executing a major compaction.

@danpi I think it should be a bug in RocksDB. We can use option 2

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -470,6 +473,10 @@ public void runWithFlags(boolean force, boolean suspendMajor, boolean suspendMin
...
+                    // submit entryLocation compaction task
+                    ledgerStorage.entryLocationCompact();
```
hangc0276, line 478: Please add a log to show entryLocationCompact is triggered.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
danpi: Great suggestion! Currently, the execution of `entryLocationCompact` is indeed coupled with the triggering of `majorCompaction`. I will modify it so that `entryLocationCompact` execution is only based on the configured trigger interval.

danpi:
@hangc0276 Addressed the feedback above. PTAL, thanks.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -87,6 +87,10 @@ public class GarbageCollectorThread implements Runnable {
     long majorCompactionMaxTimeMillis;
     long lastMajorCompactionTime;
 
+    boolean enableEntryLocationCompaction = false;
```
hangc0276, line 90: Remove this one and use `entryLocationCompactionInterval > 0` instead?

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -489,6 +505,17 @@ public void runWithFlags(boolean force, boolean suspendMajor, boolean suspendMin
                     minorCompacting.set(false);
                 }
             }
+            if (enableEntryLocationCompaction && (curTime - lastEntryLocationCompactionTime
```
hangc0276, line 508: We'd better introduce a random factor for it. If we roll restart the BookKeeper cluster, and all the bookies start time are the same, which will make all the bookies triggered RocksDB compaction at the same time and impact the read latency.

hangc0276: The random factor can be any time in [0, entryLocationCompactionInterval]

hangc0276:
> @hangc0276 Addressed the feedback above. PTAL, thanks.

@danpi Thanks for your contribution. I left more comments; please take a look. Thanks.

danpi:
> > @hangc0276 Addressed the feedback above. PTAL, thanks.
> 
> @danpi Thanks for your contribution. I left more comments; please take a look. Thanks.

@hangc0276 Add `randomCompactionDelay` to avoid all the bookies triggering compaction simultaneously. PTAL, thanks.

hangc0276 (approved):

hangc0276:
@danpi Would you please fix the check style issue?
```
Error:  /home/runner/work/bookkeeper/bookkeeper/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java:220:30: '=' is not preceded with whitespace. [WhitespaceAround]
Error:  /home/runner/work/bookkeeper/bookkeeper/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java:301: Line is longer than 120 characters (found 121). [LineLength]
Error:  /home/runner/work/bookkeeper/bookkeeper/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java:3243: Line is longer than 120 characters (found 122). [LineLength]
Audit done.
[INFO] There are 3 errors reported by Checkstyle 9.3 with buildtools/src/main/resources/bookkeeper/checkstyle.xml ruleset.
Error:  src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java:[220,30] (whitespace) WhitespaceAround: '=' is not preceded with whitespace.
Error:  src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java:[301] (sizes) LineLength: Line is longer than 120 characters (found 121).
Error:  src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java:[3243] (sizes) LineLength: Line is longer than 120 characters (found 122).
```

horizonzy:
@danpi Could you help to fix the checkstyle.

dlg99 (approved):

PR #4407, opened 2024-05-30
Fix RocksDB configuration path handling on Windows

shoothzj:
### Motivation

This PR addresses the issue where RocksDB configurations fail to correctly resolve paths on Windows systems in the `ServerConfiguration` class. See windows daily build failed in https://github.com/apache/bookkeeper/actions/runs/9294883674

> Note: Currently, windows is limited to some development jobs, there are still problems to be solved

### Changes

- Updated `ServerConfiguration.java` to utilize `java.nio.file.Paths` for path normalization and resolution.
- Refactored the `getDefaultRocksDBConf`, `getEntryLocationRocksdbConf`, and `getLedgerMetadataRocksdbConf` methods to use a new method `getFilePath`.

PR #4560, opened 2025-03-05
Fix and improve locating RocksDB config files

lhotari:
### Motivation

The solution to locate RocksDB config files has been broken since  #4407 in master branch.

### Changes

- fix issue in locating the RocksDB config files. 
  - Add proper comments to explain the reason why the method exists
  - Fixes also the issue on Windows as attempted in #4407


dlg99 (approved):
LGTM but since it is fixing a bug I think it deserves a test that reproes the problem

StevenLuMT:
Well, I added some test cases for the configuration of rocksdb before.
I think @lhotari you can add some test cases for the abnormal scenarios of this change.

eolivelli (approved):

PR #4578, opened 2025-04-10
fix: Every time use KeyValueStorage.Batch, close it in finally

StevenLuMT:

### Motivation

The rest of the places where Batch is used have been changed: use KeyValueStorage.Batch, close it in finally.

This modification is not caused by a fault. I just found that the code is not written rigorously enough and is prone to bugs when I was combing through it.

### Changes

1.  The rest of the places where Batch is used have been changed: use KeyValueStorage.Batch, close it in finally.


hezhangjian:
LGTM

PR #4581, opened 2025-04-16
Fix the coredump that occurs when calling KeyValueStorageRocksDB.count after rocksdb has been closed

zhaizhibo:

### Motivation

Fix https://github.com/apache/bookkeeper/issues/4238

### Changes

Add a boolean status check before KeyValueStorageRocksDB.count()


lhotari:
There's already #4243 to handle this issue.

lhotari:
> @AnonHxy @lhotari To fix this issue, it is unnecessary to add locks in get/put operations, as it would be too heavy. #4238

@zhaizhibo With this PR there could also be a race such as the count operation is in progress while the database gets closed. That would most likely result in a core dump as well. There's a need for a lock. 
Another question is whether all operations are protected or whether it's just the count method that is addressed.

lhotari (approved):
LGTM

lhotari:
@zhaizhibo Please fix the checkstyle issues.

zhaizhibo:
@lhotari fixed.

hezhangjian (approved):

StevenLuMT:
+1

PR #4583, opened 2025-04-16
improve: Resolve hostname only when useHostNameAsBookieID is enabled

nodece:
### Motivation

The existing implementation always resolves the default hostname for the specified network interface, even when `useHostNameAsBookieID` is disabled. This is unnecessary and can introduce resolution failures or delays in environments where DNS is misconfigured or unavailable.

Restricting hostname resolution to only when `useHostNameAsBookieID` is enabled improves efficiency and avoids potential issues related to unnecessary hostname lookups.

### Changes

- Moved hostname resolution logic under the `useHostNameAsBookieID` condition.

- When `useHostNameAsBookieID` is false, the code now directly uses `DNS.getDefaultIP(iface)` to obtain the IP address.


lhotari (approved):
LGTM

nodece:
Alternative: #4588

Issue #4585, opened 2025-04-16
Replace Java DNS resolver with Netty DNS resolver to avoid test timeouts


nodece:
Some BookKeeper tests occasionally fail due to long delays in reverse DNS lookups. The root cause is the use of `InitialDirContext.getAttributes()` in `org.apache.bookkeeper.net.DNS#reverseDns`, which can block for up to 120 seconds if DNS resolution hangs. This leads to flaky or timing-out tests such as:

- https://github.com/apache/bookkeeper/issues/4564
- https://github.com/apache/bookkeeper/issues/4565

To improve reliability, we should replace the current blocking implementation with Netty's asynchronous DNS resolver. This will allow us to set timeouts, avoid blocking test threads, and reduce flakiness.

Test with Netty DNS:

<img width="602" alt="Image" src="https://github.com/user-attachments/assets/a6bf07f7-d9ef-4053-90ff-d303babaae08" />

nodece:
I will make a PR to improve the DNS resolve.

lhotari:
@nodece Adding `-Dsun.net.inetaddr.ttl=1 -Dsun.net.inetaddr.negative.ttl=1` to test JVM args could possible help without switching to Netty DNS resolver. I don't think that DNS resolution itself hangs. It's more likely to be a result of cached addresses. The default negative cache is 10 seconds and default cache is 30 seconds on recent JVM versions.

lhotari:
I created #4586 to tune the DNS TTL settings for tests and for CI

lhotari:
Oh, now I see what there is:
https://github.com/apache/bookkeeper/blob/bcd6b52fcaa9b82cd60dbc3a8583df4ca60796e7/bookkeeper-server/src/main/java/org/apache/bookkeeper/net/DNS.java#L53-L103

@nodece I'd suggest to make this configurable so that this isn't performed at all when reverse lookup is disabled. Having this type of config mandatory in BookKeeper doesn't make much sense to me.

nodece:
@lhotari  It looks like we can get the hostname by the `java.net.InetAddress#getCanonicalHostName`.

lhotari:
@nodece Alternative approach: #4587

nodece:
Alternative approach: #4588

nodece:
Closed by #4588

PR #4586, opened 2025-04-16
Adjust DNS cache expiration in tests to reduce test timeout failures

lhotari:
### Motivation

Some tests fail due to DNS timeouts as explained in #4585.
Some of these failures might be caused by the default negative DNS cache that is 10 seconds by default or by the default positive DNS cache that is 30 seconds by default. It's better to set reduce dns cache settings to 1 second for tests.

### Changes

- pass `-Dsun.net.inetaddr.ttl=1 -Dsun.net.inetaddr.negative.ttl=1` in test JVM arguments to configure the DNS cache settings
- For OpenJDK, it's not possible to override settings in `$JAVA_HOME/conf/security/java.security` with System properties, see https://github.com/openjdk/jdk11u/blob/master/src/java.base/share/classes/sun/net/InetAddressCachePolicy.java . By default there's a negative TTL setting set to 10 seconds.
  - apply respective settings to `$JAVA_HOME/conf/security/java.security` in CI

### DIFF
--- a/.github/workflows/bk-ci.yml
+++ b/.github/workflows/bk-ci.yml
@@ -159,6 +159,13 @@ jobs:
           distribution: 'temurin'
           java-version: 11
 
+      - name: Tune Java DNS TTL settings
+        run: |
+          sudo tee -a $JAVA_HOME/conf/security/java.security <<EOF
+          networkaddress.cache.ttl=1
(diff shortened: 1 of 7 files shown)


StevenLuMT:
rerun failure checks

merlimat (approved):

PR #4593, opened 2025-04-21
should not waitBackOff when replicating open fragment

SongOf:

When replicating a ledger with open fragment, ReplicationWorker.rereplicate(long ledgerIdToReplicate) method returns false.
https://github.com/apache/bookkeeper/blob/25326dcf0e3c51bab4a3e82c1f6f90b064ef6783/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/ReplicationWorker.java#L495-L499

This will cause the replication worker to wait for rwRereplicateBackoffMs ms and be unable to replicate other ledgers. This seriously affects the efficiency of the replication worker.
https://github.com/apache/bookkeeper/blob/25326dcf0e3c51bab4a3e82c1f6f90b064ef6783/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/ReplicationWorker.java#L247-L252

### Motivation

When replicating a ledger in the open state, ReplicationWorker will choose to wait for a while before releasing the lock.
https://github.com/apache/bookkeeper/blob/25326dcf0e3c51bab4a3e82c1f6f90b064ef6783/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/ReplicationWorker.java#L496-L497

Therefore, ReplicationWorker does not need to waitBackOffTime any more after skipping replicating an open ledger.

### Changes

When replicating an open ledger, it no longer waitBackOffTime but immediately continues to replicate the next ledger.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/ReplicationWorker.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/replication/ReplicationWorker.java
@@ -111,6 +111,11 @@ public class ReplicationWorker implements Runnable {
     private final long baseBackoffForLockReleaseOfFailedLedger;
     private final BiConsumer<Long, Long> onReadEntryFailureCallback;
     private final LedgerManager ledgerManager;
+    enum ReplicateResult {
+        SUCCESS,
+        FAILED,
+        SKIP_OPEN_FRAGMENT
+    }
 
     // Expose Stats
     private final StatsLogger statsLogger;
@@ -246,7 +251,7 @@ public class ReplicationWorker implements Runnable {
         workerRunning = true;
         while (workerRunning) {
             try {
-                if (!rereplicate()) {
+                if (rereplicate() == ReplicateResult.FAILED) {
                     LOG.warn("failed while replicating fragments");
                     waitBackOffTime(rwRereplicateBackoffMs);
                 }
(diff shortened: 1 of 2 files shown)


StevenLuMT (approved):
Good jobs

liudezhi2098 (approved):

zymap (changes requested):
It seems a little dangerous that skip the backoff for one case. When you have that case, it possible you have many of ledger into the state. Then the auto recovery will try to recover it without any breaks. So maybe one way to control that is collect the ledgers who is open and release it in the future time. Then you won't block the normal ledger replication and the open ledger still have a backoff there.

SongOf:
> It seems a little dangerous that skip the backoff for one case. When you have that case, it possible you have many of ledger into the state. Then the auto recovery will try to recover it without any breaks. So maybe one way to control that is collect the ledgers who is open and release it in the future time. Then you won't block the normal ledger replication and the open ledger still have a backoff there.

When replicating a ledger in the open state, ReplicationWorker will choose to wait for a while before releasing the lock. openLedgerRereplicationGracePeriod ms in function "deferLedgerLockRelease".

zymap:
But there are still metadata operations before deferring it. If you don't give them a backoff, then the metadata service will suffer huge requests in a short time. 
And deferLedgerLockRelease adds a delay task, it will create a task and schedule it with a delay everytime. If you skip the open ledger outside, it will add many same tasks, which also do not make sense.

Issue #1124, opened 2018-02-06
The bookie runs out of memory while GCThread is extracting all ledger meta data.


kishorvpatil:
This is a bug report with bookkeeper proposal.

**BUG REPORT**

1. Please describe the issue you observed: 
While extracting all ledger log meta data, it is unnecessary it is easier to remove them if ledger no longer exists. This can help same some heap space. We have noticed exception where bookie hits OOM exception due to this issue.

- The bookie runs out of memory while extracting all ledger meta data. The heap dump shows 93% of data being pulled into the method extractMetaData method.

- We expect optimized collection of the LogEntryLedgerData if they are going to be garbage collected anyways.

Proposal PR - #&lt;#1125&gt;

PR #4599, opened 2025-04-29
refactor: Eliminate redundant cleanup logic in entry log extraction

nodece:
### Motivation

The purpose of `extractMetaFromEntryLogs` is to extract metadata, not perform garbage collection. The logic for removing entry logs with no active ledgers is handled separately in `doGcEntryLogs()`. 

### Changes

This change removes the redundant cleanup code from `extractMetaFromEntryLogs` to better separate concerns and avoid duplication.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
@@ -559,7 +559,7 @@ public class GarbageCollectorThread implements Runnable {
     /**
      * Garbage collect those entry loggers which are not associated with any active ledgers.
      */
-    private void doGcEntryLogs() throws EntryLogMetadataMapException {
+    protected void doGcEntryLogs() throws EntryLogMetadataMapException {
         // Get a cumulative count, don't update until complete
         AtomicLong activeEntryLogSizeAcc = new AtomicLong(0L);
         AtomicLong totalEntryLogSizeAcc = new AtomicLong(0L);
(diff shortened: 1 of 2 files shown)


liudezhi2098 (approved):

nodece:
rerun failure checks

nodece:
rerun failure checks

nodece:
rerun failure checks

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
```diff
@@ -559,7 +559,7 @@ private void doGcLedgers() {
     /**
      * Garbage collect those entry loggers which are not associated with any active ledgers.
      */
-    private void doGcEntryLogs() throws EntryLogMetadataMapException {
+    protected void doGcEntryLogs() throws EntryLogMetadataMapException {
```
AnonHxy, line 562: `protectd` could  be removed. Default access modifiers is enough.

zymap (changes requested):
Here is the reason why it did the cleanup: https://github.com/apache/bookkeeper/issues/1124. You could declare a new method to avoid the duplicated code.

nodece:
@StevenLuMT @zymap 

In the current implementation, all metadata is first extracted and stored in a map before performing a full GC pass over the entire dataset. This approach can result in high memory usage and delayed cleanup, especially in environments with large or fragmented metadata stores.

I suggest we should replace the current bulk extraction + GC logic with an incremental streaming approach, as follows:

Current Approach:
```
Map<String, Metadata> allMetadata = extractAllMetadata();
gc(allMetadata);
```

Proposed Approach:
```
for (String key : listAllMetadataKeys()) {
    Metadata metadata = extractMetadata(key);
    gc(metadata);
}
```

Benefits:
- Reduced memory footprint: Avoids holding all metadata in memory at once.
- More timely cleanup: GC is performed immediately after extraction.
- Improved scalability: Handles large metadata sets more gracefully.

hangc0276 (changes requested):
The current logic can cleanup those no active ledger's entrylog files immediately.

PR #4607, opened 2025-05-29
Fix the data loss issue that caused by the wrong entry log header

zymap:
---

# Motivation

We observed numerous errors in the broker that failed to read the ledger from the bookkeeper; although the ledger metadata still exists, it was unable to read from the bookkeeper. After checking the data, we found the ledger located entry log was deleted by the bookkeeper. We have a data loss issue with the bookkeeper.

The entry log file was deleted by the Garbage collector because the entry log file wrote a wrong file header.

And there is an example that the shows the header is wrong:
```
Failed to get ledgers map index from: 82.log : Not all ledgers were found in ledgers map index. expected: -1932430239 -- found: 0 -- entryLogId: 82
```

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/DefaultEntryLogger.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/DefaultEntryLogger.java
@@ -1151,6 +1151,10 @@ public class DefaultEntryLogger implements EntryLogger {
                     + " -- found: " + meta.getLedgersMap().size() + " -- entryLogId: " + entryLogId);
         }
 
+        if (header.ledgersCount == 0) {
+            throw new IOException("No ledgers map found in entryLogId " + entryLogId + ", do scan to double confirm");
+        }
+
         return meta;
     }
 
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLoggerAllocator.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/EntryLoggerAllocator.java
@@ -85,6 +85,7 @@ class EntryLoggerAllocator {
         // within the same JVM. All of these Bookie instances access this header
         // so there can be race conditions when entry logs are rolled over and
         // this header buffer is cleared before writing it into the new logChannel.
+        logfileHeader.setZero(0, DefaultEntryLogger.LOGFILE_HEADER_SIZE);
         logfileHeader.writeBytes("BKLO".getBytes(UTF_8));
         logfileHeader.writeInt(DefaultEntryLogger.HEADER_CURRENT_VERSION);
         logfileHeader.writerIndex(DefaultEntryLogger.LOGFILE_HEADER_SIZE);
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/GarbageCollectorThread.java
@@ -842,11 +842,12 @@ public class GarbageCollectorThread implements Runnable {
                 continue;
             }
 
-            LOG.info("Extracting entry log meta from entryLogId: {}", entryLogId);
 
             try {
                 // Read through the entry log file and extract the entry log meta
                 EntryLogMetadata entryLogMeta = entryLogger.getEntryLogMetadata(entryLogId, throttler);
+                LOG.info("Extracted entry log meta from entryLogId: {}, ledgers {}",
+                    entryLogId, entryLogMeta.getLedgersMap().keys());
                 removeIfLedgerNotExists(entryLogMeta);
                 if (entryLogMeta.isEmpty()) {
                     // This means the entry log is not associated with any active
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/DefaultEntryLogTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/DefaultEntryLogTest.java
@@ -391,6 +391,46 @@ public class DefaultEntryLogTest {
         assertEquals(120, meta.getRemainingSize());
     }
 
+    @Test
+    public void testLedgersMapIsEmpty() throws Exception {
+        // create some entries
+        entryLogger.addEntry(1L, generateEntry(1, 1).nioBuffer());
+        entryLogger.addEntry(3L, generateEntry(3, 1).nioBuffer());
+        entryLogger.addEntry(2L, generateEntry(2, 1).nioBuffer());
+        entryLogger.addEntry(1L, generateEntry(1, 2).nioBuffer());
+        ((EntryLogManagerBase) entryLogger.getEntryLogManager()).createNewLog(DefaultEntryLogger.UNASSIGNED_LEDGERID);
+        entryLogger.close();
+
+        // Rewrite the entry log header to be on V0 format
+        File f = new File(curDir, "0.log");
+        RandomAccessFile raf = new RandomAccessFile(f, "rw");
+        raf.seek(8);
+        // Mock that there is a ledgers map offset but the ledgers count is 0
+        raf.writeLong(40);
+        raf.writeInt(0);
+        raf.close();
+
+        // now see which ledgers are in the log
+        entryLogger = new DefaultEntryLogger(conf, dirsMgr);
+
+        try {
+            entryLogger.extractEntryLogMetadataFromIndex(0L);
+            fail("Should not be possible to recover from ledgers map index");
+        } catch (IOException e) {
+            assertEquals("No ledgers map found in entryLogId 0, do scan to double confirm", e.getMessage());
+        }
+
+        // Public method should succeed by falling back to scanning the file
+        EntryLogMetadata meta = entryLogger.getEntryLogMetadata(0L);
+        LOG.info("Extracted Meta From Entry Log {}", meta);
+        assertEquals(60, meta.getLedgersMap().get(1L));
+        assertEquals(30, meta.getLedgersMap().get(2L));
+        assertEquals(30, meta.getLedgersMap().get(3L));
+        assertFalse(meta.getLedgersMap().containsKey(4L));
+        assertEquals(120, meta.getTotalSize());
+        assertEquals(120, meta.getRemainingSize());
+    }
+
     /**
      * Explicitly try to recover using the ledgers map index at the end of the entry log.
      */


hangc0276 (approved):
Please add a unit test for this fix.

oneby-wang:
老哥给力！我们就是用的默认 DbLedgerStorage，看了下源码确实如你说的。另外按照你的说法，我感觉这条 pull request 没有解决根本问题吧？即便 logFileHeader 初始化未全填充 0，checkPoint 的时候失败了也会 redo，假设没有你说的并发时序问题，logFileHeader 最终也会被填充成正确的值。另外这个并发问题，你知道官方有相应修复的计划吗？貌似加上一些锁，然后再考虑一些时序问题就能解决，我看了下 branch-4.16 貌似并未解决。

nsnycde:
In theory, this issue shouldn't occur with the default DbLedgerStorage. The entry point for DbLedgerStorage's write, garbage collection, and checkpoint threads is getCurrentLogForLedgerForAddEntry, a method that is locked (I'm not sure where else I'm missing information). I encountered this issue with SortedLedgerStorage. In theory, rewriting and synchronizing the EntryLogManagerForSingleEntryLog.createNewLog method with a reason parameter should resolve the issue. This fix was found with SortedLedgerStorage. I haven't seen any official team address this yet.

lhotari:
> Hi, I'm wondering what situation will cause data loss(caused by wrong file header)? I read the source code, it seems the file header is updated by BufferedLogChannel.appendLedgersMap() method.

@oneby-wang @nsnycde I guess that the file might be read by GC before that happens. This PR addresses the issue that there's garbage in the initial header. The original assumption has been that the header would be initialized with zeros. However, that assumption didn't hold with Netty's optimized array allocation as explained in https://github.com/apache/bookkeeper/pull/4607#issuecomment-2918429558 . 
If there's an invalid header in the entry log file, it could get deleted in GC instead of skipped. There's also the chance that the header just happens to contain data the gets interpreted as valid data in GC. At least this is how I understood the problem and the fix.

oneby-wang:
Hi, guys, I think I figured it out, the key point is the following code in this pull request, although setting logfileHeader to zeros is also very important. I'll explain later.
```java
if (header.ledgersCount == 0) {
    throw new IOException("No ledgers map found in entryLogId " + entryLogId + ", do scan to double confirm");
}
```
I readed the journal redo code, I found that bookeeper redo will create a new entryLog file, instead of using the old entryLog file after bookie crashed. So, the old entryLog file still has the uncorrect logFileHeader(not updated by BufferedLogChannel.appendLedgersMap() method).
In DefaultEntryLogger, if getEntryLogMetadata() method returns empty entryLogMeta, the GC thread will delete the entryLog file, which may contain some ledgers can't be deleted.
```java
protected void extractMetaFromEntryLogs() throws EntryLogMetadataMapException {
	// ...
	// Read through the entry log file and extract the entry log meta
	EntryLogMetadata entryLogMeta = entryLogger.getEntryLogMetadata(entryLogId, throttler);
	removeIfLedgerNotExists(entryLogMeta);
	if (entryLogMeta.isEmpty()) {
	    // This means the entry log is not associated with any active
	    // ledgers anymore.
	    // We can remove this entry log file now.
	    LOG.info("Deleting entryLogId {} as it has no active ledgers!", entryLogId);
	    if (removeEntryLog(entryLogId)) {
	        gcStats.getReclaimedSpaceViaDeletes().addCount(entryLogMeta.getTotalSize());
	    } else {
	        gcStats.getReclaimFailedToDelete().inc();
	    }
	}
```
How does this happen? here is the old unfixed code. If logfileHeader.ledgersMapOffset != 0(random value, logfileHeader is uninitialized), and logfileHeader.ledgersCount == 0(random value, logfileHeader is uninitialized), and in while loop read ledgersMapSize < 0(some lucky value in entryLog file random location, interesting huh), the while loop is breaked out, and the entryLogMeta is empty and returned to the caller method. I think we should throw an IOException when ledgersMapSize < 0 instead of breaking out the while loop.
```java
EntryLogMetadata extractEntryLogMetadataFromIndex(long entryLogId) throws IOException {
    Header header = getHeaderForLogId(entryLogId);

    if (header.version < HEADER_V1) {
        throw new IOException("Old log file header without ledgers map on entryLogId " + entryLogId);
    }

    if (header.ledgersMapOffset == 0L) {
        // The index was not stored in the log file (possibly because the bookie crashed before flushing it)
        throw new IOException("No ledgers map index found on entryLogId " + entryLogId);
    }

    if (LOG.isDebugEnabled()) {
        LOG.debug("Recovering ledgers maps for log {} at offset: {}", entryLogId, header.ledgersMapOffset);
    }

    BufferedReadChannel bc = getChannelForLogId(entryLogId);

    // There can be multiple entries containing the various components of the serialized ledgers map
    long offset = header.ledgersMapOffset;
    EntryLogMetadata meta = new EntryLogMetadata(entryLogId);

    final int maxMapSize = LEDGERS_MAP_HEADER_SIZE + LEDGERS_MAP_ENTRY_SIZE * LEDGERS_MAP_MAX_BATCH_SIZE;
    ByteBuf ledgersMap = allocator.directBuffer(maxMapSize);

    try {
        while (offset < bc.size()) {
            // Read ledgers map size
            sizeBuffer.get().clear();
            bc.read(sizeBuffer.get(), offset);

            int ledgersMapSize = sizeBuffer.get().readInt();
            if (ledgersMapSize <= 0) {
                break;
            }
            // Read the index into a buffer
            ledgersMap.clear();
            bc.read(ledgersMap, offset + 4, ledgersMapSize);

            // Discard ledgerId and entryId
            long lid = ledgersMap.readLong();
            if (lid != INVALID_LID) {
                throw new IOException("Cannot deserialize ledgers map from ledger " + lid);
            }

            long entryId = ledgersMap.readLong();
            if (entryId != LEDGERS_MAP_ENTRY_ID) {
                throw new IOException("Cannot deserialize ledgers map from entryId " + entryId);
            }

            // Read the number of ledgers in the current entry batch
            int ledgersCount = ledgersMap.readInt();

            // Extract all (ledger,size) tuples from buffer
            for (int i = 0; i < ledgersCount; i++) {
                long ledgerId = ledgersMap.readLong();
                long size = ledgersMap.readLong();

                if (LOG.isDebugEnabled()) {
                    LOG.debug("Recovering ledgers maps for log {} -- Found ledger: {} with size: {}",
                            entryLogId, ledgerId, size);
                }
                meta.addLedgerSize(ledgerId, size);
            }
            if (ledgersMap.isReadable()) {
                throw new IOException("Invalid entry size when reading ledgers map");
            }

            // Move to next entry, if any
            offset += ledgersMapSize + 4;
        }
    } catch (IndexOutOfBoundsException e) {
        throw new IOException(e);
    } finally {
        ReferenceCountUtil.release(ledgersMap);
    }

    if (meta.getLedgersMap().size() != header.ledgersCount) {
        throw new IOException("Not all ledgers were found in ledgers map index. expected: " + header.ledgersCount
                + " -- found: " + meta.getLedgersMap().size() + " -- entryLogId: " + entryLogId);
    }

    return meta;
}
```
If extractEntryLogMetadataFromIndex() method throw a Throwable, getEntryLogMetadata() method will fall back to extractEntryLogMetadataByScanning() method, which will skip logfileHeader, and scan the rest part of entryLog file.
```java
public EntryLogMetadata getEntryLogMetadata(long entryLogId, AbstractLogCompactor.Throttler throttler)
    throws IOException {
    // First try to extract the EntryLogMetadata from the index, if there's no index then fallback to scanning the
    // entry log
    try {
        return extractEntryLogMetadataFromIndex(entryLogId);
    } catch (FileNotFoundException fne) {
        LOG.warn("Cannot find entry log file {}.log : {}", Long.toHexString(entryLogId), fne.getMessage());
        throw fne;
    } catch (Throwable e) {
        LOG.info("Failed to get ledgers map index from: {}.log : {}", entryLogId, e.getMessage());

        // Fall-back to scanning
        return extractEntryLogMetadataByScanning(entryLogId, throttler);
    }
}
```
I think the key point is the entryLog file lacks metadata validation, eg. CRC validation, so we can't figure it out it's completed with flush or corrupted due to crash? 
And I think the header.ledgersCount == 0 validation code should be writen after header.ledgersMapOffset validation(just for readability).
```java
if (header.ledgersMapOffset == 0L) {
    // The index was not stored in the log file (possibly because the bookie crashed before flushing it)
    throw new IOException("No ledgers map index found on entryLogId " + entryLogId);
}

if (header.ledgersCount == 0) {
    throw new IOException("No ledgers map found in entryLogId " + entryLogId + ", do scan to double confirm");
}
```
If bookie can't be shutdown gracefully, eg. bookie instance crashes, the old entryLog file may be deleted. So, what can we do in old version bookie after crash? First, don't start up bookie instance directly, backup your data, andd then use some hex editor to set ledgersMapOffset to zero, then start up your bookie instance, the GC thread will fall back to extractEntryLogMetadataByScanning() method, which will not delete the entryLog file.

lhotari:
@oneby-wang oh yes, based on your description, you are pointing that this is an issue:

https://github.com/apache/bookkeeper/blob/e80d0318cfdebbc79f37d9120de8df28c8c1c13a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/DefaultEntryLogger.java#L1104-L1106

It seems that this code might explain the negative value:
https://github.com/apache/bookkeeper/blob/e80d0318cfdebbc79f37d9120de8df28c8c1c13a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/DefaultEntryLogger.java#L159-L171

However it's also very strange that it assumes that `ledgersMap` is concurrently modified and has a stable iteration order after concurrent modifications:

https://github.com/apache/bookkeeper/blob/e80d0318cfdebbc79f37d9120de8df28c8c1c13a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/DefaultEntryLogger.java#L143-L188

nsnycde:
> > Hi, I'm wondering what situation will cause data loss(caused by wrong file header)? I read the source code, it seems the file header is updated by BufferedLogChannel.appendLedgersMap() method.
> 
> @oneby-wang @nsnycde I guess that the file might be read by GC before that happens. This PR addresses the issue that there's garbage in the initial header. The original assumption has been that the header would be initialized with zeros. However, that assumption didn't hold with Netty's optimized array allocation as explained in [#4607 (comment)](https://github.com/apache/bookkeeper/pull/4607#issuecomment-2918429558) . If there's an invalid header in the entry log file, it could get deleted in GC instead of skipped. There's also the chance that the header just happens to contain data the gets interpreted as valid data in GC. At least this is how I understood the problem and the fix.

When bk is running normally, gc should not read unflushed logs. You can see that the loop in the code below is entryLogger.getFlushedLogIds()
```
//org.apache.bookkeeper.bookie.GarbageCollectorThread#extractMetaFromEntryLogs

protected void extractMetaFromEntryLogs() throws EntryLogMetadataMapException {
        for (long entryLogId : entryLogger.getFlushedLogIds()) {
            // Comb the current entry log file if it has not already been extracted.
            if (entryLogMetaMap.containsKey(entryLogId)) {
                continue;
            }

            // check whether log file exists or not
            // if it doesn't exist, this log file might have been garbage collected.
            if (!entryLogger.logExists(entryLogId)) {
                continue;
            }

            LOG.info("Extracting entry log meta from entryLogId: {}", entryLogId);

            try {
                // Read through the entry log file and extract the entry log meta
                EntryLogMetadata entryLogMeta = entryLogger.getEntryLogMetadata(entryLogId, throttler);
                removeIfLedgerNotExists(entryLogMeta);
                if (entryLogMeta.isEmpty()) {
                    // This means the entry log is not associated with any active
                    // ledgers anymore.
                    // We can remove this entry log file now.
                    LOG.info("Deleting entryLogId {} as it has no active ledgers!", entryLogId);
                    removeEntryLog(entryLogId);
                    gcStats.getReclaimedSpaceViaDeletes().addCount(entryLogMeta.getTotalSize());
                } else {
                    entryLogMetaMap.put(entryLogId, entryLogMeta);
                }
            } catch (IOException | RuntimeException e) {
                LOG.warn("Premature exception when processing " + entryLogId
                         + " recovery will take care of the problem", e);
            } catch (OutOfMemoryError oome) {
                LOG.warn("OutOfMemoryError when processing {} - skipping the entry log", entryLogId, oome);
            }
        }
    }
```

nsnycde:
> @oneby-wang oh yes, based on your description, you are pointing that this is an issue:
> 
> https://github.com/apache/bookkeeper/blob/e80d0318cfdebbc79f37d9120de8df28c8c1c13a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/DefaultEntryLogger.java#L1104-L1106
> 
> It seems that this code might explain the negative value:
> 
> https://github.com/apache/bookkeeper/blob/e80d0318cfdebbc79f37d9120de8df28c8c1c13a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/DefaultEntryLogger.java#L159-L171
> 
> However it's also very strange that it assumes that `ledgersMap` is concurrently modified and has a stable iteration order after concurrent modifications:
> 
> https://github.com/apache/bookkeeper/blob/e80d0318cfdebbc79f37d9120de8df28c8c1c13a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/DefaultEntryLogger.java#L143-L188

@lhotari @oneby-wang 
Theoretically, judging that ledgerMapSize <= 0 here should not lead to incorrect deletion, because in the above judgment, if the leaderMapIndex data is not written successfully, the offset data pointed to by the logHeader header is 0, and an exception will be thrown directly before this judgment.
> https://github.com/apache/bookkeeper/blob/e80d0318cfdebbc79f37d9120de8df28c8c1c13a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/DefaultEntryLogger.java#L1079-L1082

Moreover, the offset data pointed to by the logHeader header in leaderMapIndex will only be set after the index data is written. If it crashes abnormally, the offset should not be written.
> https://github.com/apache/bookkeeper/blob/e80d0318cfdebbc79f37d9120de8df28c8c1c13a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/DefaultEntryLogger.java#L204-L208

oneby-wang:
> > @oneby-wang oh yes, based on your description, you are pointing that this is an issue:
> > https://github.com/apache/bookkeeper/blob/e80d0318cfdebbc79f37d9120de8df28c8c1c13a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/DefaultEntryLogger.java#L1104-L1106
> > 
> > It seems that this code might explain the negative value:
> > https://github.com/apache/bookkeeper/blob/e80d0318cfdebbc79f37d9120de8df28c8c1c13a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/DefaultEntryLogger.java#L159-L171
> > 
> > However it's also very strange that it assumes that `ledgersMap` is concurrently modified and has a stable iteration order after concurrent modifications:
> > https://github.com/apache/bookkeeper/blob/e80d0318cfdebbc79f37d9120de8df28c8c1c13a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/DefaultEntryLogger.java#L143-L188
> 
> @lhotari @oneby-wang Theoretically, judging that ledgerMapSize <= 0 here should not lead to incorrect deletion, because in the above judgment, if the leaderMapIndex data is not written successfully, the offset data pointed to by the logHeader header is 0, and an exception will be thrown directly before this judgment.
> 
> > https://github.com/apache/bookkeeper/blob/e80d0318cfdebbc79f37d9120de8df28c8c1c13a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/DefaultEntryLogger.java#L1079-L1082
> 
> Moreover, the offset data pointed to by the logHeader header in leaderMapIndex will only be set after the index data is written. If it crashes abnormally, the offset should not be written.
> 
> > https://github.com/apache/bookkeeper/blob/e80d0318cfdebbc79f37d9120de8df28c8c1c13a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/DefaultEntryLogger.java#L204-L208

Before this pull request, leaderMapIndex is a random value in entryLog file, it may probably not zero.

nsnycde:
> > > @oneby-wang oh yes, based on your description, you are pointing that this is an issue:哦，是的，根据您的描述，您指出这是一个问题：
> > > https://github.com/apache/bookkeeper/blob/e80d0318cfdebbc79f37d9120de8df28c8c1c13a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/DefaultEntryLogger.java#L1104-L1106
> > > 
> > > It seems that this code might explain the negative value:看来这段代码可以解释负值：
> > > https://github.com/apache/bookkeeper/blob/e80d0318cfdebbc79f37d9120de8df28c8c1c13a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/DefaultEntryLogger.java#L159-L171
> > > 
> > > However it's also very strange that it assumes that `ledgersMap` is concurrently modified and has a stable iteration order after concurrent modifications:然而它也很奇怪地假设 `ledgersMap` 是并发修改的，并且在并发修改后具有稳定的迭代顺序：
> > > https://github.com/apache/bookkeeper/blob/e80d0318cfdebbc79f37d9120de8df28c8c1c13a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/DefaultEntryLogger.java#L143-L188
> > 
> > 
> > @lhotari @oneby-wang Theoretically, judging that ledgerMapSize <= 0 here should not lead to incorrect deletion, because in the above judgment, if the leaderMapIndex data is not written successfully, the offset data pointed to by the logHeader header is 0, and an exception will be thrown directly before this judgment.理论上来说，这里判断 ledgerMapSize <= 0应该不会导致误删除，因为在上面的判断中，如果leaderMapIndex数据没有写入成功，logHeader头指向的offset数据为0，在这个判断之前会直接抛出异常。
> > > https://github.com/apache/bookkeeper/blob/e80d0318cfdebbc79f37d9120de8df28c8c1c13a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/DefaultEntryLogger.java#L1079-L1082
> > 
> > 
> > Moreover, the offset data pointed to by the logHeader header in leaderMapIndex will only be set after the index data is written. If it crashes abnormally, the offset should not be written.而且 leaderMapIndex 中 logHeader 头指向的偏移量数据只有在写入索引数据后才会设置，如果异常崩溃，则不应该写入偏移量。
> > > https://github.com/apache/bookkeeper/blob/e80d0318cfdebbc79f37d9120de8df28c8c1c13a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/DefaultEntryLogger.java#L204-L208
> 
> Before this pull request, leaderMapIndex is a random value in entryLog file, it may probably not zero.在此拉取请求之前，leaderMapIndex 是 entryLog 文件中的一个随机值，它可能不为零。

You are right, I ignored the fix of this issue

PR #4609, opened 2025-05-30
Fix Memory Leak In Netty Recycler of Bookie Client

TakaHiR07:
Main Issue: https://github.com/apache/pulsar/issues/24355

### Motivation

There is a huge memory leak risk in netty recycler of bookie client . The issue and analysis is show here: https://github.com/apache/pulsar/issues/24355

### Changes

Currently, each perChannelBookieClient has its own recycler.

After modification, All the completionKey and completionValue is no longer the inner class of perChannelBookieClient, and each perChannelBookieClient would share the same recycler. 

Notice: after the modification, it is more important to set a appropriate `io.netty.recycler.maxCapacityPerThread` config. Because I have tested that if we do not rely on cache object, the read and write performance would have a little decrease.    Since all the channel in bkClient share the same recycler, we should make the config = add entry qps * add entry average latency.


bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/AddCompletion.java
```diff
@@ -0,0 +1,130 @@
+package org.apache.bookkeeper.proto;
+
+import io.netty.util.Recycler;
+import org.apache.bookkeeper.client.BKException;
+import org.apache.bookkeeper.net.BookieId;
+import org.slf4j.MDC;
+
+class AddCompletion extends CompletionValue implements BookkeeperInternalCallbacks.WriteCallback {
```
merlimat, line 8: Refactoring by pulling the classes out masks the actual fix here. Can you, for this fix, leave the refactoring aside?

merlimat (approved):

PR #4613, opened 2025-06-05
[fix] Failed read entries after multiple decommissioning

poorbarcode:
### Motivation

**Background**
- `ReadOnlyLedgerHandle` opened without `doRecovery`, in other words, it has not been closed yet, its metadata in memory will be updated once modified.
- `ReadOnlyLedgerHandle` opened with `doRecovery`, in other words, it has been closed or it will be closed; its metadata in memory will never be updated

**Issue**
- **Background**: The ledger's metadata can also be modified by the auto-recovery component.
- There is a scenario in which a ledger handle always gets the error org.apache.bookkeeper.client.BKException$BKBookieHandleNotAvailableException: Bookie handle is not available` after multiple decommissions. The reproduction steps are as follows
  - Client service opens a read-only ledger handle, which has been closed.
  - All BKs that relate to the ledger have been decommissioned.
  - Auto recovery component moved the data into other BK instances who is alive.
  - The ledger handle in the client memory keeps connecting to the BKs who are in the original ensemble set, and the connection will always fail.
- You can reproduce the issue with the new test `testOpenedLedgerHandleStillWorkAfterDecommissioning`

### Changes

Let Bookie `LedgerHandle` always accept the metadata updates, no matter "doRecovery" or not.

<del>Provide a new API to open a read-only ledger handle with `keepUpdateMetadata`, the ledger's metadata in memory will be updated automatically after the auto-recovery component updates it.</del>


poorbarcode:
@codelipenghui 

> I'm curious about is there any cases that the ReadOnlyLedgerHandle cannot accept the updates from metadata server. As I understand, it should always accept the metadata updates no matter "doRecovery" or not. 

Changed the implementation.

poorbarcode:
rerun failure checks

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerOpenOp.java
```diff
@@ -175,8 +175,18 @@ private void openWithMetadata(Versioned<LedgerMetadata> versionedMetadata) {
...
+            // that solved happens as follows:
+            // 1. Client service open a readonly ledger handle, which has been closed.
+            // 2. All BKs that relates to the ledger have been decommissioned.
+            // 3. Auto recovery component moved the data into other BK instances who is alive.
+            // 4. The ledger handle in the client memory keeps connects to the BKs who in the original ensemble set,
+            //    and the connection will always fail.
+            // Therefore, whether the ledger handle is closed or not, it needs the watcher,
             lh = new ReadOnlyLedgerHandle(bk.getClientCtx(), ledgerId, versionedMetadata, digestType,
-                                          passwd, !doRecovery);
+                                          passwd, true);
```
zymap, line 189: When doing recovery, it is also a metadata update operation. So it doesn't allow watching the metadata update. Maybe there is a race condition risk if we always get the metadata updated from two places.

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerOpenOp.java
```diff
@@ -174,9 +195,21 @@ private void openWithMetadata(Versioned<LedgerMetadata> versionedMetadata) {
...
+            // 1. Client service open a readonly ledger handle, which has been closed.
+            // 2. All BKs that relates to the ledger have been decommissioned.
+            // 3. Auto recovery component moved the data into other BK instances who is alive.
+            // 4. The ledger handle in the client memory keeps connects to the BKs who in the original ensemble set,
+            //    and the connection will always fail.
+            // Therefore, if a user needs to the feature that update metadata automatically, he will set
+            // "keepUpdateMetadata" to "true",
             lh = new ReadOnlyLedgerHandle(bk.getClientCtx(), ledgerId, versionedMetadata, digestType,
-                                          passwd, !doRecovery);
+                                          passwd, !needRecovery);
```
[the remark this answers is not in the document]
poorbarcode: Ah, it's a bug. Has solved it

```java
- final boolean needRecovery = !doRecovery || !metadata.isClosed();
+ final boolean watchImmediately = !doRecovery || metadata.isClosed();
...
...
- new ReadOnlyLedgerHandle(...!doRecovery);
+ new ReadOnlyLedgerHandle(...watchImmediately);
```

poorbarcode:
rerun failure checks

StevenLuMT:
rerun failure checks

poorbarcode:
rerun failure checks

StevenLuMT (approved):
Good jobs

bookkeeper-server/src/main/java/org/apache/bookkeeper/client/LedgerOpenOp.java
```diff
@@ -174,9 +195,21 @@ private void openWithMetadata(Versioned<LedgerMetadata> versionedMetadata) {
         }
 
         // get the ledger metadata back
+        final boolean watchImmediately = !doRecovery || metadata.isClosed();
```
hangc0276, line 198: We changed the behavior here. When the replication worker opens ledgers without doRecovery, it may check all the ledgers in the BookKeeper cluster. What's more, most of the ledgers are in a closed state. It will register a lot of listeners and bring heavy pressure on Zookeeper.

poorbarcode:
rerun failure checks

poorbarcode:
rerun failure checks

poorbarcode:
rerun failure checks

poorbarcode:
rerun failure checks

poorbarcode:
rerun failure checks

poorbarcode:
rerun failure checks

PR #4617, opened 2025-06-11
Fix documentation terminology and method usage

liangyepianzhou:
### Motivation
1. Incorrectly uses "Enumerator" when referring to the Java standard interface `java.util.Enumeration`
2. Uses invalid method names (`hasNextElement()` instead of `hasMoreElements()`, `entry.getId()` instead of `entry.getEntryId()`)

**Impact**
1. Misleads developers searching for non-existent `Enumerator` APIs
2. Creates inconsistency with Java official documentation, reducing the professionalism of the document
3. Uses incorrect API methods that would cause compilation errors

### Changes
1. Terminology corrections:
 Replaced all `Enumerator` references with `Enumeration`
2. Method fixes:
   - Corrected `hasNextElement()` → `hasMoreElements()` in loop conditions
   - Fixed `entry.getId()` → `entry.getEntryId()` for entry ID access
   - Added missing `@Override` annotation in async callback

### DIFF
--- a/site3/website/docs/api/ledger-api.md
+++ b/site3/website/docs/api/ledger-api.md
@@ -163,13 +163,13 @@ long entryId = ledger.addEntry("Some entry data".getBytes());
 ## Reading entries from ledgers
 
 ```java
-Enumeration<LedgerEntry> entries = handle.readEntries(1, 99);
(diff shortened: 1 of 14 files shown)


StevenLuMT (approved):
Good jobs

PR #4645, opened 2025-07-30
add rate limit for gc.

thetumbled:
### Motivation

Each time the bookie gc is triggered, the read latency of zookeeper soars high to tens of seconds, threatening the stability of the cluster.
image
The unlimited zookeeper read operation in gc is responsible for the issue.
image

### Changes

Add config `gcRateLimit` to limit the read operation rate in gc.

> Otherwise:
> 
> 
> ---

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/ScanAndCompareGarbageCollector.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/ScanAndCompareGarbageCollector.java
@@ -36,6 +36,8 @@ import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
+
+import com.google.common.util.concurrent.RateLimiter;
 import lombok.Cleanup;
 import org.apache.bookkeeper.client.BKException;
 import org.apache.bookkeeper.client.api.LedgerMetadata;
@@ -84,6 +86,7 @@ public class ScanAndCompareGarbageCollector implements GarbageCollector {
     private int activeLedgerCounter;
     private StatsLogger statsLogger;
     private final int maxConcurrentRequests;
+    private final RateLimiter gcRateLimiter;
 
     public ScanAndCompareGarbageCollector(LedgerManager ledgerManager, CompactableLedgerStorage ledgerStorage,
             ServerConfiguration conf, StatsLogger statsLogger) throws IOException {
@@ -103,6 +106,7 @@ public class ScanAndCompareGarbageCollector implements GarbageCollector {
                 enableGcOverReplicatedLedger, gcOverReplicatedLedgerIntervalMillis, maxConcurrentRequests);
 
         verifyMetadataOnGc = conf.getVerifyMetadataOnGC();
+        this.gcRateLimiter = RateLimiter.create(conf.getGcRateLimit());
 
         this.activeLedgerCounter = 0;
     }
@@ -153,6 +157,7 @@ public class ScanAndCompareGarbageCollector implements GarbageCollector {
             Versioned<LedgerMetadata> metadata = null;
             while (!done) {
                 start = end + 1;
+                gcRateLimiter.tryAcquire();
                 if (ledgerRangeIterator.hasNext()) {
                     LedgerRange lRange = ledgerRangeIterator.next();
                     ledgersInMetadata = lRange.getLedgers();
@@ -175,6 +180,7 @@ public class ScanAndCompareGarbageCollector implements GarbageCollector {
                             metadata = null;
                             int rc = BKException.Code.OK;
                             try {
+                                gcRateLimiter.tryAcquire();
                                 metadata = result(ledgerManager.readLedgerMetadata(bkLid), zkOpTimeoutMs,
                                         TimeUnit.MILLISECONDS);
                             } catch (BKException | TimeoutException e) {
@@ -236,6 +242,7 @@ public class ScanAndCompareGarbageCollector implements GarbageCollector {
                 // check ledger ensembles before creating lock nodes.
                 // this is to reduce the number of lock node creations and deletions in ZK.
                 // the ensemble check is done again after the lock node is created.
+                gcRateLimiter.tryAcquire();
                 Versioned<LedgerMetadata> preCheckMetadata = ledgerManager.readLedgerMetadata(ledgerId).get();
                 if (!isNotBookieIncludedInLedgerEnsembles(preCheckMetadata)) {
                     latch.countDown();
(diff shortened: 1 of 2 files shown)


thetumbled:
PTAL, thanks. @StevenLuMT @hezhangjian @hangc0276 @nodece @zymap

StevenLuMT (changes requested):
I think adding a gcRateLimiter doesn't fundamentally solve the problem, nor is it practically feasible to implement or limit the number of calls.

I think it could be more practical to break down the rate limits into several categories: rate limits for GC calls to ZooKeeper, rate limits for GC file reads, rate limits for GC file writes, or rate limits for other GC resources.

thetumbled:
> I think adding a gcRateLimiter doesn't fundamentally solve the problem, nor is it practically feasible to implement or limit the number of calls.

Why? I believe adding a rate limiter and acquiring it prior to each zk operation would suffice.

> I think it could be more practical to break down the rate limits into several categories: rate limits for GC calls to ZooKeeper, rate limits for GC file reads, rate limits for GC file writes, or rate limits for other GC resources.

We haven't encountered any issues with GC file reads or writes. The only thing I need to limit is the rate of zk reads.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/ScanAndCompareGarbageCollector.java
```diff
@@ -153,6 +156,7 @@ public void gc(GarbageCleaner garbageCleaner) {
             Versioned<LedgerMetadata> metadata = null;
             while (!done) {
                 start = end + 1;
+                gcRateLimiter.tryAcquire();
```
hangc0276, line 159: You are not checking the `tryAcquire()` result, and `gcRateLimiter.tryAcquire();` changed nothing, so it cannot block the ZK requests.

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
```diff
@@ -114,6 +114,7 @@ public class ServerConfiguration extends AbstractConfiguration<ServerConfigurati
     protected static final String GC_OVERREPLICATED_LEDGER_WAIT_TIME = "gcOverreplicatedLedgerWaitTime";
     protected static final String GC_OVERREPLICATED_LEDGER_MAX_CONCURRENT_REQUESTS =
             "gcOverreplicatedLedgerMaxConcurrentRequests";
+    protected static final String GC_RATE_LIMIT = "gcRateLimit";
```
hangc0276, line 117: The configuration name is confusing.

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/ScanAndCompareGarbageCollector.java
thetumbled: You are right, thanks for pointing out this. I have updated the implementation.
PTAL again, thanks.

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
```diff
@@ -114,6 +114,7 @@ public class ServerConfiguration extends AbstractConfiguration<ServerConfigurati
     protected static final String GC_OVERREPLICATED_LEDGER_WAIT_TIME = "gcOverreplicatedLedgerWaitTime";
     protected static final String GC_OVERREPLICATED_LEDGER_MAX_CONCURRENT_REQUESTS =
             "gcOverreplicatedLedgerMaxConcurrentRequests";
+    protected static final String GC_ZK_OP_RATE_LIMIT = "gcZkOpRateLimit";
```
nodece, line 117: Have you considered using `gcMetadataOpRateLimit` instead? The underlying metadata store could be ZooKeeper, etcd, or something else, so using a more general name rather than one tied to a specific implementation is better.

nodece (approved):
Overall, LGTM.

StevenLuMT (approved):
LGTM: The previous definition was a bit broad. Now you have sunk to MetadataOpRateLimit, which I understand is still useful and configurable.

StevenLuMT:
By the way, is it possible to add a testcase? If it is possible, please add a testcase to verify the correctness of the function. Thank you @thetumbled

StevenLuMT:
rerun failure checks

thetumbled:
> By the way, is it possible to add a testcase? If it is possible, please add a testcase to verify the correctness of the function. Thank you @thetumbled

I will try to add a testcase later.

thetumbled:
> By the way, is it possible to add a testcase? If it is possible, please add a testcase to verify the correctness of the function. Thank you @thetumbled

A test case has been added to verify the correctness of the rate limit feature, PTAL, thanks.

thetumbled:
rerun failure checks

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
```diff
@@ -114,6 +114,7 @@ public class ServerConfiguration extends AbstractConfiguration<ServerConfigurati
     protected static final String GC_OVERREPLICATED_LEDGER_WAIT_TIME = "gcOverreplicatedLedgerWaitTime";
     protected static final String GC_OVERREPLICATED_LEDGER_MAX_CONCURRENT_REQUESTS =
             "gcOverreplicatedLedgerMaxConcurrentRequests";
+    protected static final String GC_METADATA_OP_RATE_LIMIT = "gcMetadataOpRateLimit";
```
lhotari, line 117: Since "gc" isn't used at all in the existing rate limiting settings, this raises the question of whether `gcMetadataOpRateLimit` should have a similar name such as `compactionMetadataOpRateLimit` to be consistent with the other concepts that are used ("compaction" instead of "gc").

lhotari (changes requested):
- Please check the configuration guidelines at: https://bookkeeper.apache.org/community/coding-guide/#configuration

- This configuration should also be added to the default `conf/bk_server.conf` and `site3/website/docs/reference/config.md`.

> * Names should be thought through from the point of view of the person using the config.
> * The default values should be thought as best value for people who runs the program without tuning parameters.
> * All configuration settings should be added to [default configuration file](https://github.com/apache/bookkeeper/blob/master/conf/bk_server.conf) and [documented](https://github.com/apache/bookkeeper/blob/master/site3/website/docs/reference/config.md).
 

- The new configuration should be cross-referenced with the existing "compaction" rate limiting settings since it will also impact the rate.

bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
thetumbled: This configuration affect all subtasks in gc, while compaction is just the last subtask in gc.
I think `gcMetadataOpRateLimit` is better than `compactionMetadataOpRateLimit`.

PR #4646, opened 2025-08-07
[fix]  Bookie Info lost by notification race condition. #20642

gaozhangmin:
According to pulsar's fix:   https://github.com/apache/pulsar/pull/20642
 race conditions problem is also exists in bk's [ZKRegistrationClient.java](https://github.com/apache/bookkeeper/compare/master...gaozhangmin:bookkeeper:fix_ZKRegistrationClient_1?expand=1#diff-2fd51027123dbb7985455f48cff7b47eaa1602d6d142eb7bcaec542fbc5301d5)

### DIFF
--- a/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/concurrent/FutureUtils.java
+++ b/bookkeeper-common/src/main/java/org/apache/bookkeeper/common/concurrent/FutureUtils.java
@@ -22,6 +22,7 @@ import com.google.common.base.Stopwatch;
 import com.google.common.collect.Lists;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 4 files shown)


gaozhangmin:
@StevenLuMT  
After switching the bookie from readonly to writable status online, the broker side continuously generated error logs: "Bookie Handle not available." The issue was resolved only after restarting the bookie. It was discovered that there was a bug in the ZkRegisterClient relied upon by the Pulsar broker, which caused the broker's bookie status cache update to be lost. Although the bookie was actually in a normal state, the broker failed to resolve the bookie's hostname due to the missing cache, resulting in the error: "Bookie Handle not available."

gaozhangmin:
rerun failure checks

gaozhangmin:
rerun failure checks

wenbingshen:
@gaozhangmin Is it different from this fix #4481

hangc0276:
@gaozhangmin could you rebase the master branch?

PR #4652, opened 2025-08-14
Refactor addDefaultRackBookiesIfMinNumRacksIsEnforced to Remove Redundant Logic

liangyepianzhou:
### Motivation

The current implementation of `addDefaultRackBookiesIfMinNumRacksIsEnforced` has the following issues:

- **Redundant logic**: The initialization and merging process for collecting Bookie nodes on the default rack involves unnecessary data copying and checks.
- **Incorrect logging**: The log claims to print only the defaultRack Bookies that are being excluded, while it actually prints all the nodes, including those already in `excludeBookies`. 
```java 
LOG.info("enforceMinNumRacksPerWriteQuorum is enabled, so Excluding bookies of defaultRack: {}", bookiesInDefaultRack); 
```

### Modification

- When creating `bookiesInDefaultRack`, do not copy elements from `excludeBookies`.
- Remove unnecessary null checks for `bookiesInDefaultRack`.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
@@ -372,7 +372,7 @@ public class RackawareEnsemblePlacementPolicyImpl extends TopologyAwareEnsembleP
                 }
                 bookiesInDefaultRack.add(((BookieNode) node).getAddr());
             } else {
-                LOG.error("found non-BookieNode: {} as leaf of defaultrack: {}", node, getDefaultRack());
+                LOG.error("found non-BookieNode: {} as leaf of defaultRack: {}", node, getDefaultRack());
             }
         }
         if ((bookiesInDefaultRack == null) || bookiesInDefaultRack.isEmpty()) {


bookkeeper-server/src/main/java/org/apache/bookkeeper/client/RackawareEnsemblePlacementPolicyImpl.java
```diff
@@ -364,16 +364,16 @@ protected Set<BookieId> addDefaultRackBookiesIfMinNumRacksIsEnforced(
             Set<BookieId> bookiesInDefaultRack = null;
             Set<Node> defaultRackLeaves = topology.getLeaves(getDefaultRack());
             for (Node node : defaultRackLeaves) {
-                if (node instanceof BookieNode) {
+                if (node instanceof BookieNode && ((BookieNode) node).getAddr() != null) {
```
StevenLuMT, line 367: When does node.getAddr() return null?
To prevent this, should I throw an error or use getDefaultRack?
Also, add a test case to verify this exception scenario.

eolivelli (changes requested):
which test cases are covering this change ?

liangyepianzhou:
> which test cases are covering this change ?

It seems that there was previously no test coverage for this method, so I have added some test cases for `addDefaultRackBookiesIfMinNumRacksIsEnforced`.

PR #4657, opened 2025-08-26
[fix]Wrong error code(-102) when opening a deleted ledger

poorbarcode:
### Motivation

**Background**

https://github.com/apache/bookkeeper/pull/4462 fixed the issue that the ledger can still be written even though it was deleted, and throws `LedgerFencedException` exception for the case, but the callers did not handle the error, leading to the Bookie server responding with a  `-102` error(wrong password) to the client side

**Issue**
```
Message: org.apache.bookkeeper.mledger.ManagedLedgerException: Attempted to access ledger using the wrong password error code: -102
Stacktrace:
org.apache.pulsar.broker.service.BrokerServiceException$PersistenceException: org.apache.bookkeeper.mledger.ManagedLedgerException: Attempted to access ledger using the wrong password error code: -102
        at org.apache.pulsar.broker.service.BrokerService$2.openLedgerFailed(BrokerService.java:1872)
        at org.apache.bookkeeper.mledger.impl.ManagedLedgerFactoryImpl.lambda$asyncOpen$10(ManagedLedgerFactoryImpl.java:430)
        at java.base/java.util.concurrent.CompletableFuture.uniExceptionally(CompletableFuture.java:990)
```

Actually, the error Attempted to access ledger using the wrong password error code: -102 is not caused by the password; it didn’t go to check the password
From the error stack, it throws [here](https://github.com/streamnative/bookkeeper/blob/cfea64f2c79d1395b8c103336c55ad77a23ba4b7/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/HandleFactoryImpl.java#L60). And outside wrap the exception to the EUA exception [here](https://github.com/streamnative/bookkeeper/blob/cfea64f2c79d1395b8c103336c55ad77a23ba4b7/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadEntryProcessor.java#L115). So we got the 102 at the client side. So actually, the ledger is fenced at the bk server.
```
org.apache.bookkeeper.bookie.BookieException$LedgerFencedException: null
    at org.apache.bookkeeper.bookie.BookieException.create(BookieException.java:57) ~[io.streamnative-bookkeeper-server-4.16.6.4.jar:4.16.6.4]
    at org.apache.bookkeeper.bookie.HandleFactoryImpl.getHandle(HandleFactoryImpl.java:60) ~[io.streamnative-bookkeeper-server-4.16.6.4.jar:4.16.6.4]
    at org.apache.bookkeeper.bookie.BookieImpl.fenceLedger(BookieImpl.java:1103) ~[io.streamnative-bookkeeper-server-4.16.6.4.jar:4.16.6.4]
    at org.apache.bookkeeper.proto.ReadEntryProcessor.processPacket(ReadEntryProcessor.java:81) ~[io.streamnative-bookkeeper-server-4.16.6.4.jar:4.16.6.4]
    at org.apache.bookkeeper.proto.PacketProcessorBase.run(PacketProcessorBase.java:202) ~[io.streamnative-bookkeeper-server-4.16.6.4.jar:4.16.6.4]
    at org.apache.bookkeeper.common.util.SingleThreadExecutor.safeRunTask(SingleThreadExecutor.java:128) ~[io.streamnative-bookkeeper-common-4.16.6.4.jar:4.16.6.4]
    at org.apache.bookkeeper.common.util.SingleThreadExecutor.run(SingleThreadExecutor.java:105) ~[io.streamnative-bookkeeper-common-4.16.6.4.jar:4.16.6.4]
    at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30) ~[io.netty-netty-common-4.1.121.Final.jar:4.1.121.Final]
    at java.lang.Thread.run(Thread.java:840) ~[?:?]
```

### Changes

- Instead of using `LedgerFencedException`, use a new exception `LedgerFencedAndDeletedException`
- Let all callers handle the new error

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieException.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieException.java
@@ -69,6 +69,8 @@ public abstract class BookieException extends Exception {
             return new MetadataStoreException();
         case Code.UnknownBookieIdException:
             return new UnknownBookieIdException();
+        case Code.LedgerFencedAndDeletedException:
+            return new LedgerFencedAndDeletedException();
         case Code.DataUnknownException:
             return new DataUnknownException();
         default:
@@ -95,6 +97,7 @@ public abstract class BookieException extends Exception {
         int CookieExistsException = -109;
         int EntryLogMetadataMapException = -110;
         int DataUnknownException = -111;
+        int LedgerFencedAndDeletedException = -112;
     }
 
     public int getCode() {
@@ -195,6 +198,15 @@ public abstract class BookieException extends Exception {
      */
     public static class LedgerFencedException extends BookieException {
         public LedgerFencedException() {
+            super(Code.LedgerFencedException);
+        }
+    }
+
+    /**
+     * Signals that a ledger has been fenced in a bookie. No more entries can be appended to that ledger.
+     */
+    public static class LedgerFencedAndDeletedException extends BookieException {
+        public LedgerFencedAndDeletedException() {
             super(Code.LedgerFencedException);
         }
     }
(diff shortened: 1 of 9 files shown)


poorbarcode:
rerun failure checks

hangc0276 (approved):

Issue #4655, opened 2025-08-20
A single disk being full should not affect the GC operations on other disks.


liangyepianzhou:
## Problem Description  
When a single ledger disk becomes full, `suspendMajorGC()`/`suspendMinorGC()` pauses garbage collection (GC) for **all disks**. However, a single disk failure should not affect GC operations on other disks.  

**Two scenarios:**  
- **Even Data Distribution**: Data is evenly distributed across ledger disks. When all disks are nearly full and one disk fills first.  
- **Uneven Data Distribution**: Individual disk usage spikes due to write skew or data cleanup anomalies.  

For the even Data Distribution:  
Disabling GC only on the full ledger disk first, then waiting for the next check (default `diskCheckInterval=10000` ms) to disable GC on other disks causes no harm.  

For uneven storage distribution:  
Disabling GC on all ledger disks affects normal operations of other disks.  

## Solution  

- solution 1: Add a new configuration to control whether to stop GC on other disks when any single disk becomes full.  
- solution 2: When `isReadOnlyModeOnAnyDiskFullEnabled == true`, stop GC on other disks. Otherwise, other disks should continue normal operations without GC suspension.  
- solution 3: When a single disk becomes full, only stop GC for that specific disk.

I prefer solution 2 and 3.
This isn't a new feature, but a bug fix. Fixing a bug shouldn't be considered a breaking change, as adding new configuration increases user costs.

PR #4661, opened 2025-08-30
Only stop Gc for the fill disk for DbLedgerStorage

liangyepianzhou:
Main Issue: https://github.com/apache/bookkeeper/issues/4655
## Problem Description  
When a single ledger disk becomes full, `suspendMajorGC()`/`suspendMinorGC()` pauses garbage collection (GC) for **all disks**. However, a single disk failure should not affect GC operations on other disks.  

**Two scenarios:**  
- **Even Data Distribution**: Data is evenly distributed across ledger disks. When all disks are nearly full and one disk fills first.  
- **Uneven Data Distribution**: Individual disk usage spikes due to write skew or data cleanup anomalies.  

For the even Data Distribution:  
Disabling GC only on the full ledger disk first, then waiting for the next check (default `diskCheckInterval=10000` ms) to disable GC on other disks causes no harm.  

For uneven storage distribution:  
Disabling GC on all ledger disks affects normal operations of other disks.  

## Solution  

- solution 1: Add a new configuration to control whether to stop GC on other disks when any single disk becomes full.  
- solution 2: When `isReadOnlyModeOnAnyDiskFullEnabled == true`, stop GC on other disks. Otherwise, other disks should continue normal operations without GC suspension.  
- **solution 3: When a single disk becomes full, only stop GC for that specific disk.(current fix)**

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
@@ -39,6 +39,7 @@ import java.util.Collections;
 import java.util.EnumSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.PrimitiveIterator.OfLong;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.ExecutorService;
@@ -149,6 +150,7 @@ public class SingleDirectoryDbLedgerStorage implements CompactableLedgerStorage 
 
     private final Counter flushExecutorTime;
     private final boolean singleLedgerDirs;
+    private final File currentFile;
 
     public SingleDirectoryDbLedgerStorage(ServerConfiguration conf, LedgerManager ledgerManager,
                                           LedgerDirsManager ledgerDirsManager, LedgerDirsManager indexDirsManager,
@@ -158,7 +160,7 @@ public class SingleDirectoryDbLedgerStorage implements CompactableLedgerStorage 
             throws IOException {
         checkArgument(ledgerDirsManager.getAllLedgerDirs().size() == 1,
                 "Db implementation only allows for one storage dir");
-
+        currentFile = ledgerDirsManager.getAllLedgerDirs().get(0);
         String ledgerBaseDir = ledgerDirsManager.getAllLedgerDirs().get(0).getPath();
         // indexBaseDir default use ledgerBaseDir
         String indexBaseDir = ledgerBaseDir;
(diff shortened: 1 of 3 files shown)


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
```diff
@@ -158,7 +160,7 @@ public SingleDirectoryDbLedgerStorage(ServerConfiguration conf, LedgerManager le
             throws IOException {
         checkArgument(ledgerDirsManager.getAllLedgerDirs().size() == 1,
                 "Db implementation only allows for one storage dir");
-
+        currentFile = ledgerDirsManager.getAllLedgerDirs().get(0);
```
eolivelli, line 163: cosmetic change to make the code clearer:

pass `currentFile` to getLedgerDirsListener() and make it a private field in the anonymous class (I also suggest to make it no more anonymous)

I think that we could also pass `gcThread` and make the inner class static, so that it is also easier to write unit tests to it (and I would add them)

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
```diff
@@ -1156,6 +1158,10 @@ private LedgerDirsManager.LedgerDirsListener getLedgerDirsListener() {
 
             @Override
             public void diskAlmostFull(File disk) {
+                if (!Objects.equals(disk.getPath(), currentFile.getPath())) {
```
eolivelli, line 1161: let's reduce code duplication and make this block a method

if (!isCurrentFile(disk)) {
 return;
}

bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/storage/ldb/SingleDirectoryDbLedgerStorage.java
```diff
@@ -1151,11 +1154,19 @@ public OfLong getListOfEntriesOfLedger(long ledgerId) throws IOException {
                 "getListOfEntriesOfLedger method is currently unsupported for SingleDirectoryDbLedgerStorage");
     }
 
-    private LedgerDirsManager.LedgerDirsListener getLedgerDirsListener() {
+    private LedgerDirsManager.LedgerDirsListener getLedgerDirsListener(String diskPath) {
```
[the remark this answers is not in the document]
liangyepianzhou: Changing LedgerDirsListener from an anonymous class to a member variable still doesn’t seem to identify which disk SingleDirectoryDbLedgerStorage is responsible for writing data to.  
We also need to modify the LedgerDirsListener interface to expose the disk address it’s in charge of, so that we can carry out testing.

By comparison, directly exposing the addresses of the ledger disk and the index disk might be a more straightforward way to indicate what SingleDirectoryDbLedgerStorage is responsible for.

If my understanding is off, please let me know.

StevenLuMT (approved):
Good jobs

PR #4665, opened 2025-09-14
[fix]compile error of the file native_io_jni.c for the env jdk11 & windows

poorbarcode:
### Motivation

https://github.com/apache/bookkeeper/actions/runs/17647901080/job/50151556235?pr=4613
```
[INFO] OUTPUT>native-io-jni/cpp/native_io_jni.c: In function 'Java_org_apache_bookkeeper_common_util_nativeio_NativeIOJni_fsync':
Error:  OUTPUT>native-io-jni/cpp/native_io_jni.c:168:21: error: passing argument 1 of 'fflush' makes pointer from integer without a cast [-Wint-conversion]
[INFO] OUTPUT>  168 |     int res = fsync(fd);
[INFO] OUTPUT>      |                     ^~
[INFO] OUTPUT>      |                     |
[INFO] OUTPUT>      |                     jint {aka long int}
[INFO] OUTPUT>native-io-jni/cpp/native_io_jni.c:33:26: note: in definition of macro 'fsync'
[INFO] OUTPUT>   33 | #define fsync(fd) fflush(fd)
[INFO] OUTPUT>      |                          ^~
[INFO] OUTPUT>In file included from C:\hostedtoolcache\windows\Java_Temurin-Hotspot_jdk\11.0.28-6\x64\include/jni.h:39,
[INFO] OUTPUT>                 from D:\a\bookkeeper\bookkeeper\native-io\target\nar\javah-include/org_apache_bookkeeper_common_util_nativeio_NativeIOJni.h:2,
[INFO] OUTPUT>                 from native-io-jni/cpp/native_io_jni.c:29:
[INFO] OUTPUT>C:/mingw64/x86_64-w64-mingw32/include/stdio.h:635:28: note: expected 'FILE *' {aka 'struct _iobuf *'} but argument is of type 'jint' {aka 'long int'}
[INFO] OUTPUT>  635 |   int __cdecl fflush(FILE *_File);
[INFO] OUTPUT>      |                      ~~~~~~^~~~~
[INFO] OUTPUT>native-io-jni/cpp/native_io_jni.c: In function 'Java_org_apache_bookkeeper_common_util_nativeio_NativeIOJni_pwrite':
[INFO] OUTPUT>native-io-jni/cpp/native_io_jni.c:261:26: warning: passing argument 2 of 'pwrite' discards 'const' qualifier from pointer target type [-Wdiscarded-qualifiers]
[...]

PR #4675, opened 2025-10-20
Flaky Test: Build with windows on JDK 11

liangyepianzhou:
### Motivation
Many PRs cannot be merged because of this test failure. Even after rerunning multiple times, they still fail due to the same error.

**This doesn't seem to be a flaky test, but rather a blocking test that always fails！**

`
Error:  OUTPUT>native-io-jni/cpp/native_io_jni.c:168:21: error: passing argument 1 of 'fflush' makes pointer from integer without a cast [-Wint-conversion]
`

https://github.com/apache/bookkeeper/actions/runs/18648851159/job/53162582666?pr=4674
https://github.com/apache/bookkeeper/actions/runs/18552216345/job/53161739919?pr=4673
https://github.com/apache/bookkeeper/actions/runs/18526985175/job/52800716265?pr=4672
https://github.com/apache/bookkeeper/actions/runs/18526392532/job/52798756494?pr=4653
### Changes

fflush requires a FILE*, but we're passing an int file descriptor.
   On Windows, the correct equivalent is _commit(fd), which takes an int fd.
   So, change the mapping of fsync from fflush to _commit.

### DIFF
--- a/native-io/src/main/native-io-jni/cpp/native_io_jni.c
+++ b/native-io/src/main/native-io-jni/cpp/native_io_jni.c
@@ -30,7 +30,12 @@
 
 #ifdef _WIN32
 
-#define fsync(fd) fflush(fd)
+/*  fflush requires a FILE*, but we're passing an int file descriptor.
+   On Windows, the correct equivalent is _commit(fd), which takes an int fd.
+   So, change the mapping of fsync from fflush to _commit. */
+#include <io.h>
+#define fsync(fd) _commit(fd)
+
 #define strerror_r(errno,buf,len) strerror_s(buf,len,errno)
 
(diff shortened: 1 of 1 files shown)


liangyepianzhou:
Due to the confusion caused by mixing manual reruns triggered by the committer and reruns triggered by comments,  reopen the PR

zymap (commented):
https://github.com/apache/bookkeeper/pull/4665/files

Here is the fix. You should check the platform first instead of changing all the platform to that.

liangyepianzhou:
Close this PR as the issue has already been resolved by https://github.com/apache/bookkeeper/pull/4665/files

PR #4683, opened 2025-11-14
Supports configuring TCP Keepalive related parameters in Bookie Client.

wolfstudy:
### Motivation
In a private environment, network connectivity between data centers is limited by firewall configuration. When the Broker accesses Bookkeeper infrequently, exceeding the firewall's default deactivation time (20 minutes), the firewall will actively disconnect the Broker → Bookkeeper connection. 
At this point, if a new production or consumption request is made, the Broker cannot safely access the Bookkeeper node again, causing production/consumption failures that cannot be automatically recovered from. The error message primarily indicates connection-level anomalies:
Clipboard_Screenshot_1763091013

The current Broker → Bookkeeper access chain is based on the Netty communication framework, and the keepalive function at the TCP connection layer reuses the system default SO_KEEPALIVE. The code is `PerChannelBookieClient` → `connect()`, as follows:
Clipboard_Screenshot_1763091110

If no policy is explicitly set, the "System Default" option in the table follow will be used directly:

TCP_KEEPIDLE | 7200s
-- | --
TCP_KEEPINTVL | 75s
TCP_KEEPCNT | 9

### Changes

Include the following three configuration items in ClientConfiguration:

```
    public static final String TCP_KEEPIDLE = "tcpKeepIdle";
    public static final String TCP_KEEPINTVL = "tcpKeepIntvl";
    public static final String TCP_KEEPCNT = "tcpKeepCnt";
```

To maintain compatibility, the system default configuration will still be used by default.

### DIFF
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/proto/TestPerChannelBookieClientTcpKeepalive.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/proto/TestPerChannelBookieClientTcpKeepalive.java
@@ -74,39 +74,39 @@ public class TestPerChannelBookieClientTcpKeepalive {
     public void testTcpKeepaliveConditionalLogic() {
         // Test various configuration scenarios
         ClientConfiguration conf = new ClientConfiguration();
-        
+
         // Test with positive values
         conf.setTcpKeepIdle(60);
         conf.setTcpKeepIntvl(30);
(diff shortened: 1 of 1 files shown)


liudezhi2098 (approved):

StevenLuMT (approved):
Good jobs

StevenLuMT:
rerun failure checks

PR #4686, opened 2025-11-20
feat: support `equalsAndHashcode` for BookieServerInfo

mattisonchao:
### Motivation

The `BookieServerInfo` class is exposed to the metadata store plugin. We need to support `equalsAndHashcode` to help other plugins encode or decode it with comparison. 

FYI: https://github.com/apache/pulsar/pull/25001

### Changes

- Support `equalsAndHashcode` method for `BookieServerInfo` and underlying `Endpoint`


mattisonchao:
rerun failure checks

PR #4709, opened 2026-01-29
#2974: Add hash-based journal selection for improved ledger distribution

jprieto-temporal:
Main Issue: #2974

### Motivation

When multiple journals are configured, BookKeeper selects a journal using ledgerId % numJournals. This works well when
ledger IDs are random or sequential, but produces poor distribution when ledger IDs follow strided patterns (e.g., when
multiple clients allocate IDs from disjoint ranges). Uneven journal utilization can cause hotspots and reduce write
throughput.

### Changes

- Add journalHashBasedSelection configuration option (default: false)
- When enabled, use Fibonacci hashing on ledger IDs before journal selection
- The hash function multiplies by the golden ratio constant and folds high bits into low bits to break predictable
patterns
- Add integration test verifying read/write correctness with hash-based selection

### DIFF
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/JournalIndexComputationTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/bookie/JournalIndexComputationTest.java
@@ -65,8 +65,8 @@ public class JournalIndexComputationTest {
             {4, 3, "4 journals, stride=3"},
             {4, 4, "4 journals, stride=4 (worst case without hash)"},
             {8, 8, "8 journals, stride=8 (worst case without hash)"},
-            {4, 1024, "4 journals, stride=1024"},
-            {8, 1024, "8 journals, stride=1024"},
+            {4, 1024, "4 journals, stride=1024 (worst case without hash)"},
+            {8, 1024, "8 journals, stride=1024 (worst case without hash)"},
         });
     }
 


bookkeeper-server/src/main/java/org/apache/bookkeeper/bookie/BookieImpl.java
```diff
@@ -940,8 +942,30 @@ LedgerDescriptor getLedgerForEntry(ByteBuf entry, final byte[] masterKey)
...
+     *
+     * <p>When hash-based selection is enabled, uses Fibonacci hashing to distribute ledgers
+     * across journals. The xor with the right-shifted value folds the high 32 bits into
+     * the low 32 bits, adding more mixing prior to the modulo.
+     */
     private Journal getJournal(long ledgerId) {
-        return journals.get(MathUtils.signSafeMod(ledgerId, journals.size()));
+        long index = ledgerId;
+        if (journalHashBasedSelection) {
+            index = ledgerId * FIBONACCI_HASH_CONSTANT;
```
eolivelli: what about putting this block into a static method so that we can add unit tests ?

The function gets the index and the number of journals and computes the journal index

jprieto-temporal:
@dlg99 not opposed to factoring it out and using the same hashing that is used in this PR since it provides better distribution in the general case, but I think that could be done in a separate PR.

PR #4722, opened 2026-03-10
Improve CI: add caching for Bookkeeper old release downloads and improve maven caching and error logs

lhotari:
### Motivation

CI currently fails and it's hard to investigate the issue since surefire logs aren't uploaded and the threaddump isn't captured in backward compat test timeouts.
The problem appeared to be very slow progress of old Bookkeeper release downloads from archive.apache.org. 
Maven caching wasn't usable for PR builds since the Maven cache wasn't updated as part of the daily builds. PR builds don't have write access to apache/bookkeeper cache.

### Changes

- fix condition for uploading surefire logs
- add threaddump for backward compat tests
- adds a cache for BK release downloads from archive.apache.org. In CI, GitHub Actions cache is used to persist the cache. In local development, a directory is used to cache the downloads. 
- Fix maven dependency caching by adding caching to the daily build

### DIFF
--- a/.github/workflows/bk-ci.yml
+++ b/.github/workflows/bk-ci.yml
@@ -251,6 +251,14 @@ jobs:
           networkaddress.cache.negative.ttl=1
           EOF
 
+      - name: Restore released versions cache
+        uses: actions/cache/restore@v4
+        with:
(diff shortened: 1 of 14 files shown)


eolivelli (approved):
+1

PR #4730, opened 2026-03-13
Fix read thread blocking in sendResponseAndWait causing READ_ENTRY_REQUEST p99 latency spike

hangc0276:
### Motivation

  During performance testing, we observed that bookkeeper_server_READ_ENTRY p99 latency is less than 100ms, but bookkeeper_server_READ_ENTRY_REQUEST p99 latency spikes to 3 seconds. The two metrics
  differ in scope:

  - READ_ENTRY measures the storage read time only (from before readData() to after it completes)
  - READ_ENTRY_REQUEST measures from request enqueue time (enqueueNanos) through response write

  The root cause is in PacketProcessorBase.sendResponseAndWait(). When readWorkerThreadsThrottlingEnabled=true (the default), read responses go through sendResponseAndWait, which calls future.get() to
  block the read thread pool thread until the Netty channel write completes.

  Under heavy read load, this causes a cascading failure:

  1. future.get() blocks each read thread waiting for Netty write completion
  2. If writes are slow (client backpressure, network congestion, event loop busy), threads pile up blocked
  3. The read thread pool becomes saturated — all threads blocked waiting for writes
  4. New read requests queue up, and their scheduling delay grows
  5. READ_ENTRY_REQUEST (measured from enqueueNanos) spikes, even for requests to healthy channels

  One slow consumer can block a thread pool thread, reducing capacity for all other channels — a classic head-of-line blocking problem.

### Changes
 **1. Non-blocking sendResponseAndWait (PacketProcessorBase.java)**

Replace the blocking future.get() in sendResponseAndWait() with a non-blocking ChannelFutureListener. The key design decisions:

  - Read thread freed immediately: After writeAndFlush, the thread returns to the pool and can process other requests instead of blocking on future.get()
  - Throttling semantics preserved: onReadRequestFinish() (which releases the read semaphore) is moved into the ChannelFutureListener, so it is only called after the write completes. This ensures the
  read concurrency limit still gates on write completion, without wasting a thread to enforce it.
  - Captured local variables: Since the processor may be recycled after sendResponseAndWait returns, enqueueNanos and requestProcessor are captured as local variables before the async callback.
  - Semaphore always released: Both the success and failure paths in the listener call onReadRequestFinish(), preventing semaphore leaks on write failure. (The previous code had a subtle issue: on
  ExecutionException/InterruptedException, sendResponseAndWait returned early without recording the metric, though onReadRequestFinish was still called by the caller.)

  **2. Enable maxReadsInProgressLimit by default (ServerConfiguration.java)**

  Changed the default of maxReadsInProgressLimit from 0 (unlimited) to 10000.

  With the old blocking sendResponseAndWait, the read thread pool (default 8 threads) acted as an implicit concurrency bound — each thread blocked on one write at a time, so at most 8 responses could be
  buffered. With the new non-blocking approach, threads are freed immediately and keep processing reads. Without a bound, a slow consumer could cause unbounded response buffer growth in Netty's write
  queue, leading to OOM.

  The maxReadsInProgressLimit semaphore now serves as the explicit bound. Since onReadRequestFinish() is called in the ChannelFutureListener (after the write completes), the semaphore precisely counts
  reads whose response data is in memory but not yet flushed. The default of 10,000 provides good throughput (~40MB at 4KB entries) while preventing unbounded memory growth. Users with larger entries
  should tune this lower (memoryBudget / avgEntrySize).

### Test Plan
  Added 5 new unit tests to ReadEntryProcessorTest:

  - testThrottledReadNonBlockingOnSuccess — verifies that with throttle=true, run() returns immediately and onReadRequestFinish() is deferred until the write future completes
  - testThrottledReadNonBlockingOnWriteFailure — verifies onReadRequestFinish() is still called when the write fails, ensuring the read semaphore is always released
  - testNonThrottledReadCallsOnFinishSynchronously — verifies that with throttle=false, onReadRequestFinish() is called synchronously
  - testDefaultMaxReadsInProgressLimitIsEnabled — verifies maxReadsInProgressLimit defaults to 10000
  - testThrottledReadHoldsSemaphoreUntilWriteCompletes — uses a real Semaphore(1) wired to onReadRequestStart/onReadRequestFinish to verify the full lifecycle: semaphore acquired at request creation,
  held through non-blocking run() return, blocks a second read while write is in progress, and released only when the write future completes

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java
@@ -1075,10 +1075,16 @@ public class ServerConfiguration extends AbstractConfiguration<ServerConfigurati
     /**
      * Get max number of reads in progress. 0 == unlimited.
      *
+     * <p>This limit bounds the memory used by read responses that have been read from storage
+     * but not yet flushed to the network. Since read response writes are non-blocking,
+     * without this limit a slow consumer could cause unbounded memory growth.
+     * The default value of 10000 provides a reasonable balance between throughput and memory usage.
+     * Tune based on your average entry size: memoryBudget / avgEntrySize.
+     *
      * @return Max number of reads in progress.
      */
     public int getMaxReadsInProgressLimit() {
-        return this.getInt(MAX_READS_IN_PROGRESS_LIMIT, 0);
+        return this.getInt(MAX_READS_IN_PROGRESS_LIMIT, 10000);
     }
 
     /**
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PacketProcessorBase.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PacketProcessorBase.java
@@ -19,8 +19,8 @@ package org.apache.bookkeeper.proto;
 
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelFuture;
+import io.netty.channel.ChannelFutureListener;
 import io.netty.channel.ChannelPromise;
-import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import org.apache.bookkeeper.common.util.MathUtils;
 import org.apache.bookkeeper.proto.BookieProtocol.Request;
@@ -74,10 +74,12 @@ abstract class PacketProcessorBase<T extends Request> implements Runnable {
     protected void sendReadReqResponse(int rc, Object response, OpStatsLogger statsLogger, boolean throttle) {
         if (throttle) {
             sendResponseAndWait(rc, response, statsLogger);
+            // onReadRequestFinish is called asynchronously in the ChannelFutureListener
+            // inside sendResponseAndWait to maintain throttling without blocking the thread.
         } else {
             sendResponse(rc, response, statsLogger);
+            requestProcessor.onReadRequestFinish();
         }
-        requestProcessor.onReadRequestFinish();
     }
 
     protected void sendResponse(int rc, Object response, OpStatsLogger statsLogger) {
@@ -150,27 +152,44 @@ abstract class PacketProcessorBase<T extends Request> implements Runnable {
     }
 
     /**
-     * Write on the channel and wait until the write is completed.
+     * Write on the channel and notify completion via a listener.
      *
-     * <p>That will make the thread to get blocked until we're able to
-     * write everything on the TCP stack, providing auto-throttling
-     * and avoiding using too much memory when handling read-requests.
+     * <p>This provides auto-throttling by holding the read semaphore until the write completes,
+     * without blocking the read thread pool thread. The read thread is freed immediately to
+     * process other requests, while the semaphore prevents unbounded read concurrency.
      */
     protected void sendResponseAndWait(int rc, Object response, OpStatsLogger statsLogger) {
+        // Capture fields before the processor may be recycled after this method returns.
+        final long capturedEnqueueNanos = this.enqueueNanos;
+        final BookieRequestProcessor processor = this.requestProcessor;
         try {
             Channel channel = requestHandler.ctx().channel();
             ChannelFuture future = channel.writeAndFlush(response);
-            if (!channel.eventLoop().inEventLoop()) {
-                future.get();
+            future.addListener((ChannelFutureListener) f -> {
+                if (!f.isSuccess() && logger.isDebugEnabled()) {
+                    logger.debug("Netty channel write exception. ", f.cause());
+                }
+                if (BookieProtocol.EOK == rc) {
+                    statsLogger.registerSuccessfulEvent(
+                            MathUtils.elapsedNanos(capturedEnqueueNanos), TimeUnit.NANOSECONDS);
+                } else {
+                    statsLogger.registerFailedEvent(
+                            MathUtils.elapsedNanos(capturedEnqueueNanos), TimeUnit.NANOSECONDS);
+                }
+                processor.onReadRequestFinish();
+            });
+        } catch (Exception e) {
+            if (logger.isDebugEnabled()) {
+                logger.debug("Netty channel write exception. ", e);
             }
-        } catch (ExecutionException | InterruptedException e) {
-            logger.debug("Netty channel write exception. ", e);
-            return;
-        }
-        if (BookieProtocol.EOK == rc) {
-            statsLogger.registerSuccessfulEvent(MathUtils.elapsedNanos(enqueueNanos), TimeUnit.NANOSECONDS);
-        } else {
-            statsLogger.registerFailedEvent(MathUtils.elapsedNanos(enqueueNanos), TimeUnit.NANOSECONDS);
+            if (BookieProtocol.EOK == rc) {
+                statsLogger.registerSuccessfulEvent(
+                        MathUtils.elapsedNanos(capturedEnqueueNanos), TimeUnit.NANOSECONDS);
+            } else {
+                statsLogger.registerFailedEvent(
+                        MathUtils.elapsedNanos(capturedEnqueueNanos), TimeUnit.NANOSECONDS);
+            }
+            processor.onReadRequestFinish();
         }
     }
 
--- a/bookkeeper-server/src/test/java/org/apache/bookkeeper/proto/ReadEntryProcessorTest.java
+++ b/bookkeeper-server/src/test/java/org/apache/bookkeeper/proto/ReadEntryProcessorTest.java
@@ -19,6 +19,7 @@
 package org.apache.bookkeeper.proto;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.anyLong;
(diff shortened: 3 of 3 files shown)


merlimat (approved):

wenbingshen (approved):
LGTM. The change to `maxReadsInProgressLimit` needs to be added to the release documentation in the future.

lhotari (approved):

horizonzy (approved):
Nice catch

zymap (approved):

lhotari:
**Together with the `maxReadsInProgressLimit` default change from 0 → 10000 in this same PR, this change can permanently deadlock the entire read path of a bookie.** The root cause is an interaction between this change and the server-side backpressure feature from #1410 (extended to the v2 protocol by #3324).

### Background: the #1410 admission design blocks a Netty event loop thread

`maxReadsInProgressLimit` is enforced in [`BookieRequestProcessor.onReadRequestStart`](https://github.com/apache/bookkeeper/blob/4a012a2a1e984333c65834caddb1639158ce6bac/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java#L236-L254):

```java
if (!readsSemaphore.tryAcquire()) {
    ...
    channel.config().setAutoRead(false);
    ...
    readsSemaphore.acquireUninterruptibly();   // blocks the calling thread
    channel.config().setAutoRead(true);
    ...
}
```

It is called from the read processor constructors/factories ([`ReadEntryProcessor.create`](https://github.com/apache/bookkeeper/blob/4a012a2a1e984333c65834caddb1639158ce6bac/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadEntryProcessor.java#L54), [`ReadEntryProcessorV3` constructor](https://github.com/apache/bookkeeper/blob/4a012a2a1e984333c65834caddb1639158ce6bac/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadEntryProcessorV3.java#L57)), which [`processReadRequest`](https://github.com/apache/bookkeeper/blob/4a012a2a1e984333c65834caddb1639158ce6bac/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/BookieRequestProcessor.java#L676-L699) runs on the Netty IO thread *before* submitting the request to the read thread pool. When `acquireUninterruptibly()` blocks, an entire event loop is parked. Every channel registered on that loop is affected: no requests are decoded (reads or writes, v2 or v3), no responses are flushed, no write futures complete, no future listeners run, and no channel closes are processed.

That is the latent hazard in this design: **releasing a permit requires event-loop progress (flushing a response), while acquiring a permit can block the event loop.** It stayed latent for years because `maxReadsInProgressLimit` defaulted to 0, so the semaphore was never created.

### What changed in this PR

Two things:

1. The [`getMaxReadsInProgressLimit()`](https://github.com/apache/bookkeeper/blob/4a012a2a1e984333c65834caddb1639158ce6bac/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java#L1089-L1091) default changed from 0 to 10000, arming the semaphore and the autoRead gate on every deployment. Since [`readWorkerThreadsThrottlingEnabled`](https://github.com/apache/bookkeeper/blob/4a012a2a1e984333c65834caddb1639158ce6bac/bookkeeper-server/src/main/java/org/apache/bookkeeper/conf/ServerConfiguration.java#L2135-L2137) also defaults to `true`, the throttled response path is active by default as well.
2. The permit is now released only from the `writeAndFlush` future listener in [`sendResponseAndWait`](https://github.com/apache/bookkeeper/blob/4a012a2a1e984333c65834caddb1639158ce6bac/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PacketProcessorBase.java#L155-L186):

```java
ChannelFuture future = channel.writeAndFlush(response);
future.addListener((ChannelFutureListener) f -> {
    ...
    processor.onReadRequestFinish();   // the only release
});
```

Unlike the sibling [`sendResponse()`](https://github.com/apache/bookkeeper/blob/4a012a2a1e984333c65834caddb1639158ce6bac/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PacketProcessorBase.java#L84-L110) used by the non-throttled path — which handles non-writable channels (wait up to `waitTimeoutOnBackpressureMillis`, blacklist, or drop the response) and always lets [`sendReadReqResponse`](https://github.com/apache/bookkeeper/blob/4a012a2a1e984333c65834caddb1639158ce6bac/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/PacketProcessorBase.java#L73-L82) release the permit synchronously — `sendResponseAndWait` has no non-writable-channel handling. If the remote peer stops draining responses (TCP zero window), `writeAndFlush` parks the message in the `ChannelOutboundBuffer` and the returned future can stay pending indefinitely: the listener never fires, and the permit is never released.

Before this PR, the blocking `future.get()` capped un-flushed responses at `numReadWorkerThreads`. Now the number of in-flight un-flushed responses is bounded only by the semaphore itself — whose release depends on those very flushes completing.

### The deadlock

1. The semaphore exhausts under load, by either route: **(a) slow reads** — permits are acquired at ingest, before the request is submitted to the read thread pool, so the executor backlog alone reaches the limit. With defaults the semaphore is the binding constraint: 10000 permits vs. `numReadWorkerThreads` (8) × `maxPendingReadRequestPerThread` (10000) = 80000 queue slots, so it trips long before `ETOOMANYREQUESTS` load shedding would. **(b) slow clients** — permits are held by responses stuck in the outbound buffers of open-but-non-writable channels, whose write futures never complete.
2. The next read arriving on each event loop fails `tryAcquire()` and parks that loop in `acquireUninterruptibly()`.
3. Read worker threads keep completing reads, but `writeAndFlush` from an off-loop thread only enqueues a flush task on the response channel's (parked) event loop. Neither route can release a permit anymore: flush tasks are frozen on parked loops, pending futures never complete, and even client disconnects can no longer be processed.
4. Once every IO thread has parked, no permit can ever be released, and the entire request path stalls — including writes, since parked loops can't decode any request type. `READ_ENTRY_IN_PROGRESS` stays pinned at `maxReadsInProgressLimit` with read throughput at zero until the bookie is restarted, and the bookie can still *appear* healthy: the HTTP admin/metrics endpoints are served from separate threads and don't exercise the data path.

Reproducing should be straightforward: set `maxReadsInProgressLimit` to a small value, connect a v2-protocol client, issue more concurrent reads than the limit, and have the client stop reading from its socket. A sustained cache-missing read backlog reaches the limit the same way without any client misbehavior.

### Why this change doesn't achieve its own goal either

The aim of this PR was to stop read worker threads from blocking in `future.get()` (the p99 spike). But with the semaphore now on by default, the blocking moves somewhere worse: under permit exhaustion the Netty IO threads block, stalling every channel on their event loops and making permit release impossible. A latency spike is traded for a permanent, unrecoverable read-path deadlock. Note that the throttle's own designed trigger condition — a sustained read backlog — is by itself sufficient to set up the deadlock; no client misbehavior is required.

Scope notes:

- The permit leak is specific to the v2 wire protocol (the `sendReadReqResponse` throttle branch). The v3 path releases the permit synchronously after a non-blocking write ([`ReadEntryProcessorV3.sendResponse`](https://github.com/apache/bookkeeper/blob/4a012a2a1e984333c65834caddb1639158ce6bac/bookkeeper-server/src/main/java/org/apache/bookkeeper/proto/ReadEntryProcessorV3.java#L339-L346)) and cannot leak this way — a pure-v3 overload parks event loops only transiently and recovers, because v3 releases don't depend on event-loop progress. However, the semaphore and the blocking admission are shared: once v2 permits are stuck, v3 reads also park event loops, and every channel on a parked loop stalls regardless of protocol or operation type. Note that Pulsar brokers use the v2 protocol by default (`bookkeeperUseV2WireProtocol=true`).
- Affected: master, released 4.18.0, and branch-4.17 — the pending 4.17.4 release would ship this.

### Mitigations and fix direction

Operators can break the loop with either `maxReadsInProgressLimit=0` (the previous default) or `readWorkerThreadsThrottlingEnabled=false` (routes responses through `sendResponse()`, which always releases the permit).

For a proper fix, I think two things are needed:

1. **Never block an event loop on the semaphore.** Minimal version: move the acquire off the Netty thread, e.g. into the processor's execution on the read worker (keeping acquire/release strictly paired — the current `RejectedExecutionException` path releases a permit acquired at creation). Better: make admission asynchronous — on `tryAcquire` failure, disable autoRead and queue the request with a timeout, replying `ETOOMANYREQUESTS` on timeout, so no thread ever blocks. Pulsar has a non-blocking semaphore implementation that fits this shape: [`AsyncSemaphore`](https://github.com/apache/pulsar/blob/master/pulsar-common/src/main/java/org/apache/pulsar/common/semaphore/AsyncSemaphore.java) / [`AsyncSemaphoreImpl`](https://github.com/apache/pulsar/blob/master/pulsar-common/src/main/java/org/apache/pulsar/common/semaphore/AsyncSemaphoreImpl.java) — `CompletableFuture`-based acquire with a timeout, a bounded waiter queue, a cancellation hook (e.g. for closed channels), and idempotent permit release, which makes it safe to wire the release into both the write-future listener and a channel-close listener. Being ASF/Apache-2.0 code, it could be adapted into BookKeeper.
2. **Guarantee `onReadRequestFinish()` runs exactly once even if the write future never completes** — non-writable-channel handling like `sendResponse()` has, and/or a channel-close/timeout-bound release.

Until both are in place, I'd suggest reverting the `maxReadsInProgressLimit` default to 0 — in particular for the pending 4.17.4 release.

lhotari:
PR to set `maxReadsInProgressLimit` to `0` by default: https://github.com/apache/bookkeeper/pull/4827
CI is currently broken, PR to fix it: https://github.com/apache/bookkeeper/pull/4828

PR #2457, opened 2020-10-26
BP-42: New Client API - list ledgers

nicoloboschi:
Implementation of BP-42

### Motivation
Enhance new client API, added methods for
- listing ledgers
- access to ledger metadata

### Changes

- Added new interfaces in "api" client package for ledgers listing
- Added new method `getLedgerMetadata(long ledgerId)` to BookKeeper interface
- Added `getLedgerId()` to LedgerMetadata type
-- Ledger metadata ser/deser are not changed since ledgerId is not stored inside ZK data node 

Updated all tests using `LedgerMetadataBuilder` because ledgerId is now required in order to call `build()` method

Master Issue: #2422

PR #4731, opened 2026-03-17
Issue 2422: Fix SyncLedgerIterator.hasNext() failing to iterate across ZK ledger ranges

### LINKED ISSUE #2422: BP-42: New Client API - list ledgers
This is the master ticket for tracking BP-42:

The new Client API (org.apache.bookkeeper.client.api.BookKeeper) aims to replace obsolete BookKeeper API but some features are not implemented yet, like the functionalities provided by BookKeeperAdmin. 
For example, it does not expose a method to list available ledgers, comparable to BookKeeperAdmin#listLedgers().
low.
We could improving it adding:
1) Ledgers listing 
2) Simpler access to LedgerMetadata
3) Ledger id inside LedgerMetadata

geniusjoe:
Fix #2422

Related PR: #2457

### Motivation

The `SyncLedgerIterator.hasNext()` method in `BookKeeper.java` has a logic bug that causes it to stop iterating prematurely when ledgers span across multiple ZooKeeper ranges.

ZooKeeper stores ledgers in a hierarchical tree structure (e.g. `/ledgers/00/0000/...`). Each range (znode) holds at most 10,000 leaf-level ledger IDs.
The 10,000 limit is determined by the 4-digit last level (0000-9999) as defined in `LegacyHierarchicalLedgerManager` (2-4-4 split) and `LongHierarchicalLedgerManager` (3-4-4-4-4 split). When the number of ledgers exceeds 10,000, they are distributed across multiple ranges.
image

The original `hasNext()` implementation had the following logic:
```Java
@Override
public boolean hasNext() throws IOException {
    parent.checkClosed();
    if (currentRange != null) {
        if (currentRange.hasNext()) {
            return true;
        }
    } else if (iterator.hasNext()) {
        return true;
    }
    return false;
}
```
When `currentRange` is non-null but exhausted (no more elements), it falls through to `return false` **without** checking `iterator.hasNext()` for more ranges.

### Changes

1. Fix `SyncLedgerIterator.hasNext()` in `BookKeeper.java`:
Simplified the conditional logic to properly fall through to `iterator.hasNext()` when the current range is exhausted
2. Add `testListLedgersLargeScale()` test in `LedgerMetadataTest.java`
Added a new test case with `numOfLedgers = 10010` (> 10,000 per-range limit) to verify cross-range iteration works correctly.

### DIFF
--- a/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeper.java
+++ b/bookkeeper-server/src/main/java/org/apache/bookkeeper/client/BookKeeper.java
@@ -1583,17 +1583,20 @@ public class BookKeeper implements org.apache.bookkeeper.client.api.BookKeeper {
             this.parent = parent;
         }
 
-        @Override
+        /**
+         * ZooKeeper stores ledgers in a hierarchical tree (e.g. /ledgers/00/0000/...).
+         * {@code iterator} (LedgerRangeIterator) traverses the intermediate tree nodes (ranges/buckets),
+         * while {@code currentRange} iterates over the leaf-level ledger IDs within a single range.
+         *
+         * Therefore, when {@code currentRange} has no more leaf nodes, we need to check
+         * {@code iterator.hasNext()} to determine if there are more ranges to advance to.
+         */
         public boolean hasNext() throws IOException {
             parent.checkClosed();
-            if (currentRange != null) {
-                if (currentRange.hasNext()) {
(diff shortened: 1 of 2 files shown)


hanmz (approved):
lgtm

lhotari (approved):
LGTM, good catch

wenbingshen (approved):
LGTM

